PR #5, opened 2017-08-15
Add clangd to default settings

rwols:
I've commented out the mscpp server because its path is hardcoded. I'm also unsure how the plugin responds to multiple servers being able to handle the same language. How would that work?

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -1,36 +1,48 @@
 {
-   "clients": {
-   		"pyls": {
-   			"command": ["pyls"],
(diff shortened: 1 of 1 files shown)


tomv564:
Thanks for your contribution!

I may have resolved the configuration conflict directly on your master branch, this is probably poor style :)

I removed the mscpp config, it can be documented if/when someone gets it working!

rwols:
> I may have resolved the configuration conflict directly on your master branch, this is probably poor style :)

No problem :-)

PR #17, opened 2017-08-18
Improve highlighting for diagnostics

rwols:
Somewhat more structured syntax for diagnostics:

schermafbeelding 2017-08-18 om 16 16 43

You could consider highlighting the Exxx and Dxxx numbers at this point, but that seems to be too python-specific. I'm wondering if a per-server syntax would be beneficial.

### DIFF
--- a/Syntaxes/Diagnostics.sublime-syntax
+++ b/Syntaxes/Diagnostics.sublime-syntax
@@ -1,7 +1,13 @@
 %YAML 1.2
 ---
+# [Subl]: https://www.sublimetext.com/docs/3/syntax.html
+# [LSP]: https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md
 hidden: true
 scope: output.lsp.diagnostics
+
+variables:
+  start_of_diag_body: ^\s+(?=\d)
+  filename_and_colon: ^(.*)(:)$
 
 contexts:
   main:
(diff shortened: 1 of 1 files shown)


tomv564:
This looks great, I'll give this a quick test tonight and merge it!

Syntaxes/Diagnostics.sublime-syntax
```diff
@@ -1,21 +1,49 @@
...
+              push:
+                - include: pop-at-end
+                - match: \berror\b
+                  scope: markup.deleted.lsp markup.error.lsp
+                  push: after-diagnostic-type-token
+                - match: \bwarning\b
+                  scope: markup.heading.lsp markup.warning.lsp
+                  push: after-diagnostic-type-token
+                - match: \binfo\b
+                  scope: markup.inserted.lsp markup.info.lsp
```
deathaxe: Why do you use markup.inserted, markup.heading, markup.deleted besides the others?

Are markup.error, markup.warning, markup.info not available in all color schemes?

Syntaxes/Diagnostics.sublime-syntax
```diff
@@ -1,21 +1,49 @@
...
+            3: constant.numeric.integer.decimal.lsp
+          push:
+            - include: pop-at-end
+            - match: '[^\s]+'
+              scope: comment.line.lsp
+              push:
+                - include: pop-at-end
+                - match: \berror\b
+                  scope: markup.deleted.lsp markup.error.lsp
+                  push: after-diagnostic-type-token
```
deathaxe: `push` might be replaced by `set` here.

rwols: Yep, i have to fix that

Syntaxes/Diagnostics.sublime-syntax
rwols: Yes, those scopes are basically unknown by everyone, right? We made an RFC about it in the packages repo, but that's about it. This is my sneaky way to spread it out into the wild.

tomv564: Would be awesome to spread the knowledge about those scopes - LSP is using sublimelinter's scopes which is why some people don't see the squiggly lines.

rwols: For reference: https://github.com/sublimehq/Packages/issues/1036

Syntaxes/Diagnostics.sublime-syntax
rwols: I went with a different approach in the style of @Thom1729.

Syntaxes/Diagnostics.sublime-syntax
```diff
@@ -1,49 +1,67 @@
...
-
-  after-diagnostic-type-token:
-    - include: pop-at-end
-
-  pop-at-end:
+        - ensure-diag-meta-scope
+        - expect-diag-message
+        - expect-diag-type
+        - expect-linter-type
+        - expect-line-maybe-column
```
deathaxe: Cool, wasn't aware of such constructs work.

Syntaxes/Diagnostics.sublime-syntax
```diff
@@ -1,49 +1,67 @@
...
+    - match: \berror\b
+      scope: markup.error.lsp markup.deleted.lsp
+      pop: true
+    - match: \bwarning\b
+      scope: markup.warning.lsp markup.heading.lsp
+      pop: true
+    - match: \binfo\b
+      scope: markup.info.lsp markup.inserted.lsp
+      pop: true
+
```
deathaxe: Should `match: $ pop true` be included in each of the expect sections to ensure each one is poped of at line end even if nothing was found?

rwols: added in 0ad9517

tomv564:
Thanks for collaborating on this @rwols and @deathaxe!

Some comments:
* The protocol defines an optional `code` field for Diagnostics, which we could add in the future. Pyls's linters should be filling in that field with the DXXX  / EXXX numbers.
* We should update the References syntax if it suffers from the same issues.

PR #20, opened 2017-08-18
Output panel settings

deathaxe:
I'd suggest some changes to the output panel creation.

1. Use a common function to create both panels to ensure common style.
2. Make sure output panel is marked read-only to prevent accidentally modifications


tomv564:
Thanks so much!

PR #28, opened 2017-08-20
Improve visual settings of output panels

deathaxe:
I found some further settings to improve the look and feel of the output panel yesterday, which I forgot to push to the recently merged branch. So here it is as another PR

- formatting `line:column` to ensure the colons being aligned
- (hopefully) provide enough space for each column so the tab_size can be reduced to 4 which makes the output look more compact
- added a bunch of settings to ensure to make the output panel look nicer for everyone removing unnecessary stuff
- removed the `base_dir` setting as it not used anywhere and would possibly not make sense to as it is created once but might change during lifetime of the panel, if the open project is changed on the fly.

### DIFF
--- a/main.py
+++ b/main.py
@@ -996,7 +996,7 @@ def is_at_word(view: sublime.View, event) -> bool:
         return False
 
 
-OUTPUT_SETTINGS = {
+OUTPUT_PANEL_SETTINGS = {
     "auto_indent": False,
     "draw_indent_guides": False,
     "draw_white_space": "None",
@@ -1015,7 +1015,7 @@ OUTPUT_SETTINGS = {
 def create_output_panel(window: sublime.Window, name: str) -> sublime.View:
     panel = window.create_output_panel(name)
     settings = panel.settings()
-    for key, value in OUTPUT_SETTINGS.items():
+    for key, value in OUTPUT_PANEL_SETTINGS.items():
         settings.set(key, value)
     return panel
 
--- a/main.py
+++ b/main.py
@@ -996,33 +996,27 @@ def is_at_word(view: sublime.View, event) -> bool:
         return False
 
 
+OUTPUT_SETTINGS = {
+    "auto_indent": False,
+    "draw_indent_guides": False,
+    "draw_white_space": "None",
+    "gutter": False,
+    'is_widget': True,
+    "line_numbers": False,
+    "margin": 3,
+    "match_brackets": False,
+    "scroll_past_end": False,
+    "tab_size": 4,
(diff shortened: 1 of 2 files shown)


tomv564:
I like the attention to detail!. Wish we could extract all the settings to a file and apply that on one line.

The `result_base_dir` is used by ST itself to look up where the relative `file_with_errors.py` lives on disk. See: http://docs.sublimetext.info/en/stable/reference/settings.html
We should keep it :)

Was this tested against the recently merged syntax changes ?

deathaxe:
> Wish we could extract all the settings to a file and apply that on one line.

Already thought about creating a dict object which is then applied in a loop. Most settings are ok to be hard coded but maybe a _LSP Widget.sublime-settings_ or _LSP Output.sublime_ file could hold that defaults and may be customized by user. Not that bad.

> The result_base_dir is used by ST itself ...

This answers my question how file navigation works from the output panel. I just experienced some issues with paths containing whitespace. Those settings would fix it. 

Thanks for the good link. Didn't know about it. Good stuff.

Will check that out.

Btw.: Did you consider splitting main.py into seperate modules?

main.py
```diff
@@ -996,23 +996,27 @@ def is_at_word(view: sublime.View, event) -> bool:
         return False
 
 
+OUTPUT_SETTINGS = {
```
tomv564: Nitpick: can you name this `OUTPUT_PANEL_SETTINGS` ?

main.py
```diff
@@ -819,11 +819,11 @@ def format_severity(severity: int) -> str:
 
 
 def format_diagnostic(diagnostic: Diagnostic) -> str:
-    (line, character) = diagnostic.range.start
-    location = "{}:{}".format(line + 1, character + 1)
-    formattedMessage = diagnostic.message.replace("\n", "").replace("\r", "")
-    return "\t{:<8}\t{:<8}\t{:<8}\t{}".format(
```
tomv564: About this leading tab, does the result_file_regex still work now that you have removed it? 
If it's not too much effort, a screenshot showing the visual changes would be appreciated. I've been trying to keep up with checking out your PRs locally, but it would more productive to just read the PR and approve :)

deathaxe:
Looks like that. But you may be right, the `result_line_regex` may fail, if the line number is 8 or more characters in length. 10.000.000 lines - ST dies before this happens 😈. A leading tab causes too much indention - will add one space for security.

screenshot

tomv564:
Thanks for the screenshot, I'll merge both your PRs now in case they had a dependency on one another.

PR #29, opened 2017-08-20
Scope naming

deathaxe:
We should consistently use `markup.error.lsp` to scope errors and `markup.warning.lsp` later for warnings as they are considered (new) RFC. `sublimelinter.mark.error` is kept for backward compatibility with older color schemes.

Region names should always contain or better start with the package name as they must be unique.


tomv564:
Multiple scopes is a great solution for supporting older schemes and (hopefully) newer schemes with `markup.*` scopes. Thanks!

PR #30, opened 2017-08-20
Apply scopes from diagnostics t references output panel highlighting

deathaxe:
The references output is quite static so multiple pushes may make things too complicated here.
The scope names are applied the same way we did for diagnostics.

The tab between file path and location is changed into a normal whitespace as tab does not align correctly if the file paths' length in the list differs too much. It looks better with fixed distance.

We may or may not scope the preamble more detailed, but I find a more minimalistic scoping better at the moment.

### DIFF
--- a/Syntaxes/References.sublime-syntax
+++ b/Syntaxes/References.sublime-syntax
@@ -1,14 +1,29 @@
 %YAML 1.2
 ---
 hidden: true
-scope: references
+scope: output.lsp.references
 
 contexts:
   main:
-    - match: ^(.*):$
-      scope: keyword.other.references.file
-    - match: ^\s+(.*)\s+(\d*+:\d+)$
-      captures:
-        1: meta.references.message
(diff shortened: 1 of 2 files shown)


deathaxe:
How about an unicode symbol in front of each reference?

screenshot

tomv564:
This looks good to me, we could highlight the symbol and file names in the pre-amble in the future.
I like the unicode symbol!
They could add a lot of value if we use them in completion popups to communicate things like 'class' or 'function'  or 'private' or 'public'.

I'll merge this shortly, thanks!

deathaxe:
I played a little bit with highlighting the preamble but finally came up with leaving it as is. Too many different colors make look the panel too noisy, I think.

We need to use unicode characters carefully at the moment as the current beta 3126 with GDI rendering is not able to render all of them correctly in all controls. But in general if valuable symbols exist they help making the GUI look neat with little effort.

Issue #25, opened 2017-08-19
toplevel completions are ignored


deathaxe:
The CompletionHandler drops all completions from the language server, if the cursor is not located directly after a completion character.

```python
class CompletionHandler(sublime_plugin.EventListener):
   
    ...

    def on_query_completions(self, view, prefix, locations):
        ...
                 if prev_char not in autocomplete_triggers:
                    return None
```

This prevents toplevel completions from being displayed. Jedi as an example, provides a list of all available modules after the `import` keyword. Those completions are filtered at the moment.

Instead of filtering I'd rather suggest to use the completion character published by the language server to dynamically create an auto-completion trigger for the view. Accessors like `->` could be added as `>` and the `on_query_completions` could check for the preceding `-` to decide whether to provide completions or not.

philippotto:
I agree! To rephrase it a bit: Every time ST shows autocompletions, a request should be made to the server. ST does this by default only for "alpha" characters. This is why the `autocomplete_triggers` should be used to invoke additional autocompletion queries.

deathaxe:
In #4 was discussed how to easily setup these additional triggers. With the knowledge about language-servers providing these information, I could imagine to automatically add this setting to each supported view. This way no trigger needs to be configured by user.

tomv564:
This investigation supports your conclusion:
https://github.com/Microsoft/language-server-protocol/issues/268

and I am for turning on completions for alphanumeric characters.

But I would like to verify that the performance is acceptable, some recent implementations against the javascript/typescript implementations ran into problems: https://github.com/sourcegraph/javascript-typescript-langserver/issues/331

tomv564:
@philippotto, it might be good to have the Flow language server additionally send `['.']` for triggerCharacters. Do you want to create an issue?

philippotto:
@tomv564 Sure, I created it [here](https://github.com/flowtype/flow-language-server/issues/52).

Regarding potential performance problems, this answer from your referenced issue is interesting:

> Drastically reducing how often we called completion has solved problems on Atom. Now we just trigger it on . and when manually requested.

As long as autocompletions are not blazingly fast anyway, I would leave it at that logic (on . and manually requesting). Hitting CTRL + Space for requesting autocompletions is probably idiomatic? We could still have a setting for triggering it after each keystroke (throttled), which is off by default maybe?

deathaxe:
[...]

PR #31, opened 2017-08-20
Improve compatibility with flow

philippotto:
Nice package! :)

I tried to integrate flow (see [this](https://github.com/flowtype/flow-language-server) LSP library) with this ST package and found some issues with the hover support.

One issue was that the code didn't comply with the specification for the hover payload. The contents attribute can look like this `contents: MarkedString | MarkedString[];` (see [here](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#textDocument_hover)), however it was implemented to only accept `contents: string | MarkedString[];`.

Another issue was that flow sometimes returns null instead of a MarkedString (array). This is flow's fault, but checking for that doesn't do any harm, I think.

The actual flow integration can be done by installing [this](https://github.com/flowtype/flow-language-server) package and adding the following to the config:

```
"flow":
      {
        "command": ["flow-language-server", "--stdio"],
        "scopes": ["source.js"],
        "syntaxes": ["Packages/Babel/JavaScript (Babel).sublime-syntax", "Packages/JavaScript/JavaScript.sublime-syntax"],
        "languageId": "javascript"
      },
```

### DIFF
--- a/main.py
+++ b/main.py
@@ -1466,6 +1466,10 @@ class HoverHandler(sublime_plugin.ViewEventListener):
 
     def handle_response(self, response, point):
         debug(response)
+        if response is None:
+            # Flow returns None sometimes
+            # See: https://github.com/flowtype/flow-language-server/issues/51
+            return
         contents = response.get('contents')
         if len(contents) < 1:
             return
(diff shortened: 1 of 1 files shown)


tomv564:
Thanks for reporting, this was totally a miss and your fix is elegant ! I'll do a quick review now.

main.py
```diff
@@ -1466,6 +1466,9 @@ def on_hover(self, point, hover_zone):
 
     def handle_response(self, response, point):
         debug(response)
+        if response is None:
+            # Flow returns None sometimes
```
tomv564: Do you mind creating an issue for the Flow language server and linking it in the comment?

philippotto:
I just found another problem. The spec says that the `triggerCharacters` attribute in `CompletionOptions` is optional. At the moment, LSP assumes that this is always defined.
However, I'm not sure how to resolve this best. If no `triggerCharacters` are defined, should we query for completions on every keystroke? Could this be a problem?

main.py
philippotto: Sure, I'll do that!

deathaxe:
We even should (see #25). I found top-level-completions such as keywords or imports (python) don't work as they are filtered. If ST queries for completion, LSP must forward it to the language server and show the results. How should LSP know whether the language server has something to offer or not?

`triggerCharacters` are special none-word characters, which would normally not trigger auto-completion. They are not a must have.

tomv564:
Yes, let's try to figure out what the spec means with 'triggerCharacters' in #25. I'll merge this now, thanks for reporting the issue to the Flow server project!

frou:
I'm curious about the potential of `flow`'s autocompletion engine.

Currently, [tern](http://ternjs.net) along with [tern_for_sublime](https://github.com/ternjs/tern_for_sublime) gives fantastically robust autocompletion for plain JS. 

I like the idea of flow and annotating JS (with [flow comments](https://flow.org/blog/2015/02/20/Flow-Comments/)) but until flow's autocompletion engine is as robust as tern's, it's too strange to also have to keep in mind an independent background type system that's different from what autocompletion is showing.

PR #35, opened 2017-08-20
Use urllib to translate between url and filename

deathaxe:
This commit basically contains some results of my investigation on python-language-server issues, yesterday. I've created a module uri.py, which handles the job. The client should use the same standard functions to translate between url and filename to avoid OS specific code.

As the main.py is quite large, I would suggest to just import the urllib and use the functions with full path. This is a bit more expressive and it is easier to see where the functions come from.

I use the opportunity to group and sort the imports as suggested by several official sources

First sorting level (groups):

1. standard library imports
2. sublime text default imports
3. dependencies packages imports
4. private imports (not yet used)

Second sorting level:

1. import
2. from ... import

Third sorting level:

Sort alphabetically

see also: https://packagecontrol.io/packages/Python%20Fix%20Imports

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,20 +1,24 @@
-import sublime_plugin
-import sublime
-import subprocess
-import threading
+import html
 import json
 import os
+import subprocess
 import sys
-import urllib.request as urllib
-from urllib.parse import urljoin
+import threading
 from collections import OrderedDict
-import html
-import mdpopups
+from urllib.parse import urljoin
+from urllib.parse import urlparse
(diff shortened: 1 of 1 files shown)


deathaxe:
Why did travis fail? The package works fine in ST.

deathaxe:
A comment on line 1570 ??? I can't see any there.

deathaxe:
Rebased on current master

tomv564:
Thanks for the cleanup!

PR #37, opened 2017-08-20
Add error/warning/info regions and disable phantoms by default.

deathaxe:
Had a look into the regions and phantoms stuff a little bit with more attention to the regions right now. 

1. I'd suggest to disable phantoms by default at the moment as they make files with many issues unreadable at the moment.

2. Errors/Warnings/Infos are added with different scopes so they can have different colors. Scopes are synced with diagnostics syntax. Decided to use sublimelinter scopes in diagnostics panel for now. But we may discuss which solution is better. I am not sure.

For details, please see commit messages.

### DIFF
--- a/Syntaxes/Diagnostics.sublime-syntax
+++ b/Syntaxes/Diagnostics.sublime-syntax
@@ -42,16 +42,16 @@ contexts:
 
   expect-diag-type:
     # See: https://github.com/sublimehq/Packages/issues/1036
-    # We use old markup scopes too so that old color schemes can catch up.
+    # We use sublimelinter markup scopes too so that old color schemes can catch up.
     - include: pop-at-end
     - match: \berror\b
-      scope: markup.error.lsp markup.deleted.lsp
+      scope: markup.error.lsp sublimelinter.mark.error
       pop: true
     - match: \bwarning\b
-      scope: markup.warning.lsp markup.heading.lsp
+      scope: markup.warning.lsp sublimelinter.mark.warning
       pop: true
     - match: \binfo\b
-      scope: markup.info.lsp markup.inserted.lsp
+      scope: markup.info.lsp sublimelinter.gutter-mark
       pop: true
 
   expect-linter-type:
--- a/Syntaxes/Diagnostics.sublime-syntax
+++ b/Syntaxes/Diagnostics.sublime-syntax
@@ -45,13 +45,13 @@ contexts:
     # We use sublimelinter markup scopes too so that old color schemes can catch up.
     - include: pop-at-end
     - match: \berror\b
-      scope: markup.error.lsp sublimelinter.mark.error
+      scope: markup.deleted.lsp sublimelinter.mark.error markup.error.lsp
       pop: true
     - match: \bwarning\b
-      scope: markup.warning.lsp sublimelinter.mark.warning
+      scope: markup.changed.lsp sublimelinter.mark.warning markup.warning.lsp
       pop: true
     - match: \binfo\b
-      scope: markup.info.lsp sublimelinter.gutter-mark
+      scope: markup.inserted.lsp sublimelinter.gutter-mark markup.info.lsp
       pop: true
 
   expect-linter-type:
--- a/main.py
+++ b/main.py
@@ -1166,11 +1166,11 @@ def update_diagnostics_phantoms(view: sublime.View, diagnostics: 'List[Diagnosti
     global phantom_sets_by_buffer
 
     buffer_id = view.buffer_id()
-    if show_diagnostics_phantoms and not view.is_dirty():
+    if not show_diagnostics_phantoms or view.is_dirty():
+        phantoms = None
+    else:
         phantoms = list(
             create_phantom(view, diagnostic) for diagnostic in diagnostics)
-    else:
-        phantoms = None  # type: ignore
     if phantoms:
         phantom_set = phantom_sets_by_buffer.get(buffer_id)
         if not phantom_set:
@@ -1184,7 +1184,7 @@ def update_diagnostics_phantoms(view: sublime.View, diagnostics: 'List[Diagnosti
 def update_diagnostics_regions(view: sublime.View, diagnostics: 'List[Diagnostic]', severity: int):
     region_name = "lsp_" + format_severity(severity)
     if show_diagnostics_phantoms and not view.is_dirty():
-        regions = None  # type: ignore
+        regions = None
     else:
         regions = list(create_region(view, diagnostic) for diagnostic in diagnostics
                        if diagnostic.severity == severity)
--- a/main.py
+++ b/main.py
@@ -26,7 +26,7 @@ SUBLIME_WORD_MASK = 515
 show_status_messages = True
 show_view_status = True
 auto_show_diagnostics_panel = True
-show_diagnostics_phantoms = True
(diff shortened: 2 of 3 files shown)


tomv564:
I recommend you install `mypy` to check the typing manually, or the mypy plugin for the python language server: https://pypi.python.org/pypi/pyls-mypy/0.1.2

Also a request, could you keep your PRs focused to solving one problem at a time? This way we can quickly merge obvious/important stuff, and have more in-depth reviews on the more explorative work.

deathaxe:
Will try it ;-)

deathaxe:
Hmm, pyls-mypy didn't complain locally. And honestly - If I want to initiate an object with None, I do so. There is no need to create an empty list, which is not used but for a following empty check. Is there a way to tell mypy a list initiated with `None` is ok? I ask, because I never used typing notation right now.

philippotto:
> Is there a way to tell mypy a list initiated with None is ok? I ask, because I never used typing notation right now.

I didn't use mypy before, but I guess it's similar to flow which is why you should be able to write the following:

```regions = None  # type: Optional[List[sublime.Region]]```

Maybe the type comment is not necessary if you initialize the variable above the if-block (and remove the else branch).

**Comment by deathaxe** [comment]
> Maybe the type comment is not necessary if you initialize the variable above the if-block (and remove the else branch).

CI's mypy even complains about the line 1159 where the variable is already defined.

I definitely don't find the requirement to define types in comments after a line helpful. It already brings some noise to the interface part of functions but may be useful there even though I lived well without that stuff right now. I'll definitely won't start liking that - not mentioning the tons of json files mypy creates. 🤢

**Comment by deathaxe** [comment]
Rebased my work on the current master.

**Comment by philippotto** [comment]
> CI's mypy even complains about the line 1159 where the variable is already defined.

I just gave it a try and the following code seems to work:

```python
    phantoms = None

    buffer_id = view.buffer_id()
    if show_diagnostics_phantoms and not view.is_dirty():
        phantoms = list(
            create_phantom(view, diagnostic) for diagnostic in diagnostics)
```

Mypy's problem with the current code is (as far as I can tell) that it infers the type of the `phantoms` variable by looking at its first (potential) assignment. This assignment is in the if-body and amounts to a list of phantoms. Afterwards, it checks that all references to the `phantoms` variable fulfill this type requirement, which fails when setting it to `None`. Initializing `phantoms` above the if-block with `None` works, since mypy can infer that it's optional.

Type checkers can be a bit tedious to deal with from time to time, but in the long run they can save a lot of time and improve code readability and reliability in my opinion. That's why I would avoid disabling the checks with `# type: ignore`.

Syntaxes/Diagnostics.sublime-syntax
```diff
@@ -42,16 +42,16 @@ contexts:
...
     # See: https://github.com/sublimehq/Packages/issues/1036
-    # We use old markup scopes too so that old color schemes can catch up.
+    # We use sublimelinter markup scopes too so that old color schemes can catch up.
     - include: pop-at-end
     - match: \berror\b
-      scope: markup.error.lsp markup.deleted.lsp
+      scope: markup.error.lsp sublimelinter.mark.error
       pop: true
     - match: \bwarning\b
-      scope: markup.warning.lsp markup.heading.lsp
```
tomv564: Can we have all three groups, with the sublimelinter scopes in between the markup classes? We'd have (future scopes without hacks) > (correct colors but not reliable) > (at least something is visible)

main.py
```diff
@@ -1145,37 +1148,45 @@ def update_file_diagnostics(window: sublime.Window, file_path: str, source: str,
...
-        buffer_id = view.buffer_id()
-        if buffer_id not in phantom_sets_by_buffer:
-            phantom_set = sublime.PhantomSet(view, "diagnostics")
-            phantom_sets_by_buffer[buffer_id] = phantom_set
-        else:
-            phantom_set = phantom_sets_by_buffer[buffer_id]
+def update_diagnostics_regions(view: sublime.View, diagnostics: 'List[Diagnostic]', severity: int):
+    region_name = "lsp_" + format_severity(severity)
+    if show_diagnostics_phantoms and not view.is_dirty():
+        regions = None  # type: ignore
```
tomv564: Please apply @philippotto's suggestion or revert back to `[]` for both regions and phantoms

main.py
```diff
@@ -23,15 +23,13 @@
 
 PLUGIN_NAME = 'LSP'
 SUBLIME_WORD_MASK = 515
-MARKUP_ERROR = 'markup.error.lsp sublimelinter.mark.error'
 show_status_messages = True
 show_view_status = True
 auto_show_diagnostics_panel = True
 show_diagnostics_phantoms = True
```
tomv564: For consistency's sake, can you change this value to the settings default of True?

main.py
```diff
@@ -269,7 +273,6 @@ def load_settings():
     show_view_status = settings_obj.get("show_view_status", True)
     auto_show_diagnostics_panel = settings_obj.get("auto_show_diagnostics_panel", True)
     show_diagnostics_phantoms = settings_obj.get("show_diagnostics_phantoms", True)
```
tomv564: Yes, it's duplication, but please change this to match the settings default.
We can get rid of the defaults in `settings_obj.get('my_key')` in a future PR.

main.py
deathaxe: # type: Optional[List[sublime.Region]] doesn't work for phantoms as they are already defined but mypy on CI will complain about None assignment anyway, even though the pyls-mypy does not. You'll get errors for phatoms at least, changing it. Creating an empty list, which is not used at all is nonesense.

deathaxe:
Rebased to resove merge conflicts

main.py
tomv564: Thanks for giving this a go, mypy-pyls also did not report this issue, running `mypy main.py` locally did.
I agree the type inference is not 100%, but I'm glad it could be solved without extra annotations.

deathaxe:
Just had another idea how to avoid the None assignment, by following code structure when I saw this PR already merged 😄 

```python
def update_diagnostics_phantoms(view: sublime.View, diagnostics: 'List[Diagnostic]'):
    global phantom_sets_by_buffer

    buffer_id = view.buffer_id()
    if show_diagnostics_phantoms and not view.is_dirty():
        phantoms = list(
            create_phantom(view, diagnostic) for diagnostic in diagnostics)
        if phantoms:
            phantom_set = phantom_sets_by_buffer.get(buffer_id)
            if not phantom_set:
                phantom_set = sublime.PhantomSet(view, "lsp_diagnostics")
                phantom_sets_by_buffer[buffer_id] = phantom_set
            phantom_set.update(phantoms)
            return
    phantom_sets_by_buffer.pop(buffer_id, None)


def update_diagnostics_regions(view: sublime.View, diagnostics: 'List[Diagnostic]', severity: int):
    region_name = "lsp_" + format_severity(severity)
    if not show_diagnostics_phantoms or view.is_dirty():
        regions = list(create_region(view, diagnostic) for diagnostic in diagnostics
                       if diagnostic.severity == severity)
        if regions:
            scope_name = diagnostic_severity_scopes[severity]
            view.add_regions(region_name, regions, scope_name, "dot",
                             sublime.DRAW_SQUIGGLY_UNDERLINE | UNDERLINE_FLAGS)
            return
    view.erase_regions(region_name)
```

PR #38, opened 2017-08-21
Use builtin dictionaries to create requests.

deathaxe:
JSON is a structured data set with strong syntax checks and naming. Therefore I wonder if it is required to use the much slower OrderedDict to create requests. I didn't find any statement or rules about sorting json messages in the official specification.

Specification

  see: https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md

Dictionary benchmarks

  see: https://stackoverflow.com/questions/19629682/ordereddict-vs-defaultdict-vs-dict#19643045

### DIFF
--- a/main.py
+++ b/main.py
@@ -7,7 +7,6 @@ import os
 import sys
 import urllib.request as urllib
 from urllib.parse import urljoin
-from collections import OrderedDict
 import html
 import mdpopups
 try:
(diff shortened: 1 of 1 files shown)


tomv564:
You are correct, field order should not matter.
The OrderedDict was added to support `clangd` in this commit: https://github.com/tomv564/LSP/commit/c987828f4e22a76a6fe218f5154bccc1e3df3ab2

It would be great to remove OrderedDict, but as their server appears to work with VS Code I feel it's a small effort to support them until they improve their request handling.

deathaxe:
OK, good to know.

PR #40, opened 2017-08-22
Only update settings when changed, don't re-initialize settings handling

tomv564:
Especially while developing, an enormous amount of settings reloading can be witnessed after saving a settings file.
This change should prevent registering the same settings update handler multiple time.

### DIFF
--- a/main.py
+++ b/main.py
@@ -245,6 +245,12 @@ def read_client_config(name, client_config):
 
 
 def load_settings():
+    settings_obj = sublime.load_settings("LSP.sublime-settings")
+    update_settings(settings_obj)
+    settings_obj.add_on_change("_on_new_settings", lambda: update_settings(settings_obj))
+
+
+def update_settings(settings_obj: sublime.Settings):
     global show_status_messages
     global show_view_status
     global auto_show_diagnostics_panel
@@ -254,8 +260,6 @@ def load_settings():
     global log_server
     global log_stderr
     global configs
-
-    settings_obj = sublime.load_settings("LSP.sublime-settings")
 
     configs = []
     client_configs = settings_obj.get("clients", {})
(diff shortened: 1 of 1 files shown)


deathaxe:
Is a step into the correct direction, but I am honestly in doubt with the usefulness of the construct at all. 

`sublime.load_settings()` returns the reference of a python dict object which is created by ST core and kept sync with the *.sublime-settings file automatically. Therefore calling `settings.get()` is nothing else but an ordinary dictionary operation as done a dozen of times everywhere in the plugin. 

Some packages even call `sublime.load_settings()` each time they query a setting. Did you encounter any performance issues or is there any other reason why to cache all the configs?

Another maybe more general question is: 

Client, Request, Notification are examples of well formed classes to group functions together. You created a `ClientConfig` class to group client settings. Why don't you create a `LspConfig` class to handle all settings the same way? Why do you place all the settings as globals instead? 

Please don't misunderstand my critics, but it looks a bit mixed up and may cause difficulties when it comes to split the main.py into several modules some day. 

From my point of view there are basically two strategies:

1. Create classes for everything which belongs together and therefore may have several classes in one module without loosing structure.

2. Group things in separate modules and keep structure as flat as possible. Means no classes but functions only.

In both cases you'll be able to call

```python
config.save()
if config.show_status:
    do something

```

In case 1 `config` would be an object of type `class Config`. In case 2 it was a module whose globals/functions are accessed.

... just an idea.

deathaxe:
As LSP.sublime-settings exists in your package path, `settings_obj.get(...)` will always succeed. Therefore I think storing extra default values in the code is unnecessary.

tomv564:
I like the idea of creating an `LSPConfig` class, and getting rid of the double defaults will be nice in the future, too.

Issue #18, opened 2017-08-18
No client available for view if already open on ST startup


deathaxe:
By default ST has set `hotexit: true`, which causes the previous session and all its open files to be restored the next time ST starts.

LSP does not create a client object for such views. Instead the following lines are displayed.

```
LSP: Config added: dotty 
LSP: Config added: clangd 
LSP: Config added: jsts 
LSP: Config added: rls 
LSP: Config added: pyls 
LSP: no clients found for window 2 
LSP: pyls not available for view C:\Apps\Sublime\Data\Packages\LSP\main.py in window 2 
```

The `pyls` is started as soon as the view is deactivated and activated again by minimizing and restoring ST window or switching to another view and return.

It looks like the issue may be triggered by a bug in ST core, which prevents the `on_load` and `on_activated_async` to be sent to already open views upon startup, but maybe LSP could lazily initialize lsp servers for open views in the `plugin_loaded()` hook.

tomv564:
I think it's not uncommon to have a lot of open views when launching ST after a hot exit, we may want to avoid hammering the language server at startup.

Perhaps if we begin by starting a client for the *active view* in `plugin_loaded()` then the other files should trigger didOpen as you switch to them.

Scenarios to test:
* Calling didOpen for all open views on a 20+ file session and see what happens. 
* Doing a find/replace to open a lot of files at the same time, then save them all at once.

deathaxe:
You are absolutely right. As a language server is activated on demand after switching to another view, the currently focused ones are interesting only. We just need to pay attention to multi view layouts, which may display several views a user might want to see information for initially even though they were not yet activated.

In GitGutter I use the following function (method of the EventListener in my case) to check for visible views

```python
    def is_view_visible(self):
        """Determine if the view is visible.

        Only an active view of a group is visible.

        Returns:
            bool: True if the view is visible in any window.
        """
        window = self.view.window()
        if window:
            view_id = self.view.id()
            for group in range(window.num_groups()):
                active_view = window.active_view_in_group(group)
                if active_view and active_view.id() == view_id:
                    return True
        return False
```

tomv564:
The smallest fix for the focused view is merged, a bit more work is needed to initialize multiple visible views for a single language server at once.

PR #41, opened 2017-08-22
Initialize the active window's active view at startup

tomv564:
For #18, where already-open views do not trigger the language server to startup (only toggling events or opening other views works).

In limited testing on OS-X, the non-active window had an activated event fired on its view once I changed to it, so only the active_window's active_view needed a manual startup.

### DIFF
--- a/main.py
+++ b/main.py
@@ -496,6 +496,15 @@ def plugin_loaded():
     Events.subscribe("view.on_activated_async", initialize_on_open)
     if show_status_messages:
         sublime.status_message("LSP initialized")
+    start_active_view()
+
+
+def start_active_view():
+    window = sublime.active_window()
+    if window:
+        view = window.active_view()
+        if view and is_supported_view(view):
+            initialize_on_open(view)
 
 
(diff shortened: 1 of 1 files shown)


deathaxe:
I'd rather suggest to initialize all visible views.

```python
def start_active_view():
    for window in sublime.windows():
        for group in range(window.num_groups()):
            view = window.active_view_in_group(group)
            if view and is_supported_view(view):
                initialize_on_open(view)
```

Unfortunately even though `initialize_on_open(view)` is called for all visible views, LSP handles diagnostics for the last view only. Is this an issue with stdin/stdout communication or may it be a limitation of the python-language-server?

tomv564:
There might be an issue calling `initialize_on_open` -> `get_window_client` multiple times in a row while the client is starting.
It should be sufficient to `didopen_after_initialize.append(view)` with the 2nd, 3rd etc. view.

Focusing the 2nd view triggers the `on_activated_async` event, we can look at activating all visible views once the issue you found is fixed!

PR #42, opened 2017-08-22
Create LspSettings class to manage cached settings

deathaxe:
With respect to object oriented programming which is already used for most of the language-server-protocol based stuff, the configuration of the LSP package should follow this approach, if it is intended to cache the settings.

Therefore a `class LspSettings` is created, which uses a template `LSP_SETTINGS` to initialize the object attributes at startup. In `plugin_loaded()` callback the real values from the _LSP.sublime-settings_ are loaded, mapped to the object attributes and kept up to date by the subscribed the `on_change()` event.

The one and only place to define settings to handle is the `LSP_SETTINGS` dictionary.

All managed settings can be accessed (read-only) as normal attribute by calling
```
   lsp_settings.my_setting_from_lsp_sublime_settings
```
The only drawback is mypy which complains about class attributes not found. Therefore several `# type: ignore` statements need to be added.

### DIFF
--- a/main.py
+++ b/main.py
@@ -23,16 +23,6 @@ import mdpopups
 
 PLUGIN_NAME = 'LSP'
 SUBLIME_WORD_MASK = 515
-show_status_messages = True
-show_view_status = True
-auto_show_diagnostics_panel = True
-show_diagnostics_phantoms = False
-show_diagnostics_in_view_status = True
-log_debug = True
-log_server = True
-log_stderr = False
-
-configs = []  # type: List[ClientConfig]
 
 
 class DiagnosticSeverity(object):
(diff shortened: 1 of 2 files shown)


rwols:
What about project-specific settings?

Say I have diagnostics set to false in my User settings, but I want it enabled for certain projects. Maybe this should be part of another PR once this is merged.

deathaxe:
This PR is a suggestion. Not sure, whether this is the best solution. Let's talk about what we might need.

As stated somewhere already, I am not very convinced of caching ST settings at all. The settings object is an dictionary which is synced with the sublime-settings files already. So the only real benefit of caching is to prevent us from calling `settings.get("the_setting")` all the time, but treat the settings object like a `namedtuple`.

To support project specific settings, we can't use the `LSP.sublime-settings` We'd instead need to define LSP settings in the Preferences.sublime-settings file and use `view.settings()` object to access them. This is how GitGutter supports project settings. I prefixed each of the package settings with `git_gutter_` and allow them to be placed into Preferences.sublime-settings as well. Therefore all settings may be view/project/user/global same as all ST preferences.

Today I'd rather recommend to use `lsp.` as prefix as PackageDev highlights it as namespace - what it really is.

Or we'd need to struggle with the `set_project_data/get_project_data` API which I wouldn't prefer.

rwols:
I just [use this](https://github.com/rwols/CMakeBuilder/blob/master/support/get_setting.py): check the view's settings for a setting, if it's not present fall back to the default/user settings file.

If you prefix all of your settings with `lsp.` or whatever, you can perfectly fine put those in the `"settings"` dict of a `.sublime-project` and they'll populate the view's settings.

deathaxe:
As soon as we want to use the "global" settings namespace we must always use some kind of unique settings key name in order to prevent collisions with other packages. Therefore a prefix is crucial. 

GitGutter uses the same strategy: It calls  `view.settings().get("git_gutter_enabled")` and `git_gutter_settings.get("enabled")` in case the view settings object returns `None`. 

@tomv564 : If I place the LspSettings class into a dedicated module like `modules/settings.py`, and import the LspSettings into main.py, mypy does not complain about the missing attributes, which are dynamically created. Would something like that be a valid solution? I ask, because I had `namedtuples` in mind, when I created `LspSettings`. They dynamically create a class from a tuple with the help of a string template. I guess mypy would muck about that, too, if it was done in place.

tomv564:
@deathaxe there must be a way to configure settings and typings without repeating ourselves?

```python
class LSPSettings(object):

    @classmethod
    def register_setting(cls, lookup: 'dict', key: 'str', default):
        lookup[key] = default
        return default

    def __init__(self):
        self.lookup = {}  # type: Dict[str, Any]
        self.log_debug = self.register_setting(self.lookup, 'log_debug', True)  # type: bool
        self.name = self.register_setting(self.lookup, 'name', "asdf")  # type: str

    def update(self, settings: 'sublime.Settings'):
        for key, default in self.lookup.items():
            self.__setattr__(key, settings.get(key, default))
```

tomv564:
Project-specific configuration for language servers is a must-have, but I see these global toggles (e.g. `auto_show_diagnostics_panel` as something you set once to your taste and then get used to.

deathaxe:
>     def __init__(self):
>        self.lookup = {}  # type: Dict[str, Any]
>        self.log_debug = self.register_setting(self.lookup, 'log_debug', True)  # type: bool
>        self.name = self.register_setting(self.lookup, 'name', "asdf")  # type: str

How do you want to avoid repeating this way? You just split the static LSP_SETTINGS dictionary into functions. You write the key name and default value in both the code and LSP.sublime-settings file, too.

As we can't iterate the sublime settings object, we somewhere need to add a list of known keys in the code, if we want to provide defaults during startup. We could load keys from LSP.sublime-settings, but this will be possible not earlier then with `plugin_loaded()` callback. API isn't ready for use earlier and I wouldn't suggest using python's default json module to do so.

tomv564:
> How do you want to avoid repeating this way? 

Sorry, got ahead of myself writing that comment. I liked your dict defining the settings, but anticipated we'd have to repeat all the field names to get type declarations.

I'll happily look at this PR again when it doesn't circumvent the type declarations!

deathaxe:
> ... anticipated we'd have to repeat all the field names to get type declarations.

You mean declaring them in the `__init_()` again?

I tried vice versa, `dir(..)` the attributes of the class, but this may cause unwanted keys to be updated. We'd need to write a blacklist. Didn't like it. But typing doesn't work for dynamically added attributes, which I can understand on the one hand but not on the other as it is a normal python feature to do so. Many of the stdlib packages use such methods.

> I'll happily look at this PR again when it doesn't circumvent the type declarations!

How to teach mypy to understand the relative imports of `modules.settings`?

Issue #36, opened 2017-08-20
Show tooltip even when no hover information could be found


philippotto:
Usually, when hovering over code a tooltip with some information is provided by the server. However, the server can also send "nothing" back (most simple example would be an empty line). In this case, LSP doesn't show a tooltip at all, which can be a bit annoying since it's not clear:
- if the hovering didn't work correctly or
- if LSP is still waiting for a reply or
- if there is simply no hover information available

To cut a long story short: I'd suggest to always show a tooltip on hover. If the server didn't provide information, then it could just say so ("No information available" or similar).

Agree? I could do a PR if you want.

philippotto:
> We request hovers on any word, do we really need to show "No information available" when the user rests the mouse on an if statement?

I get your point. However, I think the general question is, if popups are annoying or not. If I move my mouse and hover over something (even an `if` keyword), then I'm not much surprised if I get an (empty) tooltip. However, I might be biased since I very rarely use the mouse (hovering is one of the few use cases for me :)).
For the `if` case: Maybe we can check the scope of the hovered word and if it's a keyword of the language, then we don't show the "No information available"?

Another approach would be to show a general "loading indicator" if LSP is waiting for a server response. Maybe something for ST's status bar? Unfortunately, ST is a bit restricted in this way, which is why the status bar tends to get overloaded quite quickly.

> Some people will prefer not seeing the "No information available" hover, so a setting and a default will need to be added.

Definitely!

> In the case where you are expecting to see hover details, is there a reason why the language server can't provide it?

Hm, I don't think there is a good reason. I'd call it "insufficiency of the language server". As an example for flow: 
```export class SomeComponent extends React.PureComponent<Props, State> { ... ```

Hovering over `Props` won't show type information.

**Comment by deathaxe** [comment]
Jedi - AutoCompletions package already uses the scopes to filter when to display popups or when not. A popup might be displayed for `variable` or `entity.name` scopes. Maybe this could be configured by two settings like:

```js
{
  "show_signature_popup": true,
  "signature_popup_selector": "variable, entity.name "
}
```

Just as a hint. The PackageDev package added support for settings namespaces like used in VS Code. If too many settings will be added, this might help grouping?

```js
{
  "signature.show_popup": true,
  "signature.popup_selector": "- keyword"
}

tomv564:
@philippotto let's do it! We might want to have this on by default, so we can call the setting something like `hide_empty_hover` ?

philippotto:
[...]

PR #46, opened 2017-08-24
Always show symbol information popup

deathaxe:
This PR addresses issue #36 and intends to display tooltip even if language-server doesn't offer a description. A basic filter is used to statically exclude some scopes which a language server will most likely not have something to offer for.


main.py
```diff
@@ -1449,14 +1449,11 @@ def request_symbol_hover(self, point, hover_zone):
 
     def handle_response(self, response, point):
         debug(response)
-        if response is None:
-            # Flow returns None sometimes
```
tomv564: Please keep these comments

PR #47, opened 2017-08-25
Enable full completions

tomv564:
For issue #25, where all completion requests (even manual) were ignored unless the cursor followed a trigger character.

The following improvements were made:
* No more filtering on trigger chars (ctrl+space no longer ignored)
* Switch to ViewEventListener (easier to disable, less risk with shared state)
* Track completion states so we can discard and request new completions as the user keeps typing

To do:
- [x] verify only_complete_on_trigger_characters and make it a setting
- [x] also disable listener if server has no completionProvider
- [x] clean up debug logging
- [x] rename to CompletionHandler

### DIFF
--- a/main.py
+++ b/main.py
@@ -1511,52 +1511,89 @@ class HoverHandler(sublime_plugin.ViewEventListener):
             max_width=800)
 
 
-class CompletionHandler(sublime_plugin.EventListener):
-    def __init__(self):
-        self.completions = []  # type: List[Tuple[str, str]]
-        self.requesting = False
-        self.refreshing = False
+class CompletionState(object):
+    IDLE = 0
+    REQUESTING = 1
+    APPLYING = 2
+    CANCELLING = 3
 
-    def on_query_completions(self, view, prefix, locations):
+
+class CompletionHandler2(sublime_plugin.ViewEventListener):
+    def __init__(self, view):
+        self.view = view
+        self.autocomplete_triggers = None
+        self.completions = []  # type: List[Tuple[str, str]]
+        self.state = CompletionState.IDLE
+        self.next_request = None
+
+    @classmethod
+    def is_applicable(cls, settings):
+        syntax = settings.get('syntax')
+        return is_supported_syntax(syntax)
+
+    def initialize_triggers(self):
+        client = client_for_view(self.view)
+        if client:
+            completionProvider = client.get_capability(
+                'completionProvider')
+            if completionProvider:
+                self.autocomplete_triggers = completionProvider.get(
+                    'triggerCharacters')
+                return
+
+        self.autocomplete_triggers = []
+
+    def is_after_trigger_character(self, location):
+        if location > 0:
+            prev_char = self.view.substr(
+                sublime.Region(location - 1, location))
+            return prev_char in self.autocomplete_triggers
+
+    def on_query_completions(self, prefix, locations):
+        debug('enter at state {}'.format(self.state))
+        if self.state == CompletionState.IDLE:
+            self.do_request(prefix, locations)
+            self.completions = []
+
+        elif self.state == CompletionState.REQUESTING or self.state == CompletionState.CANCELLING:
+            debug('replacing request')
+            self.next_request = (prefix, locations)
+            self.state = CompletionState.CANCELLING
+
+        elif self.state == CompletionState.APPLYING:
+            debug('applying', self.completions)
+            self.state = CompletionState.IDLE
+
+        return self.completions, (sublime.INHIBIT_WORD_COMPLETIONS
+                                  | sublime.INHIBIT_EXPLICIT_COMPLETIONS)
+
+    def do_request(self, prefix, locations):
+        self.next_request = None
+        view = self.view
+
         if not is_supported_view(view):
             return None
 
-        if not self.refreshing and not self.requesting:
-            client = client_for_view(view)
+        client = client_for_view(view)
+        if not client:
+            return
 
-            if not client:
-                return
+        completionProvider = client.get_capability('completionProvider')
+        if not completionProvider:
+            return
 
-            completionProvider = client.get_capability('completionProvider')
-            if not completionProvider:
-                return
+        if only_complete_on_trigger_characters and not self.is_after_trigger_character():
+            return
 
-            if only_complete_on_trigger_characters:
-                autocomplete_triggers = completionProvider.get('triggerCharacters')
+        snapshot = view.substr(view.full_line(locations[0])).strip()
+        row, col = view.rowcol(locations[0])
+        debug('requesting for: "{}" {}:{}'.format(snapshot, row, col))
 
-                if locations[0] > 0:
-                    self.completions = []
-                    prev_char = view.substr(
-                        sublime.Region(locations[0] - 1, locations[0]))
-                    if prev_char not in autocomplete_triggers:
-                        return None
-
-            # TODO: cancel last completion list when backspacing.
-
-            purge_did_change(view.buffer_id())
-            row, col = view.rowcol(locations[0])
-            debug('requesting for: ', prefix, row, ':', col)
-            client.send_request(
-                Request.complete(get_document_position(view, locations[0])),
-                self.handle_response)
-            self.requesting = True
-            if not only_complete_on_trigger_characters:
-                return [], (sublime.INHIBIT_WORD_COMPLETIONS
-                            | sublime.INHIBIT_EXPLICIT_COMPLETIONS)
-
-        self.refreshing = False
-        return self.completions, (sublime.INHIBIT_WORD_COMPLETIONS
-                                  | sublime.INHIBIT_EXPLICIT_COMPLETIONS)
+        purge_did_change(view.buffer_id())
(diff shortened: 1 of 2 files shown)


main.py
```diff
@@ -1511,43 +1512,90 @@ def show_hover(self, point, contents):
...
+        self.autocomplete_triggers = None
+        self.completions = []  # type: List[Tuple[str, str]]
+        self.state = CompletionState.IDLE
+        self.next_request = None
 
-            if not client:
-                return
+    @classmethod
+    def is_applicable(cls, settings):
+        syntax = settings.get('syntax')
```
rwols: In rare cases `syntax` may be `None` (annoying ST bug), check for this just in case.

main.py
```diff
@@ -1511,43 +1512,90 @@ def show_hover(self, point, contents):
...
-            if locations[0] > 0:
-                self.completions = []
-                prev_char = view.substr(
-                    sublime.Region(locations[0] - 1, locations[0]))
-                if prev_char not in autocomplete_triggers:
-                    return None
+    def is_after_trigger_character(self, location):
+        if location > 0:
+            prev_char = self.view.substr(
+                sublime.Region(location - 1, location))
```
rwols: You can use `sublime.Region(location - 1)` I believe.

main.py
```diff
@@ -1557,20 +1605,30 @@ def format_completion(self, item) -> 'Tuple[str, str]':
...
-        items = response["items"] if isinstance(response,
-                                                dict) else response
-        self.completions = list(self.format_completion(item) for item in items)
-        self.run_auto_complete()
+        if self.state == CompletionState.REQUESTING:
+            items = response["items"] if isinstance(response,
+                                                    dict) else response
+            names = list(item.get('label') for item in items)
+            debug('got completions', names)
+            self.completions = list(self.format_completion(item) for item in items)
```
rwols: Will you be doing anything with the `sortText` property of the CompletionItem or will you let ST sort the results? The `clangd` server uses `sortText` to sort completions by priority, so you'll see properties and methods first, base classes somewhere middle, keywords last.

main.py
```diff
@@ -1511,43 +1512,90 @@ def show_hover(self, point, contents):
             max_width=800)
 
 
-class CompletionHandler(sublime_plugin.EventListener):
-    def __init__(self):
-        self.completions = []  # type: List[Tuple[str, str]]
-        self.refreshing = False
+class CompletionState(object):
```
rwols: I had some trouble seeing "old" completions when I typed fast, I hope this class fixes that.

main.py
deathaxe: You can even call `self.view.substr(location - 1)` as `substr` understands both a `Region` and a `Point`

main.py
```diff
@@ -1605,7 +1605,7 @@ def format_completion(self, item) -> 'Tuple[str, str]':
             insertText = item.get("insertText")
         if insertText[0] == '$':  # sublime needs leading '$' escaped.
             insertText = '\$' + insertText[1:]
-        return ["{}\t{}".format(label, detail), insertText]
+        return "{}\t{}".format(label, detail), insertText
```
deathaxe: Not sure, whether it's to be handled by LSP or the pyls, but `detail` returns `def function` for a function completion, which makes completions look a bit too verbose:

function    _def function_

main.py
tomv564: Interesting, is there a way to control what ST does with the completions?

main.py
tomv564: This would be something for pyls to look at.

tomv564:
Is anyone interested in giving this branch a test? I'm not 100% convinced it's working correctly.

Example: sublime still shows `for` snippet completion sometimes when you type `f`, instead of pyls's completions containing `from`

main.py
tomv564: In this case the behaviour would be as expected, but I'll add this fix anyways.
Is this documented somewhere?

rwols:
> Is anyone interested in giving this branch a test?

Yes, I'm gonna give this a spin.

main.py
deathaxe: No there is not. ST merges and sorts all items it gets internally.

rwols: Not immediately, it just presents that list as-is if the user hasn't typed more characters yet, so there might be some benefit to using `sortText`.

main.py
rwols: From some quick tests using the changes in this PR I don't see this problem anymore, nice job!

main.py
rwols: But it could slow down LSP a lot if there are tons of completion items, you'll have to experiment with it.

rwols: For reference: https://github.com/llvm-mirror/clang-tools-extra/blob/c543235c134fbb02fb630eaa53661b188068b267/clangd/ClangdUnit.cpp#L316-L317

main.py
```diff
@@ -1510,42 +1513,87 @@ def show_hover(self, point, contents):
...
+                self.do_request(prefix, locations)
+                self.completions = []  # type: List[Tuple[str, str]]
 
-            if locations[0] > 0:
-                self.completions = []
-                prev_char = view.substr(
-                    sublime.Region(locations[0] - 1, locations[0]))
-                if prev_char not in autocomplete_triggers:
-                    return None
+            elif self.state == CompletionState.REQUESTING or self.state == CompletionState.CANCELLING:
```
deathaxe: How about: `elif self.state in (CompletionState.REQUESTING, CompletionState.CANCELLING)`

deathaxe: Hmm. Seems quite easy to get the state machine stalled in CANCELLING state, once a language-server does not respond. From this point I'll never get any completion any more.

deathaxe:
Furthermore there seems to be a general issue with requesting completions. Sometimes a race condition between the onChange change notification and the completion request occures. If onChange is sent later then the completion response, python-language-server returns ArgumentError as it does not yet know about the content of the edited line.

On the other hand the onChange triggers diagnostics scan and causes python-language-server to delay completions as it is busy with linting. Especially mypy plugin causes heavy delays.

main.py
tomv564: Nice code suggestion. Why does the language server not respond?

tomv564:
@deathaxe thanks for testing!
I too have seen pyls complain about some index being out of range. didChange is always called just before requesting completions, in what case does the language server receive it _after_ the completions request?

In a discussion about the javascript-typescript server's completion performance the idea of processing diagnostics from a queue after a small delay was raised. This is worth proposing for pyls as well.

main.py
tomv564: @rwols good to know!

deathaxe:
I didn't investigate too much, but I think the race condition is related to `view.on_modified_async()' handler, which is used to push changes to the language server, while the `view.on_query_completions()` is a synchronous event. Therefore in some situations the async on_modified handler might be delayed by other operations taking place in the background. So at the point of time, the completion is requested, the queue you try to purge might still be empty. I'd suggest to use synchronous handlers in `DocumentSyncListener` as each event handler doesn't take any time to be processed. Some of the operations to read the view content might even take longer due to inter process communication.

>  the idea of processing diagnostics from a queue after a small delay was raised

I hate delays. They are caused by processing anyway. But creating a multithread or multiprocess architecture with a pool of worker threads/processes might be a good idea for pyls to process multiple requests/notifications in parallel. This might help much as pyls may already be busy with multiple linters run after each other while a completion request is received at the moment.

This requires to categorize the received requests/notifications some how. A new didChange request would need to be queued, but not processed until all pending operations on the last version of the document have finished. A new request such as a completion or signatureHelp might be added to the queue and processed immediately to respond as fast as possible, while diagnostics may still be in process. 

This requires at least one dedicated thread/process to handle the synchronization events such as didOpen/didChange/... and one dedicated thread/process to handle requests. A new didChange might would then be queued and being processed as soon as the running requests queue is empty.

deathaxe:
here's a little implementation to get a reference how quick jedi really is without LSP overhead.

```python
# -*- encoding: utf-8 -*-
import os
import sys

import sublime
import sublime_plugin


sys.path.insert(0, os.path.join(os.path.dirname(__file__), 'lib'))
from jedi.api import Script
sys.path.pop(0)


def format_completion(complete):
    """ Returns a tuple of the string that would be visible in
    the completion dialogue and the completion word

    :type complete: jedi.api_classes.Completion
    :rtype: (str, str)
    """
    display, insert = complete.name + '\t' + complete.type, complete.name
    return display, insert


class EventListener(sublime_plugin.EventListener):

    def on_load_async(self, view):
        # just load to trigger interpreter and create cache
        self._script(view, 0)

    def on_query_completions(self, view, prefix, locations):

        point = locations[0]
        if not view.match_selector(point, "source.python"):
            return None

        script = self._script(view, point)
        completions = script.completions()
        return (
            [format_completion(complete) for complete in completions],
            sublime.INHIBIT_WORD_COMPLETIONS
        )

    def _script(self, view, point):
        path = view.file_name() or ''
        region = sublime.Region(0, view.size())
        source = view.substr(region)
        line, column = view.rowcol(point)
        return Script(source, line + 1, column, path=path)
```

PR #48, opened 2017-08-25
Use classes Point and Range to translate between ST and LSP

deathaxe:
This PR intents to extend the object oriented approach already used for several things to translate between language-server-protocol's and Sublime Text's text position representations.

1. Introduce a `class Point` which represents a `(row, col)` pair.
2. Use it to define the existing `class Range`.
3. Use both classes to translate between protocol specific locations/ranges and ST.

All protocol specific stuff is packed into these two classes.
By using those objects, the number of required code lines for translation may be reduced.


deathaxe:
What's the difference between _pyls-mypy_ and _travis-mypy_ ? Even though I don't get any warnings or errors locally, travis built checks tend to fail.

tomv564:
Sorry about that, I created an issue to investigate it on `pyls-mypy` but I suspect the plugin doesn't take mypy.ini into account.
You could run `mypy main.py` to check before pushing, not the greatest workflow :/

rwols:
> You could run mypy main.py to check before pushing, not the greatest workflow :/

Maybe add a `.pre-commit` bash script at the top level and instruct contributors to do this:

```
$ ln -s .pre-commit .git/hooks/pre-commit
```

In `.pre-commit`, you can put something like this:
```
root=$(git rev-parse --show-toplevel)
mypy "$root"/main.py
```

tomv564:
@rwols thanks for the idea, this should be part of contributor documentation

@deathaxe pyls-mypy _is_ reporting these errors for me, would you mind helping troubleshoot?
From the LSP directory, on this branch, can you run:
`mypy main.py`
and if you have git bash:
`mypy --command "$(cat main.py)"`

You have no local changes against `mypy.ini`? What version of mypy do you have?

main.py
```diff
@@ -191,31 +191,37 @@ def to_payload(self):
         return r
 
 
+class Point(object):
```
rwols: Great idea!

deathaxe:
I think I found the reason. I don't use a dedicated project/workspace for LSP, but use a Project with _Data/Packages_ folder as "root". This way I can quickly jump between multiple packages or look up things quickly. 

If I open the LSP package in ST only, mypy shows the error results as well. Seems like an issue with the lookup paths of ini files? 

The python-language-server should start searching in document directory and walk up step by step until it finds an ini file. Seems pyls-mypy has some trouble with that.

deathaxe:
Rebased the work on the current master

tomv564:
> The python-language-server should start searching in document directory and walk up step by step until it finds an ini file. Seems pyls-mypy has some trouble with that.

Glad you found how to get the type checker to work. It's actually mypy that autodetects the ini/cfg files.

tests/test_point_range.py
```diff
@@ -0,0 +1,89 @@
+"""
+Tests for Point.
+
+To run this unit test install UnitTesting package and choose
+"UnitTesting: Test current project" command.
+"""
```
tomv564: I understand we may want to test _in_ sublime using the UnitTesting package eventually, but these tests should work with the standard python tools.

tests/test_point_range.py
```diff
@@ -0,0 +1,89 @@
...
+    def test_valid_point(self):
+        lsp_point = {'line': 10, 'character': 4}
+        point = Point.from_lsp(lsp_point)
+        self.assertEqual(point.row, 10)
+        self.assertEqual(point.col, 4)
+
+    def test_none_point(self):
+        # expect TypeError if range object is None
+        with self.assertRaises(TypeError):
+            Point.from_lsp(None)
```
tomv564: The tests against TypeError and KeyError confirm nothing about what LSP does, only built-in python behaviour. Please remove them.

main.py
```diff
@@ -191,31 +191,63 @@ def to_payload(self):
...
+    def __init__(self, row: int, col: int) -> None:
+        if not isinstance(row, int):
+            raise ValueError('row')
+        if not isinstance(col, int):
+            raise ValueError('col')
```
tomv564: I find the noise-to-utility ratio of this error low - it does not save the user from anything and chances are high that a developer will find the invalid range quickly.

main.py
```diff
@@ -229,8 +261,10 @@ def __init__(self, message, range, severity, source, lsp_diagnostic):
     @classmethod
     def from_lsp(cls, lsp_diagnostic):
         return Diagnostic(
-            lsp_diagnostic.get('message'),
-            Range.from_lsp(lsp_diagnostic.get('range')),
+            # crucial keys
+            lsp_diagnostic['message'],
+            Range.from_lsp(lsp_diagnostic['range']),
```
tomv564: Nice clarification

main.py
```diff
@@ -988,12 +1013,8 @@ def handle_response(self, response):
 
     def on_symbol_selected(self, symbol_index):
         selected_symbol = self.symbols[symbol_index]
-        location = selected_symbol.get("location")
-        start = location.get("range").get("start")
-        end = location.get("range").get("end")
-        startpos = self.view.text_point(start.get('line'), start.get('character'))
-        endpos = self.view.text_point(end.get('line'), end.get('character'))
-        region = sublime.Region(startpos, endpos)
```
tomv564: Great cleanup!

main.py
```diff
@@ -1362,36 +1379,20 @@ def start_server(server_binary_args, working_dir):
...
-    if point:
-        (row, col) = view.rowcol(point)
-    else:
-        view.sel()
-    uri = filename_to_uri(view.file_name())
-    position = OrderedDict(line=row, character=col)
-    dp = OrderedDict()  # type: Dict[str, Any]
-    dp["textDocument"] = {"uri": uri}
-    dp["position"] = position
-    return dp
```
tomv564: There was some filthy code in here :)

tests/test_point_range.py
```diff
@@ -0,0 +1,89 @@
...
+from LSP.main import Range
+
+
+class test_point(TestCase):
+
+    def test_valid_point(self):
+        lsp_point = {'line': 10, 'character': 4}
+        point = Point.from_lsp(lsp_point)
+        self.assertEqual(point.row, 10)
+        self.assertEqual(point.col, 4)
```
tomv564: Could test `to_lsp` at the same time
`self.assertEqual(lsp_point, Point.from_lsp(lsp_point).to_lsp())`

tests/test_point_range.py
```diff
@@ -0,0 +1,89 @@
...
+Tests for Point.
+
+To run this unit test install UnitTesting package and choose
+"UnitTesting: Test current project" command.
+"""
+
+from unittest import TestCase
+
+from LSP.main import Point
+from LSP.main import Range
```
tomv564: Was mypy working in this file? It should prevent you from sending `None` in for dict, for example?

unittesting.json
```diff
@@ -0,0 +1,9 @@
+{
+    "tests_dir" : "tests",
+    "pattern" : "test*.py",
+    "async": true,
+    "deferred": false,
+    "verbosity": 5,
+    "output": "<panel>",
+    "capture_console": true
+}
```
tomv564: Please remove until this is needed

tests/test_point_range.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class test_point(TestCase):
```
tomv564: Tests for `__repr__` would be worthwhile addition

tests/test_point_range.py
deathaxe: I never leave ST for developement, so this is the best way I found.

tests/test_point_range.py
deathaxe: How is your question related with line 11?

tests/test_point_range.py
deathaxe: This is not completely true. As you call `.get()` methods for most/all dict objects of a language-server-respond even crucial elements such as the `line` and `character` part of a point, the resulting class would contain None in that place with unpredictable results. You may trust a language-server to always return valid objects, but I don't.

The method `from_lsp` does not always need to raise an exception. It would not if a None check was added to it. With other words it could return None instead of raising the exception. This check ensures the exception to be raised.

main.py
deathaxe: You know you handle responses from an external program not under your control?

PR #49, opened 2017-08-26
Exclude some stuff from sublime-package files

deathaxe:
Some of the files and folders of the repository are meant for development only and don't need to be part of a sublime-package file. We may therefore add an .gitattributes file to exclude them from being archived/exported. This may reduce the package size somewhat and makes the package content look clean.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,19 @@
+# Stuff not being exported to ZIP and therefore
+# prevented from being packed into LSP.sublime-package
+
+## git
+.github/          export-ignore
+gh-pages/         export-ignore
+*.git             export-ignore
+*.gitignore       export-ignore
+*.gitattributes   export-ignore
+
+## unit testing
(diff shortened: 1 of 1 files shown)


tomv564:
Nice idea, is this documented somewhere?
I see some entries that don't apply to LSP, could you remove them?

deathaxe:
> I see some entries that don't apply to LSP

Which ones? Only some default git/github files/dirs , unittesting and your inis

Most such predefined files include tons of more unrelated stuff. Everything within the file is there for a certain reason.

tomv564:
Answered my own question - Package control downloads release zips from GitHub.
You named the entries I was thinking of, `.github` and `gh-pages` and unittesting.
Perhaps `docs` should be added?

rwols:
Why was this closed? A .gitattributes file can benefit all users that use Package Control.

PR #50, opened 2017-08-27
Workaround for SublimeTextIssues/Core#1877

rwols:
On some occasions the environment variables are loaded *after* `plugin_loaded` is invoked. So the `PATH` variable may not be up-to-date and consequently the path to a server may not be found. This PR is a workaround for that. It waits 2 seconds and then initializes the servers, hoping that the environment variables are loaded in those 2 seconds.

Also see: https://github.com/SublimeTextIssues/Core/issues/1877


tomv564:
Thanks for identifying what sounds like a blocking bug. Great you reported it to the ST developers, as I understand build 3142 is a dev build. 

I'm hesitant to accept a 2-second penalty for all OS-X users. Is it okay if we keep this PR as documentation for your workaround, and hope for the next dev build to fix the regression?

If in the future a larger number of OS-X users ends up stuck on that build, we could perhaps active the fix only when running in build 3142.

rwols:
Sure, let's wait and see what happens 👍

randy3k:
An alternatively solution is to execute the commands via a login shel, `/bin/bash -l -c 'pyls'` for example.

rwols:
It's not really a huge issue, I can just run "LSP: Restart Servers" from the command palette and by that time my env vars are loaded. Maybe put that in a FAQ section for OS X users.

tomv564:
@rwols I see some comments were added to the issue, it does not look like sublime will fix this immediately.
I've avoided the PATH issue on OS-X by launching with `subl` from the command line - does that circumvent your issue? There is also the old https://github.com/int3h/SublimeFixMacPath package.

rwols:
After trying SublimeFixMacPath my path seems to be correct from the get-go 👍

tomv564:
Great, then it might be an older known issue with a reliable-ish workaround.
I added some links and explanation in a new troubleshooting section of the documentation here: https://github.com/tomv564/LSP/commit/0c55bea84f592d7e114e363050e95d1496269040

I hope that's a sufficient replacement for this pull request?

braver:
Just going to throw in another perspective. 

Do you really need all the environment variables, or do you just need to know how to call the server? If it's the latter, why not let me configure the path to the server? Just let me tell LSP where it's at and it removes the need to wait for anything. Yes it's more setup, but I'd rather do a little setup than use a workaround like having to start from command line, or install another package that isn't maintained and seems a bit iffy to begin with. It should be pretty easy to make this optional too.

randy3k:
Ya, I remember I ran into some issues when using that fix and the issue was very subtle and hard to be discovered. After that, I have move it and just try my best to configure each package.

tomv564:
Since we have a few better performing workarounds (`FixMacPath`, hardcoding the command, `launchctl setenv`), I will close this PR now.

PR #63, opened 2017-09-01
didOpen requires TextDocumentItem.version

randy3k:
https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#textdocumentitem


xldenis:
I've had the same isssue when attempting to use this package.

tomv564:
Thanks!

PR #64, opened 2017-09-01
add Julia settings

randy3k:


randy3k:
well, don't merge now. It is not working smoothly as expected.

randy3k:
close for now. It turns out put we may need a script to launch the server https://github.com/JuliaEditorSupport/LanguageServer.jl/pull/177.

PR #66, opened 2017-09-02
some completions do not have details

randy3k:
as title


tomv564:
This reminds me, I had some changes to use 'kind' if details was not provided. Will merge this first, thanks!

PR #68, opened 2017-09-02
Fix bug in format_completion function

rwols:
When `insertText` is not empty, it should be used irregardless of wether `insertTextFormat` is `PlainText` or `Snippet`. In fact since Sublime handles snippets transparently, we don't even have to care about `insertTextFormat`.


tomv564:
Thanks!

PR #70, opened 2017-09-02
Use completion kind as detail if no detail provided

tomv564:


main.py
```diff
@@ -1614,6 +1638,10 @@ def format_completion(self, item) -> 'Tuple[str, str]':
         # Sublime handles snippets automatically, so we don't have to care about insertTextFormat.
         label = item.get("label")
         detail = item.get("detail")
```
rwols: Should have a default of `None` here, then?

main.py
```diff
@@ -95,6 +95,30 @@ class SymbolKind(object):
...
+    Value = 12
+    Enum = 13
+    Keyword = 14
+    Snippet = 15
+    Color = 16
+    File = 17
+    Reference = 18
+
+
+completion_item_kind_names = {v: k for k, v in CompletionItemKind.__dict__.items()}
```
rwols: Why not just write it as a dict in the first place?

randy3k: or [enum](https://docs.python.org/3.4/library/enum.html).

tomv564: Had wondered about Enum before, but documentation said "new in version 3.4".
Tried it now, getting an ImportError:
```
  File "/Users/tomv/Library/Application Support/Sublime Text 3/Packages/LSP/main.py", line 7, in <module>
    from enum import Enum
ImportError: No module named 'enum'
```

randy3k: Ai...forget that 3.3 doesn't support enum.

tomv564: It was easiest to paste and fix the list from the spec or typescript and then add the one-liner to create the reverse lookup. 
I agree the lookup is the only thing we use right now, but I'd like an enum-like lookup for future functionality.

rwols: Fair enough 👍

PR #71, opened 2017-09-03
enable word completions and explicit completions

randy3k:
why disable them?

### DIFF
--- a/main.py
+++ b/main.py
@@ -1589,10 +1589,7 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
             elif self.state == CompletionState.APPLYING:
                 self.state = CompletionState.IDLE
 
-            return (
-                self.completions,
-                sublime.INHIBIT_WORD_COMPLETIONS | sublime.INHIBIT_EXPLICIT_COMPLETIONS
-            )
+            return self.completions
 
     def do_request(self, prefix, locations):
         self.next_request = None


tomv564:
I think the idea is to prevent the "two-step" completion list (word/explicit first, followed by a "jump" in the completion list as the sublime completions are triggered).

That said, word completions are great and if the two completion sources are presented in a non-disruptive manner then I would love to remove the flags!

Looking at two other packages, the Microsoft Typescript package disables word and explicit completions, and the Anaconda package offers a `hide_snippets_on_completion`.

Is it okay if we ask some users to give feedback about how this works so we can establish a good default? Perhaps we should add an `only_show_lsp_completions` setting?

rwols:
I would like such a setting, yes.

tomv564:
I just realised that if you set `complete_all_chars` to false, there is no reason to hide sublime's completions until you press a trigger character.
We should definitely allow word/snippet completions in this case *unless* an LSP completion is being requested.

randy3k:
I would say we should enable them **even** LSP completion is requested. I don't see an obvious "jump" at all.

deathaxe:
_Jedi - Python Autocompletion_ packages provides such a setting, too. Anaconda seems to have some trouble with auto completions sometimes which results in flickering of the completion overlay. Therefore completions should kept hidden until the language-server responded. If he did a setting could decide whether to merge results from language-server with explicit and word completions or display server results only.

rwols:
On OS X when the completion widget is "updated" with more items the transition is smooth. On windows and Linux there is a slight flicker effect when updating completions.

deathaxe:
Here is how Jedi package handles completion settings. Find this best to have the choice between every possibility.

```
    // Which completions to show:
    // "all" - both jedi's and sublime's
    // "jedi" - only jedi's
    // "default" - only jedi's if it has something to show, otherwise sublime's
    "sublime_completions_visibility": "default",
```

Besides any settings, the results should returned as follows to prevent flickering

```python
            return (
                self.completions,
                0 if self.state == CompletionState.IDLE
                else sublime.INHIBIT_WORD_COMPLETIONS | sublime.INHIBIT_EXPLICIT_COMPLETIONS
            )
```

randy3k:
@rwols that why I didn't think it was an issue to show all completions.

tomv564:
A slight flicker is not a problem - as long as the user's selection into the completion list is not lost then using both built-in and lsp completions sounds a good default.

@randy3k, would you be up for applying @deathaxe's suggestion *for returning completions*, with a `only_show_lsp_completions` setting that preserves the current behaviour?

randy3k:
I support the more conservative default of showing all completions.

tomv564:
Yes, we want to show all completions by default.
But I would also like a setting that returns the INHIBIT flags if the user only wants to see LSP completions, as part of this PR. 
Alternatively, I can merge your PR and provide the setting later.

PR #73, opened 2017-09-03
Add textDocument/willSave and textDocument/willSaveWaitUntil

rwols:
For willSaveWaitUntil, it is required to *wait* in the on_pre_save ST callback in order to synchronize formatting&saving. For this reason the functionality of a request has been enhanced with a blocking mode (by default set to False) that blocks the requesting thread until the server responds, and invokes the handler on the requesting thread. A timeout of six seconds is in place in case the server doesn't respond.

Please don't merge this yet as it's a WIP. pyls doesn't understand willSaveWaitUntil and clangd just hangs.


tomv564:
This will be a nice mechanism to support, thanks for taking on the challenge!
Are any language servers using `willSave`, and if so do you have some examples?

rwols:
To be honest I have no idea if any server uses `willSave` in any meaningful way, but it's good to have this implemented anyway. My ulterior motive is to also make clangd responsive to `willSaveWaitUntil`. I'm already busy with snippet functionality for clangd :)

rwols:
@tomv564, are you in the process of refactoring now? I will probably modify this PR to only have the foundation for blocking requests once the refactoring is done.

tomv564:
Yes, I'm in the process of splitting main.py up now. I'd like to have an accessible working example of a blocking request before merging the feature - will I need to build clangd to test the feature?

rwols:
> will I need to build clangd to test the feature?

Haven't implemented willSaveWaitUntil (a.k.a. "format my doc when I save") yet, but signatureHelp is merged in now.

rwols:
> I'd like to have an accessible working example of a blocking request before merging the feature

There are a few other requests besides willSaveWaitUntil that can benefit from blocking behavior:

* shutdown: we must wait for the server's confirmation before killing the pipe.
* format doc and format range: there is no point in doing anything else while the server is formatting the code.

Issue #69, opened 2017-09-02
[Feature request] toggle language servers


randy3k:
It should be nice to be able to enable/disable specific servers.

tomv564:
Agree, this is a must-have as currently LSP is an always-on setup.
What solutions do you think will work best?
* A command that writes a project-specific setting to disable a language server by name?
* The ability to configure a language server at the project level *instead of* at the package level?

randy3k:
I guess there should be an option to disable a server globally via `LSP.sublime-settings` and also a project specific setting to enable/disable certain servers.

tomv564:
The configuration to support this was merged in #74.
There is some discussion about what the default language server configuration should look like in #76.
It would be nice to say *Enable* or *Disable* instead of Toggle, but unsure how specific the name of a command in the quick panel can be? (The "Restart Servers" command should perhaps also say "LSP - Restart pyls" for example)

randy3k:
Enabling ane disabling the servers do make more sense. I also found "Restart Servers" confusing, does it restart all running servers or just restart the server of the current window!?

May be this?

- LSP: Enable Server Globally
- LSP: Disable Server Globally
- LSP: Enable Server for Current Project
- LSP: Disable Server for Current Project
- LSP: Restart Server of Current Window

I also not that you are loading the project settings from the key `LSP` of `view.settings`. I think it should be `window.project_data`.

tomv564:
Yes, those look like good suggestions, including using `window.project_data`, thanks!
I've started some work on making these commands part of a first-time setup flow.

tomv564:
#119 added Enable/Disable language server Globally / in Project.
Please let me know if it doesn't work as expected!

Issue #51, opened 2017-08-28
Quick switch project should restart language server


tomv564:
It currently re-uses the existing client for a given language, as the window hasn't changed.

tomv564:
After watching the logs with `sublime.log_commands(True)`, I am not sure how this event can be observed directly.

rmccue:
Related question on the forum: https://forum.sublimetext.com/t/is-there-an-on-project-load-event/29660/2

tomv564:
Thanks for the tips, #95 should solve this.

PR #74, opened 2017-09-03
Per-window/project configuration

tomv564:
Should provide functionality for:
* Passing `initializationOptions` to language servers
* Disable language servers / enable per-project (#69)
* Per-project configuration of `initializationOptions`
* Initial sending of per-project `settings` via `workspace/didChangeConfiguration`

To do:
- [x] Send options in initialization
- [x] Verify that options can be configured in a compatible way for existing servers (eg. RLS, eslint)
- [x] Don't replace but extend global options with project options
- [x] Send didChangeConfiguration
- [x] Verify didChangeConfiguration works with an existing server
- [x] Add documentation

Tested this using a stdio-supporting fork of vscode-eslint (https://github.com/tomv564/vscode-eslint/tree/add-stdio-support), which refused to start up without initializationOptions (needed `nodePath`) or an initial didChangeConfiguration (with empty settings)

The python language server will get configuration support in PR https://github.com/palantir/python-language-server/pull/118

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -188,6 +188,40 @@ These can be customized as follows by adding an override in the User LSP.sublime
 * `scopes` - add language flavours, eg. `source.js`, `source.jsx`.
 * `syntaxes` - syntaxes that enable LSP features on a document, eg. `Packages/Babel/JavaScript (Babel).tmLanguage`
 * `languageId` - used both by the language servers and to select a syntax highlighter for sublime popups.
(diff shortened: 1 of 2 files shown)


randy3k:
I think this is also time to address #51 while you are working on window related code..

tomv564:
@randy3k good point, I'll have a look if this doesn't get out of hand :)

PR #80, opened 2017-09-07
Add Reason/OCaml lsp settings

chenglou:
I'm not sure why this doesn't activate for ocaml files (ocaml-language-server works for reason and both, so it shouldn't be that side's error), but for reason files this works.

Test:

```
git clone https://github.com/freebroccolo/ocaml-language-server.git
cd ocaml-language-server
npm install
npm install -g .
```

Then use this diff and open up a foo.re file and write

```reason
let a: int="hi";
```

And see the correct error at the bottom


tomv564:
Thanks for the contribution!
There is some discussion about the onboarding experience and if these configurations should be disabled by default, but let's add it to the default settings for now.
Documentation for this configuration already exists at in https://github.com/tomv564/LSP/blob/master/docs/index.md, could you review those settings and add a link to your sublime-reason package as part of this PR?

chenglou:
Done! It's now reduced to just two sentences =D

tomv564:
Thanks!
If you want to troubleshoot plain ocaml usage we could open a new issue.

PR #81, opened 2017-09-08
Use one project path instead of common prefix

### LINKED ISSUE #79: get_project_path() incorrectly determines path
If you have multiple sibling directories in a workspace with common prefix characters, it will incorrectly use a partial path for this.

I have two directories open, which are `~/Projects/example1` and `~/Projects/example2`. `os.path.commonprefix()` will hence return `~/Projects/example`, which is not a directory that exists. (The Python docs note that `os.path.commonpath()` should be used instead.)

This approach is generally problematic anyway. I have two subdirectories deep inside a common directory, and even without the partial filename problem, this is still problematic for me as it causes **a lot** more files than I care about to be indexed. This causes the initial parsing to take roughly *forever*.

VS Code is currently working on adding multi-root support (https://github.com/Microsoft/vscode/issues/28344), so I suspect they'll add support to the Language Server protocol at some point. The current behaviour in the beta is just to index the first directory in the project. IMO, this is the best behaviour until the protocol actually supports multi-root.

rmccue:
Fixes #79.

This changes from using the common prefix (which could be a partial directory name) to using only the first folder in a project. Additionally, it fixes the `is_in_workspace` check to avoid the same problem.

This removes support for multiple folders (in a way; it will now only index one folder instead of the common parent of all folders), so probably warrants a major version bump.

### DIFF
--- a/main.py
+++ b/main.py
@@ -579,10 +579,10 @@ def get_project_path(window: sublime.Window) -> 'Optional[str]':
 def get_common_parent(paths: 'List[str]') -> str:
     """
     Get the common parent directory of multiple paths.
-    """
-    if 'commonpath' in dir(os.path):
-        return os.path.commonpath(paths)
 
+    Python 3.5+ includes os.path.commonpath which does this, however Sublime
+    currently embeds Python 3.3.
+    """
     return os.path.commonprefix([path + '/' for path in paths]).rstrip('/')
 
 
--- a/main.py
+++ b/main.py
@@ -576,12 +576,33 @@ def get_project_path(window: sublime.Window) -> 'Optional[str]':
         return None
 
 
+def get_common_parent(paths: 'List[str]') -> str:
+    """
+    Get the common parent of multiple paths. Attempts to use native method
+    if available.
+    """
+    try:
+        return os.path.commonpath(paths)
+    except:
+        pass
+
+    groups = [s.split(os.path.sep) for s in paths]
+    min_len = min(len(group) for group in groups)
+    common = []
+    for i in range(min_len):
+        bits = set(group[i] for group in groups)
+        if len(bits) != 1:
+            break
+        common.append(bits.pop())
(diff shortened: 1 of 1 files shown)


rmccue:
Rats, I missed that `os.path.commonpath` was only added in 3.5, but Sublime embeds 3.3. I'll fix this up.

main.py
```diff
@@ -576,12 +576,33 @@ def get_project_path(window: sublime.Window) -> 'Optional[str]':
...
+
+    groups = [s.split(os.path.sep) for s in paths]
+    min_len = min(len(group) for group in groups)
+    common = []
+    for i in range(min_len):
+        bits = set(group[i] for group in groups)
+        if len(bits) != 1:
+            break
+        common.append(bits.pop())
+    return os.path.sep.join(common)
```
tomv564: Thanks for thinking of `is_in_workspace`! I have two requests:
* We may want to change to `commonpath` if python 3.5 becomes available, but not run through a try-catch on every call. Can you make a comment about common path's existence instead?
* Your code looks very complete, but we have a simpler use case of only expecting directories as output. Can we use `dirname` on the output of `commonprefix` instead?

rmccue: > Can we use `dirname` on the output of `commonprefix` instead?

Good catch; I didn't think about this fully. I think this should work if we also append `os.path.sep` to the project path, since:
```
commonprefix(['/usr1/', '/usr2/']) -> /usr
dirname('/usr') -> '/'

commonprefix(['/usr/', '/usr/a/']) -> /usr/
dirname('/usr/') -> '/usr'
```

I'll update.

rmccue: > to the project path

(Actually, should be safe to add to all paths.)

rmccue:
Updated for both points. I can also remove the native handler if you'd rather just stick to one code path instead.

rmccue:
Done 👍

tomv564:
Thanks, great work!

PR #83, opened 2017-09-08
Rewrite native goto definition

### LINKED ISSUE #78: [Feature request] Override ST's default goto definition?
I was wondering why ST's Goto definition didn't work. I didn't get that LSP uses its own system and doesn't plug into ST's. Would it be possible for Goto definition to first consult LSP, and if a definition isn't found, fall back to the default string-based behavior?

rmccue:
Fixes #78.

Rather than having a custom Go To Definition for LSP, this overrides the default for views that we support.

This only operates on the context menu item, the menu item (Goto -> Goto Definition) and keyboard shortcut run the `goto_definition` window (not view) command instead, but not sure the best way to translate the cursor position into our desired command yet.

(Given that #78 is still open, I figured this was desired, but wanted to check before I invest more time into fixing up the rest.)

### DIFF
--- a/main.py
+++ b/main.py
@@ -1558,6 +1558,19 @@ class Events:
             for listener in cls.listener_dict[key]:
                 listener(*args)
 
+def is_applicable(settings):
+    syntax = settings.get('syntax')
+    return syntax and is_supported_syntax(syntax)
+
+class DefinitionHandler(sublime_plugin.EventListener):
+    def on_text_command(self, view, command_name, args):
+        if not is_applicable(view.settings()):
+            return None
+
+        if command_name != 'context_goto_definition':
+            return None
+
+        return ('lsp_symbol_definition', args)
 
 class HoverHandler(sublime_plugin.ViewEventListener):
     def __init__(self, view):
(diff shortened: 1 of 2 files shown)


tomv564:
Great to see an example of how this could be accomplished!
We can discuss the proposal in the issue, as for your solution I think we are missing some cases where we can directly go to the built-in go to definition:
* When a language server has been not started or disabled
* When a language server does not support go to definition.

main.py
```diff
@@ -1558,15 +1558,27 @@ def publish(cls, key, *args):
...
+
+class DefinitionHandler(sublime_plugin.EventListener):
+    def on_text_command(self, view, command_name, args):
+        if not is_applicable(view.settings()):
+            return None
+
+        if command_name != 'context_goto_definition':
+            return None
+
+        return ('lsp_symbol_definition', args)
```
rwols: Like @tomv564 mentioned we definitely need to check wether the server is active and wether the server supports goto symbol def, otherwise this'll do nothing.

rmccue:
My bad; I'd misinterpreted "The fallback already exists" on #78 as "the command will fall back to native behaviour". :)

We should be able to use the [is_enabled check](https://github.com/tomv564/LSP/blob/030f8e3d357955b010bf9a7285df7af2c55aa3b9/main.py#L1086-L1089) from the `lsp_definition_command`, I guess.

tomv564:
@rmccue, as this PR has lost momentum I'm going to close it and document your approach in the related issue. I still think the Context Menu item is a UX problem, but hope for a way to replace it without using an `EventListener`.

PR #93, opened 2017-09-12
Lsp only completion setting

tomv564:


randy3k:
It seems that it is bugged. The first attempt showed only `Class`; the second attemp showed both LSP completions and word completions; the third attemp showed only `Class` and so on.

sep-12-2017 21-35-22

Issue #67, opened 2017-09-02
Only show diagnostic panel when uri matches current view


randy3k:
In Julia server, it starts by linting all files at the current directory. It may make sense to only show the panel when the uri of notification matches the current view. Otherwise, it will popup and refresh the panel multiple times.

tomv564:
The panel refresh in the current implementation is quite slow and distracting. Improving this would give all users are better experience.
Are you closing the panel every time and getting new lint results immediately after?
About matching URI - I think LSP should offer a way to see that a change in file A broke file B, even if file B is not the current view.

randy3k:
> Are you closing the panel every time and getting new lint results immediately after?

Yes, as I may be mentioned earlier, julia server may be slow sometimes (if there are packages need to be recompiled). I close the panel when I see it, then the panel pops up again a few second later when new notification arrives.

tomv564:
That sounds annoying, although not a fix I would suggest setting `auto_show_diagnostics_panel` to false and using the keyboard shortcut (super+shift+M / ctr+alt+M) to open the panel manually once the language server has started.

Besides improvement to the panel refresh logic, I am open to other suggestions on a clever panel-opening mechanism - although preferably without hiding errors from users who want to see them.

tomv564:
To bring this issue to some conclusion:
* The diagnostics refresh logic has come up in issue #84, so let's talk about panel improvements there.
* LSP displays diagnostics for all files in the workspace - we could configure this to only show diagnostics for *opened* files in the workspace. 

@randy3k, does that sound reasonable? I would love a PR implementing a `only_show_diagnostics_for_open_files` (or better named) setting!

deathaxe:
> LSP displays diagnostics for all files in the workspace - we could configure this to only show diagnostics for opened files in the workspace.

Sounds good. I just struggled with the same issue. I have a project which points to Sublime Text Packages path. The diagnostics panel offers diagnostics for all files I recently opened even though I am no longer interested in them. This is really annoying having a list of 30 files with tons of "errors" but not seeing the current ones.

It is especially an issue as nearly none of the packages provides configurations the python-language-server or LSP is able to handle. Therefore tons of linting issues are raised as false positives. The packages just use different rules with longer lines (120 chars) as an example.

tomv564:
I too have noticed that diagnostics remain after I closed some files - the `remove_diagnostics` function should be clearing these but perhaps it's worth adding a "LSP - Clear Diagnostics" command.

[...]

PR #96, opened 2017-09-13
Add setting to hide diagnostics received from open files that aren't open 

tomv564:
For language servers that provide full-project diagnostics at eg. startup, (eg. the julia language server as reported in #67)

TODO:
- [ ] verify this improves the experience
- [ ] document setting

### DIFF
--- a/main.py
+++ b/main.py
@@ -30,6 +30,7 @@ show_view_status = True
 auto_show_diagnostics_panel = True
 show_diagnostics_phantoms = False
 show_diagnostics_in_view_status = True
+only_show_diagnostics_for_open_files = False
 only_show_lsp_completions = False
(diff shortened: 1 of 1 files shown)


randy3k:
Does it mean that I can still see the diagonistics of the unopen files in the panel but the panel will not pop up automatically?

tomv564:
No, we'll need even more logic for that :)

tomv564:
Just to summarise: The desired behaviour is to preserve the auto-open-diagnostics behaviour but only when new diagnostics are received for _opened_ files. 
Seeing as we currently hold the panel open as long as there are diagnostics, this is a significant amount of logic to add for 1 language server's behaviour.
We can re-open this PR in the future, but for now I would like to focus on issues impacting more users.

PR #98, opened 2017-09-13
Fix exception when client does not start properly.

ehuss:


rwols:
This is a pet-peeve of mine but we keep talking about *clients* while we should be talking about *servers*.

tomv564:
@ehuss thanks for your contribution!

This project uses the `mypy` type checker, for which I need to write some contributor documentation.

The `mypy` type checker does not like applying `bool()` to client, but changing to the same check as used in other commands should work:
```
if client and client.has_capability('codeActionProvider'):
    return True
```

main.py
```diff
@@ -1929,7 +1929,7 @@ class LspCodeActionsCommand(sublime_plugin.TextCommand):
     def is_enabled(self, event=None):
         if is_supported_view(self.view):
             client = client_for_view(self.view)
-            return client and client.has_capability('codeActionProvider')
+            return bool(client) and client.has_capability('codeActionProvider')
```
tomv564: Unfortunate typing issue with mypy, but can you change the syntax to:
```
if client and client.has_capability('codeActionProvider'):
    return True
```

tomv564:
I merged the fix manually as I was unable to push the mypy fix to your branch. 
Thanks!

PR #102, opened 2017-09-15
need a space before concatting "Could not start " + config.name

braver:
If a server cannot start, there is no space between the message and the name of the server, resulting in, e.g., `Could not startjsts` instead of `Could not start jsts`

### DIFF
--- a/main.py
+++ b/main.py
@@ -1518,8 +1518,8 @@ def start_client(window: sublime.Window, config: ClientConfig):
 
         client = start_server(expanded_args, project_path)
         if not client:
-            window.status_message("Could not start" + config.name + ", disabling")
-            debug("Could not start", config.binary_args, ", disabling")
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -1518,8 +1518,8 @@ def start_client(window: sublime.Window, config: ClientConfig):
 
         client = start_server(expanded_args, project_path)
         if not client:
-            window.status_message("Could not start" + config.name + ", disabling")
-            debug("Could not start", config.binary_args, ", disabling")
+            window.status_message("Could not start " + config.name + ", disabling")
+            debug("Could not start ", config.binary_args, ", disabling")
```
rwols: Actually, it's not necessary to have the space in the `debug` function-call, it acts like `print`. But for the status message it is indeed necessary.

braver: Got it!

tomv564:
Thanks!

PR #109, opened 2017-09-17
Gives phantoms the same nubbin as build errors

braver:
This is just a very simple copy-paste from Default's exec.py to sync the styling of LSP phantoms with build errors.

before 
screen shot 2017-09-17 at 21 47 44

after
screen shot 2017-09-17 at 21 46 45

The docs suggest giving the phantoms a package specific `id` to help targeting style customisation. However, I think the idea here is to match the built-in phantoms so it's good as is at least for now. It's a shame these styles aren't easily shared. It's also not guaranteed to get out of sync.. ¯\_(ツ)_/¯

### DIFF
--- a/main.py
+++ b/main.py
@@ -1518,8 +1518,8 @@ def start_client(window: sublime.Window, config: ClientConfig):
 
         client = start_server(expanded_args, project_path)
         if not client:
-            window.status_message("Could not start" + config.name + ", disabling")
-            debug("Could not start", config.binary_args, ", disabling")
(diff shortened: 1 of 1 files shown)


tomv564:
Look great to me so far, thanks for the screenshots. 
Let me know if you were planning on doing more work in this PR, or if it should be merged!

braver:
@tomv564 No, I intend to break it up, the other stuff will probably take more time and will not touch the same code at all. So go ahead and merge!

PR #111, opened 2017-09-17
Browse function overloads with the up/down arrow keys

rwols:
~~This is a WIP.~~ ~~Ready to merge (and squash) from my point of view.~~ This is (still) a WIP.

out

out2

### DIFF
--- a/main.py
+++ b/main.py
@@ -189,6 +189,8 @@ class Request:
         r["method"] = self.method
         if self.params is not None:
             r["params"] = self.params
+        else:
+            r["params"] = dict()
         return r
 
 
@@ -235,6 +237,8 @@ class Notification:
         r["method"] = self.method
         if self.params is not None:
             r["params"] = self.params
+        else:
+            r["params"] = dict()
         return r
 
 
@@ -1735,8 +1739,9 @@ class HoverHandler(sublime_plugin.ViewEventListener):
 
 def preserve_whitespace(contents: str) -> str:
     """Preserve empty lines and whitespace for markdown conversion."""
-    contents = contents.replace('\t', '&nbsp' * 4)
-    contents = contents.replace('  ', '\a\a')
+    contents = contents.strip(' \t\r\n')
+    contents = contents.replace('\t', '&nbsp;' * 4)
+    contents = contents.replace('  ', '&nbsp;' * 2)
     contents = contents.replace('\n\n', '\n&nbsp;\n')
     return contents
 
--- a/main.py
+++ b/main.py
@@ -1,6 +1,8 @@
 import html
+import io
 import json
 import os
+import re
 import subprocess
 import sys
 import threading
@@ -479,10 +481,8 @@ class Client(object):
 
                     payload = None
                     try:
+                        debug("got json:", content)
                         payload = json.loads(content)
-                        limit = min(len(content), 200)
-                        if payload.get("method") != "window/logMessage":
-                            debug("got json: ", content[0:limit])
                     except IOError:
                         printf("Got a non-JSON payload: ", content)
                         continue
@@ -490,7 +490,7 @@ class Client(object):
                     try:
                         if "error" in payload:
                             error = payload['error']
-                            debug("got error: ", error)
+                            debug("got error:", error)
                             sublime.status_message(error.get('message'))
                         elif "method" in payload:
                             if "id" in payload:
@@ -500,12 +500,12 @@ class Client(object):
                         elif "id" in payload:
                             self.response_handler(payload)
                         else:
-                            debug("Unknown payload type: ", payload)
+                            debug("Unknown payload type:", payload)
                     except Exception as err:
                         printf("Error handling server content:", err)
 
             except IOError:
-                printf("LSP stdout process ending due to exception: ",
+                printf("LSP stdout process ending due to exception:",
                        sys.exc_info())
                 self.process.terminate()
                 self.process = None
@@ -1927,6 +1927,9 @@ class SignatureHelpListener(sublime_plugin.ViewEventListener):
     def __init__(self, view):
         self.view = view
         self.signature_help_triggers = None
+        self.active_signature = -1
+        self.signatures = []  # type: List[Dict]
+        self._visible = False
 
     @classmethod
     def is_applicable(cls, settings):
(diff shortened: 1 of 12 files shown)


braver:
Nice!

I would advise against taking over the arrow keys like that though. If the user didn't manually open the popup, you shouldn't take over elementary key bindings like those used for navigation etc. Perhaps you could combine them with a modifier?

tomv564:
Will be fantastic to have overload and current parameter support!

Some observations:
* VS Code binds the up/down keys without modifiers. Maybe it's not a big deal?
* I understand it's tempting to fix other stuff "while you're in there", but could you revert or move all changes not related to signature help when you're done?
* I don't mind if we remove syntax highlighting signature help, it was likely inaccurate with so little context.

braver:
> Maybe it's not a big deal?

Yes it is, they're just sloppy with their UX. This is a different pattern from the auto-complete list and doesn't signal clearly it will overtake up/down buttons: it's not a list.

rwols:
Yeah I went a bit overboard with the changes in a few places, I wanted to check if this was feasible. There are a few issues at hand: 1) browse function signatures 2) accentuate active parameter 3) make sure whitespace doesn't influence the signatureHelp trigger characters, I'll change the PR to only address (1).

tomv564:
Combining those three is fine with me too, I was thinking of the whitespace fixes in all the `debug` statements :)

rwols:
Ah okay, cool. I will restore those debug statements. I also want to try getting back the mdpopups syntax highlighting and see if that can work.

@braver what if we present the signatures as an actual *list*? Would that make people assume the up/down arrows are for selecting the active signature? I actually like the current approach with the `1/10` thingy. Maybe presenting the whole list is too much info, because documentation for the active signature also needs to go somewhere and documentation for the active parameter also needs to go somewhere.

braver:
@rwols I think it also depends on how the popup is triggered. 

In VS Code (last time I gave it a chance) it would show up when just navigating the code, ie. you'd use the arrow keys to move down a couple of lines and you'd get caught in this popup you didn't ask for. You'd then have to ESC out of it before you could continue navigating. So, it would disrupt what you were doing and demand attention. (VS Code does more of that and it's an attitude more fitting of an IDE IMHO).

So, if the popup is only triggered by mouse-hover, or by character insertion, or by a key-binding, then overtaking the arrow keys might not be so much of a problem. Still, without it looking like a list, it might be unexpected.

I think your approach to presenting the overload here is pretty neat, I wouldn't change that. Perhaps just land this PR first, more people will be able to have a look at it we'll figure out what works best. And if we don't figure it out, just make it configurable.

tomv564:
> In VS Code (last time I gave it a chance) it would show up when just navigating the code

I can happily report this is not the case now - only '(' triggers it in a typescript file. 

> So, if the popup is only triggered by mouse-hover, or by character insertion, or by a key-binding then overtaking the arrow keys might not be so much of a problem.

Mouse hover will not trigger this pop up (that's the Hover popup which cannot have multiple-signature support).

rwols:
pyls puts the `activeParameter` in the wrong dict? When I type `debug(` in main.py, I get this:
```
LSP: request 6: textDocument/signatureHelp 
LSP: got json:  {'result': {'activeSignature': 0, 'signatures': [{'documentation': 'Print args to the console if the "debug" setting is True.', 'parameters': [{'documentation': '', 'label': 'args'}], 'label': 'debug(*args)', 'activeParameter': 0}]}, 'id': 6, 'jsonrpc': '2.0'}
```
You can see the `activeParameter` being part of one of the signatures (only one signature in this case, actually there's always only one signature in the case of Python). But it should be sitting in the outer dict next to `actriveSignature`.

rwols:
I've noticed the logic can be flaky: sometimes the popup is visible but the `self._is_visible` bool is set to `False`, so pressing <kbd>⌘</kbd><kbd>↓</kbd> throws me to the bottom of the view. That can be annoying. I'm thinking of going back to just <kbd>↓</kbd> for the keybinding, at least then the cursor goes down a single line if the state isn't right.

tomv564:
Great find with the activeParameter being in the wrong place!
Sorry I missed your update 2 days ago, I am eager to try this out.

Code-wise I feel like a slightly terser implementation might be possible without hurting legibility. The logic gets a bit lost between the repeated `content.write`s and the `_maybe_write_documentation` calls. 

An example of what I mean: 
```python
if parameter_documentation:
  content.write(parameter_template.format(...
```

That said, writing python is not my day job and I'd love to hear more about your approach!

main.py
```diff
@@ -1974,57 +1978,192 @@ def initialize_triggers(self):
...
+        self.signatures = response.get("signatures", None)
+        x, y = self.view.viewport_extent()
+        debug("max width:", x, "max height:", y)
+        if self.signatures:
+            self.active_signature = response.get("activeSignature", -1)  # Optional.
+            self.active_parameter = response.get("activeParameter", -1)  # Optional.
+            self.view.show_popup(self._build_popup_content(), 0, point, int(x), int(y),
+                                 self._on_navigate, self._on_hide)
+        else:
+            self.view.show_popup('<html><body id="lsp"><div class="error">There are no signatures!</div></body></html>',
```
rwols: This can be "too clever" sometimes, for example when a trigger character is not actually inside of a func call this'll pop up. Maybe just remove it?

main.py
```diff
@@ -1990,57 +1994,192 @@ def initialize_triggers(self):
...
-                # TODO: this hides too soon.
-                if self.view.is_popup_visible():
-                    self.view.hide_popup()
+                index_end = index + len(parameter_label)
+                content.write(signature_label[:index])
+                content.write('<span class="active_parameter">')
+                content.write(signature_label[index:index_end])
+                content.write("</span>")  # active_parameter
+                content.write(signature_label[index_end:])
+            content.write("</span>")  # active_signature
```
tomv564: Do I understand correctly that his code does the following?
```python
signature_label.replace(parameter_label, '<span class="active_parameter">{}</span>'.format(parameter_label))
```
I would really prefer the above style unless there is a good reason against it (yes, it's not handling all the cases, but that can be fixed).
That said, I wonder if it wouldn't be safer to build the signature manually using `parameters`? This looks like it might run into problems if someone names their arguments `(instance, i)`

rwols: Ah yes, your line is much more elegant than my clumsy StringIO approach. I hope to have addressed all your issues tomorrow evening!

main.py
```diff
@@ -1990,57 +1994,192 @@ def initialize_triggers(self):
...
+                parameter = None
+                parameter_label = None
+                parameter_documentation = None
+        else:
+            signature = None
+            signature_label = None
+            signature_documentation = None
+            parameter = None
+            parameter_label = None
+            parameter_documentation = None
```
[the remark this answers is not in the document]
rwols: I couldn't find any way to improve this at first sight, do you have any suggestions?

rwols:
> Sounds good to me, we didn't warn about this before?

I removed it anyway, because it would pop up in strings too when e.g. typing a comma in Python. Maybe with some elaborate scope checking we could manage to make it reliable to only trigger when really inside of a function-call.

Btw, pyls 0.6.1 has the activeParameter in the correct place now, so the active parameter is getting highlighted 👍 .

tomv564:
Thanks for the code improvements!

New signature help in Python:
screen shot 2017-09-26 at 21 46 53

Old signature help in Python:
screen shot 2017-09-26 at 22 17 14

New signature help in Typescript:
screen shot 2017-09-26 at 21 41 47

Old signature help in Typescript:
screen shot 2017-09-26 at 21 59 51

Comparing these it's apparent that we lost some legibility:
* mdpopup's code block styling gave good contrast and spacing to the signature
* We lost newlines between documentation blocks
* We lost much-needed markdown formatting in the typescript signature.

But we did gain sans-serif fonts in plain text :+1:

Do you think you can restore some of the formatting from the old code?

As not every developer is going to check design changes for every language server, it sounds like we should build in some sample hovers / signature helps / diagnostics to design against. 
- [ ] I'll create an issue for this.

Two requests that we can take care of after merging are:
* A little clearer separation between the `1/2` and the rest of the signature when there are overloads
* Evaluate the experience keeping the signature popup open (unless the user presses Esc).

rwols:
Thanks for showing me a sample from Typescript, I tried setting up that language server but couldn't get it to work (I'm not a web developer).

I need to investigate how mdpopups translates the markdown to HTML, maybe there is already a function for that in mdpopups? I used mdpopups.md2html but it's obvious I'm missing the CSS now.

The whitespace issue should be easy to patch in.

Please don't merge yet.

tomv564:
[Mdpopups has pretty good documentation](http://facelessuser.github.io/sublime-markdown-popups/), I can't help you with personal experience as all the popups so far are 100% markdown.

rwols:
Thanks for the links, I'm gonna work on this on the weekend, been a bit busy with work-related stuff.

rwols:
I think this last commit restores most of the styling. There is some padding between the code block at the top and the edges of the popup, not sure how to get that back.

rwols:
Merged with latest upstream. `os.walk` looks like this now:

schermafbeelding 2017-09-30 om 21 45 35

Can be a bit overwhelming perhaps...

Anyway, the main feature still works (function overloads) and I think most of the markdown parsing is now there. I tried experimenting with the `preserve_whitespace` function but I'm not certain it's still needed when we call `mdpopups.md2html`.

rwols:
Actually just found a bug. I think a picture tells it most clearly:
schermafbeelding 2017-09-30 om 22 24 11

tomv564:
About the highlighting bug, I warned about this in the first review (see the outdated comments on main.py):

> That said, I wonder if it wouldn't be safer to build the signature manually using `parameters`? This looks like it might run into problems if someone names their arguments `(instance, i)`

Your fix does the job for `walk(top, topDown)`, but will the second argument ever get highlighted if the order is `walk(topDown, top)` ?

rwols:
You are correct, `walk(topDown, top)` will still highlight the `top` in `topDown` if the second parameter is to be highlighted.

But the problem with building the parameter list manually is that different languages might use different punctuation for separating their arguments. Not sure how we could fix this. I think *most* languages use a comma for a separator, but e.g. Objective-C uses whitespace.

tomv564:
OCaml and Haskell also uses spaces, but I am not sure how they trigger signature help as they don't use parentheses to wrap arguments either :)

Perhaps detecting presence of a `,` in the _parameter portion_ of the signature label can help?

Speaking of parameter portion, am I correct that a function definition `topWalk(top)` currently highlights part of the function name?

rwols:
schermafbeelding 2017-10-01 om 11 40 26

You are indeed correct 😋

rwols:
This last commit fixes a bunch of edge cases:
schermafbeelding 2017-10-01 om 12 14 59
schermafbeelding 2017-10-01 om 12 15 06
schermafbeelding 2017-10-01 om 12 15 59

rwols:
schermafbeelding 2017-10-01 om 13 50 09

This one remains as problematic.

main.py
```diff
@@ -2217,55 +2250,143 @@ def initialize_triggers(self):
...
+                                 self._on_navigate, self._on_hide)
+            self._visible = True
+
+    def _on_navigate(self, url):
+        sublime.run_command("open_url", {"url": url})
+
+    def _on_hide(self):
+        self._visible = False
+
+    def on_query_context(self, key, _, operand, __):
```
randy3k: you really like the use of `on_query_context` over the use of `ViewCommand` or `WindowCommand`.

rwols: The EventListener is there anyway, might as well override that method ;)
Otherwise, I'd have to look up the class.

rwols:
Just to give a heads up I want to try using `mdpopups.show_popup` again, but with some embedded HTML for the active signature part.

rwols:
I'm closing this because https://github.com/tomv564/LSP/pull/163 and https://github.com/tomv564/LSP/pull/162 are smaller, more digestible chunks of code to merge. The active parameter story should be handled separately.

Issue #105, opened 2017-09-16
Whitespace in popups could be better


braver:
Perhaps this is a bit too minor/visual for this stage of the package, but I'm a designer and it's bugging me 😁 

<img width="877" alt="screen shot 2017-09-16 at 13 50 24" src="https://user-images.githubusercontent.com/2543659/30512072-ac551cea-9ae6-11e7-8459-2b20f4fdaecc.png">

Text sits too tight too the edge of the popup. At the same time there is an inexplicably large amount of space between the code fragment at the top and the documentation below. It looks worse if there isn't much documentation (e.g. it just says "function").

It would look better if the text aligns and margins are consistent all around. 

<img width="332" alt="better" src="https://user-images.githubusercontent.com/2543659/30512094-4f36bedc-9ae7-11e7-8184-30635aa2e012.png">

I also noticed the phantoms are missing the little pointer nubbin that the build system phantoms have. This seems to be a function of using the mdpopups library. Are these things we can/should tune here, or should I try to land some enhancements in mdpopups?

tomv564:
Agree absolutely, would love for a PR to fix the left margin and paragraph / code block spacing.
The HTML output may need to change as well?
Perhaps can also apply the sans-serif font ( issue #82 )?

Fixing the pointer on the phantoms is probably a question of taking the right bits of the large style sheet here: https://github.com/twolfson/sublime-files/blob/master/Packages/Default/exec.py#L395

braver:
If @rmccue isn't already upto something with this, I could look into this some time next week.

rwols:
There's also the issue that we can't browse through other overloads like in VSCode.

braver:
@rwols could you point me to a demonstration of that?

rwols:
I'm working on that right now, here's a prototype screenshot:

<img width="1089" alt="schermafbeelding 2017-09-17 om 10 31 28" src="https://user-images.githubusercontent.com/2431823/30519313-87253132-9b93-11e7-8811-319ab34fd8f6.png">

I just dump all overloads to the popup box, but ideally, the "active" overload should be displayed, and the up and down arrow keys allow you to browse through the other overloads. The active parameter should be bold. This is how it's done in intelliSense in Visual Studio.

Anyway, this is somewhat of a separate issue but I thought I'd let you know of what I'm up to so you can keep it in mind for when you might change the layout.

braver:
[...]

PR #112, opened 2017-09-18
Enhancement: Keep whitespace displayed in popups

deathaxe:
mdpopups' markdown engine squashes down all multiple spaces to one.

This PR prepares the content to preserve them all and make popup content look a bit better.

### DIFF
--- a/main.py
+++ b/main.py
@@ -1712,13 +1712,21 @@ class HoverHandler(sublime_plugin.ViewEventListener):
 
         mdpopups.show_popup(
             self.view,
-            "\n".join(formatted),
+            format_popup_content("\n".join(formatted)),
             css=".mdpopups .lsp_hover { margin: 4px; } .mdpopups p { margin: 0.1rem; }",
             md=True,
             flags=sublime.HIDE_ON_MOUSE_MOVE_AWAY,
             location=point,
             wrapper_class="lsp_hover",
             max_width=800)
+
+
+def format_popup_content(contents: str) -> str:
+    """Preserve empty lines and whitespace for markdown conversion."""
+    contents = contents.replace('\t', '&nbsp' * 4)
+    contents = contents.replace('  ', '\a\a')
+    contents = contents.replace('\n\n', '\n&nbsp;\n')
+    return contents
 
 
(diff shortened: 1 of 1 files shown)


tomv564:
Could you share an example (e.g. before/after screenshot?)

Especially as there is also issue #105 and PR #111 ongoing, I'd be unfortunate if 3 people were working on improving popups in different ways.

deathaxe:
It is related to #105 yes, even though I havn't been not following that issue. Just realized it myself.

before

screenshot_before

now

screenshot

deathaxe:
If you ask me, signature/docstring popups require their own rendering module/package to be able to handle different languages and formats and create a common user experience. This is just a very short and quick hotfix.

rwols:
Yeah, I sort of agree with @deathaxe in that signature popups may vary too much between servers/languages.

It could be an idea to make LSP itself a PC package that can be added as a dependency, and then provide language-specific overrides for various servers/languages. This is how it's handled in VSCode and Atom, for example.

deathaxe:
This approach would split the server/protocol specific communication infrastructure which is some kind of backend away from presenting the results and handle user interactions which would be the frontend.

No matter how it is solved, but I think some infrastructure which splits apart the the protocol and communication stuff away from language-specific handling is required. But I like the idea to have one package, which is at least capable to provide a minimum set of functions. If you ask someone to create a frontend for each language, you'll probably not find one. 

Have a look onto the new `Python Debugger` package. It should better be named as `Sublime Debugger` as it does the same which LSP does. It provides an frontend/backend infrastructure to attach nearly each debugger to ST. I wonder if one day someone will start using it to create new debugger backends. I am in doupt with that (unfortunately).

main.py
```diff
@@ -1721,6 +1721,14 @@ def show_hover(self, point, contents):
             max_width=800)
 
 
+def format_popup_content(contents: str) -> str:
```
tomv564: To avoid confusion about what "formatting" really means, could you rename this to something like `preserve_whitespace` or `encode_whitespace` (or something else to your liking?)

tomv564:
:metal:

Issue #117, opened 2017-09-20
User settings are all-or-nothing


golddranks:
I used to have user settings like this: https://gist.github.com/urschrei/5f526b1a00378bfc049b23323b0002fb

But as one can see, the default settings already contain

```
      "rls":
      {
        "command": ["rustup", "run", "nightly", "rls"],
        "scopes": ["source.rust"],
        "syntaxes": ["Packages/Rust/Rust.sublime-syntax"],
        "languageId": "rust"
      },
```

So I tried to remove these fields from my user settings `rls` object because they seemed redundant. This made `rls` not start.

I suspect that it simply *replaces* the default `rls` settings object with the user `rls` setting object, instead of *merging* it?

tomv564:
Correct, in fact the entire contents of "clients" is replaced by your user settings. We do have configuration merging from User to Project-specific settings.

I'll link / comment this issue with an upcoming proposal on default/user settings could be managed!

tomv564:
This should be solved now that #119 is merged, but I would love some feedback.

The `clients` block in the User LSP.sublime-settings now inherits defaults from `default_clients`, so you are free to only specify overrides for  eg. `command` there.

deathaxe:
Seems to work smoothly. I modified pyls` command line in the user settings to create a logfile and added some env vars. No issues detected so far.

tomv564:
Thanks for the feedback!

PR #119, opened 2017-09-20
Language server configuration management

### LINKED ISSUE #76: [Feature request] Merge configurations?
Correct me if I'm wrong, but right now if you set a custom `"clients"` in preference, it overrides the default config rather than merging them together. Would it make sense if that configs are merged?

tomv564:
* Removes default-enabled configurations
* Adds commands to enable/disable globally or in project (solve issue #69)
* Adds setup command to help the user configure the language server (for issue #4)

### Onboarding flow

Notification when a supported syntax is detected:

screen shot 2017-09-20 at 21 46 06

Documentation shown when running "Setup Language Server": 

screen shot 2017-09-20 at 21 46 48

Clicking one of the Enable options in the bottom will copy the default config to `clients` in your user settings and set the `enabled` flags at the user/project level if needed. LSP starts up immediately.

### Copying global config vs. merging from default

The "clients" setting in the User LSP.sublime-settings fully replaced the defaults, causing #76 and #117.
Moving the default clients to their own setting key fixes #76 (as there are no default-enabled clients, you don't lose any by defining "clients" in user settings)
As we copy the entire config of a client into user settings, #117 will still not be solved: There will be no way to "use the default config, and only override the `command` field"
Is this feature worth dedicating even more lines of config-wrangling code to?

### TO DO

- [x] Determine if merging config up from `default_clients` is worthwhile (#117)
- [x] Better naming and handling of unparsed `settings` and parsed `configs`
- [x] Remove duplication around `project_data` handling
- [x] Evaluate impact to users with existing settings
- [x] Find a better way to show documentation - the popup looks terrible, just opening the MD is likely better.
-  ~Update documentation structure so we can open/link to individual language setups~


deathaxe:
> As we copy the entire config of a client into user settings, #117 will still not be solved: There will be no way to "use the default config, and only override the command field"

There is a way to do so, I think. You could try loading the settings by the following two lines:

```python
clients = settings.get("default_clients")
clients.update(settings.get("user_clients"))
```

`clients.update()` merges the two configurations basing on the default_clients.

tomv564:
@deathaxe I looked into this today, this should work if we `update` each client config. Thanks for the tip!

tomv564:
The user benefit of merging up from defaults instead of copying defaults is that if LSP improves the defaults (eg. extra syntaxes or scopes) the user will immediately get those updates. 
The downside is that every such update is potentially a breaking update (esp. when it comes to `command`). But this will probably not happen too often.

**supported_client_configs** use the settings in `default_clients`

**global_client_configs** inherit the settings from `default_clients`

In order to enable eg. `pyls` *globally*,  the User LSP.sublime-settings should have:
```json
"clients": {
  "pyls": {
    "enabled": true
  }
}
```

**project_client_configs** inherit the settings from `clients` and `default_clients`

In order to enable `pyls`, a sublime-project should have:

```json
"settings": {
  "LSP": {
    "clients": {
      "pyls": {
        "enabled": true
      }
    }
  }
}
```

deathaxe:
> The downside is that every such update is potentially a breaking update (esp. when it comes to command). But this will probably not happen too often.

If a user overrides a setting, he must ensure his changes work. You could probably do some format/type checks against the user configuration before merging. If a key is present, it must match the desired type. On the other hand, if things are merged and something goes wrong, because of invalid settings, you can be quite sure its the user settings issue and output a descriptive error.

I find the idea of merging things together better than ask the user to copy the whole structure. Can be compared with patching a color scheme vs. a theme. Later one is much easier to adjust as you just need to create a file with few lines to patch the original one.

PR #120, opened 2017-09-21
Strip leading and tailing whitespace from function documentations.

deathaxe:
Some python docstrings contain multiple newlines at the end, which results in an empty line being displayed in the tooltip for such docstrings. To avoid this issue, `preserve_whitespace` now strips whitespace from the beginning and the end of the content to display in a popup.

For consistency the replacement of multiple spaces was modified, too. Without functional effect.


tomv564:
Thanks for the fix!

PR #122, opened 2017-09-22
show other completions even when CompletionState is idle

randy3k:
Actually, I don't understand why we should inhibit word completions and `.sublime-completion` completions when CompletionState is idle. Firstly, some servers (e.g. julia server) are really slow to respond, if we inhibit word completions and the server doesn't respond fast enough or encounters an error while handling a completion request, it will essentially stop the entire auto-completion functionality. Secondly, if the reason was to prevent flickers, then I argue that `.sublime-snippet` should be also inhibited. Otherwise, there would still be flickers. (AFAIK, there is no way to stop snippet completions).

### DIFF
--- a/main.py
+++ b/main.py
@@ -1887,7 +1887,7 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
 
             return (
                 self.completions,
-                0 if self.state == CompletionState.IDLE and not only_show_lsp_completions
+                0 if not only_show_lsp_completions
                 else sublime.INHIBIT_WORD_COMPLETIONS | sublime.INHIBIT_EXPLICIT_COMPLETIONS
             )
 


randy3k:
Ai....it is not working somehow.

tomv564:
Glad I'm not the only person wondering where the snippet completion could be disabled :)

Your change looks good, but I wonder if we can have LSP even less involved in the default completions?

Your use case (`not only_show_lsp_completions`) may not need to return anything until LSP has responded?

randy3k:
If one perfer `only_show_lsp_completions`, then we should inhibit word completions, `*.sublime-completion` and `*.sublime-snippet` (though there is no way to turn off the snippet completion so far). I don't know if the use will still like to see the default completions before LSP responds.

PR #123, opened 2017-09-23
Doc changes about autocomplete triggers.

golddranks:
Fixed autocomplete_triggers -> auto_complete_triggers and added the explanation about 'User or Syntax-specific settings' to each site, because the users are likely to skip directly to the language-specific explanation. (`auto_complete_triggers` won't work in the LSP settings, but if you skip the explanation, you are likely to think that's where you're supposed to copy it.)


tomv564:
Thanks so much for helping fellow users get started with LSP!

PR #125, opened 2017-09-23
Fixed a bug where the last message from the server doesn't get read when the server terminates.

golddranks:
You were right that it doesn't print the last error message. Basically calling `subprocess.poll()` to check if the process has terminated, and if it isn't, read from stdout/stderr is a race condition. It's better to terminate the loop after `EOF` has reached directly from the pipes.

Additionally, made it so that it prints the error message even when debug logging isn't on, which is consistent with other fatal error conditions.

This fixes https://github.com/tomv564/LSP/issues/116

### DIFF
--- a/main.py
+++ b/main.py
@@ -469,13 +469,17 @@ class Client(object):
         """
         ContentLengthHeader = b"Content-Length: "
 
-        while self.process.poll() is None:
+        while True:
             try:
 
                 in_headers = True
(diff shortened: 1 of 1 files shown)


golddranks:
I checked the Python docs for it, it says that readline returns `''` on EOF (and `'\n'` on empty line) :)

PR #126, opened 2017-09-23
Added a fallback in the case where there is no project folder.

golddranks:
If there is no project folder, it considers the directory containing the file as a fallback case. Whether this is a grokkable by the language servers is their business, but at least it allows them to be able to handle the case, if they so wish.

I tested that by default, RLS can do code completion even if the project folder isn't set right.

### DIFF
--- a/main.py
+++ b/main.py
@@ -601,8 +601,15 @@ def get_project_path(window: sublime.Window) -> 'Optional[str]':
         folder_paths = window.folders()
         return folder_paths[0]
     else:
-        debug("Couldn't determine project directory")
-        return None
+        filename = window.active_view().file_name()
+        if filename:
+            project_path = os.path.dirname(filename)
+            debug("Couldn't determine project directory since no folders are open!",
+              "Using", project_path, "as a fallback.")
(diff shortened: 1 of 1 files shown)


golddranks:
Fixes https://github.com/tomv564/LSP/issues/124

main.py
```diff
@@ -587,16 +587,18 @@ def printf(*args, prefix=PLUGIN_NAME):
...
     """
     Returns the common root of all open folders in the window
     """
     if len(window.folders()):
         folder_paths = window.folders()
         return folder_paths[0]
     else:
-        debug("Couldn't determine project directory")
-        return None
+        project_path = os.path.dirname(window.active_view().file_name())
```
tomv564: We might want to guard against View.file_name() returning None (e.g. what if you create a new file and switch syntax to a supported syntax)

golddranks: Ah, indeed. I'll send another commit fixing this.

golddranks:
Rebased and added the check.

golddranks:
All right, tests passed :)

PR #130, opened 2017-09-27
Add inline type annotations and function argument hinting for Rust

coder543:
Last night, I just got really carried away with wanting Sublime Text to offer an experience to rival that of IntelliJ Rust, so I implemented a couple of big features last night and today.

I tried to be careful to make sure this PR won't affect non-Rust RLS support at all. Let me know what you think!

function parameter hinting:
function parameter hinting

variable type hinting:
screenshot_2017-09-27_15-20-49

for loop iterator variable type hinting:
screenshot_2017-09-27_15-20-14

In the case of variable type hinting, it will refrain from inserting a phantom if it detects that the type has been manually specified, so it only annotates types for variables where the type is inferred.

I'm sure parts of this could probably be reused for other languages that rely heavily on type inference too! Go would be a fine candidate, but I understand the Go RLS support is maybe a little problematic, according to the docs.

### DIFF
--- a/main.py
+++ b/main.py
@@ -1904,7 +1904,7 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
 
             return (
                 self.completions,
-                0 if self.state == CompletionState.IDLE and not only_show_lsp_completions
+                0 if not only_show_lsp_completions
                 else sublime.INHIBIT_WORD_COMPLETIONS | sublime.INHIBIT_EXPLICIT_COMPLETIONS
             )
 
--- a/main.py
+++ b/main.py
@@ -2061,7 +2061,7 @@ class TypeAnnotator(object):
             # we only want the first line of the doc string
             formatted_str = formatted_str.split("\n")[0]
 
-        docs = ""
+        docs = []
         if category == "fn":
             text = formatted_str.split("\n")
             docs = text[:-1]
--- a/main.py
+++ b/main.py
@@ -1989,6 +1989,7 @@ class TypeAnnotator(object):
             phantoms_to_generate -= 1
             return
 
+        formatted_str = re.sub("<[a-zA-Z0-9_]*>", "", formatted_str)
         formatted_str = re.sub("[^& <]*::", "", formatted_str)
 
         region = self.view.word(point)
--- a/main.py
+++ b/main.py
@@ -1723,13 +1723,21 @@ class HoverHandler(sublime_plugin.ViewEventListener):
 
         mdpopups.show_popup(
             self.view,
-            "\n".join(formatted),
+            preserve_whitespace("\n".join(formatted)),
             css=".mdpopups .lsp_hover { margin: 4px; } .mdpopups p { margin: 0.1rem; }",
             md=True,
             flags=sublime.HIDE_ON_MOUSE_MOVE_AWAY,
             location=point,
             wrapper_class="lsp_hover",
             max_width=800)
+
+
+def preserve_whitespace(contents: str) -> str:
+    """Preserve empty lines and whitespace for markdown conversion."""
+    contents = contents.replace('\t', '&nbsp' * 4)
+    contents = contents.replace('  ', '\a\a')
+    contents = contents.replace('\n\n', '\n&nbsp;\n')
+    return contents
 
 
 class CompletionState(object):
@@ -2005,7 +2013,7 @@ class SignatureHelpListener(sublime_plugin.ViewEventListener):
 
                 mdpopups.show_popup(
                     self.view,
-                    "\n".join(formatted),
+                    preserve_whitespace("\n".join(formatted)),
                     css=".mdpopups .lsp_signature { margin: 4px; } .mdpopups p { margin: 0.1rem; }",
                     md=True,
                     flags=sublime.HIDE_ON_MOUSE_MOVE_AWAY,
--- a/main.py
+++ b/main.py
@@ -983,6 +983,24 @@ def annotate_visible_types(view: sublime.View):
         var_text = var_text[:-2]
         var_start = var.begin() + var_text.rfind(" ") + 1
         annotator.request_symbol_annotate(var_start)
+    tuple_vars = view.find_all('\\blet\\b *\(([a-zA-Z0-9_, ]*)\)', 0)
+    for var in tuple_vars:
+        if var is None or var.begin() == -1:
+            continue
+        var_text = view.substr(var)
+        var_start = var.begin() + var_text.find('(') + 1
+        var_text = re.sub("let *\(", "", var_text)
+        var_text = var_text[:-1]
+        tp_vars = var_text.split(",")
+        print("variable:", var_text, view.rowcol(var_start))
+        first = True
+        for var in tp_vars:
+            annotator.request_symbol_annotate(var_start)
+            print("tuple var:", view.rowcol(var_start))
+            if first:
+                first = False
+                var_start += 1
+            var_start += 1 + len(var)
     iter_vars = view.find_all('\\bfor\\b *([a-zA-Z_][a-zA-Z0-9_]*)', 0)
     for var in iter_vars:
         if var is None or var.begin() == -1:
--- a/main.py
+++ b/main.py
@@ -1880,7 +1880,7 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
 
     def is_after_trigger_character(self, location):
         if location > 0:
-            prev_char = self.view.substr()
+            prev_char = self.view.substr(location - 1)
             return prev_char in self.trigger_chars
 
     def on_query_completions(self, prefix, locations):
--- a/main.py
+++ b/main.py
@@ -1926,10 +1926,10 @@ class TypeAnnotator(object):
         formatted = fn_name + re.sub(" *fn *", "", formatted)
         offset = 0
         if "&mut self" in formatted:
-            formatted = "&mut self." + formatted.replace("&mut self, ", "", 1).replace("(, ", "", 1)
+            formatted = "&mut self." + formatted.replace("&mut self", "", 1).replace("(, ", "", 1)
             offset = len("&mut self.")
         elif "&self" in formatted:
-            formatted = "&self." + formatted.replace("&self, ", "", 1).replace("(, ", "", 1)
+            formatted = "&self." + formatted.replace("&self", "", 1).replace("(, ", "", 1)
             offset = len("&self.")
         elif "self" in formatted:
             formatted = "self." + formatted.replace("self", "", 1).replace("(, ", "", 1)
--- a/main.py
+++ b/main.py
@@ -1836,6 +1836,7 @@ def annotate_types(view: sublime.View, current_function: bool):
     annotator.annotate_var_decl(view)
     annotator.annotate_tuple_decl(view)
     annotator.annotate_for_loops(view)
+    annotator.annotate_for_tuple_loops(view)
     if current_function:
         annotator.annotate_function(view)
     sublime.set_timeout_async(lambda: show_type_phantoms(view), 100)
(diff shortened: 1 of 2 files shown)


main.py
```diff
@@ -60,7 +65,7 @@ class DiagnosticSeverity(object):
 diagnostic_severity_scopes = {
     DiagnosticSeverity.Error: 'markup.deleted.lsp sublimelinter.mark.error markup.error.lsp',
     DiagnosticSeverity.Warning: 'markup.changed.lsp sublimelinter.mark.warning markup.warning.lsp',
-    DiagnosticSeverity.Information: 'markup.inserted.lsp sublimelinter.gutter-mark markup.info.lsp',
+    DiagnosticSeverity.Information: 'markup.inserted.lsp sublimel:inter.gutter-mark markup.info.lsp',
```
rwols: Unintentional typo?

main.py
```diff
@@ -574,6 +580,14 @@ def notification_handler(self, response):
         elif method == "window/logMessage" and log_server:
             server_log(self.process.args[0],
                        response.get("params").get("message"))
+        elif method == "rustDocument/diagnosticsBegin":
```
rwols: What are all these `rustDocument/*` methods? Where are they documented?

main.py
```diff
@@ -1399,6 +1444,7 @@ def update_file_diagnostics(window: sublime.Window, file_path: str, source: str,
 
 
 def update_diagnostics_phantoms(view: sublime.View, diagnostics: 'List[Diagnostic]'):
+    return
```
rwols: Why do we return here from this function immediately?

main.py
coder543: these are LSP extensions specific to the Rust Language Server implementation. They're [documented here.](https://github.com/rust-lang-nursery/rls/blob/master/contributing.md#extensions-to-the-language-server-protocol)

main.py
coder543: Ah... I will delete that. I was having issues with the Phantoms, so I forced LSP's existing phantoms to no longer show up, just in case that was causing the issue, but I forgot to remove that return statement. I don't think this was the cause of the issue, so I'll re-test stuff with that return statement removed.

main.py
tomv564: They are custom messages specific to RLS which any LSP consumer is free to ignore or handle.

These Rust ones are probably the most commonly used ones - the initial 'analyse' build can take a long time, the VS extension displays a spinner with some text to assure the user that things are happening :)

This specific handling should be regarded as a future feature for the LSP API - how can we expose hooks for an "LSP-contrib-rust" package to add extra handlers and features?

rwols: Cool! I think this instigates an interesting discussion on how LSP should handle server-specific code.

coder543:
I'm cleaning up the complaints from `mypy` and `flake8` now. I'm not sure how to fix the complaint from here:

    syntax = view.settings().get('syntax')
    if "Rust" not in syntax:

it complains: `error: Unsupported right operand type for in ("Optional[Any]")`

but if I try to annotate `syntax`, it complains the right-hand side of the assignment is invalid.

main.py
coder543: yep, sorry.

main.py
coder543: So, I tested, and this `update_diagnostics_phantoms` function is actually causing issues. My phantoms will show up for a short time, but as soon as this function gets triggered, it clears all Phantoms from the document, even though (or especially since) I have `"show_diagnostics_phantoms": false` in my config file.

Adding some more config options to control the new features I introduced for the Rust side is on my todo list.

coder543: But, the PhantomSet's have different ID strings, so I'm not sure why they interfere with each other.

rwols:
> but if I try to annotate syntax, it complains the right-hand side of the assignment is invalid.

You can use `# type: ignore` for these kinds of things ;)

main.py
```diff
@@ -1401,6 +1450,12 @@ def update_file_diagnostics(window: sublime.Window, file_path: str, source: str,
 def update_diagnostics_phantoms(view: sublime.View, diagnostics: 'List[Diagnostic]'):
     global phantom_sets_by_buffer
 
+    # disable the normal LSP phantoms when using a Rust document,
+    # since these interfere with the enhanced Rust phantoms
+    syntax = view.settings().get('syntax')
+    if "Rust" in syntax: # type: ignore
+        return
```
coder543: This may not be the greatest solution, long-term, but maybe it'll be enough for now.

If anyone could help me understand why the phantoms are interfering with each other, maybe there's an obvious, good long-term solution to be implemented that I'm just not aware of.

tomv564:
On first glance this looks like a really impressive effort.
I also like that LSP is inspiring people to implement IntelliJ-like features :)

But to move forward, here are some ideas:

As I alluded to in the answer to @rwols's question, the Rust protocol extensions should not be part of this package. Or at least not in main.py with the default LSP handlers. I would love it if you could work with us to implement this as a plugin, as a separate PR.

About the type annotations:

As a non-rust programmer, it took me a while to realise that the annotations were not just syntax highlighting. Would be super helpful to see this in action in a video / GIF. 

I'm really impressed by the inline phantoms! This is the kind of mechanism we could provide for all languages once we have a good example of how to use it.

Could you propose how this functionality could live in a rust extension package?
How do you feel about the completeness and future-proofness of the string-searching-and-parsing of the rust code?
How dependent is this functionality on the performance of the language server?

coder543:
I'll try to answer more in a little while when I get a chance, but to answer a few quick things:

> How do you feel about the completeness and future-proofness of the string-searching-and-parsing of the rust code?

It should be pretty robust with regards to completeness and future changes to the language. The basic syntax of Rust is not expected to change much, but it is always a consideration that they could completely alter the syntax one day in the future. I don't think there's anything that can be done to protect against that, without the Language Server Protocol itself being extended to have more features.

Rust uses `let name[: type]` for all local variable declarations. If the type is not specified, it is inferred. Spaces can optionally be placed before the colon, but that is non-standard formatting, according to the work being done on official formatting guidelines, so I only supported 1 space before the colon when detecting whether the type was manually specified or not. This is the least robust part of the string-searching-and-replacing code, and I could shore that up.

It is not possible to specify the type of the for-loop iterator variable manually, so that one is not a concern, and that syntax is _always_ `for variable in iterator`. This should never be a problem, now that I'm checking for the `in` that follows the variable name, since this syntax should be unique, and a phantom will only be created if type information is returned by the LSP.

Detecting that we are typing a function call is done only by detecting a valid identifier followed immediately by an opening parenthesis, which is conservative. I think Rust does support a certain suite of unicode values as identifiers, but this will work for the common cases and avoid false-positives in others. Support can always be extended with a more comprehensive identifier validation function in the future. The way the phantoms are implemented right now, the function signature phantom will persist until the users saves the document or opens another function call, both of which will trigger a re-annotation cycle. I think this is a good balance to be struck, since it helps when initially writing the function call, but it doesn't bother you when you're just changing a parameter in an existing one, and the information disappears once it is no longer needed. I tested just having _all_ function calls annotated, and that was extreme information overload.

> How dependent is this functionality on the performance of the language server?

My implementation earlier today was very dependent on the performance of RLS, but I switched to a model where type annotation waits for effectively as long as needed for RLS to provide complete information. I'm not a fan of spinning on a global, waiting on it to change state, but it seems to model what's actually happening very precisely, so that's what I did.

coder543:
This is a different, older Rust code base of mine, and I deleted the manual type annotation on `y_points` just to make the contrast between manually-specified and LSP-inferred types easier to see. `pvs` also has an inferred type phantom being injected into the code. The only reason I specified the types on `colors`, `x_points`, and `y_points` is to make the code easier to understand -- Rust is perfectly capable of deducing the correct types here.

screenshot_2017-09-27_16-56-08

coder543:
I need to figure out how to make some GIFs of Sublime Text so it's easy to see this stuff in action.

coder543:
> I would love it if you could work with us to implement this as a plugin, as a separate PR.

This would be a plugin for LSP, not for Sublime Text, right? Would it be separated just for modularity, or would it actually be a separate installation inside of Package Control?

> Could you propose how this functionality could live in a rust extension package?

I'm not sure what the best practices would be. I've never written an editor extension before, and it's been years since I wrote a lot of Python, which was back in the Python 2 days.

Either way, I'm happy to do what I can to help out. I would really like this to be seamlessly integrated with LSP, that way I can share it with the rest of the Rust community and they can have the same awesome Rust experience that I now have access to!

> I'm really impressed by the inline phantoms! This is the kind of mechanism we could provide for all languages once we have a good example of how to use it.

Thanks! I agree that other languages definitely could benefit from inline phantoms. I really enjoy having the full function signature in front of me while writing a function call, but another kind of inline phantom that I would love to implement (but haven't) is function parameter hints, as seen in this image of IntelliJ that I found on google:

parameter hints

tomv564:
It would be a Sublime Text package that cooperates with the LSP package to deliver the functionality. I'm open to LSP maintaining some API stability for language-specific enhancements, but I would like to know what state / methods you need access to.

deathaxe:
> I'm really impressed by the inline phantoms! This is the kind of mechanism we could provide for all languages once we have a good example of how to use it.

I know from packages like ColorHighlighter that inline phantoms often cause the text to jump around while typing. I often deleted the wrong characters when editing colors because the phantom got deleted and re-added. Terrible working - I got rid of that plugin. Therefore I don't find this idea too impressive, honestly. 

> Could you propose how this functionality could live in a rust extension package?

In order to do something like that, LSP must provide the correct architecture for it. With all the bunch of stuff in one source file, I am not sure how to accomplish something like that. We first need to split the current package into logical parts and create some kind of interface a "language-extension package" could make use of. In the end LSP must provide a general solution to access the language-server-protocol and run servers and hand over the responses to an API to give extentions like the rust thing a chance to handle them, do some formatting on output and  pass the result back to LSP which renders the popup/adds the results to output panel and so on.

See the pluggy interface of the python-language server.

This needs a bunch of architecture work before something like an extension can be added. If you don't design that stuff thoroughly, the package will end up as a big un maintainable mess after 3 or 4 such extensions have been added. We need a clear structure first!

Just as an orientation: Google's pylint limits python modules to 500 lines of code by default. Anything beyond is complained about.

I don't want to insult anybody! I am just concerned about such an important and helpful piece of software project to crash or become unusable.

The approach of being able to add a common solution to achieve code intelligence features for most languages is just too important to fail, I think.

coder543:
> I know from packages like ColorHighlighter that inline phantoms often cause the text to jump around while typing. I often deleted the wrong characters when editing colors because the phantom got deleted and re-added. Terrible working - I got rid of that plugin. Therefore I don't find this idea too impressive, honestly.

This only updates the phantoms when a document is first opened, when it is saved, and when the user types the character `(` preceded by a valid identifier. Sublime Text does a great job of handling phantoms while they're static, so, knocking it before you test it seems... unwarranted.

The rest of your post I agree with.

coder543:
At this point, while I'm open to ideas of how to make these Rust extensions coexist with the `LSP` plugin, it definitely seems like it would be an enormous amount of architectural effort to make that happen, so I'm tempted to just fork `LSP` and publish an `RLS` package that only provides LSP functionality for Rust documents, and would otherwise not interfere with the LSP package.

I'm a firm believer in configurability, so I need to add _several_ config options before I consider taking that step, since I don't want to release this until the user is able to enable and disable the various annotations it offers.

tomv564:
@coder543: If you are eager to expose this functionality in the short term to Rust users, then a fork is your best bet. I will start splitting `main.py` soon, but it's not likely that a plugin-friendly architecture will emerge from that. 

> Therefore I don't find this idea too impressive, honestly.

@deathaxe: @coder543 took it well, but wouldn't it have sufficed to say "I would like the option to turn this off"?

deathaxe:
The comments about the inline type annotations sound as it was a none plus ultra super feature, which I just can't completely follow because of the mentioned experiences. What is bad about stating possible side effects of a solution to be able to judge it objectively? I don't want to say the function is useless (the idea is indeed quite cool! 👍), but using phantoms in such a way to dynamically add inline stuff has in fact its drawbacks, which just needs to be stated as well.

rwols:
From a high-level point of view these changes want to listen to *more* notifications / override the default behavior of the notification handler, and from there inject more functionality.

This should be possible if we turn most of main.py into a base class that handles the basic protocol messages, but allows for overridability in certain places.

Again, this is how Atom and VSCode do it, and I think we should move to such an implementation, too.

coder543:
Closing this for now. It would still be nice to integrate one day, but my time has been almost entirely reallocated to other projects for the past month. I still need to at least publish a rough version as a fork on Package Control so other Rust users have the option of using my additions until my changes get merged in some form or another, if I (or someone) ever get around to putting in the work to make that happen.

PR #133, opened 2017-09-29
Add support for per language-server environment variables

deathaxe:
With this PR a user can add a dictionary to each client-configuration.

    "environment": {
        "PATH": "$PATH;${project_path};/opt/libs"
        "PYTHONPATH": [
            "${sublime_libs}",   feature removed by request
            "${packages}",
            "${project_path}"
        ]
    }

This dictionary lists environment variables to add or set to the global
environment variables `os.environ` provided by Sublime Text.

The language server will see the merged set of environment variables.

Notes about the dictionary:

1. Each variable can be either a simple string or a list of strings.
2. If a variable is a list, it is merged to a string separated by ;
   This enables easier setup of larger sets of values such as paths.
3. Each variable replaces the global environment variable.
4. Shell variables such as $PATH are expanded.
5. Point 3. and 4. allow merging of global and local variables (see $PATH)
6. The following Sublime Text variables can be added as value and will
   be expanded on assignment as well:

```
    ${packages}              // Sublime Text Packages path
    ${platform}              // Operating system
    ${project}               // Full path to project file
    ${project_base_name}     // Project name without extension
    ${project_name}          // The name of the project
    ${project_path}          // Folder in which the open Project is located

// Removed by request
//    ${sublime_libs}          // sys.path as known by Sublime Text
//    ${sublime_path}          // install path of Sublime Text
```

7. The values of a global setting are replaced by user or project settings
   as this is the default behavior of dict.update().

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -6,7 +6,14 @@
       "command": ["pyls"],
       "scopes": ["source.python"],
       "syntaxes": ["Packages/Python/Python.sublime-syntax"],
-      "languageId": "python"
+      "languageId": "python",
+      "environment": {
+        "PYTHONPATH": [
+            "${sublime_libs}",
+            "${packages}",
+            "${project_path}"
+        ]
+      }
     },
     "jsts":
     {
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -7,12 +7,7 @@
       "scopes": ["source.python"],
       "syntaxes": ["Packages/Python/Python.sublime-syntax"],
       "languageId": "python",
-      "environment": {
-        "PYTHONPATH": [
-            "${sublime_libs}",
-            "${packages}",
-            "${project_path}"
-        ]
+      "env": {
       }
     },
     "jsts":
--- a/main.py
+++ b/main.py
@@ -817,8 +817,6 @@ def apply_window_settings(client_config: 'ClientConfig', view: 'sublime.View') -
     if client_config.name in window_config:
         overrides = window_config[client_config.name]
         debug('window has override for', client_config.name, overrides)
-        merged_init_options = dict(client_config.init_options)
-        merged_init_options.update(overrides.get("initializationOptions", dict()))
         return ClientConfig(
             client_config.name,
             overrides.get("command", client_config.binary_args),
@@ -826,8 +824,10 @@ def apply_window_settings(client_config: 'ClientConfig', view: 'sublime.View') -
             overrides.get("syntaxes", client_config.syntaxes),
             overrides.get("languageId", client_config.languageId),
             overrides.get("enabled", client_config.enabled),
-            merged_init_options,
-            overrides.get("settings", dict()))
+            overrides.get("initializationOptions", client_config.init_options),
+            overrides.get("settings", client_config.settings),
+            overrides.get("env", client_config.env)
+        )
     else:
         return client_config
 
--- a/main.py
+++ b/main.py
@@ -337,7 +337,7 @@ def read_client_config(name, client_config):
         client_config.get("enabled", True),
         client_config.get("initializationOptions", dict()),
         client_config.get("settings", dict()),
-        client_config.get("environment", dict())
+        client_config.get("env", dict())
     )
 
 
@@ -458,7 +458,7 @@ def update_settings(settings_obj: sublime.Settings):
 
 class ClientConfig(object):
     def __init__(self, name, binary_args, scopes, syntaxes, languageId,
-                 enabled=True, init_options=dict(), settings=dict(), environ=dict()):
+                 enabled=True, init_options=dict(), settings=dict(), env=dict()):
         self.name = name
         self.binary_args = binary_args
         self.scopes = scopes
@@ -467,7 +467,7 @@ class ClientConfig(object):
         self.enabled = enabled
         self.init_options = init_options
         self.settings = settings
-        self.environ = environ
+        self.env = env
 
 
 def format_request(payload: 'Dict[str, Any]'):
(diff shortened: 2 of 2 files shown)


LSP.sublime-settings
```diff
@@ -6,7 +6,14 @@
...
       "syntaxes": ["Packages/Python/Python.sublime-syntax"],
-      "languageId": "python"
+      "languageId": "python",
+      "environment": {
+        "PYTHONPATH": [
+            "${sublime_libs}",
+            "${packages}",
+            "${project_path}"
+        ]
+      }
```
tomv564: Forgot to remove testing code?

main.py
```diff
@@ -1773,10 +1775,28 @@ def start_client(window: sublime.Window, config: ClientConfig):
         window.status_message("Starting " + config.name + "...")
     debug("starting in", project_path)
 
+    # Create a dictionary of Sublime Text variables
     variables = window.extract_variables()
-    expanded_args = list(sublime.expand_variables(os.path.expanduser(arg), variables) for arg in config.binary_args)
+    variables["sublime_path"] = os.path.dirname(sublime.executable_path())
+    variables["sublime_libs"] = ';'.join(sys.path)
```
tomv564: Are these variables of use to other LSP users who are not Sublime Text package developers?
If not, please remove them. I requested this in the issue:

> If this means hardcoding the path to sublime.executable_path() in the project file that's probably ok?

main.py
```diff
@@ -336,7 +336,8 @@ def read_client_config(name, client_config):
         client_config.get("languageId", ""),
         client_config.get("enabled", True),
         client_config.get("initializationOptions", dict()),
-        client_config.get("settings", dict())
+        client_config.get("settings", dict()),
+        client_config.get("environment", dict())
```
tomv564: Let's call it `env`, just like the build system (and as I'd requested in the issue)

main.py
```diff
@@ -1773,10 +1775,28 @@ def start_client(window: sublime.Window, config: ClientConfig):
...
+        for arg in config.binary_args
+    )
+
+    # Merge/expand OS environment variables
+    environ = os.environ.copy()
+    for var, value in config.environ.items():
+        # Merge vars (e.g.: PATH=$PATH;mypath)
+        # For ease of editing, allow lists for each environment variable
+        if isinstance(value, list):
+            value = ';'.join(value)
```
tomv564: As this is the first implementation, let's avoid adding extra features from the start.

The sublime build system keeps it simple:
https://github.com/twolfson/sublime-files/blob/7d2589861ca00294734379943e8b7cffb7fbac88/Packages/Default/exec.py#L55-L58

Please remove the `list` support and the sublime expand_variables.

LSP.sublime-settings
deathaxe: Was not sure how to quickly provide an example of how it works. Will remove it.

main.py
deathaxe: Those variable enable the usage of `${sublime_path}` and `${sublime_libs}` variables in the settings file next to `${project}` etc. such as in the build-system. Adding them to `variables` does neither harm nor does it take any time. The whole process of reading/expanding the environment variables takes about 0.2ms upon server startup. The `${sublime_path}` may help people using portable setups with files located relative to the sublime text executable, while `${sublime_libs}` may be most useful for package authors. But as it contains a variable set of library paths including dependencies it is not so easy to replace by hard coded paths - makes life just much easier.

These are my arguments to keep it.

tomv564: I appreciate your arguments, but they confirm my belief that it is an edge case.

The whole `env` feature is a convenience only you have requested so far. Once it has accumulated some users we can always add in functionality. Removing functionality is *much* harder.

main.py
deathaxe: > The sublime build system keeps it simple

It is one line extra compared to the linked solution and is an optional feature which does not need to be used by anyone and not really more complicated.

If people want, they can still add 2048 or more chars long and over 15 lines wrapped lists of paths in one single string. But if a bit more overview is desired, they can split it into a list of strings. As JSON does not support multiline strings it is the only way to something like that and its harmless.

I thought about removing the ';' deliminator as it may not work with Linux/OSX. This would turn the list in a real mulitline-string-only-replacement.

I used this strategy for GitGuter's status_message templete already, which is read and joined everytime something changes - not only upon startup.

https://github.com/jisaacks/GitGutter/blob/24e254e24ef4ca57586e6f5e750da15cb45d3a0d/modules/show_diff.py#L133

An it is included in the 0.2ms required to handle the environment variable expansion. Removing it will not save any time at all.

main.py
deathaxe: > The whole env feature is a convenience only you have requested so far ...

Not to forget guys like the one who requested https://github.com/palantir/python-language-server/issues/143 😄 

Most people are just happy to consume. 

Is keeping the exe path for portable users ok?

tomv564:
Thanks a lot for pioneering this configuration, I can confirm it enables completions for `sublime` when using python-language-server.

The multi-line merging functionality I removed after I caught myself assuming it would insert `;` or `:` between lines, which of course it can't. Your example config didn't show delimiters, and then I found this gotcha about PYTHONPATH as well:

> On Windows, multiple paths are separated by semicolons. On Mac OS or Linux, they are separated by colons.

I ended up with this configuration in my LSP.sublime-project:
```
"settings": {
  "LSP": {
    "pyls": {
      "env": {
        "PYTHONPATH": "/Applications/Sublime Text.app/Contents/MacOS:${packages}"
      }
    }
  }
}
```

deathaxe:
> ... I caught myself assuming it would insert ; or : between lines, which of course it can't

I removed the `;` delimiter with my latest commit to turn it into a pure-multiline-string.

>> On Windows, multiple paths are separated by semicolons. On Mac OS or Linux, they are separated by colons.

It would have been fixed by using `os.pathsep.join(values)`.

main.py
evandrocoan: > @tomv564 The whole env feature is a convenience only you have requested so far.

Me too.

PR #136, opened 2017-09-30
Add a setting to display either completion kind or details

deathaxe:
Prior to this commit the completion panel always displays the text provided by the `details` item of the language server response and falls back to the item `kind` only if `details` is not available.

Some language-servers provide quite long details which can't be displayed correctly by ST's completion panel and in some cases don't even provide further information as they are more or less a copy of `label`.

Therefore this commit adds a setting `show_completion_details` which is set `true` by default to enable LSP to always display the completion kind rather than details.

To ensure, to gracefully fail, if none of the fields is available, the CompletionItemKind object is extended to provide an `Unknown` item.

The label and the details of a completion item returned by `format_completion()` are separated by ` \t ` rather than `\t` as ST tends to display no space between them if the details string is too long otherwise.

_Note:_ All settings in the _LSP.sublime-settings_ should provide proper description comment with a list of quoted valid values to support settings tooltips and completions provided by PackageDev


tomv564:
> Some language-servers provide quite long details which can't be displayed correctly by ST's completion panel and in some cases don't even provide further information as they are more or less a copy of label

Which _language servers_ did you test?

The protocol suggests the following of the content of CompletionItem's fields:

```typescript
/**
 * A human-readable string with additional information
 * about this item, like type or symbol information.
 */
detail?: string;
/**
 * A human-readable string that represents a doc-comment.
 */
documentation?: string;
```

I would prefer if we fixed the language server issues first, and then revisit this PR.

main.py
```diff
@@ -2152,18 +2156,14 @@ def do_request(self, prefix, locations):
...
-                detail = completion_item_kind_names[kind]
-        insertText = item.get("insertText", None)
-        if not insertText:
-            insertText = label
+            detail = completion_item_kind_names[item.get("kind", CompletionItemKind.Unknown)]
+        insertText = item.get("insertText") or label
         if insertText[0] == '$':  # sublime needs leading '$' escaped.
             insertText = '\$' + insertText[1:]
-        return "{}\t{}".format(label, detail) if detail else label, insertText
+        return "{} \t {}".format(label, detail), insertText
```
rwols: Why the extra spaces between the tab character?

deathaxe: > The label and the details of a completion item returned by format_completion() are separated by \t rather than \t as ST tends to display no space between them if the details string is too long otherwise.

deathaxe:
> Which language servers did you test?

I use only python-language-server. It returns `def function_name` for a completion with the label `function_name`. Keywords display most likely the definition line. Some of them seem to be very long and don't fit into the panel.

As both are useless, Jedi - Autocompletions and Anaconda both display just the completion type.

I don't know what other language servers return in this field. Maybe it is something which needs to be issued in python-language-server, but it returns just Jedi's output. The issue ST's completion panel is not able to render some items well won't be handled there. One could file an issue at Jedi to ask to avoid such long details, ... . In the end the one or the other rendering task remains in the client. As the completion kind is already returned as integer, it is quite easy to offer an option how many details to show. VS Code uses the kind to display different icons - ST can't. 

VS Code shows these information only for highlighted items, so they don't disturb

screenshot

But ST shows them next to each item without any space in several cases. It's very distracting and sometimes even hard to read.

screenshot

In the end it's almost no effort to offer an option to let the user decide what to show.

main.py
rwols: Whoops, should have read all of your paragraphs! 👍

tomv564:
Thanks for the details and the screenshots! For the record I am for adding this option but I want to improve the experience for *all* `python-language-server` users first if we can.

I decided to install PyCharm as it is pretty highly-regarded (by those who tolerate a Java-based IDE ;)), also to test the PYTHONPATH experience.

PyCharm chooses to show the containing module's name. Jedi also has a `module_name` member in the Completion, so I assume it is technically achievable in `python-language-server`.
What do you think, should we propose this?

screen shot 2017-10-01 at 09 47 06

screen shot 2017-10-01 at 09 47 28

These completions also remind me how much more recognisable the type of a completion item is when there is a icon just before the name.

deathaxe:
### Now

I am 100% with you. We need to find a common way with best results for all language-servers with least effort and without language dependent stuff being added to the main module at the moment. 

You are right about the icons. I tried to use unicode characters in other projects to achieve something comparable, but it is quite hard to find good matching chars for everything.

The main issue is to avoid useless repetitions. The completion hints repeat the label in many cases, right now, but the kind of a completion is not visible anywhere, if not part of the details.

I like the idea of printing the module which contains a function as detail. But if it's language specific, I am not sure whether to use it at the moment. We should not create too many edge cases.

I also like the `()` being added to the label of functions to state them clearly as such. Creating completions with parentheses being added on commit would also be very nice and was even requested by a python dev in the ST forum somewhere, recently. But the rule how to create such snippets might be language dependent as well. 

### Future _(maybe off-topic)_

Anything else might be added in the future, if the structure of the package allows language specific extensions. This is why I repetitively request(ed) to split up the main module. 😉 At some point there will just always be some language-specific details which need to be handled even though the goal should be to find the biggest common base in the first place. Therefore @rwols's idea of base classes, which are derived by certain languages sounds quite sufficient.

With completions as an example it would be really helpful to have one `LSPCompletionHandler` base class defined somewhere in the package, which handles all the communication and state stuff and creates some override-able methods which could be hooked by an successor. 

```python
class CompletionHandler(object):

    def format_completion(self, item):
        """Format default completions."""
        label = item["label"]
        detail = item.get("detail")
        insertText = item.get("insertText") or label
        return "{} \t {}".format(label, detail), insertText


class PythonCompletionHandler(CompletionHandler):

    def format_completion(self, item):
        """Handle python specific completions."""
        label = item["label"]
        detail = item.get("detail") if show_completion_details else None
        if not detail:
            detail = completion_item_kind_names[item.get("kind", CompletionItemKind.Unknown)]
        insertText = item.get("insertText") or label
        if insertText[0] == '$':  # sublime needs leading '$' escaped.
            insertText = '\$' + insertText[1:]
        return "{} \t {}".format(label, detail), insertText

```

Not yet quite sure to achieve that, but something like that would be needed some day to handle language-specific details of such things.

This approach some how sounds like using LSP as a base library to create the real language servers using derived classes with the ability to create common servers using the base classes like at the moment.

deathaxe:
Another question might be which parts to be the language-server responsible for? One could also request each language server to provide proper details in a common way.

tomv564:
Tempting to reply to all your comments but we are seriously derailing now :) I hear you on the architecture concerns!

As should be linked in this PR, I created an issue (https://github.com/palantir/python-language-server/issues/149) for python-language-server to consider switching to showing the parent type.

main.py
tomv564: Would just one space suffice?

main.py
```diff
@@ -2152,18 +2156,14 @@ def do_request(self, prefix, locations):
...
-        kind = item.get("kind")
+        label = item["label"]
+        detail = item.get("detail") if show_completion_details else None
         if not detail:
-            if kind is not None:
-                detail = completion_item_kind_names[kind]
-        insertText = item.get("insertText", None)
-        if not insertText:
-            insertText = label
+            detail = completion_item_kind_names[item.get("kind", CompletionItemKind.Unknown)]
```
tomv564: So if a language server doesn't support sending `detail` or `kind`, the user will always see "Unknown" as a completion hint ?

deathaxe: I just see, what you mean. The previous solution to show nothing is better. I somehow misunderstood the ` if detail else label` in line 2166.

main.py
deathaxe: If completion label and/or the hint contains spaces and many long items of different length are displayed using 3 spaces seemed to be more readable. The italic style only doesn't cause a good distinction. May be a personal impression. What do you think?

1 space:

screenshot_before


2 spaces:

screenshot_before

main.py
deathaxe: The question just might be: If there are many items with valid hints (kinds) in the list and few don't display any, they might be realized as word completions only and the gap may look a bit strange. With a hint of _Unknown_ its obvious to be an item returned by the language-server, at least. 

Not sure, which one is the better solution.

main.py
tomv564: Agree, 2 is better!

main.py
tomv564: Yes, we can let the inconsistent "gap" in the case of the missing / mismatched kind be a hint that the language server needs a fix.
I also would like some visibility if the server is returning invalid `kind` values, so using `Unknown` for that case feels a bit too polished, perhaps `???` would draw some attention to the issue.

tomv564:
Thanks for the extra performance tweak, agree about fixing casing issues as we go, but I felt `insert_text` was a bit more instructive than `contents`.

PR #137, opened 2017-09-30
Fix exception when handling empty signature parameters.

deathaxe:
Prior to this commit an exception is raised if the function signature returned by a language server does neither contain a `parameters` nor `params` field.

The error message was:

```
   LSP: active signature {'label': 'active_window()', 'documentation': ''}
   LSP: params None
   LSP: error handling response 10
   LSP: Error handling server content: 'NoneType' object is not iterable
```

This commit fixes this issue by a fallback to an empty dictionary.


tomv564:
Thanks for fix, took the opportunity to clean up an old workaround!

PR #143, opened 2017-10-02
Add support for PackageDev settings tooltips

deathaxe:
PackageDev uses the comments above each setting to display a tooltip, if the cursor is hovered over an existing setting.

screenshot

This commit adds required comments to the LSP.sublime-settings file.

An empty "clients" object was added to prevent PackageDev's linter from highlighting the key as unknown.

The template for user defined clients is moved to the "clients" setting and a warning was added to the comment of "default_clients" to prevent a user from accidentally manipulating the wrong setting to do overrides.

The template was formatted to hopefully display well in the tooltip for most users.

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -139,8 +139,8 @@
   // Show in-line diagnostics using phantoms for unchanged files.
   "show_diagnostics_phantoms": false,
 
-  // Show diagnostics description in status bar
-  // if the cursor hovers complained code.
+  // Show the diagnostics description of the code
+  // under the cursor in status bar if available.
   "show_diagnostics_in_view_status": true,
 
   // Highlighting style of code diagnostics.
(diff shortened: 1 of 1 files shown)


LSP.sublime-settings
```diff
@@ -97,19 +124,49 @@
...
+
+  // Open and close the diagnostics panel automatically,
+  // depending on available diagnostics.
   "auto_show_diagnostics_panel": true,
+
+  // Show in-line diagnostics using phantoms for unchanged files.
   "show_diagnostics_phantoms": false,
+
+  // Show diagnostics description in status bar
+  // if the cursor hovers complained code.
```
tomv564: Should be something like "when the cursor is on code with diagnostics"

deathaxe: I am not too happy with it, too. Wanted to avoid duplicated usage of the word diagnostics.

Maybe:

```
  // Show the diagnostics description of the code
  // under the cursor in status bar if available.
```
or so?

tomv564: :+1:

Issue #154, opened 2017-10-05
LSP does not send didClose


deathaxe:
The function `notify_did_close()` is called but the `view.window()` returns None. Therefore `config_for_scope()` does so too, which prevents the didClose notification to be sent.

- ST 3144 x64
- LSP master

PR #155, opened 2017-10-06
view.window() can be none, workaround for closing

tomv564:
For #154


deathaxe:
After a quick check, it solves the issue. didClose is sent now.

tomv564:
Thanks for testing!

PR #156, opened 2017-10-06
Cancel completion after inserted tabs.

### LINKED ISSUE #152: completion menu shows too late after "tab" is hit
When <kbd>tab</kbd> is hit and there are no available completions of a given prefix, Sublime Text will trigger `insert_best_completion` and insert at a tab or spaces. 

With LSP, LSP needs to wait for the server's response to compute the completion list. While waiting for the response, `insert_best_completion` has already inserted the tab or spaces. 

And it is what we will see
oct-05-2017 13-13-29

deathaxe:
fixes #152

Disable/Hide completion popup
1. while request is on the fly to avoid flicker on Windows
2. if the cursor is preceded by a whitespace

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -115,6 +115,7 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
         self.initialized = False
         self.enabled = False
         self.trigger_chars = []  # type: List[str]
+        self.completions = []  # type: List[Tuple[str, str]]
         self.resolve = False
         self.resolve_details = []  # type: List[Tuple[str, str]]
         self.state = CompletionState.IDLE
@@ -152,7 +153,6 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
         if self.enabled:
             if self.state == CompletionState.IDLE:
                 self.do_request(prefix, locations)
-                self.completions = []  # type: List[Tuple[str, str]]
 
             elif self.state in (CompletionState.REQUESTING, CompletionState.CANCELLING):
                 self.next_request = (prefix, locations)
@@ -160,6 +160,9 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
 
             elif self.state == CompletionState.APPLYING:
                 self.state = CompletionState.IDLE
+
+            if self.state != CompletionState.IDLE:
+                self.view.run_command('hide_auto_complete')
 
             return (
                 self.completions,
@@ -182,7 +185,7 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
             if document_position:
                 client.send_request(
                     Request.complete(document_position),
-                    self.handle_response)
+                    lambda response: self.handle_response(response, prefix))
                 self.state = CompletionState.REQUESTING
 
     def format_completion(self, item) -> 'Tuple[str, str]':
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -2171,6 +2171,9 @@ def on_query_completions(self, prefix, locations):
             elif self.state == CompletionState.APPLYING:
                 self.state = CompletionState.IDLE
 
+            if self.state != CompletionState.IDLE:
+                self.view.run_command('hide_auto_complete')
+
```
randy3k: I am not getting any completion menu with these lines.

randy3k:
I am not getting any completion menu with this PR.

macOS, Sublime Text 3147

deathaxe:
I've added some debugs to get and idea what's wrong with OSX. Works on Windows correctly.

Typing `y` in python file causes the following messages for me...

```
LSP: notify: textDocument/didChange
LSP: request 4: textDocument/completion 
LSP: requesting completions
LSP: got json:  {"result":{"isIncomplete":false,"items":[{"label":"yield","kind":14, ...
LSP: display completions
```

Typing `y` followed by whitespace ...

```
LSP: notify: textDocument/didChange
LSP: request 30: textDocument/completion 
LSP: requesting completions
LSP: got json:  {"result":{"isIncomplete":false,"items":[{"label":"yield", ...
LSP: skip completions
```

main.py
```diff
@@ -2171,6 +2171,10 @@ def on_query_completions(self, prefix, locations):
             elif self.state == CompletionState.APPLYING:
                 self.state = CompletionState.IDLE
 
+            if self.state != CompletionState.IDLE:
+                debug("requesting completions")
+                self.view.run_command('hide_auto_complete')
```
deathaxe: @randy3k : Can you check, whether removing this line fixes your issue? It is in the end the only change I could imagine to cause completions not to be displayed at all under some circumstances.

randy3k:
It's odd. It is working now perfectly. Thanks.

randy3k:
A partially related issue though,

1. type yi
2. tab
3. tab

What is happening: the second tab triggers `replace_completion_with_next_completion` and somehow it may generate a new completion list or use a previously generated list.

oct-07-2017 14-59-42

deathaxe:
The whole `auto_complete` , `insert_best_completion`, `replace_completion_with_next_completion` stuff is indeed a bit weird. It looks like an overlay of different completion solutions which fight against each other. 

After first tab the `commit_completion` is called. The second tab starts `insert_best_completion` which would normally replace `yield` with the next completion. All following tabs would call `replace_completion_with_next_completion` until no more completions are available.

The way LSP needs to handle completions breaks this procedure as it requires the completions to be returned immediately instead of calling `auto_complete` some milliseconds later.

You are right, all three functions trigger `on_query_completions` event handler and therefore a completion request to the language-server. I can't see a way to find out which command fired it. LSP handles all three requests as normal `auto_complete` request and tries to show the completion panel, followed by a `commit_completion`. 

The only way to solve this issue was to block within `on_query_completions` until the language server returns a value to make ST think it's synchronous, but we don't want to do this. It causes ST to freeze during input , terrible.

The cleanest solution is to disable `"tab_completion"` setting in Preferences to avoid this issue. This disables the "in-place" completions.

main.py
deathaxe: Note: This line here hides the completion panel until the completions of a language server have arrived.

randy3k:
Another solution is to cache the LSP completion list. When `on_query_completions` is called, if the last command is `insert_best_completion`, we set `self.completions` to the previous list directly instead of querying the server.

Though, it seems that `view.command_history(0)` is not functional. It returns a `insert` command instead of `insert_best_completion`.

Putting `print(self.view.command_history(0))` in `on_query_completions` produces this
```
('insert', {'characters': 'y'}, 1)
command: insert_best_completion {"default": "\t", "exact": false}
command: replace_completion_with_next_completion
('insert', {'characters': 'yi'}, 1)
```
when I reproduce the above example. Maybe `replace_completion_with_next_completion` triggers the `insert` command before calling `on_query_completions`!? An alternative appoarch is to use `sublime_plugin.EventListener.on_text_command`. I just checked it, it should work with `on_query_completions`.

randy3k:
Perhaps we should merge this first and deal with the `replace_completion_with_next_completion` issue later.

deathaxe:
Have exactly the same possibilities in mind, but ...

1. The CompletionHandler used by LSP is a ViewEventListener, which does not support on_text_command.
2. You are right about the commands. `insert_best_completions` tries to decide, whether to trigger completions or insert a char and therefore makes use of subsequent commands.
3. I thought about caching results, too, but both solutions may cause a couple of edge cases as they depend on ST internal functionalities which might change. I struggled with such kind of issues in PackageDev recently, as it always added some dots somewhere under some circumstances. It uses `on_text_command` to find out when a completion was committed or canceled. Not sure whether we should go this way.

deathaxe:
@randy3k : Possibly found a solution for the most recent issue. Can you check it?

randy3k:
Prefect! It seems working.

randy3k:
why is it closed?

PR #157, opened 2017-10-06
Shutdown language-server by protocol specification.

deathaxe:


main.py
```diff
@@ -505,6 +509,18 @@ def has_capability(self, capability):
     def get_capability(self, capability):
         return self.capabilities.get(capability)
 
+    def shutdown(self):
+        self.send_request(Request.shutdown(), self.exit)
```
rwols: Maybe we should make this a blocking request instead of a non-blocking one?

tomv564: I'd like to understand the case where this is important, and I'd like to know the impact to sublime text / LSP's own process before we do this.

main.py
```diff
@@ -505,6 +509,24 @@ def has_capability(self, capability):
     def get_capability(self, capability):
         return self.capabilities.get(capability)
 
+    def shutdown(self):
+        try:
+            self.send_request(Request.shutdown(), self.exit)
+            self.process.wait(timeout=5000)
+        except OSError as err:
+            exception_log("Error shutdown server", err)
```
tomv564: :+1: We MUST show errors when language servers misbehave. Please change the message to "Error shutting down server".

main.py
```diff
@@ -718,43 +736,46 @@ def start_active_view():
...
-def unload_window_clients(window_id: int):
+def unload_window_clients(window_id: int) -> None:
     global clients_by_window
-    if window_id in clients_by_window:
-        window_clients = clients_by_window[window_id]
-        del clients_by_window[window_id]
-        for config, client in window_clients.items():
-            debug("unloading client", config, client)
-            unload_client(client)
+    clients_by_config = clients_by_window.pop(window_id)
```
tomv564: Nice improvement here

main.py
```diff
@@ -718,43 +736,46 @@ def start_active_view():
...
+def unload_old_clients(window: sublime.Window):
+    project_path = get_project_path(window)
+    clients_by_config = window_clients(window)
+    clients_to_unload = []
+    for config_name, client in clients_by_config.items():
+        if client and client.get_project_path() != project_path:
+            debug('unload', config_name, 'project path changed from ', client.get_project_path())
+            clients_to_unload.append(config_name)
+    for config_name in clients_to_unload:
+        clients_by_config.pop(config_name).shutdown()
```
tomv564: Mypy should complain that pop() can return `None` here? (Not that this _will_ happen at runtime, but  it should reveal the issue anyways)

main.py
```diff
@@ -718,43 +736,46 @@ def start_active_view():
...
-            closed_windows.append(id)
+    for clients_by_config in clients_by_window.values():
+        for client in clients_by_config.values():
+            client.shutdown()
+    clients_by_window.clear()
+
+
+def check_window_unloaded() -> None:
+    open_window_ids = set(window.id() for window in sublime.windows())
+    closed_windows = set(id for id in clients_by_window if id not in open_window_ids)
```
tomv564: Nice cleanup :+1:

deathaxe:
> These PRs continue to grow beyond the original intent (here: let's make it blocking just in case, and let's moving some stuff around while we're at it).

Not a bad will, sorry. 

1. The blocking stuff was added after being requested. Didn't have that in mind.
2. Try to think of splitting up commits into changes/refactors. But it's honestly sometimes hard to 
  a) find the root through the code as the whole functions are scattered all over the file
  b) and therefore to decide where to cut as many things (don't want to call it issue) are revealed during search.

Those "refactoring" changes came up during testing and debugging changes. See the exception handling in the shutdown() function as an example. I honestly found a bunch of further issues, which need to be handled by starting/stopping servers, which I did not even mention, but from my point of view belong to this topic. The Client objects as an example seem never to be cleaned up by the garbage collector, ...

So I really tried to limit the changes as much as possible without just leaving some hacked lines of code.

Again, it's not to bother you.

main.py
deathaxe: Nope. Don't have any warning here. Neither in the diagnostics panel nor by running mypy directly.

rwols:
> These PRs continue to grow beyond the original intent

Apologies, what I had in mind was that the functionality from that willSave PR could be used once this got merged. You can always ask people to revert stuff and do changes in smaller steps by the way.

rwols:
I tried this out with an (I believe) correct shutdown+exit sequence for clangd. I get this in my console:
```
  File "/Users/raoulwols/Library/Application Support/Sublime Text 3/Packages/LSP/plugin/core/clients.py", line 68, in check_window_unloaded
    unload_window_clients(closed_window_id)
  File "/Users/raoulwols/Library/Application Support/Sublime Text 3/Packages/LSP/plugin/core/clients.py", line 75, in unload_window_clients
    client.shutdown()
  File "/Users/raoulwols/Library/Application Support/Sublime Text 3/Packages/LSP/plugin/core/rpc.py", line 58, in shutdown
    self.terminate()
  File "/Users/raoulwols/Library/Application Support/Sublime Text 3/Packages/LSP/plugin/core/rpc.py", line 65, in terminate
    self.process.terminate()
  File "./python3.3/subprocess.py", line 1765, in terminate
  File "./python3.3/subprocess.py", line 1760, in send_signal
```

deathaxe:
Looks like python behaves different on MacOS. On windows it fails silently if the process no longer exists. So need to catch that exception. Thanks.

rwols:
Yep, that fixes it.

Ideally, we should check for exit status 0. According to protocol the server returns 0 if `shutdown` was sent before `exit` and 1 otherwise.

deathaxe:
Any benefit from it? More than a message?

rwols:
A message would be nice. Imagine a server serializing data to disk in-between the shutdown request and shutdown reply. But maybe that's for another PR ;)

rwols:
I really want this to get merged :/

PR #162, opened 2017-10-07
Browse function overloads with the up/down arrow keys (with no other stuff)

rwols:
This just adds the browsing functionality, let's leave the parameter highlighting for another PR.

Note the usage of `mdpopups.update_popup`now.

### DIFF
--- a/main.py
+++ b/main.py
@@ -2269,16 +2269,19 @@ class SignatureHelpListener(sublime_plugin.ViewEventListener):
 
     def __init__(self, view):
         self.view = view
-        self.signature_help_triggers = None
+        self._initialized = False
+        self._signature_help_triggers = []  # type: List[str]
         self._visible = False
         self._language_id = ""
+        self._signatures = []  # type: List[Any]
+        self._active_signature = -1
 
     @classmethod
     def is_applicable(cls, settings):
         syntax = settings.get('syntax')
         return syntax and is_supported_syntax(syntax)
 
-    def initialize_triggers(self):
+    def initialize(self):
         client = client_for_view(self.view)
         if client:
             signatureHelpProvider = client.get_capability(
@@ -2286,19 +2289,20 @@ class SignatureHelpListener(sublime_plugin.ViewEventListener):
             if signatureHelpProvider:
                 self.signature_help_triggers = signatureHelpProvider.get(
                     'triggerCharacters')
-                return
 
-        self.signature_help_triggers = []
+        config = config_for_scope(self.view)
+        if config:
+            self._language_id = config.languageId
+
+        self._initialized = True
 
     def on_modified_async(self):
         pos = self.view.sel()[0].begin()
         last_char = self.view.substr(pos - 1)
         # TODO: this will fire too often, narrow down using scopes or regex
-        if self.signature_help_triggers is None:
-            self.initialize_triggers()
-        config = config_for_scope(self.view)
-        if config:
-            self._language_id = config.languageId
+        if not self._initialized:
+            self.initialize()
+
         if self.signature_help_triggers:
             if last_char in self.signature_help_triggers:
                 client = client_for_view(self.view)
--- a/main.py
+++ b/main.py
@@ -2375,15 +2375,16 @@ class SignatureHelpListener(sublime_plugin.ViewEventListener):
 
     def _build_popup_content(self) -> str:
         signature = self._signatures[self._active_signature]
-        debug("active signature", signature)
         formatted = []
+
         if len(self._signatures) > 1:
-            label = "**{}** of **{}** overloads (use the arrow keys to navigate):\n```{}\n{}\n```\n".format(
-                str(self._active_signature + 1), str(len(self._signatures)),
-                self._language_id, signature.get('label'))
-        else:
-            label = "```{}\n{}\n```".format(self._language_id, signature.get('label'))
+            signature_navigation = "**{}** of **{}** overloads (use the arrow keys to navigate):\n".format(
+                str(self._active_signature + 1), str(len(self._signatures)))
+            formatted.append(signature_navigation)
+
+        label = "```{}\n{}\n```".format(self._language_id, signature.get('label'))
         formatted.append(label)
+
         params = signature.get('parameters')
         if params:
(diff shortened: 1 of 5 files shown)


stubs/mdpopups.pyi
```diff
@@ -17,3 +17,14 @@ def show_popup(
...
     allow_code_wrap=False
 ): ...
+
+def update_popup(
+    view: sublime.View,
+    content: str,
+    md=True,
+    css=None,
+    wrapper_class=None,
+    template_vars=None,
```
tomv564: Can you check what typings these should be?

main.py
```diff
@@ -2303,40 +2309,69 @@ def on_modified_async(self):
...
+                                    on_hide=self._on_hide)
+                self._visible = True
+
+    def on_query_context(self, key, _, operand, __):
+        if key != "lsp.signature_help":
+            return False  # Let someone else handle this keybinding.
+        elif not self._visible:
+            return False  # Let someone else handle this keybinding.
+        elif self.active_signature not in range(0, len(self.signatures)):
+            return False  # Let someone else handle this keybinding.
```
tomv564: What happens if we just check for `-1` here?

main.py
```diff
@@ -2303,40 +2309,69 @@ def on_modified_async(self):
...
-                    preserve_whitespace("\n".join(formatted)),
-                    css=".mdpopups .lsp_signature { margin: 4px; } .mdpopups p { margin: 0.1rem; }",
-                    md=True,
-                    flags=sublime.HIDE_ON_MOUSE_MOVE_AWAY,
-                    location=point,
-                    wrapper_class="lsp_signature",
-                    max_width=800)
+            self._config = config_for_scope(self.view)
+            self.signatures = response.get("signatures")
+            self.active_signature = response.get("activeSignature", -1)
```
tomv564: Can you add `signatures` and `active_signature` to `__init__` ?

main.py
```diff
@@ -2303,40 +2309,69 @@ def on_modified_async(self):
...
-                mdpopups.show_popup(
-                    self.view,
-                    preserve_whitespace("\n".join(formatted)),
-                    css=".mdpopups .lsp_signature { margin: 4px; } .mdpopups p { margin: 0.1rem; }",
-                    md=True,
-                    flags=sublime.HIDE_ON_MOUSE_MOVE_AWAY,
-                    location=point,
-                    wrapper_class="lsp_signature",
-                    max_width=800)
+            self._config = config_for_scope(self.view)
```
tomv564: Can we store only `languageId`, and do it once in `on_modified_async` just like `initialize_triggers`?

main.py
```diff
@@ -2263,9 +2263,15 @@ def run_auto_complete(self):
...
 class SignatureHelpListener(sublime_plugin.ViewEventListener):
+
+    css = ".mdpopups .lsp_signature { margin: 4px; } .mdpopups p { margin: 0.1rem; }"
+    wrapper_class = "lsp_signature"
+
     def __init__(self, view):
         self.view = view
         self.signature_help_triggers = None
+        self._visible = False
+        self._config = None  # type: ClientConfig
```
tomv564: This should be `Optional[ClientConfig]` but prefer replacing with languageId.

main.py
```diff
@@ -2303,40 +2309,69 @@ def on_modified_async(self):
...
-                    md=True,
-                    flags=sublime.HIDE_ON_MOUSE_MOVE_AWAY,
-                    location=point,
-                    wrapper_class="lsp_signature",
-                    max_width=800)
+            self._config = config_for_scope(self.view)
+            self.signatures = response.get("signatures")
+            self.active_signature = response.get("activeSignature", -1)
+            debug("got signatures, active is", len(self.signatures), self.active_signature)
+            if len(self.signatures) > 0 and self._config:
```
tomv564: Updating the popup shouldn't care about `_config`, `_build_popup_content` can do its own checking and still function if `languageId` is not available.

main.py
```diff
@@ -2303,40 +2309,69 @@ def on_modified_async(self):
...
+
+    def _on_hide(self):
+        self._visible = False
+
+    def _build_popup_content(self) -> str:
+        signature = self.signatures[self.active_signature]
+        debug("active signature", signature)
+        formatted = []
+        if len(self.signatures) > 1:
+            label = "```{}\n{}/{} {}\n```".format(self._config.languageId, str(self.active_signature + 1),
```
tomv564: It might be less confusing to the user if the `1 / 2` was not formatted as a code block like the signature itself. Would it be okay to show it below or above the code block?

main.py
rwols: When there are zero signatures, you might get an `activeSignature` of `0` from a server, and a `signatures == []`. In that case, `signatures[active_signature]` will raise an out of bounds exceptions. I like this approach of checking the range because it accounts for both invalid numbers for the active signature as well as for an empty signature list.

tomv564: We should guard against this case in `handle_response` and at least log an error.
See https://github.com/Microsoft/vscode-languageserver-node/issues/191 for a discussion with one of the protocol maintainers about defaulting to 'null' instead of '0'

rwols: I see. I should fix that in clangd then 💃. For the time being, I think checking if the active signature is not out-of-bounds is a good safe-guard for faulty servers.

tomv564:
Thanks for the fixes, works great for me in javascript / typescript!

I did some cleanups, please let me know if I missed your intentions somewhere.
Otherwise I'll merge this tonight.

rwols:
LGTM

main.py
```diff
@@ -2303,40 +2315,87 @@ def on_modified_async(self):
...
+        elif not self._visible:
+            return False  # Let someone else handle this keybinding.
+        elif len(self._signatures) < 2:
+            return False  # Let someone else handle this keybinding.
+        else:
+            # We use the "operand" for the number -1 or +1. See the keybindings.
+            new_index = self._active_signature + operand
+
+            # clamp signature index
+            new_index = max(0, min(new_index, len(self._signatures) - 1))
```
rwols: This changes the behaviour, but I'm okay with this.

PR #163, opened 2017-10-07
Look at the last non-whitespace character for signature help triggers

rwols:
Look at the last non-whitespace character for signature help triggers.


tomv564:
Can you give an example of when this is needed?

rwols:
Sure:
lastchar

Observe that the popup disappears when you press a space after the `(` or `,` before this PR.

randy3k:
Should it be the job of the server? I just wonder how vscode triggers it.

rwols:
No it's up to the client (us), because there is no way for the server to know where the cursor is.

randy3k:
What if a server put `space` in `SignatureHelpOptions.triggerCharacters`. Though, one may argue that it should be avoided.

rwols:
Yeah, that would a problem (Objective-C? Although clangd doesn't advertise a space as a signature help trigger). I don't think there are many languages that use a space for separating the arguments, we can probably make exceptions for those.

randy3k:
At least we may check if space is in `signature_help_triggers`

rwols:
Cool, thanks to mypy I discovered a bug already (4bcf353).

tomv564:
Agree a fix is needed for this.

Putting a space in `signature_help_triggers` sounds like a terrible idea, let's not complicate LSP with handling for such scenarios

Since Sublime Text is already doing a lot of hard work parsing the code, could we do something with the scopes it provides while we are in a signature?

screen shot 2017-10-09 at 09 11 15

rwols:
> could we do something with the scopes it provides while we are in a signature?

Nice question, but I would say unfortunately no.

Ideally, we would receive the `meta.function-call.arguments` scope if we're inside a signature. And ideally, trigger characters are scoped as `punctuation.separator` / `punctuation.section`. So just check if we're inside `meta.function-call.arguments`, and then peek behind for a `punctuation`, no? Well, it's too risky to rely on that because in my experience scopes may differ wildly or be only partially implemented or be only partially according to the scope naming guide. This approach is purely text-based, but still uses ST's api to look behind.

rwols:
Rebased & force-pushed.

tomv564:
Thanks! Sorry for the delay in merging.

PR #166, opened 2017-10-08
Implement textDocument/rangeFormatting

rwols:
There might be race bugs when using this with multiple selections.

When this is invoked with multiple selections, I try to send requests from "highest range" to "lowest range", so that local edits don't affect other local edits.

I'm not too worried about this at the moment.

### DIFF
--- a/main.py
+++ b/main.py
@@ -179,6 +179,10 @@ class Request:
         return Request("textDocument/formatting", params)
 
     @classmethod
+    def rangeFormatting(cls, params: dict):
+        return Request("textDocument/rangeFormatting", params)
+
+    @classmethod
     def documentSymbols(cls, params: dict):
         return Request("textDocument/documentSymbol", params)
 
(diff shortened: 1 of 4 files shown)


main.py
```diff
@@ -1411,6 +1415,43 @@ def handle_response(self, response, pos):
...
+                "uri": filename_to_uri(self.view.file_name())
+            },
+            "range": Range.from_region(self.view, region).to_lsp(),
+            "options": {
+                "tabSize": 4,  # TODO: Fetch these from the project settings / global settings
+                "insertSpaces": True
+            }
+        }
+        if not self.client:
+            return
```
rwols: I couldn't manage to silence mypy without this check, but it is essentially redundant, because it is checked for in the `is_enabled` method. How can I tell mypy that?

main.py
```diff
@@ -1398,7 +1402,7 @@ def run(self, edit):
                     "uri": filename_to_uri(self.view.file_name())
                 },
                 "options": {
-                    "tabSize": 4,
+                    "tabSize": 4,  # TODO: Fetch these from the project settings / global settings
```
deathaxe: Just saw this merged. The tab size can be easily read by `"tabSize": self.view.settings().get("tab_size", 4)` This includes all available sources: view specific setting, project setting, user setting, default.

rwols: Yeah. It’s also hardcoded in the format document request, so I felt that copying that behavior was the best course of action. We should fix it in another PR (so fix both format doc and format selection options).

PR #168, opened 2017-10-08
Improve whitespace in diagnostics popup

### LINKED ISSUE #82: Display text in sans-serif font
VS Code's overlays use sans-serif fonts for regular text within overlays. It would be neat if we could do the same here.

(e.g. https://code.visualstudio.com/images/intellisense_paramater_info.png)

Given that this is potentially down to user preference, this could be a preference rather than a default.

(This is already possible through mdpopups' [user CSS preference](http://facelessuser.github.io/sublime-markdown-popups/settings/#mdpopupsuser_css), but IMO would make a good default.)

### LINKED ISSUE #105: Whitespace in popups could be better

braver:
Fixes #105 

Currently, descriptive text just sits in the parent element and cannot be targeted with CSS without also affecting code snippets. This PR solves that by wrapping it in a `div.text`.

However, if there is a newline between `div`s it is rendered as a blank line. This shouldn’t happen in regular HTML or Markdown, but this isn’t regular. This causes two problems:

1. A disproportionate amount of white space between code snippets and text. This is solved by shifting text half a line upwards, making it sit nicely centered vertically.
2. Succeeding lines have a blank line between them. This isn’t terrible, but it wasn’t there before the PR, so I solved it by removing the `\n` before a `div`. Fenced Code blocks should be followed by a newline so that’s ensured when the block is added.

I had a little side-issue with any text after a `\r\n` not being rendered. I dunno why, also don’t know why there would be a `\r\n`, must be something to do with TypeScript being a Microsoft thing and giving us Windows line endings or something. It also caused double newlines to not be respected. I solved that by converting `\r\n` to `\n`.

Content in a `div.text` is now also rendered with a sans-serif font, and thereby fixes #82 

Before
screen shot 2017-10-08 at 18 32 56

After
screen shot 2017-10-08 at 20 01 56

(replaces #167)

### DIFF
--- a/main.py
+++ b/main.py
@@ -193,6 +193,10 @@ class Request:
         return Request("textDocument/formatting", params)
 
     @classmethod
+    def rangeFormatting(cls, params: dict):
+        return Request("textDocument/rangeFormatting", params)
+
+    @classmethod
     def documentSymbols(cls, params: dict):
         return Request("textDocument/documentSymbol", params)
 
@@ -1417,7 +1421,7 @@ class LspFormatDocumentCommand(sublime_plugin.TextCommand):
                     "uri": filename_to_uri(self.view.file_name())
                 },
                 "options": {
-                    "tabSize": 4,
+                    "tabSize": 4,  # TODO: Fetch these from the project settings / global settings
                     "insertSpaces": True
                 }
             }
@@ -1428,6 +1432,36 @@ class LspFormatDocumentCommand(sublime_plugin.TextCommand):
     def handle_response(self, response, pos):
         self.view.run_command('lsp_apply_document_edit',
                               {'changes': response})
+
+
+class LspFormatDocumentRangeCommand(sublime_plugin.TextCommand):
+    def is_enabled(self):
+        if is_supported_view(self.view):
+            client = client_for_view(self.view)
+            if client and client.has_capability('documentRangeFormattingProvider'):
+                if len(self.view.sel()) == 1:
+                    region = self.view.sel()[0]
+                    if region.begin() != region.end():
+                        return True
+        return False
+
+    def run(self, _):
+        client = client_for_view(self.view)
+        if client:
+            region = self.view.sel()[0]
+            params = {
+                "textDocument": {
+                    "uri": filename_to_uri(self.view.file_name())
+                },
+                "range": Range.from_region(self.view, region).to_lsp(),
+                "options": {
+                    "tabSize": 4,  # TODO: Fetch these from the project settings / global settings
+                    "insertSpaces": True
+                }
+            }
+            client.send_request(Request.rangeFormatting(params),
+                                lambda response: self.view.run_command('lsp_apply_document_edit',
+                                                                       {'changes': response}))
 
 
 class LspSymbolDefinitionCommand(sublime_plugin.TextCommand):
@@ -2283,16 +2317,22 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
 
 
 class SignatureHelpListener(sublime_plugin.ViewEventListener):
+
     def __init__(self, view):
         self.view = view
-        self.signature_help_triggers = None
+        self._initialized = False
+        self._signature_help_triggers = []  # type: List[str]
+        self._visible = False
+        self._language_id = ""
+        self._signatures = []  # type: List[Any]
+        self._active_signature = -1
 
     @classmethod
     def is_applicable(cls, settings):
         syntax = settings.get('syntax')
         return syntax and is_supported_syntax(syntax)
 
-    def initialize_triggers(self):
+    def initialize(self):
         client = client_for_view(self.view)
         if client:
             signatureHelpProvider = client.get_capability(
@@ -2300,16 +2340,19 @@ class SignatureHelpListener(sublime_plugin.ViewEventListener):
             if signatureHelpProvider:
                 self.signature_help_triggers = signatureHelpProvider.get(
                     'triggerCharacters')
-                return
 
-        self.signature_help_triggers = []
+        config = config_for_scope(self.view)
+        if config:
+            self._language_id = config.languageId
+
+        self._initialized = True
 
     def on_modified_async(self):
         pos = self.view.sel()[0].begin()
         last_char = self.view.substr(pos - 1)
         # TODO: this will fire too often, narrow down using scopes or regex
-        if self.signature_help_triggers is None:
-            self.initialize_triggers()
+        if not self._initialized:
+            self.initialize()
 
         if self.signature_help_triggers:
             if last_char in self.signature_help_triggers:
@@ -2323,40 +2366,87 @@ class SignatureHelpListener(sublime_plugin.ViewEventListener):
                             lambda response: self.handle_response(response, pos))
             else:
                 # TODO: this hides too soon.
-                if self.view.is_popup_visible():
+                if self._visible:
                     self.view.hide_popup()
 
     def handle_response(self, response, point):
         if response is not None:
-            config = config_for_scope(self.view)
-            signatures = response.get("signatures")
-            activeSignature = response.get("activeSignature")
-            debug("got signatures, active is", len(signatures), activeSignature)
-            if len(signatures) > 0 and config:
-                signature = signatures[activeSignature]
-                debug("active signature", signature)
-                formatted = []
-                formatted.append(
-                    "```{}\n{}\n```\n".format(config.languageId, signature.get('label')))
-                params = signature.get('parameters')
-                if params:
-                    for parameter in params:
-                        paramDocs = parameter.get('documentation')
-                        if paramDocs:
-                            formatted.append("**{}**\n".format(parameter.get('label')))
-                            formatted.append("* *{}*\n".format(paramDocs))
+            self._signatures = response.get("signatures", [])
+            self._active_signature = response.get("activeSignature", -1)
 
-                formatted.append(signature.get('documentation'))
+            if self._signatures:
+                if not 0 <= self._active_signature < len(self._signatures):
+                    debug("activeSignature {} not a valid index for signatures length {}".format(
+                        self._active_signature, len(self._signatures)))
+                    self._active_signature = 0
+            else:
+                if self._active_signature != -1:
+                    debug("activeSignature should be -1 or null when no signatures are returned")
+                    self._active_signature = -1
 
-                mdpopups.show_popup(
-                    self.view,
-                    preserve_whitespace("\n".join(formatted)),
-                    css=popup_css,
-                    md=True,
-                    flags=sublime.HIDE_ON_MOUSE_MOVE_AWAY,
-                    location=point,
-                    wrapper_class="lsp_popup",
-                    max_width=800)
+            if len(self._signatures) > 0:
+                mdpopups.show_popup(self.view,
+                                    self._build_popup_content(),
+                                    css=popup_css,
+                                    md=True,
+                                    flags=sublime.HIDE_ON_MOUSE_MOVE_AWAY,
+                                    location=point,
+                                    wrapper_class="lsp_popup",
+                                    max_width=800,
+                                    on_hide=self._on_hide)
+                self._visible = True
+
+    def on_query_context(self, key, _, operand, __):
+        if key != "lsp.signature_help":
+            return False  # Let someone else handle this keybinding.
+        elif not self._visible:
+            return False  # Let someone else handle this keybinding.
+        elif len(self._signatures) < 2:
+            return False  # Let someone else handle this keybinding.
+        else:
+            # We use the "operand" for the number -1 or +1. See the keybindings.
+            new_index = self._active_signature + operand
+
+            # clamp signature index
+            new_index = max(0, min(new_index, len(self._signatures) - 1))
+
+            # only update when changed
+            if new_index != self._active_signature:
+                self._active_signature = new_index
+                mdpopups.update_popup(self.view,
+                                      self._build_popup_content(),
+                                      css=popup_css,
+                                      md=True,
+                                      wrapper_class="lsp_popup")
+
+            return True  # We handled this keybinding.
+
+    def _on_hide(self):
+        self._visible = False
+
+    def _build_popup_content(self) -> str:
+        signature = self._signatures[self._active_signature]
+        formatted = []
+
+        if len(self._signatures) > 1:
+            signature_navigation = "**{}** of **{}** overloads (use the arrow keys to navigate):\n".format(
+                str(self._active_signature + 1), str(len(self._signatures)))
+            formatted.append(signature_navigation)
+
+        label = "```{}\n{}\n```\n".format(self._language_id, signature.get('label'))
+        formatted.append(label)
+
+        params = signature.get('parameters')
+        if params:
+            for parameter in params:
+                paramDocs = parameter.get('documentation', None)
+                if paramDocs:
+                    formatted.append("**{}**\n".format(parameter.get('label')))
+                    formatted.append("* *{}*\n".format(paramDocs))
+        sigDocs = signature.get('documentation', None)
+        if sigDocs:
+            formatted.append(sigDocs)
+        return preserve_whitespace("\n".join(formatted))
 
 
 def get_line_diagnostics(view, point):
--- a/main.py
+++ b/main.py
@@ -2024,29 +2024,17 @@ class HoverHandler(sublime_plugin.ViewEventListener):
             if language:
                 formatted.append("```{}\n{}\n```\n".format(language, value))
             else:
-                formatted.append("<div class='text'>{}</div>".format(value))
+                formatted.append("<div class='description'>{}</div>".format(value))
 
         mdpopups.show_popup(
             self.view,
-            preserve_whitespace("\n".join(formatted)),
+            preserve_whitespace("".join(formatted)),
             css='''
-                .mdpopups .lsp_hover {
-                    margin: 0.25rem;
+                .mdpopups .lsp_hover .highlight {
+                   border-width: 0;
                 }
-                .mdpopups .lsp_hover div.highlight,
-                .mdpopups .lsp_hover pre.highlight {
-                    margin: 0;
-                }
-                .mdpopups .lsp_hover .text,
-                .mdpopups .lsp_hover p {
-                    margin: 0;
-                    padding: 0;
-                }
-                .mdpopups .lsp_hover p .text {
-                    position: relative;
-                    top: -0.5rem;
-                    margin: 0 .5rem;
-                    padding: 1px;
+                .mdpopups .lsp_hover .description {
+                    margin: 0 0.5rem;
                     font-family: sans-serif;
                 }
             ''',
@@ -2061,7 +2049,6 @@ def preserve_whitespace(contents: str) -> str:
     """Preserve empty lines and whitespace for markdown conversion."""
     contents = contents.strip(' \t\r\n')
     contents = contents.replace('\r\n', '\n')
-    contents = contents.replace('\n<div', '<div')
     contents = contents.replace('\t', '&nbsp;' * 4)
     contents = contents.replace('  ', '&nbsp;' * 2)
     contents = contents.replace('\n\n', '\n&nbsp;\n')
--- a/main.py
+++ b/main.py
@@ -2429,7 +2429,7 @@ class SignatureHelpListener(sublime_plugin.ViewEventListener):
         formatted = []
 
         if len(self._signatures) > 1:
-            signature_navigation = "**{}** of **{}** overloads (use the arrow keys to navigate):\n".format(
+            signature_navigation = "**{}** of **{}** overloads (use the ↑ ↓ keys to navigate):\n".format(
                 str(self._active_signature + 1), str(len(self._signatures)))
             formatted.append(signature_navigation)
 
--- a/main.py
+++ b/main.py
@@ -61,6 +61,7 @@ popup_css = '''
     .lsp_popup p {
         margin-bottom: 0.5rem;
         padding: 0 0.5rem;
+        font-family: system;
     }
 '''
 
@@ -1085,6 +1086,7 @@ class LspSetupLanguageServerCommand(sublime_plugin.WindowCommand):
             css='''
                 .lsp_documentation {
                     margin: 1rem 1rem 0.5rem 1rem;
+                    font-family: system;
                 }
                 .lsp_documentation h1,
                 .lsp_documentation p {
--- a/main.py
+++ b/main.py
@@ -2024,7 +2024,7 @@ class HoverHandler(sublime_plugin.ViewEventListener):
             if language:
                 formatted.append("```{}\n{}\n```\n".format(language, value))
             else:
-                formatted.append("<div class='description'>{}</div>".format(value))
+                formatted.append("<p class='description'>{}</p>".format(value))
 
         mdpopups.show_popup(
             self.view,
--- a/main.py
+++ b/main.py
@@ -2023,36 +2023,23 @@ class HoverHandler(sublime_plugin.ViewEventListener):
                 language = item.get("language")
 
             if language:
-                formatted.append(
-                    mdpopups.md2html(
-                        sublime.active_window().active_view(),
-                        "```{}\n{}\n```\n".format(language, value)
-                    )
-                )
+                formatted.append("```{}\n{}\n```\n".format(language, value))
             else:
-                formatted.append(
-                    mdpopups.md2html(
-                        sublime.active_window().active_view(),
-                        "<div class='description' markdown='1'>{}</div>".format(preserve_whitespace(value))
-                    )
-                )
+                formatted.append(preserve_whitespace(value))
 
         mdpopups.show_popup(
             self.view,
-            "".join(formatted),
+            "\n".join(formatted),
             css='''
-                .lsp_hover .highlight {
-                    border-width: 0;
+                .lsp_hover {
+                    margin: 0.5rem 0.5rem 0 0.5rem;
                 }
-                .lsp_hover div.highlight,
-                .lsp_hover pre.highlight {
-                    margin-bottom: 0;
-                }
-                .lsp_hover .description {
-                    padding: 0.5rem 0.5rem 0 0.5rem;
+                .lsp_hover p {
+                    margin-bottom: 0.5rem;
+                    padding: 0 0.5rem;
                 }
             ''',
-            md=False,
+            md=True,
             flags=sublime.HIDE_ON_MOUSE_MOVE_AWAY,
             location=point,
             wrapper_class="lsp_hover",
@@ -2065,7 +2052,6 @@ def preserve_whitespace(contents: str) -> str:
     contents = contents.replace('\r\n', '\n')
     contents = contents.replace('\t', '\u00A0' * 4)
     contents = contents.replace('  ', '\u00A0' * 2)
-    contents = contents.replace('\n\n', '\n\u00A0\n')
     return contents
 
 
--- a/main.py
+++ b/main.py
@@ -2034,6 +2034,10 @@ class HoverHandler(sublime_plugin.ViewEventListener):
                 .lsp_hover {
                     margin: 0.5rem 0.5rem 0 0.5rem;
                 }
+                .lsp_hover .highlight {
+                    border-width: 0;
+                    border-radius: 0;
+                }
                 .lsp_hover p {
                     margin-bottom: 0.5rem;
                     padding: 0 0.5rem;
--- a/main.py
+++ b/main.py
@@ -2049,9 +2049,9 @@ def preserve_whitespace(contents: str) -> str:
     """Preserve empty lines and whitespace for markdown conversion."""
     contents = contents.strip(' \t\r\n')
     contents = contents.replace('\r\n', '\n')
-    contents = contents.replace('\t', '&nbsp;' * 4)
-    contents = contents.replace('  ', '&nbsp;' * 2)
-    contents = contents.replace('\n\n', '\n&nbsp;\n')
+    contents = contents.replace('\t', '\u00A0' * 4)
+    contents = contents.replace('  ', '\u00A0' * 2)
+    contents = contents.replace('\n\n', '\n\u00A0\n')
     return contents
 
 
--- a/main.py
+++ b/main.py
@@ -2021,24 +2021,38 @@ class HoverHandler(sublime_plugin.ViewEventListener):
             else:
                 value = item.get("value")
                 language = item.get("language")
+
             if language:
-                formatted.append("```{}\n{}\n```\n".format(language, value))
+                formatted.append(
+                    mdpopups.md2html(
+                        sublime.active_window().active_view(),
+                        "```{}\n{}\n```\n".format(language, value)
+                    )
+                )
             else:
-                formatted.append("<p class='description'>{}</p>".format(value))
+                formatted.append(
+                    mdpopups.md2html(
+                        sublime.active_window().active_view(),
+                        "<div class='description' markdown='1'>{}</div>".format(preserve_whitespace(value))
+                    )
+                )
 
         mdpopups.show_popup(
             self.view,
-            preserve_whitespace("".join(formatted)),
+            "".join(formatted),
             css='''
-                .mdpopups .lsp_hover .highlight {
-                   border-width: 0;
+                .lsp_hover .highlight {
+                    border-width: 0;
                 }
-                .mdpopups .lsp_hover .description {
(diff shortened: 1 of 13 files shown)


braver:
Ehm, I think this also needs to be applied to the signature popup. I could extract and share the styling between them.

rwols:
> Ehm, I think this also needs to be applied to the signature popup. I could extract and share the styling between them.

Agreed, but we're definitely going to have merge conflicts ;)

braver:
@rwols hurry up a bit with those PR’s then 😉

tomv564:
It looks a lot better!

I am not convinced this is a maintainable approach to getting good-looking hovers, hopefully it's okay if we take a more holistic look at mdpopups / minihtml limitations after the signature help changes are merged.

braver:
> I am not convinced this is a maintainable approach to getting good-looking hovers

Yeah, you kinda end up fighting against markdown, mdpopups and minihtml all at the same time. The html and css we need is actually very limited. I could mock one up that relies less on markdown and mdpopups' built-in templates?

braver:
This is as good as I can get it without any hacky attempts to deal with the blank line after the fenced code block. It's a cleaner look overall and without the nested margins and borders it takes hardly any css to look good. 

screen shot 2017-10-09 at 11 36 10

screen shot 2017-10-09 at 11 36 25


I read some more about what mdpopus provides and I definitely like how it handles things like indentation in `pre` blocks. I thought minihtml would render it as pre-formatted but apparently you have to do that on the plugin end. The more I learn about minihtml the more I appreciate a library taking on the heavy lifting if we can get away with it.

Basically the only thing I don't like about this setup right now is how a line break between block elements is rendered as a blank line, and that really sounds like a minihtml limitation that perhaps mdpopups can provide a solution for. It's not the worst thing in the world either, so perhaps it's better not to muck about with it.

deathaxe:
Your changes totally break python hover tooltips!

Result with your change (broken):

screenshot

Before (ok):

screenshot

You MUST decide whether to create your own html and disable markdown processing or let mdpopups do its job. You MUST NOT mix up markdown and html.

main.py
```diff
@@ -2040,9 +2048,10 @@ def show_hover(self, point, contents):
 def preserve_whitespace(contents: str) -> str:
     """Preserve empty lines and whitespace for markdown conversion."""
     contents = contents.strip(' \t\r\n')
-    contents = contents.replace('\t', '&nbsp;' * 4)
-    contents = contents.replace('  ', '&nbsp;' * 2)
-    contents = contents.replace('\n\n', '\n&nbsp;\n')
+    contents = contents.replace('\r\n', '\n')
```
deathaxe: Which language-server returns `\r\n`?

main.py
```diff
@@ -2022,14 +2022,22 @@ def show_hover(self, point, contents):
                 value = item.get("value")
                 language = item.get("language")
             if language:
-                formatted.append("```{}\n{}\n```".format(language, value))
+                formatted.append("```{}\n{}\n```\n".format(language, value))
             else:
-                formatted.append(value)
+                formatted.append("<p class='description'>{}</p>".format(value))
```
deathaxe: This line mixes html with markdown and breaks markdown processing.

main.py
rwols: Should be all of them, schorsing to protocol?

main.py
rwols: But markdown can have embedded HTML, right?

main.py
braver: jsts in my case at least

main.py
braver: That’s what I thought. Since when exactly can’t you have html in your markdown? Is that an mdpopups thing? 

Also, you don’t have to capitalize “must”, I can read just fine and it just makes you come across as angry and frustrated. If you are, please let me know what I did to have that effect. 

I’ll look into what happens with the python tool tips.

braver:
This is waiting for other PR’s, so please just consider it an investigation into what gives the best level of control over layout and readability vs. effort and maintainability. Currently it feels like servers don’t return clean markdown, mdpopus doesn’t convert lines

main.py
deathaxe: Not at all, frustrated. And yes it seems an mdpopups thing to not support embedded html this way as well as forcing `<br>` rather than using paragraphs.

A possible solution was to disable markdown processing `md=False`, create all lines as you did here and use `mdpopups.syntax_highlight()` to get highlighted html for source code embedding it into a `<div class="highlight">` manually. A little more effort was needed to do so, but it's solvable.

But I am concerned about all of this might grow quickly beyond the desired scope as it might create more and more per language edge cases needed to struggle with?

The only thing I am not too fine with is enforcing font-faces on a per package base.

braver: > Not at all, frustrated.

Pfieuw 😄 

> A little more effort was needed to do so, but it's solvable.

Yeah, but if the content is in markdown, wouldn't that mean we would also have to convert all that? You use mdpopups in GitGutter right? Is that a custom HTML?

> But I am concerned about all of this might grow quickly beyond the desired scope

It kinda feels like we can't get away with doing nothing (i.e. just push everything through mdpopups) and we definitely don't want do do a lot. I think we can go two ways here:

- get mdpopups / @facelessuser involved to tweak mdpopups based on our findings
- find a middle ground where we process content as little as possible and do as little styling as possible

> The only thing I am not too fine with is enforcing font-faces on a per package base.

Well, it's not a font-face but just a generic family, but I understand your point. I personally see no harm in using `sans-serif` here, but it's not my call.

rwols: If mdpopups ignores HTML, then that's still okay; it'll be consumed by Sublime's minihtml like we want. If mdpopups doesn't ignore HTML but processes stuff like `*hello*` inside of a HTML tag, then that's a problem. We *could* still work around that, if that so happens the case. @braver, are you willing to experiment wether mdpopups recognizes HTML, but ignores it, *or*, wether mdpopups doesn't recognize HTML and processes it as if it's Markdown?

braver: Well, it seems mdpopups will recognise `**bolded**` text (see "function" in one of my earlier screenshots) inside an html element, but it doesn't convert whitespace which messes up the python output.

Also seems like pyls doesn't wrap code snippets in any useful way. That would be another reason to stick with the monospace/editor font at least until servers are more mature and clean up their output.

facelessuser: What do you guys need?  You can mix markdown with HTML, but you have to do it right.   There is a TON of info going on here.  Do you guys have  a specific question?

braver: @facelessuser 

1.

```
<div>some content</div>
<div>some more content</div>
```

A blank line is rendered between the two divs. It's not padding or margin or can be styled to go away Is that intentional? Is that minihtml or mdpopups?

```
<div>some content</div><div>some more content</div>
```
Has the desired effect, but it's harder to achieve.

2.

I'd expect this markdown:

some text\n
\n
some more text

to be converted to
```
<p>some text</p>
<p>some more text</p>
```

But that doesn't seem to be the case.

deathaxe: The main question here is, how can mdpopups convert plain text docstrings which are said to be markdown by language-server-specs to proper looking html, even though each available server returns its own formatted text which is no proper markdown. 😆

deathaxe: But seriously, what @braver tried is to create real paragraphs for each returned description item. But somehow mdpopups doesn't create such, but inserts line breaks `<br>` only. Any way to create proper paragraphs?

deathaxe: Another question was about preserving white space in normal paragraphs to keep indention of ascii docstrings such as used by python in-tact.

facelessuser: Okay. To answer the first question, this is what mdpopups spits out.  What's the problem?

```py
>>> mdpopups.md2html(sublime.active_window().active_view(), 'some text\n\nsome more text\n')
'<p>some text</p><p>some more text</p>'
```

Is that not what you want?

deathaxe: Yes it is. For some reason I had the feeling it wouldn't be the case. After some research, the reason is the comments' lines not to be separated with double `\n`. On the other hand, who wants to decide, which lines to separate as paragraph and which one not?

facelessuser: What do you mean by preserve whitespace? Can you give me an example?

deathaxe: @braver: Paragraphs work with `preserve_whitespace("\n\n".join(formatted))` if we remove `contents = contents.replace('\n\n', '\n&nbsp;\n')` from `preserve_whitespace`.

But this won't solve any problems, because the CSS defines `.mdpopups p { margin: 0.1rem; }` which effectively removes any spacing between paragraphs.

braver: If there were a line break between those two paragraphs in your example output, am I correct to assume that would be rendered as a blank line by minihtml as if there was a <br>? That is, minihtml doesn’t eat the white space between two block elements?

facelessuser: I'll go ahead and answer some earlier speculation while I'm at it.   Python Markdown (which mdpopups uses) should process markdown in inline tags.  Block tags are excluded.  By default though, I include an extension that will process markdown in block tags if they have the attribute `markdown="1"`.  Usually you to do something like this (notice I have block the content with new lines, also notice that the opening tag is on its own line and the closing on its own):

```
<div markdown="1">

Some **markdown** content.

</div>
```

Markdown doesn't usually care about white space between paragraphs except to separate the paragraphs.  Spaces and tabs usually collapse as they would in any HTML.  Except where syntax specifically cares about whitespace.

Code tags are usually the only element where whitespace is generally perserved.

deathaxe: > What do you mean by preserve whitespace? Can you give me an example?

Some (python) - most are actually simple ASCII - docstrings use elements which are indented using several spaces (e.g. Google style docstrings, see my most recent screenshot). Passing those strings to markdown would squash down all the spaces to one. IMHO, this is default markdown behavior, but doesn't work well, when trying to handle ASCII as markdown.

facelessuser: So, if you want to protect HTML, put it in a block `div` and you are okay.  Nothing **should** get parsed.

But there is always a caveat. Python Markdown has a weakness in the way it processes fenced blocks.  I won't go into elaborate details, but fenced blocks are parsed as a pre processor, then the file is parsed for HTML blocks and sometimes a fenced block in an HTML block that is not meant to get parsed turns into soemthing like this:

before:

````
<div>

```
somecode
```
 
</div>

````

after:

````
<div>

w23134870280

</div>
````

The weird letters and numbers would be a place holder that never gets restored.

Basically fences and HTML are handled as separate  tasks when really you kind of need something more advanced that handles them simultaneously (or inline together).

Comment (facelessuser) [reply main.py]: I guess you could turn off fences and not have to deal with that.

Comment (facelessuser) [reply main.py]: I should also mention that setting that setting `md=False` when calling show popup will just pass through HTML without applying markdown.

Anyways, if you are hoping to have docstrings parsed as markdown-ish, you may have to process yourselves as markdown-ish may not work.  It's possible you could write your own extensions for Python Markdown and use those though....

Anyways, it sounds like you guys have a lot to figure out.  If you have more specific questions, I'd be happy to answer them.

Comment (rwols) [reply main.py]: Maybe we should handle this server-side?

Comment (braver) [reply main.py]: @rwols in what sense exactly? Getting the servers to standardize their output?

Comment (braver) [reply main.py]: Learned *so much* in the above. Adding the `markdown='1'` attribute allows us to wrap the content in something without losing the markdown processing. Also, minihtml isn't as weird as I feared it might be, but I *am* getting a <br> where I don't want one. 

I'll fiddle with it some more.

```
<p><div class="highlight"><pre>function&nbsp;scaleOrdinal&lt;string&gt;(range?:&nbsp;string[]):&nbsp;ScaleOrdinal&lt;string,&nbsp;string&gt;&nbsp;(+1&nbsp;overload)<br></pre></div><br /><div class='description' markdown='1'><strong>function</strong> <em>(exported, ambient)</em></div><div class='description' markdown='1'>Constructs a new ordinal scale with an empty domain and the specified range.<br />If a range is not specified, it defaults to the empty array; an ordinal scale always returns undefined until a non-empty range is defined.<br /><br />By default, the domain is configured to generate implicitly, if the scale is invoked with an unknown value.<br />See the "unknown(...)" method of the scale to change this behavior.<br /><br />The generic corresponds to the data type of range elements.</div></p>
```

**Comment by braver** [comment]
So, what I decided to do now is pass all contents through `mdpopups.md2html` and process them this way before joining the pieces together. In `show_popup` markdown is set to false so it doesn't try to format it again. This prevents issues with linebreaks between blocks which can results in extra `<br>` elements being added. And it seems to work fine with pyls now too.

I removed the `font-face` and did as little layout tweaking as possible. I'm happy with the results visually and the overall concept feels pretty reasonable to me.

The `preserve_whitespace` function is moved to pre-process descriptions before they're passed to mdpopups. In my test cases it seems to work fine without that function though, so I'm not sure what to think of it. I think it doesn't have to be used on code snippets because mdpopups already normalizes whitespace/indentation in `<pre>` elements. Perhaps someone else can shine some light on the necessity of `preserve_whitespace`.

I'm new to mypy, any help with the errors?

**Comment by deathaxe** [comment]
Indeed, looks well for most python docstrings. The extra empty line is actually appreciated with python as it separates "paragraphs" very clearly. There are some issues with signatures including line breaks in the argument list, but it appears for little functions only.

File: main.py [hunk 3]
```diff
@@ -2040,9 +2062,10 @@ def show_hover(self, point, contents):
 def preserve_whitespace(contents: str) -> str:
     """Preserve empty lines and whitespace for markdown conversion."""
     contents = contents.strip(' \t\r\n')
-    contents = contents.replace('\t', '&nbsp;' * 4)
-    contents = contents.replace('  ', '&nbsp;' * 2)
-    contents = contents.replace('\n\n', '\n&nbsp;\n')
+    contents = contents.replace('\r\n', '\n')
+    contents = contents.replace('\t', '\u00A0' * 4)
+    contents = contents.replace('  ', '\u00A0' * 2)
+    contents = contents.replace('\n\n', '\n\u00A0\n')
```
deathaxe: I am no longer sure about that line. I originally added it as I experienced absolutely no spaces between paragraphs. Indeed it breaks the creation of `<p>paragraph1</p><p>paragrah2</p>` The line causes markdown to create `<p>paragraph1<br><br>paragrah2</p>`. 

If we remove this line and tweak the CSS a little bit to add proper padding, we may not even need the `md2html()` functions as merging the lines into `div`s is effectively nothing else then creating several paragraphs. The solution would be far easier then anyone of us thought yesterday 😄  

You may try ...

```python
            ...
            if language:
                formatted.append("```{}\n{}\n```\n".format(language, value))
            else:
                formatted.append(preserve_whitespace(value))

        mdpopups.show_popup(
            self.view,
            "".join(formatted),
            css='''
                .lsp_hover {
                    margin: 0.5rem;
                }
                .lsp_hover .highlight {
                    border-width: 0;
                    margin-bottom: 0;
                }
                .mdpopups .lsp_hover p {
                    padding: 0.5rem 0.5rem 0 0.5rem;
                }
            ''',
            md=True,
            flags=sublime.HIDE_ON_MOUSE_MOVE_AWAY,
            location=point,
            wrapper_class="lsp_hover",
            max_width=800)


def preserve_whitespace(contents: str) -> str:
    """Preserve empty lines and whitespace for markdown conversion."""
    contents = contents.strip(' \t\r\n')
    contents = contents.replace('\r\n', '\n')
    contents = contents.replace('\t', '\u00A0' * 4)
    contents = contents.replace('  ', '\u00A0' * 2)
    return contents
```

braver: That's also pretty clean. However,  it looks a lot better if the description text aligns with the code text. Now it aligns with the code box:

screen shot 2017-10-10 at 20 00 37

Some screenshots to compare:

screen shot 2017-10-10 at 20 04 27

screen shot 2017-10-10 at 20 11 49

screen shot 2017-10-10 at 20 11 20

Also note in the first screenshot how there is whitespace missing after `function _(exported, ambient)_`. But disregarding that, I'm leaning towards the second or third style. 


----

As an aside, I don't think we have to prefix css selectors with `.mdpopups` (e.g. `.mdpopups .lsp_hover`). Our css follows after mdpopup's so there is no need to increase the specificity.

braver: As for sticking closely to mdpopups, this is in line with @deathaxe's suggestion, but with some fixes and even less lines of code:

```python
        for item in contents:
            value = ""
            language = None
            if isinstance(item, str):
                value = item
            else:
                value = item.get("value")
                language = item.get("language")

            if language:
                formatted.append("```{}\n{}\n```\n".format(language, value))
            else:
                formatted.append(preserve_whitespace(value))

        mdpopups.show_popup(
            self.view,
            "\n".join(formatted),
            css='''
                .lsp_hover {
                    margin: 0.5rem 0.5rem 0 0.5rem;
                }
                .lsp_hover p {
                    margin-bottom: 0.5rem;
                }
            ''',
            md=True,
            flags=sublime.HIDE_ON_MOUSE_MOVE_AWAY,
            location=point,
            wrapper_class="lsp_hover",
            max_width=800)


def preserve_whitespace(contents: str) -> str:
    """Preserve empty lines and whitespace for markdown conversion."""
    contents = contents.strip(' \t\r\n')
    contents = contents.replace('\r\n', '\n')
    contents = contents.replace('\t', '\u00A0' * 4)
    contents = contents.replace('  ', '\u00A0' * 2)
    return contents
```

screen shot 2017-10-10 at 20 38 35

screen shot 2017-10-10 at 20 39 00

I still think it's prettier to align the text, but I also very much like how little code this is.

To align text, it's now enough to add `padding: 0 0.5rem;` to `.lsp_hover p`. 

screen shot 2017-10-10 at 20 41 18

I also kinda like removing the border around the code, by adding `.lsp_hover .highlight { border-width: 0; border-radius: 0; }`

screen shot 2017-10-10 at 20 43 50

rwols: LGTM, you probably want to experiment around with the new signature help browser now, too.

deathaxe: > I don't think we have to prefix css selectors with .mdpopups

agree.

> However, it looks a lot better if the description text aligns with the code text.

Can be tweaked by setting the margin of `<p>`, i think. The highlight is a `<div>` and is therefore not effected. I just didn't see as pyls does return the signature as ordinary text unfortunately.

> there is whitespace missing after 

I saw such issues with signatures, too. Would be interesting how the raw text looks like, and if there is a way to fix it before passing it to mdpopups.

braver: Changed my mind about the border around the code box. I'll push an update and will look into some of the other popups later on. Gotta watch a football match right now, will get back to the code as soon as we're a couple of goals behind ;)

deathaxe: I see, it might be caused by applying `preserve_whitespace()` to each line separately. It strips tailing newlines, and therefore should be called with the joined lines.

main.py
rwols: >  in what sense exactly? Getting the servers to standardize their output?

Yes. Different languages have different traditions for documentation. e.g. Python has its own ascii style, C++ has Doxygen, Java has JavaDoc, etc.

braver:
Alrighty, that's the final push from my end, ready for feedback.

- merged with master and tested with the new overload browsing (which is very nifty by the way)
- removed the border around `.highlight` code box because it messes with alignment and doesn't add anything of value
- there is no need to have more than 1 wrapper-classname and more that 1 stylesheet, so all popups now use the same
- except for the documentation/setup popup, but I did clean that up to match overal whitespace etc. of the other popups
- I triple checked and`sans-serif` renders as Arial, which is standard for browsers but nobody's favourite font and will certainly not match anyone's UI theme. I decided to remove that from all popups, perhaps we can get a `--ui-font` variable for these situations in future ST versions
- I replaced the hint "use the arrow keys" with "use the ↑ ↓ keys", because it's a bit clearer, [utf8 icons](https://www.utf8icons.com) are cool, and also a little bit because I could

Nice clean, uniform popups all around. And in the end it wasn't even that hard. Why did it look so hard a week ago 🤷‍♂️ ?

---

I did notice this with pyls:

popup

Gif compression makes it hard to see, but the hover popup doesn't pick up that first line as a code block, but the signature popup does. I don't think that's new or related, but I thought it was a bit weird.

braver:
I asked around on Discord and JPS suggested using the "system" generic font family. This gives the system's UI font, not just Arial. I think it looks pretty good like that, but I'll leave it up to @tomv564 to decide if this should stay or not.

facelessuser:
>  And in the end it wasn't even that hard. Why did it look so hard a week ago :man_shrugging: ?

I find that is often the case.  Solutions are usually much smaller and simpler than I expect.  It is the journey to understand what the solution is that usually is hard :slightly_smiling_face: .

braver:
Yeah, this has been pretty typical. Try 500 things, 50% stupid hacks because “why doesn’t it work”, find out along the way how it actually works, throw everything away and rewrite it in a fraction of the code. I like to think of it as iteration, not just barely controlled chaos 😉 @facelessuser thanks for providing some much needed clarity.

deathaxe:
GOOD JOB! 😉 

The result looks pretty good and simple now. Using `system` font face looks good on a first glance, too.

One little suggestion:

The `.highlight` class defines a font-size of 0.9rem, which looks good in inline code blocks, but causes the signature of the first line to stand behind the description.

How about increasing the fenced code block's font size to 1.0rem or even 1.05rem?

```
    .lsp_popup div.highlight {
        font-size: 1.05rem;
    }
```

screenshot

braver:
> GOOD JOB!

Thanks, you did half of it though ;)

Changing a monospace font to 0.9em is typically done because they tend to "look" larger. If it needs 0.8 or 0.95 or whatever heavily dependent on the specific font (GitHub puts inline code at 0.85), but it's typical to tweak it a bit. I kinda don't like messing with it unless I'm sure what font it is though. 

I could put it back to 1em, but I don't know what reasoning @facelessuser put behind this. Unless you're adjusting the body font-family like we're doing there shouldn't be a reason to adjust highlighted inline code (it's all the same font anyway).

facelessuser:
@braver, I don't think people normally decrease mono because they "look" larger, at least that's not why I did it.  I do it because inline code provides a little padding on top and bottom `0.05`, so you can see the highlight better around the text.  So I decrease the font so it still takes up the same size as text on the same line.  If there was no padding on top and bottom, the font could remain the same height as everything else on the line.

Basically, I didn't want weird line spacing around a line that had inline code.  The line should take up as much height as a line without inline code.

facelessuser:
So yeah, block code could be increased to normal size, but inline (at least in mdpopups) would need to remain a little smaller.

braver:
> I don't think people normally decrease mono because they "look" larger

Hm, I must say I don't know why people do it generally, but they do look larger (x-height is usually quite a bit higher for on-screen readability), so if people are doing it for other reasons that also good. As long as they're doing it 😁 

Having some space for the background and highlighting is also good. 

So, is this something you'd do in the mdpopups default stylesheet? Move `font-size: 0.9rem` from the `.mdpopups .highlight` selector to `.mdpopups code.highlight`? Or should we do it here?

facelessuser:
I could look into it or even take a pull request.  As long as it doesn't botch up inline, I'm fine.  I always prefer plugins to only override what they *need* to in order to have consistent popups across plugins, and general global improvements to the basics to be made in mdpopups default CSS.  But that's my personal opinion.

If a plugin is going to make changes to global stuff, please, please, put a wrapper div class so users can target your changes and override if they like (and probably target your non-global, plugin specific changes as well).  The default CSS doesn't really provide a consistent UI if all the plugins are overriding the globals anyways, but I'm not going to police anyone :slightly_smiling_face:.

rwols:
> I really like how we're setting a repository record of almost 60 comments about a small +42/−23 PR about cosmetics.

Everybody has an opinion when it's about look&design 🎨

facelessuser:
Lives have been lost over cosmetics :slightly_smiling_face:.

braver:
> Lives have been lost over cosmetics

No animals were harmed in the creation of this repository though ;)

We don't even increase specificity over mdpopups here, or introduce new classnames other than the wrapper, so if anyone tweaks basic mdpopup styling these should just follow. Or they can target just the LSP popups super easily. I think we're good. As if anyone is going to do that anyway.

@facelessuser I think adding `font-size: 1rem;` at this point would do the trick:
https://github.com/facelessuser/sublime-markdown-popups/blob/717c21a56029e90207584bb07342082e282cf310/css/default.css#L201
Want a PR for that? It's probably going to affect a bunch of users of mdpopups, e.g. GitGutter?

facelessuser:
> It's probably going to affect a bunch of users of mdpopups, e.g. GitGutter?

it can't be helped some times, and this is honestly a pretty insignificant difference.  I'm sure some people will notice it, but many won't. If generally it is good positive improvement, I don't mine making it.

So yeah, a pull is fine.

tomv564:
Looking forward to trying this out, thanks to @facelessuser for the extra support for the awesome library!

deathaxe:
> e.g. GitGutter?

Wouldn't do any harm here at all. The diffpopup shows the .highlight div only besides the toolbar. 0.9 or 1.0 doesn't matter.

deathaxe:
> I really like how we're setting a repository record of almost 60 comments about a small +42/−23 PR about cosmetics.

The GUI is the first thing a user sees, responsible for the first impression. If it looks to weird or broken, everything else is considered sluggish as well. Therefore it's worth the effort, to reflect the intend of create something great.

braver:
Indeed, still it's pretty awesome to see this level of dedication for it. It's not every day I get a group of developers to join in a discussion about exactly which font at which size on the pixel level. 

Anyway, a [PR](https://github.com/facelessuser/sublime-markdown-popups/pull/39) has been sent @ mdpopups.

Also when working on that PR, I got highlighting in the popup (I guess the dependency is behind master?) and it looks pretty rad.

screen shot 2017-10-11 at 21 19 37

facelessuser:
Mdpopups behind?  I've seen Package Control not update dependencies some times, maybe that was it?  I don't know, it isn't a major feature behind.

Anyways, thanks for the PR!

facelessuser:
FYI, you should add the dependency `pymdownx` and `pyyaml`.  Currently mdpopups is using an internal copy of `pymdownx` plugins that will be going away in the future in favor of the dependency. And pyyaml is required for certain frontmatter features, though I am in a period where I am not forcing it.

braver:
> Mdpopups behind? 

I dunno, I just replaced the mdpopup in my Packages dir with my fork and all of a sudden I had highlighting. 🤷‍♂️

facelessuser:
It might be because recently I switched to using the Sublime highlighter by default, but that was I think back around June.  Either way, bonus!

braver:
Well, I removed my fork and let ST download the dependency again and highlighting is gone again... it was fun while it lasted 😞

rwols:
I've never seen highlighting in fenced code blocks personally, too. Maybe investigate this further @facelessuser ?

facelessuser:
Really?  What version of mdpopups is on your system:

```py
import mdpopups
mdpopups.version()
```

Maybe you guys picked up my secret version 😏 ...yeah, there's no secret version...or is there?

braver:
`(2, 1, 1)`

facelessuser:
Hmm, that's the latest...can you show me the fence source (raw markdown you use)?  Or do you render it directly via the some other method?

braver:
The attached file is what comes out of `self._build_popup_content()` for 
```
mdpopups.show_popup(self.view,
                                    self._build_popup_content(),
                                    css=popup_css,
                                    md=True,
                                    flags=sublime.HIDE_ON_MOUSE_MOVE_AWAY,
                                    location=point,
                                    wrapper_class="lsp_popup",
                                    max_width=800,
                                    on_hide=self._on_hide)
```

[print.txt](https://github.com/tomv564/LSP/files/1377062/print.txt)

facelessuser:
Yup it should get highlighted.  Let me know how mdpopups_test runs.  It's possible I accidentally broke and then fixed something without even knowing 😕 .

rwols:
Version 2.1.1 here as well, and running `mdpopups: test` looks like this:
schermafbeelding 2017-10-11 om 22 29 23

facelessuser:
Ruh roh. I done broke something.  I guess I need to cut a release.

braver:
Test looks like this for me:

```
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/mdpopup_test-master/test.py", line 83, in <lambda>
    window.show_quick_panel(tests, lambda v, fm=fmatter, md=md_file: run_test(v, fm, md))
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/mdpopup_test-master/test.py", line 80, in run_test
    getattr(this, 'mdpopups_%s_test' % test)(fm, md)
  File "/Users/koenlageveen/Library/Application Support/Sublime Text 3/Packages/mdpopup_test-master/test.py", line 126, in mdpopups_popup_format_test
    show_popup(mdpopups.format_frontmatter(fm) + sublime.load_resource(md))
  File "/Applications/Sublime Text.app/Contents/MacOS/sublime.py", line 192, in load_resource
    raise IOError("resource not found")
OSError: resource not found
```

Did I break it? 😢

facelessuser:
Weird...

facelessuser:
I'll probably merge the popup PR this evening and cut a release.  I'll make sure my version of mdpopups_test is pristine and up to date with what's on tip and ensure everything looks good.  You guys can then update the dependency and mdpopups_test, and if after that something is still broken, I'll dig deeper.

rwols:
Just wanted to note that both @braver and me are on Mac, maybe the problem is there? Just guessing.

facelessuser:
Nah, I'm a mac user too...just not at work.  I probably broke something and there aren't any plugins relying on the highlighting, so no one reported it to me.

facelessuser:
Do either of you have this in your preferences file set to False?  `"mdpopups.use_sublime_highlighter": true,`

braver:
Nope, didn't know that was a thing

facelessuser:
Just making sure.  If it's false it will use pygments. But even pygments should highlight it.  All of my machines seem to be running mdpopups tip, so I at least know that that mdopups_tests is working, so it looks like I just need to release mdpopups latest. Hopefully that will fix it.

braver:
It might be nothing, but I noticed with master that I also get a mdpopups output panel that I can select

screen shot 2017-10-11 at 23 15 16

It's not there when installed via "satisfy dependencies".

facelessuser:
Probably because it hasn't used the sublime highlighter to highlight code.  It actually puts the code snippet in a hidden output (I should disable it showing up in the menu though) and renders it with Sublime.  It then uses logic I ripped from my ExportHtml plugin to extract the real colors from the view and construct the HTML.  If it doesn't go down that path (which clearly it isn't for you before tip), it won't create the output panel.

facelessuser:
> BTW. what I did now was download a zip of master, put it in Packages instead of the original and copy the dependency-metadata.json to it from the original.

I point to my own channel which pulls the latest commit instead of tag.

facelessuser:
Alright.  Mdpopups 2.2.0 released.  Let me know if that fixes things.

braver:
Yes, it works now!

For Python at least. Also in GitGutter by the way, never had highlighting there before. What determines if it works for a language? I get no highlighting in TypeScript for something like this:

\`\`\`typescript
let y_scale: ScaleLinear(value: number | {
    valueOf(): number;
}) => number
\`\`\`

facelessuser:
It's kind of late here, so I'll answer and then go to bed 🙂 .

The Sublime Text highlighter is the default highlighter.  And I've created a mapping (https://github.com/facelessuser/sublime-markdown-popups/blob/master/st3/mdpopups/st_mapping.py) for most of the default syntaxes, but maybe not all, and definitely not for a lot of the 3rd party sytnaxes out there.

Feel free to add new ones if needed.  In general I try to match names to what Pygments uses so if a user switches between them, it will still work.  One day, I may drop Pygments for mdpopups, but I will most likely always use something like Pygments as a reference for a known sytnax convention for fences.

braver:
Nice! Colors everywhere!

screen shot 2017-10-12 at 08 59 42

facelessuser:
Well, my work here is done.  Don't forget to update dependencies:

```js
{
    "*": {
        ">=3124": [
            "pygments",
            "python-markdown",
            "mdpopups",
            "python-jinja2",
            "markupsafe",
            "pymdownx",
            "pyyaml"
        ]
    }
}
```

I don't have an exact date, but sometimes soonish I plan make `pymdownx` and `pyyaml` required and not optional.

braver:
Replaced by #176

PR #170, opened 2017-10-11
Split up main.py

tomv564:
Goals are:
- [x] make code for a given feature easy to find
- [x] introduce some layering to avoid dependency problems
- [x] preserve mypy checking and linting
- [x] avoid modules being loaded multiple times by sublime text?

Remaining:
- [x] move `plugin_loaded` and `plugin_unloaded` to `__init__.py` so AutoPackageReloader can work
- [x] extract the edit functionality from `main.py`

There were some calls about reorganising that might be worth another look in the future:
- Should we enforce a consistent way of creating popups by moving something into `core`?
- `document.py` is maybe not the most cohesive set of functions, could be split.

Other fixes:
- As part of fixing some startup issues I hopefully tackled the race conditions where document commands were sent before `didOpen` (see https://github.com/tomv564/LSP/pull/170/commits/02d6e48c47c35df3baf3803242c9e27a089e12c8)

Expected side effects:
- Will need to use AutoPackageReloader while developing
- I'm not sure how file history will look in master after this.

Apologies for the rebasing effort on existing pull requests. I may hop in to help fit existing work into the new structure if needed.

This PR will be squash-merged once the last fixes are done, I appreciate any bug reports!


deathaxe:
Makes reading and following the single functions much easier, cool.

plugin/core/main.py
```diff
@@ -0,0 +1,300 @@
...
+    unload_window_clients, unload_all_clients
+)
+from .events import Events
+from .documents import (
+    initialize_document_sync, notify_did_open
+)
+from .diagnostics import handle_diagnostics, remove_diagnostics
+
+
+def plugin_loaded():
```
deathaxe: `plugin_loaded()` and `plugin_unloaded()` should be part of the top-level __init__.py. Otherwise reloading the plugin with AutomaticPackageReloader doesn't work. These functions placed here are called only once upon ST startup.

plugin/core/main.py
```diff
@@ -0,0 +1,300 @@
...
+
+class LspStartClientCommand(sublime_plugin.TextCommand):
+    def is_enabled(self):
+        return is_supported_view(self.view)
+
+    def run(self, edit):
+        start_active_view()
+
+
+class LspApplyWorkspaceEditCommand(sublime_plugin.WindowCommand):
```
deathaxe: Would it make sense to move this and the following class to plugin/edit.py or plugin/apply_edit.py?

It looks like a normal command such as Symbols / Diagnostics / Code Actions ...

plugin/core/main.py
tomv564: Good to know, I will try this out and add instructions so people know to use it.

plugin/core/main.py
tomv564: Good point, might as well move it out of main!
 I'd like to have part of the edit functionality in the core package because it's much more a service to other features rather than something the user invokes directly.

boot.py
```diff
@@ -13,3 +15,11 @@
 from .plugin.code_actions import *
 from .plugin.symbols import *
 from .plugin.rename import *
+
+
+def plugin_loaded():
+    main.startup()
```
rwols: I find this to be too enterprisey, almost Java-esque. what can you possibly put here other than `main.startup`? Other than that, start to look good to me. Can't wait for this to get merged.

deathaxe:
> avoid modules being loaded multiple times by sublime text?

This should not be a problem with the new structure as all imports are placed within sub directories.

plugin/core/main.py
tomv564: I made the changes and reloading appears to work (hovers still work after reloading the package, at least). But can you explain why these hooks living in `main.py` would prevent reloading from working?

boot.py
tomv564: I'm open to other suggestions for both `boot.py` and `startup/shutdown`! I renamed to `boot.py` as calling it `__init__.py` caused multiple instances of the packages to be started.

plugin/core/main.py
deathaxe: It's the way ST (re)loads its plugins. The reason is basically the same as for not reloading modules from sub directories.

It is designed to load top-level modules and look for `plugin_loaded` within their `__dict__` object. The way those modules are reloaded and the imports are handled then, causes this `__dict__` object to not contain `plugin_loaded`.
see: sublime_plugin.py L 101ff.

I guess it is related with the existing "issues" with reloading python modules. `__dict__` is an attribute, which are not or only partly updated by the `imp.reload()` or `importlib.import_module`.

But I am not too much of an python expert to give a real technical explanation for it.

PR #171, opened 2017-10-12
Update docs for clangd regarding compilation databases

rwols:
Any project of non-trivial size probably has include dirs and whatnot. Make users aware of how clangd handles this.


tomv564:
Awesome guide!
Do you also have instructions for `compd` ?
As an experienced user it's perhaps a bit much detail, what do you think of extracting the guide into a separate page? Then the people who are aware of what compilation databases are can just start configuring CMake/compdb and carry on.

rwols:
> what do you think of extracting the guide into a separate page? Then the people who are aware of what compilation databases are can just start configuring CMake/compdb and carry on.

Is it really too much text? Reading time is about 2 minutes... I'm not sure if it's worth the hassle to create a separate page for clangd.

tomv564:
Thanks for the update about `compdb`!

I had hoped that we could distill the setup into something short for the experts, but I get the feeling no such thing exists. I'll ask you some questions about clangd on gitter later.

If extracting a page is too much hassle I can take care of that for you :)

PR #173, opened 2017-10-13
Fix exception with signature help after project change

deathaxe:
This commit fixes an issue with LSP throwing an exception after opening another project within the self window using project switcher or ProjectManager plugin.

The traceback:
```
LSP: pyls not available for view None in window 3
  File "C:\Apps\Sublime\sublime_plugin.py", line 417, in run_async_view_listener_callback
    vel.__class__.__dict__[name](vel)
  File "C:\Apps\Sublime\Data\Packages\LSP\plugin\signature_help.py", line 60, in on_modified_async
    if self.signature_help_triggers:
AttributeError: 'SignatureHelpListener' object has no attribute 'signature_help_triggers'
```


rwols:
Thanks for fixing my broken PR 😋

PR #175, opened 2017-10-15
Blocking requests

rwols:
Commit 9e735fe67bed28e62bb0b5bd72b21c783f5e0641 adds the infrastructure for blocking requests.

Commit 2caff31e74819ff5943972df03a152a784d922dd enables blocking requests for `textDocument/formatting` and `textDocument/rangeFormatting`.

The `shutdown` request from https://github.com/tomv564/LSP/pull/157 should also be blocking.

This PR is the successor of https://github.com/tomv564/LSP/pull/73.


deathaxe:
`shutdown` in #157 is actually blocking already. It blocks for at most 5s to wait for the server to exit. If the server does not in this time, terminate() is called.

rwols:
You are totally right. And we should call terminate after the timeout. There is no support for doing something different on a timeout with this PR. I guess we can let `shutdown` be a special case.

rwols:
I noticed clangd already exits when receiving the `shutdown` request and doesn't send anything back. That's a bug in clangd.

tomv564:
I would like to wait with this infrastructure until it is needed for `willSaveWaitUntil`.

To answer your suggestions from the previous pull request:

> shutdown: we must wait for the server's confirmation before killing the pipe.

In the new shutdown logic the client is removed from lookups before the shutdown request is made - no blocking needed to avoid new requests.

> format doc and format range: there is no point in doing anything else while the server is formatting the code.

Blocking here would only help users trying to break things while formatting?

rwols:
> I would like to wait with this infrastructure until it is needed for willSaveWaitUntil.

Fair enough.

rwols:
Let us wait for a language server to implement willSaveWaitUntil.

PR #179, opened 2017-10-21
Implement textDocument/onTypeFormatting

rwols:
- Create ViewEventListener
- Add onTypeFormatting classmethod to Request
- Add "format_on_type" setting
- Refactor options handling in formatting requests

### DIFF
--- a/plugin/core/settings.py
+++ b/plugin/core/settings.py
@@ -63,6 +63,7 @@ def update_settings(settings: Settings, settings_obj: sublime.Settings):
     settings.completion_hint_type = read_str_setting(settings_obj, "completion_hint_type", "auto")
     settings.complete_using_text_edit = read_bool_setting(settings_obj, "complete_using_text_edit", False)
     settings.resolve_completion_for_snippets = read_bool_setting(settings_obj, "resolve_completion_for_snippets", False)
+    settings.format_on_type = read_bool_setting(settings_obj, "format_on_type", False)
     settings.log_debug = read_bool_setting(settings_obj, "log_debug", False)
     settings.log_server = read_bool_setting(settings_obj, "log_server", True)
     settings.log_stderr = read_bool_setting(settings_obj, "log_stderr", False)
--- a/plugin/formatting.py
+++ b/plugin/formatting.py
@@ -92,6 +92,7 @@ class DocumentOnTypeFormattingListener(sublime_plugin.ViewEventListener):
         self._client = None  # type: Optional[Client]
         self._initialized = False
         self._enabled = False
+        self._undo_was_invoked = False
 
     def on_modified_async(self) -> None:
         if not settings.format_on_type:
@@ -101,6 +102,9 @@ class DocumentOnTypeFormattingListener(sublime_plugin.ViewEventListener):
         if self._enabled:
             client = client_for_view(self.view)
             if client:
+                if self._undo_was_invoked:
+                    self._undo_was_invoked = False
+                    return
                 for region in self.view.sel():
                     self._on_type_formatting_async(client, region.begin())
 
@@ -129,6 +133,10 @@ class DocumentOnTypeFormattingListener(sublime_plugin.ViewEventListener):
         client.send_request(Request.onTypeFormatting(params),
                             lambda response: self.view.run_command("lsp_apply_document_edit",
                                                                    {"changes": response}))
+
+    def on_text_command(self, name, _) -> None:
+        if name == "undo":
+            self._undo_was_invoked = True
 
 
 def get_formatting_options(view: sublime.View) -> dict:
--- a/plugin/formatting.py
+++ b/plugin/formatting.py
@@ -1,8 +1,24 @@
+import sublime
+import sublime_plugin
 
+from .core.configurations import is_supported_syntax
+from .core.documents import get_document_position
+from .core.events import global_events
 from .core.protocol import Request
+from .core.registry import LspTextCommand, client_for_view, session_for_view
+from .core.settings import settings
 from .core.url import filename_to_uri
-from .core.registry import client_for_view, LspTextCommand
 from .core.views import region_to_range
+
+# typing only
+from .core.rpc import Client
+assert Client
+
+try:
+    from typing import List, Optional
+    assert List and Optional
+except ImportError:
+    pass
 
 
 class LspFormatDocumentCommand(LspTextCommand):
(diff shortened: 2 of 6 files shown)


tomv564:
Wow, didn't even know about this feature.
Which language server are you testing this against? The Omnisharp c# language server has support, which I haven't tried yet.

rwols:
clangd too, it's very cool. It'll activate when the user types a closing brace `}`. Unfortunately Sublime auto-pairs braces, so it won't activate very often "by accident", the user has to be aware of this feature I guess.

rwols:
rebased to latest upstream master.

plugin/formatting.py
```diff
@@ -56,11 +67,74 @@ def run(self, _):
...
+    def __init__(self, view: sublime.View) -> None:
+        super().__init__(view)
+        self._triggers = []  # type: List[str]
+        self._client = None  # type: Optional[Client]
+        self._initialized = False
+
+    def on_modified_async(self) -> None:
+        if not settings.format_on_type:
+            return
+        if not self._ensure_initialized():
```
tomv564: For clarity I prefer what the CompletionHandler does:
```python
if not self._initialized:
    self._initialize()  # sets self._enabled

if self._enabled:
```

(in your implementation, `_ensure_initialized` means both "has been initialised" and "is enabled")

plugin/formatting.py
```diff
@@ -56,11 +67,74 @@ def run(self, _):
...
+
+    def _ensure_initialized(self) -> bool:
+        if self._initialized:
+            return bool(self._client)
+        self._initialized = True
+        self._client = client_for_view(self.view)
+        if self._client:
+            options = self._client.get_capability("documentOnTypeFormattingProvider")
+            if options:
+                self._triggers = [options.get("firstTriggerCharacter")] or []
```
tomv564: I may be misunderstanding, but if "firstTriggerCharacter" is absent the result would be:
```python
# Python 2.7.12 (default, Jun 29 2016, 14:05:02)
# [GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)] on darwin
# Type "help", "copyright", "credits" or "license" for more information.
>>> [None] or []
[None]
```
Following the spec, "firstTriggerCharacter" must be present. I'd prefer to disable the feature if it's missing.

plugin/formatting.py
```diff
@@ -56,11 +67,74 @@ def run(self, _):
...
+    def _on_type_formatting_async(self, point: int) -> None:
+        previous_character = self.view.substr(point - 1)
+        if previous_character not in self._triggers:
+            return
+        params = get_document_position(self.view, point)
+        if not params:
+            return
+        params["ch"] = previous_character
+        params["options"] = get_formatting_options(self.view)
+        purge_did_change(self.view.buffer_id())
```
tomv564: Typing `}}` here might cause a similar experience with repeated requests from typing `...` for the CompletionHandler. With completions we try to not block and discard old results, we might need something similar here.

plugin/core/settings.py
```diff
@@ -64,6 +65,7 @@ def update(self, settings_obj: sublime.Settings):
         self.complete_all_chars = read_bool_setting(settings_obj, "complete_all_chars", True)
         self.completion_hint_type = read_str_setting(settings_obj, "completion_hint_type", "auto")
         self.resolve_completion_for_snippets = read_bool_setting(settings_obj, "resolve_completion_for_snippets", False)
+        self.format_on_type = read_bool_setting(settings_obj, "format_on_type", True)
```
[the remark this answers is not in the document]
rwols: done

plugin/formatting.py
```diff
@@ -56,11 +67,74 @@ def run(self, _):
...
+        if not syntax:
+            return False
+        elif not is_supported_syntax(syntax):
+            return False
+        return True
+
+    def __init__(self, view: sublime.View) -> None:
+        super().__init__(view)
+        self._triggers = []  # type: List[str]
+        self._client = None  # type: Optional[Client]
```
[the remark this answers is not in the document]
rwols: done

plugin/formatting.py
rwols: done

plugin/formatting.py
rwols: done

plugin/formatting.py
rwols: This ViewEventListener uses `on_modified_async`, so nothing will be executed on Sublime's UI/main thread.

I'm not sure how to solve your concern at this point.

tomv564: We don't need to solve it immediately :) Thanks for the updates, I'll look at merging this weekend!

plugin/formatting.py
```diff
@@ -56,11 +67,73 @@ def run(self, _):
...
+        if previous_character not in self._triggers:
+            return
+        params = get_document_position(self.view, point)
+        if not params:
+            return
+        params["ch"] = previous_character
+        params["options"] = get_formatting_options(self.view)
+        purge_did_change(self.view.buffer_id())
+        # We ignore mypy here because it'll complain about an optional type, while that's been accounted for in
+        # _ensure_initialized.
```
tomv564: I believe the mypy ignore and this comment can be removed?

plugin/core/settings.py
```diff
@@ -47,6 +47,7 @@ def __init__(self):
         self.complete_all_chars = False
         self.completion_hint_type = "auto"
         self.resolve_completion_for_snippets = False
+        self.format_on_type = True
```
tomv564: Would like this value to default False as well

plugin/core/settings.py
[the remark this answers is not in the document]
tomv564: Please update the default value to False here.

tomv564:
Trying this out, I'm noticing that `undo` when the last character is a trigger character does not work (it undoes, notices the last trigger char being reinserted and re-applies the format?)
Does it do this for you as well when undoing the autoformat on  `}` blocks?

rwols:
Yes, I’ve noticed this too.

tomv564:
Question about both this PR and the blocking requests PR. 

I've been running SublimeJsPrettier which auto-formats on save, and I found that:
* As-you-type formatting is only a temporary "help" until prettier fixes up the entire document.
* The penalty of the extra didChange is not so noticeable (with the typescript language server)

Would adding either PR bring significant value to clangd users, given the auto-bracket-matching feature Sublime Text already offers itself?

rwols:
> Would adding either PR bring significant value

No, I'd say textDocument/onTypeFormatting is "nice to have". But like you said, other workflows may already achieve the desired goal. Still, I'd like to get this PR up to standards when I find the time.

rwols:
I have rebased this to a recent `master`. I have also disabled it by default. I'm still thinking of a way to handle the annoying behavior that occurs when undoing a trigger character with ctrl+Z.

tomv564:
Let's re-open the PR when it's ready.

PR #180, opened 2017-10-22
Fixes for tab completion and backspacing past completion start

tomv564:
* Cancel completion popup when backspacing past completion start

Then, when tab-completing (`insert_best_completion`) from a hidden completions list:
* If last completions are still relevant, return them immediately
* If no completions are available, undo tab insertion just before LSP completions are shown.

Credit to Typescript plugin and Jedi plugins for documenting behaviour.


tomv564:
This should fix #152.
A side effect is that inserted tabs after autocompleted insertions sometimes get undone.

PR #185, opened 2017-10-26
fix looking up document state in notify_did_save

dten:
was trying to lookup a file in the window ids map. pretty certain should be using this function instead

map is keyed on int as per https://github.com/dten/LSP/blob/0f8b8b0feb40878330953ea66a3b579caa7b1bcd/plugin/core/documents.py#L53

### DIFF
--- a/plugin/core/documents.py
+++ b/plugin/core/documents.py
@@ -160,8 +160,9 @@ def notify_did_close(view: sublime.View):
 
 def notify_did_save(view: sublime.View):
     file_name = view.file_name()
-    if file_name:
-        if file_name in document_states:
+    window = view.window()
+    if window and file_name:
+        if has_document_state(window, file_name):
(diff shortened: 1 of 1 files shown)


plugin/core/documents.py
```diff
@@ -161,7 +161,7 @@ def notify_did_close(view: sublime.View):
 def notify_did_save(view: sublime.View):
     file_name = view.file_name()
     if file_name:
-        if file_name in document_states:
+        if has_document_state(view.window(), file_name):
```
tomv564: `view.window()` can return None - best to store in a `window` variable and check `if file_name and window:`

dten: Oh didn't know it can be None.. wonder what causes that. I'll change it

tomv564:
Sorry for the extreme delay merging this!

PR #188, opened 2017-10-29
Use sortText if it's present in completion items

rwols:
This sorts the completions as defined by the language server (if there's a sortText property).

At least python-language-server and clangd provide this key.

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -241,6 +241,9 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
         if self.state == CompletionState.REQUESTING:
             items = response["items"] if isinstance(response,
                                                     dict) else response
+            if len(items) > 1 and "sortText" in items[0]:
+                # If the first item has a sortText key, assume all of them have a sortText key.
+                items = sorted(items, key=lambda item: item["sortText"])
             self.completions = list(self.format_completion(item) for item in items)
 
             if self.has_resolve_provider:


tomv564:
Implementation looks good!

If I understand correctly, sublime will preserve the given item order on the initial presentation of the list, but start re-shuffling entries using a fuzzy match algorithm as soon as a single character is typed (see https://forum.sublimetext.com/t/autocomplete-suggestions-are-not-in-expected-order/4728)

plugin/completion.py
```diff
@@ -241,6 +241,9 @@ def handle_response(self, response: dict):
         if self.state == CompletionState.REQUESTING:
             items = response["items"] if isinstance(response,
                                                     dict) else response
+            if len(items) > 0 and "sortText" in items[0]:
```
tomv564: Can be `len(items) > 1` :)

rwols:
> If I understand correctly, sublime will preserve the given item order on the initial presentation of the list, but start re-shuffling entries using a fuzzy match algorithm as soon as a single character is typed

Yes. There is much value in showing the most relevant items first in the list (most relevant according to the server). For example, clangd will show public member variables and methods of a class instance first, and way at the bottom will it show things like macro's, private methods and compiler-generated methods like destructors. Member variables and methods of base classes are somewhere in the middle of that list.

I don't know pyls prioritizes its completion items.

As soon as the user starts typing more, Sublime Text will take over and starts fuzzy matching.

plugin/completion.py
rwols: done.

PR #189, opened 2017-11-02
use auto-generated region.<color> scopes

braver:
In recent dev builds new auto-generated scopes have become available that provide a way to reliably get a color without having color schemes to support package-specific scopes. With SublimeLinter we're currently testing this new approach and it's and excellent way of getting red underlines etc. without color schemes having to add `sublimelinter.mark.error` etc. Here you can do away with throwing a bunch of scopes at it in hopes that color schemes will have a match for at least one of them. 

I don't believe you can do a version check in `sublime-syntax` files, so for the time being you can keep the markup.error/deleted etc. scopes in there. Perhaps the sublimelinter scopes should be kept for now too, no sure how effective they are. 

One thing to be aware of though is that for instance `region.redish` has both red fore- and background, although with reasonable contrast. So the diagnostics panel will look a bit different (but not badly so).

screen shot 2017-11-02 at 15 58 33

All in all I think adopting `region.<color>` scopes is a better approach than hoping color schemes will adopt specific scopes. Not sure how you feel about having version checks in the code and supporting experimental features, but I think it's a good way to get some exposure for the new approach and it indicated how we intend to move forwards with this.

### DIFF
--- a/Syntaxes/Diagnostics.sublime-syntax
+++ b/Syntaxes/Diagnostics.sublime-syntax
@@ -42,16 +42,15 @@ contexts:
 
   expect-diag-type:
     # See: https://github.com/sublimehq/Packages/issues/1036
-    # We use sublimelinter markup scopes too so that old color schemes can catch up.
     - include: pop-at-end
     - match: \berror\b
-      scope: markup.deleted.lsp sublimelinter.mark.error markup.error.lsp
+      scope: region.redish markup.deleted.lsp markup.error.lsp
       pop: true
     - match: \bwarning\b
-      scope: markup.changed.lsp sublimelinter.mark.warning markup.warning.lsp
+      scope: region.orangish markup.changed.lsp markup.warning.lsp
       pop: true
     - match: \binfo\b
-      scope: markup.inserted.lsp sublimelinter.gutter-mark markup.info.lsp
+      scope: region.bluish markup.inserted.lsp markup.info.lsp
       pop: true
 
   expect-linter-type:
--- a/plugin/diagnostics.py
+++ b/plugin/diagnostics.py
@@ -24,12 +24,20 @@ diagnostic_severity_names = {
     DiagnosticSeverity.Hint: "hint"
 }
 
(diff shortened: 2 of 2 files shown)


tomv564:
I think the version check is okay. Looks like build 3148 supported this, so `> 3147` should be enough?

The contrast is not great in the diagnostics panel, would be usable without that background color.

screen shot 2017-11-10 at 21 28 14

braver:
Contrast is definitely not good there. I know they were contemplating foreground only regions, this might be a good case. Let’s wait for that.

braver:
FYI, I'm working on a similar panel for SublimeLinter, learning some tricks too. I intend to update this PR later to make them match.

braver:
@tomv564 we ended up using the following in SublimeLinter:

Highlights:
```
warning:  region.yellowish markup.changed.sublime_linter markup.warning.sublime_linter
error:    region.redish    markup.deleted.sublime_linter markup.error.sublime_linter
```

Text:
```
warning:  markup.changed.sublime_linter markup.warning.sublime_linter
error:    markup.deleted.sublime_linter markup.error.sublime_linter
```

Perhaps some day we get `region.colorish` scopes for just the foreground, but until that happens it should work just fine without for most color schemes.

With SublimeLinter 4, we will no longer generate color schemes. If a color scheme is used that doesn't provide colors for the above scopes, we provide an option to change the scopes for highlights. For text (i.e. in the output panel), you're just out of luck. Current ST dev releases provide an easy way for users to extend color schemes, so that will eventually sort itself out. At some point there is only so much you can do for ancient and unmaintained color schemes, at least on the package end. I think that's a better approach than generating color schemes and messing with user space stuff like settings etc.

For LSP this means `sublimelinter.<something>` will stop doing anything very soon. You can leave it in because it doesn't hurt anyone, but consider it legacy and deprecated at this point.

tomv564:
Agree, nice work with moving on from the manipulated color schemes (which worked well for me for many years)!
Too bad the `colorish` option didn't work out, thanks for driving this proposal!

PR #196, opened 2017-11-15
fix: inserted spaces get undone

randy3k:
inserted spaces get undone as mentioned at https://github.com/tomv564/LSP/pull/180#issuecomment-338468746

remove these two lines because the bug #152 seems to be fixed even without
undoing the previous insertion.


randy3k:
Oh. I was wrong. These two lines are needed. So another fix is probably needed.

PR #197, opened 2017-11-15
fix: inserted space get undone

randy3k:
inserted spaces get undone as mentioned at https://github.com/tomv564/LSP/pull/180#issuecomment-338468746

this PR checks actively if the last text command is
insert_best_completion

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -109,6 +109,15 @@ class CompletionSnippetHandler(sublime_plugin.EventListener):
                 exception_log("Error inserting snippet: " + insertText, err)
 
 
+last_text_command = None
+
+
+class CompletionHelper(sublime_plugin.EventListener):
+    def on_text_command(self, view, command_name, args):
+        global last_text_command
+        last_text_command = command_name
+
+
 class CompletionHandler(sublime_plugin.ViewEventListener):
     def __init__(self, view):
         self.view = view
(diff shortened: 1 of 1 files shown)


plugin/completion.py
```diff
@@ -252,7 +261,10 @@ def handle_response(self, response: dict):
             # if insert_best_completion was just ran, undo it before presenting new completions.
             prev_char = self.view.substr(self.view.sel()[0].begin() - 1)
             if prev_char.isspace():
-                self.view.run_command("undo")
+                if last_text_command == "insert_best_completion":
```
tomv564: Do we need any other completion events in here ?

plugin/completion.py
```diff
@@ -252,7 +261,10 @@ def handle_response(self, response: dict):
...
+                    self.view.run_command("undo")
+                else:
+                    return
```
tomv564: This `else: return` has some unintended consequences which I missed before merging :D

randy3k: I put the return there to avoid things like this
screen shot 2017-11-17 at 12 23 30 am

It could be reproduced by typing `y` and `<space>` very quickly or when the language server responds slowly.

We might need to check if the context which triggered the completion is still a prefix of the current cursor.

tomv564: Thanks for explaining and sorry I stomped on the changes. I’ll try to integrate it into the existing stale-request handling logic

PR #198, opened 2017-11-15
add julia language server

randy3k:
only works on unix/mac though.

PC users may use (untested)

```
"command": ["julia", "--startup-file=no", "--history-file=no", "-e", "using LanguageServer; server = LanguageServer.LanguageServerInstance(STDIN, STDOUT, false); run(server);"]
```

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -123,6 +123,14 @@
       "syntaxes": ["Packages/Java/Java.sublime-syntax"],
       "languageId": "java"
     },
+    "julials":
+    {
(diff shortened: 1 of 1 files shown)


randy3k:
perhaps LSP should support platform-dependent settings along this direction

```
"command" : {
     "osx": "<osx command>",
     "windows": "<osx command>",
     "linux": "<linux command>"
}
```

tomv564:
Agree about platform-specific paths.

Is there some kind of convention where the `command` will be a correct default for most people (eg. can you install this language server to your PATH?)

randy3k:
Julia doesn't install any executables to PATH, so I don't think it will be useful to point to a file in PATH and ask users to copy the script file there.

PR #202, opened 2017-11-16
Prevent selection of CompletionItem hint if its label is empty

### LINKED ISSUE #201: Autocompletion hints fail if CompletionItem label is empty
This could happen when an user requests autocompletion hints for all the members and functions of some package like OpenCV (``cv2``), using python-language-server as LSP backend.

andfoy:
Fixes #201

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -230,7 +230,7 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
                 hint = completion_item_kind_names[kind]
         # label is an alternative for insertText if insertText not provided
         insert_text = item.get("insertText") or label
-        if insert_text[0] == '$':  # sublime needs leading '$' escaped.
+        if len(insert_text) > 0 and insert_text[0] == '$':  # sublime needs leading '$' escaped.
             insert_text = '\$' + insert_text[1:]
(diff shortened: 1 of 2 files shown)


plugin/completion.py
```diff
@@ -230,7 +230,7 @@ def format_completion(self, item: dict) -> 'Tuple[str, str]':
                 hint = completion_item_kind_names[kind]
         # label is an alternative for insertText if insertText not provided
         insert_text = item.get("insertText") or label
-        if insert_text[0] == '$':  # sublime needs leading '$' escaped.
+        if len(insert_text) > 0 and insert_text[0] == '$':  # sublime needs leading '$' escaped.
```
tomv564: It's tempting to omit these empty entries from the list (as the user gets no utility from them).
On the other hand perhaps users / developers should know that the language server is sending empty completion items.
What do you think?

andfoy: Well, I have this kind of question both as an user and also as an LSP client developer for Spyder. I've found too bothersome to display empty strings on the autocompletion menu. However, I think it's good to have a second verification at the client level, because sometimes this empty string suggestions come directly from the autocompletion libraries, e.g., Jedi or Rope on Python

PR #214, opened 2017-11-26
Send textDocument/didClose notification

rwols:
didClose never got sent.

This is because a closed view doesn't have a window ID anymore.

This patch walks through the relevant global dictionaries "by hand" and sends a didClose notification if applicable.

### DIFF
--- a/plugin/core/documents.py
+++ b/plugin/core/documents.py
@@ -14,8 +14,14 @@ from .protocol import Notification, Point
 from .settings import settings
 from .url import filename_to_uri
 from .configurations import config_for_scope, is_supported_view, is_supported_syntax, is_supportable_syntax
-from .clients import client_for_view, window_clients, check_window_unloaded
+from .clients import client_for_view, clients_by_window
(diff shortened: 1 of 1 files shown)


plugin/core/documents.py
```diff
@@ -14,9 +14,15 @@
 from .settings import settings
 from .url import filename_to_uri
 from .configurations import config_for_scope, is_supported_view, is_supported_syntax, is_supportable_syntax
-from .clients import client_for_view, window_clients, check_window_unloaded
+from .clients import client_for_view, clients_by_window
```
rwols: mypy complains about clients_by_window, not sure how to fix that?

tomv564:
Thanks for noticing, I just submitted the 'ugly but works' fix in f187449.
I don't have time to review a rewrite of the entire didClose mechanism, but I'll happily look at a more minimal set of changes.

rwols:
Thanks! I think your one liner will work fine in most practical cases.

PR #222, opened 2017-12-03
is_in_workspace to check for member in all project folders

stevenkaras:
This should allow diagnostics to be run for files that are not in the primary
project folder. I have not made the effort to adapt other parts of the plugin
that should take into account multiple project folders


tomv564:
The `is_in_workspace` check has now been removed.
I welcome future contributions to follow VS Code's multi-folder support!

PR #227, opened 2017-12-10
Fixed incorrect retrieval of 'error' state

vogelsgesang:
My last pull request contained a mistake 🙁 
Sorry for that...


tomv564:
Ah, typical thing a review should catch! There should be tests on that code, too :)

PR #228, opened 2017-12-10
Treat a capability announced as `False` as unsupported

vogelsgesang:
So far, the presence of a capability's key in ServerCapabilities was treated as if the server supported that capability.
E.g., `{"codeActionProvider": false}` was interpreted as if the server supported code actions.

While not clearly stated in the LSP spec, the intent of such a `false` capability is rather to explicitly state that the server does *not* support the capability.

I was running into this while experimenting with my own dummy language server implementation

### DIFF
--- a/plugin/core/rpc.py
+++ b/plugin/core/rpc.py
@@ -43,7 +43,7 @@ class Client(object):
         return self.project_path
 
     def has_capability(self, capability):
-        return capability in self.capabilities
+        return capability in self.capabilities and self.capabilities[capability] is not False
 
     def get_capability(self, capability):
         return self.capabilities.get(capability)


plugin/core/rpc.py
```diff
@@ -43,7 +43,7 @@ def get_project_path(self):
         return self.project_path
 
     def has_capability(self, capability):
-        return capability in self.capabilities
+        return capability in self.capabilities and self.capabilities[capability] is not False
```
rwols: Perhaps is `return self.capabilities.get(capability, False)` is slightly less verbose?

tomv564: @rwols that syntax reads better but eg `self.capabilities.get('completionProvider', False)` would return a dict with options instead of a boolean.

tomv564:
Thanks so much for the bug report and fix!

PR #229, opened 2017-12-13
Add polymer-ide to the default_clients list in settings.

rictic:
I tested it locally, the enable server command that then sets up configuration for the user is quite nice.

`npm i -g polymer-editor-service` is the prerequisite command if you'd like to test it out.

This HTML should produce an error diagnostic on the url:

```html
<link rel="import" href="./does-not-exist.html">
```

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -147,7 +147,7 @@
         "Packages/JavaScript/JavaScript.sublime-syntax",
         "Packages/JavaScript/JSON.sublime-syntax"
       ],
-      "languageId": "polymer",
+      "languageId": "javascript",
       "settings": {
         "polymer-ide": {
           "analyzeWholePackage": false,
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -139,7 +139,8 @@
         "text.html",
         "source.html",
         "source.js",
-        "source.css"
+        "source.css",
+        "source.json"
       ],
       "syntaxes": [
         "Packages/HTML/HTML.sublime-syntax",
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -130,6 +130,30 @@
       "scopes": ["source.julia"],
       "syntaxes": ["Packages/Julia/Julia.sublime-syntax"],
       "settings": {"runlinter": true}
+    },
+    "polymer-ide":
+    {
+      "command": ["polymer-editor-service"],
+      "scopes": [
+        "text.html.basic",
+        "text.html",
+        "source.html",
+        "source.js",
+        "source.css"
+      ],
+      "syntaxes": [
+        "Packages/HTML/HTML.sublime-syntax",
+        "Packages/CSS/CSS.sublime-syntax",
+        "Packages/JavaScript/JavaScript.sublime-syntax",
+        "Packages/JavaScript/JSON.sublime-syntax"
+      ],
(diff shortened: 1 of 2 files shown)


rwols:
You should add a small section in the docs beside the other language servers on how to get started.

LSP.sublime-settings
```diff
@@ -130,6 +130,30 @@
...
+        "text.html",
+        "source.html",
+        "source.js",
+        "source.css"
+      ],
+      "syntaxes": [
+        "Packages/HTML/HTML.sublime-syntax",
+        "Packages/CSS/CSS.sublime-syntax",
+        "Packages/JavaScript/JavaScript.sublime-syntax",
+        "Packages/JavaScript/JSON.sublime-syntax"
```
tomv564: Does the language server provide services in documents of all these types (even JSON?)?

LSP.sublime-settings
```diff
@@ -130,6 +130,30 @@
...
+      ],
+      "languageId": "polymer",
```
tomv564: I guess it's hard to pick a languageId that covers all the document types, but we use this value to select a syntax highlighter for Signature Help popups. If you eg. only have Signature Help for Javascript, putting `javascript` here might be a better choice.

rictic: Ah, interesting. We don't have signature help currently, but we could inside of Polymer databinding expressions, in which case we'd want javascript. Updated.

rictic: On a related note, looks like in the latest version of the protocol a client can declare support for markdown in signature help and completion items documentation.

LSP.sublime-settings
rictic: Is this something where syntaxes and scopes should be different? I'm not sure how they're used by LSP.

tomv564: Sorry, I thought I'd merged this long ago :blush:

The scope vs. syntax is mostly an implementation detail:

LSP uses two listener Apis: EventListener (global) and ViewEventListener (per file)

Where possible, we use ViewEventListeners for views of the right syntax. This way we don't slow down the UI or cause unintended behaviour for irrelevant files.

The EventListeners are global, so there we can check the scope of the code around the cursor. Example, if sublime detects javascript embedded within a document with HTML syntax, LSP can provide services for that.

tomv564:
A few months ago I added functionality to offer LSP functionality for an opened file if the file type was supported.

screen shot 2017-12-22 at 20 30 25

After merging this configuration, I get offered polymer-ide every time I open a JSON file. As it stands right now I'd have to instruct users to edit their config file to stop this.

I'm thinking it is time to stop adding default configurations, and instead use the SublimeLinter model of installing separate packages to enable LSP for a given language/environment. 

I'd like to resolve this JSON file distraction before releasing a new version soon. Do you have a proposal? Some discussions on what extra things can be done in a Sublime LSP-polymer package are discussed in https://github.com/tomv564/LSP/issues/212.

rictic:
Yeah, polymer-ide is a bit unusual in this regard. What about a regex that – if present on a default client – must match the first 4KB of the open file before LSP will suggest the language server? For Polymer something like the following:

```javascript
/link rel=['"]import['"]|Polymer\(|customElements\.define/
```

That will match an HTML import, a custom element definition, or an invocation of the Polymer function. Any of those three is a strong indication that the user would see benefit from polymer-ide.

PR #230, opened 2017-12-14
Implement textdocument/documentHighlight

rwols:
- Using a queue/purge model like in notifying document changes.
- Using "text" scope for unknown and text kinds, "markup" for read/write.
- Add the kind of document highlights to protocol.py
- Allow users to change the presentation style:
  * Users can choose underline, stippled or squiggly.
  * Users can change the scope per kind.
- Timeout of 500ms after no selection changes before sending out a document
  highlight request.

clangd recently added textdocument/documentHighlight 🎉

### DIFF
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -62,7 +62,6 @@ class DocumentHighlightKind(object):
     Text = 1
     Read = 2
     Write = 3
-    Count = 4
 
 
 class Request:
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -57,6 +57,14 @@ class CompletionItemKind(object):
     Reference = 18
 
 
+class DocumentHighlightKind(object):
+    Unknown = 0
+    Text = 1
+    Read = 2
+    Write = 3
+    Count = 4
+
+
 class Request:
     def __init__(self, method: str, params: 'Optional[dict]') -> None:
         self.method = method
@@ -110,6 +118,10 @@ class Request:
     @classmethod
     def documentSymbols(cls, params: dict):
         return Request("textDocument/documentSymbol", params)
+
+    @classmethod
+    def documentHighlight(cls, params: dict):
+        return Request("textDocument/documentHighlight", params)
 
     @classmethod
     def resolveCompletionItem(cls, params: dict):
--- a/plugin/highlights.py
+++ b/plugin/highlights.py
@@ -42,6 +42,12 @@ class DocumentHighlightListener(sublime_plugin.ViewEventListener):
             self._clear_regions()
             self._queue()
 
+    def _initialize(self) -> None:
+        self._initialized = True
+        client = client_for_view(self.view)
+        if client:
+            self._enabled = client.get_capability("documentHighlightProvider")
+
     def _queue(self) -> None:
         self._version += 1
         current_version = self._version
@@ -94,9 +100,3 @@ class DocumentHighlightListener(sublime_plugin.ViewEventListener):
                 scope = settings.document_highlight_scopes.get(kind_str, None)
                 self.view.add_regions("lsp_highlight_{}".format(kind_str),
                                       regions, scope=scope, flags=flags)
-
-    def _initialize(self) -> None:
-        self._initialized = True
-        client = client_for_view(self.view)
-        if client:
-            self._enabled = client.get_capability("documentHighlightProvider")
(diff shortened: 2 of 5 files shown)


tomv564:
Super cool, I've been wanting to see how this could work in Sublime Text.
Any chance you could add a screenshot / gif of this in action?

rwols:
documenthighlight

tomv564:
Thanks for the GIF, looks great!
Especially in your c++ example there is a lot colouring / underlining going on. I'm suspect the read/write thing will be lost on some users (they might mistake it for errors depending on what color the scheme renders).
I'm also curious how this cooperates with eg. diagnostics highlighting.
Going to try this soon, exciting!

tomv564:
I added support for this in my typescript server (https://github.com/tomv564/lsp-tsserver/commit/3486ed4070a1947f25c087979df8d5dffc75a7b7), here it is in action:

Cursor on "text":

screen shot 2017-12-16 at 13 43 26

Combined with diagnostics on edited "text"

screen shot 2017-12-16 at 13 43 44

As noisy as I could get it:

screen shot 2017-12-16 at 13 44 22

plugin/highlights.py
```diff
@@ -0,0 +1,102 @@
...
+        current_version = self._version
+        sublime.set_timeout_async(lambda: self._purge(current_version), 500)
+
+    def _purge(self, this_version: int) -> None:
+        if this_version == self._version:
+            self._on_document_highlight()
+
+    def _clear_regions(self) -> None:
+        for kind in settings.document_highlight_scopes.keys():
+            self.view.erase_regions("LspDocumentHighlightListener_{}".format(kind))
```
tomv564: In edit.py we use `lsp_edit`, could we use `lsp_highlight_{kind}` here?

plugin/highlights.py
```diff
@@ -0,0 +1,102 @@
...
+        elif settings.document_highlight_style == "squiggly":
+            flags |= sublime.DRAW_SQUIGGLY_UNDERLINE
+        self._clear_regions()
+        for kind_str, regions in kind2regions.items():
+            if regions:
+                scope = settings.document_highlight_scopes.get(kind_str, None)
+                self.view.add_regions("LspDocumentHighlightListener_{}".format(kind_str),
+                                      regions, scope=scope, flags=flags)
+
+    def _initialize(self) -> None:
```
tomv564: Prefer this method higher up after `on_selection_modified_async`

plugin/highlights.py
```diff
@@ -0,0 +1,102 @@
...
+                client.send_request(request, self._handle_response)
+
+    def _handle_response(self, response: list) -> None:
+        if not response:
+            return
+        kind2regions = {}  # type: Dict[str, List[sublime.Region]]
+        for kind in range(0, DocumentHighlightKind.Count):
+            kind2regions[_kind2name[kind]] = []
+        for highlight in response:
+            if highlight:
```
tomv564: Is this check needed?

plugin/core/protocol.py
```diff
@@ -57,6 +57,14 @@ class CompletionItemKind(object):
     Reference = 18
 
 
+class DocumentHighlightKind(object):
+    Unknown = 0
+    Text = 1
+    Read = 2
+    Write = 3
+    Count = 4
```
tomv564: I had to double check, but `Count` is a clever definition not in the spec?
Prefer using an Enum here if possible and `len(DocumentHighlightKind)` for count.

plugin/highlights.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def on_selection_modified_async(self) -> None:
+        if not self._initialized:
+            self._initialize()
+        if self._enabled:
+            self._clear_regions()
+            self._queue()
+
+    def _queue(self) -> None:
+        self._version += 1
```
tomv564: I'm sure it would take a lot of editing to hit the max bounds of `_version`, but I wonder if storing and checking the position of the cursor(s) could work equally well?

plugin/highlights.py
rwols: I think strictly no. I've removed the check in cc3f1a7.

plugin/core/protocol.py
rwols: We can't use enums? we're on Python 3.3. Not entirely sure what you mean here. I've removed the `Count` class member and replaced it with a `4` in the `x in range(0, 4)`.

tomv564:
Thanks for the fixes, great feature worthy of a release in the near future!

PR #234, opened 2017-12-29
Fix: _is_visible property of SignatureHelpListener is unreliable

rwols:
I noticed that `self._is_visible == False` after pressing <kbd>space</kbd> when the signature help popup is visible. Thus, pressing the down key results in the cursor moving down, instead of browsing through function overloads. I'm not entirely sure what causes this, but apparently `self._is_visible` is set to `True` after the call to `mdpopups.show_popup`, however this gets undone by the `self._on_hide` handler somehow.

I've decided to replace this construct with self.view.is_popup_visible. The downside is that this method
also return `True` when the signature help popup is hidden, but the on-hover
popup is visible. Thus, we try to make sure that the signature help popup is
"truly" visible/hidden when the _signatures list is empty.

### DIFF
--- a/plugin/signature_help.py
+++ b/plugin/signature_help.py
@@ -71,8 +71,11 @@ class SignatureHelpListener(sublime_plugin.ViewEventListener):
                             lambda response: self.handle_response(response, pos))
             else:
                 # TODO: this hides too soon.
-                if self._visible:
+                if len(self._signatures) > 0:
                     self.view.hide_popup()
+                    # This makes sure that we don't trigger accidentally in on_query_context while another type of
+                    # popup is visible (e.g. one triggered by a mouse hover).
+                    self._signatures = []
 
     def handle_response(self, response, point):
         if response is not None:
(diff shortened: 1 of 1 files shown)


tomv564:
Pressing space also does a full query for the same signature help data. Looking at the logs, the events are like this:
'(' -> get signature help -> show_popup (1)
' ' -> get signature help -> show_popup (2) -> on_hide (1)

I'm going to look at how https://github.com/tomv564/LSP/tree/keep-signature-help-open behaves under these conditions.

rwols:
I tried out that `keep-signature-help-open` branch, and it is *amazing*... when it works! For instance, it won't work on C++ methods if we don't also add `meta.method-call`. See what I mean with the uncertainty of the scopes? I think in practice it can definitely be mostly reliable, but updating the scopes that should match the selector will be a maintenance job. I agree that work should be done first to enhance the sig help behavior before moving to highlighting active params.

tomv564:
The triggering logic was creating some of our own problems, I took the liberty to fix both issues mentioned in the commit https://github.com/tomv564/LSP/commit/43c9d6eafaab898b8e624575bf6663f13abf93dc.

I'd like to merge the parameter highlighting PR next, would be great if you're interested in rebasing your branch (and giving the triggers a sanity check), otherwise I can help you with it soon!

rwols:
> I'd like to merge the parameter highlighting PR next, would be great if you're interested in rebasing your branch (and giving the triggers a sanity check), otherwise I can help you with it soon!

Gonna rebase this evening!

PR #235, opened 2017-12-29
Add "vscode" style signature help popup

rwols:
- Can switch between "sublime" style and "vscode" style in the LSP preferences.
- Tried to keep both style as similar as possible. I think only the top label is now different.

Related: https://github.com/tomv564/LSP/issues/231

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -180,6 +180,13 @@
   // Valid values are "underline" or "box"
   "diagnostics_highlight_style": "underline",
 
+  // Signature help style.
+  // Valid values are "sublime" or "vscode".
+  //
+  // - "sublime": Uses mdpopups faithfully.
+  // - "vscode": Tries to imitate VSCode.
+  "signature_help_style": "vscode",
+
   // Highlighting style of "highlights": accentuating nearby text entities that
   // are related to the one under your cursor.
   // Valid values are "underline", "stippled" or "squiggly".
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -180,12 +180,9 @@
   // Valid values are "underline" or "box"
   "diagnostics_highlight_style": "underline",
 
-  // Signature help style.
-  // Valid values are "sublime" or "vscode".
-  //
-  // - "sublime": Uses mdpopups faithfully.
-  // - "vscode": Tries to imitate VSCode.
-  "signature_help_style": "vscode",
+  // If true, highlight the active parameter of the currently active signature.
+  // Note that this disables mdpopups highlighting.
+  "highlight_active_signature_parameter": true,
 
   // Highlighting style of "highlights": accentuating nearby text entities that
   // are related to the one under your cursor.
--- a/plugin/signature_help.py
+++ b/plugin/signature_help.py
@@ -223,12 +223,12 @@ class SignatureHelpListener(sublime_plugin.ViewEventListener):
 
     def _replace_active_parameter(self, signature: str, parameter: str) -> str:
         if parameter[0].isalnum() and parameter[-1].isalnum():
-            pattern = r'\b{}\b'.format(parameter)
+            pattern = r'\b{}\b'.format(re.escape(parameter))
         else:
             # If the left or right boundary of the parameter string is not an alphanumeric character, the \b check will
             # never match. In this case, it's probably safe to assume the parameter string itself will be a good pattern
             # to search for.
-            pattern = parameter
+            pattern = re.escape(parameter)
         replacement = '<span style="font-weight: bold; text-decoration: underline">{}</span>'.format(parameter)
         # FIXME: This is somewhat language-specific to look for an opening parenthesis. Most languages use parentheses
         # for their parameter lists though.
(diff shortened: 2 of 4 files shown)


tomv564:
For maintenance's sake I would like to have only 1 signature help implementation. Especially if we somehow gain 90 lines of code!

rwols:
> For maintenance's sake I would like to have only 1 signature help implementation.

Are you sure? The only "branch point" really is just the `_build_popup_content`. I predict some people are loving the ST-style highlighting, others may prefer highlighting the active param.

tomv564:
Good point! I somehow imagined adding parameter highlighting to the existing signature would work, but I guess updating already formatted content is asking for trouble.

I think your way of building the signature help is a good first step.
We could try re-adding highlighting by passing the best-supported scope names into `view.style_for_scope`, afterwards!

rwols:
Just rebased, but I didn't test it thoroughly, although it looks to be working fine from first glance. Please test before merging.

LSP.sublime-settings
```diff
@@ -180,6 +180,13 @@
   // Valid values are "underline" or "box"
   "diagnostics_highlight_style": "underline",
 
+  // Signature help style.
+  // Valid values are "sublime" or "vscode".
+  //
+  // - "sublime": Uses mdpopups faithfully.
+  // - "vscode": Tries to imitate VSCode.
+  "signature_help_style": "vscode",
```
rwols: Do we want this to be `vscode` by default or `sublime` by default?

plugin/signature_help.py
```diff
@@ -126,28 +116,59 @@ def on_query_context(self, key, _, operand, __):
...
+
+    def _update_popup(self) -> None:
+        mdpopups.update_popup(self.view,
+                              self._build_popup_content(),
+                              css=popup_css,
+                              md=True,
+                              wrapper_class=popup_class)
+        self._visible = True
+
+    def _hide_popup(self) -> None:
```
tomv564: No longer used?

plugin/signature_help.py
```diff
@@ -126,28 +116,59 @@ def on_query_context(self, key, _, operand, __):
...
+                            on_navigate=self._on_hover_navigate)
+        self._visible = True
+
+    def _update_popup(self) -> None:
+        mdpopups.update_popup(self.view,
+                              self._build_popup_content(),
+                              css=popup_css,
+                              md=True,
+                              wrapper_class=popup_class)
+        self._visible = True
```
tomv564: Should not be needed, even though it was being reset to True in the old code

plugin/signature_help.py
```diff
@@ -126,28 +116,59 @@ def on_query_context(self, key, _, operand, __):
...
+            return self._build_popup_content_style_sublime()
+
     def _on_hide(self):
         self._visible = False
 
     def _on_hover_navigate(self, href):
         webbrowser.open_new_tab(href)
 
-    def _build_popup_content(self) -> str:
+    def _build_informational_header(self) -> str:
```
tomv564: Would prefer `build_overload_selector` or something more specific than "informational".

plugin/signature_help.py
```diff
@@ -167,3 +188,59 @@ def _build_popup_content(self) -> str:
...
+            # If the left or right boundary of the parameter string is not an alphanumeric character, the \b check will
+            # never match. In this case, it's probably safe to assume the parameter string itself will be a good pattern
+            # to search for.
+            pattern = parameter
+        replacement = '<span style="font-weight: bold; text-decoration: underline">{}</span>'.format(parameter)
+        # FIXME: This is somewhat language-specific to look for an opening parenthesis. Most languages use parentheses
+        # for their parameter lists though.
+        index = signature.find('(')
+        # Note that this works even when we don't find an opening parenthesis, because .find returns -1 in that case.
+        string = signature[index + 1:]
```
tomv564: Please rename `string` and preferably `index` to explain what you are doing here.

plugin/signature_help.py
```diff
@@ -167,3 +188,59 @@ def _build_popup_content(self) -> str:
...
+        return signature[:index + 1] + re.sub(pattern, replacement, string, 1)
```
tomv564: Curious what mypy thinks of the type of `pattern` here - if sets a regexp and else sets a string

plugin/signature_help.py
```diff
@@ -167,3 +188,59 @@ def _build_popup_content(self) -> str:
         if sigDocs:
             formatted.append(sigDocs)
         return "\n".join(formatted)
+
+    def _build_popup_content_style_vscode(self) -> str:
+        # Fetch all the relevant data.
+        signature = {}  # type: Dict[str, Any]
+        signature_label = ""
+        signature_documentation = ""
+        parameter = {}  # type: Dict[str, Any]
```
tomv564: `signature` and `parameter` are not used outside their respective if blocks below, no need to declare them here?

LSP.sublime-settings
tomv564: How about a boolean setting named something like `highlight_active_signature_parameter`?
This might be more helpful for users not familiar with VS Code.

We can merge it set to True and then decide before next release what works best.

rwols:
This last commit db0e5e8 is not part of your review but I noticed things like `int foo(float *bar);` would not highlight, because the `*` was seen as "match anything" while it should be a literal asterisk.

PR #242, opened 2018-01-16
cancel current completion if the previous input is an space

### LINKED ISSUE #211: space wrongly triggers autocompletion
following https://github.com/tomv564/LSP/pull/197#discussion_r151605795

screen shot 2017-11-17 at 12 23 30 am

It could be reproduced by typing `y` and `<space>` very quickly or when the language server responds slowly.

We might need to check if the context which triggered the completion is still a prefix of the current cursor.

randy3k:
Pre https://github.com/tomv564/LSP/issues/211#issuecomment-345973254
> We could also check if the last character is whitespace and set the completion state to CANCELLING in on_modified?

Cancel current completion If you receive a space while waiting for the completion reply.

fix #211

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -168,6 +168,10 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
         if self.view.sel()[0].begin() < self.last_location:
             self.last_location = 0
             self.view.run_command("hide_auto_complete")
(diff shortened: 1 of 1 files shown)


randy3k:
bump. @tomv564 how do you think about this PR?

tomv564:
Thanks for reminding me!

PR #261, opened 2018-01-26
Added support for ${sublime_path} and ${sublime_libs} variables,

evandrocoan:
join "env" value lists accordingly to OS path value delimiter then
each variable can be either a simple string or a list of strings.

Follows from https://github.com/tomv564/LSP/pull/133

### DIFF
--- a/plugin/core/main.py
+++ b/plugin/core/main.py
@@ -1,4 +1,5 @@
 import os
+import sys
 import subprocess
 
 try:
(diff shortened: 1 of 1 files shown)


tomv564:
As mentioned in the first line of the contributor guidelines:

>  LSP is a universal language server package. It is not the place for language-specific logic, features or workarounds.

evandrocoan:
You seem to be confusing LSP protocol and plugin for Sublime Text. If this was the [python-language-server](https://github.com/palantir/python-language-server) you are indeed correct. However this is not a language server, it is a client and is tight couple with Sublime Text and can integrate with it as much as it likes.

tomv564:
>   it is a client and is tight couple with Sublime Text and can integrate with it as much as it likes.

This is not a package that specialises in _developing Sublime Text packages_. There is already enough complexity just supporting LSP for every day use.

You are free to create a package that uses LSP to provide a better package authoring experience.

PR #267, opened 2018-01-30
Hide Context.sublime-menu when working on projects without language

### LINKED ISSUE #266: Hide Context.sublime-menu when working on projects without language servers
image

evandrocoan:
servers, fixes #266

### DIFF
--- a/plugin/core/clients.py
+++ b/plugin/core/clients.py
@@ -37,20 +37,18 @@ class ConfigState(object):
 
 
 class LspTextCommand(TextCommand):
-    def __init__(self, view, capability='', last_check=lambda: True):
+    def __init__(self, view, capability=''):
         super().__init__(view)
         self.capability = capability
-        self.last_check = last_check
 
     def is_visible(self):
         return is_supported_view(self.view)
 
-    def is_enabled(self):
+    def has_client_with_capability(self):
         client = client_for_view(self.view)
         if client and client.has_capability(self.capability):
-            return self.last_check()
+            return True
         return False
-
 
 def window_configs(window: sublime.Window) -> 'Dict[str, ConfigState]':
     if window.id() in clients_by_window:
--- a/plugin/core/clients.py
+++ b/plugin/core/clients.py
@@ -23,9 +23,8 @@ except ImportError:
 clients_by_window = {}  # type: Dict[int, Dict[str, Client]]
 
 
-class LspTextCommand(TextCommand):
-    # Passing capability='' due Sublime Text bug: https://github.com/SublimeTextIssues/Core/issues/2175
-    def __init__(self, view, capability='', last_check=lambda: True):
+class LspTextCommand(object):
+    def __init__(self, view, capability, last_check=lambda: True):
         super(LspTextCommand, self).__init__(view)
         self.capability = capability
         self.last_check = last_check
--- a/plugin/core/clients.py
+++ b/plugin/core/clients.py
@@ -1,7 +1,5 @@
 import sublime
 
-from sublime_plugin import TextCommand
-
 from .logging import debug, exception_log
 from .configurations import config_for_scope, is_supported_view
 from .protocol import Notification, Request
--- a/plugin/core/clients.py
+++ b/plugin/core/clients.py
@@ -1,7 +1,9 @@
 import sublime
 
+from sublime_plugin import TextCommand
+
 from .logging import debug, exception_log
-from .configurations import config_for_scope
+from .configurations import config_for_scope, is_supported_view
 from .protocol import Notification, Request
 from .workspace import get_project_path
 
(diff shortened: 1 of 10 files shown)


rwols:
Nice work @evandrocoan. I have a few comments about this. Although not immediately applicable to your current problem, I would also suggest to use a mixin class instead, to allow things like this:

```
class LspContextMenu:
    def is_visible(self):
        # etc ...

class FooCommand(sublime_plugin.TextCommand, LspContextMenu):
    def run(self, edit):
        # etc...

class BarCommand(sublime_plugin.WindowCommand, LspContextMenu):
    def run(self):
        # etc ...
```
Notice how *both* text commands as well as window commands can now use `LspContextMenu`. Furthermore, Sublime won't "instantiate" the LspContextMenu class, because it doesn't inherit from sublime_plugin.TextCommand. This idea is not new, because this is exactly [how GitSavvy does it](https://github.com/divmain/GitSavvy/blob/3c5b1ab49adc9b223d0a461d3e353a2e5cbfc64b/core/git_command.py#L48-L64).

Furthermore, because `LspContextMenu`'s purpose is *really* to allow "is_visible" functionality when the view is supported, I would suggest to rename the class to something that better communicates what it actually does, for instance `LspIsViewSupportedMixin`.

evandrocoan:
@rwols, if I do a mixing I need to do 2 constructors calls on all base classes instead of only 1. See this example from https://github.com/SublimeTextIssues/Core/issues/2175#issuecomment-362117507:
> ```python
> from sublime_plugin import TextCommand
> 
> class LspTextCommand(object):
>     def __init__(self, capability):
>         self.capability = capability
> 
> class LspFormatDocumentCommand(TextCommand, LspTextCommand):
>     def __init__(self, view):
>         TextCommand.__init__(view)
>         LspTextCommand__init__('other_stuff')
> 
> class LspSymbolDefinitionCommand(LspTextCommand):
>     def __init__(self, view):
>         TextCommand.__init__(view)
>         LspTextCommand.__init__('more_other_stuff')
> ```
> This forces on all derived classes to do 2 super constructors calls `TextCommand` and `LspTextCommand`. But inhering `TextCommand` on `LspTextCommand`, I can do everything just with one constructor call in all my derived classes:
> ```python
> from sublime_plugin import TextCommand
> 
> class LspTextCommand(TextCommand):
>     def __init__(self, view, capability):
>         super(LspTextCommand, self).__init__(view)
>         self.capability = capability
> 
> class LspFormatDocumentCommand(LspTextCommand):
>     def __init__(self, view):
>         super(LspFormatDocumentCommand, self).__init__(view, 'other_stuff')
> 
> class LspSymbolDefinitionCommand(LspTextCommand):
>     def __init__(self, view):
>         super(LspFormatDocumentCommand, self).__init__(view, 'more_other_stuff')
> ```

Unless I am missing something.

plugin/core/clients.py
```diff
@@ -23,9 +23,8 @@
 clients_by_window = {}  # type: Dict[int, Dict[str, Client]]
 
 
-class LspTextCommand(TextCommand):
-    # Passing capability='' due Sublime Text bug: https://github.com/SublimeTextIssues/Core/issues/2175
-    def __init__(self, view, capability='', last_check=lambda: True):
+class LspTextCommand(object):
+    def __init__(self, view, capability, last_check=lambda: True):
         super(LspTextCommand, self).__init__(view)
```
rwols: This cannot work?

plugin/core/clients.py
```diff
@@ -23,9 +23,8 @@
 clients_by_window = {}  # type: Dict[int, Dict[str, Client]]
 
 
-class LspTextCommand(TextCommand):
-    # Passing capability='' due Sublime Text bug: https://github.com/SublimeTextIssues/Core/issues/2175
-    def __init__(self, view, capability='', last_check=lambda: True):
+class LspTextCommand(object):
```
rwols: Would it not make more sense to rename this to something like `AutoDisableMixin` ?

evandrocoan: I deprecated the mixin because it is requires to write more code then usual, on the last commit https://github.com/evandroforks/LSP/commit/329038ab04d360302146c166036a9548ddcb5ceb you can see how much code I removed.

Also, would be required to add an exception for `mypy`
1. https://github.com/SublimeTextIssues/Core/issues/2066#issuecomment-362456473
1. https://travis-ci.org/tomv564/LSP/builds/336313027?utm_source=github_status&utm_medium=notification since it is failing with this:
```
mypy -p plugin
plugin/core/clients.py:26: error: Too many arguments for "__init__" of "object"
```

evandrocoan:
Implement it also for b6a3b9dca2cf04fd14fe6602d6cb1fb22ae51679 ?

1. b6a3b9dca2cf04fd14fe6602d6cb1fb22ae51679 Expose hover as command

rwols:
> if I do a mixing I need to do 2 constructors calls on all derivated classes instead of only 1.

Hmm yeah good point. I tried out your branch locally and it seems to work fine on first glance.

evandrocoan:
So far, all the commands which extends `LspTextCommand` require a server `capability`.

tomv564:
@evandrocoan agreed, but thanks for making the change. I'll clean up the diff and merge your changes shortly!

tomv564:
Sorry, I was unable to push to the branch on your fork (even though GitHub suggested it would work). Merged in https://github.com/tomv564/LSP/commit/67158b3f2b7c9b3009fdb6192a28332679a6593a, thanks for the UX and code improvements!

PR #279, opened 2018-02-18
Send what features we support to the server.

LoneBoco:
Some servers will listen to what features a client says it supports and will disable those features, returning "False" in the server capabilities reply.  Which, in turn, disables that feature in our code.

### DIFF
--- a/plugin/core/main.py
+++ b/plugin/core/main.py
@@ -212,7 +212,7 @@ def start_client(window: sublime.Window, config: ClientConfig):
     client.set_crash_handler(lambda: handle_server_crash(window, config))
 
     initializeParams = {
-        "processId": client.process.pid if client.process else None,
+        "processId": os.getpid(),
         "rootUri": filename_to_uri(project_path),
         "rootPath": project_path,
         "capabilities": {
--- a/plugin/core/main.py
+++ b/plugin/core/main.py
@@ -217,17 +217,42 @@ def start_client(window: sublime.Window, config: ClientConfig):
         "rootPath": project_path,
         "capabilities": {
             "textDocument": {
-                "completion": {
-                    "completionItem": {
-                        "snippetSupport": True
-                    }
-                },
                 "synchronization": {
                     "didSave": True
-                }
+                },
+                "hover": {
+                    "contentFormat": ["plaintext"]
+                },
+                "completion": {
+                    "completionItem": {
+                        "snippetSupport": True,
+                        "documentationFormat": ["plaintext"]
+                    },
+                    "completionItemKind": {
+                        "valueSet": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25]
+                    }
+                },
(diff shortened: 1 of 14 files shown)


LoneBoco:
The server in question where I encountered this issue was https://github.com/OmniSharp/omnisharp-roslyn

I'm not sure if sending everything in completionItemKind is correct or not.  I figured I may as well.

plugin/core/main.py
```diff
@@ -217,17 +217,80 @@ def start_client(window: sublime.Window, config: ClientConfig):
...
+                            CompletionItemKind.Reference
+                            ]
                     }
                 },
-                "synchronization": {
-                    "didSave": True
-                }
+                "signatureHelp": {
+                    "signatureInformation": {
+                        "documentationFormat": ["plaintext"]
```
rwols: I'm fairly certain you can set this to `markdown`.

LoneBoco: I've added markdown to the list.

rwols:
I’m eager to try this out tomorrow!

tomv564:
Whoops, I see the URI fix is also in here, could you keep that in the other PR?

LoneBoco:
@tomv564 I thought I reverted my accidental merge.  Is it still showing?

tomv564:
It's likely I only saw the old commit and didn't check the diff - sorry to bother you.
Thanks for the initiative, and good to hear omnisharp being used with LSP!

rwols:
@LoneBoco do you happen to work with macOS? I tried several times to install the OmniSharp LSP server but never got it to work.

LoneBoco:
@rwols I do not.  I use [omnisharp-roslyn](https://github.com/OmniSharp/omnisharp-roslyn) in Windows.  According to the README.md, you will need mono (>=5.2.0) to run it on OSX.  You should be able to clone it and run `build.sh` to have it download everything, build it, and run all the tests.

These are the settings I use with LSP:
```
{
    "clients":
    {
        "csharp":
        {
            "command": ["C:/Projects/omnisharp-roslyn/artifacts/scripts/OmniSharp.Stdio.cmd", "-lsp"],
            "scopes": ["source.cs"],
            "syntaxes": ["Packages/C#/C#.sublime-syntax"],
            "languageId": "csharp"
        }
    }
}
```

One thing to note is that omnisharp-roslyn has issues with the current master branch.  Some of them are partially fixed, but haven't been merged back yet.

I've [forked omnisharp-roslyn](https://github.com/LoneBoco/omnisharp-roslyn) and pushed some of the fixes into my repository, as well as added my own fixes for https://github.com/OmniSharp/omnisharp-roslyn/issues/1113, https://github.com/OmniSharp/omnisharp-roslyn/issues/1119, and https://github.com/OmniSharp/omnisharp-roslyn/issues/1125.  If you want the most problem free version of omnisharp-roslyn to mess with, I would recommend you clone from me for the time being.

Also, I would recommend you run with logging enabled.  I had originally tried to open a project and it didn't work.  The logs showed that msbuild was not able to process the solution as the project was using .NET 4.7 and I didn't have that version installed on my machine.  That might be a reason why you are having issues.

PR #280, opened 2018-02-18
The URI conversion will sometimes fail on file:// handlers.

LoneBoco:
Sometimes, a Windows pathname will be converted incorrectly.  This is a bandaid fix on the problem.

Previously, we would have a path that started like this: \C:\Users\...

When we tossed that path into Sublime, it would not be able to find the view associated to that file.  This bandaid fix would remove the preceding \ in that pathname.

This will mainly fix issues with diagnostics.

### DIFF
--- a/plugin/core/url.py
+++ b/plugin/core/url.py
@@ -2,6 +2,7 @@ from urllib.parse import urljoin
 from urllib.parse import urlparse
 from urllib.request import pathname2url
 from urllib.request import url2pathname
+import re
 
 
 def filename_to_uri(path: str) -> str:
@@ -9,4 +10,8 @@ def filename_to_uri(path: str) -> str:
 
 
 def uri_to_filename(uri: str) -> str:
-    return url2pathname(urlparse(uri).path)
+    prog = re.compile(r'\\[a-zA-Z]:\\.+')
+    ret = url2pathname(urlparse(uri).path)
+    if prog.match(ret):
+        ret = ret[1:]
+    return ret


tomv564:
Thanks for proposing a fix.
I need a little more information to be convinced to stop using python's built-in url translation logic.
`\C:\Users` doesn't look like a valid path, what URI did it come from?
Which language server are you using, and do you have a log / repro?

LoneBoco:
@tomv564 I am using omnisharp-roslyn for my language server.  Here is a log:

```
LSP:      {'diagnostics': [{'range': {'start': {'line': 3, 'character': 0}, 'end': {'line': 3, 'character': 37}}, 'code': 0, 'source': 'csharp', 'message': 'Unnecessary using directive.', 'severity': 4}, {'range': {'start': {'line': 3, 'character': 0}, 'end': {'line': 3, 'character': 37}}, 'code': 0, 'source': 'csharp', 'message': 'Unnecessary using directive.', 'severity': 4}], 'uri': 'file:///C%3A/Clients/clientname/file.cs'}
LSP: maybe_file_uri: file:///C%3A/Clients/clientname/file.cs
LSP: file_path: \C:\Clients\clientname\file.cs
```

I updated core\diagnostics.py to print out the above.
```
debug("maybe_file_uri: " + maybe_file_uri)
file_path = uri_to_filename(maybe_file_uri)
debug("file_path: " + file_path)
```

tomv564:
Encoding the `:` as `%3A` _should_ be unnecessary and is a bit unexpected as they are likely using standard Microsoft APIs to translate paths to URIs?
Any chance we can check on omnisharp's side if they can have a closer look at it? We can refer to https://en.wikipedia.org/wiki/File_URI_scheme#Windows or https://blogs.msdn.microsoft.com/ie/2006/12/06/file-uris-in-windows/

LoneBoco:
Looking into it further, it does appear that a properly formed URI should use the `:` instead of `%3A`.

```
>>> url2pathname('/C%3A/file.cs')
'\\C:\\file.cs'
>>> url2pathname('/C:/file.cs')
'C:\\file.cs'
```
So I believe this to be an omnisharp bug.  Whoops!

LoneBoco:
@tomv564 However, this may be an intentional thing because VSCode uses `%3A` encoding on uris.  The language server spec doesn't say anything about specially encoding the uris, but the main implementation does this encoding.  omnisharp-roslyn appears to specifically mangle the path to work with VSCode.  I've popped a question at the omnisharp-roslyn team about this, so maybe this will need to be something that will need to be handled by this extension.  If that is the case, I will make another attempt at this (probably by replacing `%3A` with `:` before throwing the uri into urllib)

PR #285, opened 2018-02-27
Feature to toggle diagnostic panel

### LINKED ISSUE #283: Feature request: Toggle diagnostic panel
Hi Tom, I have a suggestion.

Right now I have a keyboard shortcut to open the diagnostic panel `ctrl+shift+m`.
As you can see. When I open the panel with that keybinding, and press the keybinding again the panel is still visible. 

notoggle

Wouldn't it be better if the panel could be toggled like this? :)

output3

If the answer is yes? I have already done that and I could create a pull request if you want :)

predragnikolic:
Closes #283 

If open, the diagnostic panel will close. Else, it will open. :)


tomv564:
Thanks for adding symmetry to this operation! I forgot to mention, Escape can be used to dismiss any panel in Sublime Text.

PR #288, opened 2018-03-04
Allow "" for document_highlight_style in order to disable it entirely

rwols:
- Add "" option for document_highlight_style
- Add document_highlight_* settings to index.md docs

closes https://github.com/tomv564/LSP/issues/286

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -200,7 +200,8 @@
 
   // Highlighting style of "highlights": accentuating nearby text entities that
   // are related to the one under your cursor.
-  // Valid values are "underline", "stippled" or "squiggly".
+  // Valid values are "underline", "stippled", "squiggly" or "".
+  // When set to the empty string (""), no document highlighting is requested.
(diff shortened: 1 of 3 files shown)


tomv564:
Looks good!
Should we replace this later with a setting that disables the whole listener? i.e. `is_applicable` should return false if the user doesn't want document highlight.

rwols:
> Should we replace this later with a setting that disables the whole listener? i.e. is_applicable should return false if the user doesn't want document highlight.

Yeah I had that at first actually. But after testing it, the ViewEventListener stays inactive when enabling it through a setting. So that felt unintuitive. The `is_applicable` method only gets called when a property of the view is changed (so changing something from our home-brew global `settings` object doesn't invoke `is_applicable`).

tomv564:
> So that felt unintuitive.
Agree that it's nice when settings apply immediately, but a good portion of LSP settings already don't. 
But that's optimisation anyways, this is the fix we need now. Thanks!

Issue #186, opened 2017-10-27
Error panel pops up too fast


jakeparis:
I'm finding that the error pane is popping up as I'm in the middle of typing a phrase. Obviously while the line is unfinished, the syntax is not correct, that's just distracting. As such, I think it would be helpful to add some kind of delay before triggering the error popup panel. Even better would be to make the delay configurable.

rwols:
What language server is this?

jakeparis:
Javascript

tomv564:
Agree that the panel (and the squiggly lines) showing up mid-statement can be distracting! 
I usually leave the error panel open for that reason.

Would you want a delay on:
* The panel opening
* Diagnostics updating in the panel
* Diagnostics updating in the panel _and_ the document?

Perhaps you can try the same edits in VS Code so we can see where the diagnostics handling differs?

jakeparis:
I think the most _important_ delay would be on the panel opening, as that large movement within your field of vision is the biggest problem. I have my panel set to only open when there's a problem, and that's mainly to maximize editing real estate on my screen. 

I think it makes the most sense to have at least a slight delay on all of the things you mentioned. I don't have VSCode set up.

tomv564:
I see two options:
* We add the file's diagnostics count to the status area, you would open the panel manually when you see the diagnostics there (no need to decide what delay suits you)
* We wait to hear if other users are interested in this delay (as I'd like to avoid adding a setting for just one user)

braver:
Automatically opening the panel, regardless of delays, is super annoying IMO. I'd rather have a hint (e.g. a counter) that tells me something is in there and then I can open it when I want to, and close it when I'm done. So, that first option sounds great to me.

rmccue:
I quite like the SublimeLinter behaviour, which has [lint modes](https://sublimelinter.readthedocs.io/en/latest/lint_modes.html). The save mode is super useful, since it only shows you the errors when they're relevant (i.e. when you're going to run the code).

That said, background mode is quite useful. SublimeLinter by default has [a delay of 0.25s](https://github.com/SublimeLinter/SublimeLinter3/blob/master/SublimeLinter.sublime-settings#L4) before it shows the errors. In terms of how that actually works, it appears to be handled in SL's [queueing](https://github.com/SublimeLinter/SublimeLinter3/blob/master/lint/queue.py).

predragnikolic:
I like the idea of a counter. 
[...]

PR #292, opened 2018-03-13
Add ability to delay diagnostics

rmccue:
See #186. This adds a new `auto_show_delay` setting which can be used to debounce the `show_panel` command.

i.e. if you set it to `0.3`, it will wait 300ms after a diagnostics response before showing the panel. If another diagnostics response is received in the meantime, it will cancel the original timer and start a new one.

This allows you to avoid having the panel show up constantly while typing code. This doesn't affect panel updates, which are still instant (after receiving the diagnostics response).

For convenience, I've simply copied across the [SublimeLinter queue functionality](https://github.com/SublimeLinter/SublimeLinter/blob/8ca2a23ba4f2c6da814904dbc60814b7b576c38e/lint/queue.py) and credited it in the license file. If this is undesirable, I can remove it and write it from scratch instead.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -19,3 +19,30 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
+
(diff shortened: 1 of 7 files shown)


tomv564:
Thanks for your initiative!
I would like to evaluating showing diagnostics updates in a less disruptive way in the  status bar - okay if we pursue that first instead?

rmccue:
Manage the project however you like, I just found this useful personally. :)

I’d actually rather have a show-on-save feature, but didn’t have time to implement that as well.

tomv564:
Created issue #381 for your show-on-save feature - that idea sounds more deterministic to implement!

PR #295, opened 2018-03-20
Try to utilize LS completion textEdit if possible

### LINKED ISSUE #294: Add support for CompletionItem.textEdit (& additionalTextEdits)
* Windows 10 / macOS 10.13.3
* Latest LSP from Package Control
* EditorServiceHost (PowerShell) language server accoring to recipe at: https://github.com/PowerShell/PowerShellEditorServices/pull/629#issuecomment-373787797

LSP currently supports completion with `CompletionItem.insertText` only, while it may be good enough for some servers e.g. [python-language-server](https://github.com/palantir/python-language-server), for some cases like EditorServiceHost this is simply not enough as `insertText` requires some extra logic on editor side.

Concrete example for EditorServiceHost
1. `$tru`<kbd>Ctrl</kbd><kbd>Space</kbd> gets completed to `$$true` (double `$$`) instead of `$true`

    Payload:
    ~~~
    [
    {'filterText': '$true', 'insertText': '$true', 'sortText': '0001true', 'label': 'true', 'additionalTextEdits': None, 'detail': None, 'data': None, 'kind': 6, 'command': None,
     'textEdit': {'newText': '$true', 'range': {'end': {'line': 5, 'character': 2}, 'start': {'line': 5, 'character': 0}}}, 'commitCharacters': None, 'range': None, 'documentation': None}, {'filterText': '$tokens', 'insertText': '$tokens', 'sortText': '0002tokens', 'label': 'tokens', 'additionalTextEdits': None, 'detail': '[Token[]]', 'data': None, 'kind': 6, 'command': None, 'textEdit': {'newText': '$tokens', 'range': {'end': {'line': 5, 'character': 2}, 'start': {'line': 5, 'character': 0}}}, 'commitCharacters': None, 'range': None, 'documentation': None}, {'filterText': '$this', 'insertText': '$this', 'sortText': '0003this', 'label': 'this', 'additionalTextEdits': None, 'detail': None, 'data': None, 'kind': 6, 'command': None, 'textEdit': {'newText': '$this', 'range': {'end': {'line': 5, 'character': 2}, 'start': {'line': 5, 'character': 0}}}, 'commitCharacters': None, 'range': None, 'documentation': None}
    ]
    ~~~

2. `$env:U`<kbd>Ctrl</kbd><kbd>Space</kbd> gets completed to `$env:$env:USERPROFILE` instead of `$env:USERPROFILE`:

    Payload:
    ~~~
    [
    {'filterText': '$env:UATDATA', 'insertText': '$env:UATDATA', 'sortText': '0001UATDATA', 'label': 'UATDATA', 'additionalTextEdits': None, 'detail': None, 'data': None, 'kind': 6, 'command': None, 'textEdit': {'newText': '$env:UATDATA', 'range': {'end': {'line': 5, 'character': 6}, 'start': {'line': 5, 'character': 0}}}, 'commitCharacters': None, 'range': None, 'documentation': None}, {'filterText': '$env:USERDNSDOMAIN', 'insertText': '$env:USERDNSDOMAIN', 'sortText': '0002USERDNSDOMAIN', 'label': 'USERDNSDOMAIN', 'additionalTextEdits': None, 'detail': None, 'data': None, 'kind': 6, 'command': None, 'textEdit': {'newText': '$env:USERDNSDOMAIN', 'range': {'end': {'line': 5, 'character': 6}, 'start': {'line': 5, 'character': 0}}}, 'commitCharacters': None, 'range': None, 'documentation': None}, {'filterText': '$env:USERDOMAIN', 'insertText': '$env:USERDOMAIN', 'sortText': '0003USERDOMAIN', 'label': 'USERDOMAIN', 'additionalTextEdits': None, 'detail': None, 'data': None, 'kind': 6, 'command': None,
     'textEdit': {'newText': '$env:USERDOMAIN', 'range': {'end': {'line': 5, 'character': 6}, 'start': {'line': 5, 'character': 0}}}, 'commitCharacters': None, 'range': None, 'documentation': None},
    {'filterText': '$env:USERDOMAIN_ROAMINGPROFILE', 'insertText': '$env:USERDOMAIN_ROAMINGPROFILE', 'sortText': '0004USERDOMAIN_ROAMINGPROFILE', 'label': 'USERDOMAIN_ROAMINGPROFILE', 'additionalTextEdits': None, 'detail': None, 'data': None, 'kind': 6, 'command': None, 'textEdit': {'newText': '$env:USERDOMAIN_ROAMINGPROFILE', 'range': {'end': {'line': 5, 'character': 6}, 'start': {'line': 5, 'character': 0}}}, 'commitCharacters': None, 'range': None, 'documentation': None}, {'filterText': '$env:USERNAME', 'insertText': '$env:USERNAME', 'sortText': '0005USERNAME', 'label': 'USERNAME', 'additionalTextEdits': None, 'detail': None, 'data': None, 'kind': 6, 'command': None, 
     'textEdit': {'newText': '$env:USERNAME', 'range': {'end': {'line': 5, 'character': 6}, 'start': {'line': 5, 'character': 0}}}, 'commitCharacters': None, 'range': None, 'documentation': None},
    {'filterText': '$env:USERPROFILE', 'insertText': '$env:USERPROFILE', 'sortText': '0006USERPROFILE', 'label': 'USERPROFILE', 'additionalTextEdits': None, 'detail': None, 'data': None, 'kind': 6, 'command': None, 
     'textEdit': {'newText': '$env:USERPROFILE', 'range': {'end': {'line': 5, 'character': 6}, 'start': {'line': 5, 'character': 0}}}, 'commitCharacters': None, 'range': None, 'documentation': None}
   ]
   ~~~

Quoting https://microsoft.github.io/language-server-protocol/specification:

> The `insertText` is subject to interpretation by the client side. Some tools might not take the string literally. For example VS Code when code complete is requested in this example `con<cursor position>`  and a completion item with an `insertText` of `console` is provided it will only insert `sole`. Therefore it is recommended to use `textEdit` instead since it avoids additional client side interpretation.

Therefore for to avoid additional logic and misinterpretations `textEdit` (when provided by server) should be supported and used instead `insertText` by LSP and also (optionally `additionalTextEdits`).

nanoant:
Sublime Text does not support explicit replacement with completion at given range, instead it replaces current completion prefix with the completion text. Knowing the completion prefix boundaries at plugin
side, we can trim `textEdit` range and text given by language server to the boundaries of the prefix.

Fixes #294

E.g. for PowerShell grammar and PowerShellEditorServices LS:

    $tru<Ctrl+Space> (prefix "tru") -> now "true" instead of "$true"
    $env:U<Ctrl+Space> (prefix "U") -> now "USER" instead of "$env:USER"

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -10,7 +10,7 @@ except ImportError:
 from .core.protocol import Request
 from .core.settings import settings
 from .core.logging import debug, exception_log
-from .core.protocol import CompletionItemKind
+from .core.protocol import CompletionItemKind, Range
 from .core.clients import client_for_view
 from .core.configurations import is_supported_syntax
 from .core.documents import get_document_position, purge_did_change
@@ -241,12 +241,28 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
             kind = item.get("kind")
             if kind:
                 hint = completion_item_kind_names[kind]
-        # label is an alternative for insertText if insertText not provided
-        insert_text = item.get("insertText") or label
+        # label is an alternative for insertText if neither textEdit nor insertText is provided
+        insert_text = self.text_edit_text(item) or item.get("insertText") or label
         if len(insert_text) > 0 and insert_text[0] == '$':  # sublime needs leading '$' escaped.
-            insert_text = '\$' + insert_text[1:]
+            insert_text = '\\$' + insert_text[1:]
         # only return label with a hint if available
         return "\t  ".join((label, hint)) if hint else label, insert_text
+
+    def text_edit_text(self, item) -> 'Optional[str]':
+        # try to handle textEdit if present
+        text_edit = item.get("textEdit")
+        if text_edit:
+            edit_range, edit_text = text_edit.get("range"), text_edit.get("newText")
+            if edit_range and edit_text:
+                edit_range = Range.from_lsp(edit_range)
+                last_start = self.last_location - len(self.last_prefix)
+                last_row, last_col = self.view.rowcol(last_start)
+                if last_row == edit_range.start.row == edit_range.end.row and edit_range.start.col <= last_col:
+                    # sublime does not support explicit replacement with completion
+                    # at given range, but we try to trim the textEdit range and text
+                    # to the start location of the completion
+                    return edit_text[last_col - edit_range.start.col:]
+        return None
 
     def handle_response(self, response: 'Optional[Dict]'):
         global resolvable_completion_items


plugin/completion.py
```diff
@@ -243,6 +243,24 @@ def format_completion(self, item: dict) -> 'Tuple[str, str]':
...
+                    start_row, start_col = range_start.get("line"), range_start.get("character")
+                    end_row, end_col = range_end.get("line"), range_end.get("character")
+                    if all([start_row is not None, start_col is not None,
+                            end_row is not None, end_col is not None]):
+                        last_start = self.last_location - len(self.last_prefix)
+                        last_row, last_col = self.view.rowcol(last_start)
+                        if last_row == start_row == end_row and start_col <= last_col:
+                            # sublime does not support explicit replacement with completion
+                            # at given range, but we try to trim the textEdit range and text
+                            # to the start location of the completion
```
rwols: Have you tried using the `LspApplyDocumentEditCommand`? (it's in `edit.py`)

nanoant: The problem is that `insert_text` is prepared for each completion before the choice is made by the user.

After choice is made Sublime simply replaces prefix with whatever was the completion text returned by plugin in previous step, at this point there is no way to run plugin callback such as `LspApplyDocumentEditCommand`.

On the other hand language server does not know what is the prefix Sublime tries to complete, it does receive only the location you want to complete and knows document content, hence when you do `$tru`<kbd>Ctrl</kbd><kbd>Space</kbd> language server does not know if the prefix was `tru` or `$tru`.

In my Powershell example Sublime prefix is `tru` however language server is returning `$true`, and after completion we end up with `$$true` (double dollar sign) instead `$true`.

My PR uses extra information from `text_edit` response (if present) to make sure the completion starts exactly where the prefix begins, so Sublime does the proper replacement.

It is not perfect solution, it does assumption that everything happens on single line, and may not work as intended in extreme cases e.g. if language server returns `%true` for `$tru`<kbd>Ctrl</kbd><kbd>Space</kbd>, even it does not make sense, it will be still completed to `$true` instead of `%true`, but we can't do anything about this with the way Sublime currently handles completions.

nanoant:
FYI fixed `editText` -> `textEdit` in my PR commit message and reworked a bit the description.

plugin/completion.py
rwols: Good points.

plugin/completion.py
```diff
@@ -243,6 +243,24 @@ def format_completion(self, item: dict) -> 'Tuple[str, str]':
                 hint = completion_item_kind_names[kind]
         # label is an alternative for insertText if insertText not provided
         insert_text = item.get("insertText") or label
+        # try to handle textEdit if present
+        text_edit = item.get("textEdit")
+        if text_edit:
```
tomv564: Can we hide this behaviour behind a global LSP setting for now? I'd like to avoid hitting this code path automatically for all language servers that happen to send textEdit.

plugin/completion.py
```diff
@@ -243,6 +243,24 @@ def format_completion(self, item: dict) -> 'Tuple[str, str]':
...
+            edit_range, edit_text = text_edit.get("range"), text_edit.get("newText")
+            if edit_range and edit_text:
+                range_start, range_end = edit_range.get("start"), edit_range.get("end")
+                if range_start and range_end:
+                    start_row, start_col = range_start.get("line"), range_start.get("character")
+                    end_row, end_col = range_end.get("line"), range_end.get("character")
+                    if all([start_row is not None, start_col is not None,
+                            end_row is not None, end_col is not None]):
```
tomv564: The range is probably specified to be required, in which case I like to avoid the None checks (I'm fine with errors from non-compliant servers).
There is existing logic for parsing ranges in `protocol.py`

nanoant:
@tomv564 No pbm for late review, I'll adapt my changes to your comments beginning of the next week. Cheers.

nanoant:
@tomv564 I have applied your remarks to my patch, except your request to make `textEdit` optional. I would argue to make it default -- justification quoting https://microsoft.github.io/language-server-protocol/specification:

> The insertText is subject to interpretation by the client side. Some tools might not take the string literally. For example VS Code when code complete is requested in this example con<cursor position> and a completion item with an insertText of console is provided it will only insert sole. Therefore it is recommended to use textEdit instead since it avoids additional client side interpretation.

I think `CompletionHelper.text_edit_text` checks (single line edit and position matching) should be safe enough to make it default. However I am open for a discussion, especially if you can think of any example that can break `text_edit_text`.

plugin/completion.py
```diff
@@ -241,13 +241,29 @@ def format_completion(self, item: dict) -> 'Tuple[str, str]':
             kind = item.get("kind")
             if kind:
                 hint = completion_item_kind_names[kind]
-        # label is an alternative for insertText if insertText not provided
-        insert_text = item.get("insertText") or label
+        # label is an alternative for insertText if neither textEdit nor insertText is provided
+        insert_text = self.text_edit_text(item) or item.get("insertText") or label
         if len(insert_text) > 0 and insert_text[0] == '$':  # sublime needs leading '$' escaped.
-            insert_text = '\$' + insert_text[1:]
+            insert_text = '\\$' + insert_text[1:]
```
tomv564: I'm a bit worried about this change as it was a working fix for completions from PHP language servers. Can you tell me what is problematic, or should we take this change in a separate PR?

nanoant: It is just about escaping in Python -- `'...'` still respects backslash escape codes, `\$` is not proper escape code so Python leaves it as is, however Linters complain about this invalid escape code, hence backslash should be escaped, or raw Python string should be used.

nanoant:
@tomv564 Thanks a lot!

PR #313, opened 2018-04-26
make request and notication classes customizable

randy3k:
This looks like a huge PR, but actually there is only one significant change - the customizable classes at [here](https://github.com/randy3k/LSP/blob/fc829fc9c31abe7f256592ab0bf36155217f9e2e/plugin/core/rpc.py#L249-L250). It allows plugin developers to override the default Request and Notification classes if necessary. A potential usage is to honor the [extended protocol](https://github.com/llvm-mirror/clang-tools-extra/blob/849d20a96082bf06134957850445d2e33dba3bc9/clangd/Protocol.h#L354-L354) used by clangd. See also https://github.com/Microsoft/language-server-protocol/issues/255.

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -9,7 +9,7 @@ except ImportError:
 
 from .core.clients import client_for_view
 from .core.clients import LspTextCommand
-from .core.protocol import Request, Range
+from .core.protocol import Range
 from .core.documents import get_position
 from .core.diagnostics import get_point_diagnostics
 from .core.url import filename_to_uri
@@ -42,7 +42,7 @@ class LspCodeActionsCommand(LspTextCommand):
                 sel.clear()
                 sel.add(sublime.Region(pos))
 
-            client.send_request(Request.codeAction(params), self.handle_codeaction_response)
+            client.send_request(client.request_class.codeAction(params), self.handle_codeaction_response)
 
     def handle_codeaction_response(self, response):
         titles = []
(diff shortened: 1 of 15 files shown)


randy3k:
Check https://gist.github.com/randy3k/b18ceb94a4daf010528c1d281935cc65 for such a implemented LanguageHandler.

tomv564:
Ambitious change - replacing the standardised APIs behaviour feels like it undermines the idea of LSP (especially as now the custom notification class is responsible for maintaining the "didOpen" implementation in LSP). 

* I think LSP would be better off exposing a much smaller extension surface
* Cquery has less invasive ways of getting flags, as discussed in https://github.com/cquery-project/cquery/issues/486#issuecomment-370171388

Perhaps replacing document metadata is something already under consideration by the clangd developers?

rwols:
> Perhaps replacing document metadata is something already under consideration by the clangd developers?

With a `compile_commands.json` file at the root of the project, there is no need for sending metadata.

randy3k:
I have switched to cquery and the following project setting solve my problem. As it is not likely that this PR will be merged. I would close it for now.

```
	"settings":
	{
		"LSP":
		{
			"cquery-r":
			{
				"enabled": true,
				"initializationOptions":
				{
					"cacheDirectory": "/tmp/cquery",
					"extraClangArguments": [
						"-I/Library/Frameworks/R.framework/Headers"
					]
				},
			}
		}
	}
```

PR #314, opened 2018-04-26
Add diagnostics count to statusbar

### LINKED ISSUE #186: Error panel pops up too fast

predragnikolic:
closes #186 

### CHANGED
- changed the default value for `auto_show_diagnostics_panel` to false,
- before `auto_show_diagnostics_panel` was responsible for automatically opening and closing the diagnostics panel,
now it is responsible only for opening, as the panel will automatically close when there are no errors(no need to show an empty panel),
- updated the docs for the settings I changed and added.

### ADDED
- new setting for configuring the visibility of the count in the status bar `show_diagnostics_count_in_view_status`, defaults to true,
- added diagnostics count to the status bar.

Waiting for feedback, if you have something you would change, or reject, please say so.. :)

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -167,10 +167,13 @@
 
   // Open and close the diagnostics panel automatically,
   // depending on available diagnostics.
-  "auto_show_diagnostics_panel": true,
+  "auto_show_diagnostics_panel": false,
 
   // Show in-line diagnostics using phantoms for unchanged files.
   "show_diagnostics_phantoms": false,
+
+  // Show errors and warnings count in the status bar 
+  "show_diagnostics_count_in_view_status": true,
 
   // Show the diagnostics description of the code
   // under the cursor in status bar if available.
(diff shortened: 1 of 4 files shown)


predragnikolic:
Messed up the commit history, that's why I closed it. Tried to get it in order, but didn't succeed. 

Sorry for the boder :-)

predragnikolic:
Ok Tom :) 

Have you any suggestion on how to avoid looping over diagnostics?

tomv564:
Sorry, I really haven't had time to have a closer look. Tested it last night, merged it now, going to add some small fixes in master. Nicely done!

predragnikolic:
Thank you Tom :)

PR #339, opened 2018-06-13
Give priority to more specific scopes. Fixes #338

### LINKED ISSUE #338: Scopes priorities while selecting configuration
## Bug:

When there are multiple language servers configured, all of which are for similar scopes (Ex. `source.json`, `source.json.sublime.settings`) the configuration with the most specific scope should be preferred; however right now one or the other could "win", some times leading to erroneous configuration.

Example comes from configuring **vscode-json-languageserver** to work with both `json` and `jsonc` languageIds.

### Suggestion:

Give priority to the configuration with the most specific scope that matches.

Kronuz:
This pull request makes it so `get_scope_client_config()` returns the most specific configuration that matches the currently selected scope.

### DIFF
--- a/plugin/core/configurations.py
+++ b/plugin/core/configurations.py
@@ -17,6 +17,10 @@ window_client_configs = dict()  # type: Dict[int, List[ClientConfig]]
 
 
 def get_scope_client_config(view: 'sublime.View', configs: 'List[ClientConfig]') -> 'Optional[ClientConfig]':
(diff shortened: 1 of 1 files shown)


Kronuz:
I've added a comment for that now :)

Also, I believe we should take that `sel = view.sel()` and `sel[0].begin()` out of the loops... but that's other issue.

PR #341, opened 2018-06-13
Allow using partial words to match syntaxes. Fixes #340

### LINKED ISSUE #340: Use case-insensitive Regexps for matching syntaxes
### Suggestion:

When using `settings.get('syntax')` to filter configurations, it's needed such configurations to match exactly. Many times there are more than one single package for any given syntax (and oftentimes there are packages that haven't been updated in some time and for which syntax files have moved or extension definition changed from `tmLanguage` to `sublime-syntax`). This can lead in the need to configure server languages "guessing" what syntax files are configured for it.

Real-world examples of such:

**Markdown**:
1. `Packages/Markdown/Markdown.sublime-syntax`
2. `Packages/Markdown Extended/Syntaxes/Markdown Extended.sublime-syntax`
3. `Packages/Markdown Extended/Syntaxes/Markdown Extended.tmLanguage`

**SCSS**:
1. `Packages/Sass/Syntaxes/SCSS.sublime-syntax`
2. `Packages/SCSS/SCSS.tmLanguage`

**JSON**:
1. `Packages/JavaScript/JSON.sublime-syntax`
2. `Packages/Babel/JSON (Babel).sublime-syntax`
3. `Packages/PackageDev/Package/Sublime Text Settings/Sublime Text Settings.sublime-syntax`

It could very well be better to have a set of case insensitive regexps to match the language from the syntax (after all scopes are later used to confirm). That way we could have, accordingly, the following syntaxes that would do a `re.search()`.

```json
  "syntaxes": [
    "Markdown"
  ]
```

```json
  "syntaxes": [
    "SCSS"
  ]
```

```json
  "syntaxes": [
    "JSON",
    "Sublime Text Settings"
  ]
```

_Alternatively_, we could add `languages` as the list of regexps that match, keeping `syntaxes` for exact matches and using `languages` as fallbacks to figure out the language configuration.

Kronuz:
This is compatible with the current configurations, but also supports using partial words for syntaxes. 

Ex., to match any syntax with "JSON" in it:

```json
"syntaxes": ["json"]
```

### DIFF
--- a/plugin/core/configurations.py
+++ b/plugin/core/configurations.py
@@ -1,3 +1,4 @@
+import re
 import sublime
 
 from .settings import ClientConfig, client_configs
@@ -105,14 +106,14 @@ def apply_window_settings(client_config: 'ClientConfig', view: 'sublime.View') -
 def is_supportable_syntax(syntax: str) -> bool:
     # TODO: filter out configs disabled by the user.
     for config in client_configs.defaults:
-        if syntax in config.syntaxes:
+        if re.search(r'|'.join(r'\b%s\b' % re.escape(s) for s in config.syntaxes), syntax, re.IGNORECASE):
             return True
     return False
 
 
 def is_supported_syntax(syntax: str) -> bool:
     for config in client_configs.all:
-        if syntax in config.syntaxes:
+        if re.search(r'|'.join(r'\b%s\b' % re.escape(s) for s in config.syntaxes), syntax, re.IGNORECASE):
             return True
     return False
 


tomv564:
Looks good, we can ignore the test coverage for now.
Forgive my inexperience, but why the '\b' surrounding each escaped string?

Kronuz:
It means word boundary, so that if for instance I add `[“c”]` that won’t match `Coffescript`, only “C”

Kronuz:
@tomv564, could you please merge this one in?

I'd like to start working in #349 now so I can get json/jsonc, c/c++/objective-c, css/scss/sass/less plugins ready. 😁

tomv564:
Thanks, sorry for the slow response!

PR #357, opened 2018-06-26
Bugfix: format completion

### LINKED ISSUE #348: Unknown `CompletionItemKind`s are not handled properly
### Bug:
As it stands now, this LSP client specifies which `CompletionItemKind`s it supports in `TextDocumentClientCapabilities `. [According to the spec](https://microsoft.github.io/language-server-protocol/specification#initialize), if `InitializeParams.capabilities.textDocument.completion.completionItemKind.valueSet` is set, the client also guarantees that any value outside of `valueSet` will be handled gracefully and fall back to a default value:
```typescript
completionItemKind?: {
	/**
	 * The completion item kind values the client supports. When this
	 * property exists the client also guarantees that it will
	 * handle values outside its set gracefully and falls back
	 * to a default value when unknown.
	 *
	 * If this property is not present the client only supports
	 * the completion items kinds from `Text` to `Reference` as defined in
	 * the initial version of the protocol.
	 */
	valueSet?: CompletionItemKind[];
},
```
However, this is not the case here, the client emits an error instead of handling the unknown value:
```
Error handling server payload
  File "C:\Users\laure\AppData\Roaming\Sublime Text 3\Installed Packages\LSP.sublime-package\plugin/core/rpc.py", line 322, in receive_payload
  File "C:\Users\laure\AppData\Roaming\Sublime Text 3\Installed Packages\LSP.sublime-package\plugin/core/rpc.py", line 341, in response_handler
    self._response_handlers[handler_id](result)
  File "C:\Users\laure\AppData\Roaming\Sublime Text 3\Installed Packages\LSP.sublime-package\plugin/completion.py", line 261, in handle_response
    self.completions = list(self.format_completion(item) for item in items)
  File "C:\Users\laure\AppData\Roaming\Sublime Text 3\Installed Packages\LSP.sublime-package\plugin/completion.py", line 261, in <genexpr>
    self.completions = list(self.format_completion(item) for item in items)
  File "C:\Users\laure\AppData\Roaming\Sublime Text 3\Installed Packages\LSP.sublime-package\plugin/completion.py", line 237, in format_completion
KeyError: 19
```
This fallback behavior is also mandatory for `InitializeParams.capabilities.workspace.symbol.symbolKind.valueSet` and `InitializeParams.capabilities.textDocument.documentSymbol.symbolKind.valueSet`, although I haven't tested them in this client yet, so I don't know if they are affected as well.

rwols:
- Update completion item kinds
- More safely fetch the kind (if it doesn't exist, it reverts to None)

I haven't tested this, but this should fix #348. Ping @gobijan and @LaurentTreguier, could you give this a spin with the Ruby server?

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -240,7 +240,7 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
         elif settings.completion_hint_type == "kind":
             kind = item.get("kind")
             if kind:
-                hint = completion_item_kind_names[kind]
+                hint = completion_item_kind_names.get(kind)
         # label is an alternative for insertText if neither textEdit nor insertText is provided
         insert_text = self.text_edit_text(item) or item.get("insertText") or label
         if len(insert_text) > 0 and insert_text[0] == '$':  # sublime needs leading '$' escaped.
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -54,6 +54,13 @@ class CompletionItemKind(object):
     Color = 16
     File = 17
     Reference = 18
+    Folder = 19
+    EnumMember = 20
+    Constant = 21
+    Struct = 22
+    Event = 23
+    Operator = 24
+    TypeParameter = 25
 
 
 class DocumentHighlightKind(object):


tomv564:
Looks good to me, thanks!

We won't notice as quickly when even more CompletionItemKinds are introduced this way, but if servers are not required to respect a client's "supported completion item kinds", then we may as well gracefully degrade all the time.

PR #358, opened 2018-06-27
Dynamic configuration of settings and languageId

### LINKED ISSUE #356: ClientConfig should be configured per window and/or view
### Suggestion:

As it currently is right now, there's a single global `ClientConfig` object per language handler, and it later gets overridden by the project/window settings. Instead, it should be created when needed, already with the window settings in place, and it should be created by the language handlers dynamically; allowed to be configured differently per window or even view.
 
#### Reson

The reason is simple: some language servers could decide dynamically, per window and even view, to use different configurations and settings, additionally to those set up by the ST project settings. For example: At the window level, Python language handler could decide to dynamically add ST Packages directory as an extra Python path; or disable/enable the client configuration based upon other heuristics particular to the project; or yet other language handler could be adding extra library paths based on the current open project base folders (`window.folders()`). At the view level, the currently open view could decide different `languageId` from the scopes in the view, and there currently is just one allowed per server.

#### Proposal

One way this could be fixed/implemented without a huge overhaul is if `ClientConfig` had methods instead of properties to extract certain configurations. For example, for languageId, it'd be receiving the current view so it returns the proper languageId: `ClientConfig.get_language_id(view)`, analogously, `settings` would be receiving a window: `ClientConfig.get_settings(window)` so it could get settings tailored to the project (added extra paths from project folders).

Kronuz:
Dynamic configuration of settings (per window) and languageId (per view).

I'm using this to allow getting custom extra paths per window and different languageId per view (for the same server). This fixes #356


plugin/core/main.py
```diff
@@ -167,7 +167,7 @@ def handle_initialize_result(result, client, window, project_path, config):
     client.send_notification(Notification.initialized())
     if config.settings:
         configParams = {
-            'settings': config.settings
+            'settings': config.get_settings(window)
```
tomv564: This get-settings-by-window function looks incomplete (and perhaps not needed), correct?

tomv564:
Functionality has since been implemented.

PR #359, opened 2018-06-27
Show diagnostic source as part of the message

Kronuz:
### Suggestion:

When using plugins for language servers, or language servers that natively return the source of a diagnostic, the source should be shown as part of the hover popup.

For example, Python's pycodestyle, flake8 and mypy all return the source for each diagnostic; without adding the source, it's difficult to guess where the errors are coming from, this is a suggestion for. This pull request adds the source, as in: "[mypy] some message" or "[pyflakes] E501: line too long".

### DIFF
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -77,7 +77,7 @@ class LspHoverCommand(LspTextCommand):
 
     def diagnostics_content(self, diagnostics):
         formatted_errors = list(
-            "<pre>{}</pre>".format(diagnostic.message)
+            "<pre>{}</pre>".format("[{}] {}".format(diagnostic.source, diagnostic.message) if diagnostic.source else "{}".format(diagnostic.message))
             for diagnostic in diagnostics
             if diagnostic.severity == DiagnosticSeverity.Error)
         formatted = []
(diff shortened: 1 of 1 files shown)


tomv564:
I think you pushed the wrong commit to this branch!

Kronuz:
I think I might 🤔 ...i certainly did. I’ll rectify that as soon as I can

Kronuz:
@tomv564, done! I got the wrong commit for this branch; fixed it!

It's getting difficult to manage all these branches... you think we could work in merging some? I could help more if I had commit access, I'd love to chat a bit about some things related to the project.

rwols:
I think we can all trust @tomv564 at being a good maintainer and leading this project in the right direction. If your pull requests needs work according to @tomv564, then please respect that.

tomv564:
No worries, I see your PRs often contain  a similarly named commit several times over - this repo must not make your workflow easier !
Speaking of commit access, would it be a good time to move the repo elsewhere? Do I create an org, or is there a good org for community-maintained sublime packages?

plugin/hover.py
```diff
@@ -89,7 +89,7 @@ def diagnostics_content(self, diagnostics):
             formatted.append("</div>")
 
         formatted_warnings = list(
-            "<pre>{}</pre>".format(diagnostic.message)
+            "<pre>{}</pre>".format("[{}] {}".format(diagnostic.source, diagnostic.message) if diagnostic.source else "{}".format(diagnostic.message))
```
tomv564: The else case can be just `diagnostic.message` - or perhaps we want to only have maximum 1 format call per diagnostic?

Kronuz:
@tomv564, yes, that part can be without that extra `format()`, that is right. I’ve already started putting together a few repositories for plugins for different languages. I’d want to have a word with you some time soon if you have a chance. Do you use WhatsApp, Telegram, Google Chat, or something else to have a chat?

PR #360, opened 2018-06-27
Testable sessions

tomv564:
Promotes ConfigState to a Session object that also handles initialising and shutting down LSP clients.
Introduces the process module for process-related functionality
Removes a chunk of specialised code out of main.py, including shorting the session initializing callback.


tomv564:
Connecting to already-running TCP servers might now work to some degree.
Unfortunately this will probably re-introduce bugs around session startup / shutdown, but I don't have time for exhaustive testing.

PR #361, opened 2018-06-28
Add Ruby/Rails and Javascript-Typescript to default settings.

gobijan:


tomv564:
Hi, as people have started to get conflicting language servers in the default settings, I am trying to move away from them.
Instead, would you mind opening a PR against the documentation so we can document the settings there?
The future way of helping users automatically configure a language server is by implementing a LanguageHandler, this base class provides hooks for installing and configuring the language server.
An example is in the LSP-tss, LSP-dotty or for example https://github.com/randy3k/LSP-julia

tomv564:
Thanks for adding the documentation, closing this issue now!

PR #364, opened 2018-06-29
Prevent os.path.relpath exception for project and file on 2 drives

nanoant:
Otherwise os.path.relpath may emit for example:
`ValueError: path is on mount 'D:', start on mount 'C:'`

This is just a tiny fix for Windows users. Problem occurs when you have your project on one drive, but you add a folder from different drive.


tomv564:
Multiple folders are not really supported well by this package, it's been an endless source of issues (search the issues for "path is on mount") so I'm happier discouraging people using LSP in that way.

I'll merge this workaround for now, thanks for providing a workaround.

Issue #206, opened 2017-11-16
Evaluate new Command Palette


tomv564:
New in dev build 3154, for any commands where we currently ask for user input/choice (eg. Rename) or would like to in the future (Stop/Start/Enable/Disable server)
https://forum.sublimetext.com/t/dev-build-3154/33286

PR #367, opened 2018-07-03
Add rename text input handler

predragnikolic:
Ask for new symbol name though the command palette, instead of the window input panel. #206

output

There is a [comment](https://github.com/tomv564/LSP/compare/master...predragnikolic:add-new-rename-text-input-handler?expand=1#diff-c80a16ac420975c347556b49acdda80cR27) I left for you to have a look. 

Can you please tell me, what the purpose of the `event` argument in the `run` method and when do you pass it by to the command?

```python
class LspSymbolRenameCommand(LspTextCommand):
    #...
    def run(self, edit, new_name, event=None):
        pos = get_position(self.view, event)
        params = get_document_position(self.view, pos)
```
Also Tom, is there something you would like to change? :)


rwols:
You could also choose to change the minimum required build number in your package control submission.

predragnikolic:
Thanks for explaining :)
> Did you test the rename from the context menu after your changes?

Sorry, I didn't, but I fixed it. 
And now renaming works in all three cases, when it is triggered through the:
* keyboard shortcut
* context menu
* command palette

> You added Rename to the panel so it will appear to work even when the server does not support it?

Just like in the context menu, if the `is_enabled` method returns `False` the command is grayed out in the menu.
As for the command palette, Sublime won't show the command if the `is_enabled` returns 'False', so the command will only be visible if the user can actually rename the symbol. 

>  we could do a feature detection?

I will let you decide if you want a feature detection :D
But I think people should upgrade to a newer version if they use an older build.
I don't know if they lose something by doing that? If you want a feature detection, I will try to do it :D

> **Note**: In the python language server, if I rename a symbol, and try to rename it again, it wont take affect. But if I rename the symbol, save the file, and then rename it again, it works.
> In your typescript language server, everything works great, I can rename the symbol as much as I want, the above problem only affects `pyls` and that problem existed even before this PR.

PR #377, opened 2018-07-17
Add Ruby / Ruby on Rails config to docs.

gobijan:


tomv564:
Thanks for your contribution. I believe `source.ruby` and a Syntax entries with just "Ruby" and "Rails" should work as well - that feature should be documented!

PR #385, opened 2018-07-29
Fix: log_stderr didn't work anymore

rwols:


rwols:
What's up with these unit tests?

tomv564:
There are two sets of tests now: unit tests which run outside of sublime, and functional tests using UnitTesting.

You cannot import modules using the sublime API (eg settings) in unit-tested code. It needs to be passed in instead if we want to keep the option configurable. I’ve also been tempted to remove settings if the default is likely to suit everyone

tomv564:
Thanks, please ignore my earlier comment about the unit tests. Had I looked at the failure in travis, I could have helped you with the simple fix.

rwols:
Thanks for the fix! I don't have much time during regular weekdays, but wanted to dig into the unit test stuff this weekend.

PR #395, opened 2018-08-17
Allow to merge global and project settings

st4lk:
This is my global LSP settings (Preferences -> Package Settings -> LSP -> Settings):
```
// Settings in here override those in "LSP/LSP.sublime-settings",

{
  "clients": {
    "pyls":
    {
      // ...
      "settings": {
        "pyls": {
          "follow_imports": true 
        }
      },
      "env": {
        "global_var": "Merge works!"
      },
  }
}
```

And this is my project settings (Project -> Edit Project):
```
{
  // ...
  "settings":
  {
    "LSP":
    {
      "pyls":
      {
        "env":
        {
          "custom_var": "Testing LSP Sublime!"
        },
        "settings":
        {
          "pyls":
          {
            "python_interpreter": "/path/to/venv/bin/python",
            "python_virtualenv": "/path/to/venv/"
          }
        }
      }
    }
  }
}

```

My expectation is that final settings sent to language server would be:
```json
{
  "pyls":
  {
    "python_interpreter": "/path/to/venv/bin/python",
    "python_virtualenv": "/path/to/venv/",
    "follow_imports": true
  }
}
```

But they are
```
{
  "pyls":
  {
    "python_interpreter": "/path/to/venv/bin/python",
    "python_virtualenv": "/path/to/venv/"
  }
}
```
So they are taken from project settings. Package settings are completely ignored.

This patch adds an ability to merge settings and env. To do it, need to specify it explicitly in package settings (Preferences -> Package Settings -> LSP -> Settings):
```
{
  "clients": {
    "pyls":
    {
      // ...
      "settings": {
        "pyls": {
          "follow_imports": true 
        }
      },
      "env": {
        "global_var": "Merge works!"
      },
  },
  "merge_client_env": true,
  "merge_client_settings": true
}
```

The same behaviour is for `"env"`.

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -213,5 +213,11 @@
   "log_stderr": false,
 
   // Show full JSON-RPC responses in the console
-  "log_payloads": false
+  "log_payloads": false,
+
+  // Merge project env with global env. If false, then override.
+  "merge_client_env": false,
+
+  // Merge project settings with global settings. If false, then override.
+  "merge_client_settings": false
 }
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -213,11 +213,5 @@
   "log_stderr": false,
 
   // Show full JSON-RPC responses in the console
-  "log_payloads": false,
-
-  // Merge project env with global env. If false, then override.
-  "merge_client_env": false,
-
-  // Merge project settings with global settings. If false, then override.
-  "merge_client_settings": false
+  "log_payloads": false
 }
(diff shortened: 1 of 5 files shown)


tomv564:
This is an example where I'd prefer to avoid extra settings and have 1 predicable behaviour good enough for all users.

If we consider always merging those two values, what is the risk? Users would have to become aware of what they put in global client configs?

st4lk:
Agree, have removed settings.

PR #413, opened 2018-09-17
Change look of references

### LINKED ISSUE #412: Show line of reference in panel
### Suggestion:

When displaying references, it would be nice if the line was also copied into the panel at the bottom. Maybe with a configurable number of surrounding lines for better context and the actual usage highlighted with an outline box.

Would also be helpful in remembering whose references you are actually seeing.

I'm using pyls, FWIW.

predragnikolic:
closes #412

The only thing about this PR that I don't understand is the `References.sublime-syntax` file. 
I just copied the `Diagnostics.sublime-syntax` and made some adjustments, like deleted some things, renamed some.
The syntax looks ok to me, but I would suggest someone else who knows better to have a look at it, just to check if everything is alright.  

If you have any question, concern, let me know.

### DIFF
--- a/Syntaxes/References.sublime-syntax
+++ b/Syntaxes/References.sublime-syntax
@@ -1,7 +1,13 @@
 %YAML 1.2
 ---
+# [Subl]: https://www.sublimetext.com/docs/3/syntax.html
+# [LSP]: https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md
 hidden: true
 scope: output.lsp.references
+
+variables:
+  start_of_reference_body: ^\s+(?=\d)
+  filename_and_colon: ^\s*(\S)\s+(.*)(:)$
 
 contexts:
   main:
@@ -9,26 +15,33 @@ contexts:
     - include: references-body
 
   references-preamble:
-    # References to "function" at path\file.py:
-    - match: ^(.*)(:)$
+    - match: '{{filename_and_colon}}'
       captures:
-        0: meta.references.preamble.lsp
-        1: string.unquoted.lsp
-        2: punctuation.separator.lsp
+        0: meta.reference.preamble.lsp
+        1: punctuation.section.references.preample.lsp
+        2: string.unquoted.lsp
+        3: punctuation.separator.lsp
 
   references-body:
-    # Must match the
-    #    - "result_file_regex" setting
-    #    - reference output format
-    # Examples:
-    #    ◌ c:\path name\file name.py 423:32
-    #    ◌ /unixpath/file.py 815:1
-    - match: ^\s+(\S)\s+((\S.+)\s+(\d+)(?:(:)(\d+))?)$
+    - match: '{{start_of_reference_body}}'
+      push:
+        - ensure-reference-meta-scope
+        - expect-line-maybe-column
+
+  ensure-reference-meta-scope:
+    - meta_scope: meta.reference.body.lsp
+    - match: ""  # match the empty string
+      pop: true
+
+  expect-line-maybe-column:
+    - include: pop-at-end
+    - match: (\d+)(?:(:)(\d+))?
       captures:
-        0: meta.references.body.lsp
-        1: punctuation.section.path.reference.lsp
-        2: meta.path.reference.lsp
-        3: entity.name.file.lsp
-        4: constant.numeric.integer.decimal.lsp
-        5: punctuation.separator.lsp
-        6: constant.numeric.integer.decimal.lsp
+        1: constant.numeric.integer.decimal.lsp
+        2: punctuation.separator.lsp
+        3: constant.numeric.integer.decimal.lsp
+      pop: true
+
+  pop-at-end:
+    - match: $
+      pop: true
--- a/Syntaxes/References.sublime-syntax
+++ b/Syntaxes/References.sublime-syntax
@@ -19,14 +19,21 @@ contexts:
       captures:
         0: meta.reference.preamble.lsp
         1: punctuation.section.references.preample.lsp
-        2: string.unquoted.lsp
+        2: string.unquoted.lsp entity.name.file.references.lsp
         3: punctuation.separator.lsp
 
   references-body:
     - match: '{{start_of_reference_body}}'
       push:
         - ensure-reference-meta-scope
+        - code
         - expect-line-maybe-column
+      
+  code:
+    - match: '(?=\S)'
+      set:
+        - meta_scope: source
+        - include: pop-at-end
 
   ensure-reference-meta-scope:
     - meta_scope: meta.reference.body.lsp
--- a/plugin/references.py
+++ b/plugin/references.py
@@ -102,11 +102,11 @@ class LspSymbolReferencesCommand(LspTextCommand):
         Example:
         {
             'main.py': [
-                { Region, 'from .src.one import one' },
-                { Region, 'two = 1 + one()' }
+                { Point, 'from .src.one import one' },
+                { Point, 'two = 1 + one()' }
             ],
             'src/one.py': [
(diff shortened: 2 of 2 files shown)


Syntaxes/References.sublime-syntax
```diff
@@ -1,34 +1,47 @@
...
-    # References to "function" at path\file.py:
-    - match: ^(.*)(:)$
+    - match: '{{filename_and_colon}}'
       captures:
-        0: meta.references.preamble.lsp
-        1: string.unquoted.lsp
-        2: punctuation.separator.lsp
+        0: meta.reference.preamble.lsp
+        1: punctuation.section.references.preample.lsp
+        2: string.unquoted.lsp
```
FichteFoll: I would even go as far and assign `entity.name.file.references.lsp` here.

Syntaxes/References.sublime-syntax
```diff
@@ -1,34 +1,47 @@
...
-    # Must match the
-    #    - "result_file_regex" setting
-    #    - reference output format
-    # Examples:
-    #    ◌ c:\path name\file name.py 423:32
-    #    ◌ /unixpath/file.py 815:1
-    - match: ^\s+(\S)\s+((\S.+)\s+(\d+)(?:(:)(\d+))?)$
+    - match: '{{start_of_reference_body}}'
+      push:
+        - ensure-reference-meta-scope
```
FichteFoll: Since you already have the multi-push, you might as well go ahead and highlight any following referenced code as `source` (unspecified which one as we don't know). Some color schemes, like mine, have background color highlighting for inline code segments on `source source`, for example.

I.e.
```yaml
  references-body:
    - match: '{{start_of_reference_body}}'
      push:
        - ensure-reference-meta-scope
        - code
        - expect-line-maybe-column
      
  code:
    - match: '(?=\S)'
      push:
        - meta_scope: source
        - include: pop-at-end
```

Syntaxes/References.sublime-syntax
predragnikolic: Is this alright? 
```diff
    0: meta.reference.preamble.lsp
    1: punctuation.section.references.preample.lsp
-   2: string.unquoted.lsp
+   2: string.unquoted.lsp entity.name.file.references.lsp 
    3: punctuation.separator.lsp
```

Syntaxes/References.sublime-syntax
predragnikolic: Thanks for the instructions. 
2018-09-21-132934_683x564_scrot
But how do I stop the scope `source` from spanning to the next line?

```
          38:13   file_path = uri_to_filename(location.get("uri"))
	  40:47   file_location = "{}:{}:{}".format(file_path, start.row + 1, start.col + 1)
  scopes  ^ output.lsp.references meta.reference.body.lsp source

Comment (FichteFoll) [reply Syntaxes/References.sublime-syntax]: Ah, my bad. You need to change toe "push" to a "set", otherwise you won't be able to leave it.

File: Syntaxes/References.sublime-syntax (hunk 1 above)
Comment (FichteFoll) [reply Syntaxes/References.sublime-syntax]: Yeah, that's good. This way it even has a fallback scope with the string.

**Comment by predragnikolic** [comment]
Danke @FichteFoll  :)

File: plugin/references.py [hunk 3]
```diff
@@ -57,44 +58,85 @@ def run(self, edit, event=None):
...
+        for reference in references:
+            file_path = uri_to_filename(reference.get("uri"))
+            relative_file_path = os.path.relpath(file_path, base_dir)
+
+            point = Point.from_lsp(reference.get('range').get('start'))
+            # get line of the reference, to showcase its use
+            reference_line = linecache.getline(file_path, point.row + 1).strip()
+
+            # we don't want to cache the line, we always want to get fresh data
+            linecache.clearcache()
```
tomv564: Should `clearcache` be called in a loop? 
Is clearing the line cache necessary for "grouping references" or should it be done in another part of the code ?
I assume the point is to use _some_ caching per references request?

plugin/references.py
```diff
@@ -57,44 +58,85 @@ def run(self, edit, event=None):
...
     def want_event(self):
         return True
 
-
-def format_reference(reference, base_dir):
-    start = Point.from_lsp(reference.get('range').get('start'))
-    file_path = uri_to_filename(reference.get("uri"))
-    relative_file_path = os.path.relpath(file_path, base_dir)
-    return " ◌ {} {}:{}".format(relative_file_path, start.row + 1, start.col + 1)
+    def _get_formated_references(self, references: 'List[Dict]', base_dir) -> str:
```
tomv564: Spelling: "formatted"

plugin/references.py
```diff
@@ -57,44 +58,85 @@ def run(self, edit, event=None):
...
+        formated_references = self._get_formated_references(response, base_dir)
+
+        panel = ensure_references_panel(window)
+        panel.settings().set("result_base_dir", base_dir)
+
+        panel.set_read_only(False)
+        panel.run_command("lsp_clear_panel")
+        window.run_command("show_panel", {"panel": "output.references"})
+        panel.run_command('append', {
+            'characters': "{} references for '{}'\n\n{}".format(references_count, word, formated_references),
```
tomv564: Spelling: "formatted"

plugin/references.py
```diff
@@ -57,44 +58,85 @@ def run(self, edit, event=None):
...
+        Example:
+        {
+            'main.py': [
+                { Point, 'from .src.one import one' },
+                { Point, 'two = 1 + one()' }
+            ],
+            'src/one.py': [
+               { Point, 'def one():' }
+            ]
+        }"""
```
tomv564: I personally feel doc comments are not really that useful when they take up 50% on an internal method.
The beauty of python is that reading the code should be just as clear!

plugin/references.py
```diff
@@ -57,44 +58,85 @@ def run(self, edit, event=None):
...
+        dict = {}  # type: Dict[str, List[Dict]]
```
tomv564: Please name this variable according to its purpose or contents

plugin/references.py
```diff
@@ -57,44 +58,85 @@ def run(self, edit, event=None):
...
 
-
-def format_reference(reference, base_dir):
-    start = Point.from_lsp(reference.get('range').get('start'))
-    file_path = uri_to_filename(reference.get("uri"))
-    relative_file_path = os.path.relpath(file_path, base_dir)
-    return " ◌ {} {}:{}".format(relative_file_path, start.row + 1, start.col + 1)
+    def _get_formated_references(self, references: 'List[Dict]', base_dir) -> str:
+        grouped_references = self._group_references_by_file(references, base_dir)
+        formated_references = self._format_references(grouped_references)
```
tomv564: Spelling: "formatted"

plugin/references.py
```diff
@@ -57,44 +58,85 @@ def run(self, edit, event=None):
...
+        # return if there are no references
+        if references_count < 1:
             window.run_command("hide_panel", {"panel": "output.references"})
             window.status_message("No references found")
+            return
+
+        word = self.view.substr(self.view.word(pos))
+
+        base_dir = get_project_path(window)
+        formated_references = self._get_formated_references(response, base_dir)
```
tomv564: Spelling: "formatted"

plugin/references.py
predragnikolic: `clearcache` should't be inside of a loop.
I need to call it only once.

plugin/references.py
predragnikolic: Agree, it is ugly, and not that helpful

predragnikolic:
> I wonder if references can share some code (eg syntax) with diagnostics again?

I don't know the answer to this question. :D

Issue #405, opened 2018-09-06
Show popup for 'info' level diagnostics


tophattom:
### Question:

Is there a way to show popup dialog for diagnostics of `info` level? Currently I can only see the popup for `warning` messages. `info` diagnostics are only shown in the status bar at the bottom of the screen.

*Please include in your report:*

* OS and language server: OS X 10.11.6
* How you installed LSP (Package Control)

### Suggestion:

Maybe add a setting to control which diagnostics open a popup.

tomv564:
Good suggestions, and a good opportunity to remove some duplication around error and warning displays.

tomv564:
Added `auto_show_diagnostics_panel_level` setting to control when the panel is opened. 
Please show me if the popups on hover are unmanageable!

PR #418, opened 2018-09-30
Render all diagnostics in hovers

tomv564:
For #405

### DIFF
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -38,6 +38,14 @@ class HoverHandler(sublime_plugin.ViewEventListener):
 _test_contents = []  # type: List[str]
 
 
+class_for_severity = {
+    DiagnosticSeverity.Error: 'errors',
+    DiagnosticSeverity.Warning: 'warnings',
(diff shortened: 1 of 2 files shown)


plugin/hover.py
```diff
@@ -38,6 +38,14 @@ def on_hover(self, point, hover_zone):
...
 
 
+class_for_severity = {
+    DiagnosticSeverity.Error: 'errors',
+    DiagnosticSeverity.Warning: 'warnings',
+    DiagnosticSeverity.Information: 'info',
+    DiagnosticSeverity.Hint: 'hints'
+}
+
+
```
predragnikolic: You could use [`diagnostic_severity_names`](https://github.com/tomv564/LSP/blob/25e63ba9c7df8b92d21bc282cdf0fe1910e268a1/plugin/diagnostics.py#L23) for determining the class name?

tomv564: Good suggestion! As I'd have to move that lookup and change all the CSS names to singular I've avoided it for now.

Issue #419, opened 2018-10-04
OSX PATH issues


matt-allan:
Hello,

I was having a lot of difficulty getting LSP to to successfully start the PHP language server on one of my machines.  I reviewed the troubleshooting guide before posting this.

I have two machines with identical shell environments and sublime text configurations.  On one machine the language server starts without issue, but on the other one it fails immediately.  When this happens the console displays:

```
LSP: starting ['php', '/Users/matt/code/php-language-server/bin/php-language-server.php']
LSP:  --> initialize
LSP: window 2 added session phpls
LSP: LSP stream logger stopped.
LSP: LSP stdout process ended.
```

Eventually I was able to determine this was caused by the incorrect `php` binary being found.  I was able to fix the issue by changing `php` in the `command` configuration to the absolute path.

The strange thing is, I always start Sublime using the `subl` binary.  If I dump my path in the Sublime console (using `from os import environ; environ['PATH']`) I can see the PHP binary is in my path, and the path is identical between both machines.  I also tried installing the SublimeFixMacPath package and that did not resolve the issue.  Any idea why it would be failing to find the correct binary?

After reading through the plugin code I would have expected to see something in the logs if the process failed to start.  I ran the subprocess in the console manually and was able to recover an error message on stdout and a non zero exit code.  Apparently it's using an old version of PHP, I'm guessing the one in `/usr/bin/php` that comes with macOS.  It seems like we might be able to log the exit code and read stdout if the process quits unexpectedly?

```
>>> import subprocess
>>> from subprocess import PIPE
>>> process = subprocess.Popen(["php", "/Users/matt/code/php-language-server/bin/php-language-server.php"], stdout=PIPE, stderr=PIPE)
>>> stdout, stderr = process.communicate()
>>> stderr
b''
>>> stdout
b'\nParse error: parse error, expecting `"identifier (T_STRING)"\' in /Users/matt/code/php-language-server/bin/php-language-server.php on line 3\n'
>>> process.returncode
255
```

LSP settings:

```json
{
	"clients":
	{
		"phpls":
		{
			"command":
			[
				"php",
				"/Users/matt/code/php-language-server/bin/php-language-server.php"
			],
			"enabled": true,
			"languageId": "php",
			"scopes":
			[
				"source.php",
				"embedding.php"
			],
			"syntaxes":
			[
				"Packages/PHP/PHP.sublime-syntax"
			]
		}
	},
	"log_debug": true,
	"log_payloads": true,
	"log_server": true,
[...]

PR #420, opened 2018-10-04
Update default PHP settings to match docs

yuloh:
I was a little confused by the default settings which made it seem like tilde expansion in the `command` should work.  This PR changes the example to match the docs, which should hopefully make it clearer that you need to use the absolute path.

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -73,7 +73,7 @@
     },
     "phpls":
     {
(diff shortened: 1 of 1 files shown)


rchl:
Tilda is expanded https://github.com/tomv564/LSP/blob/ac1c858b30120b3a1f73bbb77539df65a580c907/plugin/core/clients.py#L25

yuloh:
Hm, I thought it wasn't but I guess it probably wasn't working because of the issues I described in #419.

Thanks!

PR #431, opened 2018-10-14
Add missing CRLF after JSON RPC message

ad-si:
This is not explicitly mentioned in the specification, but shouldn't it be this way?


tomv564:
You're right, it is not mentioned in the docs that I try to follow: https://github.com/Microsoft/language-server-protocol/blob/gh-pages/specification.md

The reasoning for no termination for content is likely that "content-length" provides the recipient with all it needs to read the message.

ad-si:
Yeah I guess that makes sense.
It would probably be a bad idea anyways as some JSON RPC messages contain the whole file or parts of it and this would make pretty complicated to differentiate between a \r\n embedded in the file and the \r\n of the JSON RPC message.

Issue #435, opened 2018-10-18
Exclude some folders in project for LSP


gubsky90:
Can i exclude some folders in project for LSP?
I have some protected folders in mys project, so i got:

LSP:      {'message': "UnexpectedValueException: RecursiveDirectoryIterator::__construct(/home/www-data/services/db/data/mysql): failed to open dir: Permission denied in...

Problem in php-language-server (https://github.com/felixfbecker/php-language-server)

tomv564:
The PHP language server itself is wandering into that directory? I don’t think Sublime Text or this package are asking it to go in there. Can you report at the server’s repo so they can help you?

gubsky90:
Yes, php-language-server explore all php files in rootPath (now sublime project_path). 

Please see my pull request : https://github.com/tomv564/LSP/pull/436

tomv564:
In general this package doesn't try to take on features for just one language server. Fixing the issue in PHP language server would help people using other editors as well.

I found the server's failure to handle permissions here:
https://github.com/felixfbecker/php-language-server/issues/462

PR #436, opened 2018-10-19
Exclude some folders in project for LSP #435

gubsky90:
Some language servers (like php-language-server (https://github.com/felixfbecker/php-language-server)) process all file in rootPath. If in project added protected folders, server throw fatal error.

I add sub_folder param for client config.

Config <project>.sublime-project example:
```
{
  ...
  "settings":
  {
    "LSP":
    {
      "phpls":
      {
        "enabled": true,
        "sub_folder": "repo/spygasm",
      }
    },
    ...
  }
}
```

### DIFF
--- a/plugin/core/configurations.py
+++ b/plugin/core/configurations.py
@@ -77,6 +77,7 @@ def apply_window_settings(client_config: 'ClientConfig', window: 'sublime.Window
             client_settings,
             client_env,
             overrides.get("tcp_host", client_config.tcp_host),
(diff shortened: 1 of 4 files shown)


rwols:
I think this should be gracefully handled by the server instead.

tomv564:
I like that this is a generic solution that might be useful beyond the PHP server problem it is addressing. That said, I see no other editors providing an override, and if the server is fixed then the need for this feature is gone, too.

PR #440, opened 2018-10-26
Add enabled flag to the go client configuration example

yuloh:
It took me a minute to realize I needed to add `"enabled": true` for the go language server to startup.


tomv564:
Note that the goal is not to teach everyone to edit their LSP settings - instead the package should offer to enable your golang server.

PR #450, opened 2018-11-05
Dont scroll to end when the diagnostics panel is open  

### LINKED ISSUE #446: Don't move cursor at the end when populating the diagnostics panel
When the diagnostics are populated the cursor in the panel is moved to the end. See [here](https://github.com/tomv564/LSP/blob/2869978d8b46d717da27eb0ac7a7e234840b218d/plugin/core/panels.py#L53-L56).

Is there a reason for this? 

Because I can't use the `f4` keybinding to goto the next result when it is opened.
Instead I need first to press the `shift + f4`, which is the backward direction.

Here is a simple solution.
```diff
     def run(self, edit, characters):
         self.view.replace(edit, sublime.Region(0, self.view.size()), characters)
 
-        # Move cursor to the end
+        # Clear the selection
         selection = self.view.sel()
         selection.clear()
-        selection.add(sublime.Region(self.view.size(), self.view.size()))
```

predragnikolic:
Dont scroll to end when the diagnostics panel is open, but keep the clear selection, so that panel text is not selected when opened.

Closes  #446

### DIFF
--- a/plugin/core/panels.py
+++ b/plugin/core/panels.py
@@ -50,7 +50,6 @@ class LspUpdatePanelCommand(sublime_plugin.TextCommand):
     def run(self, edit, characters):
         self.view.replace(edit, sublime.Region(0, self.view.size()), characters)
 
-        # Move cursor to the end
+        # Clear the selection
         selection = self.view.sel()
         selection.clear()
-        selection.add(sublime.Region(self.view.size(), self.view.size()))


rwols:
We'll see how it goes!

predragnikolic:
Raoul, you got merge permission :confetti_ball: congratulation :)
What could possibly go wrong.

rwols:
Well, for simple changes like these I think it's okay for me to merge ;) The serious pull requests I'll leave up to Tom.

PR #458, opened 2018-11-09
Fix: we should be able to handle a "null" result from a server

rwols:
If there's an error response, handle that.
Otherwise, handle the result.

### DIFF
--- a/plugin/core/rpc.py
+++ b/plugin/core/rpc.py
@@ -154,18 +154,16 @@ class Client(object):
         if self.settings.log_payloads:
             debug('     ' + str(result))
         handler, error_handler = self._response_handlers.pop(request_id, (None, None))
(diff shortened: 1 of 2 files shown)


rwols:
> If we continue with this logic, why did we drop the "invalid response payload" case?

the logic seemed somewhat complex:

no error key, no result key ==> invalid response
error key, no result key ==> handle error
no error key, result key ==> handle result
error key, result key ==> invalid response

I thought: If there's an error key present we should handle it regardless of the result key. Because of that, it turned out that the "invalid response" logical path went away.

> Especially since it wasn't needed to get your memory leak fix in place?

You're right, I've reverted to the old behavior and added unit tests for that behavior.

tomv564:
Thanks for the quick fix!

PR #462, opened 2018-11-12
Add ability to fold diagnostics

AmjadHD:
This PR adds the ability to fold diagnostics in the diagnostics panel and keep them folded when the panel is updated.

### DIFF
--- a/plugin/core/panels.py
+++ b/plugin/core/panels.py
@@ -1,3 +1,5 @@
+from .settings import settings
+
 import sublime
 import sublime_plugin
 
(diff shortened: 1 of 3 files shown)


plugin/core/panels.py
```diff
@@ -48,8 +49,19 @@ class LspUpdatePanelCommand(sublime_plugin.TextCommand):
...
+        folded_regions = (view.substr(view.line(fr.a)) for fr in view.unfold(sublime.Region(0, view.size())))
+        view.replace(edit, sublime.Region(0, view.size()), characters)
+
+        for fr in folded_regions:
+            # get all file lines
+            for ln in view.find_by_selector("meta.diagnostic.preamble.lsp"):
+                if view.substr(ln) == fr:
+                    # fold the region spanning from the end of the line of the
+                    # filename to the to the end of the indented body
+                    view.fold(sublime.Region(ln.b, view.indented_region(ln.b + 2).b - 1))
```
rwols: `indented_region` is an undocumented ST API function. Beware.

AmjadHD: as I have noticed by experimenting, indented_region returns the region of the bloc of code with the highest indentation level containing the point:
the point here is represented by the cursor:
screenshot_1
screenshot_2
screenshot_3

tomv564:
I love using Sublime's built-in features like folding where possible, but the logic looping over and string matching the output while using undocumented APIs - it's a big much. 
Instead of giving users this crutch, I also feel they'd be better served by configuring their diagnostics to their needs.

AmjadHD:
Although the logic seems unintuitive, it isn't that bad as the common case is 2 or 3 folded regions and 4 or 5 diagnostics entries, so that is efficient enough, as for the undocumented function, i see nothing wrong of using it as long as we know how to use it.
If you have a different implementation i'm all ears.
If users don't like this, we can make it optional by adding a setting entry such as `"fold_diagnostics: false"`.

PR #466, opened 2018-11-19
Add ability to fold diagnostics in the diagnostics panel

AmjadHD:
This PR adds the ability to fold diagnostics in the diagnostics panel

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -274,6 +274,9 @@
   // depending on available diagnostics.
   "auto_show_diagnostics_panel": true,
 
+  // enable folding diagnostics in the diagnostics panel
+  "fold_diagnostics": false,
+
   // Open the diagnostics panel automatically
   // when diagnostics level is equal to or less than:
   // error: 1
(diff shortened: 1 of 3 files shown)


AmjadHD:
Sorry, things got messed up with the old PR (#462) so I had to create a new one, I made folding optional and disabled by default.

AmjadHD:
I don't know why Travis CI Build is failing 
```
plugin/core/settings.py:71: error: "Settings" has no attribute "fold_diagnostics"
plugin/core/panels.py:53: error: "Settings" has no attribute "fold_diagnostics"
The command "mypy -p plugin" exited with 1.
```
why is mypy only complaining about the additions i made ?

tomv564:
It appears your new setting has not been added to `plugin/core/types.py` - that should be enough to make mypy happy. A setting defaulted to false is great, thanks for adding it!

Did you consider why the folds have to be manually restored for each update?
I wonder if the panel update logic can't be implemented in a way so the regions _under_ each file get replaced, keeping the folds intact.

AmjadHD:
build is still failing, it doesn't update to e6f346a

rwols:
@AmjadHD try force pushing without changing anything: `git push -f`.

AmjadHD:
it says: "everything up to date"

Issue #33, opened 2017-08-20
How to handle windows with multiple root folders?


tomv564:
The LSP package currently handles windows with multiple folders by finding a common prefix.

As discovered in #24, a common usage with Sublime workspaces is to link in whole trees of projects, perhaps even from multiple drives (e.g. `C:\Users\Tom\Projects` and `D:\Work\`)

For this scenario, it is increasingly unlikely that we can find a common prefix to give the language server as `rootUri`.

Looking at the language server protocol, the workspace concept (of which there is only one in a language server process) offers these actions:

`workspace/symbols` (would you really want to execute this across both drives ?) `workspace/didChangeConfiguration` (configuration is likely to be project-specific, if anything)

An improvement would be to create an LSP instance per `window.folders()`.

Users who choose to mount the parent folder of multiple projects will have to accept that:
* workspace symbol listing is for projects
* The LSP package will not be able to help with per-project configuration.

randy3k:
And it should also handle the case with no folders, i.e. when editing signal files.

tomv564:
This has been solved by #81 (only use first folder) and #126 (support single files)

haferburg:
This issue is still mentioned on https://github.com/tomv564/LSP#troubleshooting

tomv564:
Thanks, pointed it to issue #82 instead!

PR #472, opened 2018-12-01
Fix #471

### LINKED ISSUE #471: LS always starts in first folder of workspace
LSP always starts a language server in the first project of your workspace, regardless of which one you're working on. For example, with the following workspace:

screenshot_20181201_151846

When I open any Rust files in `bserver`, RLS is still started in `LSP`, since it appears first in the list. This causes RLS to throw a warning:

screenshot_20181201_151010

and effectively breaks all useful functionality of the LSP plugin--nothing works, because RLS is staring at the wrong directory.

I'm still digging as to why this is, but it looks like the issue is [an oversight with branching right here](https://github.com/tomv564/LSP/blob/master/plugin/core/workspace.py#L16). I'll submit a PR shortly.

scriptis:
Fixes a logic error that causes `get_project_path` to always return the first directory in the current workspace.

### DIFF
--- a/plugin/core/workspace.py
+++ b/plugin/core/workspace.py
@@ -13,7 +13,7 @@ def get_project_path(window: 'Any') -> 'Optional[str]':
     """
     Returns the first project folder or the parent folder of the active view
     """
-    if len(window.folders()):
+    if len(window.folders()) == 1:
         folder_paths = window.folders()
         return folder_paths[0]
     else:


scriptis:
While this _does_ fix the CWD of spawned language servers, LSP still seems to stop functioning entirely  when working in a non-primary directory in a workspace... I'm going to leave this open for the moment, but keep poking around to see if I can fix anything else.

plugin/core/workspace.py
```diff
@@ -13,7 +13,7 @@ def get_project_path(window: 'Any') -> 'Optional[str]':
     """
     Returns the first project folder or the parent folder of the active view
     """
-    if len(window.folders()):
+    if len(window.folders()) == 1:
```
rwols: I downloaded RLS and can reproduce your problem locally. Could you try using this instead?

```
import os
try:
    from typing import List, Optional, Any, Iterable
    assert List and Optional and Any and Iterable
except ImportError:
    pass

from .logging import debug
from .types import ViewLike


def get_filename_from_view(view: ViewLike) -> 'Optional[str]':
    if not view:
        debug("No view is active in current window")
        return None  # https://github.com/tomv564/LSP/issues/219
    filename = view.file_name()
    if not filename:
        debug("Couldn't determine project directory since no folders are open",
              "and the current file isn't saved on the disk.")
    return filename


def get_directory_name(view: ViewLike) -> 'Optional[str]':
    filename = get_filename_from_view(view)
    if filename:
        project_path = os.path.dirname(filename)
        return project_path
    return None


def find_path_among_multi_folders(folders: 'Iterable[str]',
                                  view: ViewLike) -> 'Optional[str]':
    filename = get_filename_from_view(view)
    if not filename:
        return None
    folders = [os.path.realpath(f) for f in folders]
    file = view.file_name()
    if not file:
        return None
    file = os.path.realpath(file)
    while file not in folders:
        file = os.path.dirname(file)
        if os.path.dirname(file) == file:
            # We're at the root of the filesystem.
            file = None
            break
    debug('project path is', file)
    return file


def get_project_path(window: 'Any') -> 'Optional[str]':
    """
    Returns the project folder or the parent folder of the active view
    """
    if not window:
        return None
    num_folders = len(window.folders())
    if num_folders == 0:
        return get_directory_name(window.active_view())
    elif num_folders == 1:
        folder_paths = window.folders()
        return folder_paths[0]
    else:  # num_folders > 1
        return find_path_among_multi_folders(
            window.folders(),
            window.active_view())
```

also we should really get some unit tests going for get_project_path

scriptis: This works perfectly! I'll throw together some tests and update this.

rwols: There is a MockWindow class sitting in test_windows.py that you can use

rwols: just run `tox -q` on the command-line to run all tests (except for the unittesting plugin suite)

rwols:
What's the progress on this?

scriptis:
Haven't had time to work on this the past week, should have it sorted today. Sorry about the wait.

scriptis:
Look good? @rwols

tomv564:
Thanks for the contribution, but I will revert this change now in favour of a full solution.

The existing behaviour was not a "logic error", the function even documented that it would return the first folder. (Also see the note in the [troubleshooting section](https://github.com/tomv564/LSP/blob/master/README.md#troubleshooting) and issue #33 linked from it)

I'm sure we can make multi-folder projects work *reliably*, but if we need to look closer at the usages of `get_project_path` - what further changes are needed so `get_project_path` is always correct?

tomv564:
LSP should probably start communicating the current limitation by warning users and refusing server startup when files under the 2nd folder are opened.

scriptis:
@tomv564: I don't follow your reasoning in reverting this. For workspaces with more than one active project,  [this evaluates to true](https://github.com/tomv564/LSP/blob/master/plugin/core/workspace.py#L16), always returning the first directory of the workspace, even though the function is documented to return `the parent folder of the active view`. As far as I can tell, this PR made the function *conform* to the previous documentation, not stray from it?

tomv564:
The `parent folder of the active view` part of the documentation/function is intended as a fallback, it probably should have been worded as such including pointing out it is a filesystem directory it wants to return, not a sublime folder.

But perhaps this is a good time to look at supporting multi-directory workspaces:

Say you have a window with two python projects:
```
/project1
/project2
```

**Server sessions are shared per window**

* You open `/project1/main.py` - the python language server starts with rootUri `file:///project1`
* You then open `/project2/main.py` - the window already has a python session, so it re-uses that.
* LSP sends didOpen `project2/main.py` on a session based on `project1`.

This is fixable with the cost of looking up the session by window root folder.

**`get_project_path` uses `window.active_view()`**

* Request references in `/project1/main.py`
* You switch active view to `/project2/main.py`
* References response arrives, calls `get_project_path` and receives `/project2` instead of `/project1`, breaking the relative path lookup

**Support configuration per folder?**

A follow-up consideration: If users can have different sessions for `/project1` and `/project2` in the same window,  should they not be able to configure the language server differently for each folder?

PR #474, opened 2018-12-03
Allow user to trigger custom commands

ayoub-benali:
This PR allows the user to send custom commands to the LSP server as `workspace/executeCommand` This offers extensibility to the LSP client as Servers can always have custom "endpoints"

Commands can be added either 
 - via custom sublime commands, see [documentation](http://docs.sublimetext.info/en/latest/reference/command_palette.html) and later invoked via palette.

- as key bindings like:
```json
{"keys": ["f10"], "command": "lsp_execute", "args":{"command_name": "build-import", "command_args":{}}}
```


plugin/configuration.py
```diff
@@ -73,6 +74,33 @@ def _on_done(self, index):
             self.window.status_message("{} enabled, starting server...".format(config_name))
 
 
+class LspExecuteCommand(sublime_plugin.WindowCommand):
```
rwols: You might want to inherit from LspTextCommand (located in plugin/core/registry.py). This handles disabling the command if the view does not support LSP.

rwols:
Can you give some examples where this used?

ayoub-benali:
> Can you give some examples where this used?

Here is an example were it would be used https://scalameta.org/metals/docs/editors/new-editor.html#import-build
Metals is a new LSP server for Scala, @olafurpg and @gabro can probably provide more examples

Commands/Default.sublime-commands
```diff
@@ -59,5 +59,9 @@
     {
         "caption": "LSP: Rename Symbol",
         "command": "lsp_symbol_rename"
+    },
+    {
+        "caption": "LSP: Execute command",
```
olafurpg: Would it be possible for users to register custom commands via JSON configuration? Instead of running "LSP: Execute command" and manually type "build-import" (making sure there are no typos) it would good to additionally support the following workflow

1. register commands in LSP client settings 
```json
{
  "Metals": {
    "commands": [
      { "command": "build-import", "arguments": [] }
    ]
  }
}
```
2. run the "Metals: Import build" command directly without quick panel prompt

gabro:
Woops, I misread the scope of this PR. I thought it was about allowing users to define custom commands.

What I said above is then not directly relevant to this, although it would be a cool feature indeed!

Commands/Default.sublime-commands
ayoub-benali: The example on how to register such commands is in the PR description. Users don't have to type manually "build-import", they have to add it to the client config once. 
Next time they can select the commands that they registered.

Once Metal is stable enough we can add a default client + config to this project, this way users don't event have to configure the client, it comes with the plugin and with a set of predefined commands.

olafurpg: I'm sorry I totally missed that part! Some more thoughts

- It would be good to allow a more readable title for the command "Metals: Import build" instead of "build-import".
- (not a big deal) It would be nice if we could avoid the quick panel and allow users to run "Build import" directly as a unique command.
- Note that `workspace/executeCommand` supports `CommandParams.arguments`, we don't use it in Metals but it's part of the spec.

ayoub-benali: make sense. I will add labels for the commands and allow arguments support

rwols: @ayoub-benali also consider using the more fancy sublime_plugin.ListInputHandler and sublime_plugin.TextInputHandler classes (see plugin/rename.py for an example usage).

plugin/configuration.py
```diff
@@ -73,6 +74,33 @@ def _on_done(self, index):
             self.window.status_message("{} enabled, starting server...".format(config_name))
 
 
+class LspExecuteCommand(LspTextCommand):
```
predragnikolic: Maybe this class should live is own file?
I would suggest creating a `plugin/execute_command.py` file.

Commands/Default.sublime-commands
ayoub-benali: @rwols and the proper way to test this would be through sublime testing like `test_edit.py` ?

tests/test_execute_command.py
```diff
@@ -0,0 +1,41 @@
...
+
+        session = Session(text_config, dirname(__file__), MockClient())
+        session.state = ClientStates.READY
+        wm._sessions[text_config.name] = session
+
+        self.view.run_command('lsp_execute')
+
+        # popup should be visible eventually
+        yield self.view.is_popup_visible()
+        self.view.update_popup("command1")
```
ayoub-benali: @rwols Do you know how to select an element from the popup and hitting "enter" via the sublime api or the testing framework ?
Otherwise we can't test that the client is receiving the command as expected ?

rwols: Good question. I don't know the answer to that. Maybe @randy3k can clarify. How do we check the content of a quick panel or select an item from the quick panel? Is this even possible?

plugin/execute_command.py
```diff
@@ -0,0 +1,36 @@
...
+        pass
+
+    def _on_done(self, index: int) -> None:
+        if index > -1:
+            command = self._commands[index]
+            client = client_for_view(self.view)
+            if client:
+                request = {
+                    "command": command
+                }
```
rwols: How are you planning to use string arguments?

ayoub-benali: yes I am planning to forward the command args as well

ayoub-benali: BTW according to the spec the command type is a string
https://microsoft.github.io/language-server-protocol/specification#workspace_executeCommand

rwols: > BTW according to the spec the command type is a string

I'm not sure I totally follow you here. The `arguments` is an `any`, so can potentially be a list or something. My question was: how do we know if the command should supply a string argument, or a null argument? I think it's good to restrict us to no arguments and a single string argument for now.

ayoub-benali: ouuhh sorry totally miss read your initial question. 

> I'm not sure I totally follow you here. The `arguments` is an `any`, so can potentially be a list or something. My question was: how do we know if the command should supply a string argument, or a null argument? I think it's good to restrict us to no arguments and a single string argument for now.

I am fine with treating the arguments for now as an optional single one of type string. But assuming we want to be more flexible, is there a problem to read the arguments as Dict[String, Any] ? 

@olafurpg is any of the current/future Metals commands need arguments ?

plugin/execute_command.py
```diff
@@ -0,0 +1,36 @@
...
+    def run(self, lsp_command) -> None:
+        self._commands = []   # type: List[str]
+        for config in client_configs.all:
+            if config.commands:
+                self._commands.extend(config.commands)
+
+        if len(self._commands) > 0:
+            self.view.window().show_quick_panel(self._commands, self._on_done)
+
+    def _handle_response(self, response: 'Optional[Any]') -> None:
```
ayoub-benali: @rwols given the response can be `Any` what do you think should be the way to format and display the response?
Should we update the diagnostic panel ?

rwols: What does metals return? Have you tried printing it to the status bar (`sublime.status_message`) and/or the console (`print`)?

Commands/Default.sublime-commands
ayoub-benali: @rwols since I moved "class LspExecuteCommand" to it own file `plugin/execute_command.py` the command doesn't appear in the command pallet. Is there some mapping I need to do? `lsp_execute` -> `LspExecuteCommand`

plugin/execute_command.py
ayoub-benali: From the doc, Metals commands don't take arguments for now
https://scalameta.org/metals/docs/editors/new-editor.html#metals-server-commands

plugin/execute_command.py
ayoub-benali: response is empty so far. I will display a popop in case of an error and probably just log the response for now

Commands/Default.sublime-commands
ayoub-benali: editing "boot.py" fixed the issue

tests/test_execute_command.py
randy3k: If what you what is just to hit the enter, you could change the content of the quick panel (every panel is a regular view) and then execute `window.run_command("select")`.

If you need more sophisticated interactions, you might need to use https://github.com/randy3k/Keypress. See [this](https://github.com/randy3k/Keypress/blob/master/tests/test.py) for an example.

ayoub-benali: @randy3k the command (TextCommand) that I am working on shows a pannel a require user interaction, so to test it properly Iwould need to:
- check the content of the quick pannel that is generated
- select one of the elements
- press enter 

Do you know how to do that ?

plugin/execute_command.py
olafurpg: I don't have any planned commands in mind for Metals that will need arguments or return meaningful responses (they all return `{}`).

plugin/core/settings.py
```diff
@@ -149,9 +149,19 @@ def read_language_configs(client_config: dict) -> 'List[LanguageConfig]':
     return list(map(read_language_config, client_config.get("languages", [])))
 
 
+def read_commands(commands: 'List[Dict[str, Any]]') -> 'List[Command]':
+    validCommands = []
```
rwols: style nitpick: please use snake_case variable names (so `valid_commands` instead of `validCommands`)

plugin/execute_command.py
```diff
@@ -0,0 +1,56 @@
...
+            self._command_names = []   # type: List[str]
+            self._command_args = dict()   # type: Dict[str, Dict[str, Any]]
+            for config in client_configs.all:
+                for command in config.commands:
+                    self._command_names.append(command.name)
+                    self._command_args[command.name] = command.args
+
+            # TODO handle when command name is passed
+            if len(self._command_names) > 0:
+                self.view.window().show_quick_panel(self._command_names, lambda i: self._on_done(client, i))
```
rwols: I still highly recommend using a `sublime_plugin.ListInputHandler` instead of `show_quick_panel`.

plugin/execute_command.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class LspExecuteCommand(LspTextCommand):
+    def __init__(self, view):
+        super().__init__(view)
+
+    def run(self, command_name) -> None:
+        client = client_for_view(self.view)
+        if not client:
+            self.view.show_popup("No connection with server")
```
rwols: Just return silently here. There should always be a client available. This command is not visible from the command palette if there is no ClientConfig present for the view.

plugin/execute_command.py
```diff
@@ -0,0 +1,56 @@
...
+        else:
```
rwols: When silently returning from the above `if not client:`, you can omit the `else` clause here.

plugin/execute_command.py
```diff
@@ -0,0 +1,56 @@
...
+            self._command_names = []   # type: List[str]
+            self._command_args = dict()   # type: Dict[str, Dict[str, Any]]
+            for config in client_configs.all:
+                for command in config.commands:
+                    self._command_names.append(command.name)
+                    self._command_args[command.name] = command.args
+
+            # TODO handle when command name is passed
```
rwols: Can you clarify what this means?

plugin/execute_command.py
ayoub-benali: ok but do you know how to load the config only for clients that are Initialized ?
If I have client A and B in my settings but only A I have a current session, how can I get only configs of A ?

plugin/execute_command.py
ayoub-benali: to my understanding the run method can take extra arguments, I want allowing the user to create a key binding to this sublime command directly (without having the open the quick pallet) and pass directly the LSP command name and maybe the args like 
```
  {"keys": ["f10"], "command": "lsp_execute", "args":{"command_name": "build-import"}}
```
So far this leads to 
```
  File "/opt/sublime_text/sublime_plugin.py", line 1067, in run_
    return self.run(edit, **args)
TypeError: run() got multiple values for argument 'command_name'
```

plugin/execute_command.py
ayoub-benali: no problem but since the run method returns None, how to link the `LspTextCommand` to to an instance of `ListInputHandler` ?

plugin/execute_command.py
rwols: Okay I see. The signature of `run` should I think be

```
def run(self, edit, command_name: str) -> None:
   ...
```
Because it is a sublime_plugin.TextCommand. These types of commands receive an `edit` object with which one can edit the view (we don't use it here though).

ayoub-benali: tried it already and it get `plugin/execute_command.py:19: error: Signature of "run" incompatible with supertype "TextCommand"` with mypy

using `def run(self, edit, **command_name) -> None:` works but now is a Dict[str, Any] which defeats the purpose :(

ayoub-benali: extra arguments need default value so this did the trick
```
    def run(self, edit, command_name: 'Optional[str]'=None, command_args: 'Dict[str, Any]'=dict()) -> None:
```

plugin/execute_command.py
rwols: > how to link the LspTextCommand to to an instance of ListInputHandler

You can define a new method for LspExecuteCommand:
```
    def input(self, args):
        return None if 'command_name' in args else CommandNameInputHandler(self.view)
```

where CommandNameInputHandler would be something like this:

```
class CommandNameInputHandler(sublime_plugin.ListInputHandler):
    def __init__(self, view):
        super(CommandNameInputHandler, self).__init__()
        self.view = view
        self.commands = []  # ... get a list of all the commands

    def list_items(self):
        return self.commands

    def placeholder(self):
        return self.commands[0]
```

ayoub-benali:
```
if type(command_name) is list:
                name = command_name[0]
                args = command_name[1] if len(command_name) == 2 else dict()
                self._send_command(client, name, args)
            elif type(command_name) is str:
                self._send_command(client, command_name, command_args)
```

This is the workaround that I found but had to removed type hints from the run method signature otherwise we get this from mypy
```
plugin/execute_command.py:48: error: Argument 3 to "_send_command" of "LspExecuteCommand" has incompatible type "Collection[str]"; expected "Dict[str, Any]"
```

not sure this is better than using plain `show_quick_panel`

tests/test_execute_command.py
randy3k: You need to get the view if of the panel, change the content of the view and execute the command `select`.

tests/test_execute_command.py
```diff
@@ -0,0 +1,51 @@
...
+        # TODO: check the content of the pallet, select one element and hit enter
+
+    def test_execute_command(self):
+        wm = windows.lookup(self.view.window())
+        wm._configs.all.append(text_config)
+
+        session = Session(text_config, dirname(__file__), MockClient())
+        session.state = ClientStates.READY
+        wm._sessions[text_config.name] = session
+        self.view.run_command("lsp_execute_command", {"command_name": "command1"})
```
ayoub-benali: @rwols I am seeing "no config found or enabled for view" in the console when running the test. Do you know why the config is `None`? even though  I do like the other tests  
```
wm = windows.lookup(self.view.window())
wm._configs.all.append(text_config)
```

tomv564:
Great solution for cheaply adding fire-and-forget commands. 

Does the LSP protocol not provide a way (other than code actions) for exposing commands?
The concern with this solution is having to update everyone's configs if your server decides to deprecate/rename/add arguments to a command. 
(For the "legacy config" reason I've been trying to _reduce_ the amount of content in client config)

Also, with this feature we should take the chance to decide what it is _not_ for. For example, if more client logic needs to be attached to the server response, writing a LanguageHandler in another Sublime package would be preferred. Can we anticipate users wanting dialog boxes, etc?

Big thumbs up for adding tests!

olafurpg:
Good job everyone with the detailed review and tests 👍 Looking forward to give this a try :) 

>  For example, if more client logic needs to be attached to the server response, writing a LanguageHandler in another Sublime package would be preferred.

I agree, and this would be consistent with how LSP editor support works in vscode/atom/vim/emacs. 

> Does the LSP protocol not provide a way (other than code actions) for exposing commands?

Commands are used in other places as well, for example `CompletionItem.command`

```ts
/**
	 * An optional command that is executed *after* inserting this completion. *Note* that
	 * additional modifications to the current document should be described with the
	 * additionalTextEdits-property.
	 */
	command?: Command;
```

tests/test_execute_command.py
tomv564: These tests can be a bit painful to work on as they are sensitive to state changes from other tests and your current editing work. I recommend re-opening Sublime and running only 1 test case at a time to keep things sane. Also, if you make implementation changes, reloading package from the sublime palette may be needed.

"no config found or enabled for view" might be a red herring (eg. from `new_file()` in setUp)  - adding additional logging is unfortunately the best advice I can give.
I've never tried starting a debugger on the sublime python scripts, maybe that works for you?

ayoub-benali: I am running this test file alone, so it shouldn't be related to previous state. 
for some reason [score selector](https://github.com/tomv564/LSP/blob/master/plugin/core/configurations.py#L42) is returning 0 even though I open `/testfile.txt` via the view and I register a config for txt files `LanguageConfig("test", ["text.plain"], ["Plain text"])`

tomv564:
After looking at your editor integration documentation (https://scalameta.org/metals/docs/editors/new-editor.html#import-build), I think your users will eventually be better served by a LanguageHandler implementing your protocol extensions.

Alternatively, could we perhaps start by only implementing the support for invoking commands from keybindings and .sublime-commands files? It would save us both time and effort with typing, testing (and my concern for extending ClientConfig)

ayoub-benali:
> I think your users will eventually be better served by a LanguageHandler implementing your protocol extensions.

Can you please elaborate ? would that be done in this project or a separate package ? if so how would it interact with LSP package ?

>Alternatively, could we perhaps start by only implementing the support for invoking commands from keybindings and .sublime-commands files

That is fine as the result is the same for user experience (key binding + pallet), I will update the PR to remove the config update. 

I would just need your help debugging the weird test case, I am already several days on it...

tomv564:
If you create a sublime text package that implements LSP's LanguageHandler, you will get callbacks to integrate with LSP. An example here:
https://github.com/tomv564/LSP-rust/blob/master/plugin.py#L36

I'll try to checkout this branch and run your test this week sometime!

ayoub-benali:
I updated the PR to limit the feature to just sending commands and no client config. Thanks for the help with the test case

tomv564:
So, I tested this out with your language server and found the following:

* I had to add `sbt-bloop` manually to my new project to make `sbt bloopInstall` work
  * Is that supposed to be available or automated for me? I didn't see it in your instructions?

After setting up this in a Default.sublime-commands:
```
    {
        "caption": "LSP: Execute command",
        "command": "lsp_execute",
        "args": {
            "command_name": "build-import"
        }
    }
```

I ran the command and noticed the only feedback was in the console log.
Would be great if the user can see progress in Sublime while it's running?
If running sbt directly works, then Sublime's "exec" command will get you a free output panel
```python
        self.window.run_command(
            "exec", {
                "cmd": ["sbt", "-Dsbt.log.noformat=true", "bloopInstall"],
                "working_dir": sublime.active_window().folders()[0]
            })
```

image

Regarding the server announcing available commands, I noticed your server sends the following:
```
'executeCommandProvider': {'commands': ['build-import', 'build-connect', 'sources-scan', 'doctor-run']}
```
Could be something LSP reads and offers to blindly execute without args.

Regarding the server, diagnostics updated (but only after saving the file) and go to definition worked as well. Great work!

I had previously written some code to make dotty integrate with LSP, here's a repro with some of the code updated for metals: https://github.com/tomv564/lsp-metals

tests/test_execute_command.py
tomv564: The command is not invoked your test because Sublime's command naming is a bit tricky. Try
`lsp_execute` instead of `lsp_execute_command`!

Much of the test setup should work like test_hover.py

tomv564: Forgot to add, client responses are hardcoded in MockClient (see test_session.py). You'll want to add:
```'workspace/executeCommand': {}``` to its `responses` dict.

**Comment by ayoub-benali** [comment]
> I had to add sbt-bloop manually to my new project to make sbt bloopInstall work
Is that supposed to be available or automated for me? I didn't see it in your instructions?

I my case I just followed the [insctructions](https://scalameta.org/metals/docs/editors/sublime.html) for sublime. Basically you just need to install Metals. Bloop get installed automatically by Metals.

> I ran the command and noticed the only feedback was in the console log.

This is due to the default Metals behavior with Sublime. From the known issues [section](https://scalameta.org/metals/docs/editors/sublime.html#known-issues):
`The Sublime Text client uses an invasive alert window for window/showMessage, so Metals uses window/logMessage instead.`
Maybe this should change @olafurpg ? for me the diagnostic panel is better suited to show this kind of messages than a message dialog, which `window/showMessage` is currently doing.

> Would be great if the user can see progress in Sublime while it's running?

we can't do that due to reason above.

> Could be something LSP reads and offers to blindly execute without args.

I agree, we can save this somehow and make the commands available.

> I had previously written some code to make dotty integrate with LSP, here's a repro with some of the code updated for metals: https://github.com/tomv564/lsp-metals

Thanks for the repo ! @olafurpg I think if we want to provide an equal experience to VS code plugin, we should create metals package for sublime leveraging LSP package and having so nice defaults. Could that be hosted at https://github.com/scalameta ?

**Comment by olafurpg** [comment]
@tomv564 Thank you for the feedback and trying it out! 😊 

> I had to add sbt-bloop manually to my new project to make sbt bloopInstall work 

Did you execute the "Import build" command?  It should not be necessary to manually add sbt plugins, Metals "Import build" takes care of installing sbt-bloop and running `sbt bloopInstall` for you.

The first time Metals encounters a new sbt build it sends a `window/showMessageRequest` to ask for permission to import the build. The `-Dmetals.client=sublime` setting disables `showMessageRequest` for Sublime since the endpoint was not implemented when I first tried it. If the Sublime LSP package now supports `showMessageRequest` I can update `-Dmetals.client=sublime` to enable `showMessageRequest` as normal.

>  Would be great if the user can see progress in Sublime while it's running?

That would be great indeed! Metals offers two ways to show background progress

- `window/logMessage`: this should match the output channel for `sbt bloopInstall` in your screenshot (along with additional Metals logs)
- `metals/status`: opt-in extension via `-Dmetals.status-bar=on` to place text contents in the editor status bar which should always be visible to the user. Docs https://scalameta.org/metals/docs/editors/new-editor.html#metals-status Example output: imageedit_3_3576623823


> Could be something LSP reads and offers to blindly execute without args.

Yes, I think that would be a great. Although, I wish LSP had a way to attach readable titles and descriptions the commands 😏 

> Regarding the server, diagnostics updated (but only after saving the file) and go to definition worked as well. Great work!

Thanks! It's expected diagnostics are only triggered after file save, I actually like it that way so I'm not distracted by red squiggles as I'm typing incomplete programs.

>  I think if we want to provide an equal experience to VS code plugin, we should create metals package for sublime leveraging LSP package and having so nice defaults. Could that be hosted at https://github.com/scalameta ?

@ayoub-benali I agree that would be a good idea. If I create the repository, would you be interested in helping maintain it?

File: tests/test_execute_command.py (hunk 12 above)
Comment (ayoub-benali) [reply tests/test_execute_command.py]: @tomv564  I copied the same code as in `test_hover.py` and still `no config found or enabled for view LSP/tests/testfile.txt` :(
I have no clue why `score` in `get_scope_client_config` is zero even though the client config is found. 

Did it work for you ?

File: tests/test_execute_command.py [hunk 13]
```diff
@@ -0,0 +1,58 @@
+from unittesting import DeferrableTestCase
+import sublime
+from os.path import dirname
+from LSP.plugin.core.types import ClientConfig, ClientStates, LanguageConfig
+from LSP.plugin.core.test_session import MockClient
+from LSP.plugin.core.sessions import Session
+from LSP.plugin.core.registry import windows  # , session_for_view
+from LSP.plugin.core.settings import client_configs
+
+test_file_path = __file__
```
ayoub-benali: @tomv564 I got the tests to work but only with a python file using the `testfile.txt` with scope `text.plain` didn't work for the client.
See: https://github.com/tomv564/LSP/pull/474#discussion_r241889838
While debugging the issue I discovered that the file content was somehow 0 (using view.size()) for sublime, I don't why, but opening a python file solved the issue and selector score was > 0

rwols:
How are we moving forward?

tomv564:
I merged the minimal functionality as with the use of `message_dialog` I am pretty sure the end-to-end tests would no longer work. Thanks a ton for the effort with the tests, though!

You'll have to add a keybinding / command in your user config to invoke `lsp_execute_command` with this implementation.

`showMessageRequest` was recently added, perhaps it's not released yet?

tomv564:
@ayoub-benali @olafurpg release 0.8.0 is now out containing both your features!

PR #489, opened 2018-12-24
Add a bulb

### LINKED ISSUE #485: Visual indication if code actions exist
VS code uses a bulb to indicate if code actions exist.

2018-12-15-200218_683x384_scrot
 
Currently LSP doesn't show any visual indication that code action exist.
But here is what I come up with:

output

What do you think about that?

predragnikolic:
closes #485 

### Changed

Code action request is send:
* when the cursor stops moving (it is debounced),
* after didChange notification, 

Each time a request is send, the response is cached.
The code action request is not sent when the `LspCodeActionsCommand` is run (it uses the cache to get the commands),  unless you pass a `make_request` flag.
This flag is used in hover popups for errors, warnings, hints...  When the user clicks `code actions` in hovers, behind the scenes, the caret get placed to a new position (see [here](https://github.com/tomv564/LSP/blob/6111498aca20c6b85808eebda9348335fd598678/plugin/hover.py#L167-L175)) and based on the new position we need to fire a new request, so the user gets the right code actions.

Couple of questions for you.
Should the bulb be behind a flag, maybe some people don't like it?
For people who use GitGutter, they need to update `protected_regions` to include `lsp_bulb`. 
If this PR gets through I can make a suggestion/PR to GitGutter to include it.


plugin/code_actions.py
```diff
@@ -1,3 +1,5 @@
+import sublime_plugin
+from sublime import Region
```
predragnikolic: Delete this

predragnikolic:
This definitely needs more work. There are some problems I haven't noticed at first. 
I will close this and reopen it when I think it is ready.

predragnikolic:
I fixed some problems I had with servers who don't support code actions and some edge cases I found with the mouse. The final result look like this.

output

predragnikolic:
I hope there are no big problems. But If you do notice something unexpected strange please say.
I think this gives a better user experience overall.

predragnikolic:
How can I make mypy happy? Except using `# type: ignore`
```
plugin/code_actions.py:38: error: Argument 1 to "_handle_response" of "CodeAction" has incompatible type "Optional[Any]"; expected "List[Dict[Any, Any]]"
```

tomv564:
Nice demo, excited to get this in. Do you have time for some investigations to simplify and isolate this feature a bit?

In my testing, `on_selection_modified_async` was fired for all editing commands. Will it work if we don't listen to `didChange` but instead look up code actions _after_ the 500ms wait for `purge_did_change`? 

Can we leave LspCodeActionsCommand and `hover.py` unchanged? (Your caching idea is thoughtful but not a necessary optimisation for the first implementation, I think?).

>Should the bulb be behind a flag, maybe some people don't like it?

Yes :) I anticipate we'll enable it by default in the future though!

predragnikolic:
I think this is ready for a final review. 
Or what people in the JS world woud say... reached stage 3 :)

plugin/core/helpers.py
```diff
@@ -0,0 +1,17 @@
+from threading import Timer
```
rwols: Did you try using `sublime.set_timeout_async` instead of a `threading.Timer`?

According to the [docs](https://docs.python.org/3/library/threading.html#timer-objects), a Timer is a subclass of a Thread. So as it stands now a thread is constantly created and destroyed whenever the view changes. This seems wasteful to me. While on the other hand the `sublime.set_timeout_async` function runs a function after a timeout on a dedicated "async" thread (from what I can understand from the ST api). This "async" thread keeps on living forever and various tiny functions may be run on it.

plugin/code_actions.py
```diff
@@ -1,56 +1,124 @@
...
+            else:
+                code_action.hide_bulb()
+
+        if callback is not None:
+            callback(response)
+
+    def show_bulb(self) -> None:
+        region = self.view.sel()[0]
+        flags = sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE
+        self.view.add_regions('lsp_bulb', [region], 'markup.changed', 'Packages/LSP/icons/lightbulb.png', flags)
```
rwols: This is pretty cool :)

plugin/code_actions.py
```diff
@@ -1,56 +1,124 @@
...
+
+
+class LspCodeActionsCommand(LspTextCommand):
+    def is_enabled(self):
+        return self.has_client_with_capability('codeActionProvider')
+
+    def run(self, edit):
+        self.commands = []  # type: List[Dict]
+
+        self.code_action = CodeAction(self.view)
```
rwols: I'm not entirely sure why the `CodeAction` class should exist. Why is it a bad idea to merge the CodeAction class and this LspCodeActionsCommand class?

plugin/core/helpers.py
predragnikolic: I will try it and let you know how it turns out :)

plugin/code_actions.py
predragnikolic: It is not a bad idea. 

But first let me tell you why it first exist. When I started this, my initial idea was to have a class CodeAction that is responsible for the state and other things related to code actions. That way I have one place to send the request, cache data, etc. 
CodeAction class would be the source of truth. And other classes that inherit sublime_plugin ViewEventListener, TextCommand ... could just ask CodeActions what they need and do their thing based on the data they received :D

Well that was the initial idea. But I have nothing against simplicity, so now that I have no caching, I see no need to have a CodeAction class :)

plugin/code_actions.py
predragnikolic: :+1:  :D

plugin/code_actions.py
tomv564: @rwols I gave the feedback to keep `LspCodeActionsCommand` unchanged as it makes it easier to add feature-flag and reason about the bulb behaviour separately. We can always merge the two later.

plugin/code_actions.py
```diff
@@ -1,56 +1,124 @@
...
+    def is_applicable(cls, _settings):
+        if settings.show_bulb:
+            return True
+        return False
+
+    def on_selection_modified_async(self):
+        self.fire_request()
+
+    def on_modified_async(self):
+        self.handle_modified_async()
```
tomv564: You explained this trigger with:
> It is needed to check if the code actions are valid after we made some changes.

What if we hide the bulb whenever the selection_modified_async fires?

plugin/code_actions.py
rwols: Sure!

LSP.sublime-settings
```diff
@@ -325,6 +325,9 @@
   // Valid values are "bookmark", "circle", "cross", "dot" or ""
   "diagnostics_gutter_marker": "dot",
 
+  // Show a bulb in the gutter when code actions are available
+  "show_bulb": false,
```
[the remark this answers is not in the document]
predragnikolic: Done

plugin/code_actions.py
predragnikolic: Yes, great suggestion.
I simplified that a bit, it this better 5079c9b?

plugin/core/helpers.py
predragnikolic: Well I expected that the `set_time_out_async` return something like a **`timer`**, but it returns `None`. And the sublime api doesn't have a `clear_time_out_async` function that would allow me to cancel that given **`timer`**. 

So I don't seem to know how to implement this with a `set_time_out_async`. 
https://github.com/tomv564/LSP/blob/ba832ee73dcf5d26cef8e64193a71c2a87dcfc5c/plugin/core/helpers.py#L11

rwols: Right, there's no way to cancel a timer with the ST api unfortunately.

predragnikolic:
I had to push it :D for a better user experience

The code actions debounces  the requests every 0.5s `on_selection_modified_async` (witch feels snappy),
Conclusion: 0.8s for `on_selection_modified_async` feels slow.

and when a `on_modified_async` than requests are debounced every 0.8s (the slight delay is not noticeable when editing).
Conclusion: 0.8s for `on_modified_async` doesn't feels slow.

I can revert the last commit if you don't agree with this

tomv564:
@predragnikolic can you explain how the user experience is improved?

You have one debounced function calling another debounced function here, is that intentional? Feels like this creates a lot of combinations of timings with different results?

predragnikolic:
> Not creating CodeAction objects on every cursor move - perhaps it can be merged with the ViewEventListener, and named CodeActionBulbListener or something.

Yes I will do that. 

> Using sublime's timeouts instead of creating threads (do we really need to cancel timeouts?)

As far as this is concerned, I still have no idea how I would do it differently. There is probably a way to delay a function call by not using timers. I will see how the highlight request is send.  

But to be honest I plan just to relax for a couple of days :)

PR #491, opened 2018-12-29
Set jump back history

### LINKED ISSUE #480: LSP needs a LspJumpBackCommand
When executing LSP goto definition, 
and then executing the built in `jump_back` command, 
the cursor won't be placed in the previous place, as I expect.

To fix this, we can introduce a `LspJumpBackCommand`.
We can do the same thing as TernForSublime [did](https://github.com/ternjs/tern_for_sublime/blob/91a27a39b1b0a33a9043aa685e1ee48c64a58274/tern.py#L564).

predragnikolic:
closes #480

### DIFF
--- a/plugin/definition.py
+++ b/plugin/definition.py
@@ -5,6 +5,7 @@ from .core.protocol import Request, Point
 from .core.documents import get_document_position, get_position, is_at_word
 from .core.url import uri_to_filename
 from .core.logging import debug
+from Default.history_list import get_jump_history_for_view
 try:
     from typing import List, Dict, Optional, Any
     assert List and Dict and Optional and Any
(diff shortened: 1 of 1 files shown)


predragnikolic:
I used ` # type: ignore ` to make Travis happy. 
> To silence a single missing import error, add a # type: ignore at the end of the line containing the import. (source: https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports. )

tomv564:
Thanks guys, I added the needed stub so the `#type: ignore` statement could be removed in 
https://github.com/tomv564/LSP/commit/cc1ad460b044091aed8c6bdb97a538db9c4fdcee

PR #493, opened 2019-01-13
Merege CodeAction class to LspCodeActionBulbListener

predragnikolic:
I merged the CodeAction class and changed the behavior that @rwols reported:
> the bulb behavior is kind of weird with clangd
the bulb doesn't show up until I click on "Code Actions..." in the hover popup of a diagnostic

Based on what others editors do, the bulb is tied to the cursor position, not the mouse. 
So I chose not to show the bulb when code actions are triggered from the popup menu.

Before - the bulb **will** show up when you click on the `code actions` link
before

After  - the bulb **will not** show up when you click on the `code actions` link
after

If you have other suggestions, I am listening? :)


plugin/code_actions.py
```diff
@@ -12,37 +12,64 @@
...
+            "uri": filename_to_uri(view.file_name())
+        },
+        "range": region_to_range(view, region).to_lsp(),
+        "context": {
+            "diagnostics": list(diagnostic.to_lsp() for diagnostic in point_diagnostics)
+        }
+    }
+    session.client.send_request(
+        Request.codeAction(params),
+        lambda response: on_response_recieved(response))
```
tomv564: This turned out to be a really nice cleanup!

plugin/code_actions.py
```diff
@@ -12,37 +12,64 @@
...
+        self.hide_bulb()
+        self.fire_request()
+
+    def fire_request(self):
+        current_point = self.view.sel()[0].begin()
+        if self._stored_point != current_point:
+            self._stored_point = current_point
+            sublime.set_timeout_async(lambda: self._purge(current_point), 800)
+
+    def _purge(self, current_point: int) -> None:
```
tomv564: I used the word `purge` for the document sync logic as it empties a queue of updates - here we only have a single value instead of a list. Perhaps we can call these two `schedule_request` and `fire_request` ?

plugin/code_actions.py
predragnikolic: Thanks :D

PR #495, opened 2019-01-13
Fix: a language server may send null for the completion items

### LINKED ISSUE #494: TypeError: 'NoneType' object is not iterable - golsp completion
Output from console with logging enabled:
```
LSP:  --> textDocument/completion
LSP:      {'isIncomplete': False, 'items': None}
Error handling server payload
  File "/Users/margus/Library/Application Support/Sublime Text 3/Packages/LSP/plugin/core/rpc.py", line 141, in receive_payload
  File "/Users/margus/Library/Application Support/Sublime Text 3/Packages/LSP/plugin/core/rpc.py", line 160, in response_handler
    handler(response["result"])
  File "/Users/margus/Library/Application Support/Sublime Text 3/Packages/LSP/plugin/completion.py", line 310, in handle_response
    items = sorted(items, key=lambda item: item.get("sortText") or item["label"])
TypeError: 'NoneType' object is not iterable
```

Not sure how to reproduce exactly as I haven't noticed the exact moment when it happens. But sometimes the whole completion stops working after this issue and all I see in the console (until sublime restart) is `LSP:  --> textDocument/didChange`

It seems to be protocol implementation issue as null values in response seem to be allowed: https://microsoft.github.io/language-server-protocol/specification#textDocument_completion

> result: CompletionItem[] | CompletionList | null. If a CompletionItem[] is provided it is interpreted to be complete. So it is the same as { isIncomplete: false, items }

* OS and language server:
  * macOS
  * https://github.com/saibing/bingo
```
$ git rev-parse HEAD
263b5b30b0744d777a3bdfb77aa174abf6b1664b
```

* User config:
```
{
	"clients":
	{
		"golsp":
		{
			"command":
			[
				"/Users/margus/src/go/bin/bingo"
			],
			"enabled": true,
			"env":
			{
				"GOPATH": "/Users/margus/src/go"
			},
			"languageId": "go",
			"scopes":
			[
				"source.go"
			],
			"syntaxes":
			[
				"Packages/Go/Go.sublime-syntax"
			]
		}
	},
	"log_debug": true,
	"log_payloads": true,
	"log_stderr": true,
	"only_show_lsp_completions": true
}
```

rwols:
Fixes #494.

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -304,7 +304,7 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
         if self.state == CompletionState.REQUESTING:
             items = []  # type: List[Dict]
             if isinstance(response, dict):
-                items = response["items"]
+                items = response["items"] or []
             elif isinstance(response, list):
                 items = response
             items = sorted(items, key=lambda item: item.get("sortText") or item["label"])


rwols:
@kerma, I downloaded the bingo server and this seems to fix the issue. Could you try this one-liner diff too?

kerma:
@Kronuz So far so good 👍

PR #497, opened 2019-01-15
Add support for Djaneiro Python syntax

OlegSmelov:
If you install [Djaneiro](https://github.com/squ1b3r/Djaneiro), Python language server does not activate automatically because Djaneiro defines its own scope (`source.python.django`) and syntax (`Python (Django)`).

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -47,7 +47,7 @@
     {
       "command": ["pyls"],
       "scopes": ["source.python"],
-      "syntaxes": ["Packages/Python/Python.sublime-syntax"],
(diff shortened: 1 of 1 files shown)


tomv564:
Can you test if it works without adding the extra scope? (I expect "source.python" still matches your django sub-scope)

I'm okay with adding it to the default as I realize Djaneiro has a pretty good number of users, but I hope not to have to accommodate a huge list of syntaxes per language server in the end 😅

OlegSmelov:
It does work without the additional scope. 👍 

I understand your concern about having to maintain a list of custom syntaxes, but I guess adding support for a reasonable number of custom syntaxes will save a lot of frustration for users.

Thanks a lot for the plugin!

PR #501, opened 2019-01-27
add support for workspace/symbol

ayoub-benali:
Add support for [workspace/symbol](https://github.com/Microsoft/language-server-protocol/blob/master/versions/protocol-2-x.md#workspace_symbol)

Ideally we send a query on every keystroke but `TextInputHandler` doesn't have an `on_change` method. So I had to do it in two steps. First wait for the user to press `enter` then show results in a list via `ListInputHandler`. 

`show_input_panel` has an `on_change` method by you can't show two panels (for result) in the same time :( 

@tomv564 can I add more symbol kinds [here](https://github.com/tomv564/LSP/blob/master/plugin/symbols.py#L14) ? LSP 3 has more symbol kinds and it is backward compatible, see [spec](https://microsoft.github.io/language-server-protocol/specification#textDocument_documentSymbol)
As you can see in the demo, the panel is showing a number instead of the symbol kind, it would be nice to complete the list.

Demo:
workspace-symbol-5

cc @olafurpg


predragnikolic:
@ayoub-benali hello, I tried this locally.

And now I see what you meant by `issues getting the user experience right` :D 
I tried to see if I could make it better, but I couldn't. So no help from me.
It would be a lot easier if there was a `on_change` hook on the `TextInputHandler`, but currently there is none.

Also here is something that I noticed. 
While this command works with `rust` language server, 
When triggering the command with the `pyls` server, it completely freezes Sublime. 

One more thing. Although I like when people contribute to LSP
This PR made me think about the functionality that is already built in Sublime. 
Sublime already has a feature like this built in `goto_symbol_in_project` and I think it works great. What are the benefits you get with the `workspace/symbol`?  :)

ayoub-benali:
Hi @predragnikolic thanks for testing it out, I am glad it worked so well for rust :)

> When triggering the command with the `pyls` server, it completely freezes Sublime.

Probably because the server never replied? have you seen the server replying in the console ?
Maybe we can check for the server capability first before making this command available.  Should add a timeout as well.
 
> Sublime already has a feature like this built in `goto_symbol_in_project` and I think it works great. What are the benefits you get with the `workspace/symbol`? :)

 - `workspace/symbol` leads to a better experience, since the LSP server has a better context about your project then sublime, you get more info during your search like package name, symbol kind, etc you get what you want quicker.
- by default  `goto_symbol_in_project` indexes only the project source while `workspace/symbol` allows you to browse through the source code that is part of your "class path" without having to manage that your self by :
   - downloading the source dependencies yourself, by default your get only the compiled ones in several build tools / languages like java, scala.
   - adding source directories to the sublime project directories. 
 - it offloads the search part to the LSP server rather than having sublime loading all the dependencies and searching through them. Sublime can get a bit slow when it index a huge number of files, that is unfortunately common in my experience.

predragnikolic:
The `LSP: Workspace symbol` command is not available for `pyls` which means that `pyls` didn't support `workspace/symbol`, so that issue is solved :)

But I still don't like that I need to press `enter` for the search to happen. Is there a way to solve this? 

Also with `rls` there is a difference when typing 
**main**(press enter) main.rs (shows the main function in the list)
and 
**m**(press enter)ain main.rs (doesn't shows the main function in the list)
output

That is all from me, I have no other suggestions or issues :)

plugin/workspace_symbol.py
```diff
@@ -0,0 +1,94 @@
...
+    def validate(self, txt):
+        if txt != "":
+            self._completed_request = False
+            self._request_time = 0
+            request = Request.workspaceSymbol({"query": txt})
+            if self.client:
+                self.client.send_request(request, self._handle_response, self._handle_error)
+
+            # wait for request to complete
+            while not self._completed_request or not self._timeout:
```
tomv564: Sublime UI features do not accommodate async tasks, but holding the UI thread hostage with a time.sleep might result in complaints about freezes.

It's a nice feature that the text and list input handlers can be chained using `next_input`, but I'd prefer that we simply fire the request and let the response handler run a new command with the results provided to a ListInputHandler.

I hope we can remove the timeout, no idea if 3 seconds is a reasonable limit - we want users to feedback on the server to improve the behaviour :)

plugin/workspace_symbol.py
```diff
@@ -0,0 +1,94 @@
...
+        self.view = view
+        self.client = client_for_view(self.view)
+        self._symbols = []  # type: 'List[Dict]'
+        self._completed_request = False
+        self._request_time = 0
+
+    def _timeout(self) -> bool:
+        return self._request_time > 3
+
+    def _handle_response(self, symbols: 'List[Dict[str, Any]]'):
```
tomv564: This response can be null / `None` and should be `Optional`

plugin/workspace_symbol.py
```diff
@@ -0,0 +1,94 @@
...
+import os
+
+try:
+    from typing import List, Optional, Dict, Any
+    assert List and Optional and Dict and Any
+except ImportError:
+    pass
+
+
+class Symbol(sublime_plugin.ListInputHandler):
```
tomv564: Would prefer `ResultsHandler` or `SymbolListHandler`.

plugin/workspace_symbol.py
rwols: I agree here and have serious doubts about the user experience.

Can't we use something like a `view.set_status` while the request is in-flight, and then clear the `view.set_status` when the request handler gets invoked? And then show a quick_panel?

ayoub-benali: * The reason I combined the two `InputListener` with a wait in the middle is that it could be that your  search query doesn't lead to the results you want (it happens a lot to me). So you want to delete your input (query string) and issue a request again, like in the demo.
If we hide the input field and show a quick panel after and you don't like the results you have to first cancel the quick panel and then invoke the command again. This IMO breaks the user experience.
All this is a workaround because sublime doesn't have `on_change` to make the user experience interactive like in VS code :(

* during my testing using [metals](https://scalameta.org/metals/), the response time was never over a second. very often it was betwee 0.25 and 0.6. So in term of user experience, you don't actually see the UI freeze because the waiting time is very short and you are waiting for the list to populate anyway.
@predragnikolic Maybe you can tell us what was the response time for you with `rls` ?

I can reduce the timeout time but if you guys are still against the current design I can implement what Tom suggested.

ayoub-benali: >I'd prefer that we simply fire the request and let the response handler run a new command with the results provided to a ListInputHandler.

@tomv564 I tried your suggestion just to see the user experience and it doesn't work :/
when running the second command, sublime doesn't show the  `ListInputHandler` containing the list of symbols and goes straight to the the `run` method. 
The `input(args)` method returning the `ListInputHandler` is taken into consideration only if the command is run via the command palette, which is not the case here ...

predragnikolic: Sorry for the late reply, I tested this with the `lsp-tsserver` in a small js project. 
```
ended
3.3723504543304443   <-- this was the first request
ended
0.05594778060913086
ended
0.04345417022705078
ended
0.19172906875610352
ended
0.025737762451171875
ended
0.11135029792785645
ended
0.05542397499084473
```

Regarding rust. I am just learning rust, I usually experiment and just declare a couple of functions. 
These are the result for the `rls. I had 3 functions defined :D
```
ended
0.0030100345611572266
ended
0.0070781707763671875
ended
0.009324789047241211
```

ayoub-benali: @predragnikolic thanks for the insights. I changed the implementation to avoid the thread sleep now

ayoub-benali:
@tomv564 I changed the feature implementation to show a quick panel now, so no need for the thread sleep.
I had to remove the test because I couldn't validate the quick panel, when it shows up, in the test file even using [keypress](https://github.com/randy3k/Keypress)

plugin/workspace_symbol.py
olafurpg: I would avoid hardcoded timeouts if possible, or if you're gonna keep the timeout I recommend to least increase it to something like 20s. If a language server is super slow then that's an issue of the language server, the LSP package is not to blame. The exact response time depends on several factors: the query itself, number of results, and the size of the project, whether it's the first request that triggers indexing, computer hardware, operating system (I/O is noticeably slower on Windows), ...

In Metals we search both in project sources and library dependencies so a query can a lot of results in some cases leading to ~2s response times based on my experiments https://scalameta.org/metals/blog/2019/01/22/bloom-filters.html#response-times I'm sure we can optimize some parts down the road, maybe truncate the number of results for generic queries like `"S"` but at the moment I can't guarantee all responses will be <3s (esp. not for large codebases on Windows machines).

olafurpg:
If you send a request on every keystroke I recommend canceling old requests when the user enters new characters. This allows the language server to free up resources to compute the latest request.

A request is cancelled by sending a `$/cancelRequest` notification https://microsoft.github.io/language-server-protocol/specification#cancelRequest

plugin/workspace_symbol.py
ayoub-benali: thanks for the detailed description. I removed time out now because the results panel is shown asynchronously.

tomv564:
Thanks @ayoub-benali!

PR #505, opened 2019-01-30
add new symbol kinds

ayoub-benali:
fix https://github.com/tomv564/LSP/issues/504


tomv564:
Thanks, sorry for the slow response. Could you update the symbols here as well?

https://github.com/tomv564/LSP/blob/e9e0ad3fb92fc1d381cc315a6ccd850d9d48bbc7/plugin/core/sessions.py#L101

ayoub-benali:
@tomv564 done

PR #511, opened 2019-02-07
Implement quick panel for symbol references

alin23:
Usually the quick panel is better suited for showing symbol references
because you can cycle through the references using the keyboard.

It is also closer to the original Sublime Text behavior.

I chose to make this opt-in to avoid confusing existing users.

### DIFF
--- a/plugin/references.py
+++ b/plugin/references.py
@@ -6,7 +6,7 @@ from .core.documents import is_at_word, get_position, get_document_position
 from .core.panels import ensure_panel
 from .core.protocol import Request, Point
 from .core.registry import client_for_view, LspTextCommand
-from .core.settings import PLUGIN_NAME
+from .core.settings import PLUGIN_NAME, settings
 from .core.url import uri_to_filename
 from .core.workspace import get_project_path
 
@@ -25,6 +25,7 @@ def ensure_references_panel(window: sublime.Window) -> 'Optional[sublime.View]':
 class LspSymbolReferencesCommand(LspTextCommand):
     def __init__(self, view):
         super().__init__(view)
+        self.reflist = []  # type: List[List[str]]
 
     def is_enabled(self, event=None):
         if self.has_client_with_capability('referencesProvider'):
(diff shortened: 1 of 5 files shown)


tomv564:
Huge improvement in UX!

screenshot 2019-02-14 at 21 32 10

I'll add one review question, but I'd like to propose this as a default experience in a future version!

plugin/references.py
```diff
@@ -57,37 +58,87 @@ def handle_response(self, response: 'Optional[List[Dict]]', pos) -> None:
...
-        # highlight all word occurrences
-        regions = panel.find_all(r"\b{}\b".format(word))
-        panel.add_regions('ReferenceHighlight', regions, 'comment', flags=sublime.DRAW_OUTLINED)
-        panel.set_read_only(True)
+        if settings.show_references_in_quick_panel:
+            window.show_quick_panel(
+                self.reflist,
+                lambda index: self.on_ref_choice(base_dir, index),
+                sublime.MONOSPACE_FONT | sublime.KEEP_OPEN_ON_FOCUS_LOST,
+                self.get_current_ref(base_dir, word_region.begin()),
```
tomv564: I was unable to figure out what `get_current_ref` does - I commented it out and passed -1 instead. Does it select a meaningful default? Otherwise I'd love to exclude that logic.

alin23: `get_current_ref` is supposed to find the command palette entry that corresponds to the cursor position from which you sent the `find_reference` command.

So for example if you were in `plugin/core/test_rpc.py:15:20` (row 15 column 20) and sent a `find_reference` command, then `get_current_ref` should return `4` and highlight that specific entry in the command palette (`4` is the zero-based index from the list of references)

Can you try and see if that works for you? If it doesn't then every time, `find_reference` will highlight the first entry and open that specific file which is not really good if you were on another file.

tomv564: Thanks for explaining, that makes sense! 

I didn't think language servers guaranteed that a "self reference" would be included in the results. Some servers do (pyls included `types.py:11:7`, which is the declaration I invoked `find_reference` from).

In any case, you set up the opened views to be transient - so even if the self reference is missing, the user can press `ESC` and they will be where they left off.

alin23:
Thanks for the review and for merging! @tomv564 
Also thank you for making this tool, it is the final thing I missed from full fledged IDEs like IntelliJ an Xcode.

ayoub-benali:
thanks @alin23 really nice improvement

PR #512, opened 2019-02-07
Always reset completion state to IDLE when in CANCELLING

simonklb:
Previously adding a space while requesting completion would
cause completion to stay in CANCELLING indefinitely.

Fixes https://github.com/tomv564/LSP/issues/510

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -323,10 +323,10 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
             self.view.run_command("hide_auto_complete")
             self.run_auto_complete()
         elif self.state == CompletionState.CANCELLING:
+            self.state = CompletionState.IDLE
             if self.next_request:
                 prefix, locations = self.next_request
                 self.do_request(prefix, locations)
-            self.state = CompletionState.IDLE
         else:
             debug('Got unexpected response while in state {}'.format(self.state))
 
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -326,7 +326,7 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
             if self.next_request:
                 prefix, locations = self.next_request
                 self.do_request(prefix, locations)
-                self.state = CompletionState.IDLE
+            self.state = CompletionState.IDLE
         else:
             debug('Got unexpected response while in state {}'.format(self.state))
 


simonklb:
Don't merge this yet.

@tomv564 could you or someone that knows this codebase better than me please take a look and confirm if resetting the CANCELLING state to IDLE like this is ok or not?

My language server crashed with the error message `textDocument/completion: illegal Pos value` and this patch might be the culprit since I've not seen it before.

rwols:
I’ll try it out locally today or tomorrow.

We should really have some sort of integration test for these kinds of things.

tomv564:
Yes, you've found the fix for a problem I've seen from time-to-time: When responses for two completion requests arrive in a certain sequence, the completion state cannot not leave CANCELLING because no next request is scheduled.

Before merging I want to understand exactly what case triggers this, both potentially for testing and to make sure this is the right place for the fix.

Any chance you can repro with logging of the state transitions on both the request and response logic?

simonklb:
The issue I experienced was the case where it goes into `CANCELLING` state after you add a space during a request here: https://github.com/tomv564/LSP/blob/3f39c3dc5d3c8632e50eca54ad733768e99898be/plugin/completion.py#L204-L205

After this you never get out of the `CANCELLING` state since `self.next_request` is `None` here: https://github.com/tomv564/LSP/blob/3f39c3dc5d3c8632e50eca54ad733768e99898be/plugin/completion.py#L326-L329

The next time you try to do a completion you enter `on_query_completions` and `self.next_request` is set here:
https://github.com/tomv564/LSP/blob/3f39c3dc5d3c8632e50eca54ad733768e99898be/plugin/completion.py#L227-L229

But the problem is that you never get to the `handle_response` because `do_request` is never run while in the `CANCELLING` state.

Simple debug logging of a completion -> space -> failed completion attempts:
```
LSP: on_query_completions state = 0 # Initial completion
LSP:  --> textDocument/didChange
LSP:  --> textDocument/completion 
LSP:  --> textDocument/didChange # Followed by a quick spacebar
...
LSP: handle_response state = 3 # Completion response, cancelled by the space
LSP: handle_response next_request:  None
LSP: on_query_completions state = 3 # Failed completion attempt
LSP:  --> textDocument/didChange
LSP: on_query_completions state = 3 # Failed completion attempt
LSP:  --> textDocument/didChange
```

simonklb:
In the case where you do two request in succession, and the second request gets cancelled and next_request is set, the race condition you might be seeing is that next_request is set _after_ handle_response has completed on the first request.

simonklb:
I believe I've fixed the consecutive requests bug by moving the CANCELLING->IDLE state change before calling `do_request` for the `next_request`.

I discovered that the log would sometimes say `Got unexpected response while in state 0`. The second commit seems to have fixed that.

@tomv564 Not sure if this is the problem you were describing, but my completions are working much better now.

tomv564:
Thanks for documenting your findings, we now have some evidence to look at in case we need to fine-tune the logic further. Great work!

PR #515, opened 2019-02-18
Fix exception when kind is None

kkurian:
At least when using this package in conjunction with https://github.com/guillermooo/dart-sublime-bundle, there are cases where kind is None, resulting in an exception. This resolves that issue.


plugin/highlights.py
```diff
@@ -103,7 +103,8 @@ def _handle_response(self, response: 'Optional[List]') -> None:
         for highlight in response:
             r = range_to_region(Range.from_lsp(highlight["range"]), self.view)
             kind = highlight.get("kind", DocumentHighlightKind.Unknown)
-            kind2regions[_kind2name[kind]].append(r)
+            if kind is not None:
```
rwols: Indeed it seems the LSP specification allows `"kind"` to be `None`. LGTM.

PR #528, opened 2019-03-07
Remove trailing commas in pyls configuration

OlegSmelov:
https://github.com/tomv564/LSP/pull/522 introduced a trailing comma that shouldn't be there.


jfcherng:
Just in my humble opinion, trailing commas are good for editing and diff when adding/deleting a line.

OlegSmelov:
@jfcherng Trailing commas are not valid in JSON. I'm not sure if Sublime's parser specifically allows them in this file, but I've had issues with trailing commas in similar situations.

jfcherng:
@OlegSmelov ST's config is NOT JSON :)

OlegSmelov:
@jfcherng You are right 👍

PR #530, opened 2019-03-13
add magicpython for pyls

jgirardet:


tomv564:
Can you tell me a little bit more about the magicpython syntax? Why does it exist? 
As I commented when Djaneiro support was requested in #497, I have no way to reason about which extra syntaxes should be supported but would like to avoid accumulating _all_ of them. Is there a way we can make it easy for your users to make the config change themselves?

jgirardet:
I don't have any strong argument for it. it's not hard to set but since it's not straightforward maybe some user might go away due to this.
maybe something more detailed in the python section of the doc ?
feel free to close it.

imposeren:
Having ability to add a syntax using a local settings is good already. I don't think it's a good idea. I personally like MagicPython (I don't even remember why... maybe f-strings were not highlighed well using default syntax but that migth have changed already), but it has far less monthly installs than Djaneiro.

Maybe to prevent "users go away" it will be better to add "Enable Language Server for Syntax" command similar to "Enable Language Server for Project".

BTW: Troubleshooting sections in the docs suggests `sublime.active_window().active_view().settings().get("syntax")` command to get current syntax, and `LSP.sublime-settings` suggests `view.settings().get("syntax")` . Is there any difference? Maybe it's better to "sync" both instructions?

tomv564:
@imposeren great points, "enable for current syntax" might be an easy variation on "enable language server" to implement.
I was unaware that `view` resolves in the console, as well as `window`. We should absolutely simplify the instructions in that case!

tomv564:
Read a bit about magic python itself - it looks to be ambitious and relevant so let's add it!

PR #533, opened 2019-03-18
 Update docs on how to install clangd

jfcherng:
Add some problems that I encountered while trying to install and use the `clangd` LSP today.


tomv564:
Thanks!

PR #539, opened 2019-03-27
Update Reason's syntax path

jeremejevs:


LSP.sublime-settings
```diff
@@ -169,7 +169,7 @@
     {
       "command": ["ocaml-language-server", "--stdio"],
       "scopes": ["source.reason"],
-      "syntaxes": ["Packages/sublime-reason/Reason.tmLanguage"],
```
rwols: For reference, I'm assuming the package is this one: https://github.com/reasonml-editor/sublime-reason

and the pull request to change it into "Reason" was this one: https://github.com/wbond/package_control_channel/pull/7045

jeremejevs: Yep, that's correct, thanks for the links. Surprising that nobody has had an issue with this for so long :slightly_smiling_face:

Issue #544, opened 2019-03-30
LSP doesn't restart the server after it crashed


haferburg:
During development I kill the server process to recompile. When the client notices this, it asks "Language server ... has crashed, do you want to restart it?". But it never does when I click Yes. It still sends a shutdown request. Maybe it's waiting for a response.

Console output:

    Failure writing to stdout
    Traceback (most recent call last):
      File "AppData\Roaming\Sublime Text 3\Packages\LSP\plugin\core\transports.py", line 201, in write_stdin
        self.process.stdin.flush()
    OSError: [Errno 22] Invalid argument

    LSP: unloading session
    LSP:  --> shutdown {"jsonrpc": "2.0", "method": "shutdown", "id": 4}
    ending client

Possible duplicate of #250 and #236. Maybe this is a regression?

rwols:
fixed by #545

PR #545, opened 2019-03-30
After the transport fails, don't try to use it anymore. If a shutdown…

haferburg:
… request fails, still end the session. (#544)

### DIFF
--- a/plugin/core/rpc.py
+++ b/plugin/core/rpc.py
@@ -92,7 +92,7 @@ class Client(object):
             self._response_handlers[self.request_id] = (handler, error_handler)
             self.send_payload(request.to_payload(self.request_id))
         else:
-            debug(' sending ' + request.method + ' omitted')
+            debug('Unable to send', request.method)
             if error_handler is not None:
                 error_handler()
 
@@ -101,7 +101,7 @@ class Client(object):
             debug(' --> ' + notification.method)
             self.send_payload(notification.to_payload())
         else:
-            debug(' sending ' + notification.method + ' omitted')
+            debug('Unable to send', notification.method)
 
     def send_response(self, response: Response) -> None:
         self.send_payload(response.to_payload())
@@ -120,6 +120,7 @@ class Client(object):
         self._transport_fail_handler = handler
 
     def handle_transport_failure(self) -> None:
+        debug('Transport failed')
         self.transport = None
         if self._transport_fail_handler is not None:
             self._transport_fail_handler()
--- a/plugin/core/rpc.py
+++ b/plugin/core/rpc.py
@@ -72,7 +72,7 @@ def try_terminate_process(process: 'subprocess.Popen') -> None:
 
 class Client(object):
     def __init__(self, transport: Transport, settings) -> None:
-        self.transport = transport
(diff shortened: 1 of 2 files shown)


haferburg:
I will continue the fight with the type checker later...

rwols:
Just run `tox` in your terminal (install with `python3 -m pip install tox`) to check what the build server does (approximately)

plugin/core/rpc.py
```diff
@@ -87,13 +87,21 @@ def __init__(self, transport: Transport, settings) -> None:
...
         self.request_id += 1
-        debug(' --> ' + request.method)
-        self._response_handlers[self.request_id] = (handler, error_handler)
-        self.send_payload(request.to_payload(self.request_id))
+        if self.transport is not None:
+            debug(' --> ' + request.method)
+            self._response_handlers[self.request_id] = (handler, error_handler)
+            self.send_payload(request.to_payload(self.request_id))
+        else:
+            debug(' sending ' + request.method + ' omitted')
```
rwols: why did you prefix the word "sending" with a space here?

it might be best to use this:

```
debug('sending', request.method, 'omitted')
```
the `debug` function concatenates all of its arguments together with spaces in-between, just like the `print` function.

plugin/core/rpc.py
```diff
@@ -87,13 +87,21 @@ def __init__(self, transport: Transport, settings) -> None:
...
+                error_handler()
 
     def send_notification(self, notification: Notification) -> None:
-        debug(' --> ' + notification.method)
-        self.send_payload(notification.to_payload())
+        if self.transport is not None:
+            debug(' --> ' + notification.method)
+            self.send_payload(notification.to_payload())
+        else:
+            debug(' sending ' + notification.method + ' omitted')
```
rwols: same here.

haferburg: @rwols The console now reads

    LSP:  --> textDocument/formatting
    Failure writing to stdout
      File "\AppData\Roaming\Sublime Text 3\Packages\LSP\plugin\core\transports.py", line 201, in write_stdin
        self.process.stdin.flush()
    OSError: [Errno 22] Invalid argument

    LSP: transport failed
    LSP: unloading session dkmv
    LSP: unable to send shutdown
    LSP: unable to send exit

plugin/core/rpc.py
haferburg: Good point. I added the space because the other message is `debug(' --> '+...).`

haferburg:
Console output looks like this now:

    LSP:  --> textDocument/formatting
    Failure writing to stdout
      File "D:\Users_SSD\Queequeg_2\AppData\Roaming\Sublime Text 3\Packages\LSP\plugin\core\transports.py", line 201, in write_stdin
        self.process.stdin.flush()
    OSError: [Errno 22] Invalid argument

    LSP: transport failed
    LSP: unloading session dkmv
    LSP: unable to send shutdown
    LSP: unable to send exit

haferburg:
Thanks, everyone!

rwols:
Thanks for this nice piece of work :)

PR #546, opened 2019-03-30
use CompletionItem.label for completion menu

ayoub-benali:
The label should be displayed to the user and insert text can be trimmed [here](https://github.com/tomv564/LSP/blob/master/plugin/completion.py#L296)

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -275,7 +275,7 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
                 hint = completion_item_kind_names.get(kind)
         # label is an alternative for insertText if neither textEdit nor insertText is provided
         insert_text = self.text_edit_text(item) or item.get("insertText") or label
-        trigger = insert_text
+        trigger = label
         if len(insert_text) > 0 and insert_text[0] == '$':  # sublime needs leading '$' escaped.
             insert_text = '\\$' + insert_text[1:]
         # only return trigger with a hint if available


plugin/completion.py
```diff
@@ -275,7 +275,7 @@ def format_completion(self, item: dict) -> 'Tuple[str, str]':
                 hint = completion_item_kind_names.get(kind)
         # label is an alternative for insertText if neither textEdit nor insertText is provided
         insert_text = self.text_edit_text(item) or item.get("insertText") or label
-        trigger = insert_text
+        trigger = label
```
rwols: I'm actually working on exactly this, too!

The thing with the label is that it might be totally different than the trigger characters. Consider what the completion tuple for SublimeText should be:

```
[TRIGGER\tHINT, REPLACEMENT]
```

Here, **trigger** should contain the prefix of what the user is trying to type. Now note that the *label* of a CompletionItem might be totally different than the prefix of what the user is trying to type!

For instance, consider what clangd does: the label looks like

```
<prefix>function-or-variable-name<possibly opening parens><possibly args without types><possibly closing parens>
```

Here, <prefix> is either a space (` `) or a dot (`⚈`) depending on wether `additionalInsertText` is present.

However, the *filterText* of a CompletionItem *must* be a prefix of what the user is trying to type. So for this reason I think we should actually use *filterText* instead of *label*.

There is a little bit more stuff in my work, but it contains the above idea.

ayoub-benali:
@rwols I will close this one then and wait for your PR

PR #551, opened 2019-04-04
Fix: Use "filterText" (fallback to "label") for the trigger of the completion item

rwols:
An option is added to always prefer "label" over the "filterText" if the user so desires. However, certain language servers might use the "label" to denote extra data, e.g. a circular dot in front of the word-to-complete in case it returns a completion with an "additionalTextEdits" key. In this case, using the "label" for the ST trigger characters can give wrong behavior. An example of this wrong behavior is: https://github.com/tomv564/LSP/issues/368

One can take a look at the completion_test.py file to get a feel for what this change entails.

ping @ayoub-benali @frou

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -355,6 +355,11 @@
   // Resolve completions and apply snippet if received.
   "resolve_completion_for_snippets": false,
 
+  // When presenting completions, prefer the "label" over the "filterText" key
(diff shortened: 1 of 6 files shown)


ayoub-benali:
>However, certain language servers might

Another argument to have the "label" actually the default and using the "filterText" as the exception. i.e. I would flip default of `prefer_label_over_filter_text` flag

PR #554, opened 2019-04-06
Speed up edit command (resolves #548)

### LINKED ISSUE #548: LspApplyDocumentEditCommand is slow
I'm testing my formatter with a 5k LoC file every now and then. My server computes the required edits quickly. LSP then freezes Sublime for 5 s while it applies the ~800 edits. For 1360 changes it takes 30 s. I traced this to the `LspApplyDocumentEditCommand`, which looks like it's at least in O(n²).

The algorithm uses the Sublime API to repeatedly query regions for each change, so Sublime needs to build that list again and again.

Alternatively, we could apply the changes in reverse order, so that the regions we're about to edit are still valid. We're sorting the changes anyways, all it takes is a `reverse=True`. The `TextEdit[]` specification says "_Text edits ranges must never overlap, that means no part of the original document must be manipulated by more than one edit._"

Preparing a PR from https://github.com/haferburg/LSP/tree/faster-edit-command. After my change, it always takes less than 1 s.

Am I missing something?

The LSP spec also says "_If multiple inserts have the same position, the order in the array defines the order in which the inserted strings appear in the resulting text._" I'm not sure if that already works, but I think it does. Needs more thorough testing.

haferburg:
I don't think this is working properly yet.


haferburg:
Also the git history is a bit messed up, is that a problem?

rwols:
> Also the git history is a bit messed up, is that a problem?

No worries, pull request commits are squashed and then merged.

plugin/core/edit.py
```diff
@@ -100,16 +82,24 @@ def changes_sorted(self, changes: 'List[dict]') -> 'List[Dict]':
...
+        # The spec reads:
+        # > However, it is possible that multiple edits have the same start position: multiple
+        # > inserts, or any number of inserts followed by a single remove or replace edit. If
+        # > multiple inserts have the same position, the order in the array defines the order in
+        # > which the inserted strings appear in the resulting text.
+        # So we sort by start position. But if multiple text edits start at the same position,
+        # we use the index in the array as the key.
+        indexed_changes = [(i, c) for i, c in enumerate(changes)]
+        indexed_changes = sorted(indexed_changes, key=get_start_position, reverse=True)
+        return [ic[1] for ic in indexed_changes]
```
haferburg: I don't know about these lines. It works, but seems a bit sloppy. Is there a faster of more elegant way to express this?

rwols: you can just do `indexed_changes = sorted(enumerate(changes), key=get_start_position, reverse=True)`, no?

You could also write it like this:

```
def second(tup: 'Tuple[any, any]') -> any:
    return tup[1]

...

return list(map(second, sorted(enumerate(changes), key=get_start_position, reverse=True)))

```

rwols: In fact I don't think you have to convert to a list at all:

```
return map(second, sorted(enumerate(changes), key=get_start_position, reverse=True))
```

rwols: ... But then you'd have to change the return type of `changes_sorted_in_reverse` from `List[dict]` to `Iterable[dict]`

haferburg:
I don't understand what I need to do. My local mypy is happy. Travis' mypy complains in [build 1248](https://travis-ci.org/tomv564/LSP/builds/516534153?utm_source=github_status&utm_medium=notification) about an include. I remove the include, then it complains in [build 1253](https://travis-ci.org/tomv564/LSP/builds/516544725?utm_source=github_status&utm_medium=notification) that the include is missing.

rwols:
You need to `assert Any` when you import it

```
    from typing import List, Dict, Optional, Any
    assert List and Dict and Optional and Any
```
this is a workaround for python3.3 embedded into SublimeText.

haferburg:
Oh, I see. Thanks!
I was only able to reproduce it by specifying

    flake8 == 3.5.0
    mypy == 0.570

in the `tox.ini`. Also `envlist = py33`. Can I commit that? How does Travis choose versions?

plugin/core/edit.py
```diff
@@ -55,41 +56,29 @@ def open_and_apply_edits(self, path, file_changes):
 
 class LspApplyDocumentEditCommand(sublime_plugin.TextCommand):
     def run(self, edit, changes: 'Optional[List[dict]]' = None, show_status=True):
+        # Apply the changes in reverse, so that we don't invalidate the range
+        # of any change that we haven't applied yet.
+        start = timer()
+        changes2 = changes or []
```
rwols: you could also just call this `changes` again: `changes = changes or []`, no?

tests/test_edit.py
```diff
@@ -1,53 +1,69 @@
...
+            text_edit((1, 2), (1, 3), '6'),  # replace the h
+            text_edit((0, 1), (0, 1), '1'),  # insert after a
+            text_edit((0, 1), (0, 1), '2'),
+            text_edit((0, 1), (0, 1), '3'),
+        ]
+        expected = (
+            'a123bcde\n'
+            'fg456ij\n'
+        )
+        self.run_test(original, expected, file_changes)
```
rwols: thanks for this test!

plugin/core/edit.py
```diff
@@ -55,41 +56,29 @@ def open_and_apply_edits(self, path, file_changes):
...
             window = self.view.window()
             if window:
                 base_dir = get_project_path(window)
                 file_path = self.view.file_name()
                 relative_file_path = os.path.relpath(file_path, base_dir) if base_dir else file_path
-                message = 'Applied {} change(s) to {}'.format(len(changes), relative_file_path)
+                message = 'Applied {} change(s) to {} in {:.1f} ms'.format(
+                    len(indices), relative_file_path, elapsed * 1000)
                 window.status_message(message)
+                debug(message)
```
rwols: the extra `debug(message)` here feels unnecessary.

plugin/core/edit.py
```diff
@@ -55,41 +56,29 @@ def open_and_apply_edits(self, path, file_changes):
...
-                # current region was removed, don't advance index.
-                last_region_count = len(self.view.get_regions('lsp_edit'))
-
-        self.view.erase_regions('lsp_edit')
         if show_status:
             window = self.view.window()
             if window:
                 base_dir = get_project_path(window)
                 file_path = self.view.file_name()
                 relative_file_path = os.path.relpath(file_path, base_dir) if base_dir else file_path
```
rwols: we can compute `elapsed = timer() - start` at this point, instead of earlier.

plugin/core/edit.py
haferburg: I added it because the window.status_message disappears after a second or so, and I often missed it.

plugin/core/edit.py
haferburg: Unfortunately not. Since `changes` might be null, mypy won't allow `changes[index]`. So I introduced another variable with a different type (not `Optional`).

Would appreciate a better solution.

plugin/core/edit.py
haferburg: True, but I wanted to time only the algorithm. Are you worried about something specific?

rwols:
Thanks for the nice work :) let's see how it behaves for various language servers.

haferburg:
Thank you! You guys are very nice and helpful.

PR #556, opened 2019-04-06
Move window/logMessage handler to the rest of the notification handlers in window.py

rwols:


rwols:
> Note that the old implementation had an early return on window/logMessage

Ah yes, I hadn't thought of that.

PR #559, opened 2019-04-07
Completions that start with a "$" are not extracted correctly

rwols:
The last_row, last_col logic causes $_POST to become _POST et cetera.
I have removed this logic in favor of getting intelephense to work correctly.
We should rethink the edit range of the TextEdit.

@nanoant, I would like to test these changes with https://github.com/PowerShell/PowerShellEditorServices. I fired up my windows machine, installed Powershell 6, cloned the repository, and ran the instruction in the repository. But I don't understand how to actually run the server with SublimeText. What do I put in my `"clients"` dict to run the server?

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -45,7 +45,7 @@ def extract_hint(completion_item: 'Dict[str, Any]') -> 'Optional[str]':
     return hint
 
 
-def extract_replacement(completion_item: 'Dict[str, Any]', fallback: str) -> str:
+def extract_replacement(completion_item: 'Dict[str, Any]', fallback: str, prefix: str, word_separators: str) -> str:
     replacement = None
     text_edit = completion_item.get("textEdit")
     if text_edit:
@@ -53,17 +53,21 @@ def extract_replacement(completion_item: 'Dict[str, Any]', fallback: str) -> str
         replacement = text_edit.get("newText")
     if not replacement:
         replacement = completion_item.get("insertText") or fallback
-    # We make a choice here. If no "insertTextFormat" key is present, what should we choose as default?
-    # The specification is not clear about the absence of this key. We will assume PlainText.
     if completion_item.get("insertTextFormat", 1) == 1:
+        if replacement[0] in word_separators and prefix and prefix[0] not in word_separators:
+            repl = replacement[1:]
+            for sep in word_separators:
+                if sep in repl:
+                    replacement = repl
+                    break
         replacement = replacement.replace("$", "\\$")
     return replacement
 
 
-def format_completion(completion_item: 'Dict[str, Any]') -> 'Tuple[str, str]':
+def format_completion(completion_item: 'Dict[str, Any]', prefix: str, word_separators: str) -> 'Tuple[str, str]':
     trigger = extract_trigger(completion_item)
     hint = extract_hint(completion_item)
-    replacement = extract_replacement(completion_item, trigger)
+    replacement = extract_replacement(completion_item, trigger, prefix, word_separators)
     return "{}\t  {}".format(trigger, hint) if hint else trigger, replacement
 
 
@@ -298,11 +302,11 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
             if document_position:
                 client.send_request(
                     Request.complete(document_position),
-                    self.handle_response,
+                    lambda response: self.handle_response(prefix, response),
                     self.handle_error)
                 self.state = CompletionState.REQUESTING
 
-    def handle_response(self, response: 'Optional[Dict]'):
+    def handle_response(self, prefix: str, response: 'Optional[Dict]'):
         global resolvable_completion_items
 
         if self.state == CompletionState.REQUESTING:
@@ -312,7 +316,8 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
             elif isinstance(response, list):
                 items = response
             items = sorted(items, key=lambda item: item.get("sortText") or item["label"])
-            self.completions = list(format_completion(item) for item in items)
+            word_separators = self.view.settings().get("word_separators") or ""
+            self.completions = list(format_completion(item, prefix, word_separators) for item in items)
 
             if self.has_resolve_provider:
                 resolvable_completion_items = items
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -11,7 +11,7 @@ from .core.protocol import Request
 from .core.events import global_events
 from .core.settings import settings
 from .core.logging import debug, exception_log
-from .core.protocol import CompletionItemKind, Range
+from .core.protocol import CompletionItemKind
 from .core.registry import session_for_view, client_for_view
 from .core.configurations import is_supported_syntax
 from .core.documents import get_document_position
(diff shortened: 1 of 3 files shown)


nanoant:
@rwols Hi, no need for Windows, it works also well on Linux/macOS. Actually I have left my Windows dev machine at work so I don't have it home. So if you don't mind I will try to give you recipe for running Editor Services on macOS/Linux.

1. Install `PowerShell` Sublime module, so you get `source.powershell` scope.

2. Install `pwsh` on Linux/macOS, either via brew or something else. I use NixOS package manager for the example below.

3. Download latest (I think `2.0.0-prev`) [PowerShellEditorServices](https://github.com/PowerShell/PowerShellEditorServices)

4. Extract `PowerShellEditorServices/PowerShellEditorServices` subfolder (NOT main folder) from `PowerShellEditorServices.zip` into PowerShell Modules folder, i.e.:

    * `%UserProfile%\Documents\WindowsPowerShell\Modules` on Windows
    * `~/.local/share/powershell/Modules` on Linux/macOS

4. Use following entry in your user `LSP.sublime-settings`:

	~~~json
	"pwsh":
	{
		"command":
		[
			"~/.nix-profile/bin/pwsh",
			"~/.local/share/powershell/Modules/PowerShellEditorServices/Start-EditorServices.ps1",
			"-BundledModulesPath", "~/.local/share/powershell/Modules/PowerShellEditorServices/bin",
			"-HostName", "SublimeText",
			"-HostProfileId", "SublimeText",
			"-HostVersion", "3.0",
			"-Stdio",
			"-LogPath", "$TMPDIR/PowerShellEditorServices.log",
			"-LogLevel", "Diagnostic",
			"-SessionDetailsPath", "$TMPDIR/SessionDetails.json",
			"-FeatureFlags", "PSReadLine"
		],
		"enabled": true,
		"scopes": ["source.powershell"],
		"syntaxes": ["Packages/PowerShell/Support/PowershellSyntax.tmLanguage"],
		"languageId": "PowerShell"
	},
	~~~

	NOTES:
	* `~/.nix-profile/bin/pwsh` is my macOS PowerShell interpreter path, may be different for you. On Windows command is called `powershell` not `pwsh`.
	* `-BundledModulesPath` seems to be necessary for macOS (my installation).
	* `-FeatureFlags` seems to be necessary too on macOS, otherwise you get cryptic error, and you need to look into `$TMPDIR/PowerShellEditorServices.log` to know what happened.
	* `-Stdio` has to be explicitly specified, otherwise it goes through TCP I believe. I had once a problem with `-Stdio` on Windows, but on macOS seems to work just fine.
	* Finally I recommend first trying running the command directly from command line, e.g. `pwsh ~/.local/share/powershell/Modules/PowerShellEditorServices/Start-EditorServices.ps1`

6. Run Sublime with some simple `Test.ps1` file, you should see some completion when trying to type `Get-` e.g.

	image

Good luck & let me know if it works for you.

rwols:
Nice! I actually got it working on my MacBook. Some initial impressions:

1) "filterText" over "label" really is the Right Way (tm) for SublimeText.
2) Completion for `$env:` is replaced in my text buffer with `$$env:`, oops!
3) Completion for something that does *not* end with a colon seems to be working alright: `$Error`, `$ErrorView` etc.
4) As an aside it is strange that `$` is not a trigger character according to the server:
```
[{'selector': 'text.html', 'characters': '<'}, {'selector': 'source.powershell', 'characters': '.-:\\'}]
```

rwols:
The fact that we end up with two dollar-signs for `$env` might also be a bug of sublime text. Consider this simple completion provider:

```
import sublime_plugin


class TestEventListener(sublime_plugin.ViewEventListener):
    def on_query_completions(self, prefix, locations):
        return [
            ("$foooooooooo", "\\$foooooooooo"), #1
            ("%foooooooooo", "%foooooooooo"),   #2
            ("$foooooo:ooo", "\\$foooooo:ooo"), #3
            ("%foooooo:ooo", "%foooooo:ooo"),   #4
            ("foo$oooooooo", "foo\\$oooooooo"), #5
            ("foo%oooooooo", "foo%oooooooo"),   #6
            ("@foo$ooooooo", "@foo\\$ooooooo"), #7
            ("!foo!ooo!ooo", "!foo!ooo!ooo")    #8
        ]

```
* (1) and (2) work correctly.
* With (3) and (4), we end up with double dollar-signs, or double percent-signs. It does not matter wether it is a dollar-sign, it seems to happen with every non-word character.
* (5) and (6) work correctly, perhaps because the initial character is not a non-word character.
* (7) and (8) behave like (3) and (4) in that the initial non-word character `@` and `!` are not replaced.

In all the bad cases (3), (4), (7), (8) there are more than one non-word characters in the replacement text.

rwols:
Also, the following completion item works perfectly fine:

```
("foo$o$oooooo", "foo\\$o\\$oooooo")
```

So there seems to be a bug in the following situation:

1) The first character of the replacement text is a non-word character.
2) There is another non-word character in the replacement text.

rwols:
I think these two commits make intelephense and PowerShellEditorServices work nicely with as minimal view state as possible, using only the word separators of the view. I have also added completion samples for PowerShellEditorServices so that anybody can screw around with the completion handling logic and verify that things keep working as intended.

tomv564:
Great that you're testing the powershell server and documenting the behaviour.

It looks like we're trading one complexity for another, as I didn't fully understand the previous behaviour either I'd really like @nanoant 's comments on the new logic?

nanoant:
It took me a while but I think I have finally nailed down the problem. I switched to latest LSP's `master` and started testing [PowerShellEditorServices](https://github.com/PowerShell/PowerShellEditorServices) and surprisingly the completion was **not** at first working as intended, when completing `$Env:U` the whole text was replaced to `USER` but I should end up with `$Env:USER` there 🙄 

Sublime2

But then I was thinking what has changed since my commit ed7d7a2a90a227dfa92f8eb1d23fe30c20d58c22 and I came across @rwols commit f4b4ce93ef9ef4da4ec97171fdcc6536b2c1aa3f and new `prefer_label_over_filter_text`. I switched it on, so the went back to the old behavior as in time of my commit. And it worked fine, this is what I got:

Sublime3

But wait, the only difference between `prefer_label_over_filter_text` on and off is the label, but apparently the label makes a difference how to Sublime Text completes the text. My sane assumption was that Sublime Text is replacing the `prefix` as it is supplied to `on_query_completions(self, prefix, locations)` but I was **WRONG**, apparently it tries to replace the label! 🤒 

But what if the label is even not there in the text!? Then I have no clue what is the replacement.

I would really would like to have @wbond and official Sublime Text powers involved in this LSP support and I believe that only reasonable way to handle this is to have deterministic way to replace the buffer using custom range and text when completion is selected. Otherwise we can just guess what is Sublime Text logic and we will never come to some solid solution.

Therefore I will reference also here this: https://github.com/SublimeTextIssues/Core/issues/2754

@rwols And finally I have tested this PR `bugfix/dollar-sign-completions` branch, and unfortunately
it does not work at all with PowerShell, neither with:
1. `settings.prefer_label_over_filter_text` on
    Sublime4

2. or `settings.prefer_label_over_filter_text` off
    Sublime5

So I believe we will need to find other way.

rwols:
It seems I was too quick to open this draft work for review :)

> when completing $Env:U the whole text was replaced to USER but I should end up with $Env:USER there

Yes, I only tested it by typing $Env:, that part works.

> new prefer_label_over_filter_text. I switched it on, so the went back to the old behavior as in time of my commit. And it worked fine, this is what I got:

It looks like this also doesn't work as intended here :(

Note that in f4b4ce93ef9ef4da4ec97171fdcc6536b2c1aa3f I [had also removed this line](https://github.com/tomv564/LSP/commit/f4b4ce93ef9ef4da4ec97171fdcc6536b2c1aa3f#diff-ccb284e84187b2ce90d374d118bb2d9eL278), which seems to make things work for Powershell & PHP. git blame shows that this was done to close [this issue](https://github.com/tomv564/LSP/issues/368) (the issue boils down to the language server having "decorated labels") However, it has the undesired effect of [showing snippets verbatim](https://github.com/tomv564/LSP/issues/547). So in order to fix that I chose to use filterText (fallback to label), which now breaks PHP and Powershell.

> I believe that only reasonable way to handle this is to have deterministic way to replace the buffer using custom range and text when completion is selected.

Agreed. It's impossible to maintain all language server's assumptions. Explicit replacement ranges are the true fix. [We haven't even tried tackling this yet](https://github.com/tomv564/LSP/issues/536), which I think is impossible without an explicit replacement range.

> So I believe we will need to find other way.

Yep. Though, are you sure using the label works correctly for Powershell? It doesn't work for PHP.

Let's not merge this right now.

james2doyle:
Just found this from the other thread about the first character being cut off. Here is the PHP example:

Screen Recording 2019-05-22 at 08 59 AM

PR #561, opened 2019-04-11
Add configuration for gopls

frou:
There is an official Go language server taking shape: https://github.com/golang/go/wiki/gopls

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -240,6 +240,13 @@
       "syntaxes": ["Packages/Go/Go.sublime-syntax"],
       "languageId": "go"
     },
+    "gopls":
+    {
+      "command": ["gopls"],
+      "scopes": ["source.go"],
+      "syntaxes": ["Packages/Go/Go.sublime-syntax"],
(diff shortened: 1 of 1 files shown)


tomv564:
Thanks! Would be nice to help go developers choose and install this new version if it improves over the existing one.

frou:
gopls is still in flux, but completions/hover/goto-definition seem to work quite well just now (including being able to jump all the way into stdlib source files with goto-definition!). After the next release of this package, I'll edit their wiki page to mention Sublime

rwols:
Does signature help work as expected?

frou:
@rwols The popup with the appropriately underlined parameter works (apart from a varargs final parameter, e.g. `xs ...T`). I may not be aware of all capabilities of Signature Help though

frou:
I've updated their wiki to mention Sublime+LSP:

https://github.com/golang/go/wiki/gopls#integration-with-your-text-editor

(Look at the amount of fragmentation for Vim and Vim-likes!)

PR #567, opened 2019-04-15
Correctly handle code action of CodeAction type (fixes #566)

### LINKED ISSUE #566: Code actions of CodeActionLiteral type are not supported (used by eslint-server)
While experimenting with [eslint-server](https://github.com/Microsoft/vscode-eslint/tree/master/server), I've noticed that its code actions for fixing lint issues are not handled properly.

I found two issues:
 - handling CodeActions action list as an array of `Command` while protocol also allows array of `CodeAction`.
 - Passing `args` param to `workspace/executeCommand` instead of expected `arguments` param.

rchl:
Fixes two issues:
  - The Command's arguments passed to workspace/executeCommand should be
    passed through 'arguments' property rather than 'args'.
  - Fix handling of code action in case textDocument/codeAction [1] returns
    list of CodeAction's rather than Command's [2].

    Instead of having LspCodeActionsCommand duplicate logic of
    LspExecuteCommand, pass data from former to the latter to
    execute code action. Had to modify latter to not trigger a
    dialog box on success as it's not the best user experience in case
    when executing action from a popup menu.
  - Advertise support for workspace/executeCommand as we clearly support
    that. That wasn't necessary to fix anything but it felt right to do.

[1] https://microsoft.github.io/language-server-protocol/specification#textDocument_codeAction
[2] https://microsoft.github.io/language-server-protocol/specification#command

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -7,9 +7,10 @@ try:
 except ImportError:
     pass
 
-from .core.registry import LspTextCommand
+from .core.registry import client_for_view, LspTextCommand
 from .core.protocol import Request
 from .diagnostics import get_point_diagnostics
+from .core.edit import parse_workspace_edit
 from .core.url import filename_to_uri
 from .core.views import region_to_range
 from .core.registry import session_for_view
@@ -80,6 +81,11 @@ class LspCodeActionBulbListener(sublime_plugin.ViewEventListener):
         self.view.erase_regions('lsp_bulb')
 
 
+def is_command(command_or_code_action: dict) -> bool:
+    command_field = command_or_code_action.get('command')
+    return isinstance(command_field, str)
+
+
 class LspCodeActionsCommand(LspTextCommand):
     def is_enabled(self):
         return self.has_client_with_capability('codeActionProvider')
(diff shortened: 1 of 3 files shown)


plugin/code_actions.py
```diff
@@ -108,11 +108,18 @@ def show_popup_menu(self) -> None:
...
         if index > -1:
-            client = client_for_view(self.view)
-            if client:
-                client.send_request(
-                    Request.executeCommand(self.commands[index]),
-                    self.handle_command_response)
-
-    def handle_command_response(self, response):
-        pass
+            # TODO for CodeAction, handle potential WorkspaceEdit before Command.
```
tomv564: I'd like to see this supported as part of the CodeAction implementation

plugin/code_actions.py
```diff
@@ -108,11 +108,18 @@ def show_popup_menu(self) -> None:
...
+            if command:
+                command_name = command.get('command')
+                command_args = command.get('arguments', None)
+
+                args = {
+                    "command_name": command_name,
+                    "command_args": command_args,
+                    "modal_result": False,
+                }
+                self.view.run_command('lsp_execute', args)
```
tomv564: Would prefer if the command execution didn't use `LspExecuteCommand`, that class is for letting the user directly run commands from the quick panel and not to be invoked from code.

It's fine if we want them to share functionality though!

plugin/execute_command.py
```diff
@@ -14,27 +14,32 @@ class LspExecuteCommand(LspTextCommand):
     def __init__(self, view):
         super().__init__(view)
 
-    def run(self, edit, command_name=None, command_args=None) -> None:
+    def run(self, edit, command_name=None, command_args=None, modal_result=True) -> None:
```
tomv564: I like the idea of letting the user specify `modal_result`, but this functionality was meant to be separate from the code action command execution. Can we move it to another PR?

plugin/execute_command.py
```diff
@@ -14,27 +14,32 @@ class LspExecuteCommand(LspTextCommand):
...
         msg = "command {} failed. Reason: {}".format(command, error.get("message", "none provided by server :("))
         sublime.message_dialog(msg)
 
-    def _send_command(self, client: Client, command_name: str, command_args: 'Dict[str, Any]') -> None:
+    def _send_command(self, client: Client, command_name: str, command_args: 'Optional[List[Any]]',
+                      modal_result) -> None:
         request = {
             "command": command_name,
-            "args": command_args
+            "arguments": command_args
```
tomv564: Sorry, I jumped the gun and fixed this in master as it was a clear bug!

plugin/core/sessions.py
```diff
@@ -136,7 +136,8 @@ def get_initialize_params(project_path: str, config: ClientConfig):
             },
             "workspace": {
                 "applyEdit": True,
-                "didChangeConfiguration": {}
+                "didChangeConfiguration": {},
+                "executeCommand": {},
```
tomv564: Nice catch here as well!

AmjadHD:
https://github.com/rchl/LSP/blob/1e048eb498a01a5e49a1617da3e18c662f59ded4/plugin/code_actions.py#L69
that failes if the response is None, which is the case of [lua ls](https://github.com/tomv564/LSP/issues/596).

tomv564:
Tested with lsp-tsserver, hope it works for your use cases as well. Thanks again @rchl !

ayoub-benali:
Thanks for the PR completion @tomv564

Issue #482, opened 2018-12-14
Option to hide links in hover


predragnikolic:
Currently the hover looks like this:
![screenshot from 2018-12-14 17-03-08](https://user-images.githubusercontent.com/22029477/50014040-1b81d800-ffc3-11e8-8dc4-13cbd907d572.png)

But I would at least appreciate a flag to remove the links, so it would look like:
![screenshot from 2018-12-14 17-04-24](https://user-images.githubusercontent.com/22029477/50014091-3e13f100-ffc3-11e8-8907-6ce55da0046c.png)

PR #568, opened 2019-04-16
Show symbol actions by client capabilities

ayoub-benali:
Show only the actions that the server can handle during the hover event

### DIFF
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -82,21 +82,23 @@ class LspHoverCommand(LspTextCommand):
             all_content += self.diagnostics_content(point_diagnostics)
 
         all_content += self.hover_content(point, response)
-        all_content += self.symbol_actions_content()
+        all_content += self.symbol_actions_content(point)
 
         _test_contents.clear()
         _test_contents.append(all_content)  # for testing only
         self.show_hover(point, all_content)
 
-    def symbol_actions_content(self):
+    def symbol_actions_content(self, point):
         actions = []
-        if self.has_client_with_capability('definitionProvider') and is_at_word(self.view, None):
+        (x, y) = self.view.rowcol(point)
+        event = {"x": x, "y": y}
+        if self.has_client_with_capability('definitionProvider') and is_at_word(self.view, event):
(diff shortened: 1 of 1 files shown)


tomv564:
Can you please put in the effort and write a description for your pull request? For example, I guess this is for issue #482 ?

plugin/hover.py
```diff
@@ -82,18 +82,25 @@ def handle_response(self, response: 'Optional[Any]', point) -> None:
...
-    def symbol_actions_content(self):
+    def symbol_actions_content(self, point):
         actions = []
-        # TODO: filter by client capabilities
-        actions.append("<a href='{}'>{}</a>".format('definition', 'Definition'))
-        actions.append("<a href='{}'>{}</a>".format('references', 'References'))
-        actions.append("<a href='{}'>{}</a>".format('rename', 'Rename'))
+        (x, y) = self.view.rowcol(point)
+        event = {"x": x, "y": y}
+        if self.has_client_with_capability('definitionProvider') and is_at_word(self.view, event):
```
tomv564: I would suggest not including `is_at_word` in every statement but just doing it once. 

But even more so, this check is already done before the hover request is made (see `is_likely_at_symbol`) so perhaps we can avoid doing the same check a different way again?

ayoub-benali:
> Can you please put in the effort and write a description for your pull request? For example, I guess this is for issue #482 ?

The PR is simply taking care of a todo left in the code. Regarding https://github.com/tomv564/LSP/issues/482 I can add flag if needed

plugin/hover.py
ayoub-benali: @tomv564  so you are saying `is_at_word` check can be ignore and the client capability check is enough ?

ayoub-benali:
@tomv564 are you happy with the changes ?

plugin/hover.py
tomv564: Changes are good to the first part of my comment!

But did you see that we already check if the cursor is at a symbol? Is there a need to check it again here?

ayoub-benali: @tomv564 you are right, no need to check

tomv564:
Thanks, great work!

PR #570, opened 2019-04-18
Change flake8 rules formatting to make compatible with pyls

rchl:
Pyls language server is not capable of parsing previous formatting of
flake8 configuration. It does its own parsing that expects rule codes
to be separated by comma.


tomv564:
Thanks! (there was a "rebase and merge" button for me)

PR #571, opened 2019-04-20
Add configuration for Metals and update Scala doc

ayoub-benali:
Add config for Scala's new LSP server, [Metals](https://scalameta.org/metals/)

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -203,6 +203,13 @@
       }
       ]
     },
+    "Metals":
+    {
+      "command": ["metals-sublime"],
+      "languageId": "scala",
+      "scopes": ["source.scala"],
+      "syntaxes": ["Packages/Scala/Scala.sublime-syntax"]
+    },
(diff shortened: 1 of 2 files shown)


ayoub-benali:
It was always there, it is only setting a bunch of flags. Now that the server is mature I think it is safe having Metals by default.

@olafurpg Once this released we can update the doc here: https://github.com/scalameta/metals/blob/master/docs/editors/sublime.md#installing-the-plugin

LSP.sublime-settings
```diff
@@ -203,6 +203,13 @@
       }
       ]
     },
+    "Metals":
```
tomv564: There was an unspoken convention of lowercase names here, as it was only for looking up configs/clients case-sensitively by name in code. 
I understand it's shown to users now, but is it ok if we stick with lowercase? I'd like to avoid users having issues overriding a "Metals" config with a "metals" key in their user config.

ayoub-benali:
@tomv564 PR updated

PR #577, opened 2019-04-27
Text editing cleanup/testability

tomv564:
Separate sublime-tied logic to plugin/edit.py.
Add tests for remaining logic (response parsing, sorting)
Remove debugging status/logging when applying doc edits.

Part of an effort to improve unit test coverage.

### DIFF
--- a/plugin/core/edit.py
+++ b/plugin/core/edit.py
@@ -1,55 +1,41 @@
-import sublime
-import sublime_plugin
-
+from .url import uri_to_filename
 try:
-    from typing import List, Dict, Optional, Any, Iterable
-    assert List and Dict and Optional and Any and Iterable
+    from typing import List, Dict, Optional, Any, Iterable, Tuple
+    TextEdit = Tuple[Tuple[int, int], Tuple[int, int], str]
+    assert List and Dict and Optional and Any and Iterable and Tuple
 except ImportError:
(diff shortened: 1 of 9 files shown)


plugin/formatting.py
```diff
@@ -18,6 +18,11 @@ def options_for_view(view: ViewLike) -> 'Dict[str, Any]':
     }
 
 
+def apply_reponse_to_view(response, view):
```
rwols: typo: reponse -> response

plugin/core/edit.py
```diff
@@ -1,34 +1,25 @@
...
-            for uri, file_changes in changes.items():
-                path = uri_to_filename(uri)
-                self.open_and_apply_edits(path, file_changes)
-                documents_changed += 1
-        elif document_changes:
-            for document_change in document_changes:
-                uri = document_change.get('textDocument').get('uri')
-                path = uri_to_filename(uri)
-                self.open_and_apply_edits(path, document_change.get('edits'))
+            for (document_uri, document_changes) in changes.items():
```
rwols: There is no need to use parentheses around the key/value pair here.

plugin/formatting.py
tomv564: Thanks for looking!

PR #584, opened 2019-05-02
Move parsing and formatting of signature help to core with tests

tomv564:
To do:
- [x] clean up
- [x] add parameter/signature documentation fields to test data
- [x] test case around overloads
- [x] test case around parameter highlighting


rwols:
Really nice work here separating the rendering of the sighelp from the ST API logic.

tomv564:
Thanks! If you want to help, diagnostics and references have a fair bit of extractable code as well!

PR #587, opened 2019-05-08
stdio transport fix

bscottm:
Give the stdio transport a bit of love, fix a fairly subtle bug that might have mistaken an empty string as EOF, make the processing loop state-driven like the TCP transport.

The LSP protocol does insert a blank line between the headers to signal where the content starts. There was a subtle bug in the stdio processing loop that could have mistaken a blank line as EOF, which is what `readline()` returns when it reads EOF.

I came across this while debugging an issue in `haskell-ide-engine`.

### DIFF
--- a/plugin/core/transports.py
+++ b/plugin/core/transports.py
@@ -39,7 +39,14 @@ class Transport(object, metaclass=ABCMeta):
 
 STATE_HEADERS = 0
 STATE_CONTENT = 1
+STATE_EOF = 2
 
+StateStrings = { STATE_HEADERS: 'STATE_HEADERS'
+               , STATE_CONTENT: 'STATE_CONTENT'
+               , STATE_EOF:     'STATE_EOF'}
+
+def state_to_string(state: int) -> None:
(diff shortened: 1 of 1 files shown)


tomv564:
Thanks for the bug fix -  I'll try to read this closely in the upcoming days!
Probably want to comment out or remove the `debug` statement.
Do you have any more details on how you found it? Curious if we got lucky with all other language servers or if people are seeing this with other languages as well!

bscottm:
I suspect that very few language servers fork/spawn and talk over pipes (vice network over localhost.) I was debugging an issue with `haskell-ide-engine`, which does prefer to work over a pipe. Turns out that the problem wasn't LSP, but the `stdio` code looked like it hadn't been touched compared to the TCP transport. For example, there are other headers that could be sent that LSP ignores. The blank line is the only way the code knows when headers are done and to slurp the content payload, which wasn't really being checked in the original code.

I'm happy to comment out the debug statements. I usually leave them in for anyone who has to diagnose an issue that I've created.

tomv564:
Thanks for the cleanups!

PR #595, opened 2019-05-26
Add underline in active parameter

predragnikolic:
Hello,

I just noticed that we now have syntax highlighting and active parameter in signature help, which is great! :)  (although I don't use a very colorful colorscheme)

But I had some problems with determining visually what the active parameter is. 
Here are a few screenshots... By first glance it is hard to tell what is the active parameter, right?
Screen Shot 2019-05-26 at 12 07 36 PM
Screen Shot 2019-05-26 at 12 06 27 PM
Screen Shot 2019-05-26 at 12 06 43 PM

With this PR, I just add a underline. The result is that you can immediately see what the active parameter is.
Screen Shot 2019-05-26 at 12 08 27 PM
Screen Shot 2019-05-26 at 12 05 48 PM

### DIFF
--- a/plugin/signature_help.py
+++ b/plugin/signature_help.py
@@ -38,9 +38,9 @@ class ColorSchemeScopeRenderer(object):
 
     def _wrap_with_scope_style(self, content: str, scope: str, emphasize: bool = False, escape: bool = True) -> str:
         color = self._scope_styles[scope]["color"]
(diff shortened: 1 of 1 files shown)


ayoub-benali:
This a great ergonomic improvement !

tomv564:
Nice fix, it definitely is needed with color schemes like yours, using white text in the signature!

Issue #443, opened 2018-10-28
Support configuring child projects in single workspace


tomv564:
As suggested on gitter by @plasticine:

>For example, say I have some ruby code in services/app and some JS code in services/ui, and then a bunch of top level stuff in the project root; config files, scripts and other junk.

>How can I used something like solargraph for the ruby codebase in services/app, and flow for the code in services/ui? Flow for example expects to find a .flowconfig file in the root from which it’s run, and it absolutely doesn’t make sense to put that in the root of the project.

Some servers don't mind starting in an arbitrary folder and detecting projects in the tree (eg. the typescript server just looks for package.json).

>I do know that VSCode extensions often support projectPath or workingDirectories config keys that would allow for this.

We should read the docs to understand how language servers are configured - per child project? One process or processes for each child project?

plasticine:
@tomv564 Hey 👋🏼 I had a bit of a stab at seeing if I could get this running. It’s super hacked in for the moment (apologies, it’s been a long time since I wrote Python!), but it is working.

https://github.com/tomv564/LSP/compare/master...plasticine:spike/client-path

``` json
{
	"folders":
	[
		{
			"path": "."
		}
	],
	"settings": {
		"LSP": {
			"flow": {
				"enabled": true,
				"command": ["flow", "lsp"],
				"languageId": "javascript",
				"client_path": "services/ui",
				"scopes": ["source.js"],
				"syntaxes": [
					"Packages/Babel/JavaScript (Babel).sublime-syntax",
					"Packages/JavaScript/JavaScript.sublime-syntax"
				]
			}
		}
	}
}

```

tomv564:
I also _just_ closed https://github.com/tomv564/LSP/pull/436 which is the same solution, but then used to workaround a broken language server.

I'd like to understand the monorepo use cases a bit better before merging solutions.
This setting works for one-project-per-language workspaces.
But if you have a monorepo with multiple flow projects, how should LSP accommodate the flow language server?

plasticine:
@tomv564 Yeah for sure — I assumed as much which is why I didn’t bother with a PR. I’ll just maintain my fork for a while until a direction is decided upon I guess.

FWIW, it seems to me that `client_path` (if set) could just be a part of tracking the session, which would allow for multiple clients of the same type in a tree, right?

plasticine:
Also, as reference, here’s two VSCode examples that have this sort of pattern that I’m personally familiar with;

[...]

PR #601, opened 2019-06-04
Allow LSP clients to operate in project sub-paths

plasticine:
This PR addresses #443.

I’ve been using this patch for quite some time, but figured I’d contribute it here and see what people think. It’s pretty naive and hacked in but if there is interest I’m keen to get this shippable. I think there was some resistance to this approach initially so I’m keen to gauge interest before I go too deep.

It allows for configuring LSP client to operate at sub-paths, for example the following project config would configure the flow client 

```json
{
  "settings": {
    "LSP": {
      "flow": {
        "enabled": true,
        "client_path": "vendor/frontend"
      }
    }
  }
}
```

### DIFF
--- a/plugin/core/clients.py
+++ b/plugin/core/clients.py
@@ -38,6 +38,7 @@ def get_window_env(window: sublime.Window, config: ClientConfig) -> 'Tuple[List[
 
 def start_window_config(window: sublime.Window,
                         project_path: str,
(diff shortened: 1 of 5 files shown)


tomv564:
Closing this as it's getting out of date - I hope the recently added multi-folder support can help the use case!

Issue #536, opened 2019-03-21
bug in auto completion when edit behind the completion position


ayoub-benali:
When autocomplete has to edit before the completion position the trimming introduced in https://github.com/tomv564/LSP/pull/295 leads to a bug, see [line](https://github.com/tomv564/LSP/blob/master/plugin/completion.py#L296) 

You can see an example bellow using Scala with [Metals](https://scalameta.org/metals/):
![broken-autocomplete](https://user-images.githubusercontent.com/1632384/54742160-dde3ab00-4bc0-11e9-9f73-a9078ddecec7.gif)

Here is the `CompletionItem` sent back by the server:
```
{
	'insertTextFormat': 2,
	'data': {
		'symbol': 'example/Foo#myFunction().',
		'target': 'file:/home/ayoub/workspace/testproject/?id=root'
	},
	'detail': 'override def myFunction(): Unit',
	'sortText': '00000',
	'filterText': 'override def myFunction',
	'preselect': True,
	'label': 'override def myFunction(): Unit',
	'kind': 2,
	'additionalTextEdits': [],
	'textEdit': {
		'newText': 'override def myFunction(): Unit = ${0:???}',
		'range': {
			'start': {
				'line': 7,
				'character': 2
			},
			'end': {
				'line': 7,
				'character': 18
			}
		}
	}
}
```

### Expected behavior
Before:
```scala
def myF<ctrl+space>
```
After:
```scala
override def myFunction(): Unit = ???
```

Please note the bug doesn't happen when the prefix is fully specified:
![broken-autocomplete-2](https://user-images.githubusercontent.com/1632384/54742571-189a1300-4bc2-11e9-9c9a-77b77158bee6.gif)

* OS and language server: Ubuntu 18.04
* How you installed LSP (Package Control or from git?): Package Control version 0.8.2

Any ideas on how to fix this kind of issues ? I am willing to make a PR but first would like to get your thoughts

cc @nanoant

ayoub-benali:
One idea could be that we:
1. remove the trimming applied [here](https://github.com/tomv564/LSP/blob/master/plugin/completion.py#L296) 
2. after completion we add extra logic in [CompletionSnippetHandler](https://github.com/tomv564/LSP/blob/master/plugin/completion.py#L67) to remove any leftovers. 

We know:
- the last prefix 
- the range of the replacement present in `textEdit` 
- that Sublime replaces only the last prefix and not the whole range specified in `textEdit`  

[...]

Issue #529, opened 2019-03-12
additionalTextEdits support in auto completion


kadircet:
I believe this issue was also mentioned in #294 but it was optional so it didn't get addressed(if there were any other reasons, I couldn't see within the issue itself).

LSP supports `additionalTextEdits` field in completion items, and at least clangd makes use of it to provide user with include insertions. I was wondering if there was any particular reason for not supporting this. I believe it could be added in a similar way to resolve support.

rwols:
It is a somewhat daunting task to start implementing this, because the ST API is simply not capable of arbitrary text replacements for a completion item. An initial (naive) thought on how to implement this would require a callback for which completion item the user selected from the ST completion widget, but currently we cannot attach a callback to that event. Thus we're limited to providing the ST completion widget one single textEdit (as its API demands).

rwols:
I hope Jon and Will can one day find the time to re-think the completion widget a-la Atom/VSCode.

kadircet:
I see that auto completion lacks the functionality to do arbitrary insertions but it seems like you've a way to perform actions on the `view` after a completion item has been selected (https://github.com/tomv564/LSP/blob/master/plugin/completion.py#L110).

And it seems like there are replace/insert/erase functionalities at arbitrary positions https://www.sublimetext.com/docs/3/api_reference.html#sublime.View

Why these two cannot be combined?

predragnikolic:
I think this is doable.

Here is an example of what I think would work...
```diff
    def handle_resolve_response(self, response, view):
-        # replace inserted text if a snippet was returned.
-        if current_completion and response.get('insertTextFormat') == 2:  # snippet
-            insertText = response.get('insertText')
-            try:
-                sel = view.sel()
-                sel.clear()
-                sel.add(current_completion.region)
-                view.run_command("insert_snippet", {"contents": insertText})
-            except Exception as err:
-                exception_log("Error inserting snippet: " + insertText, err)

def handle_resolve_response(self, response):
+    view.run_command("example", {"response": response})

+ class ExampleCommand(sublime.TextCommand):
+    def run(self, edit, response):
+       # handle the response there
```

The advantage of calling our own TextCommand, 
instead of running `insert_snippet` command 
is that you get an edit token. With it you can `insert`, `erase`, `replace`.

rwols:
[...]

PR #602, opened 2019-06-05
Completion with text edits

tomv564:
Sets up a post-completion hook so we can:
* Fix the completion using the item's textEdit (issue #536)
* Apply additionalTextEdits (issue #529)

`completionItem/resolve` is used to get the additionalTextEdits if the server indicates support.

Code is tested using examples given in issues, but more live testing with specific servers is needed!
Assumptions:
* If a command like `'commit_completion', 'insert_best_completion', 'auto_complete'` runs, the next `on_modified` is the insertion of the completion.
* It is always safe to trim the difference between the start of the current word and the textEdit's start.
* All additionalTextEdits should be immediately applied after the completion is inserted.

To do:
- [x] Get feedback from clangd , metals and others?
- [x] Clean up code
- [x] Look for holes / add tests
- [x] UX considerations

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -52,6 +52,7 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
         self.initialized = False
         self.enabled = False
         self.trigger_chars = []  # type: List[str]
+        self.resolve = False
         self.state = CompletionState.IDLE
         self.completions = []  # type: List[Any]
         self.next_request = None  # type: Optional[Tuple[str, List[int]]]
@@ -74,6 +75,7 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
                 'completionProvider')
             if completionProvider:
                 self.enabled = True
+                self.resolve = completionProvider.get('resolveProvider') or False
                 self.trigger_chars = completionProvider.get(
                     'triggerCharacters') or []
                 if self.trigger_chars:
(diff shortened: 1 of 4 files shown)


ayoub-benali:
Thanks @tomv564 for taking a stab a this. I just made a test with Metals and both issues seems to be fixed :tada: 

autocomplete

@olafurpg Do you have other use cases were this might still break ?

olafurpg:
This looks great! Nice job 👏 

The only example I can think of is when `textEdit` spans multiple lines. This is prohibited by the spec and VS Code filters out such completions. In Metals we can still sometimes returns those, for example

```scala
// before
val name = "John"
val message = """
Hello $nam<COMPLETE>
"""
// after
val name = "John"
val message = s"""
Hello $name
"""

```

ayoub-benali:
@tomv564 Have you considered saving the `additionalTextEdits` contained in `CompletionItem` to apply them right after instead of doing a roundtrip via `completionItem/resolve` ? 
LSP could be sending `additionalTextEdits` in the initial completion and not support  `completionItem/resolve`

tomv564:
Thanks for testing this so quickly, great to see your demo. Forgot to mention here, but @ayoub-benali your suggestions in #536 were key to getting this done!

In your example it's pretty clear that an import occurred, but imports added to the top won't be visible when you're two pages down. What do you think of showing a popup briefly (or less visibly, a message in the status bar)?

ayoub-benali:
> In your example it's pretty clear that an import occurred, but imports added to the top won't be visible when you're two pages down. What do you think of showing a popup briefly (or less visibly, a message in the status bar)?

I think a status bar message is fine because in my example an import was added but in general we don't know which kind of text edits are sent by the server

ayoub-benali:
> This looks great! Nice job 
> 
> The only example I can think of is when `textEdit` spans multiple lines. This is prohibited by the spec and VS Code filters out such completions. In Metals we can still sometimes returns those, for example
> 
> ```scala
> // before
> val name = "John"
> val message = """
> Hello $nam<COMPLETE>
> """
> // after
> val name = "John"
> val message = s"""
> Hello $name
> """
> ```

completion inside double quoted text isn't even triggered by sublime text see https://github.com/SublimeTextIssues/Core/issues/2793

Something has to be on sublime side before we can test this case

tomv564:
@ayoub-benali LSP should already do what you suggested, the logic is roughly
```python
if item.additionalTextEdits:
  apply_additional_edits(item.additionalTextEdits)
elif resolveProvider:
  resolve_completion_item(item)
```

ayoub-benali:
Awesome !

Issue #599, opened 2019-06-02
[Request] Add option for character limit in autocompletion


LiteCatDev:
This so far is working awesome. But there is something missing.

Again, using Dart Language Server the autocompletion popup shows, but the description is so big that it goes out of the screen [since i'm using a big font size] . It would be great if in the configuration you could set a character limit for the description text and just add "..." if the description is too big. I have implemented something like this but since i don' t know python that well, i don't have the courage to make a pull request 

My implementation is this one 

In the file LSP/plugin/core/completion.py line 37 i think

> if len(replacement) > 0 and replacement[0] == '$':  # sublime needs leading '$' escaped.
>         replacement = '\\$' + replacement[1:]
>     //if len(trigger) + len (hint) > 80:     <- this is checking if it exceeds the character limit
>       //  hint = hint[0:80 - len(trigger)]  <-this changes the hint so it can't be greater that the character limit
>     return "\t  ".join((trigger, hint)) if hint else trigger, replacement

again this is only an idea, it is of no importance , just customization

tomv564:
Can you share a screenshot of what it looks like now? 

As over time we've accumulated many settings for many preferences, I'd like to instead find a solution that works for all users. Perhaps `kind` is a better default for the `completion_hint_type`?

tomv564:
Can re-open if this is a common issue, but so far only 1 report with Dart language server, and a workaround is available with the `completion_hint_type` setting.

PR #603, opened 2019-06-06
Added "completion_maxsize" property

LiteCatDev:
Again, using dart language server, the "hints" are going of my screen , so i dug a bit and found that i can reduce the hint lenght  to a limit .

By default it's disabled  so it is not noticeable until you manually activate it.

Useful for long hints , can be used as a way to hide hints but it's not the purpose. The purpose is to keep the start of a hint without completely hiding it . ot to cut the hint before it goes offscreen


tomv564:
Waiting for a comment in issue #599!

tomv564:
Thanks for the effort, but this PR is far from mergeable - looks like you've merged in from your fork and from this repo?
Anyways, `completion_hint_type` already allows you to select a shorter field to use in the completion hints.

Issue #596, opened 2019-05-27
support for lua-language-server


AmjadHD:
I know this was discussed before in #322 but this implementation of the [lua-language-server](https://github.com/sumneko/lua-language-server) is actively maintained, and have much more features than [lua-lsp](https://github.com/Alloyed/lua-lsp), you can download the vscode extension from [here](https://marketplace.visualstudio.com/_apis/public/gallery/publishers/sumneko/vsextensions/lua/0.9.7/vspackage)
```json
"lua-ls":
        {
            "command":
            [
                "D:/Downloads/Archives/sumneko.lua-0.9.7/extension/server/bin/lua-language-server",
                "-E",
                "D:/Downloads/Archives/sumneko.lua-0.9.7/extension/server/main.lua"
            ],
            "enabled": true,
            "languageId": "lua",
            "scopes":
            [
                "source.lua",
            ],
            "syntaxes":
            [
                "Packages/Lua/Lua.sublime-syntax"
            ]
        },
```
see this [issue](https://github.com/sumneko/lua-language-server/issues/39) for further highlights.

tomv564:
@AmjadHD I fixed the error you mentioned in pull request #567 (would have liked a new issue for it instead, but thanks for reporting!)

Neither lua server is in the docs, perhaps you can add a section for lua linking to the two servers?

PR #634, opened 2019-06-30
Add lua-ls server

AmjadHD:
#596


rwols:
By the way what is up with the travis build script?

PR #638, opened 2019-07-02
Some minor mostly stylistic changes

AmjadHD:
Sorry for these OCD-ish changes, but I wish to have a good level of consistency in this project.

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -39,11 +39,9 @@ class CompletionHelper(sublime_plugin.EventListener):
 
 class LspTrimCompletionCommand(sublime_plugin.TextCommand):
 
-    def run(self, edit, range: 'Optional[Tuple[int, int]]'=None):
+    def run(self, edit, range: 'Optional[Tuple[int, int]]' = None):
         if range:
-            start, end = range
-            region = sublime.Region(start, end)
-            self.view.erase(edit, region)
+            self.view.erase(edit, sublime.Region(*range))
 
 
 class CompletionHandler(sublime_plugin.ViewEventListener):
@@ -65,7 +63,7 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
     @classmethod
     def is_applicable(cls, settings):
         syntax = settings.get('syntax')
-        return is_supported_syntax(syntax, client_configs.all) if syntax else False
+        return syntax is not None and is_supported_syntax(syntax, client_configs.all)
 
     def initialize(self):
         self.initialized = True
@@ -108,9 +106,8 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
             self.view.settings().set('auto_complete_triggers', completion_triggers)
 
     def is_after_trigger_character(self, location):
-        if location > 0:
-            prev_char = self.view.substr(location - 1)
-            return prev_char in self.trigger_chars
+        prev_char = self.view.substr(location - 1)
+        return location > 0 and prev_char in self.trigger_chars
 
     def is_same_completion(self, prefix, locations):
         if self.response_incomplete:
@@ -142,13 +139,14 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
 
     def on_modified(self):
 
+        current_location = self.view.sel()[0].begin()
         # hide completion when backspacing past last completion.
-        if self.view.sel()[0].begin() < self.last_location:
+        if current_location < self.last_location:
             self.last_location = 0
             self.view.run_command("hide_auto_complete")
 
         # cancel current completion if the previous input is a space
-        prev_char = self.view.substr(self.view.sel()[0].begin() - 1)
+        prev_char = self.view.substr(current_location - 1)
         if self.state == CompletionState.REQUESTING and prev_char.isspace():
             self.state = CompletionState.CANCELLING
 
@@ -165,8 +163,7 @@ class CompletionHandler(sublime_plugin.ViewEventListener):
     def on_completion_inserted(self):
         # get text inserted from last completion
         word = self.view.word(self.last_location)
-        region = sublime.Region(word.begin(), self.view.sel()[0].end())
-        inserted = self.view.substr(region)
+        inserted = self.view.substr(sublime.Region(word.begin(), self.view.sel()[0].end()))
 
         item = self.find_completion_item(inserted)
         if item:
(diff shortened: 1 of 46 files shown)


rwols:
Please submit the stylistic changes in smaller chunks and single topics. For instance: you decided that writing `some_dict = {}` everywhere looks better than writing `some_dict = dict()` everywhere (I agree!). Write a pull request for that single topic.

AmjadHD:
It'll take a lot of time to create PRs for all those changes, I instead divided them to smaller commits if you don't mind.

plugin/completion.py
```diff
@@ -162,13 +154,12 @@ def on_modified(self):
                 if self.response_incomplete:
                     # debug('incomplete, triggering new completions')
                     self.view.run_command("hide_auto_complete")
-                    sublime.set_timeout(self.run_auto_complete, 0)
```
rwols: This might seems strange, but what this does is it makes sure that `self.run_auto_complete` is ran from Sublime's main thread. I am unsure wether that is necessary though.

Remember that these "handle response" methods always run on a Python thread.

plugin/core/signature_help.py
```diff
@@ -171,31 +169,30 @@ def __init__(self, signatures: 'List[SignatureInformation]',
     def build_popup_content(self, renderer: ScopeRenderer) -> str:
         parameter_documentation = None  # type: Optional[str]
 
-        formatted = []
+        formatted = ""
```
rwols: joining a list of strings is "significantly faster": https://stackoverflow.com/questions/3055477/how-slow-is-pythons-string-concatenation-vs-str-join

plugin/core/windows.py
```diff
@@ -450,7 +435,7 @@ def _apply_workspace_edit(self, params: 'Dict[str, Any]', client: Client, reques
     def _handle_session_started(self, session, project_path, config):
         client = session.client
         client.set_crash_handler(lambda: self._handle_server_crash(config))
-        client.set_error_display_handler(lambda msg: self._window.status_message(msg))
```
rwols: Removing the unnecessary lambdas everywhere is nice!

plugin/diagnostics.py
```diff
@@ -322,7 +324,7 @@ def update_diagnostics_panel(window: sublime.Window):
             panel.settings().set("result_base_dir", base_dir)
 
             auto_open_panel = False
-            to_render = []
+            to_render = ""
```
rwols: joining a list of strings is "significantly faster": https://stackoverflow.com/questions/3055477/how-slow-is-pythons-string-concatenation-vs-str-join

plugin/hover.py
```diff
@@ -112,15 +112,12 @@ def diagnostics_content(self, diagnostics):
         by_severity = {}  # type: Dict[int, List[str]]
         for diagnostic in diagnostics:
             by_severity.setdefault(diagnostic.severity, []).append(self.format_diagnostic(diagnostic))
-        formatted = []
+        formatted = ""
```
rwols: joining a list of strings is "significantly faster": https://stackoverflow.com/questions/3055477/how-slow-is-pythons-string-concatenation-vs-str-join

plugin/rename.py
```diff
@@ -27,12 +27,7 @@ def validate(self, name):
         return len(name) > 0
 
     def get_current_symbol_name(self):
-        pos = get_position(self.view)
-        current_name = self.view.substr(self.view.word(pos))
-        # Is this check necessary?
-        if not current_name:
-            current_name = ""
-        return current_name
+        return self.view.substr(self.view.word(get_position(self.view)))
```
rwols: Have you checked wether `view.substr` always returns a `str`?

plugin/core/signature_help.py
AmjadHD: https://stackoverflow.com/a/3055558/8993294 shows the opposit.

plugin/completion.py
AmjadHD: what happens if it doesn't run in sublime's main thread ?

plugin/rename.py
AmjadHD: Yes, I tested it in different edge cases, like `Region(0)`, `Region(-1)`, `Region(view.size() + 100)`, `0`, `-3` ...
additionally The ST API already states that they return a `str`
> substr(region) | str | Returns the contents of the region as a string.
> -- | -- | --
> substr(point) | str | Returns the character to the right of the point.

plugin/core/signature_help.py
rwols: Let's ignore the stackoverflow crowd and do the experiment ourselves. In my Sublime Text console I do  the following:

```
>>> from timeit import timeit as f
>>> f('x="".join([str("asdfasdfasdf") for i in range(0, 100)])')
17.87444483214216
>>> f('x="".join(["asdfasdfasdf" for i in range(0, 100)])')
4.023197690061409
>>> f('x=""\nfor i in range(0, 100): x += "asdfasdfasdf"')
13.988012291068856
>>> f('x=""\nfor i in range(0, 100): x += str("asdfasdfasdf")')
28.351181478097544
```
The `"".join([])` variant seems fastest. However, it becomes slower once we use a call to `str`. I'm not sure why. The `+=` variant also becomes slower when wrapping it in `str`.

plugin/completion.py
rwols: I don't fully understand the consequences of that. So just to be safe, my suggestion is to revert this line.

AmjadHD: Ok I've reverted this line

plugin/core/signature_help.py
AmjadHD: Ok I reverted that in f0587379784f496d017022a86223f4c189a3f82e

tomv564:
46 files modified, and then "Try to make tests pass" :scream:

I understand there's a lot of low-hanging fruit in the codebase, but as with the documentation PR I think there are more important things to do for the end-user.

If we want consistency, do you have some suggestions on how to set up a linter and autoformatter? That way we can make it easy for contributors to keep the code clean.

Alternatively, I think it's great to help keep the code clean by participating in reviews and doing some cleanups while updating/bugfixing a feature.

AmjadHD:
> 46 files modified, and then "Try to make tests pass" 😱

Yeah, It got really messy 🤣, that's my fault, I deviated from the original purpose of this PR to modifying code logic but I reverted the changes that made tests fail. You can close this if you want, setting up an autoformatter and linter is probably better, but I don't think that would remove those ugly dict(...) and list(filter(map(lambda...))) or unused lambdas.

tomv564:
Please open a new issue for auto formatters and standards if you want to drive it forward! Closing this now.

PR #640, opened 2019-07-02
Show a list of available LSP commands

AmjadHD:
this should prove helpful, as it reduces the need for many shortcuts.

### DIFF
--- a/Keymaps/Default (Linux).sublime-keymap
+++ b/Keymaps/Default (Linux).sublime-keymap
@@ -1,13 +1,14 @@
 [
     // Show a list of available LSP commands
-    {
-        "keys": ["f1"],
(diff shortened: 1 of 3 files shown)


rwols:
If people want to bind <kbd>F1</kbd> to a command I think they can do that themselves. I am not too keen on Sublime plugins providing default keybindings, as probably someone out there already has <kbd>F1</kbd> bound to something that he/she deems important.

For instance, the binding of <kbd>CTRL</kbd><kbd>.</kbd> stood in the way of someone's workflow: https://github.com/tomv564/LSP/issues/620

rwols:
Could you force-push your commit? I think there was something wrong with the travis install script during the great Bitbucket API change.

Issue #645, opened 2019-07-08
[Completion Regression] text_edit_text returns None constantly 


ayoub-benali:
Not sure when the regression happened but this column check:
```python
            if edit_range.start.col <= word_col:
                # if edit starts at current word, we can use it.
                # if edit starts before current word, use the whole thing and we'll fix it up later.
                return edit_text
```
[here](https://github.com/tomv564/LSP/blob/master/plugin/core/completion.py#L76) returns always False now. Which leads the completion to fallback to `label` instead of using `textEdit`. As you can see bellow:

![autocomplete-3](https://user-images.githubusercontent.com/1632384/60817794-a3684300-a19c-11e9-9050-90a7dac2988e.gif)

Logs with some debug variables:
```
item: {'textEdit': {'newText': 'apply($0)', 'range': {'end': {'line': 3, 'character': 7}, 'start': {'line': 3, 'character': 7}}}, 'label': 'apply[A](xs: A*): List[A]', 'sortText': '00000', 'preselect': True, 'insertTextFormat': 2, 'filterText': 'apply', 'data': {'symbol': 'scala/collection/immutable/List.apply().', 'target': 'file:/home/ayoub/workspace/testproject/?id=root'}, 'kind': 2}

LSP start: {'line': 3, 'character': 7}
_last_row: 3, last_col: 6
LSP: textEdit from col 7, 7 applied at col 6
edit_range.start.col: 7, word_col: 6
```

For me the condition [here](https://github.com/tomv564/LSP/blob/master/plugin/core/completion.py#L76) is incomplete and doesn't handle the case when the completion happens after last word beginning. In this case a `.` 

I think we are safe to remove the condition. 

I can make a PR but wanted first to confirm the issue with @tomv564 

* OS and language server: Ubuntu with [Metals](https://scalameta.org/metals/)
* How you installed LSP: [Master](https://github.com/tomv564/LSP/commit/3722f0b5c26ecbd3768502ee62f562ee57d98cc1)
* `prefer_label_over_filter_text` is set to True

tomv564:
Thanks for reporting. I think `word_col` is incorrectly provided to `format_completion` by **plugin/completion.py**  

The logic there looks up the start of the current word being completed. In your example sublime tells us that the `.` is a word - we should probably only use the start-of-word lookup if the cursor is at a word, see `is_at_word`.  

Can have a look at a fix for this soon, or feel free to open a PR.

ayoub-benali:
[...]

PR #653, opened 2019-07-17
Add failing test for textedit on non-word completion

tomv564:
For issue #645 

TextEdits from completion items are being skipped because LSP incorrectly calculates the "current" completing word.
In this case it works: `List.a|` -> word_start is char 6
In this case it fails: `List.|` -> word_start is char 5

The fix is to check if the cursor position _is_ at a word to begin with.


ayoub-benali:
Thank you !

PR #658, opened 2019-07-19
Do not escape hover content

### LINKED ISSUE #657: hover regression
After the latest commits hover popups produce errors, I suspect c9a73907f00972a415e8e979d5dcc96c11ec3f09 is responsible for this
* OS: Windows 10
* language server: pyls
* How you installed LSP: git
* Minimal reproduction steps
    1. open a .py file
    2. hover over an function name
* Log
```
LSP:  --> textDocument/hover
LSP:      {'contents': "is_supported_syntax(syntax: str, configs: 'List[ClientConfig]') -> bool"}
Parse Error: #x27;List[ClientConfig]&#x27;) -&gt; bool</p><p><a href='definition'>Definition</a> | <a href='references'>References</a> | <a href='rename'>Rename</a></p></div></div> code: Unexpected character
Parse Error: 27;List[ClientConfig]&#x27;) -&gt; bool</p><p><a href='definition'>Definition</a> | <a href='references'>References</a> | <a href='rename'>Rename</a></p></div></div> code: Unexpected character
Parse Error: 7;List[ClientConfig]&#x27;) -&gt; bool</p><p><a href='definition'>Definition</a> | <a href='references'>References</a> | <a href='rename'>Rename</a></p></div></div> code: Unexpected character
Parse Error: #x27; code: Unknown entity
Parse Error: #x27;) -&gt; bool</p><p><a href='definition'>Definition</a> | <a href='references'>References</a> | <a href='rename'>Rename</a></p></div></div> code: Unexpected character
Parse Error: 27;) -&gt; bool</p><p><a href='definition'>Definition</a> | <a href='references'>References</a> | <a href='rename'>Rename</a></p></div></div> code: Unexpected character
Parse Error: 7;) -&gt; bool</p><p><a href='definition'>Definition</a> | <a href='references'>References</a> | <a href='rename'>Rename</a></p></div></div> code: Unexpected character
Parse Error: #x27; code: Unknown entity
```

rwols:
Close #657.

~Otherwise mdpopups will complain.~
~Also add two hover tests to cement this behavior.~

cc @LoneBoco

### DIFF
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -139,9 +139,9 @@ class LspHoverCommand(LspTextCommand):
             value = ""
             language = None
             if isinstance(item, str):
-                value = escape(item)
+                value = item
             else:
-                value = escape(item.get("value"))
+                value = item.get("value")
                 language = item.get("language")
             if language:
                 formatted.append("```{}\n{}\n```\n".format(language, value))


jfcherng:
image

PHP seems not to be fixed by this PR.

rwols:
You're right. I also noticed in clangd that template parameters are escaped too many times now. Let us revert this small change and think of another solution.

LoneBoco:
Is the problem the fact that some things are returning HTML entities and by escaping, we turn `&lt;` into `&amp;lt;`?  I've been reading some of the Markdown specs some more (there is no one Markdown spec at this time, which makes everything really complicated), and it seems that it is probably safest to ensure that the LSP servers themselves are encoding all angle brackets and ampersands.  It is probably best to just rollback #650 and I can work with the Omnisharp team to end up with something that works well with LSP plugins.

PR #659, opened 2019-07-20
Add support for all the modern "goto" requests

### LINKED ISSUE #636: feature request: goto declaration in hover
when i click definition in the hover it goes to the implementation. sublime in turn with "go to definition" actually goes to the declaration. can we have both in the hover, as declaration/implemntation maybe?

rwols:
- Add support for textDocument/typeDefinition
- Add support for textDocument/declaration
- Add support for textDocument/implementation

close #636

### DIFF
--- a/tests/test_goto.py
+++ b/tests/test_goto.py
@@ -1,0 +1,72 @@
+from setup import TextDocumentTestCase, close_test_view
+import sublime
+from os.path import dirname
+
+TEST_FILE_PATH = dirname(__file__) + "/goto.txt"
+
+RESPONSE = [
+    {
+        'uri': TEST_FILE_PATH,
+        'range':
+        {
+            'start':
+            {
+                'character': 2,
+                'line': 0
+            },
+        }
+    }
+]
+
+
+class GotoTestCase(TextDocumentTestCase):
+
+    def setUp(self):
+        super().setUp()
+        yield 100
+        self.view.run_command('insert', {"characters": 'hello there'})
+        yield 100
+        self.goto_view = sublime.active_window().open_file(TEST_FILE_PATH)
+        yield 100
+        self.goto_view.run_command('insert', {'characters': 'foo'})
+        yield 100
+        self.view.window().focus_view(self.view)  # go back to first view
+        yield 100
+
+    def tearDown(self):
+        close_test_view(self.goto_view)
+        super().tearDown()
+
+
+class LspGotoCommandTests(GotoTestCase):
+
+    def do_common_checks(self):
+        yield 100
+        view = sublime.active_window().active_view()
+        self.checkTrue(view.file_name().endswith("goto.txt"))
+        line1, col1 = view.rowcol(view.sel()[0].a)
+        line2, col2 = view.rowcol(view.sel()[0].b)
+        self.checkEqual(line1, 0)
+        self.checkEqual(col1, 2)
+        self.checkEqual(line2, 0)
+        self.checkEqual(col2, 2)
+
+    def test_definition(self):
+        self.client.responses['textDocument/definition'] = RESPONSE
+        self.view.run_command('lsp_symbol_definition')
+        self.do_common_checks()
+
+    def test_type_definition(self):
+        self.client.responses['textDocument/typeDefinition'] = RESPONSE
+        self.view.run_command('lsp_symbol_type_definition')
+        self.do_common_checks()
+
+    def test_declaration(self):
+        self.client.responses['textDocument/declaration'] = RESPONSE
+        self.view.run_command('lsp_symbol_declaration')
+        self.do_common_checks()
+
+    def test_implementation(self):
+        self.client.responses['textDocument/implementation'] = RESPONSE
+        self.view.run_command('lsp_symbol_implementation')
+        self.do_common_checks()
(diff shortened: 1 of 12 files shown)


tests/test_goto.py
```diff
@@ -0,0 +1,72 @@
...
+    def tearDown(self):
+        close_test_view(self.goto_view)
+        super().tearDown()
+
+
+class LspGotoCommandTests(GotoTestCase):
+
+    def do_common_checks(self):
+        yield 100
+        view = sublime.active_window().active_view()
```
tomv564: I think the travis mypy is not checking test files (it perhaps should), but as view is `Optional[View]`, the in-editor mypy is complaining about the unchecked accesses to `view` below - we need an `if view:` or equivalent.
```
       46:25  	mypy        	error     	Item "None" of "Optional[View]" has no attribute "file_name"
       46:25  	mypy        	error     	Item "None" of "Union[str, None, Any]" has no attribute "endswith"
       47:24  	mypy        	error     	Item "None" of "Optional[View]" has no attribute "rowcol"
       47:36  	mypy        	error     	Item "None" of "Optional[View]" has no attribute "sel"
       48:24  	mypy        	error     	Item "None" of "Optional[View]" has no attribute "rowcol"
       48:36  	mypy        	error     	Item "None" of "Optional[View]" has no attribute "sel"
```

tests/test_goto.py
```diff
@@ -0,0 +1,72 @@
...
+        self.checkTrue(view.file_name().endswith("goto.txt"))
```
tomv564: Should be assertTrue?

tests/test_goto.py
```diff
@@ -0,0 +1,72 @@
...
+        line1, col1 = view.rowcol(view.sel()[0].a)
+        line2, col2 = view.rowcol(view.sel()[0].b)
+        self.checkEqual(line1, 0)
```
tomv564: Should be assertEqual?

tests/test_goto.py
```diff
@@ -0,0 +1,72 @@
...
+        self.checkEqual(col1, 2)
+        self.checkEqual(line2, 0)
+        self.checkEqual(col2, 2)
+
+    def test_definition(self):
+        self.client.responses['textDocument/definition'] = RESPONSE
+        self.view.run_command('lsp_symbol_definition')
+        self.do_common_checks()
```
tomv564: I believe you may need to `yield self.do_common_checks()` here to turn `test_definition` into a generator (AKA an async test method). I have a feeling I've tried this before and given up on it, would be awesome if we can get it working for the DRY factor.

tests/test_goto.py
rwols: Huh, I didn't know I could enable mypy with pyls. Thanks for that.

If the `view` and/or `file_name()` is invalid I decided to "skip the test" instead, otherwise I predict annoying sporadic Travis build failures. We could also opt to hard-fail the test instead. I'll let you be the judge.

tests/test_goto.py
rwols: You're totally right. I had made the test for textDocument/definition correctly and then decided to put the common checks into a method. They're most definitely working now with the latest changes.

tests/test_goto.py
```diff
@@ -0,0 +1,72 @@
...
+        yield 100
+        self.view.window().focus_view(self.view)  # go back to first view
+        yield 100
+
+    def tearDown(self):
+        close_test_view(self.goto_view)
+        super().tearDown()
+
+
+class LspGotoCommandTests(GotoTestCase):
```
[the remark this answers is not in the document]
rwols: You're right. I've collapsed this into a single class.

tomv564:
Looking at the latest failures in travis, you might have a race condition between GotoTests opening 'goto.txt' and its base class TextDocumentTestCase opening 'testfile.txt'.

Would it be possible to use 'testfile.txt' for these tests as well? I think overriding 'setUp' is likely causing issues.

Isyielding from setUp handled by DeferredTestCase?

rwols:
> Looking at the latest failures in travis, you might have a race condition between GotoTests opening 'goto.txt' and its base class TextDocumentTestCase opening 'testfile.txt'.

Yeah :( they're working fine locally.

> Would it be possible to use 'testfile.txt' for these tests as well? I think overriding 'setUp' is likely causing issues.

I'll try that tomorrow. The override of setUp works locally, and yielding seems to work locally.

predragnikolic:
@rwols hello,
I just noticed something,
In the initialize request, where the client capabilities are defined,
nowhere did I find that you specified that the client support declaration, typeDefinition, implementation? :) 
https://github.com/tomv564/LSP/blob/67a55b7978651f29d8c7d1e0873c8f20deff8ce1/plugin/core/sessions.py#L90-L92

Shouldn't those capabilities also be defined in the client capabilities or am I missing something?

rwols:
Hmm :thinking: clangd and ccls worked correctly with these changes. I guess those servers advertise `implementationProvider`, `declarationProvider`, etc, without the client advertising them. We should probably advertise that we are able to handle those messages too. Want to make a PR? Do you have a language server that refuses to provide those capabilities?

predragnikolic:
> Want to make a PR? 

Done #672

> Do you have a language server that refuses to provide those capabilities?

No, I just accidentally saw that it wasn't specified in the capabilities.

PR #665, opened 2019-07-25
Handle allowed messages before initialize

### LINKED ISSUE #654: Handle allowed messages from server while initializing.
As would have been helpful in the #613 troubleshooting:

> during the initialize request the server is allowed to send the notifications 
> window/showMessage, window/logMessage and telemetry/event as well as the window/showMessageRequest request to the client.

rwols:
I had to move the MockWindow and MockView to test_sublime.py for this.

close #654

### DIFF
--- a/plugin/core/test_session.py
+++ b/plugin/core/test_session.py
@@ -89,25 +89,25 @@ class SessionTest(unittest.TestCase):
 
     def test_can_get_started_session(self):
         project_path = "/"
-        post_initialize_callbacak = unittest.mock.Mock()
+        post_initialize_callback = unittest.mock.Mock()
         session = self.assert_if_none(
             create_session(config=test_config,
                            project_path=project_path,
                            env=dict(),
                            settings=Settings(),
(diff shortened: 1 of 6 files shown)


rwols:
> Make the caller explicitly call initialize?

Tried this first, got endlessly stuck on tests.

> expose an on_initializing parameter in the Session constructor

Taken this route now. It works!

plugin/core/test_session.py
```diff
@@ -89,39 +89,62 @@ def test_can_create_session(self):
 
     def test_can_get_started_session(self):
         project_path = "/"
-        created_callback = unittest.mock.Mock()
+        post_initialize_callbacak = unittest.mock.Mock()
```
tomv564: small typo: "callbacak"

rwols: Hmm, does this test not run then?

rwols: Never mind, I mindlessly used CTRL+D to change that variable name.

PR #666, opened 2019-07-27
Add settings id

predragnikolic:
So other LSP Handler Plugins settings can be added to the same sublime menu.

Screen Shot 2019-07-27 at 9 38 36 AM


predragnikolic:
Hahahahahahahah PR number #666

PR #667, opened 2019-08-01
Fix new line to break

predragnikolic:
I tried the elm language server, and noticed one thing.
The diagnostics for errors were really hard tor read. They were all concatenated into one big hard to read string. 

So this PR fixes that.

Before
2019-08-01-073810_653x728_scrot

After
2019-08-01-080238_1366x768_scrot

I know  mdpopups has a `nl2br` option, but setting it won't cause the new lines to translate to breaks.
https://github.com/tomv564/LSP/blob/67a55b7978651f29d8c7d1e0873c8f20deff8ce1/plugin/hover.py#L169-L179

So if you know a better way how to fix this, please close this PR :)

### DIFF
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -167,6 +167,7 @@ class LspHoverCommand(LspTextCommand):
         return mdpopups.md2html(self.view, "\n".join(formatted))
 
     def show_hover(self, point, contents):
+        contents = contents.replace('\n', '<br>')
         mdpopups.show_popup(
             self.view,
             contents,
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -119,10 +119,11 @@ class LspHoverCommand(LspTextCommand):
         return "<p>" + " | ".join(actions) + "</p>"
 
     def format_diagnostic(self, diagnostic):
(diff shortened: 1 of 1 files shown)


tomv564:
`nl2br` won't work because this popup is HTML, not markdown.
The `<pre>` tag we're using in `format_diagnostic` would be a good way to preserve text formatting but it's not supported by minihtml? Does `<code>` work? Maybe you can find another tag/style combo supported by minihtml (https://www.sublimetext.com/docs/3/minihtml.html)?

predragnikolic:
I tried out `code` and that doesn't preserve white space.
There is a `white-space: pre;` property in css in browsers, but subime mini-html doesn't support that, nor the `pre` tag.

So I guess this can only be solved my manually replacing the new lines with breaks.

tomv564:
Sorry to take so long to look at this - the place where you are replacing newlines looks like it could accidentally affect other hover content. 
Could we replace the newlines on `diagnostic.message` in  `format_diagnostic` instead?

predragnikolic:
Done :)

PR #671, opened 2019-08-03
Advertise all 'goto' capabilities to servers

rwols:
Also, be explicit about not supporting dynamic registration.


rwols:
Will make a new PR for the explicit dynamic capabilities.

PR #673, opened 2019-08-04
Organize our textdocument and workspace capabilities

rwols:
Organize our textdocument and workspace capabilities.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -8,8 +8,8 @@ from .logging import debug
 import os
 from .protocol import completion_item_kinds, symbol_kinds
 try:
-    from typing import Callable, Dict, Any, Optional
-    assert Callable and Dict and Any and Optional
+    from typing import Callable, Dict, Any, Optional, List, Union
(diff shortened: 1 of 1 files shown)


tomv564:
The organisation and highlighting of missing capabilities is really great, but I'd like to know why need to send 'dynamicRegistration: false' in the first place - the absence should be "falsy" and back-wards compatible?

rwols:
> the absence should be "falsy" and back-wards compatible?

You're right. I'll remove the explicitness after #689 is merged.

rwols:
Now that #689 is merged I added colorProvider to the capabilities.

tomv564:
I don't think this PR on the whole improves the initialisation params. The result has become harder to read, and we now have a bunch of function calls and typing just to provide constant dicts. It probably made more sense when there was some repetition with the dynamic registration property?

We probably also shouldn't use the codebase for documenting missing capabilities as the protocol will always continue adding more. We can use the wiki and users can open issues to help bring them to our attention.

rwols:
Agreed.

PR #700, opened 2019-08-26
Preserve the order in the array of text edits if they have the same location

### LINKED ISSUE #690: Formatting adding trailing newline clears last line
* OS and language server
Linux + Gopls
Package Control
* Minimal reproduction steps
```go
package main

import (
	"fmt"
)

func main() {
	fmt.Println("Hello, world")
} // No newline!
```

Format

```go
package main

import (
	"fmt"
)

func main() {
	fmt.Println("Hello, world")
```
* Log
No diagnostic output. 

Initially reported to [gopls](https://github.com/golang/go/issues/33717), but they pointed out that the gopls commandline does the right thing.

Is this a LSP issue or Sublime itself?
Let me know if I can provide any other helpful information!

rwols:
Related to #690.

There's a bug in our sorting of the text edits. The original order in the array should be preserved if text edits have the same start location. Consequently, a test was wrong :) but is now fixed.

This does not end #690 because I'm pretty sure now that it is a bug in gopls. Or at least, gopls assumes there is an "infinite buffer" which ST doesn't have. One may argue gopls sends invalid positions outside of the text buffer. I have documented this behavior in the form of a skipped test called test_remove_line_and_then_insert_at_that_line_at_end.

### DIFF
--- a/plugin/core/edit.py
+++ b/plugin/core/edit.py
@@ -35,7 +35,7 @@ def sort_by_application_order(changes: 'Iterable[TextEdit]') -> 'List[TextEdit]'
 
     def get_start_position(pair: 'Tuple[int, TextEdit]'):
         index, change = pair
-        return change[0][0], change[0][1], index
+        return change[0][0], change[0][1], -index
 
     # The spec reads:
     # > However, it is possible that multiple edits have the same start position: multiple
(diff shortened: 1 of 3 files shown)


rwols:
Never mind, I cannot read.

> the order in the array defines the order in which the inserted strings appear in the resulting text.

PR #701, opened 2019-08-27
Fix inline color phantom makes line height larger a little bit

jfcherng:
The phantom created by LSP is make the height of the line larger a little bit.

- Similar issue that I reported to ColorHelp weeks ago: https://github.com/facelessuser/ColorHelper/pull/122


tomv564:
Thanks for fixing it here too!

Issue #629, opened 2019-06-27
Option to disable autocomplete feature


arsham:
* OS and language server
Linux/gopls
* How you installed LSP
git

I appreciate it if you would provide an option to disable the autocomplete functionality.

Thank you.

tomv564:
Almost a duplicate of #574. 

We could let the user blacklist some capabilities in an LSP setting and filter those out of our `initialize` request, saving us the effort of creating a new setting for every feature users want to disable.

tomv564:
There is a new `disabled_capabilities` setting, adding "completion" to the array in your LSP settings should stop LSP from contributing to Sublime's autocompletion UI.

arsham:
Amazing work, thank you.

Issue #574, opened 2019-04-26
Feature request: Please add an option for disabling popups on hover.


agilesteel:
I am constantly finding myself hiding my mouse cursor in the corners of my screen since even when I don't touch my mouse for a while some code always slides under the cursor and some random popup steals my focus.

It is already possible to bind a key to display a popup like so: `{ "keys": ["ctrl+space"], "command": "lsp_hover"}` however this does not disable the default behavior which triggers a popup as soon as the mouse cursor finds itself hovering over something it considers relevant.

tomv564:
Sounds annoying!
I tried to repro this on a recent dev build on OS-X but noticed that the mouse cursor disappears as soon as I start typing, and I don't get hover popups until I use the touchpad again.

I'm not sure if it's always been like this, but this behaviour sounds like it should prevent your problem. Is the mouse cursor active while typing in your sublime text? Do other editors (eg. IntelliJ?) have some smart way of avoiding this on your platform?

agilesteel:
Interesting, it might be related to the fact that I'm running it inside of a VM. My host is Windows 10 and my guest is Ubuntu 16.04. I'm not sure about other editors. I haven't used IntelliJ for a long time now. In fact for similar reasons. IDEs keep doing some stuff all the time that I haven't asked them to do, Sublime doesn't which is why I love it. It stays out of my way and I usually set it up in such a way that I don't even see auto complete popups (only triggered by a key binding).

BTW VS code doesn't have the hover problem since its LSP respects the `"editor.hover.enabled": false` setting. Maybe Sublime LSP should also respect the native `"show_definitions": false` Sublime setting? Just a suggestion...

tomv564:
We probably can introduce a new setting instead! Good to see an example of another editor disabling them.

tomv564:
There is a new `disabled_capabilities` setting, adding "hover" to the array in your LSP settings should disable the automatic hover popups.

skobes:
FYI it appears the setting is now "hoverProvider", e.g.

```
{
  "disabled_capabilities": { "hoverProvider": true }
}
```

Just in case anyone else finds this issue while looking for the way to disable popups.

Issue #703, opened 2019-08-27
Reduce impact of LSP color listener by default.


tomv564:
@predragnikolic: Noticed while closing views after find-replace in pyls:

```
Traceback (most recent call last):
  File "/Applications/Sublime Text.app/Contents/MacOS/sublime_plugin.py", line 506, in run_async_view_listener_callback
    vel.__class__.__dict__[name](vel)
  File "/Users/tomv/Library/Application Support/Sublime Text 3/Packages/LSP/plugin/color.py", line 47, in on_activated_async
    self.schedule_request()
  File "/Users/tomv/Library/Application Support/Sublime Text 3/Packages/LSP/plugin/color.py", line 53, in schedule_request
    current_point = self.view.sel()[0].begin()
  File "/Applications/Sublime Text.app/Contents/MacOS/sublime.py", line 649, in __getitem__
    raise IndexError()
IndexError
```

* We should probably fix the above issue (maybe it's rare but I ran into this edge case pretty quickly)
* We could avoid creating ViewEventListeners for a few kinds of irrelevant views: Read-only, transient, unsupported syntaxes ( Maybe `DocumentSyncListener` `is_applicable` is suitable?)
* We could avoid running `LspColorListener`'s `schedule_request` until a session with colorProvider for a given view has been found.

predragnikolic:
Hello @tomv564 .
Can you tell me what steps do I need to take to get that error?
I have been trying to reproduce it, but I didn't succeed.

tomv564:
Sorry, it should be a "replace in files".
Replace all 'debug' with 'dbg' in all of LSP. You should get a bunch of changed files opened.
Close these these files (ctrl-w or cmd-w) selecting "Don't Save" and you should see the error.

predragnikolic:
Thanks, I reproduced it. Now I can work on a fix.

predragnikolic:
- I fixed the IndexError.

- The schedule_request is not called until a session is found.
-  Added a `is_supported_syntax` check to `is_applicable` method in LspColorListener. 

What I didn't do from the above is: 
`"avoid creating ViewEventListeners for a few kinds of irrelevant views: Read-only, transient"`.

Why the need to avoid Read-only and transient views?

PR #704, opened 2019-08-28
Add disabled_capabilities settings for hover and completion

tomv564:
For #629 and #574

We should probably move the colorProvider toggle into this as well.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -82,7 +82,7 @@ Global plugin settings and settings defined at project level are merged together
 * `diagnostics_gutter_marker` `"dot"` *gutter marker for code diagnostics: "dot", "circle", "bookmark", "cross" or ""*
 * `show_code_actions_bulb` `false` *show a bulb in the gutter when code actions are available*
 * `show_color_box` `true` *Show a box next to the color if the language server provides color support*
-* `disabled_capabilities`, `[]` *Turn off client capabilities (features), like "hover" or "completion"
+* `disabled_capabilities`, `[]` *Turn off client capabilities (features): "hover", "completion", "documentHighlight", "colorProvider"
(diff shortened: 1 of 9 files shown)


tomv564:
Added colorProvider and documentHighlight as both have existing ways of being turned off in the settings.
We should remove the show_color_box setting before the next release, perhaps part of fixing #703

plugin/color.py
```diff
@@ -39,9 +39,11 @@ def __init__(self, view: sublime.View) -> None:
 
     @classmethod
     def is_applicable(cls, _settings):
-        if settings.show_color_box:
-            return True
-        return False
+        if 'colorProvider' in settings.disabled_capabilities:
+            return False
+        elif not settings.show_color_box:
```
rwols: Perhaps just remove this setting?

docs/features.md
```diff
@@ -82,6 +82,7 @@ Global plugin settings and settings defined at project level are merged together
 * `diagnostics_gutter_marker` `"dot"` *gutter marker for code diagnostics: "dot", "circle", "bookmark", "cross" or ""*
 * `show_code_actions_bulb` `false` *show a bulb in the gutter when code actions are available*
 * `show_color_box` `true` *Show a box next to the color if the language server provides color support*
+* `disabled_capabilities`, `[]` *Turn off client capabilities (features), like "hover" or "completion"
```
rwols: Please clarify that valid identifiers are coming from `TextDocumentClientCapabilities`.

tomv564: Good point, added all valid options here. Currently the subset is fairly small, so I don't want to mislead people into thinking all the capabilities can be toggled this way.

plugin/color.py
tomv564: Done!

PR #707, opened 2019-08-31
Improve color provder

### LINKED ISSUE #703: Reduce impact of LSP color listener by default.

predragnikolic:
closes #703

### DIFF
--- a/plugin/color.py
+++ b/plugin/color.py
@@ -24,7 +24,6 @@ class LspColorListener(sublime_plugin.ViewEventListener):
         self.color_phantom_set = None  # type: Optional[sublime.PhantomSet]
         self._stored_point = -1
         self.enabled = False
-        self.session = None
 
     @classmethod
     def is_applicable(cls, _settings):
@@ -34,12 +33,12 @@ class LspColorListener(sublime_plugin.ViewEventListener):
         return is_supported and not disabled
 
     def on_activated_async(self):
-        self.session = session_for_view(self.view)
-        if not self.session:
+        session = session_for_view(self.view)
+        if not session:
             self.initialize_session()
             return
 
-        self.enabled = self.session.has_capability('colorProvider')
+        self.enabled = session.has_capability('colorProvider')
         if self.enabled:
             self.send_color_request()
 
@@ -50,16 +49,15 @@ class LspColorListener(sublime_plugin.ViewEventListener):
     def initialize_session(self):
         config = config_for_scope(self.view)
         if config:
-            print('add listener', "initialized.{}".format(config.name))
-            global_events.subscribe("initialized.{}".format(config.name), self.on_session_initialized)
+            sublime.set_timeout_async(lambda: self.maybe_session_initialized(), 1000)
 
-    def on_session_initialized(self, session):
-        print('remove listener', 'initialized.{}'.format(session.config.name))
-        global_events.unsubscribe('initialized.{}'.format(session.config.name), self.on_session_initialized)
+    def maybe_session_initialized(self):
+        session = session_for_view(self.view)
+        if not session:
+            return
 
         self.enabled = session.has_capability('colorProvider')
         if self.enabled:
-            self.session = session
             self.send_color_request()
 
     def schedule_request(self):
@@ -80,12 +78,16 @@ class LspColorListener(sublime_plugin.ViewEventListener):
         if is_transient_view(self.view):
             return
 
+        session = session_for_view(self.view)
+        if not session:
+            return
+
         params = {
             "textDocument": {
                 "uri": filename_to_uri(self.view.file_name())
             }
         }
-        self.session.client.send_request(
+        session.client.send_request(
             Request.documentColor(params),
             self.handle_response
         )
--- a/plugin/color.py
+++ b/plugin/color.py
@@ -17,16 +17,12 @@ from .core.configurations import is_supported_syntax
 
 
 def send_color_request(view, on_response_recieved: 'Callable'):
-    session = session_for_view(view)
-    if not session or not session.has_capability('colorProvider'):
-        # the server doesn't support colors, just return
-        return
-
     params = {
         "textDocument": {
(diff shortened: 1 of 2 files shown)


predragnikolic:
MyPy is complaining, should I add a check
    if session is None:
        return

plugin/color.py
```diff
@@ -45,11 +43,26 @@ def is_applicable(cls, _settings):
         disabled = 'colorProvider' in settings.disabled_capabilities
         return is_supported and not disabled
 
+    def initialize(self):
+        session = session_for_view(self.view)
+        if session is None:
+            # not yet initialized, try again
+            sublime.set_timeout_async(self.initialize, 400)
```
tomv564: Will this not create an endless loop if no session is ever created?
Hard to guess how long it takes for a session to start up, perhaps some servers take longer than 400ms.
Perhaps we can hook up to another view event instead of this set_timeout for re-trying initialisation?

ayoub-benali:
Thanks for working on this. I uncounteted this issue multiple times

predragnikolic:
This commit 68508009b40dc1b6525e9e6d741c868f712e5806 is just an experiment. 
I left all the print statements in the code, 
so you can try this branch locally and in sublime console get a better feeling of what is happening in the background. 

But here is a how it works.
https://github.com/tomv564/LSP/blob/68508009b40dc1b6525e9e6d741c868f712e5806/plugin/color.py#L35-L39
When you open a python file for the first time.
Session is None,
so we call `initialize_session` 

https://github.com/tomv564/LSP/blob/68508009b40dc1b6525e9e6d741c868f712e5806/plugin/color.py#L49-L53

 `initialize_session` sets up a listener, 
where the `key` has this form `"initialized.{client_config_name}"`. 
In our case, we opened a python file, the event `key` name will be `"initialized.pyls"`. 

> NOTE: The event name could be `initialized`. 
The reason it is not done that way is this. 
What if some servers take long to initialize and you open a file css file. 
And it that moment the server that took long to initialize, initialized before the css server?? I don't what would happen, but I think nothing good.

In `windows.py` after we send a Notification to the server that the client is initialized.
We now emit a global_event, that a session has been created. 
Where the event `key` will be `"initialized.pyls"` and the payload will be the `session`.
https://github.com/tomv564/LSP/blob/68508009b40dc1b6525e9e6d741c868f712e5806/plugin/core/windows.py#L498-L499 

Than when that event is emited,
a listener is triggered, in our case it is `on_session_initialized`
https://github.com/tomv564/LSP/blob/68508009b40dc1b6525e9e6d741c868f712e5806/plugin/color.py#L56-L62

We immediately unsubscribe from the event, because we don't need it anymore. 
And do with the `session` payload what we want.

The flow I described above is only when you open a file for the first time, and no session is found.
https://github.com/tomv564/LSP/blob/68508009b40dc1b6525e9e6d741c868f712e5806/plugin/color.py#L35-L43 
The second time you open a file,
we will have a session for sure.

predragnikolic:
Just know that this approach may have problems I am not aware of.  
If you need a quick way to checkout PR. You can use https://github.com/predragnikolic/GithubPR.

predragnikolic:
Before `Center color box on the line`:

2019-09-02-220755_650x320_scrot

After:

2019-09-02-220841_652x296_scrot

tomv564:
Thanks for sharing the "listen for session created" mechanism. A similar thing exists in WindowDocumentHandler (`add_session`) and you've shown that we need it in a few more places. Perhaps it's better to work on this problem outside of this PR?

Some notes on the implementation of your idea.
* We want to move away from using `global_events` as events in other windows can easily leak into your current window. A WindowManager can probably provide the event instead.
* We want to avoid storing session unnecessarily as it's easy to get bugs when sessions are ended/removed/restarted. I think your implementation doesn't really require it either?

Anyways, there are some great fixes in this PR and I'd like to get them into master!
For a simpler solution of the "activated but not session not yet ready" case, can we do something like:
1. Check if a config is enabled for a view? (`is_supported_view`)-> Disable if not.
2. Is a session available? (`session_for_view`) -> Enable based on capabilities.
3. Is a session starting? (something like `_can_start_config` in WindowManager) -> Timeout 1 second, then step 2 again.

predragnikolic:
Tom, I tried to do as you suggested

>Check if a config is enabled for a view? (is_supported_view)-> Disable if not.
Is a session available? (session_for_view) -> Enable based on capabilities.
Is a session starting? (something like _can_start_config in WindowManager) -> Timeout 1 second, then step 2 again.

I am going on vacation today and I am not bringing a laptop with me :)
so I if you request some changes, could you apply them?
If not, it is ok. I will be back in two weeks. :)

tomv564:
Great work, made 1 tweak to try to initialise without timeout as well (see commit) and to stop initialising on subsequent activation.
Enjoy your time away!

PR #709, opened 2019-09-01
Expand variables on initializationOptions

jfcherng:
This PR adds the ability to use some variables in `initializationOptions`.

This PR comes from the following use case. In the settings of `intelephense` server, there are some path-related settings in the `initializationOptions`. The `licenceKey` can be a path of a file containing a license key. I would like to keep it cross-platform and portable so I want it to be in ST's directory (or at least, in user's home directory `~`).

```js
"intelephense": {
    // ...
    "initializationOptions": {
        "clearCache": false,
        "storagePath": "${temp_dir}/intelephense",
        "globalStoragePath": "${temp_dir}/intelephense",
        "licenceKey": "${packages}/User/intelephense-license.txt",
    },
},
```

### DIFF
--- a/plugin/core/clients.py
+++ b/plugin/core/clients.py
@@ -47,14 +47,15 @@ def get_expanding_variables(window: sublime.Window) -> dict:
     return variables
 
 
-def expand_variables_for_dict(window: sublime.Window, dict_: dict) -> dict:
-    for key, value in dict_.items():
+def expand_variables_for_dict(window: sublime.Window,
+                              d: Dict[str, Any]) -> Dict[str, Any]:
+    for key, value in d.items():
         if isinstance(value, dict):
-            dict_[key] = expand_variables_for_dict(window, value)
+            d[key] = expand_variables_for_dict(window, value)
         elif isinstance(value, str):
-            dict_[key] = sublime.expand_variables(value, get_expanding_variables(window))
+            d[key] = sublime.expand_variables(value, get_expanding_variables(window))
 
-    return dict_
+    return d
 
 
 def start_window_config(window: sublime.Window,
--- a/plugin/core/clients.py
+++ b/plugin/core/clients.py
@@ -1,5 +1,6 @@
 import sublime
 import os
+import tempfile
 from .sessions import create_session, Session
 
 # typing only
@@ -36,6 +37,26 @@ def get_window_env(window: sublime.Window, config: ClientConfig) -> 'Tuple[List[
     return expanded_args, env
 
 
+def get_expanding_variables(window: sublime.Window) -> dict:
+    variables = window.extract_variables()
+    variables.update({
+        "home": os.path.expanduser('~'),
+        "temp_dir": tempfile.gettempdir(),
+    })
+
+    return variables
+
+
+def expand_variables_for_dict(window: sublime.Window, dict_: dict) -> dict:
+    for key, value in dict_.items():
+        if isinstance(value, dict):
+            dict_[key] = expand_variables_for_dict(window, value)
+        elif isinstance(value, str):
+            dict_[key] = sublime.expand_variables(value, get_expanding_variables(window))
+
+    return dict_
+
+
 def start_window_config(window: sublime.Window,
                         project_path: str,
                         config: ClientConfig,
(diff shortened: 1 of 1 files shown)


plugin/core/clients.py
```diff
@@ -36,6 +37,26 @@ def get_window_env(window: sublime.Window, config: ClientConfig) -> 'Tuple[List[
...
+    variables = window.extract_variables()
+    variables.update({
+        "home": os.path.expanduser('~'),
+        "temp_dir": tempfile.gettempdir(),
+    })
+
+    return variables
+
+
+def expand_variables_for_dict(window: sublime.Window, dict_: dict) -> dict:
```
rwols: NIT: I don't like the underscore suffix for a variable :(

Also, we can pin down what kind of dict this is, right? It must be a `Dict[str, Union[str, dict]]`. That should also be the return type if I read it correctly.

jfcherng: > I don't like the underscore suffix for a variable :(

I dont either. But is using `dict` as a variable name a good practice? Or, what'a suggested one?

jfcherng: > Also, we can pin down what kind of dict this is, right? It must be a Dict[str, Union[str, dict]]. That should also be the return type if I read it correctly.

If the target is not a more general `dict`, i.e., `Dict[str, Any], should we change the function name?

jfcherng: >  It must be a Dict[str, Union[str, dict]]

What about `bool`, `int`, `float`?

```
"intelephense": {
    // ...
    "initializationOptions": {
        "clearCache": false, // <-------- bool
    },
},
```

rwols: You're right. It's probably best to leave it at `Dic[str, Any]`.

plugin/core/clients.py
```diff
@@ -36,6 +37,27 @@ def get_window_env(window: sublime.Window, config: ClientConfig) -> 'Tuple[List[
...
+    variables.update({
+        "home": os.path.expanduser('~'),
+        "temp_dir": tempfile.gettempdir(),
+    })
+
+    return variables
+
+
+def expand_variables_for_dict(window: sublime.Window,
+                              d: Dict[str, Any]) -> Dict[str, Any]:
```
rwols: I don't think the old python3.3 from ST3 will accept this. Shouldn't this be in quotes?

`d: 'Dict[str, Any]') -> 'Dict[str, Any]'`

jfcherng: Ah, yes!

jfcherng: Why don't we use a real `typing` module?
https://github.com/packagecontrol/typing

rwols: Interesting! I was not aware that ST dependency exists.

jfcherng:
> I welcome this change and I'm looking forward to doing this in more places, especially the "command" key.

@rwols Done via https://github.com/tomv564/LSP/pull/709/commits/184b58fc95b493971c2dca5ee315a6b586511221

plugin/core/clients.py
tomv564: About the dependency, do we really need `typing` to be importable for endusers?
We _could_ replace the try blocks with this (mypy sets MYPY to always True when checking):
```python
MYPY = False
if MYPY:
    from typing import Dict
```

jfcherng: I was just curious why `typing` is not actually imported and using "quoted" type hintings in codes. For whatever concerned, you are in charge anyway. :smile: I am just lazy so I directly make it as a dependency in my plugins.

tomv564:
If we do this, someone will suggest we infinitely recurse though the `settings` member as well?

Looking at the reason for the request, it looks like the specific language server is the major need, combined with your additional desire to have your config portable. 

Appreciate offering to do the work yourself, but I want to discuss the value of this complexity to end users vs. maintenance burden.
Implementing it in a LanguageHandler might be a better outcome for all of us?

predragnikolic:
hello @jfcherng, if you don't mind  I added your code to LSP-intelephese :) 

https://github.com/sublimelsp/LSP-intelephense/commit/2f8d525b82482ddce1bc8d45d4c57ed96af56060

jfcherng:
@predragnikolic sure. feel free to use it.

PR #717, opened 2019-09-13
Create a new line for edits after the document

### LINKED ISSUE #690: Formatting adding trailing newline clears last line

tomv564:
Fixes #690 where gopls is returning edits for a line that doesn't exist - this happens to work with VS Code because implementation details.

To do:
- [x] ~remove list constructor so we can just pass an iterator from sorted in `sort_by_application_order`~ (Did not work with `reversed`)
- [x] update comments in the test case

### DIFF
--- a/plugin/edit.py
+++ b/plugin/edit.py
@@ -51,12 +51,14 @@ class LspApplyDocumentEditCommand(sublime_plugin.TextCommand):
                 start, end, newText = change
                 region = sublime.Region(self.view.text_point(*start), self.view.text_point(*end))
 
-                # Handle when a language server (eg gopls) inserts at a row beyond the document
-                # some editors create the line automatically, sublime needs to have the newline prepended.
                 if start[0] > last_row and newText[0] != '\n':
-                    newText = '\n' + newText
-
-                self.apply_change(region, newText, edit)
+                    # Handle when a language server (eg gopls) inserts at a row beyond the document
+                    # some editors create the line automatically, sublime needs to have the newline prepended.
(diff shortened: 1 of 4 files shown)


plugin/edit.py
```diff
@@ -41,13 +41,21 @@ def open_and_apply_edits(self, path, file_changes):
...
-            for change in sort_by_application_order(changes):
+            last_row, last_col = self.view.rowcol(self.view.size())
+            for change in reversed(sort_by_application_order(changes)):
                 start, end, newText = change
                 region = sublime.Region(self.view.text_point(*start), self.view.text_point(*end))
+
+                # Handle when a language server (eg gopls) inserts at a row beyond the document
+                # some editors create the line automatically, sublime needs to have the newline prepended.
+                if start[0] > last_row and newText[0] != '\n':
+                    newText = '\n' + newText
```
rwols: Looks like an OK workaround.

tomv564: Thanks for the test you wrote and the research!

PR #723, opened 2019-09-19
Use sublime's auto_complete_selector setting for filtering completions

tomv564:
Sublime has a pretty good scope-based filter for determining if we need completions. 

Switching to this setting:
* Allows the user to change the completion scopes without editing the package source
* Disables completion in comments (no more completion on '.' in comments)
* Allows completion in string literals (at least, works in typescript for me)

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -18,8 +18,6 @@ from .core.documents import get_document_position, is_at_word
 from .core.sessions import Session
 from .core.edit import parse_text_edit
 
-NO_COMPLETION_SCOPES = 'comment, string'
-
 
 class CompletionState(object):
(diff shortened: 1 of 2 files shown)


rchl:
Just a note that completions in comments can be useful. For example when writing annotating code with jsdoc comments.

tomv564:
The most significant change is that LSP will start sending completion requests within regular `""` and `''` strings for most users. 

They might tolerate the built-in sublime completions within strings because those are fast. If the server is slow, completion popups come up late and might be more distracting.

I wonder if users would like the ability to say: Give me LSP completions in source, but only sublime word completion within strings?

PR #731, opened 2019-09-25
Format on save

tomv564:
For issue #8 

Adds support for sync request (blocking the editor for up to 1 second)

Default off, enable by setting `lsp_format_on_save` to true in preferences.

### DIFF
--- a/plugin/core/rpc.py
+++ b/plugin/core/rpc.py
@@ -15,7 +15,6 @@ from .logging import debug, exception_log
 from .protocol import Request, Notification, Response
 from .types import Settings
 
-
 TCP_CONNECT_TIMEOUT = 5
 
 # RequestDict = TypedDict('RequestDict', {'id': 'Union[str,int]', 'method': str, 'params': 'Optional[Any]'})
@@ -75,14 +74,19 @@ class Client(object):
         self._response_handlers = {}  # type: Dict[int, Tuple[Optional[Callable], Optional[Callable[[Any], None]]]]
         self._request_handlers = {}  # type: Dict[str, Callable]
         self._notification_handlers = {}  # type: Dict[str, Callable]
+        self._sync_request_results = {}  # type: Dict[int, Optional[Any]]
         self.exiting = False
         self._crash_handler = None  # type: Optional[Callable]
         self._transport_fail_handler = None  # type: Optional[Callable]
         self._error_display_handler = lambda msg: debug(msg)
         self.settings = settings
 
-    def send_request(self, request: Request, handler: 'Callable[[Optional[Any]], None]',
-                     error_handler: 'Optional[Callable[[Any], None]]' = None) -> None:
+    def send_request(
+            self,
+            request: Request,
+            handler: 'Callable[[Optional[Any]], None]',
+            error_handler: 'Optional[Callable[[Any], None]]' = None,
+    ) -> 'None':
         self.request_id += 1
         if self.transport is not None:
             debug(' --> ' + request.method)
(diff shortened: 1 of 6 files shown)


rwols:
I have something to say about this, but I need a little more time, so please don't merge this yet.

rchl:
It might be related to this PR or might not but I think this will not work for eslint server. I believe (after a short look at its code) that the way eslint server formats code on saving, is that it expects to get `textDocument/willSaveWaitUntil` notification and from there it handles the formatting.

Some references:
 - https://github.com/microsoft/vscode-languageserver-node/blob/master/server/src/main.ts#L282-L288
 - https://github.com/microsoft/vscode-eslint/blob/52a3bba443b92fdf33ada0758eeeb2b93aa387d8/server/src/eslintServer.ts#L679-L698

tomv564:
@rchl good to know, eslint doesn't have a formattingProvider so `willSaveWaitUntil` is the only way to make it work. I think the implementation would be very similar to this one, if you're keen on trying it out!

plugin/formatting.py
```diff
@@ -37,20 +37,22 @@ def is_applicable(cls, view_settings: dict) -> bool:
...
-                    "options": options_for_view(self.view)
-                }
-                request = Request.formatting(params)
-                response = client.execute_request(request)
-                if response:
-                    apply_response_to_view(response, self.view)
-                    global_events.publish("view.on_purge_changes", self.view)
+        file_path = self.view.file_name()
+        if file_path:
+            if self.view.settings().get("lsp_format_on_save"):
```
rchl: I don't know what's the style of this package but my preference would be to create as little nesting as possible. So either merge those two `if`'s or even return early instead of nesting.

Feel free to totally ignore me. :)

rchl:
@tomv564 I'm quite intimidated by this code base and didn't have a time to properly learn it. So we'll see if I get the courage. :)

plugin/core/rpc.py
```diff
@@ -92,6 +97,29 @@ def send_request(self, request: Request, handler: 'Callable[[Optional[Any]], Non
...
+            debug('unable to send', request.method)
+            return None
+
+        debug(' ==> ' + request.method)
+        self.request_id += 1
+        request_id = self.request_id
+        self.send_payload(request.to_payload(self.request_id))
+
+        current_time = start_time = time.time()
+        while current_time < start_time + timeout:
```
rwols: Condition variables are perfect for this use-case: we have some resource updated in another thread but want to access it from the current thread. I suggest to change the while-loop here with a condition variable. Consider `git cherry-pick`ing this commit I made that already does that: 65e979a

plugin/formatting.py
```diff
@@ -13,25 +17,52 @@
...
+                client = client_from_session(session_for_view(self.view, 'documentFormattingProvider'))
+                if client:
+                    params = {
+                        "textDocument": {
+                            "uri": filename_to_uri(file_path)
+                        },
+                        "options": options_for_view(self.view)
+                    }
+                    request = Request.formatting(params)
+                    response = client.execute_request(request)
```
rwols: When I make a bunch of changes and then quickly hit CTRL+S, the textDocument/didChange notification is sent too late. We should have an additional `global_events.publish("view.on_purge_changes", self.view)` just above this line to make sure that the server sees the most recent document. Consider `git cherry-pick`ing this commit that already does that: 65e979a

tomv564: Good one, I wonder if our `didSave` should also purge changes before being sent?

plugin/core/rpc.py
tomv564: Thanks, this is a nice improvement! Cherry picking gave me both your suggestions.

tomv564:
Good reminder from @rwols that this is the first time we introduce a regular sublime setting (at the view level) instead of an LSP setting. I didn't think much about it beyond the necessity of turning this feature on granularly (per project/syntax) - is it the best way to solve this?

tomv564:
@rchl I added the `willSaveWaitUntil` logic to this PR as it was so similar. If you checkout branch `workspace_configuration_request` then you will get these changes _and_ a minimal `workspace/configuration` implementation for the eslint server.

rchl:
@tomv564 It does actually work (after I've adjusted my LSP-eslint to expose setting from a correct place). Very nice, thank you for your work. :)

plugin/formatting.py
rchl: Don't know if relevant but with `pyls` server, I often see issues not disappearing on first save after fixing. This happens with `pylint` linter which is one of many that `pyls` runs and that one only runs on save.

plugin/formatting.py
tomv564: Good point, changed it!

PR #735, opened 2019-09-26
Add option to hide symbol action links in hover popup

### LINKED ISSUE #482: Option to hide links in hover

jwortmann:
This will add a new setting "show_symbol_action_links" with default value "true" to enable/disable the symbol action links (e.g. "References", "Rename") which are shown in hover pupups.

I think this option would be nice, if the user (like me) doesn't really use these links and would like to have a less distracting popup.

Closes #482


predragnikolic:
Maybe also add the check here? :)
```
  formatted.extend(items) 
  if settings.show_symbol_action_links:
            formatted.append("<a href='{}'>{}</a>".format('code-actions', 
                                                   'Code Actions')) 
  formatted.append("</div>")
```


https://github.com/tomv564/LSP/blob/c0b86c0bc5a50866daf75d1b2f86137078a6f8dd/plugin/hover.py#L138-L143

jwortmann:
The "Code Actions" link is indeed quite similar to the other links at the bottom of the popup and the setting name that I chose could imply to include that link too. Personally I would welcome to remove it as well, because with the language servers I use, the "Code Actions" link doesn't seem to work at all.

But on the other hand, "Code Actions" seems to be displayed only on warnings/errors, whereas "Definition", "References" and "Rename" (for the language servers I use) are added to all hover popups, even in cases where it absolutely doesn't make sense in that particular context. So I think that the "Code Action" link is much less annoying and probably more useful (well, I can only guess regarding the latter, because it just doesn't work for me...)

Perhaps a third opinion would be helpful for this.

tomv564:
Good reasoning about code actions - one improvement is to combine its logic with the code actions bulb, so we really only show it when there are fixes.

As for the other links, they're really only there for discoverability - I don't ever click them and I suspect few people do. So perhaps we can remove them permanently and come up with a better idea so people can quickly see what actions are available at a given point in a document, including shortcuts?

ayoub-benali:
I don't use the links as well. Do a deletion is fine by me

jwortmann:
I'm fine with removing the links permanently, because that functionality can still be accessed from the right-click context menu and it feels kind of redundant to have them in the hover popups as well. But I expect there are some users that use them, so I think it could be better to make them optional. If you prefer to not introduce another setting, feel free to close this PR.

For the code actions link, I saw [there should be another popup](https://github.com/tomv564/LSP/issues/10#issuecomment-323165232) with "No actions available" if there aren't any (?), however this popup isn't shown for me.

tomv564:
#741 will deal with the code actions link (you probably discovered a bug with the "No actions available" hover - it works when not triggered from the hover popup)

I'm still hoping to avoid adding a setting, but keep this PR open because using a setting with default off can also be a deprecation path.

jwortmann:
Thank you for merging!

But I noticed one small bug now in https://github.com/tomv564/LSP/blob/e2b715b417a63aee728ba20ffb0de6c687ff6558/plugin/hover.py#L197
This will append the links (if set to true) even if `contents` is empty, which will then bypass the following check to avoid empty popups. So I suggest to replace this line with
```python
if contents and settings.show_symbol_action_links:
```

Issue #699, opened 2019-08-26
Support workspace/configuration?


tomv564:
Related to #526 where the ESLint server makes the request and subsequently fails to start.

Spec here:
https://github.com/Microsoft/language-server-protocol/blob/gh-pages/specification.md#workspace_configuration

Do other language servers need it?
What is authoritative on these scope names used in the URI? VS Code? EditorConfig?
How does it apply to our package: try to read Sublime settings, or store them in EditorConfig or each language client's settings?

rwols:
As with all `any[]` types in the LSP spec that go directly against the spirit of LSP we can't really do anything generic.

My idea about this was to let a LanguageHandler handle this.

Here's a basic outline of how that would look:

```
windows.py

        client.on_request(
            "workspace/configuration",
            lambda params, request_id: self._handle_configuration(params, client, request_id))


    def _handle_configuration(self, params, client, id) -> None:
        # LanguageHandler.get_configuration would be a new interface method
        maybe = self._handlers.get_configuration(params)
        if maybe:
            client.send_response(maybe ...)
            return
        # fall back to just getting the view settings
        view = self.window.active_view()
        setting = view.settings().get(params[ ... ])
        client.send_response(setting ...)
```

tomv564:
Some interesting background in this discussion: https://github.com/microsoft/language-server-protocol/issues/567#issuecomment-420589320

tomv564:
@rwols, to respond to your suggestion:

Looking at the code, a LanguageHandler can replace the default implementation by calling `client.on_request("workspace/configuration", my_handler)` in the `on_initialized` hook. 

As for storing server configuration in the Sublime settings structure instead of the LSP client configs, I think it's interesting (as we discussed in the format on save PR). But I think that servers likely expect the same content as they got from `didChangeConfiguration`?

tomv564:
As an implementation is merged now, I will close the issue until we have more servers that challenge the current functionality.

PR #737, opened 2019-09-29
Support workspace/configuration

tomv564:
This is enough to get the eslint server to pull configuration from the client config's `settings` object.

The kotlin language server also used the request briefly (with section parameter set to "kotlin") but reverted to `workspace/didChangeConfiguration`.

Goal is to cover needs discussed in issue #699

Need to figure out how to handle:
-  ~section (eslint doesn't use it, but I guess we could treat them as keys under the client config settings object?)~ Lack of consistent use, we wait for more server to adopt.
-  ~scopeUri (at least send None for URIs outside the workspace?)~ wait for multiple workspace support 
-  ~do we need to check server support? (we could send `workspace/didChangeConfiguration` notifications with no content, unclear if needed)~ There is no server capability.

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -464,6 +464,14 @@ class WindowManager(object):
         # reconstruct/get the actual Client object back. Maybe we can (ab)use our homebrew event system for this?
         client.send_response(Response(request_id, {"applied": True}))
 
+    def _get_session_config(self, params: 'Dict[str, Any]', session: Session, client: Client, request_id: int) -> None:
+        items = []  # type: List[Any]
+        requested_items = params.get("items") or []
+        for requested_item in requested_items:
+            items.append(session.config.settings)
+
+        client.send_response(Response(request_id, items))
+
     def _handle_pre_initialize(self, session: 'Session') -> None:
         client = session.client
         client.set_crash_handler(lambda: self._handle_server_crash(session.config))
@@ -488,6 +496,10 @@ class WindowManager(object):
         client.on_request(
             "workspace/applyEdit",
             lambda params, request_id: self._apply_workspace_edit(params, client, request_id))
+
+        client.on_request(
+            "workspace/configuration",
+            lambda params, request_id: self._get_session_config(params, session, client, request_id))
 
         client.on_notification(
             "textDocument/publishDiagnostics",
--- a/plugin/core/rpc.py
+++ b/plugin/core/rpc.py
@@ -14,11 +14,8 @@ except ImportError:
 from .logging import debug, exception_log
 from .protocol import Request, Notification, Response
 from .types import Settings
-from threading import Condition
-from threading import Lock
 
 TCP_CONNECT_TIMEOUT = 5
-DEFAULT_SYNC_REQUEST_TIMEOUT = 1.0
 
 # RequestDict = TypedDict('RequestDict', {'id': 'Union[str,int]', 'method': str, 'params': 'Optional[Any]'})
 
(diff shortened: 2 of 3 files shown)


plugin/core/windows.py
```diff
@@ -464,6 +464,14 @@ def _apply_workspace_edit(self, params: 'Dict[str, Any]', client: Client, reques
         # reconstruct/get the actual Client object back. Maybe we can (ab)use our homebrew event system for this?
         client.send_response(Response(request_id, {"applied": True}))
 
+    def _get_session_config(self, params: 'Dict[str, Any]', session: Session, client: Client, request_id: int) -> None:
+        items = []  # type: List[Any]
+        requested_items = params.get("items") or []
+        for requested_item in requested_items:
+            items.append(session.config.settings)
+
+        client.send_response(Response(request_id, items))
```
rchl: Would it be a lot more work to do it "properly" from the start and expose a hook that LSP plugins (`LanguageHandler` implementations) could react to?

I don't know if it would be really worth it, as with static configuration it will probably be good enough for 99% of the cases, but just in case... :)

plugin/core/windows.py
```diff
@@ -464,6 +464,14 @@ def _apply_workspace_edit(self, params: 'Dict[str, Any]', client: Client, reques
         # reconstruct/get the actual Client object back. Maybe we can (ab)use our homebrew event system for this?
         client.send_response(Response(request_id, {"applied": True}))
 
+    def _get_session_config(self, params: 'Dict[str, Any]', session: Session, client: Client, request_id: int) -> None:
+        items = []  # type: List[Any]
+        requested_items = params.get("items") or []
+        for requested_item in requested_items:
+            items.append(session.config.settings)
```
rchl: If we will go with this hardcoded solution, maybe we should consider not using root `settings` for that but have a more specific property in `settings` for this case. Should be better in case there is a need later to provide some other settings also.

plugin/core/windows.py
tomv564: Right, it would be nice if language handlers can replace this built-in implementation. Would it not be valuable for LSP to provide a functional example for other plugins to improve on (with better defaults for example) ?

plugin/core/windows.py
tomv564: I found no explanation that this configuration request expects a different data structure than the one sent by didChangeConfiguration. With your example, users would suddenly have to move their config into a new property just because the server “upgraded” to the workspace/configuration request.

rchl: OK, I thought it's some arbitrary settings. Didn't know it's something more specific that is already sent with specific message. Then it makes sense to do it like that.

plugin/core/windows.py
rchl: I guess you are right. Providing default with a way to override would be the best.

For Eslint server, response can include properties like `workspaceFolder` and `workingDirectory`. To provide those, we'd need to be able to override response in `LanguageHandler` but to be fair, I've been running without those two and things are generally working without them. So I'm not sure how important are those.

tomv564:
The use of "section" in VS code's eslint: Server sends a blank section, and assumes the Code extension only returns eslint config (see: https://github.com/microsoft/vscode-eslint/blob/master/client/src/extension.ts#L490)

Hopefully this means `section` is a path _after_ the language server's config. 
The Kotlin server sent "kotlin" as section, meaning they expected they were querying a "global" set of settings (which exists in VS Code, but in LSP the settings start at the language client config).

Ignoring `section` for now is likely incorrect, but works for both implementations that I found.

tomv564:
ScopeUri is probably best implemented once we have multiple workspaces.

PR #741, opened 2019-10-01
Pre-request code actions

tomv564:
The Hover popup currently always shows a Code Actions link, even if no code actions are available. This PR aims to query the server, then show information about available code actions in the popup.

This shows an initial implementation, running with a typescript and eslint server.
[
Screenshot 2019-10-01 at 11 30 23
](url)

**Two functional issues need to be resolved:**
- [x] The list of code actions is long - change to a "Code Actions (3)" link?
- [x] Only the first codeActionProvider for the document is being queried - we should query all and show a "Code Actions" link for each.
  - (I'm hesitant to get into this, but of all places to combine LSP results, this use case probably makes the most sense)

**UI opinions wanted**

The hover UI is getting messy. Code actions feel like they "belong" to the diagnostics, but there can be many blocks of diagnostics (errors and warnings, per server) so it's unclear where to place the links.

**Technical fixes**
- [x] Potentially cooperating with the bulb logic (so fetched actions can be used at once)

### DIFF
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -9,7 +9,7 @@ from .core.registry import session_for_view, LspTextCommand
 from .core.protocol import Request, DiagnosticSeverity, Diagnostic
 from .core.documents import get_document_position
 from .core.popups import popup_css, popup_class
-from .code_actions import run_code_action_or_command, request_code_actions_with_diagnostics
+from .code_actions import actions_manager, run_code_action_or_command  # , request_code_actions_with_diagnostics
 from .core.settings import client_configs, settings
 
 try:
@@ -79,7 +79,6 @@ class LspHoverCommand(LspTextCommand):
     def run(self, edit: 'Any', point: 'Optional[int]' = None) -> None:
         hover_point = point or self.view.sel()[0].begin()
         self._hover = None  # type: Optional[Any]
-        self._requested_action_configs = []  # type: List[str]
         self._actions_by_config = {}  # type: Dict[str, List[dict]]
         self._diagnostics_by_config = {}  # type: Dict[str, List[Diagnostic]]
 
@@ -88,7 +87,7 @@ class LspHoverCommand(LspTextCommand):
 
         self._diagnostics_by_config = point_diagnostics_by_config(self.view, hover_point)
         if self._diagnostics_by_config:
-            self._requested_action_configs = self.request_code_actions(hover_point)
+            self.request_code_actions(hover_point)
             self.request_show_hover(hover_point)
 
     def request_symbol_hover(self, point: int) -> None:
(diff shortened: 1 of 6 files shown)


tomv564:
Last commit achieved the two feature goals: All code action features to query all servers, and the hover shows an actions link per server diagnostics section.

**Examples**

Both servers report a diagnostic, only typescript offers actions:

Screenshot 2019-10-03 at 13 04 55

Eslint with error, warning and actions block:

Screenshot 2019-10-03 at 13 05 08

**UI polish goals (please suggest how you'd improve this!)**
* Reduce the amount of color in diagnostics
* Get the Code Actions link to pop out a bit more (is it rendering darker than the diagnostics texts?)

jwortmann:
I just tested this quickly:
* The logic to not show a 'Code Actions' link if the server doesn't provide it seems to work fine
* Something isn't right with the popup HTML/CSS yet, because the `warnings` class is omitted for me now:
Before:
before
After:
after

My opinion for the UI:
I would prefer to not have the background color `--accent` for the 'Code Actions' link CSS, because it could be an arbitrary color from the color scheme, which might even be the same as `--yellowish` or `--redish` for warnings and errors. In your case it is blue, but even that doesn't seem optimal if the server provides a info/hint message, which has also blue background. I think the traditional underlined blue text link without background color (like in your opening comment) would be better here, so that users immediately see that it is a clickable link.

tomv564:
Thanks for the feedback! I added the missing class and adjusted the colors - I hear your argument for the underlined link, but I'd like to try a button-like style as well to differentiate LSP actions from documentation links.

Screenshot 2019-10-03 at 16 09 18

plugin/hover.py
```diff
@@ -126,28 +131,38 @@ def symbol_actions_content(self) -> str:
     def format_diagnostic(self, diagnostic: 'Diagnostic') -> str:
         diagnostic_message = escape(diagnostic.message, False).replace('\n', '<br>')
         if diagnostic.source:
-            return "<pre>[{}] {}</pre>".format(diagnostic.source, diagnostic_message)
+            return "<pre class=\"{}\">[{}] {}</pre>".format(class_for_severity[diagnostic.severity], diagnostic.source,
+                                                            diagnostic_message)
         else:
-            return "<pre>{}</pre>".format(diagnostic_message)
+            return "<pre class=\"{}\">{}</pre>".format(diagnostic_message)
```
jwortmann:
```suggestion
            return "<pre class=\"{}\">{}</pre>".format(class_for_severity[diagnostic.severity], diagnostic_message)
```

plugin/hover.py
```diff
@@ -173,7 +188,15 @@ def hover_content(self, point: int, response: 'Optional[Any]') -> str:
...
+    def request_show_hover(self, point: int) -> None:
+        sublime.set_timeout(lambda: self.show_hover(point), 50)
+
+    def show_hover(self, point: int) -> None:
+        contents = self.diagnostics_content() + self.hover_content()
+
+        _test_contents.clear()
+        _test_contents.append(contents)  # for testing only
+
         mdpopups.show_popup(
```
jwortmann: I think there should be
```
if contents:
    mdpopups.show_popup(
    ...
    ...)
```
to prevent empty popups.

rchl:
I've tried it out just for a minute and one thing that bothered me immediately was that the hit zone for the `Core Actions` link is only on the characters and not on the spaces. Very unline normal links on web pages. :)

tomv564:
Reverted the Code Actions link to the classic blue underlined after your feedback @jwortmann  and @rchl !

PR #747, opened 2019-10-04
Add terraform-lsp server

juliosueiras:
Add support for terraform-lsp

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -319,6 +319,25 @@ Install the [bash language server](https://github.com/mads-hartmann/bash-languag
 
 `npm i -g bash-language-server`
 
+### Terraform<a name="terraform">
+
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -319,6 +319,25 @@ Install the [bash language server](https://github.com/mads-hartmann/bash-languag
...
+```json
+"terraform":
+{
+  "command":
+  [
+    "terraform-lsp"
+  ],
+  "enabled": true,
+  "scopes": ["source.terraform"],
+  "complete_all_chars": true,
```
tomv564: The `complete_all_chars` setting doesn't apply here - it is for end-users to apply globally in their LSP settings, not per server config.

juliosueiras: no problem

Issue #381, opened 2018-07-22
Investigate showing diagnostics panel on save


tomv564:
Instead of the debounced delay logic in #292

predragnikolic:
Having an option to show the panel on save would be cool. But there is one other thing that needs improvement.

This plugin currently handles diagnostics on this way:
* Open a file, diagnostics are shown for that file.
* Close a file, diagnostics are removed for that file.

I think this is a better approach.
* Open a file, diagnostics are shown for that file.
* Close a file, diagnostics are shown for that file.

I thought about when should the diagnostics be removed, and it is only in this two cases:
* When a file is opened and it doesn't belong to the project. As soon as it is closed, the diagnostics for that file should be removed.
* When a file is opened in a project, but it belongs to a vendor(PHP), node_module(JavaScript) or some other folder that should be ignored. When the file is closed all diagnostics should be removed, we don't want to care for code that we didn't write. This would need a new setting, like `ignore_diagnostics_paths` which would take a list of folder names to ignore. like ['vendor', 'node_modules'].

The advantage to this is when you close a file and it had errors, you would still see those error. Before you wouldn't know because as soon as you close it, all diagnostics for the given file are removed. 

I already made like a prototype(I hardcoded the ignored paths). if you want you can try it. 
In `core/windows.py` go to line 449 and replace the contents of `_handle_view_closed` with this:
```python
 def _handle_view_closed(self, view: View, session):
        file_name = view.file_name()
        base_dir = get_project_path(self._window)

        if not base_dir or not file_name:
            return

        file_in_project = base_dir in file_name  # type: bool
        file_in_ignored_paths = self._is_in_ignored_paths(base_dir, file_name)

        if not file_in_project or file_in_ignored_paths:
            self._diagnostics.remove(view, session.config.name)
            self._sublime.set_timeout_async(lambda: self._check_window_closed(), 500)

    def _is_in_ignored_paths(self, base_dir, file_name) -> bool:
        """ Returns a bolean indicating if the file should be ignored """
        ignored_paths = ['stubs', 'tests']  # this will be populated by the settings
        for ignored_path in map(lambda path: os.path.join(base_dir, path), ignored_paths):
            if ignored_path in file_name:
                return True
        return False
```
So now when you open a file in the LSP project or any Project that has `stubs` or `tests` at the root, as soon as you close the file that are in these folders the diagnostics will go away. Same for files that you open but don't belong to the currently open project. Only errors that matter to you will stay visible. 

[...]

PR #748, opened 2019-10-07
WIP: Open diagnostics panel on save/publish

tomv564:
An implementation to investigate #381.
Rather than opening the panel when diagnostics arrive - show the panel when the user has saved.
As relevant diagnostics can have arrived or still be pending at save, a bit of tracking state is needed.
Note: It turns out on_post_save_async is invoked even if the view was not dirty, additional check had to be added for is_dirty.

Unsolved:
- [x] The panel should probably show when the view is opened
- [ ] Add setting that combines logically with the existing `auto_show_diagnostics_panel` (default True)
- [x] Think of how to handle diagnostics for files that are not opened.

### DIFF
--- a/plugin/diagnostics.py
+++ b/plugin/diagnostics.py
@@ -279,6 +279,10 @@ expecting_diagnostics = {}  # type: Dict[str, Optional[Callable[[], None]]]
 
 
 class DiagnosticsOnSaveListener(sublime_plugin.ViewEventListener):
+    def __init__(self, view: sublime.View) -> None:
+        self._received_first_diagnostics = False
+        super().__init__(view)
+
     @classmethod
     def is_applicable(cls, view_settings: dict) -> bool:
         # if not settings.show_diagnostics_in_view_status:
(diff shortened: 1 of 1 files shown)


tomv564:
At this point the implementation only shows the panel if the current file has diagnostics.
This is probably fine/expected for people used to single-file linters.

But servers that run compilation/type checking will probably report errors in related files that are not open yet. I believe the panel should also open for these cases: If I comment out a function in file B, I want the panel to pop up saying "file A cannot import this function".

rmccue:
@tomv564 Alas, I switched to VS Code, so no longer use Sublime/LSP. Sorry! :)

predragnikolic:
Initial impression is good :) great job 

But I will be using this tomorrow the whole day and see if I encounter something unexpected and give you my feedback

predragnikolic:
Everything seems ok :)
It shows the diagnostics panel on save.

One possible improvement, but not related to this card:

I am in a file`src/hello.js`. 
Whenever the diagnostics panel open,
we should scroll the diagnostics panel view port to the currently open file section, 
in this case `◌ src/hello.js`.  
And only scroll to the section, if the file exist in the panel. 

But that can be done in a different PR. :)

Issue #722, opened 2019-09-19
Remove show_view_status setting?


tomv564:
As an effort of shrinking the package as we grow in other directions, I'd like to propose removing the `show_view_status` setting, showing the list of running language servers in the status bar, which is on by default.

I understand that some people like to turn off "noise" in the UI, but anyone have this setting turned off ? 

**Discussed cleanups:**
- [x] `show_diagnostics_phantoms` REMOVED
- [x] `prefer_label_over_filter_text` REMOVED
- [x] `quick_panel_monospace_font` REMOVED

**Proposals:**
- `show_view_status`: Create a single `"ui": "minimal"` setting that hides all UI in status bar, phantoms, hovers that can be seen "on-request" instead (eg. no close/next/previous buttons on the phantoms, no `show_symbol_action_links`,  no server names in status bar.)
- `show_diagnostics_in_view_status`: I'd like to remove this as you can use the new diagnostic phantoms instead.
- `auto_show_diagnostics_panel_level` and `show_diagnostics_severity_level`: We need to be able to show `hint` and `info` somewhere without the same urgency as errors and warnings. Is making one setting out of them worth the user pain / migration logic needed? <strike>That said looks like `show_diagnostics_severity_level` is only used by the phantoms, not the panel or hover. I think VS Code had an interesting UI discussion about this.</strike> Fixed, regions/hovers on all diagnostics, panel and navigation respect `show_diagnostics_severity_level`. All that remains is removing the `panel_level` setting.

predragnikolic:
Me, I like minimal UI :)

But If you are interested,
Here is a list of settings that I don't find useful.
I just never caught myself using phantoms for diagnostics.


    // Show in-line diagnostics using phantoms for unchanged files.
    "show_diagnostics_phantoms": false,

Diagnostics messages are usually long, and can't fit in the status bar. 
So I turned this setting off, and use the diagnostics panel instead.


    // Show the diagnostics description of the code
    // under the cursor in status bar if available.
    "show_diagnostics_in_view_status": true,

Can these two settings be combined?

    // Open the diagnostics panel automatically
    // when diagnostics level is equal to or less than:
    // error: 1
    // warning: 2
    // info: 3
    // hint: 4
    "auto_show_diagnostics_panel_level": 3,


    // Show the diagnostics with level less than or equal to
    // the given value.
    // error: 1
    // warning: 2
    // info: 3
    // hint: 4
    "show_diagnostics_severity_level": 3,

[...]

PR #754, opened 2019-10-15
Diagnostics rework

tomv564:
The "Show diagnostics panel after saving" investigation (#748) showed the functionality could be "tacked" on, but would work better if components were set up in proper relations instead of querying global state.

In this PR, a DiagnosticsPresenter is introduced which gets access to DocumentHandler's events.

Clean ups:
- [x] Deprecate the `show_diagnostics_phantoms` feature (as discussed in #722)

Desired fixes:
- [x] Stop using global_events for pushing diagnostics updates to ui
- [x] Remove duplication of looping logic over diagnostics after an update: `update_diagnostics_in_status_bar `, `update_diagnostics_panel`, `window_has_relevant_diagnostics`, `update_view_diagnostics` (trying a strategy to update many things in a single walk through)
- [x] Remove duplication querying diagnostics at a point: `point_diagnostics_by_config` vs `get_line_diagnostics` (perhaps share a cache if cursor listener and bulb listener both want diagnostics?)

- [x] Try to remove the need for `get_window_diagnostics` / `get_view_diagnostics` - it should be easy to access the DiagnosticsStorage, and it should expose useful functions for accessing/filtering data.

### DIFF
--- a/plugin/diagnostics.py
+++ b/plugin/diagnostics.py
@@ -4,7 +4,6 @@ import sublime
 import sublime_plugin
 
 from .core.configurations import is_supported_syntax
-from .core.events import global_events
 from .core.logging import debug
 from .core.panels import ensure_panel
 from .core.protocol import Diagnostic, DiagnosticSeverity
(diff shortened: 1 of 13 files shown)


plugin/diagnostics.py
```diff
@@ -292,16 +223,175 @@ def saved(self) -> None:
...
+    """ Iterate over diagnostics structure"""
+
+    def __init__(self, subs: 'List[DiagnosticsUpdateWalk]') -> None:
+        self._subscribers = subs
+
+    def walk(self, diagnostics_by_file: 'Dict[str, Dict[str, List[Diagnostic]]]') -> None:
+        self.invoke_each(lambda w: w.begin())
+
+        if diagnostics_by_file is not None:
+            if diagnostics_by_file:
```
predragnikolic: Why the double if for `diagnostics_by_file`? :-)

tomv564: Good catch, should have left a TODO there. The old `update_diagnostics_panel` had both checks and did nothing at all if `diagnostics_by_file` was `None` - I think that case can't happen anymore, so I'll remove the check!

tomv564:
@AmjadHD  this PR is more about refactoring code instead of UI polish - if we can do something that looks good in all fonts then I'd like to see an example.

PR #766, opened 2019-11-10
Add an extra comma so as to not have an invalid.illegal scope

rwols:
Although the JSON (with comments) is valid, basically because ST3 parses it and doesn't complain, the sublime syntax definition for JSON incorrectly scopes the whitespace just before `// "settings"` as `invalid.illegal`. Adding a comma works around this issue.

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -162,7 +162,7 @@
       "command": ["pyls"],
       "scopes": ["source.python"],
       "syntaxes": ["Packages/Python/Python.sublime-syntax", "Packages/MagicPython/grammars/MagicPython.tmLanguage", "Packages/Djaneiro/Syntaxes/Python Django.tmLanguage"],
-      "languageId": "python"
+      "languageId": "python",
       // "settings": {
       //   "pyls": {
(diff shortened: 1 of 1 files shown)


tomv564:
I have an opposite problem - the json server complains about all the trailing commas in the file as invalid (and I personally don't care for them either)

The settings open as "Sublime Text Settings (JSON)" and I don't see any highlighting  - do we have different syntaxes applied?

rwols:
I guess you're using the syntax file from PackageDev. I use the default syntax Packages/Javascript/JSON.sublime-syntax. I've made an issue about it: https://github.com/sublimehq/Packages/issues/2193, it should be auto-fixed in the future.

tomv564:
You are correct, it's the one from PackageDev. 
Getting the syntax fixed is a great move. 

The LSP-json plugin sets that syntax up as a "jsonc" language which helps with the comments but not with the trailing commas.
But I think the utility of a json language server is limited for the sublime settings file, so the trailing comma errors are more something for @predragnikolic to consider.

rwols:
This'll be auto-fixed once a new release of ST drops.

PR #767, opened 2019-11-10
Put the context menu into a separate child menu

rwols:
The number of LSP capabilities grows larger and larger :) I think it makes sense to put the items of the context menu into a child menu.

Screenshot from 2019-11-10 10-41-15


tomv564:
Sounds great, unsure if any LSP plugins have contributed LSP menu items, but if so theirs should also be moved! (Maybe the Metals plugin?)

PR #770, opened 2019-11-11
Teach LSP about WorkspaceFolder

rwols:
Introduce the Workspace class in protocol.py. It has a name and a ~URI~ path.
Refactor most of plugin/core/workspace.py to work with the new Workspace
class.

In the WindowManager class, use the new Workspace class at the boundary
of the session starter callable.

So, the Session class doesn't yet understand what a workspace is, but
the WindowManager does.

This diff does absolutely nothing feature-wise, it just refactors code.

One of the building blocks of #697.

### DIFF
--- a/plugin/core/test_windows.py
+++ b/plugin/core/test_windows.py
@@ -99,6 +99,9 @@ class WindowManagerTests(unittest.TestCase):
         # session must be started (todo: verify session is ready)
         view = MockView(__file__)
 
+        # WindowManager will call window.active_view() at some point during wm.activate_view
+        window._files_in_groups = [[view]]
+
         wm.activate_view(view)
         self.assertIsNotNone(wm.get_session(TEST_CONFIG.name))
         self.assertEqual(len(docs._sessions), 1)
@@ -161,16 +164,18 @@ class WindowManagerTests(unittest.TestCase):
         # change project_path
         new_project_path = tempfile.gettempdir()
         test_window.set_folders([new_project_path])
+        test_window.set_project_data({"folders": [{"name": "foo", "path": new_project_path}]})
         # global_events.publish("view.on_close", MockView(__file__))
         another_view = MockView(None)
         another_view.settings().set("syntax", "Unsupported Syntax")
+        test_window._files_in_groups[0][0] = another_view
         wm.activate_view(another_view)
 
         self.assertEqual(len(wm._sessions), 0)
         self.assertEqual(len(docs._sessions), 0)
 
         # don't forget to check or we'll keep restarting sessions!
-        self.assertEqual(wm._project_path, new_project_path)
+        self.assertEqual(wm.get_project_path(), new_project_path)
 
     def test_offers_restart_on_crash(self):
         docs = MockDocuments()
--- a/plugin/core/test_windows.py
+++ b/plugin/core/test_windows.py
@@ -164,7 +164,6 @@ class WindowManagerTests(unittest.TestCase):
         # change project_path
         new_project_path = tempfile.gettempdir()
         test_window.set_folders([new_project_path])
-        test_window.set_project_data({"folders": [{"name": "foo", "path": new_project_path}]})
         # global_events.publish("view.on_close", MockView(__file__))
         another_view = MockView(None)
         another_view.settings().set("syntax", "Unsupported Syntax")
--- a/plugin/core/test_windows.py
+++ b/plugin/core/test_windows.py
@@ -1,5 +1,5 @@
 from . import test_sublime as test_sublime
-from .diagnostics import WindowDiagnostics
+from .diagnostics import DiagnosticsStorage
 from .events import global_events
 from .sessions import create_session
 from .sessions import Session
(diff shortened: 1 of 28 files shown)


plugin/core/workspace.py
```diff
@@ -1,42 +1,116 @@
...
-            project_path = os.path.dirname(filename)
-            debug("using", project_path, "as a fallback.")
-            return project_path
-        else:
-            debug("no fallback path possible because the current file isn't saved to disk.")
-            return None
-    else:
-        debug("no view is active in current window")
-        return None  # https://github.com/tomv564/LSP/issues/219
+    return os.path.commonprefix([path + '/' for path in paths]).rstrip('/')
```
tomv564: `os.path.commonprefix` can return invalid paths (it does a character-based match instead of stopping at the last common directory).

plugin/edit.py
```diff
@@ -2,9 +2,11 @@
 import sublime_plugin
 from .core.edit import sort_by_application_order
 try:
-    from typing import List, Dict, Optional, Any, Iterable, Tuple
+    from typing import List, Dict, Optional, Any, Iterable
     from .core.edit import TextEdit
-    assert List and Dict and Optional and Any and Iterable and Tuple and TextEdit
+    assert List and Dict and Optional and Any and Iterable
+    # plugin/edit.py:9: error: The type alias to Tuple is invalid in runtime context
```
tomv564: I cherry-picked the fixes (and removed the type: ignore) to master to keep them outside of the workspace effort. Thanks!

rwols:
> The code suggests a window has multiple workspaces,

That is correct.

> but Sublime only has 1 workspace (which can contain multiple folders).

Disagree. A sublime-workspace file is just a cache file. It has IMO nothing to do with an LSP workspace.

> Parsing project files introduces a lot of logic over just calling window.folders(). Is the rarely used name field really worth it?

For sure.  *Suppose that* we agree that an LSP-Workspace is a folder entry in a sublime-project. A user can supply a name to each folder entry of a sublime-project:

```javascript
{
    "folders":
    [
        {
            "name": "Docs",
            "path": "some/relative/path/to/documents"
        },
        {
            "name": "Misc",
            "path": "/etc/absolute/path"
        },
        {
            "path": "."
        }
    ],
    "settings":
    {
        "tab_size": 4,
        "translate_tabs_to_spaces": true,
        "rulers": [120]
    }
}
```
This is a sublime-project with three LSP-Workspaces, named "Docs", "Misc", and whatever the name is of the folder that the sublime-project lives in.

> It's more consistent with LSP state to store paths instead of URIs (and convert when we send to server).

That's fair. I'll change that.

> Some code is introduced about common prefix of workspace folders, but I think it's unused?

Yep, that's unused. It was used for plugin/references.py in my WIP PR. I'll remove it.

rwols:
I believe travis is broken :(

tomv564:
Good catch, I had accidentally disabled builds on pushing to master while trying to fix the double builds on pull requests. Now fixed, reverted mypy to stay on python 3.3.

tomv564:
Let's call the path(+name) things `WorkspaceFolder` to be consistent with the naming used in protocol at least? 

I propose we proceed without project file parsing for now, I understood how the field works in Sublime but I think we'll be fine sending the `basename` for now.

rwols:
> I propose we proceed without project file parsing for now, I understood how the field works in Sublime but I think we'll be fine sending the basename for now.

Alright :) changes coming soon

plugin/core/windows.py
```diff
@@ -440,24 +444,34 @@ def end_session(self, config_name: str) -> None:
...
-            self._project_path = get_project_path(self._window)
-            if self._project_path is None and self._projectless_root_path is None:
+    def _ensure_workspace(self) -> 'Optional[WorkspaceFolder]':
+        if self._workspace is None:
+            self._workspace = maybe_get_first_workspace_from_window(self._window)
+            if self._workspace is None and self._projectless_workspace is None:
                 # the projectless fallback will only be set once per window.
-                self._projectless_root_path = get_active_view_path(self._window)
-        return self._project_path or self._projectless_root_path
+                view = self._window.active_view()
```
tomv564: I preferred this logic tucked away in `get_active_view_path`, now this function's logic is getting quite nested.

plugin/core/test_windows.py
```diff
@@ -161,16 +164,18 @@ def test_ends_sessions_when_quick_switching(self):
         # change project_path
         new_project_path = tempfile.gettempdir()
         test_window.set_folders([new_project_path])
+        test_window.set_project_data({"folders": [{"name": "foo", "path": new_project_path}]})
```
tomv564: Should not be needed?

plugin/core/windows.py
```diff
@@ -324,8 +327,8 @@ def __init__(self, window: WindowLike, configs: ConfigRegistry, documents: Docum
         self._sublime = sublime
         self._handlers = handler_dispatcher
         self._restarting = False
-        self._project_path = get_project_path(self._window)
-        self._projectless_root_path = None  # type: Optional[str]
+        self._workspace = maybe_get_first_workspace_from_window(self._window)  # type: ignore
```
tomv564: I think this `# type: ignore` is not needed?

tomv564:
So close!
```
plugin/core/workspace.py:3:1: F401 '.types.ViewLike' imported but unused
```

Will approve / merge as soon as fixed

PR #772, opened 2019-11-12
Extend completion heuristics for metals' 'implement all members'

### LINKED ISSUE #771: Completion inputs label instead of textEdit newText
scalameta/metals#1031 was recently merged which adds an "implement all members" completion option. However, in Sublime it seems to not show up in the same order in the completions as the other editors. It seems to be triggered by e for some reason. Apart from that, if you do decide to use that completion, the completion seems to instead of implementing the `newText` it implements the `label`.

I'm on MacOS using the [Metals Language Server](https://github.com/scalameta/metals) with this SNAPSHOT `0.7.6+224-b3ea857f-SNAPSHOT`

Here is a gif illustrating what I'm talking about

mygif

And here is the snippet of lsp json that shows the completion item

```
[Trace - 08:54:53 AM] Received request 'completionItem/resolve - (30)'
Params: {
  "kind": 12,
  "sortText": "00002",
  "filterText": "e",
  "insertTextFormat": 2,
  "textEdit": {
    "range": {
      "start": {
        "line": 9,
        "character": 3
      },
      "end": {
        "line": 9,
        "character": 4
      }
    },
    "newText": "def foo: Int \u003d ${0:???}\n   def boo: Int \u003d ${0:???}"
  },
  "data": {
    "target": "file:/Users/ckipp/Documents/scala-workspace/test-project/?id\u003droot",
    "symbol": "local6"
  }
}


[Trace - 08:54:53 AM] Sending response 'completionItem/resolve - (30)'. Processing request took 1ms
Result: {
  "kind": 12,
  "sortText": "00002",
  "filterText": "e",
  "insertTextFormat": 2,
  "textEdit": {
    "range": {
      "start": {
        "line": 9,
        "character": 3
      },
      "end": {
        "line": 9,
        "character": 4
      }
    },
    "newText": "def foo: Int \u003d ${0:???}\n   def boo: Int \u003d ${0:???}"
  },
  "data": {
    "target": "file:/Users/ckipp/Documents/scala-workspace/test-project/?id\u003droot",
    "symbol": "local6"
  }
}
```

If I can provide any more details, just let me know!

rwols:
@ckipp01 I haven't actually tested this with a snapshot build of MetaLS. Could you try this branch?

close #771

### DIFF
--- a/plugin/core/completion.py
+++ b/plugin/core/completion.py
@@ -54,12 +54,7 @@ def format_completion(item: dict, word_col: int, settings: 'Settings') -> 'Tuple
         elif trigger[0] == ' ' or trigger[0] == '•':
             trigger = trigger[1:]  # remove clangd insertion indicator
         else:
-            debug("replacement prefix does not match trigger!")
-            if trigger == "Implement all members":
-                # MetaLS workaround
-                replacement = item["textEdit"]["newText"]
-            else:
-                replacement = item.get("insertText") or trigger
+            debug("WARNING: Replacement prefix does not match trigger '{}'".format(trigger))
 
     if len(replacement) > 0 and replacement[0] == '$':  # sublime needs leading '$' escaped.
         replacement = '\\$' + replacement[1:]
--- a/plugin/core/completion.py
+++ b/plugin/core/completion.py
@@ -55,7 +55,11 @@ def format_completion(item: dict, word_col: int, settings: 'Settings') -> 'Tuple
             trigger = trigger[1:]  # remove clangd insertion indicator
         else:
             debug("replacement prefix does not match trigger!")
-            replacement = item.get("insertText") or trigger
+            if trigger == "Implement all members":
+                # MetaLS workaround
+                replacement = item["textEdit"]["newText"]
+            else:
+                replacement = item.get("insertText") or trigger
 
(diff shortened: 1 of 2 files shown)


ayoub-benali:
I know that we are limited by Sublime API https://github.com/SublimeTextIssues/Core/issues/2754
but this solution will break the moment Metals rename the trigger :/

rwols:
> but this solution will break the moment Metals rename the trigger :/

unfortunately, yes :(

ckipp01:
@rwols thanks for looking at this so quickly!

I tested this out locally, and yes, this does the trick. Probably not ideal that we have to have that coded in there, but I'm not sure of what else to do for it, so I'll leave it up to you if you want to add this in or not.

plugin/core/completion.py
```diff
@@ -55,7 +55,11 @@ def format_completion(item: dict, word_col: int, settings: 'Settings') -> 'Tuple
             trigger = trigger[1:]  # remove clangd insertion indicator
         else:
             debug("replacement prefix does not match trigger!")
-            replacement = item.get("insertText") or trigger
+            if trigger == "Implement all members":
```
tomv564: Let's remove the server-specific logic, remove the assignments to ` replacement`, and update the warning to:
```python
debug("WARNING: Replacement prefix does not match trigger '{}'".format(trigger))
```

PR #773, opened 2019-11-13
Parse and show related information in diagnostics hover content

### LINKED ISSUE #492: Support for relatedInformation diagnostic property
Some language servers support a `relatedInformation` field that supplements diagnostic messages with additional context. Here's the [spec](https://microsoft.github.io/language-server-protocol/specification#diagnostic) and the [LSP proposal](https://github.com/Microsoft/language-server-protocol/issues/260) from a year ago.

For example, in Hack (Facebook's PHP replacement), the language server delivers diagnostic messages when an argument type does not match the type of the formal parameter. Here's a [sample payload](https://gist.github.com/jeffomatic/ee031e11bbe42a7228f64e07d1165c4f).

There are a couple of things the ST LSP plugin could do in order to make this data truly useful:

- Within the diagnostic pane and the error hover popup, the LSP plugin surfaces the top-level diagnostic message, but what you really want to see is an expansion containing the two underlying messages, one at the callsite (the nested message reports the deduced type of the argument) and one at the declaration (the nested message reports the declared type of the parameter).
- The plugin successfully adds highlighting effects to the function callsite, but it ought to highlight the function definitions as well.

Probably, the first is easier to accomplish than the latter, but both would be amazing.

tomv564:
Closes #492 

It's pretty ugly, though RLS's duplicate content in the diagnostics doesn't help:

Screenshot 2019-11-13 at 17 59 41

Discussion of how VS Code implemented the UI: https://github.com/microsoft/vscode/issues/10271

### DIFF
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -1,3 +1,4 @@
+from .url import uri_to_filename
 try:
     from typing import Any, List, Dict, Tuple, Callable, Optional, Union, Mapping
     assert Any and List and Dict and Tuple and Callable and Optional and Union and Mapping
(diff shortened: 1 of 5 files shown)


rwols:
I'm going to try this out on clangd

rwols:
Schermafbeelding 2019-11-13 om 22 27 52

All of the "Candidate constructor ..." lines are DiagnosticRelatedInformation.

In both rust and clangd's case, these related diagnostics would stand out better if they had a different background color. Like an "informational" or "note" or "warning" color. Because in both cases they are informational.

Like in the terminal:

Schermafbeelding 2019-11-13 om 22 35 57

tomv564:
Agree, thankfully there are lots of shades of grey to use:

Screenshot 2019-11-14 at 22 45 20

From your clangd example the absolute file paths would be better left alone instead of made relative - fixed this as well.

predragnikolic:
50 shades of popup :-D

PR #775, opened 2019-11-13
Remove "prefer_label_over_filter_text" setting

tomv564:
As discussed in #722 and seen causing confusion with metals in #771.


ayoub-benali:
I will update the doc in Metals once there is a release.

Issue #646, opened 2019-07-09
Signature Help height too short (need to scroll)


FHTheron:
I am not sure if this is SublimeText or LSP. The popup cuts off after about two lines.

* Ubuntu, ccls
* Package Control or git (both do the same)
* Create a function with many long parameters.

![Screenshot from 2019-07-08 15-11-41](https://user-images.githubusercontent.com/16208294/60887754-6d3aca00-a255-11e9-83e6-3aff595c57eb.png)

AmjadHD:
possible duplicate of #486

tomv564:
Closing duplicate, please follow the linked issue  for updates :)

Issue #604, opened 2019-06-06
Hover popups scroll vertically instead of fitting to content


AmjadHD:
![image](https://user-images.githubusercontent.com/33235928/59054655-702c4e80-8894-11e9-9aab-3f2f886157c1.png)

* OS: Windows 10
* How you installed LSP (Package Control or from git?): git
* Minimal reproduction steps
1. open a css file with vscode-css
2. hover over e.g `sans-serif`

tomv564:
I've seen this too! Would be great if someone could try to address it with our `popup_css` ... assuming Sublime Text has not introduced new layout behaviour in newer versions.

AmjadHD:
sorry, css is not my cop of tea :)

AmjadHD:
possible duplicate of #486

tomv564:
Yeah, it's a duplicate, but nice to see some solutions in linked issues from #486. Closing this now!

PR #776, opened 2019-11-16
Break long spans in popups

### LINKED ISSUE #486: Why there are scroll bars on the houver pop up with so much screen available for it?
When reporting the issue: 
1. https://github.com/palantir/python-language-server/issues/464 Go to definition has a annoying behavior sometimes

I noticed this:
fsg

Are the scroll bars being put there by LSP or Sublime Text? If by LSP, would be nice a setting to extend their limits/maximum sizes. They do not seem big enough.

tomv564:
Closes #486

Specific fixes for the related issues #604 and #646 

Long hover before:
Screenshot 2019-11-16 at 18 19 48
after:
Screenshot 2019-11-16 at 18 22 02

Long signature before:
Screenshot 2019-11-16 at 18 20 54

after:
Screenshot 2019-11-16 at 18 22 17

Worth noting:
* Line breaks added at 80 characters, I'm aware there's logic in SublimeLinter that sets a viewport-relative limit but perhaps this is enough?
* Splitting does not work in signature labels that don't use commas.  

To do:
- [x] update tests
- [x] clean code
- [x] add screenshots.

### DIFF
--- a/plugin/core/signature_help.py
+++ b/plugin/core/signature_help.py
@@ -143,9 +143,33 @@ def create_signature_help(response: 'Optional[Dict]') -> 'Optional[SignatureHelp
 
 def render_signature_label(renderer: ScopeRenderer, sig_info: SignatureInformation,
                            active_parameter_index: int = -1) -> str:
-
     if sig_info.parameters:
         label = sig_info.label
+        params_to_newline_before = []
+        max_length = 80
+        if len(label) > max_length and '\n' not in label:
+            param_index = 0
+            start_offset = 0
+            line_offset = 0
+            while True:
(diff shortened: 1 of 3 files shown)


kaste:
We also only hardwrap at 80 columns. 🤫

Serega124:
For some reason span break does not work with [typescript-language-server](https://github.com/theia-ide/typescript-language-server) even when signature label contains commas:

изображение

LSP v0.9.1

tomv564:
That case is not covered yet, I left server-provided code blocks in hovers as text wrapping code might look funny. But looking at your example I think I’ll add a fix for it now.

PR #777, opened 2019-11-16
Docs: Add TypeScript disable auto-import feature

roblav96:
This `tsserver` default configuration frustrated me, thought I'd document a quick way to disable this feature until additional LSP option's are made available. 🙂

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -33,7 +33,7 @@ Have you added multiple folders to your Sublime workspace? LSP may not handle yo
 
 ### 4. [TypeScript] Disable auto-import out of scope autocompletions
 
-![typescript-auto-import](https://user-images.githubusercontent.com/1457327/68996892-42e5c280-086e-11ea-98e5-fffb003a1261.gif "TypeScript Auto Import")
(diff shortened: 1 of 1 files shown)


tomv564:
Yes, I remember turning this on in lsp-tsserver thinking it would ruin everything. Surprised no servers have exposed the setting, have your reported your request to their respective repos?

roblav96:
To be honest I'm still wrapping my head around all the layers that go into _Language Server Protocol_ as a whole, but I think the package `typescript-language-server` first needs to support a configuration option for [includeExternalModuleExports@lsp-server.ts#L432](https://github.com/theia-ide/typescript-language-server/blob/3114c557b58eb6b286e3c63f68ffe56c26800224/server/src/lsp-server.ts#L432), maybe read it from the `initializationOptions`?

tomv564:
@roblav96  exactly, they could also read it from the "workspace configuration". Either way, you know enough about the code - all you'd have to do is convince them to make the change :)

I think your documentation is only of value to serious hackers, and I'd prefer to link to the config option once it exists in `typescript-language-server`!

tomv564:
I hope you've reported the issue, I will close the PR now but will refer to it if more LSP users ask about unexpected extra typescript completion items.

PR #778, opened 2019-11-16
Read line from open view buffer, else fall-back to linecache

predragnikolic:
The bug:

When triggering find references,
if the view buffer has changed,
because we didn't yet saved the file,
we might end up reading a line which content had changed.

> because linecache read content of the file that was saved to disc,
and it can't read the content of the currently edited buffer(opened view).  

Ending up with a blank string reference `7:1  [no reference string]` 

bug

This PR fixes it:

output


plugin/references.py
```diff
@@ -191,3 +191,19 @@ def _group_references_by_file(self, references: 'List[ReferenceDict]'
         linecache.clearcache()
```
predragnikolic: Is it ok to leave the `linecache.clearcache()` function call here? :)

plugin/references.py
```diff
@@ -191,3 +191,19 @@ def _group_references_by_file(self, references: 'List[ReferenceDict]'
...
 
         return grouped_references
+
+
+def get_line(file_name: str, row: int) -> str:
+    '''
+    Get the line from the buffer if the view is open, else fallback to linecache.
+    '''
+    view = sublime.active_window().find_open_file(file_name)
```
rwols: I don't really like the `sublime.active_window()` call here. We already know what the view is in `_group_references_by_file`, it is `self.view`, and so we also know what the window is, namely `self.view.window()`. Perhaps `get_line` can get an extra parameter for the `window`?

plugin/references.py
rwols: I don't think it will cause any problems

rwols:
Looks like the build server broke :)

```
Downloading archive: https://storage.googleapis.com/travis-ci-language-archives/python/binaries/ubuntu/14.04/x86_64/python-3.3.tar.bz2

$ curl -sSf --retry 5 -o python-3.3.tar.bz2 ${archive_url}

curl: (7) Failed to connect to storage.googleapis.com port 443: Connection timed out

Unable to download 3.3 archive. The archive may not exist. Please consider a different version.
```

tomv564:
I think it was a temporary issue as my recent builds are breaking for other reasons :)

predragnikolic:
Hmm,
the build is failing because of 
`"get_line" has incompatible type "Optional[Window]"; expected "Window"`

`self.view.window()` returns an "Optional[Window].

This starts me wondering, can there be a view, if there is no window?

I looked at `sublime.pyi` file and saw this.
```
class View:
...
 def window(self) -> Optional[Window]:
```

And I thought it was a mistake,
Than I opened the real sublime.py file and saw this:

```
class View:
    ...
    def window(self):
        window_id = sublime_api.view_window(self.view_id)
        if window_id == 0:
            return None
        else:
            return Window(window_id)
```

And apparently it wasn't a mistake.
thew `window` method may return `None` if the window_id == 0;

but when is window_id == 0? :)
is that ever the case, maybe it is, and if so when? I know this is more a question for the sublime devs. Maybe you guys have an answer for that.

tomv564:
I think when the window is closed you can still be holding a view or window object in the python world that doesn’t really exist in the editor anymore. 

It’s an edge case but lsp events/responses can arrive after a window is closed.

PR #779, opened 2019-11-16
Fix bulb not showing up in selections even if code action exist

predragnikolic:
Send Code Action request only if the Region didn't changed for 800ms.

Previously we only checked if the .begin() of the region didn't changed for 800ms.

But in that case we might end up in situation when making a selection,
where we had code actions but didn't show a bulb, because the begin in selections didn't change.

https://github.com/tomv564/LSP/blob/f9c68aa98266e23f61b1254c3f41ee0af8c00f64/plugin/code_actions.py#L136

The bug:

bug

After this PR gets merged:
good

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -63,8 +63,9 @@ class CodeActionsManager(object):
         else:
             self._requests.clear()
             if diagnostics_by_config is None:
-                diagnostics_by_config = point_diagnostics_by_config(view, point)
-            self._requests[current_location] = request_code_actions(view, point, actions_handler)
+                diagnostics_by_config = point_diagnostics_by_config(view, view.sel()[0])
+                print('here are diagnostics', diagnostics_by_config)
+            self._requests[current_location] = request_code_actions(view, diagnostics_by_config, point, actions_handler)
 
     def get_location_key(self, view: sublime.View, point: int) -> str:
         return "{}#{}:{}".format(view.file_name(), view.change_count(), point)
@@ -73,13 +74,7 @@ class CodeActionsManager(object):
 actions_manager = CodeActionsManager()
 
 
-def request_code_actions(view: sublime.View, point: int,
-                         actions_handler: 'Callable[[CodeActionsByConfigName], None]') -> 'CodeActionsAtLocation':
-    diagnostics_by_config = point_diagnostics_by_config(view, point)
-    return request_code_actions_with_diagnostics(view, diagnostics_by_config, point, actions_handler)
-
-
-def request_code_actions_with_diagnostics(view: sublime.View, diagnostics_by_config: 'Dict[str, List[Diagnostic]]',
+def request_code_actions(view: sublime.View, diagnostics_by_config: 'Dict[str, List[Diagnostic]]',
                                           point: int, actions_handler: 'Callable[[CodeActionsByConfigName], None]'
                                           ) -> 'CodeActionsAtLocation':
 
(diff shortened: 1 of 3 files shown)


tomv564:
The bug to me is the bulb state changing at all when the selection begin() didn't change? I still see that the selection change doesn't effect the `position` sent to request code actions.

predragnikolic:
> I still see that the selection change doesn't effect the position sent to request code actions.

It does affect the code actions get from the server.

https://github.com/tomv564/LSP/blob/1fb0957075294e193e5644dc663ebf714bf2f355/plugin/code_actions.py#L139

This point(`current_region.begin()`) is later used to get the sessions in view for that point.

```
def request_code_actions_with_diagnostics(...):
    for session in sessions_for_view(view, point):
```

Based on that point we extract `point_diagnostics`
if we have a `point_diagnostics`
we will send point_diagnostics[0].range
else we will send the whole selected region.
```
else region_to_range(view, view.sel()[0])
```

Here is the place where the Code action request is send

```
def request_code_actions_with_diagnostics(view: sublime.View, diagnostics_by_config: 'Dict[str, List[Diagnostic]]',
                                          point: int, actions_handler: 'Callable[[CodeActionsByConfigName], None]'
                                          ) -> 'CodeActionsAtLocation':

    actions_at_location = CodeActionsAtLocation(actions_handler)

    for session in sessions_for_view(view, point):

        if session.has_capability('codeActionProvider'):
            if session.config.name in diagnostics_by_config:
                point_diagnostics = diagnostics_by_config[session.config.name]
                file_name = view.file_name()
                relevant_range = point_diagnostics[0].range if point_diagnostics else region_to_range(
                    view,
                    view.sel()[0])
                if file_name:
                    params = {
                        "textDocument": {
                            "uri": filename_to_uri(file_name)
                        },
                        "range": relevant_range.to_lsp(),
                        "context": {
                            "diagnostics": list(diagnostic.to_lsp() for diagnostic in point_diagnostics)
                        }
                    }
                    if session.client:
                        session.client.send_request(
                            Request.codeAction(params),
                            actions_at_location.collect(session.config.name))
    return actions_at_location
```

I don't know if this is good behaviour or not :)

tomv564:
Ah yes you are right, the user could “select” more diagnostics! I had a hard time imagining if any users would intentionally do this. 

What if we just send the range of the whole line? And only hide the bulb when the row has changed?

predragnikolic:
we should send the whole region, and we should send all diagnostics between that region :)
Here is how vs code does it.

we have a diagnostics at the variable `d`

Screenshot 2019-11-22 at 22 04 50

we have a diagnostics at the variable `d` and `c` so both diagnostics are send,

Screenshot 2019-11-22 at 22 05 11


we have a diagnostics at the variable `d`, `c` and `b`.

Screenshot 2019-11-22 at 22 05 25

we have a diagnostics at the variable `d`, `c`, `b`, `a` and `x`.

Screenshot 2019-11-22 at 22 05 35

I tried implementing this, but I hadn't had success. 
But still I will push those commits so you could see what I tried.

tomv564:
But this is great research (and a really useful example!) So what is preventing us from matching VS Code in this case?

predragnikolic:
> So what is preventing us from matching VS Code in this case?

Although I am pretty sure I sent the selected Sublime region, and all the diagnostics in between that region. The server seems like it only sends a response of the code actions underneath the cursor... maybe that is the way that the theia typescript-language-server implemented code actions. Im not sure. 
I will need a little more time to figure this out 🙂

tomv564:
I merged this, but really hope for some follow up work to fix the real issues:
* We currently filter diagnostics by point (instead of any diagnostic that _overlaps_ the selection)
* We send the first diagnostic's range as range, instead of the selection.

Would be awesome if we can find some of VS Code's reasoning about their implementation.

PR #782, opened 2019-11-19
Adjust link to discord chat

rchl:
Use the invite link. Fewer steps for people who are not invited yet but more steps for those who are (still have to click "accept").


tomv564:
Thanks, good you mentioned about the invite link, it's super annoying if you just want to go into the channel. 

Linked to the getting help section instead where the actions are explained. A smarter invite link (or a pair of buttons) at the top would be awesome.

FichteFoll:
The invite link should put you in the #lsp channel directly. No need for the extra reference.

PR #783, opened 2019-11-19
Show previous/next diagnostic with phantom

### LINKED ISSUE #746: How to jump to next/previous diagnostic?
I saw #350 but that didn't really answer my question.

I see the `F4` and `shift+F4` bindings are disabled by default:
https://github.com/tomv564/LSP/blob/81631161d920206f64291d62ff70fdfd2ffe27ba/Keymaps/Default%20(OSX).sublime-keymap#L19-L20

But if I enable them, they work, but only until I trigger search-in-files (which I do often). Then the results from search-in-files take over the shortcuts and I can no longer use them just within given file with diagnostic shown

If diagnostics panel is open and focused, I can use F4 and Shift+F4 to jump through diagnostics but I'd rather not have to open the panel as it's less convenient.

So I think proper solution would be to add extra View commands that I could trigger manually from my key bindings.

tomv564:
Press  <kbd>F8</kbd> to open/next, <kbd>shift+F8</kbd> for previous, <kbd>esc</kbd> to close. (see suggested keybindings)

Screenshot 2019-11-22 at 10 34 08

Screenshot 2019-11-26 at 15 19 08

Essentially like VS Code's "Next Problem" feature. 

Suggested keybindings
```json
{ "keys": ["f8"], "command": "lsp_next_diagnostic"},
{ "keys": ["shift+f8"], "command": "lsp_previous_diagnostic"},
{ "keys": ["escape"], "command": "lsp_hide_diagnostic", "context":
	[
		{ "key": "setting.lsp_diagnostic_phantom"}
	]
}
```

Closes #746:
* Allows navigation for those who don't like to use the panel for it (SublimeLinter style)
* Allows a faster and more readable way to see a diagnostic + related info than hovering.

Implementation details:
* The DiagnosticsPresenter passes user actions (forward/back) and server diagnostic updates into the DiagnosticsCursor
* The cursor selects a different walking logic based on an already-selected diagnostic or a cursor position in a file.
* The Escape key is "stolen" if the view has a `lsp_diagnostic_phantom` setting applied. There is some risk with stealing a so often-used key.
* Phantoms are not hidden if a diagnostic disappears.

To do:
- [x] Implement UI + wire up Esc for dismissing phantoms
- [x] Implement UI for related info
- <strike> Update UI for code actions (if keeping at all)</strike>
- [x] Review behaviour when diagnostics removed/moved/added.
- [x] Only include warnings and errors (See https://github.com/Microsoft/vscode/issues/45436)
- [x] UI polish (add prev/next, related info spacing and color, top arrow, rounded corners?)
- [x] Organize diagnostics code and add unit tests

### DIFF
--- a/Commands/Default.sublime-commands
+++ b/Commands/Default.sublime-commands
@@ -62,6 +62,16 @@
         "args": {}
     },
     {
+        "caption": "LSP: Next Diagnostic",
+        "command": "lsp_next_diagnostic",
+        "args": {}
+    },
+    {
+        "caption": "LSP: Previous Diagnostic",
+        "command": "lsp_previous_diagnostic",
+        "args": {}
+    },
+    {
         "caption": "LSP: Workspace Symbol",
         "command": "lsp_workspace_symbols"
     },
--- a/phantoms.css
+++ b/phantoms.css
@@ -12,9 +12,16 @@ div.warning-arrow {
 }
 div.container {
     margin: 0;
+    border-radius: 0 0.2rem 0.2rem 0.2rem;
 }
 div.content {
     padding: 0.4rem 0.4rem 0.4rem 0.7rem;
+}
+div.content p {
+    margin: 0.2rem;
+}
+div.content p.additional {
+    margin-top: 0.7rem;
 }
 div.toolbar {
     padding: 0.2rem 0.7rem 0.2rem 0.7rem;
--- a/phantoms.css
+++ b/phantoms.css
@@ -1,3 +1,15 @@
+div.error-arrow {
+    border-top: 0.4rem solid transparent;
+    border-left: 0.5rem solid color(var(--redish) blend(var(--background) 30%));
+    width: 0;
+    height: 0;
+}
+div.warning-arrow {
+    border-top: 0.4rem solid transparent;
+    border-left: 0.5rem solid color(var(--yellowish) blend(var(--background) 30%));
+    width: 0;
+    height: 0;
+}
 div.container {
     margin: 0;
 }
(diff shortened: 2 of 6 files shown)


rwols:
I haven't really looked at all the code yet, but I noticed these two things:

1.  phantomdiags
  The whiteish/bluish box around the related diags is not present. Is this intentional? It feels to me as if this part should share the same css code as the hover popup.

2. The command lsp_next_diagnostic does not cycle back to the beginning of the phantoms list, but lsp_previous_diagnostic _does_ cycle forward to the end of the phantoms list if invoking it on the first diag. I think they should both have the cycle behavior.

tomv564:
Good catch, even had a TODO for testing the missed wrap-forward from diagnostic logic

The color was doable in this manner: 

Screenshot 2019-11-27 at 00 10 57

But I feel the extra markup/css is a bit unfortunate.
The amount of styling on hovers gets a little distracting, like this example:

Screenshot 2019-11-27 at 00 11 11

For me the additional info is part of the error and could just be red.

rwols:
> For me the additional info is part of the error and could just be red.

I'm okay with having it red, but IMO there should be _some_ sort of separation between the main diag and the related diags. Indentation for the related diags perhaps? Or a horizontal line?

tomv564:
I like your suggestion of the horizontal line a lot (I tried indent too but this is more space-efficient)

Screenshot 2019-11-27 at 13 42 54

rwols:
Looks good!

tomv564:
I will merge this now and document it when we get to a release soon!

PR #793, opened 2019-11-28
Deprecate global events and "setup supported server" detection

tomv564:
Global events is a holdover from when LSP maintained all state globally, it allowed decoupling code that otherwise result in unneeded/circular dependencies. 

Now that feature code (event listeners, commands etc.) have a relation with an LSP WindowManager, we can stop firing events that all WindowManagers have to listen to (only to check if `view.window() == self.window` to ignore views from other windows)

The "setup supported server" logic also used global events and has seemed broken for a long time. I thought this was a good time to remove it in favour of LanguageHandlers / documentation.


tomv564:
`test_simple_label` test is flaky, will take a look at that later.

PR #794, opened 2019-11-29
Add elixir-ls support

dvcrn:
Add configuration and instructions to make [elixir-ls](https://github.com/elixir-lsp/elixir-ls) work

Sadly still needs a bit of manual configuration since this probably won't ever be in the path


dvcrn:
Yeah, Elixir dependencies/packages are project context specific so a runnable global install isn't a common concept. Packaging a binary is one option but bundling a random binary doesn't sound like a good idea 😅

PR #796, opened 2019-11-29
Make a trivial handler for 'LocationLink'

### LINKED ISSUE #764: 'lsp_symbol_definition' via keybinding  using `ccls' as client.
While I was trying to bind the 'lsp_symbol_definition'-command to my custom keys, I had so suffer from a `... type str does not support Buffer API´.
On revisiting the relevant file ("plugin/goto.py") I found the TODO comment which mentions the future implementation of DocumentLink-support.

By dumping the response-payload of the 'lsp_symbol_definition'-command I found out that the received payload has different keys as beeing parsed by the 'handle_response()' function.
See here:
```
[{targetUri: file:///usr/include/bash/builtins/bashgetopt.h, targetRange: {end: {line: 38, character: 30}, start: {line:
38, character: 0}}, targetSelectionRange: {end: {line: 38, character: 26}, start: {line: 38, character: 11}}}]
```

### Solution

So I added the following snippet to provide the right parsing method in case of this setup/scenario:

```python
            if 'targetUri' in location:
                file_path = uri_to_filename(location['targetUri'])
                start = Point.from_lsp(location['targetRange']['start'])
            else:
                file_path = uri_to_filename(location["uri"])
                start = Point.from_lsp(location['range']['start'])
```

By adding this the desired function of my keybinding worked.

I just wanted to share this in the hope it might be useful.
Best regards !

rwols:
For our fellow ccls users :) Verified with ccls.

close #764

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -110,10 +110,10 @@ def get_initialize_params(project_path: str, config: ClientConfig) -> dict:
                 },
                 "formatting": {},
                 "rangeFormatting": {},
-                "declaration": {},
-                "definition": {},
-                "typeDefinition": {},
-                "implementation": {},
+                "declaration": {"linkSupport": True},
+                "definition": {"linkSupport": True},
+                "typeDefinition": {"linkSupport": True},
+                "implementation": {"linkSupport": True},
                 "codeAction": {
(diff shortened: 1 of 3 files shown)


tomv564:
Haha, the close button strikes again!

PR #800, opened 2019-11-29
Apply severity filter to diagnostics panel and navigation, not underline/hover

### LINKED ISSUE #795: Diagnostic navigation ignores severity levels
I have `show_diagnostics_severity_level` and `auto_show_diagnostics_panel_level` set to 1
and when navigating though diagnostics (https://github.com/tomv564/LSP/pull/783) I reach `warn` diagnostic while I should have only `error`.

Maybe these two settings should also be merged ?

I tested with metals.

tomv564:
Closes #795

Next/previous diagnostic was hardwired to Errors and Warnings, but as we have a setting controlling what the user thinks is relevant (`show_diagnostics_severity_level`), we should read that instead.

Also drops the severity filter for in-view diagnostic regions. This is because as a developer I want quick access to errors and warnings, but still want to see server suggestions somewhere.

- [x] Should the default for `show_diagnostics_severity_level` change from 3 (`info`) to 2 (`warning`) ?

- <strike>For those offended by `info` and `hint` squigglies in their code, should we expose `diagnostic_severity_scopes` as a setting? (We could drop some of the fallback scopes and give the user the chance to not set any scopes at all for some diagnostics)</strike> Can do this later.

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -14,7 +14,7 @@
   // warning: 2
   // info: 3
   // hint: 4
-  "auto_show_diagnostics_panel_level": 3,
+  "auto_show_diagnostics_panel_level": 2,
 
   // Show errors and warnings count in the status bar
   "show_diagnostics_count_in_view_status": false,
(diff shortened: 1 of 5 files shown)


ayoub-benali:
issue fixed thanks @tomv564

PR #802, opened 2019-11-30
Add a UnitTesting test for the format-on-save feature

rwols:
This doesn't actually cover the transport/threads handling as we're dealing with a MockClient. It only verifies the "business logic" :)

### DIFF
--- a/plugin/core/test_mocks.py
+++ b/plugin/core/test_mocks.py
@@ -286,14 +286,7 @@ class MockClient():
             on_success(response)
 
     def execute_request(self, request: Request) -> 'Any':
-        response = None
-
-        def save_response(x):
-            nonlocal response
-            response = x
-
-        self.send_request(request, save_response)
-        return response
+        return self.responses.get(request.method)
 
     def send_notification(self, notification: Notification) -> None:
         self._notifications.append(notification)
--- a/plugin/core/test_mocks.py
+++ b/plugin/core/test_mocks.py
@@ -34,7 +34,8 @@ basic_responses = {
             'definitionProvider': True,
             'typeDefinitionProvider': True,
             'declarationProvider': True,
-            'implementationProvider': True
+            'implementationProvider': True,
+            'documentFormattingProvider': True
         }
     }
 }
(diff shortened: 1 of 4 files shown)


tests/setup.py
```diff
@@ -53,11 +53,17 @@ def close_test_view(view):
...
+        self.client = None
+
     def setUp(self) -> None:
-        self.view = sublime.active_window().open_file(test_file_path)
+        self.view = sublime.active_window().open_file(self.test_file_path)
         self.view.settings().set("auto_complete_selector", "text.plain")
         self.wm = windows.lookup(self.view.window())
-        self.client = MockClient(async_response=sublime_delayer(100))
+        if self.client is None:
+            self.client = MockClient(async_response=sublime_delayer(100))
```
tomv564: Can you explain why test_file_path must be configurable and why the change in client assignment?

tests/test_pre_save.py
```diff
@@ -0,0 +1,42 @@
...
+        super().tearDown()
+
+    def test_format_on_save(self):
+        """
+        Open the (empty) file on_pre_save.txt, put the character 'a' in it, and then save it.
+        The test language server should remove the 'a' character.
+        """
+        yield 100
+        setup = self.view.settings().set
+        setup("ensure_newline_at_eof_on_save", False)
```
tomv564: Prefer a specific name like `set_view_setting` (as `setup` might suggest something more involved)

tests/test_pre_save.py
```diff
@@ -0,0 +1,42 @@
...
+        setup("trim_trailing_white_space_on_save", False)
+        setup("lsp_format_on_save", True)
+        self.client.responses['textDocument/formatting'] = [{
+            "newText": "",
+            "range": {
+                "start": {"line": 0, "character": 0},
+                "end":   {"line": 0, "character": 1}
+            }
+        }]
+        run = self.view.run_command
```
tomv564: Not sure if we need to alias the occurrence of two not terribly-long calls?

plugin/core/test_mocks.py
```diff
@@ -284,6 +285,16 @@ def send_request(self, request: Request, on_success: 'Callable', on_error: 'Call
         else:
             on_success(response)
 
+    def execute_request(self, request: Request) -> 'Any':
+        response = None
```
tomv564: This could be as simple as 

```python
return self.responses.get(request.method)
```

tests/setup.py
rwols: The test saves the file. If the test fails, then other tests would use the content in that file.

As for the customizable MockClient, my initial thought was to make a test for the sync timeouts, but then I realized that we were dealing with mock transports anyway. I reverted that part.

Issue #249, opened 2018-01-21
TypeError: can't concat bytes to str


evandrocoan:
I am not sure how to reproduce, but I have enabled debug and this popped some time after:
```
:
2018-01-20 23:03:54,977 parso.python.diff diff parser start:
2018-01-20 23:03:54,978 parso.python.diff diff parser calculated:
2018-01-20 23:03:54,979 parso.python.diff diff: line_lengths old: 66, new: 66:
2018-01-20 23:03:54,979 parso.python.diff diff equal old[1:32] new[1:32]:
2018-01-20 23:03:54,980 parso.python.diff diff actually copy 1 to 7:
2018-01-20 23:03:55,013 parso.python.diff parse_part from 8 to 33 (to 33 in part parser):
2018-01-20 23:03:55,013 parso.python.diff diff replace old[33:33] new[33:33]:
2018-01-20 23:03:55,014 parso.python.diff diff equal old[34:66] new[34:66]:
2018-01-20 23:03:55,016 parso.python.diff parse_part from 34 to 34 (to 34 in part parser):
2018-01-20 23:03:55,017 parso.python.diff diff actually copy 35 to 65:
2018-01-20 23:03:55,018 parso.python.diff diff parser end:
2018-01-20 23:03:55,219 parso.cache pickle loaded: d:\user\dropbox\softwareversioning\pycodestyle\pycodestyle.py:
2018-01-20 23:03:55,289 pyls.config.config   finish pyls_hover --> [hook]:
Exception in thread Thread-112:
Traceback (most recent call last):
  File "./python3.3/threading.py", line 901, in _bootstrap_inner
  File "./python3.3/threading.py", line 858, in run
  File "F:\SublimeText\Data\Packages\LSP\plugin\core\rpc.py", line 164, in read_stderr
    server_log(decoded.strip())
  File "F:\SublimeText\Data\Packages\LSP\plugin\core\logging.py", line 18, in server_log
    printf(*args, prefix=binary)
  File "F:\SublimeText\Data\Packages\LSP\plugin\core\logging.py", line 23, in printf
    print(prefix + ":", *args)
TypeError: can't concat bytes to str
```

Perhaps use `print(prefix + ":", str(*args))` so we can know what it was trying to print?

tomv564:
Are you able to make this change locally in hopes of trapping the error?

evandrocoan:
Sure, I shall put a try...catch for the exception then print like 10 `\n` new lines to make sure I see it.

[...]

PR #803, opened 2019-12-01
Replace undecodeable UTF-8 sequences by '?'

rwols:
There's nothing in the LSP spec about the encoding of the stderr stream (or in the case of the TCP transport the stdout stream). We're assuming UTF-8, which is a good choice I think, but if the encoding fails we can try UTF-16 as a last resort.

Related: #249

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -33,6 +33,10 @@ disallow_untyped_defs = False
 check_untyped_defs = True
 disallow_untyped_defs = False
 
+[mypy-plugin.core.test_process]
+check_untyped_defs = True
+disallow_untyped_defs = False
+
 [mypy-plugin.core.test_protocol]
 check_untyped_defs = True
 disallow_untyped_defs = False
--- a/plugin/core/process.py
+++ b/plugin/core/process.py
@@ -1,8 +1,7 @@
 from .logging import debug, exception_log, server_log
-import locale
+import subprocess
 import os
 import shutil
-import subprocess
 import threading
 
 try:
@@ -64,7 +63,6 @@ def log_stream(process: 'subprocess.Popen', stream: 'IO[Any]') -> None:
     Reads any errors from the LSP process.
     """
     running = True
-    preferred_encoding = locale.getpreferredencoding()
     while running:
         running = process.poll() is None
 
@@ -72,14 +70,18 @@ def log_stream(process: 'subprocess.Popen', stream: 'IO[Any]') -> None:
             content = stream.readline()
             if not content:
                 break
+            decoded = None  # type: Optional[str]
             try:
                 # There's nothing in the specification about what the encoding should be of stderr (or, in the case of a
-                # TCP transport, the encoding of stdout). Let's assume the encoding of the current locale.
-                # On Linux and macOS, this is UTF-8 (even when using a Dutch locale on macOS). On Windows, this might be
-                # cp5212.
-                message = content.decode(preferred_encoding).strip()
+                # TCP transport, the encoding of stdout). Let's assume UTF-8.
+                decoded = content.decode("UTF-8")
             except UnicodeDecodeError:
-                message = "Unable to decode bytes! (tried decoding with {})".format(preferred_encoding)
+                # Guess it wasn't UTF-8. Let's try UTF-16 as a last resort.
+                try:
+                    decoded = content.decode("UTF-16")
+                except UnicodeDecodeError:
+                    decoded = None
+            message = decoded.strip() if decoded is not None else "Unable to decode bytes! (tried UTF-8 and UTF-16)"
             server_log('server', message)
         except IOError as err:
             exception_log("Failure reading stream", err)
--- a/plugin/core/process.py
+++ b/plugin/core/process.py
@@ -70,7 +70,7 @@ def log_stream(process: 'subprocess.Popen', stream: 'IO[Any]') -> None:
             content = stream.readline()
             if not content:
                 break
-            server_log('server', content.decode('UTF-8', 'replace'))
+            server_log('server', content.decode('UTF-8', 'replace').strip())
         except IOError as err:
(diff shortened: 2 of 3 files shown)


rwols:
closing and reopening due to test flakiness

tomv564:
We could also look at using `locale.getpreferredencoding` ? UTF-16 is perhaps only commonly seen on windows?

rwols:
I really that idea, let me try to implement that.

rwols:
test flakiness

rwols:
@AmjadHD just to be sure because we're not decoding with UTF-16 anymore (but with cp1252 as you reported on Discord), does this actually for you now?

AmjadHD:
The problem was fixed, but some characters are no shown correctly:
`server: Informationÿ: impossible de trouver des fichiers pour le(s) modŠle(s) sp‚cifi‚(s).`
it should look like:
`server: Information: impossible de trouver des fichiers pour le(s) modèle(s) spécifié(s).`

tomv564:
Looks like the preferred encoding is misleading, and UTF-16 was the likely best fallback (in Windows). Can we change the client to read stderr as UTF-16 as soon as a UTF-8 decode fails?

AmjadHD:
tried:
```diff
diff --git a/plugin/core/process.py b/plugin/core/process.py
index da0122c..0c2060e 100644
--- a/plugin/core/process.py
+++ b/plugin/core/process.py
@@ -71,7 +71,7 @@ def log_stream(process: 'subprocess.Popen', stream: 'IO[Any]') -> None:
             if not content:
                 break
             try:
-                decoded = content.decode("UTF-8")
+                decoded = content.decode("UTF-16")
             except UnicodeDecodeError:
                 # todo: do we still need this ?
                 decoded = content  # type: ignore
```
and I get this:
```湉潦浲瑡潩ｮ›浩潰獳扩敬搠⁥牴畯敶⁲敤⁳楦档敩獲瀠畯⁲敬猨 潭詤敬猨 灳掂晩艩猨⸩਍:```

**Comment by rwols** [comment]
@AmjadHD To rule out the possibility of SublimeText setting the wrong locale in Windows, can you open a Powershell command prompt and run the following?

```
[System.Text.Encoding]::Default
```

https://superuser.com/questions/670197/finding-out-the-default-character-encoding-in-windows

**Comment by AmjadHD** [comment]
```
PS C:\Users\Lehdhili> [System.Text.Encoding]::Default

Preamble          :
BodyName          : utf-8
EncodingName      : Unicode (UTF-8)
HeaderName        : utf-8
WebName           : utf-8
WindowsCodePage   : 1200
IsBrowserDisplay  : True
IsBrowserSave     : True
IsMailNewsDisplay : True
IsMailNewsSave    : True
IsSingleByte      : False
EncoderFallback   : System.Text.EncoderReplacementFallback
DecoderFallback   : System.Text.DecoderReplacementFallback
IsReadOnly        : True
CodePage          : 65001
```

**Comment by rwols** [comment]
This confuses me even more :+1:

**Comment by rwols** [comment]
So, the language server can choose whatever encoding it wants. It *should* use UTF-8 because payloads are in UTF-8, and otherwise *should* use the default system locale. It should not matter which language server we're looking at, but to get more context which language server are you using @AmjadHD?

I am also interested in the following:

1. What is your system locale? Run the following in Powershell: `Get-UICulture` https://serverfault.com/questions/173630/how-to-get-system-locale-in-windows-7-cmd
2. If you run Python in a REPL from the command-line (so not from within SublimeText), what does `import locale; locale.getpreferredencoding()` print?

**Comment by tomv564** [comment]
From the issue:
```
LSP: starting ['typescript-language-server.cmd', '--stdio']
```

I also read somewhere that windows console's default encoding is UTF-16, so the use of a `.cmd` wrapper might influence the chosen encoding

**Comment by AmjadHD** [comment]
```powershell
PS C:\Users\Lehdhili> Get-UICulture

LCID             Name             DisplayName
----             ----             -----------
1036             fr-FR            Français (France)
```
```
Python 3.8.0 (tags/v3.8.0:fa919fd, Oct 14 2019, 19:37:50) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import locale; locale.getpreferredencoding()
'cp1252'
```

rwols:
Perhaps we should stick to UTF-8 and fallback to locale.getpreferredlocale() if that fails.

If a French person using cp1252 is working for a Chinese customer, working on filenames with Chinese characters in it, then pretty much all logging through stderr would fail if using locale.getpreferredlocale().
On the other hand, UTF-8 can handle both French and Chinese.

I'm guessing we don't want another setting in LSP.sublime-settings for this.

So I propose:
1. Decode as UTF-8.
2. If that fails, decode as locale.getpreferredencoding().
3. If that fails, print a decoding error message.

The strange characters from @AmjadHD 's example are unfortunate but somewhat readable nonetheless, in contrast to using UTF-16.

I've also looked at the logging code of typescript-language-server

https://github.com/theia-ide/typescript-language-server/blob/f5fccedf1369b7e8db9f6b11d3a8427bcadf42b2/server/src/logger.ts#L60-L82

but I don't see anything standing out as strange.

AmjadHD:
I like 2, raw bytes is more readable than the output of `decode('utf-8', 'ignore')`

rwols:
But we have to convert to a string at *some* point though. We want to prefix the message with the server name. Theoretically speaking we could convert the server prefix to bytes then, as it's probably an ASCII name.

There's a third option called `'replace'`:

```
>>> print(b'Information\xff: impossible de trouver des fichiers pour le(s) mod\x8ale(s) sp\x82cifi\x82(s).\r\n'.decode('utf-8', 'replace'))
Information�: impossible de trouver des fichiers pour le(s) mod�le(s) sp�cifi�(s).
```

tomv564:
I like replace, best of both worlds!

PR #806, opened 2019-12-01
fix: remove reference to obsolete setting

Narretz:
While debugging, I found a reference to a removed setting.
I think it's nice to have this setting, so instead of removing it, I used the show_view_status setting. Let me know if I should just remove it altogether.


tomv564:
Let's remove the setting altogether!

Narretz:
Hey @tomv564 thanks for the quick response.

The setting I changed  the condition too is still used in another place (to show LSP status in the status bar)
Do you mean I should:
- remove the condition completely (always display the initialization message)
- remove the condition and the status display
- none of the above ;)

tomv564:
Let's remove `show_status_message` from all code ( I think there's another occurrence in `types.py`), and always display the initialisation message.

It looks like the setting was never read from sublime preferences, hardwired to always True :)

Narretz:
Right, I missed the ref in types.py

All done! 

Test Failure looks unrelated

PR #807, opened 2019-12-01
Add debug note if there's no matching config for a file

Narretz:
While failing to set up a config for a custom client, I was confused by LSP's silence. I finally tracked down the error to an incorrect syntax setting. 
This debug message should help others to find out where exactly LSP was missing something.

Note that I had to create a list from the filter - I suspect that any does not handle filter objects well (even though they are iterable). In my tests, any would return False for the filter object.


Narretz:
I realized that this isn't quite right, as this block checks the started status of the configs, not if they match the doc. will fix and  reopen

PR #809, opened 2019-12-02
Multiple-folder project support

### LINKED ISSUE #443: Support configuring child projects in single workspace

### LINKED ISSUE #785: Multiple folder with gopls
When I use multiple folders in one workspace with gopls sometimes a window pops up with the error:

> Your workspace root is X, but your module root is Y. Please add Y as your workplace folder.

This happens on OSX and I'm using go modules (i.e. neither X nor Y are in the GOPATH).

I found a couple of issues that sound related and it is also mentioned in the troubleshooting section (https://lsp.readthedocs.io/en/latest/troubleshooting/). However all the issues have been merged/closed quite a while ago.

tomv564:
Closes #443 - if people are satisfied they can setup multiple project folders into their monorepo-style workspace
Closes #785 - if someone can confirm this makes the "module in wrong path" error from Go go away.

**Implementation details**
* A window now stores a list of sessions per configuration, instead of just one session per config.
* Each session has a list of paths it handles, which can be queried with `session.handles_path(file_path)`
* Servers are sent all folders at startup, the first folder being the one a file was originally opened in.
  * If the server confirms support for multiple folders, the session keeps all folders. 
  * If not, only the folder with the opened file is kept. 
* We can send `didChangeWorkspaceFolders` when opening a file in a newly added folder. Added in https://github.com/tomv564/LSP/pull/809/commits/f765ea193c53c451f3cdcffcdcf6a0f3f5c22f49

**Not solved**

* <strike>We won't send `workspace/didChangeWorkspaceFolders` yet, need a reliable way to detect folders changing.</strike> See above
* Unsure if some servers do not support overlapping folders (and solve it in their VS Code extension). 

**To do:**

- [x] Support `workspace/workspaceFolders` request
- [x] Testing and fixing of projects that have child folders of another folder.
- [x] Review of TODOs and debug logging in code
- [x] Remove fake directory for folderless sessions (send `None` for rootUri as VS Code does)
- [x] Manual testing of quick switch project
- [x] Manual review of places relative paths are shown and navigated to (hover, diagnostics, references)

### DIFF
--- a/plugin/core/documents.py
+++ b/plugin/core/documents.py
@@ -98,6 +98,6 @@ class DocumentSyncListener(LSPViewEventListener):
         self.manager.documents.handle_view_saved(self.view)
 
     def on_close(self) -> None:
-        if self.view.file_name() and self.view.is_primary():
+        if self.view.file_name() and self.view.is_primary() and self.has_manager():
             self.manager.handle_view_closed(self.view)
             self.manager.documents.handle_view_closed(self.view)
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -3,13 +3,12 @@ from .protocol import Request
 from .transports import start_tcp_transport, start_tcp_listener, TCPTransport, Transport
 from .rpc import Client, attach_stdio_client
 from .process import start_server
-from .url import filename_to_uri
 from .logging import debug
 import os
 from .protocol import completion_item_kinds, symbol_kinds, WorkspaceFolder
 try:
     from typing import Callable, Dict, Any, Optional, List
-    assert Callable and Dict and Any and Optional and Transport and List
+    assert Callable and Dict and Any and Optional and Transport and List and WorkspaceFolder
 except ImportError:
     pass
 
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -176,7 +176,7 @@ class Session(object):
         self._on_post_exit = on_post_exit
         self.capabilities = dict()  # type: Dict[str, Any]
         self.client = client
-        self._workspace_folders = workspace_folders  # TODO: perhaps not until initialized?
+        self._workspace_folders = workspace_folders
         if on_pre_initialize:
             on_pre_initialize(self)
         self._initialize()
@@ -216,15 +216,14 @@ class Session(object):
             lambda result: self._handle_initialize_result(result))
 
     def _supports_workspace_folders(self) -> bool:
-        # 'capabilities': {'workspace': {'workspaceFolders': {'supported': True
         workspace_cap = self.capabilities.get("workspace", {})
         workspace_folder_cap = workspace_cap.get("workspaceFolders", {})
         return workspace_folder_cap.get("supported")
 
     def _handle_initialize_result(self, result: 'Any') -> None:
-        # only keep supported amount of folders
         self.capabilities = result.get('capabilities', dict())
 
+        # only keep supported amount of folders
         if self._supports_workspace_folders():
             debug('multi folder session:', self._workspace_folders)
         else:
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1,7 +1,7 @@
 from .types import ClientConfig, ClientStates, Settings
 from .protocol import Request
 from .transports import start_tcp_transport, start_tcp_listener, TCPTransport, Transport
-from .rpc import Client, attach_stdio_client
+from .rpc import Client, attach_stdio_client, Response
 from .process import start_server
 from .logging import debug
 import os
@@ -201,9 +201,18 @@ class Session(object):
             debug('single folder session:', self._workspace_folders[0])
         else:
             debug('multi folder session:', self._workspace_folders)
+
         self.state = ClientStates.READY
+
+        self.client.on_request(
+            "workspace/workspaceFolders",
+            lambda _params, request_id: self._handle_workspace_folders(request_id))
+
         if self._on_post_initialize:
             self._on_post_initialize(self)
+
+    def _handle_workspace_folders(self, request_id: int) -> None:
+        self.client.send_response(Response(request_id, [wf.to_dict for wf in self._workspace_folders]))
 
     def end(self) -> None:
         self.state = ClientStates.STOPPING
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -6,7 +6,7 @@ from .process import start_server
 from .url import filename_to_uri
 from .logging import debug
 import os
-from .protocol import completion_item_kinds, symbol_kinds
+from .protocol import completion_item_kinds, symbol_kinds, WorkspaceFolder
 try:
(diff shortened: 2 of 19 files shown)


plugin/core/sessions.py
```diff
@@ -71,11 +71,13 @@ def with_client(client: Client) -> 'Session':
     return session
 
 
-def get_initialize_params(project_path: str, config: ClientConfig) -> dict:
+def get_initialize_params(workspace_folders: 'List[WorkspaceFolder]', config: ClientConfig) -> dict:
```
rwols: Instead of `List[WorkspaceFolder]`, you could use `Iterable[WorkspaceFolder]` everywhere. The first item can be retrieved with `next(iter(workspace_folders))`

plugin/core/windows.py
```diff
@@ -379,45 +384,56 @@ def activate_view(self, view: ViewLike) -> None:
...
             return
 
-        project_path = workspace.path
+        # TODO: this throws away previous ad-hoc folders.
+        # The previous fix was to only set an ad-hoc folder once per window
+        self._workspace_folders = get_workspace_folders(self._window, file_path)
+
         self._window.status_message("Starting " + config.name + "...")
-        debug("starting in", project_path)
+        debug("starting in", self._workspace_folders[0].path)
```
rwols: You can use
```
debug("starting in", self._workspace_folders[0])
```
since `WorkspaceFolder` has a `__str__`

Although shouldn't it be

```
debug("starting in", self._workspace_folders)
```
? I guess it depends on wether the session supports multi folders or not?

plugin/core/sessions.py
tomv564: I really like this strategy for lists that are expensive-ish to compute and filtered-on multiple times when we only want 1 item.
In the case of WorkspaceFolder, we end up sending all of them anyways. 
Is there a performance benefit to programming against Iterable when receiving List? I'm a bit scared of things like empty generators being truthy, and iterators only being iterable once.

plugin/core/windows.py
tomv564: It's a bit misleading, but I took this message as stating the working directory of the server process. 

That probably also could use some improvement, if you have `/project/A` and `/project/B` folders, the server should perhaps have its working directory at `/project`?

tomv564: Given that we don't always start an executable, perhaps that logging should be moved and we could keep the logging after initialization that confirms which folders are part of the session.

rwols: > That probably also could use some improvement, if you have /project/A and /project/B folders, the server should perhaps have its working directory at /project?

I thought of `sublime.cache_path()/LSP/servername`, but that broke some assumptions for clangd. Perhaps the cwd should stay at the first workspace folder for now so as to not break any assumptions. Leave that problem for another PR :)

plugin/core/documents.py
```diff
@@ -98,6 +98,6 @@ def on_post_save_async(self) -> None:
         self.manager.documents.handle_view_saved(self.view)
 
     def on_close(self) -> None:
-        if self.view.file_name() and self.view.is_primary():
+        if self.view.file_name() and self.view.is_primary() and self.has_manager():
```
tomv564: Unrelated to PR, but closing windows/views resulted in an error

plugin/core/sessions.py
```diff
@@ -42,7 +41,8 @@ def with_client(client: Client) -> 'Session':
             tcp_port = socket.getsockname()[1]
             server_args = list(s.replace("{port}", str(tcp_port)) for s in config.binary_args)
 
-        process = start_server(server_args, project_path, env, settings.log_stderr)
+        working_dir = workspace_folders[0].path if workspace_folders else None
+        process = start_server(server_args, working_dir, env, settings.log_stderr)
```
tomv564: Possible breaking change: folder-less sessions no longer set a `cwd`.

plugin/core/sessions.py
```diff
@@ -71,11 +71,13 @@ def with_client(client: Client) -> 'Session':
...
 
 
-def get_initialize_params(project_path: str, config: ClientConfig) -> dict:
+def get_initialize_params(workspace_folders: 'List[WorkspaceFolder]', config: ClientConfig) -> dict:
+    first_folder = workspace_folders[0] if workspace_folders else None
     initializeParams = {
         "processId": os.getpid(),
-        "rootUri": filename_to_uri(project_path),
-        "rootPath": project_path,
+        "rootUri": first_folder.uri() if first_folder else None,
```
tomv564: Possible breaking change: Sending `None` for rootUri was always allowed according to the spec but we never did it.

plugin/core/sessions.py
```diff
@@ -160,28 +176,78 @@ def __init__(self,
...
     def get_capability(self, capability: str) -> 'Optional[Any]':
         return self.capabilities.get(capability)
 
-    def _initialize(self, project_path: str) -> None:
-        params = get_initialize_params(project_path, self.config)
+    def handles_path(self, file_path: 'Optional[str]') -> bool:
+        if not file_path:
+            return False
+
+        if not self._workspace_folders:
```
tomv564: Sessions without rootUri are expected to handle _any_ file - we only want one of them per window.

plugin/core/sessions.py
```diff
@@ -160,28 +176,78 @@ def __init__(self,
...
 
     def has_capability(self, capability: str) -> bool:
         return capability in self.capabilities and self.capabilities[capability] is not False
 
     def get_capability(self, capability: str) -> 'Optional[Any]':
         return self.capabilities.get(capability)
 
-    def _initialize(self, project_path: str) -> None:
-        params = get_initialize_params(project_path, self.config)
+    def handles_path(self, file_path: 'Optional[str]') -> bool:
```
tomv564: `Optional[str]` here because almost all the time we end up wanting to pass `view.file_name()` into it.

plugin/core/workspace.py
```diff
@@ -1,37 +1,67 @@
...
-    elif hasattr(view_or_window, 'active_view'):
-        view = view_or_window.active_view()
-        if not view:
-            return None
-        filename = view.file_name()
+    if file_path:
+        sorted_folders = []  # type: List[str]
+        if folders:
+            for folder in folders:
+                if file_path and file_path.startswith(folder):
```
AmjadHD: checking `file_path` is redundant

PR #810, opened 2019-12-02
Upgrade python and tools to recent version

tomv564:
Unsure if we really want to do this for ST3, but this PR will track the impact:

- [x] flake8: `E252 missing whitespace around parameter equals` on argument default values.
- [x] mypy: two small issues (incorrect format() call and new need for a type annotation.
- [x] unittest (mock) was using non-existing methods "assert_called_once" and "assert_not_called" (and silently passing)
- [x] Checking formatting correctness with `yapf --diff` works

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ python:
 before_install:
     - pip install "mypy==0.750"
     - pip install "flake8==3.7.9"
-    - pip install yapf
+    - pip install "yapf==0.29.0"
     - pip install python-coveralls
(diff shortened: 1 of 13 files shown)


tomv564:
I think the updates to mock and mypy has already caught more mistakes than expected, so I'd like to merge this shortly

rwols:
What about tox.ini? Does it need any updates? I like to have my local dev environment in sync with travis as much as possible.

tomv564:
Good point, updated and tested tox ( I don't use it myself much)

PR #811, opened 2019-12-02
Server logs to panel

rwols:
This works pretty good already.

* Move forward with OutputPanel from sublime_lib? What to do about the mypy errors?
* Should the server log panel be a circular buffer? It will append lines indefinitely now.
* It appears `attach_tcp_client` was an unused function? I removed it.
* Moved attachment of stderr to the WindowManager.

Screenshot from 2019-12-02 20-30-35


rchl:
So this will split server logs only? Does it means that it will be hard to correlate specific server messages with the protocol communication? I would sometimes do `console.error()` calls in a server to debug something.

I would love to have to have all LSP-related logs in one single panel and be able to decide what kind of logs should be shown there (protocol / server / etc. - could be done with a bitmask).

tomv564:
Yeah, correlating server output to payloads is nice, we could perhaps always have them in a server log panel?

The main console should then only contain LSP logging about package initialisation, configuration and failures to start servers - things new users would be most interested in.

Another thought - if you are running 2 servers in a given window, it becomes somewhat painful to identify which messages came from which server. We could create an output panel per running configuration?

plugin/core/windows.py
```diff
@@ -22,7 +26,7 @@
     assert LanguageConfig
     assert WorkspaceFolder
 except ImportError:
-    pass
+    from sublime_lib import OutputPanel
```
tomv564: Strange place for an import - is this where you are running into mypy errors?

plugin/core/windows.py
```diff
@@ -140,7 +149,7 @@ def get_document_state(self, path: str) -> DocumentState:
     def has_document_state(self, path: str) -> bool:
         return path in self._document_states
 
-    def _get_applicable_sessions(self, view: ViewLike, notification_type: 'Optional[str]'=None) -> 'List[Session]':
+    def _get_applicable_sessions(self, view: ViewLike, notification_type: 'Optional[str]' = None) -> 'List[Session]':
```
tomv564: I hope this will avoid conflicts depending on which PR merges first

plugin/core/rpc.py
```diff
@@ -28,35 +25,8 @@ def format_request(payload: 'Dict[str, Any]') -> str:
     return json.dumps(payload, sort_keys=False)
 
 
-def attach_tcp_client(tcp_port: int, process: 'subprocess.Popen', settings: Settings) -> 'Optional[Client]':
```
tomv564: I think this is old code that was not reviewed before refactoring (probably tcp_mode=host related).
It looks like `set_transport_failure_handler` is no longer hooked up for TCP clients.
TCP servers can also log on `stdout` and `stderr`, think that also has gone missing?
Perhaps we should take that fix outside of this PR.

plugin/core/windows.py
```diff
@@ -435,6 +449,11 @@ def _start_client(self, config: ClientConfig) -> None:
         if session:
             debug("window {} added session {}".format(self._window.id(), config.name))
             self._sessions[config.name] = session
+            transport = session.client.transport
+            if isinstance(transport, StdioTransport) and settings.log_stderr:
```
tomv564: Fishing up a process after the client is created is not the easiest as we've tried to abstract it from much of the code outside session :) 
Perhaps we should just expose an `Optional` process somewhere on Client and/or Session?

Otherwise, can `create_session` help? It is the only place that definitely knows if a process is created.

plugin/diagnostics.py
```diff
@@ -315,36 +304,35 @@ def end(self) -> None:
...
-            self._file_content += item + "\n"
+            self._file_content.append(self.format_diagnostic(diagnostic))
 
     def end_file(self, file_path: str) -> None:
         if self._file_content:
             panel_file_path = os.path.relpath(file_path, self._base_dir) if self._base_dir else file_path
-            self._to_render.append(" ◌ {}:\n{}".format(panel_file_path, self._file_content))
+            self.print(" ◌ {}:".format(panel_file_path))
+            for content in self._file_content:
+                self.print(content)
```
tomv564: There have been issues about the diagnostics panel flickering, which a generous contributor fixed by collecting output and only writing once. Perhaps we should keep that logic as you and I may not notice the flickering (or does OutputPanel improve it somehow even with `force_writes=True`)?

plugin/panels.py
```diff
@@ -0,0 +1,20 @@
...
+
+assert ensure_server_panel
+assert ensure_diagnostics_panel
+
+
+class LspTogglePanel(WindowCommand):
+    def run(self, panel_type: str) -> None:
+        if not isinstance(panel_type, str):
+            return
+        ensure_func = globals().get("ensure_{}_panel".format(panel_type), None)
```
tomv564: Given the low number of variations of panel I would prefer a simple `if` statement. I'm guessing pyflakes and mypy don't help us here either :)

plugin/panels.py
```diff
@@ -0,0 +1,20 @@
...
+from sublime_lib import Panel
+from sublime_plugin import WindowCommand
+
+assert ensure_server_panel
+assert ensure_diagnostics_panel
+
+
+class LspTogglePanel(WindowCommand):
+    def run(self, panel_type: str) -> None:
+        if not isinstance(panel_type, str):
```
tomv564: Even though theoretically users can put anything in the command arguments, I don't really feel the need for us to protect them from invalid input (especially as it's a silent failure now). Otherwise, the argument should be typed as `Any` ?

plugin/diagnostics.py
rwols: I decided to back off from sublime_lib. It has exactly the problems you mentioned.

plugin/core/rpc.py
rwols: I would like to keep this function removed, otherwise I would have to modify it for an extra `log_callback` parameter for `log_stream` while it's unused.

plugin/core/windows.py
rwols: When I put it above, I get massive mypy errors. Got frustrated, gave up with sublime_lib.

rwols:
- [x] Circular buffer for the panel
- [x] LSP payloads are also dumped in the panel
- [x] Highlight all the things!
- [ ] Fix pytest errors

Screenshot from 2019-12-06 12-11-28

Screenshot from 2019-12-06 12-22-24

rwols:
> Curious about your thoughts about OutputPanel, assuming we solve the import/typing issue, did it save us complexity?

Not that much. And as you pointed out, it doesn't buffer its `print` calls, so flickering might occur for some users. Decided to revert.

rchl:
Looking very nice!

I see on the screenshot that debug lines are still prefixed with `LSP`. Since all logs are LSP in that panel, should it be dropped? Can we show language server name instead?

tomv564:
Wow, that syntax looks amazing!!

rwols:
I think the on_post_exit handler is not invoked properly somehow in the tests. Have to investigate that.

Other than that, here are some things to consider with this diff:

1. sync requests are broken. Logging to a panel means invoking the ST api.
2. `"word_wrap"` would be nice, but it slows down the UI a lot somehow.

> I see on the screenshot that debug lines are still prefixed with LSP. Since all logs are LSP in that panel, should it be dropped? Can we show language server name instead?

Having "LSP" as a prefix allows me to syntax-highlight those lines differently from the server logs.

plugin/core/windows.py
rwols: Fixed by passing a callback along in `create_session`.

Commands/Default.sublime-commands
```diff
@@ -60,9 +60,14 @@
         "args": {}
     },
     {
-        "caption": "LSP: Show Diagnostics",
-        "command": "lsp_show_diagnostics_panel",
```
ayoub-benali: just noticed that this command is still referenced in `*.sublime-keymap`

tomv564:
If users have a key binding set up for that command, it will break right?
Perhaps this is not the most necessary refactoring

ayoub-benali:
Yes it broke for me, but I don't consider it a big issue. What we could do is keep the old command so that it simply forwards to "lsp_toggle_panel" for diagnostic. This way we acheive retro-compatibility. Also add a depreciation warning and remove it in future release ?

rchl:
> Having "LSP" as a prefix allows me to syntax-highlight those lines differently from the server logs.

1) You could probably highlight based on the `-->`/`<--` characters alone?
2) There should be a mention of language-server name on those lines. Otherwise, you don't know from which server that log is coming from.
3) Ideally, there would be separate panels for each server as I don't think it's useful for debugging to see them all together in the same context. But if not that, at least see point 2.

plugin/core/main.py
```diff
@@ -16,12 +18,18 @@
     pass
 
 
+def ensure_server_panel(window: sublime.Window) -> 'Optional[sublime.View]':
+    return ensure_panel(window, "server", "", "", "Packages/LSP/Syntaxes/ServerLog.sublime-syntax")
```
rchl: The "server" name for the panel is a bit ambiguous. Not clear that it's LSP's logging.

I mean when looking here:
Screenshot 2019-12-09 at 14 43 22

rwols:
> There should be a mention of language-server name on those lines. Otherwise, you don't know from which server that log is coming from.

Agreed, I'll modify this. What do you suggest for the panel's name? Just "LSP" perhaps? "LSP Logs"?

plugin/core/main.py
tomv564: Perhaps: "Language Servers" ?

tomv564:
Perhaps: "Language Servers" ?

rwols:
```

FAIL: test_formats_on_save (test_formatting.DocumentFormattingTests)

  File "/home/travis/.config/sublime-text-3/Packages/UnitTesting/unittesting/core/st3/case.py", line 28, in _executeTestPart

    yield from deferred

  File "/home/travis/.config/sublime-text-3/Packages/LSP/tests/test_formatting.py", line 47, in test_formats_on_save

    self.assertEquals(self.transport.sent[3]["method"], "textDocument/didChange")

AssertionError: 'textDocument/didOpen' != 'textDocument/didChange'

- textDocument/didOpen

?                 ^^ -

+ textDocument/didChange

?                 ^^^^^

```

:thinking:

tomv564:
We have a new flaky test I think. It eventually passed on master after 2 manual requests: https://travis-ci.org/tomv564/LSP/builds/623412425

rwols:
@rchl could you check if these latest changes are satisfactory?

rchl:
> @rchl could you check if these latest changes are satisfactory?

I can't say that I had a chance to play with logs too much but I've been running on this branch for a couple of days now and didn't see any serious issues.

I've noticed that syntax highlighting of really long log lines can fail (payload shows up in white) but that seems to be a SublimeText "issue".

Otherwise, I just saw this highlighting problem:
Screenshot 2019-12-12 at 16 07 38
Highlighting gets confused with string within a string (`"Foo 'bar'"`).

rwols:
That highlighting seems to be a bug in the Python syntax, perhaps you can post that log line here in text-form?

rchl:
I'll do in a while. Currently in transit.

I wonder though if we should switch to JavaScript syntax instead. It should handle sick payloads well.

rchl:
```
:: <-- lsp-vue textDocument/publishDiagnostics: {'diagnostics': [{'message': "Argument of type '{ props: { translations: { type: (ObjectConstructor | FunctionConstructor)[]; required: boolean; }; }; data: () => { showBanner: boolean; fallback: { title: string; text: string; agreeButtonText: string; closeTitle: string; }; }; computed: { ...; }; mounted: typeof mounted; methods: { ...; }; }' is not assignable to parameter of type 'new (...args: any[]) => any'.\n  Type '{ props: { translations: { type: (ObjectConstructor | FunctionConstructor)[]; required: boolean; }; }; data: () => { showBanner: boolean; fallback: { title: string; text: string; agreeButtonText: string; closeTitle: string; }; }; computed: { ...; }; mounted: typeof mounted; methods: { ...; }; }' provides no match for the signature 'new (...args: any[]): any'.", 'code': 2345, 'severity': 1, 'range': {'end': {'character': 0, 'line': 0}, 'start': {'character': 0, 'line': 0}}, 'source': 'Vetur'}], 'uri': 'file:///Users/cookie.vue'}
```

rwols:
It highlights a bit weird, but is with regards to matching brackets totally correct:
```
{
    'diagnostics':
    [
        {
            'message': "Argument of type '{ props: { translations: { type: (ObjectConstructor | FunctionConstructor)[]; required: boolean; }; }; data: () => { showBanner: boolean; fallback: { title: string; text: string; agreeButtonText: string; closeTitle: string; }; }; computed: { ...; }; mounted: typeof mounted; methods: { ...; }; }' is not assignable to parameter of type 'new (...args: any[]) => any'.\n  Type '{ props: { translations: { type: (ObjectConstructor | FunctionConstructor)[]; required: boolean; }; }; data: () => { showBanner: boolean; fallback: { title: string; text: string; agreeButtonText: string; closeTitle: string; }; }; computed: { ...; }; mounted: typeof mounted; methods: { ...; }; }' provides no match for the signature 'new (...args: any[]): any'.",
            'code': 2345,
            'severity': 1,
            'range':
            {
                'end':
                {
                    'character': 0,
                    'line': 0
                },
                'start':
                {
                    'character': 0,
                    'line': 0
                }
            },
            'source': 'Vetur'
        }
    ],
    'uri': 'file:///Users/cookie.vue'
}
```

rchl:
The part within the string that highlights in purple is still part of double-quoted string so I'm not sure why it would highlight differently. Unless python treats that in some special way but in LSP protocol, it shouldn't be handled in any special way, so from that perspective, it doesn't make sense.

The purple-highlight has extra scope `constant.other.placeholder.python` and it's not really a constant in my view so I think it's just buggy highlighting.

rwols:
True, there's some lookahead gone wrong again as is usual with smart lookaheads.

rwols:
I can reduce this highlighting bug to the following python one-liner
```
x="{:{ {}}"
```
The syntax thinks it's a formatting string apparently

But my Python interpreter thinks otherwise
```
>>> x="{:{ {}}".format("1", "2", "3")
  File "<stdin>", line 1, in <module>
ValueError: unmatched '{' in format spec
```

rwols:
I think this is pretty much done
Screenshot from 2019-12-12 23-06-59

rchl:
Is the highlighting issue we've talked about above considered fixed or not? It looks like it isn't from my testing.

rwols:
@rchl I'm of the opinion that it is a bug in the python syntax. I've made an issue for it: https://github.com/sublimehq/Packages/issues/2232. if that gets fixed there then it'll be auto-fixed for the language servers panel too.

plugin/core/windows.py
```diff
@@ -618,20 +675,38 @@ def __init__(self, configs: GlobalConfigs, documents: 'Any',
...
+    def set_server_panel_factory(self, factory: 'Callable') -> None:
+        self._server_panel_factory = factory
+
+    def set_settings_factory(self, settings: Settings) -> None:
+        self._settings = settings
+
     def lookup(self, window: 'Any') -> WindowManager:
         state = self._windows.get(window.id())
         if state is None:
+            if not self._settings:
```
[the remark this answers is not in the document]
tomv564: Why are the `settings` being moved to Window as part of this pull request?

plugin/core/panels.py
```diff
@@ -78,8 +105,26 @@ def run(self, edit: sublime.Edit, characters: 'Optional[str]' = "") -> None:
...
+
+
+class LspUpdateServerPanelCommand(sublime_plugin.TextCommand):
+    def run(self, edit: sublime.Edit, prefix: str, message: str) -> None:
+        with mutable(self.view):
+            self.view.insert(edit, self.view.size(), "{}: {}\n".format(prefix, message))
+            total_lines, _ = self.view.rowcol(self.view.size())
+            point = 0  # Starting from point 0 in the panel ...
+            regions = []  # type: List[sublime.Region]
+            for _ in range(0, max(0, total_lines - SERVER_PANEL_MAX_LINES)):
```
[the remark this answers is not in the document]
rwols: > We are erasing multiple lines while only writing 1 line?

It's entirely possible that a window/logMessage contains line terminators. In fact clangd does this sometimes :)

> Not that I mind erasing multiple lines - there's an opportunity to only clean up every 5 or 10 lines logged to amortise the cost a bit.

I want to leave that optimization for another PR, this one is already quite large :(

plugin/core/main.py
```diff
@@ -1,13 +1,16 @@
 import sublime
 
-from .settings import settings, load_settings, unload_settings
-from .logging import set_debug_logging, set_server_logging
-from .registry import windows, load_handlers, unload_sessions
+from .logging import set_debug_logging
+from .logging import set_exception_logging
```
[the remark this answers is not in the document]
rwols: I compactified the imports in 11b5452. However, I believe writing out every import on a line is better:

1) I can select all the imports and hit <kbd>F5</kbd> (sort lines) to sort all the imports
2) It's clearer to the reader (IMO) what is being used, and how much of a module is being used

plugin/core/rpc.py
```diff
@@ -1,13 +1,12 @@
...
-from .transports import TCPTransport, StdioTransport, Transport
-from .process import attach_logger
+from .transports import StdioTransport, Transport
 try:
     import subprocess
-    from typing import Any, List, Dict, Tuple, Callable, Optional, Union
+    from typing import Any, List, Dict, Tuple, Callable, Optional, Union, Mapping
     # from mypy_extensions import TypedDict
     assert Any and List and Dict and Tuple and Callable and Optional and Union and subprocess
+    PayloadLike = Union[Optional[Mapping[str, Any]], Optional[List], None]
```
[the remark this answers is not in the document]
rwols: I removed PayloadLike entirely

plugin/core/windows.py
```diff
@@ -607,6 +623,47 @@ def _handle_server_crash(self, config: ClientConfig) -> None:
...
+        if self._settings.log_stderr:
+            self._handle_server_message(name, message)
+
+    def _handle_show_message(self, name: str, params: 'Any') -> None:
+        self._sublime.status_message("{}: {}".format(name, extract_message(params)))
+
+    def _handle_log_payload(self, name: str, direction: str, method: 'Optional[str]', request_id: 'Optional[int]',
+                            payload: 'PayloadLike') -> None:
+        # If "log_payloads" == True, ignore "log_debug" and just print the entire line.
+        if not self._settings.log_debug and not self._settings.log_payloads:
```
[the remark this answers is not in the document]
rwols: Replaced by a logger class (called PreformattedPayloadLogger) in rpc.py with dedicated methods. I started out with an abstract base class but then felt that was pointless.

The WindowManager now only has to adjust the server_name and the sink of that PreformattedPayloadLogger.

One nice consequence of this setup is that if we leave the sink alone, the Client will simply log to the console again :)

plugin/core/windows.py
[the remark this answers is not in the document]
rwols: WindowManager needs to know wether it's okay to print stderr logging to the panel: see _handle_stderr_log method

plugin/core/rpc.py
```diff
@@ -69,11 +38,76 @@ def try_terminate_process(process: 'subprocess.Popen') -> None:
...
+        direction = "unhandled" if unhandled else Direction.Incoming
+        self.log(self.format_notification(direction, method), params, self.settings.log_payloads)
+
+
 class Client(object):
     def __init__(self, transport: Transport, settings: Settings) -> None:
         self.transport = transport  # type: Optional[Transport]
         self.transport.start(self.receive_payload, self.on_transport_closed)
         self.request_id = 0
+        self.logger = PreformattedPayloadLogger(settings, "server", debug)
```
rwols: A client will print to the ST console by default. The sink is modified in WindowManager to print to panel instead.

rwols:
> Perhaps we can do an int(payload["id"]) if "id" in payload else None or something?

You're right...

> Nice changes in the last round!

Thanks!

> I'd still like to avoid the deprecation of LspShowDiagnosticsPanelCommand, but otherwise it's ready to merge!

I was about to write this, hold on for just this evening :)

PR #813, opened 2019-12-03
Closes #812 added R to default config and to documentation

### LINKED ISSUE #812: There is no mention of R in the docs, nor how to set it up in context of this package
Been setting up LSP today and noticed there was no documentation for R strangely, might be worth letting people know that they have to first do:

```
R$> install.packages("languageserver")
```

and then add something to the tune of:

```
		"rlang":
		{
			"command":
			[
				"R", "--slave", "-e", "languageserver::run()"
			],
			"languageId": "r",
			"scopes":
			[
				"source.r"
			],
			"syntaxes":
			[
				"Packages/R/R.sublime-syntax"
			],
			"enabled": true
		}
```

to their `LSP.sublime-settings` to get stuff working, or maybe even add this to the default clients as I think it should be the same in almost all cases?

HarrisonWilde:
Following on from issue #812 have added R to the default list of clients in `LSP.sublime-settings` and added a note to the docs to install `languageserver` from CRAN first.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -100,15 +100,7 @@ Requires installation of the `languageserver` package from CRAN:
 
 See the [CRAN mirrored package on GitHub](https://github.com/cran/languageserver) for more information and up-to-date installation instructions.
 
-```json
-"rlang":
-{
-  "command": ["R", "--slave", "-e", "languageserver::run()"],
-  "scopes": ["source.r"],
-  "syntaxes": ["Packages/R/R.sublime-syntax"],
-  "languageId": "r"
-}
-```
+Once this has been done the language server need only be enabled to work, as it is included amongst the default clients.
 
 
 ### Scala<a name="scala"></a>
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -92,6 +92,25 @@ Goes well with the [Rust Enhanced package](https://github.com/rust-lang/rust-enh
...
+"rlang":
+{
+  "command": ["R", "--slave", "-e", "languageserver::run()"],
+  "scopes": ["source.r"],
+  "syntaxes": ["Packages/R/R.sublime-syntax"],
+  "languageId": "r"
+}
+```
+
+
```
rwols: Does this still need documentation if it's going to be part of the default clients? Should the documentation not  just say to `enable` the language server?

HarrisonWilde: Good point, was following other examples in the docs but realised they weren't all just repeating defaults in `clients`, will change to as you say

PR #819, opened 2019-12-07
Rename TestRenderer to MockRenderer to fix pytest warning

rwols:
This fixes the following warning when running tox:

```
=========================================== warnings summary ============================================
plugin/core/test_signature_help.py:118
  /home/raoul/.config/sublime-text-3/Packages/LSP/plugin/core/test_signature_help.py:118: PytestCollectionWarning: cannot collect test class 'TestRenderer' because it has a __init__ constructor (from: plugin/core/test_signature_help.py)
    class TestRenderer(ScopeRenderer):

-- Docs: https://docs.pytest.org/en/latest/warnings.html
```


tox.ini
```diff
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py35
+envlist = py3
```
tomv564: unintentional?

rwols: Apologies for the wording in the title. This is intentional as I don't have python 3.5 installed on my system (I don't think many people have 3.5 installed, instead 3.7 or 3.8).

I could use python virtual envs, but changing the requirement from py35 to py3 just seemed more practical.

tomv564: Yeah, I'm in the same boat, change it to `py36` every time (and have been close to committing that change). 

The idea was to keep it the same as the CI version - perhaps the difference is not that great with newer python versions?

PR #820, opened 2019-12-07
Prevent an AttributeError in pytests

rwols:
Between the "while self.process" and "self.process.stdout", the process
might become None.

This catches the following AttributeError that might occur at times when running tox:

```
~/.c/s/P/LSP (master|✚1) $ tox
py3 installed: attrs==19.3.0,coverage==4.5.4,entrypoints==0.3,flake8==3.7.9,importlib-metadata==1.2.0,mccabe==0.6.1,more-itertools==8.0.1,mypy==0.750,mypy-extensions==0.4.3,packaging==19.2,pkg-resources==0.0.0,pluggy==0.13.1,py==1.8.0,pycodestyle==2.5.0,pyflakes==2.1.1,pyparsing==2.4.5,pytest==5.3.1,pytest-cov==2.8.1,six==1.13.0,typed-ast==1.4.0,typing-extensions==3.7.4.1,wcwidth==0.1.7,yapf==0.29.0,zipp==0.6.0
py3 run-test-pre: PYTHONHASHSEED='1114725583'
py3 runtests: commands[0] | mypy plugin
Success: no issues found in 59 source files
py3 runtests: commands[1] | flake8 plugin tests
py3 runtests: commands[2] | pytest --quiet plugin
......................................................................Exception in thread Thread-2:
  File "/usr/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.7/threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "/home/raoul/.config/sublime-text-3/Packages/LSP/plugin/core/transports.py", line 197, in read_stdout
    header = self.process.stdout.readline()
AttributeError: 'NoneType' object has no attribute 'stdout'

........................sss [ 85%]
................                                                                                  [100%]

```


plugin/core/transports.py
```diff
@@ -215,7 +215,7 @@ def read_stdout(self) -> None:
                         content_length = 0
                     state = STATE_HEADERS
 
-            except IOError as err:
+            except (AttributeError, IOError) as err:
```
tomv564: I'd like a different fix for this - there is no clarification why AttributeError is in there and if it's only the tests that trigger it then the fix should be in test code.

rwols: After thinking about it for a while the process may also become None in production code. Who makes sure that the process is still alive after checking wether it is alive?

The only way to handle this is by "just reading from it", and using exception mechanism to handle sudden failure.

It's kind of the same situation with reading files. You can check wether a file is present on the filesystem, but to really make sure that the file exists you have to open it and handle an exception if opening the file fails.

tomv564: You are right. It would be nice to avoid catching _all_ AttributeErrors in this code. I thought about adding a getter to `self.process.stdout` but the point is we _do_ want the same exception handling logic anyways.

If there is no obvious way to only catch the missing stdout then let's go ahead with this.

tomv564: This error still feels really broad to catch here.

 I think we can declare a `UnexpectedProcessExitError` and do something like this:
```python
def safe_stdout(self) -> 'IO[Any]':
  if self.process:
    return self.process.stdout
  else:
    raise UnexpectedProcessExitError()
```

rwols: I'll see what I can do :)

tomv564:
Thanks @rwols !

plugin/core/transports.py
```diff
@@ -185,6 +185,12 @@ def close(self) -> None:
         self.send_queue.put(None)  # kill the write thread as it's blocked on send_queue
         self.on_closed()
 
+    def _checked_stdout(self) -> 'IO[Any]':
+        if self.process:
+            return self.process.stdout
+        else:
+            raise UnexpectedProcessExitError()
+
```
rwols: I'm going to come across as super nitpicky now, but this is incorrect again. There's the possibility that an AttributeError will bubble up from this function.

The absolute sure way to handle this is

```python
def _checked_stdout(self) -> 'IO[Any]':
    try:
        return self.process.stdout
    except AttributeError as ex:
        raise UnexpectedProcessExitError() from ex
```

tomv564: Sure, but does it happen in your test now?

PR #821, opened 2019-12-08
Add testing transport, use real client, test sync and formatting

tomv564:


tomv564:
You could, perhaps keep it separate? I also wanted to have mypy coverage reports somewhere for example. 

I'd like to avoid mixing it because with these tests you are running logic in unrelated handlers (think of all the code running on selection change) without ever checking the result.

PR #824, opened 2019-12-12
Update Python Language Server setup instructions

OlegSmelov:
I wanted to make Python Language Server installation instructions as clear as possible to make life easier for others.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -34,27 +34,55 @@ Be sure to install [Vue Syntax Highlight](https://packagecontrol.io/packages/Vue
 
 ### Python<a name="python"></a>
 
-`pip install python-language-server`
+There are at least two language servers, use either one.
 
-See: [github:palantir/python-language-server](https://github.com/palantir/python-language-server)
+#### Palantir Python Language Server
 
-Alternatively, Microsoft's python language server (using .NET Core runtime)
+```sh
+pip install python-language-server
+```
 
-[Instructions here](https://github.com/Microsoft/python-language-server/blob/master/Using_in_sublime_text.md)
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -34,28 +34,52 @@ Be sure to install [Vue Syntax Highlight](https://packagecontrol.io/packages/Vue
...
+Make sure you can run `pyls` in your terminal. If you've installed it into a virtualenv, you might need to override the path to `pyls` in global LSP settings (Package Settings -> LSP -> Settings):
 
-Use virtualenv adding the following settings :
 
 ```json
-"settings": {
+{
+    "clients": {
+        "pyls": {
+            "command": ["/Users/mike/.virtualenvs/pyls-virtual-env/bin/pyls"]
```
rwols: Perhaps you can mention that `Users/mike` should be replaced by the path to the reader's `pyls`. For an example of that, see the `dart` configuration.
You might also want to mention to set `"enabled": true` here because there have been users in the past that got confused about that boolean setting.

docs/index.md
```diff
@@ -34,28 +34,52 @@ Be sure to install [Vue Syntax Highlight](https://packagecontrol.io/packages/Vue
...
+```
+
+If you use a virtualenv for your current project, add a path to it in your [project configuration](https://www.sublimetext.com/docs/3/projects.html) (Project -> Edit Project):
+
+```json
+{
+    "settings": {
         "LSP": {
             "pyls": {
+                "enabled": true,
```
rwols: And then mention here, that if the user desires to disable `pyls` by default, but enable it on a per-project basis, then this is possible via this route.

OlegSmelov:
Updated. Added some comments in sample setting files.

PR #826, opened 2019-12-17
Fix links to documentation in README

jwortmann:
This just corrects two broken links to the documentation which didn't work correctly and fell back to "Docs > Home" instead.


rwols:
Thanks!

PR #827, opened 2019-12-18
check if last_location is -1

randy3k:
it may happen with the server returns error on `textDocument/completion`


tomv564:
Seems reasonable but how do we end up in `on_completion_inserted` when the server returns no items? `on_query_completions` should already set the `last_location` to something non-negative?

randy3k:
In my case, it happens when the server has an run time error and returns an error code and the menu is showing regular word completions.

PR #831, opened 2019-12-19
WIP: Remove blocking requests in favor of saving twice

rwols:
formatting.py has an annoying state machine now, like in completion.py.

I couldn't account for the sequential nature of applying and sending
didChange notifications to multiple sessions. That functionality has
been omitted and I'm printing a warning now. Such a feature would be
easy to write with asyncio.

- [ ] All applicable sessions must be able to format, in a sequential manner


rwols:
@rchl, you might want to try this branch and see if your timeouts disappear.

I'm not sure we should go ahead with this. The nice thing about synchronous requests is that it's easy to write sequential code. Whereas with the callback-based approach for `send_request` it's going to be hard to do things sequentially. It also saves twice, but in practise I don't think that's a major concern.

On the other hand, this does unblock the UI entirely.

rchl:
When working with *.vue files that have both `lsp-eslint` ad `lsp-vue` servers running, I'm getting:
```
LSP: WARNING: Can only run formatting for the first language server active
```

rwols:
OK, so we definitely want all of the applicable sessions to format the document, in a sequential manner.

frou:
I have my own older (non-LSP) formatting plugin that saves the file to disk twice, and the bad thing about that is that when using a build tool or similar that watches the filesystem for changes, the double-save causes it to unnecessarily thrash.

rchl:
I'm also usually working with projects that have file watchers set up, but not sure it would cause a problem to save file twice as those watchers are configured with some small delay before change handlers are run. But for sure it's not ideal to save twice.

rwols:
Looks like this approach has no future (pun intended :)).

Making the programming flow sequential in the face of callbacks is absolutely annoying. I tried using/writing a Future object, but I realized that was going to be a massive undertaking so I stopped. Given that `asyncio` has coroutines (a nice abstraction around Future/Promise objects), it would become really easy to write sequential non-blocking code.

That said, once we return from the `on_pre_save` event callback, ST will save the buffer. So we're going to be stuck with locks anyway, even with `asyncio`.

rchl:
GitGutter is using an implementation of a Promise/Future (which I believe I've written ;)) - https://github.com/jisaacks/GitGutter/blob/master/modules/promise.py

I'm not sure how feasible it would be to use it here. Maybe @deathaxe's experience can help in t

deathaxe:
GitGutter's _Promise_ class is a very basic and limitted version of a python _Future_ with the goal to encapsulate all the callback handling required for synchronizing/chaining asynchronous requests.

Building a project on such an approach is a design decision which needs to be made at the beginning. Such a change would have a very big impact on nearly everything.

From a today's perspective I wouldn't go the _Promise_ way again. It makes even simple things quite complicated looking.

I am with @rwols about `async/await` using the `asyncio` library would propably result in much more simple/natural/pythonic looking code even though it is a big design decision as well.

PR #832, opened 2019-12-21
Remove quotes from types

rwols:
This adds a typing.py file that takes care of all the annoying boilerplate. Single-quotes can be removed from type signatures now. Also important is that we don't have to `assert` all the types anymore, allowing `flake8` to tell us which types are imported but unused.

It's a process that can be gradual, so I'll stop here and make a pull request.


tomv564:
This will become legacy baggage as soon as we're all on python 3.8 :(

rwols:
Hmm, in what way would these changes make it harder to transition to py38? They make it easier to transition to py38. All you'd have to do is to remove the typing.py file and replace "from .typing" with "from typing" in all files.

tomv564:
> Hmm, in what way would these changes make it harder to transition to py38?

It would just be easier to do this at the same time as we cut off python 3.3 support, no need for the typing.py add-on then. 

It's not that it's a bad idea, I just won't have the time and energy to look into these kind of fixes.

rwols:
No worries Tom, I'll gradually convert each file over time.

PR #833, opened 2019-12-21
WIP: Test language server

rwols:
This is a "fake" language server so that we can do integration tests.

You can request `$test/getReceived` to await a sent method.
You can set its mock response via `$test/setResponse`.

I've also made sure to await until the session has really started, and await until the session has truly ended. The `TextDocumentTestCase` tests the entire real chain from the transport all the way to the view.

I've also removed all `yield 100` awaits by various `yield from self.await_message(...)` message calls.

I've put all "single document tests" into the `test_single_document.py` file for easier bookkeeping. It's possible to revert this.

In total the test time is now about 10 seconds.


rwols:
Whelp, it needs more work to get it working on travis.

PR #836, opened 2019-12-24
Document next/prev diagnostics in keymaps

predragnikolic:


Keymaps/Default (Linux).sublime-keymap
```diff
@@ -28,6 +28,15 @@
...
     { "keys": ["shift+f12"], "command": "lsp_symbol_references", "context": [{"key": "setting.lsp_active"}]},
 
+    // Next/Prevous diagnostics
+    { "keys": ["f8"], "command": "lsp_next_diagnostic"},
+    { "keys": ["shift+f8"], "command": "lsp_previous_diagnostic"},
+    { "keys": ["escape"], "command": "lsp_hide_diagnostic", "context":
+        [
+            { "key": "setting.lsp_diagnostic_phantom"}
+        ]
+    },
```
rwols: All these keybindings should have the `{"key": "setting.lsp_active"}` context, otherwise some people could complain  about how this plugin hijacks other plugins' keybindings.

rwols:
With the "escape" key bound to "lsp_hide_diagnostic", it seems I cannot close my panels anymore (with the "escape" key). How can we ensure that "escape" is handled by "lsp_hide_diagnostic" only when a diagnostic is visible?

rwols:
Perhaps we need an `is_enabled` method in the `LspHideDiagnosticCommand` class...

predragnikolic:
I just noticed that the toggling of the diagnostics panel doesn't work.

if you open `Default (Windows).sublime-keymap`
and scroll to 
```
    // Show/Hide Diagnostics Panel
    { "keys": ["ctrl+alt+m"], "command": "lsp_toggle_panel", 
```

I noticed that there is a `lsp_toggle_panel` command.

But there is nowhere in the lsp project a class named `LspTogglePanelCommand`? 
I belive that that is the reason why I can't toggle the diagnostics panel.

Which exact panels you cannot close? 
Because the escape key closes all panels, that I tried.

rwols:
> Which exact panels you cannot close? Because the escape key closes all panels, that I tried.

When I have focus on a view that is not a panel, I can press "escape" to close the (not in focus) panel.

Now, when I press "escape" nothing happens. I have to focus the panel first, by clicking in it, and then I can close it with "escape".

> I noticed that there is a lsp_toggle_panel command. But there is nowhere in the lsp project a class named LspTogglePanelCommand? I belive that that is the reason why I can't toggle the diagnostics panel.

Ah, forgot to change those command names in the keybindings file :)

predragnikolic:
Ok I tried to give focus to the view, or to the sidebar.
I can still press `esc` and the panel will close.
good

rwols:
I guess it's some weird combination of keybindings on my end then.

predragnikolic:
Try deleting all your custom keybinding(create a backup :D).
and see if it is still happening.

I tired that and it still works as expected.

PR #837, opened 2019-12-24
Maybe simpler naming of commands for toggeling the panels? :)

predragnikolic:
I know, I am nitpicking,

But I just can't seem to remember the new commands for toggling the panel. :man_facepalming: 
Can we name them with less words and maybe rearrange the words to be more easy to remeber? :D

### DIFF
--- a/Commands/Default.sublime-commands
+++ b/Commands/Default.sublime-commands
@@ -60,11 +60,11 @@
         "args": {}
     },
     {
-        "caption": "LSP: Toggle Panel: Language Servers",
+        "caption": "LSP: Toggle Log Panel",
         "command": "lsp_toggle_server_panel",
     },
     {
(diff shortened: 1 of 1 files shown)


Commands/Default.sublime-commands
```diff
@@ -60,11 +60,11 @@
         "args": {}
     },
     {
-        "caption": "LSP: Toggle Panel: Language Servers",
+        "caption": "LSP: Toggle Log Panel",
```
predragnikolic: Maybe event "LSP: Toggle Console"

rwols: I'm okay with any name :)

rwols:
Hmm, it seems the travis reporting mechanism is broken. It is still building pull-requests though.

predragnikolic:
Should I close this?

rchl:
I think new names are better.

PR #847, opened 2019-12-29
Prevent showing the AC on word separators

predragnikolic:
Not all characters should trigger opening the auto complete panel.
This PR will remove `complete_all_chars` in favor of a more configurable setting `auto_complete_ignored_triggers`.

Before:
Note. the autocomplete opens as soon as I type `,` which is not desired in this case. Because I intended to enter a new line.
bad

After:
Note. the autocomplete will not open as soon as I type `,` which is desired in this case. Because I intended to enter a new line.
good

This issue was reported on discord as well.
2019-12-29-234134_738x314_scrot


plugin/completion.py
```diff
@@ -260,7 +255,10 @@ def do_request(self, prefix: str, locations: 'List[int]') -> None:
         if not client:
             return
 
-        if settings.complete_all_chars or self.is_after_trigger_character(locations[0]):
+        prev_point = locations[0] - 1 if locations[0] - 1 >= 0 else 0
+        prev_char = view.substr(prev_point)
+
+        if prev_char in self.trigger_chars or prev_char not in self.ignored_chars:
```
rwols: Perhaps we can keep the setting as it is, but change its behavior so that it only triggers for word characters. We already have a function for that: https://github.com/sublimelsp/LSP/blob/96e28970c48680d00141b88631feedcb9c141cf2/plugin/core/documents.py#L46-L51

```suggestion
        if prev_char in self.trigger_chars or (settings.complete_all_chars and position_is_word(self.view, prev_point)):
```

predragnikolic: I tried using `position_is_word` :) 
Here is what the problem is with that approach.

If I use:
```python
def do_request(self, prefix: str, locations: 'List[int]') -> None:
        ...
        if prev_char in self.trigger_chars or(settings.complete_all_chars and position_is_word(self.view, prev_point)):
```
and I have a file empty python file. (the cursor is `|`)
```python
| 
```
And I manually trigger the autocomplete.
I will get no completions.
But I expect to see completions.

Because the previous cursor character is not a trigger char
and the character under cursor is not a word. 

With this PR, as it is now, 
you will get completions suggestion for, import, assert, class, def, ... 

Second example:
Lets say we have a python file:

```
# there is one space between cursor and 1
if 2 > 1 |
```
And you manually trigger the autocomplete.
You will get will get no completions.
Because the previous cursor character is not a trigger char
and the character under cursor is not a word. 

With this PR as it is now, you will get completions suggestion for, and, is, in, not.

The problem is even if we have completions at some position, 
and that position is not in the trigger chars and not a word,
when we trigger the autocomplete manually, 
we will get no results.

The solution to that problem is that we need a flag, `is_autocomplte_manually` in the `do_request` function.
 And if it is manually triggered, we will send the completion request to the server,
regardless if the character under cursor is a:
- trigger char 
- or a word 
- or neither of both.

That means no matter the case, if the autocomplete was triggered manually, we should send a completion request.

But I don't know how to determine if autocomplete was trigger manually or not. 
Or to be more precise, 
I don't know how determine if the `on_query_command` was called by the user pressing `crtl+space` or by the plugin like in the on_modified() defined in `CompletionHandler` class.
```
    def on_modified(self) -> None:
        ...
        if self.committing:
            self.committing = False
            self.on_completion_inserted()
        else:
            if self.view.is_auto_complete_visible():
                if self.response_incomplete:
                    # debug('incomplete, triggering new completions')
                    self.view.run_command("hide_auto_complete") # this here will hide the autocomplete
                    sublime.set_timeout(self.run_auto_complete, 0) # this will trigger on_query_completion
```

So instead of asking what characters should trigger the autocomplete,
I ask the question what characters should not trigger the autocomplete.
That is the reason why I did it this way.

But If you know a way of how to determine if the autocomplete was triggered manually,
let me know :)

rwols: Good point about the manual autocomplete. Still, we should strive for no extra settings. Every view has a setting called `word_separators`, perhaps we can use that?

```
>>> view.settings().get("word_separators")
'./\\()"\'-:,.;<>~!@#$%^&*|+=[]{}`~?'
```

predragnikolic: I removed the `auto_complete_ignored_triggers` setting and brought back `complete_all_chars` (although I have nothing against removing that setting too).

I also tried to simplify the completion file,
there was a lot of logic there.
Hope people will understand it easily now.

I made sure that all functionality is still there, but this need to be tested. I tested it with, LSP-typescript, LSP-intelephense(with this server I tested the is_incomplete flag), pyls, LSP-json...

Everything seems ok, 
but I would not merge this, until I test this for a week of real usage.

rwols: There are already a bunch of completion tests. Run "UnitTesting: Test Current File" from the command palette when you've opened tests/test_completion.py to check and expand your work.

rwols:
I think e74b430 should be a separate pull request...

plugin/completion.py
predragnikolic: thanks for the tip

predragnikolic:
I opened a separate PR #851,
This PR will be on hold until #851 is done.

predragnikolic:
I changed the PR,
Basically all I want is to prevent showing the AC on word separators.

I have a suggestion to remove the `complete_all_chars` settings?
Setting it to `false` will send completion requests only after a trigger char. 
Which is not that useful, IMO. 

I didn't used `complete_all_chars` setting at all in this PR. 
I can add it back if you want,
but I would set `complete_all_chars` to `false` only if I want to disable completions entirely... 
On the other hand I can use`"disabled_capabilities": ["completion"],` if I really want that...

If you want, 
I can add `complete_all_chars` back.
If not, 
I can make a commit to remove it entirely, in favor of having `"disabled_capabilities": ["completion"]` if someone ever wanted to set `complete_all_chars` to `false`

plugin/completion.py
```diff
@@ -260,7 +256,14 @@ def do_request(self, prefix: str, locations: 'List[int]') -> None:
...
 
-        if settings.complete_all_chars or self.is_after_trigger_character(locations[0]):
+        prev_point = locations[0] - 1 if locations[0] - 1 >= 0 else 0
+        prev_char = view.substr(prev_point)
+
+        if (
+            settings.complete_all_chars or
+            prev_char in self.trigger_chars or
+            prev_char not in self.ignored_trigger_chars
+        ):
```
rwols: This looks strange:

a) If settings.complete_all_chars is true, then the expression in the if-statement is always true, right?
b) If settings.complete_all_chars is false, then the expression in the if-statement should only be true when prev_char in self.trigger_chars is true, but it is also true when prev_char not in self.ignored_trigger_chars, which is most of the time.

I don't think we even need self.ignored_trigger_chars:

```suggestion
    if prev_char in self.trigger_chars or (settings.complete_all_chars and prev_char not in self.view.settings().get("word_separators")):
```

This says: if the previous character is a trigger character, then do the completion dance, otherwise, if complete_all_chars is true and the previous character is not a word separator, do the completion dance.

plugin/completion.py
```diff
@@ -260,7 +256,14 @@ def do_request(self, prefix: str, locations: 'List[int]') -> None:
         if not client:
             return
 
-        if settings.complete_all_chars or self.is_after_trigger_character(locations[0]):
+        prev_point = locations[0] - 1 if locations[0] - 1 >= 0 else 0
```
rwols: Nitpick: you can use `prev_point = max(0, locations[0] - 1)` here.

predragnikolic: I dind't know that, I like it :D

predragnikolic:
Hello, last night I wanted to write the test in completions you asked for...
 
> Try to add an extra test in test_completion.py. The cursor is, for example, at a comma. Then trigger completions, and then check that no completion widget is visible

But to my surprise, when running `UnitTesting: Test Current file` when in `test_completion.py`. Sublime started to freeze each time I run the command. 

So I looked a bit and found the cause of it.
It is this commit 946165c73390035113419de3b0b51f2f1ea6c933
I merged the latest master into this branch locally, and than I experienced constant frezess when trying to run tests in `test_completion.py`

I think it has something to do with the `self.ready_lock` but I am not a thread expert.

rwols:
It is strange that I don’t experience this. But I will push a new commit tonight. This was also reported in the chat for eslint.

tomv564:
Just caught up on discord and saw you tested the comma trigger issue. It's only on ST4, right?

Did anyone from the ST team confirm why completions trigger more often? I would be careful to start fixing ST4 issues unless behaviour changes (like these) are confirmed (and explained why) by them. Perhaps we can ask their advice on filtering the requests (instead of burdening the user with configuring a setting).

There is also the `st4-exploration` branch that you can park these changes in and pick them over to master once you decide to officially support ST4.

rwols:
I was under the impression that this happened for ST3 too. If it's only for ST4 then we should defer these changes for later.

In any case, perhaps we can also solve this through the `auto_complete_selector` view setting, by adding more punctuation scopes (except for the ones that are trigger characters).

predragnikolic:
> Just caught up on discord and saw you tested the comma trigger issue. It's only on ST4, right?

I was also under the impression that this happened for ST3(but it is not). 
I am going to immediately close this PR,  because the completion  file will be  rewritten for the ST4.

ST3
Screenshot 2020-01-12 at 02 40 51

ST4
Screenshot 2020-01-12 at 02 41 19

rwols:
Perhaps we can salvage the PR by writing a test that verifies that no AC widget is open upon writing a comma character?

It would succeed in ST3, but fail in ST4 (on build 4061 at least), giving us a reminder to fix this.

Moral of the story: test-driven-development ;)

PR #849, opened 2019-12-31
Allowing user to pick from multiple implementations when using 'Go to symbol implementation'

mrbrian:
When navigating to a symbol's implementation, the existing behaviour was to go to the first implementation it receives from LSP.

This pull request will pop up a dialog and let the user pick from multiple implementations.

Gif preview: https://github.com/mrbrian/LSP/blob/big_gifs/docs/images/goto_implementation.gif

I've only tested my changes with C++ & CCLS so far.  I'm not sure how to setup a test with Javascript.

### DIFF
--- a/plugin/goto.py
+++ b/plugin/goto.py
@@ -83,7 +83,7 @@ class LspGotoCommand(LspTextCommand):
                     flags=sublime.TRANSIENT | sublime.ENCODED_POSITION | sublime.FORCE_GROUP)
 
         window = sublime.active_window()
-        orig_sel = []
+        orig_sel = []  # type: List[sublime.Region]
         if window.active_view():
             view = sublime.View(window.active_view())
             sel = view.sel()
--- a/plugin/goto.py
+++ b/plugin/goto.py
@@ -73,10 +73,7 @@ class LspGotoCommand(LspTextCommand):
                 open_location(window, locations[idx])
             else:
                 if orig_view:
-                    orig_view.sel().clear()
-                    # orig_view.sel().add_all(orig_sel)
                     window.focus_view(orig_view)
-                    # orig_view.show(orig_sel[0])
 
         def highlight_entry(
                 window: sublime.Window,
@@ -90,9 +87,10 @@ class LspGotoCommand(LspTextCommand):
                     flags=sublime.TRANSIENT | sublime.ENCODED_POSITION | sublime.FORCE_GROUP)
 
         window = sublime.active_window()
+        view = self.view
         if response:
             # Save to jump back history.
-            get_jump_history_for_view(self.view).push_selection(self.view)
+            get_jump_history_for_view(view).push_selection(view)
             # TODO: DocumentLink support.
             if isinstance(response, dict):
                 locations = [process_response(response)]
@@ -103,7 +101,7 @@ class LspGotoCommand(LspTextCommand):
             elif len(locations) > 1:
                 window.show_quick_panel(
                     items=[display_name for file_path, display_name, rowcol in locations],
-                    on_select=lambda x: select_entry(window, locations, x, self.view),
+                    on_select=lambda x: select_entry(window, locations, x, view),
                     on_highlight=lambda x: highlight_entry(window, locations, x),
                     flags=sublime.KEEP_OPEN_ON_FOCUS_LOST)
             # TODO: can add region here.
--- a/plugin/goto.py
+++ b/plugin/goto.py
@@ -40,10 +40,10 @@ class LspGotoCommand(LspTextCommand):
                     request, lambda response: self.handle_response(response, pos))
 
     def handle_response(self, response: 'Optional[Any]', position: int) -> None:
-        def process_response_array(array):
-            return [process_response(x) for x in array]
+        def process_response_list(responses: list):
+            return [process_response(x) for x in responses]
 
-        def process_response(response):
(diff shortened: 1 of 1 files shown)


plugin/goto.py
```diff
@@ -8,7 +8,7 @@
 from Default.history_list import get_jump_history_for_view
 
 try:
-    from typing import List, Dict, Optional, Any
+    from typing import List, Dict, Optional, Any, Tuple
     assert List and Dict and Optional and Any
```
rwols: Python 3.3, the python version that SublimeText 3 uses, has no support for `typing`. We get around this issue by trying to import the module, catching an `ImportError` in SublimeText.

This implies that anything imported from the `typing` is unknown to the python interpreter in SublimeText, so anything imported from there needs to be in _single quotes_.

To make flake8 happy, you have to put anything you import from the typing module in the `assert` expression here.

You can check your work by running `tox` locally in the repository root folder.

plugin/goto.py
```diff
@@ -40,22 +40,71 @@ def run(self, edit: sublime.Edit, event: 'Optional[dict]' = None) -> None:
                     request, lambda response: self.handle_response(response, pos))
 
     def handle_response(self, response: 'Optional[Any]', position: int) -> None:
+        def process_response_list(responses: list) -> List[Tuple[str, str, Tuple[int, int]]]:
```
rwols: For instance, ST doesn't understand what this `List` class is, so you must wrap it in single quotes:

```
def process_response_list(responses: list) -> List[Tuple[str, str, Tuple[int, int]]]
```

plugin/goto.py
```diff
@@ -40,22 +40,71 @@ def run(self, edit: sublime.Edit, event: 'Optional[dict]' = None) -> None:
...
+        def highlight_entry(window: sublime.Window, locations: List[Tuple[str, str, Tuple[int, int]]], idx: int) -> None:
+            fname, file_path_and_row_col, rowcol = locations[idx]
+            row, col = rowcol
+            window.open_file(
+                    file_path_and_row_col,
+                    group=window.active_group(),
+                    flags=sublime.TRANSIENT | sublime.ENCODED_POSITION | sublime.FORCE_GROUP)
+
         window = sublime.active_window()
+        orig_sel = []
```
rwols: I also noticed by checking your branch out locally that this needs to be type-hinted.

```suggestion
        orig_sel = []  # type: List[sublime.Region]
```

plugin/goto.py
mrbrian: Thanks!  I think I now have tox running with the same environment as the github actions (python 3.7) and without errors.  However, now it fails on the did_close test in test_single_document.SingleDocumentTestCase.  

It did pass the test in an earlier run, so I don't think its related to this PR:
https://github.com/sublimelsp/LSP/commit/94a4759518ca44328325ed5bbf93dd586fe69227/checks?check_suite_id=379337072

rwols: Yeah, it's a flaky test, don't worry about it. I'm still in the process of fixing that.

plugin/goto.py
```diff
@@ -40,22 +40,72 @@ def run(self, edit: sublime.Edit, event: 'Optional[dict]' = None) -> None:
...
+                idx: int,
+                orig_view: sublime.View) -> None:
+            if idx >= 0:
+                open_location(window, locations[idx])
+            else:
+                if orig_view:
+                    orig_view.sel().clear()
+                    # orig_view.sel().add_all(orig_sel)
+                    window.focus_view(orig_view)
+                    # orig_view.show(orig_sel[0])
```
rwols: Nitpick: please remove commented-out code.

plugin/goto.py
```diff
@@ -40,22 +40,72 @@ def run(self, edit: sublime.Edit, event: 'Optional[dict]' = None) -> None:
...
+        def select_entry(
+                window: sublime.Window,
+                locations: 'List[Tuple[str, str, Tuple[int, int]]]',
+                idx: int,
+                orig_view: sublime.View) -> None:
+            if idx >= 0:
+                open_location(window, locations[idx])
+            else:
+                if orig_view:
+                    orig_view.sel().clear()
```
rwols: orig_view.sel().clear() removes all cursors. We've had some comments in the past that this is undesirable. Do we even need to call orig_view.sel().clear()? Perhaps we can just leave it out. Otherwise, we should add back the original selection. The choice is up to you.

plugin/goto.py
```diff
@@ -40,22 +40,72 @@ def run(self, edit: sublime.Edit, event: 'Optional[dict]' = None) -> None:
...
+                    # orig_view.sel().add_all(orig_sel)
```
rwols: Nitpick: please remove commented-out code.

plugin/goto.py
mrbrian: For simplicity I removed the selection clear, but ideally it would restore the original selection.  The only thing is that mypy falsely complains that sublime.Selection isn't iterable.  (looks like its a known issue in mypy https://github.com/python/mypy/issues/2220)  So I'm not sure if there's a nice way to ignore the error?

rwols: You may use `# type: ignore` after an expression/statement to make mypy ignore that line.

rwols:
Let's see how this fares in `master`.

Issue #845, opened 2019-12-29
Go to Definition by LSP seems not to utilize LSP properly for the new tab


jfcherng:
## Env

LSP: 0.9.3 with [LSP-intelephense](https://github.com/sublimelsp/LSP-intelephense)
LSP Server: [intelephense](https://github.com/bmewburn/vscode-intelephense) 
Windows 7 x64

By the way, LSP 0.9.2 is fine.

## Issue

New tabs opened by "LSP: Go to Definition" seem not to utilize LSP properly. After "lsp-intelephense initialized", there is no LSP functionality and no "lsp-intelephense" shown at the bottom-left corner. After closing the opened tab, the original tab has 2 `lsp-intelephense` at the bottom-left corner. If I do these steps again, there can be 3 `lsp-intelephense` and more.

Maybe related to Windows' path separator, idk. 

```
LSP: path not in existing lsp-intelephense session: e:\WWW\nthu-ee-iframe\public\include\list_faculty_functions.php
LSP: path not in existing lsp-intelephense session: e:\WWW\nthu-ee-iframe\public\include\list_faculty_functions.php
LSP: window 2 requests lsp-intelephense for e:\WWW\nthu-ee-iframe\public\include\list_faculty_functions.php
LSP: starting ['node', 'C:\\Users\\Clover\\AppData\\Roaming\\Sublime Text 3\\Packages\\LSP-intelephense\\node_modules\\intelephense\\lib\\intelephense.js', '--stdio']
LSP: window 2 added session lsp-intelephense
LSP: multi folder session: [WorkspaceFolder('nthu-ee-iframe', 'E:\WWW\nthu-ee-iframe')]
```

## Video Record 

- https://streamable.com/20wq7

predragnikolic:
I noticed this too.

It is not just LSP-intelephense :)
![good](https://user-images.githubusercontent.com/22029477/71551305-a2de9580-29e4-11ea-9f33-d685a7694feb.gif)

This bug can be reproduced with all servers :)

The minimal steps to reporduce is for LSP-json (or any other language server).
Open (or create) a `.json` file.
Open command pallete.
Type `View Pack File`
Type `.json` and select a json file not from the currently open folder.
This will print a message in the status line  `lsp-json` initialized. But the server di

rwols:
I can also reproduce this with python-language-server by doing "Go To Definition" on an imported module outside the project (for example `import html` in a py file and then goto def for that `html` token).

[...]

PR #850, opened 2020-01-01
Fix "goto def" for a foreign file (outside the workspace)

rwols:
- Take the first available language server if it's workspace-unaware
- Take the one and only language server if it's workspace-aware
- Ensure things work when ST is started with a foreign file
- Ensure sessions know what type of language server they are dealing with
  (i.e. workspace-aware or workspace-unaware) with a mutex lock.
  Actually, you can use that same mutex lock to check wether a
  session is ready (thus simplifying some test code)

The DocumentHandler needs to know wether a file is inside or outside the workspace for its _get_applicable_sessions method.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -5,6 +5,7 @@ from .rpc import Client, attach_stdio_client, Response
 from .process import start_server
 from .logging import debug
 import os
+import threading
 from .protocol import completion_item_kinds, symbol_kinds, WorkspaceFolder
 try:
     from typing import Callable, Dict, Any, Optional, List, Tuple
@@ -177,6 +178,7 @@ class Session(object):
         self._on_post_exit = on_post_exit
         self.capabilities = dict()  # type: Dict[str, Any]
         self.client = client
+        self.ready_lock = threading.Lock()
         self._workspace_folders = workspace_folders
         if on_pre_initialize:
             on_pre_initialize(self)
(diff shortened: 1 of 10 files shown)


rwols:
needs to be tested on windows (ping @jfcherng)

jfcherng:
@rwols tested ok for my use case in https://github.com/sublimelsp/LSP/issues/845

plugin/core/sessions.py
```diff
@@ -191,40 +193,53 @@ def get_capability(self, capability: str) -> 'Optional[Any]':
...
+            self._handle_initialize_result,
+            self._handle_initialize_error)
 
     def _supports_workspace_folders(self) -> bool:
+        assert self.ready_lock.locked()
         workspace_cap = self.capabilities.get("workspace", {})
         workspace_folder_cap = workspace_cap.get("workspaceFolders", {})
         return workspace_folder_cap.get("supported")
 
+    def supports_workspace_folders(self) -> bool:
```
predragnikolic: Why do we have two definitions for `supports_workspace_folders` and `_supports_workspace_folders`?
Can we have one?

rwols: _supports_workspace_folders is the actual implementation, supports_workspace_folders just puts a mutex lock around it. I've renamed _supports_workspace_folders to _unsafe_supports_workspace_folders. It's not supposed to be called outside the class.

rwols:
Let's bring this to master and see if any issues arise.

Issue #797, opened 2019-11-29
Some UnitTesting tests are failing with ST builds >= 4000


rwols:
The failing tests are

```
test_window_without_configs (test_configs.WindowConfigTests) ... FAIL
test_additional_edits (test_completion.QueryCompletionsTests) ... FAIL
test_dash_missing_from_label (test_completion.QueryCompletionsTests) ... FAIL
test_edit_after_nonword (test_completion.QueryCompletionsTests) ... FAIL
test_edit_before_cursor (test_completion.QueryCompletionsTests) ... FAIL
test_implement_all_members_quirk (test_completion.QueryCompletionsTests) ... FAIL
test_resolve_for_additional_edits (test_completion.QueryCompletionsTests) ... FAIL
test_simple_inserttext (test_completion.QueryCompletionsTests) ... FAIL
test_simple_label (test_completion.QueryCompletionsTests) ... FAIL
test_space_added_in_label (test_completion.QueryCompletionsTests) ... FAIL
test_var_prefix_added_in_insertText (test_completion.QueryCompletionsTests) ... FAIL
test_var_prefix_added_in_label (test_completion.QueryCompletionsTests) ... FAIL
test_var_prefix_using_label (test_completion.QueryCompletionsTests) ... FAIL
```

My installed packages:

```
		"AutomaticPackageReloader",
		"AvrAssembler",
		"ChaiScript",
		"CMake",
		"Dockerfile Syntax Highlighting",
		"FileIcons",
		"fish",
		"GitSavvy",
		"HexViewer",
		"OpenGL Shading Language (GLSL)",
		"Package Control",
		"PackageDev",
		"Protobuf Syntax Hightlighting",
		"SideBarTools",
		"Terminus",
		"TypeScript Syntax",
		"UnitTesting"
```

Narretz:
Where did you get Sublime Text v >4000 from?

rwols:
@Narretz see this post on the forums https://forum.sublimetext.com/t/emmet-2-0-alpha/47933/10

Narretz:
Thankls!

randy3k:
Ha, I am involved in 4/17 of your packages. Maybe I spent too much time in Sublime Text. 🤓

rwols:
fixed by #866

PR #851, opened 2020-01-03
Refactor completions

predragnikolic:
The goal of this PR is to make the completions file easier to understand.
All tests are passing except
```
FAIL: test_edit_before_cursor (test_completion.QueryCompletionsTests)
  File "/home/predrag/.config/sublime-text-3/Packages/UnitTesting/unittesting/core/st3/case.py", line 28, in _executeTestPart
    yield from deferred
  File "/home/predrag/.config/sublime-text-3/Packages/LSP/tests/test_completion.py", line 317, in test_edit_before_cursor
    '  override def myFunction(): Unit = ???')
AssertionError: '  def myF' != '  override def myFunction(): Unit = ???'
-   def myF
+   override def myFunction(): Unit = ???
```
That test is failing in the latest master also.
I know this has something to do with metals... That's all I know. :D don't know how to fix it

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -58,8 +58,9 @@
   // Show symbol action links in hover popup if available
   "show_symbol_action_links": false,
 
-  // Auto complete won't be triggered for these characters
-  "auto_complete_ignored_triggers": [","],
+  // Request completions for all characters if set to true,
+  // or just after trigger characters only otherwise.
+  "complete_all_chars": true,
 
   // Controls which hints the completion panel displays
   // "auto": completion details if available or kind otherwise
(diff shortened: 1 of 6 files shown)


rwols:
> That test is failing in the latest master also.

That is due to ST build 40xx changing the way completions behave: #797. 4061 somehow fixed most of those tests that used to fail, so I guess they have some sort of test suite at SublimeHQ too (hopefully). The only one still failing is the one you're experiencing now.

You should test your changes with the old ST, build 32xx.

That said, I think it's too risky and not worth our time to change the behavior of the completion.py files. It works for almost all users now. And with ST4, we're going to rewrite the completion logic entirely anyway.

predragnikolic:
I downloaded ST v3.2.2.
all tests are passing.

You have a point.
The completions file will be rewritten for the new ST version.
I will close this PR, so we can focus on other things.

PR #853, opened 2020-01-06
Session state

rwols:
A session is ready _or_ initializing if and only if
1) it is present in wm._sessions

A session is ready if and only if
1) it is present in wm._sessions
2) a call to session.handles_path(...) or session.supports_workspace_folders() returns

A session that is exiting is immediately popped from wm._sessions, so we cannot obtain a session that is exiting.

Since wm.get_session(...) always calls session.handles_path, we are guaranteed a ready session.

We have to await the initialize request because we need to determine wether the session is workspace-aware or workspace-unaware before calling session.handles_path.

Language servers that don't respond to the initialize request are disabled temporarily for the window.

- ~Let's try adding a Windows CI~
- [x] Fix all tests on Windows. Some processes were left hanging, others needed filename_to_uri.
- ~Find and make sure that language servers that do elaborate stuff in the initialize request like downloading source code still work.~

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,3 +37,28 @@ jobs:
     - run: |
         pip3 install coverage==4.5.4 codecov==2.0.15
         codecov
+
+  Windows:
+    runs-on: windows-latest
+    env:
+      CODECOV_TOKEN: ${{secrets.CODECOV_TOKEN}}
+      SUBLIME_TEXT_VERSION: "3"
+      SUBLIME_TEXT_ARCH: x64
+      PACKAGE: LSP
+    steps:
+    - uses: actions/checkout@v1
+    - run: python3 tests/server.py --version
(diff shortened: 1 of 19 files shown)


tomv564:
After reading about the issue with servers blocking on `initialize`, I remembered how LSP used to park a list of "documents to open while waiting for a session to start" somewhere on the WindowManager. That functionality was removed but it could be re-introduced to implement a "only start 1 session of a kind at a time" logic. 
Would be really nice to avoid blocking here!

rwols:
Interesting, perhaps we can go that route :)

"only start 1 session at a time"? Why would we want that? I'm confused about that sentence.

Multiple sessions can initialize at the same time right now. The current implementation is such that `get_sessions_for_view` returns an iterable of sessions when all sessions have initialized (they initialize concurrently).

The current design also allows an easy translation to asyncio, replacing the threading.Lock with an asyncio.Lock. That means we'll await non-blocking, instead of in a blocking manner right now.

PR #857, opened 2020-01-14
FIX AttributeError: module 'asyncio' has no attribute 'get_running_loop'

predragnikolic:
When running: `UnitTesting: Test current file` in test_completion.py
sublime would not run the tests 
and it would show a popup saying that the server didn't initialize in 3sec, 
because 'asyncio' has no attribute 'get_running_loop'.

All credit goes to @rwols.


rwols:
thanks

PR #862, opened 2020-01-24
Disable a server only for a specific folder of the project, if it fails to init

### LINKED ISSUE #860: 0.9.3 reintroduced issue #79
My project has multiple folders defined, with some being direct ancestors of one another. e.g.:
* `/foo/`
* `/foo/bar/`

The LSP server (`Flow`) is only valid inside a subfolder (e.g. `/foo/bar/`) and only applicable to files within.

Previously, (with https://github.com/sublimelsp/LSP/pull/81) I was able to just set my project's first folder entry to `/foo/bar/` to have LSP plugin start the LSP server in the correct directory.

Now that it uses workspaceFolders instead, I have no way of controlling the order of these folders and as a result LSP plugin tries to start the server in `/foo/` and fails.

I was able to temporarily fix it for myself by modifying `sorted_workspace_folders` in `workspace.py` to comment out the sorting logic:
```
def sorted_workspace_folders(folders: List[str], file_path: str) -> List[WorkspaceFolder]:
    sorted_folders = []  # type: List[WorkspaceFolder]
    for folder in folders:
        # if file_path and file_path.startswith(folder):
        #     sorted_folders.insert(0, WorkspaceFolder.from_path(folder))
        # else:
            sorted_folders.append(WorkspaceFolder.from_path(folder))
    return sorted_folders
```

This sorting logic probably needs to handle the case where multiple folders satisfy the prefix path requirement. Longest prefix length should probably win.

rwols:
Moreover, account for nested workspace folders.
Added a test for this.
Also, disable a language server if it responds with an error for the initialize request.
Flow returns an error when its version doesn't match with what is stated in .flowconfig.

Tested with Flow language server and a sublime project with the structure /foo, /foo/bar.

This results in the following behavior:
1. Language server is started for some folder /foo
2. It fails somehow. "blacklist" the server name + folder pair
3. A file is opened in another workspace.
4. Language is not blacklisted for this server name + folder pair, so we attempt to start a process
5. It succeeds.
6. User can move on with his/her work.

Thus there are two ways to disable a lang server:

1. Disable locally per workspace folder. We do this if it succeeds to start, but fails to initialize
2. Disable the configuration for the window. We do this if it doesn't start.

close #860

### DIFF
--- a/plugin/core/clients.py
+++ b/plugin/core/clients.py
@@ -1,7 +1,7 @@
 from .protocol import WorkspaceFolder
 from .sessions import create_session, Session
 from .settings import ClientConfig, settings
-from .typing import List, Dict, Tuple, Callable, Optional
+from .typing import Any, List, Dict, Tuple, Callable, Optional
 import os
 import sublime
 
@@ -28,15 +28,17 @@ def get_window_env(window: sublime.Window, config: ClientConfig) -> Tuple[List[s
 
 def start_window_config(window: sublime.Window,
                         workspace_folders: List[WorkspaceFolder],
+                        designated_folder: Optional[WorkspaceFolder],
                         config: ClientConfig,
                         on_pre_initialize: Callable[[Session], None],
-                        on_post_initialize: Callable[[Session], None],
+                        on_post_initialize: Callable[[Session, Optional[Dict[str, Any]]], None],
                         on_post_exit: Callable[[str], None],
                         on_stderr_log: Optional[Callable[[str], None]]) -> Optional[Session]:
     args, env = get_window_env(window, config)
     config.binary_args = args
     return create_session(config=config,
                           workspace_folders=workspace_folders,
+                          designated_folder=designated_folder,
                           env=env,
                           settings=settings,
                           on_pre_initialize=on_pre_initialize,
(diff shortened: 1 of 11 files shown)


rwols:
> The need to disable per folder seems reasonable (although fixing a fundamentally broken approach).

Yes, but I agree that a LanguageHandler is perhaps better suitable for this. It can override `on_start` and return `False` if there's not a .flowconfig in the folder. I should split this PR into two:

- one for the "disabling per workspace folder" functionality (debatable),
- and one for the "nested folders" bug (mergeable).

> I'm not sure why designated_folder is added everywhere when there is workspace_folders[0] ?

I found `workspace_folders` having two purposes to be too vague. Moreover, sorting the folders is too much work. All we need is the folder that contains the file path from the view that initially started the session. Hence the `max(...)` in workspace.py.

> What is hash in protocol for?

I want to put a WorkspaceFolder into a set:
```python
        self._blacklist = {}  # type: Dict[str, Set[Optional[WorkspaceFolder]]]
```
So I must define how to hash a WorkspaceFolder.

rwols:
After mulling it over for a while I think we can go ahead with this. I've added tests for the new _disable_temporarily functionality in test_windows.py.

PR #866, opened 2020-01-30
Add new completion API

### LINKED ISSUE #757: Incorrect additionalTextEdit applied when auto-importing PHP namespaces
* OS and language server
  * macOS 10.15 - 
  * Package Control
* Minimal reproduction steps
  * create a minimal php package with autoloading
  * create a `src/First` and `src/Second` folders
  * add an App class with the appropriate namespaces, so we have e.g. `Acme\First\App` and `Acme\Second\App`
  * In a different class e.g. `Acme/Start` try writing App in a function. It will show two App instance, `Acme\First\App` and `Acme\Second\App`.
  * Select the second one
  * It will use the first App's additionalTextEdits and adds `use Acme\First\App` when I clearly selected the second (tried using tab or enter, doesnt matter)
* Log
```
{'textEdit': {'range': {'end': {'character': 4, 'line': 6}, 'start': {'character': 2, 'line': 6}}, 'newText': 'App'}, 'additionalTextEdits': [{'range': {'end': {'character': 0, 'line': 4}, 'start': {'character': 22, 'line': 2}}, 'newText': '\n\nuse Acme\\Test\\App;\n\n'}], 'kind': 7, 'detail': 'use Acme\\Test\\App', 'sortText': 'App', 'label': 'App', 'data': 1264852955825781},
{'textEdit': {'range': {'end': {'character': 4, 'line': 6}, 'start': {'character': 2, 'line': 6}}, 'newText': 'App'}, 'additionalTextEdits': [{'range': {'end': {'character': 0, 'line': 4}, 'start': {'character': 22, 'line': 2}}, 'newText': '\n\nuse Acme\\Third\\App;\n\n'}], 'kind': 7, 'detail': 'use Acme\\Third\\App', 'sortText': 'App', 'label': 'App', 'data': 4436429581058069}
```

From the logs, it's clearly seen that the use expression is correct for the different classes. It just that it always uses the first one and there is no way to import the second namespace

predragnikolic:
NEW:
- completion kind symbols in autocomplete (symbols will probably be a subject of change)
Screenshot from 2020-01-30 19-53-15

- support deprecated flag in completion items(don't know any server that implements this, but here is an image of a deprecated completion item)

Screenshot from 2020-01-30 20-10-28

- this PR also fixes #757, because we now know exactly what completion item was chosen

I haven't been able to run any completion unit test, just to tell you. 

There are some things in this PR, that I wish I had done better, but I was not able to(due to my limited knowledge, I will put some comments to mark those things)


plugin/completion.py
```diff
@@ -1,25 +1,85 @@
...
+
+            range = Range.from_lsp(text_edit['range'])
+            edit_region = range_to_region(range, self.view)
+
+            if insert_text_format == InsertTextFormat.Snippet:
+                self.view.run_command("insert_snippet", {"contents": new_text})
+            else:
+                # use current point as the end of the edit region
+                current_point = self.view.sel()[0].begin()
+                edit_range = sublime.Region(edit_region.begin(), current_point)
```
predragnikolic: this is the only part of the PR that diverges from the spec.

I use the text edit, start ragne, 
but I can not use the text edit end range. 

This is because of the following reasons:
```
   | - is the curson

When choosing a command completion item,  1| pref|
                                          2|

                                          # removed
The typed prefix will be remove,          1| 
                                          2|

Because of the removed prefix, we cannot relay anymore on the range end position,
because the end position is outdated and it would select the wrong region.

We could potentionaly substract the prefix length from the end range.
But we cannot relay on that eather,
because sometimes the completion request is sent before the `textDocument/didChange`. 
(`textDocument/didChange` has also a debounce of 500ms);

I would not debounce the completion request for 500ms, because it feels really slow.

The only solution I found now is to use the current cursor postion. 
When choosing a command completion item,  1| pref|
                                          2|

The typed prefix will be remove,          1| 
                                            ^ this will be the end range position
                                          2|

and we will insert the completion item.   1| prefix|
                                          2|

I know this sounds complicated. But it is the only way I managed to make it work.
```

plugin/completion.py
```diff
@@ -1,25 +1,85 @@
...
+                self.view.insert(edit, current_point, completion)
+
+        # import statements, etc. some servers only return these after a resolve.
+        additional_edits = item.get('additionalTextEdits')
+        if additional_edits:
+            self.apply_additional_edits(additional_edits)
+
+        self.do_resolve(item)
+
+    def do_resolve(self, item: dict) -> None:
```
predragnikolic: I am not proud of how the do_resolve function looks

predragnikolic: any suggestion on how to make it better?

predragnikolic:
TODO: 
remove `completion_hint_type` because it is no longer used

predragnikolic:
Also this PR, doesn't trigger autocomplete on `,` on sublime text 4

plugin/completion.py
```diff
@@ -1,25 +1,85 @@
...
+
+        text_edit = item.get('textEdit')
+        if text_edit:
+            new_text = text_edit.get('newText')
+
+            range = Range.from_lsp(text_edit['range'])
+            edit_region = range_to_region(range, self.view)
+
+            if insert_text_format == InsertTextFormat.Snippet:
+                self.view.run_command("insert_snippet", {"contents": new_text})
```
rwols: I think handling of the edit region is missing here. Imagine this situation: The user types `aaa$bbb`.
The server returns one completion item with range (2, 6) and textEdit `foobar($0)`. So we should end up with `aafoobar()`. This code will end up with `aaa$bbbfoobar()`.

My guess:

1.   Remove the region from the buffer in all cases: `self.view.erase(edit_region)`.
2a. If it's a Snippet, then run `insert_snippet {"content": new_text}`.
2b. If it's a PlainText, then run `insert_snippet {"content": snippet_escape(new_text)}`, where `snippet_escape` replaces all `$` signs with `\$`.

In particular, I'd like to know if a completion item like this one would work: https://github.com/sublimelsp/LSP/blob/2d7bed9bc58eb25fe670e472813cc810bf77852b/tests/test_completion.py#L351-L370

Here you can see that we have a single letter `I` in the buffer, and the completion is something completely different.

plugin/completion.py
```diff
@@ -1,25 +1,85 @@
...
+        client = client_from_session(session)
+        if not client:
+            return
+
+        completion_provider = session.get_capability('completionProvider')
+        has_resolve_provider = completion_provider and completion_provider.get('resolveProvider', False)
+        if has_resolve_provider:
+            client.send_request(Request.resolveCompletionItem(item), self.handle_resolve_response)
+
+    def handle_resolve_response(self, response: 'Optional[Dict]') -> None:
```
rwols: You don't have to put quotes around these mypy types anymore. Just write `Optional[Dict]` ! :dancer:

plugin/completion.py
rwols: Also that completion example from metals is supposed to be a snippet I believe: `def foo: Int = ${1:???}\n   def boo: Int = ${2:???}`

plugin/core/completion.py
```diff
@@ -1,76 +1,61 @@
...
+    item_kind = item.get("kind")
+    if item_kind:
+        kind = compleiton_kinds.get(item_kind, sublime.KIND_AMBIGUOUS)
+
+    is_deprecated = item.get("deprecated", False)
+    if is_deprecated:
+        list_kind = list(kind)
+        list_kind[1] = '⚠'
+        list_kind[2] = "⚠ {} - Deprecated".format(list_kind[2])
+        kind = tuple(list_kind)  # type: ignore
```
predragnikolic: mypy complained here that I tried to assign `Tuple[object, ...]` to `Tuple[int, str, str]`.

I tried to use `kind = cast(Tuple[int, str, str], tuple(list_kind))` here but sublime cant import cast from typings it, it results in a run-time error :)

so I put # type: ignore

is this ok?

plugin/completion.py
```diff
@@ -1,35 +1,85 @@
...
+from .core.views import range_to_region
 from .core.typing import Any, List, Dict, Tuple, Optional, Union
 
 
-class CompletionState(object):
-    IDLE = 0
-    REQUESTING = 1
-    APPLYING = 2
-    CANCELLING = 3
+class LspSelectCompletionItemCommand(sublime_plugin.TextCommand):
```
rwols: One way to test these changes thoroughly is to have a unit test that runs this command.

e.g. something like this

```python
import sublime
from unittesting import DeferrableTestCase

class SelectCompletionItemTest(DeferrableTestCase):

    def setUp(self) -> Generator:
        super().setUp()
        self.window = sublime.active_window()
        self.view = self.window.new_file()
        yield lambda: not self.view.is_loading()
        self.view.set_scratch(True)

    def tearDown(self) -> None:
        self.view.close()
        super().tearDown()

    def test_a(self) -> None:
        self.view.run_command("insert", {"characters": "a"})
        item = { ... }
        self.view.run_command("lsp_select_completion", {"item": item})
        actual = self.view.substr(sublime.Region(0, self.view.size()))
        self.assertEqual("b", actual)

File: plugin/completion.py (hunk 3 above)
Comment (rwols) [reply plugin/completion.py]: Also try this situation:

The buffer is `#include <io|>` ( `|` denotes the cursor position)
The completion is `iostream>` with edit range (10, 12), so it should remove `io>` and replace it with `iostream>`.

Comment (predragnikolic) [reply plugin/completion.py]: > 2b. If it's a PlainText, then run insert_snippet {"content": snippet_escape(new_text)}, where snippet_escape replaces all $ signs with \$

Is there any reason why you suggested to run insert_snippet for a PlainText format completion item and escape the $ sign?

Why not just use view.insert() or view.replace()? 
After all I would run `insert_snippet` command only for completion items that are  InsertTextFormat.Snippet.

> I think handling of the edit region is missing here.
Yeah, I will add it :)

Comment (predragnikolic) [reply plugin/completion.py]: > Also try this situation:
> 
> The buffer is `#include <io|>` ( `|` denotes the cursor position)
> The completion is `iostream>` with edit range (10, 12), so it should remove `io>` and replace it with `iostream>`.

I tested this with the JSON language server, it works as expected :)

good

File: plugin/completion.py [hunk 7]
```diff
@@ -24,25 +24,25 @@ def run(self, edit: Any, item: dict) -> None:
 
             range = Range.from_lsp(text_edit['range'])
             edit_region = range_to_region(range, self.view)
+            # subtract the typed prefix form the end of the text edit
+            edit_region = sublime.Region(edit_region.begin(), edit_region.end() - len(CompletionHandler.last_prefix))
```
rwols: Can it happen that `edit_region.begin() > edit_region.end() - len(CompletionHandler.last_prefix)` ?

plugin/completion.py
```diff
@@ -24,25 +24,25 @@ def run(self, edit: Any, item: dict) -> None:
...
-                current_point = self.view.sel()[0].begin()
-                edit_range = sublime.Region(edit_region.begin(), current_point)
-                self.view.replace(edit, edit_range, new_text)
+                self.view.replace(edit, edit_region, new_text)
 
                 # move the cursor to the end of the text edit
                 sel = self.view.sel()
                 sel.clear()
-                sel.add(edit_range.begin() + len(new_text))
+                sel.add(edit_region.begin() + len(new_text))
```
rwols: Do we *really* want to modify the selection? What if there is more than one cursor? Is it possible to make an implementation without touching the selection?

plugin/completion.py
rwols: > Is there any reason why you suggested to run insert_snippet for a PlainText format completion item and escape the $ sign?

You're right, just use self.view.replace(...)

plugin/completion.py
predragnikolic: It can or should I say it could (I removed that logic)
I concluded that is best to NOT touch the TextEdit Range. 
And instead when the user select a command completion item.
the prefix gets removed (I cannot stop that behavior, thats how sublime works...)
so the text edit range is not valid anymore(and I need to calculate the offset, which from my point of view now, seems impossible)

But the solutions is simple.
I just insert back the removed prefix, before continuing with the insert of the TextEdit 
which makes the text edit range valid again. 

If it all makes sense, if not I can try to explain it better.

plugin/completion.py
predragnikolic: Not anymore, I removed that :)

predragnikolic:
I still don't know how to fix the completion test. 
I found a workaround, but it is not good eather.

When calling `self.set_response('textDocument/completion', label_completions)` I concluded that it doesnt work. That is the reason why I created a test_completion property on the CompletionHandler class.

plugin/completion.py
```diff
@@ -166,7 +165,7 @@ def do_request(self, completion_list: sublime.CompletionList, prefix: str, locat
             return
 
         self.manager.documents.purge_changes(self.view)
-        document_position = get_document_position(self.view, locations[0])
+        document_position = text_document_position_params(self.view, locations[0])
         if document_position:
```
rwols: I removed the need for this `if` check. text_document_position_params always returns a dict, wereas get_document_position returned an optional dict.

tests/test_completion.py
```diff
@@ -161,167 +194,148 @@ def await_message(self, msg: str) -> 'Generator':
...
 
-        Metals: label="override def myFunction(): Unit"
+    #     Metals: label="override def myFunction(): Unit"
 
-        """
-        self.view.run_command('append', {'characters': '  def myF'})
-        self.view.run_command('move_to', {'to': 'eol'})
+    #     """
+    #     handler = self.get_view_event_listener("on_query_completions")
+    #     handler.test_completions = edit_before_cursor
```
rwols: Looking forward to seeing this working :)

plugin/core/completion.py
```diff
@@ -1,76 +1,61 @@
...
-            # debug('textEdit from col {}, {} applied at col {}'.format(
-            #     edit_range.start.col, edit_range.end.col, word_col))
-
-            if edit_range.start.col <= word_col:
-                # if edit starts at current word, we can use it.
-                # if edit starts before current word, use the whole thing and we'll fix it up later.
-                return edit_text
-
-    return None
+compleiton_kinds = {
```
rwols: typo: compleiton_kinds -> completion_kinds

plugin/completion.py
```diff
@@ -72,274 +111,81 @@ def initialize(self) -> None:
...
-        # cancel current completion if the previous input is an space
-        prev_char = self.view.substr(self.view.sel()[0].begin() - 1)
-        if self.state == CompletionState.REQUESTING and prev_char.isspace():
-            self.state = CompletionState.CANCELLING
-
-        if self.committing:
-            self.committing = False
-            self.on_completion_inserted()
+        # this is for tests
+        if self.test_completions:
```
predragnikolic: I really would not like to add code like this just to make this testable. 
Can we take a look on why the self.set_response and self.await_message('textDocument/completion') don't work as expected in tests? 🙂

rwols: I took the time to check out the changes locally.

This works for me:

```python
    def test_simple_label(self) -> 'Generator':
        self.set_response("textDocument/completion", label_completions)
        self.type("a")
        count = self.view.change_count()
        yield from self.select_completion()
        yield from self.await_message("textDocument/completion")
        yield from self.await_view_change(count + 2)
        self.assertEquals(self.read_file(), 'asdf')
```

plugin/core/completion.py
rwols: Sounds OK to me for now.

plugin/completion.py
```diff
@@ -1,36 +1,83 @@
...
-    CANCELLING = 3
+class LspSelectCompletionItemCommand(sublime_plugin.TextCommand):
+    def run(self, edit: Any, item: dict) -> None:
+        insert_text_format = item.get("insertTextFormat")
 
+        text_edit = item.get('textEdit')
+        if text_edit:
+            # insert the removed command completion item prefix
+            # so we don't have to calculate the offset for the textEdit range
+            for sel in self.view.sel():
```
rwols: Regarding handling multiple cursors: it is much more complex. If we want to do multiple cursors, we would have to request for each cursor location completions, and then take the *intersection* of all results. The remaining completion items are valid for all cursor locations. What we do here is insufficient. So let's not pretend we support this feature.

plugin/completion.py
```diff
@@ -1,36 +1,83 @@
...
 from .core.views import text_document_position_params
 
 
-class CompletionState(object):
-    IDLE = 0
-    REQUESTING = 1
-    APPLYING = 2
-    CANCELLING = 3
+class LspSelectCompletionItemCommand(sublime_plugin.TextCommand):
+    def run(self, edit: Any, item: dict) -> None:
```
rwols: You can rewrite this signature as `def run(self, edit: sublime.Edit, **item: Any) -> None:`,

and then in plugin/core/completion.py, you can pass the item dict as-is:

```
    return sublime.CompletionItem.command_completion(
        trigger,
        command="lsp_select_completion_item",
        args=item,
        annotation=annotation,
        kind=kind
    )
```
This saves us another key lookup, and it saves us a dict allocation.

plugin/completion.py
```diff
@@ -72,274 +111,81 @@ def initialize(self) -> None:
...
-                    sublime.set_timeout(self.run_auto_complete, 0)
+            self.do_request(completion_list, prefix, locations)
 
-    def on_completion_inserted(self) -> None:
-        # get text inserted from last completion
-        begin = self.last_location
+        return completion_list
 
-        if begin < 0:
+    def do_request(self, completion_list: sublime.CompletionList, prefix: str, locations: List[int]) -> None:
```
rwols: So as I mentioned before, let's focus on getting the basics right and not support multiple cursors for now.

plugin/completion.py
```diff
@@ -1,36 +1,83 @@
...
+            new_text = text_edit.get('newText')
 
+            range = Range.from_lsp(text_edit['range'])
+            edit_region = range_to_region(range, self.view)
 
-class CompletionHelper(sublime_plugin.EventListener):
-    def on_text_command(self, view: sublime.View, command_name: str, args: Optional[Any]) -> None:
-        global last_text_command
-        last_text_command = command_name
+            self.view.erase(edit, edit_region)
```
rwols: Let's try to think of a math solution for this, so that we can replace an `erase`+`insert` operation with a single `replace` operation. If the completion is a snippet, then we are forced to use `erase` I think. But in the case of a plaintext completion, I hope we can do it with a single `replace`!

plugin/completion.py
```diff
@@ -72,274 +111,81 @@ def initialize(self) -> None:
...
-    def do_request(self, prefix: str, locations: List[int]) -> None:
-        self.next_request = None
-        view = self.view
-
-        # don't store client so we can handle restarts
-        client = client_from_session(session_for_view(view, 'completionProvider', locations[0]))
-        if not client:
-            return
-
-        if settings.complete_all_chars or self.is_after_trigger_character(locations[0]):
```
rwols: Where did this functionality go? Are we always requesting completions now?

predragnikolic: Not always, only when on_query_completion is called. (I noted that are all AZ-az chars, and all `auto_complete_triggers` chars)

plugin/completion.py
predragnikolic: We can use replace :)

plugin/completion.py
predragnikolic: Support for multicursor `TextEdit`-s is more complex than I thought :)

tests/test_completion.py
```diff
@@ -161,167 +194,117 @@ def await_message(self, msg: str) -> 'Generator':
...
+            self.read_file(),
+            '-UniqueId')
 
     def test_edit_before_cursor(self) -> 'Generator':
         """
 
         Metals: label="override def myFunction(): Unit"
 
         """
-        self.view.run_command('append', {'characters': '  def myF'})
```
predragnikolic: I must admit that I needed to change some tests so they pass. for example this one... 
The test would fail if we have a completion item with a label  `override def myFunction(): Unit`, 
and I type `myF` sublime wont show the autocomplete popup. 

But if I type `omyF` that will match **o**verride**myF**unction|. But `myF` wont match this completion item.

tests/test_mocks.py
```diff
@@ -24,7 +24,7 @@
             'hoverProvider': True,
             'completionProvider': {
                 'triggerCharacters': ['.'],
-                'resolveProvider': False
+                'resolveProvider': True
```
predragnikolic: Is this OK? 
This was the easiest way to test the `resolveProvider` functionality.

plugin/completion.py
predragnikolic: Thanks for the help. :)
I updated all the tests. I didn't use `await_view_change` because I think that the tests look simpler without it.

BTW, I am so happy to see the completions tests working again :)
I only need to update the test_completion_core.py and to be done with the tests.(and after that I can consider this PR as done, unless you have further change requests)

tests/test_mocks.py
rwols: Sounds good to me.

tests/test_completion.py
rwols: Perhaps @ayoub-benali can check out this branch locally and test the changes for metals?

predragnikolic: I tried it with metals now.
I found the example of @ayoub-benali posted on github
https://github.com/sublimelsp/LSP/pull/602#issuecomment-499453180
and recreated that.
good

I think st4 filters out that completion item(`myF` don't match the label `override def myFunction(): Unit`, but `omyF` does match `override def myFunction(): Unit` )

ayoub-benali: thanks @predragnikolic for testing so quickly. For me ST4 should provide completion for `myF`. Should I create an issue ?

predragnikolic: If you have the time please fill an issue, at least they can tell us if this is intentional behavior or not :)

I just noted on the `.gif` that `additionalTextEdits` gets applied twice. I will fix it

predragnikolic: @ayoub-benali setting this sublime setting `"auto_complete_preserve_order": "none"` default is ("some"), fixes the problem.

good

ayoub-benali: this is nice ! good catch

ayoub-benali: Can LSP set `auto_complete_preserve_order` setting ? or does the user has to do it explicitly ?

predragnikolic: Yes we can with `sublime.load_settings("Preferences.sublime-settings")` and than we can change the desired setting, 
but I pretty sure that it is bad... Imagine that a X plugin messes up with your user preferences without you knowing it. 
But looking at this case, setting it to "none" leads to better user experience... :)

rwols: @predragnikolic `auto_complete_preserve_order`, nice find!

willrowe: Would this still be a bug in ST4? It seems that the `auto_complete_preserve_order` setting should only affect the order of what is displayed, not remove items from being displayed completely.

predragnikolic: I agree the setting only mentions that it reorders items, not remove them. :)

plugin/completion.py
```diff
@@ -47,8 +47,8 @@ def run(self, edit: Any, **item: Any) -> None:
         additional_edits = item.get('additionalTextEdits')
         if additional_edits:
             self.apply_additional_edits(additional_edits)
-
-        self.do_resolve(item)
+        else:
+            self.do_resolve(item)
```
rwols: Please add a test for this.

tests/test_completion.py
rwols: I think in the meantime we can modify the tests by doing `self.view.settings().set("auto_complete_preserve_order", "none")` in the `setUp` and worry about what ST4 should be doing later.

plugin/completion.py
predragnikolic: added 49ccda6 :)

tests/test_completion_core.py
```diff
@@ -32,141 +32,3 @@ def test_dict_response(self):
     def test_incomplete_dict_response(self):
         self.assertEqual(parse_completion_response({'items': [], 'isIncomplete': True}), ([], True))
 
-
-class CompletionFormattingTests(unittest.TestCase):
-
-    def test_only_label_item(self):
```
predragnikolic: There is a test in test_completions for this:
- test_simple_label

tests/test_completion_core.py
```diff
@@ -32,141 +32,3 @@ def test_dict_response(self):
...
-        result = format_completion({"label": "asdf"})
-        self.assertEqual(len(result), 2)
-        self.assertEqual("asdf", result[0])
-        self.assertEqual("asdf", result[1])
-
-    def test_prefers_insert_text(self):
```
predragnikolic: There is a test in test_completions for this:
- test_prefer_insert_text_over_label

tests/test_completion_core.py
```diff
@@ -32,141 +32,3 @@ def test_dict_response(self):
...
-                    # Replace the single character that triggered the completion request.
-                    'end': {'character': 13, 'line': 6},
-                    'start': {'character': 12, 'line': 6}
-                }
-            }
-        }
-        result = format_completion(item)
-        self.assertEqual(result, ('const\t  Keyword', 'const'))
-
-    def test_text_edit_intelephense(self):
```
predragnikolic: I think we can just remove these tests:
- test_text_edit_intelephense
- test_text_edit_clangd
- test_missing_text_edit_but_we_do_have_insert_text_for_pyls 
because they are no longer relevant.

predragnikolic:
@rwols I think I am done.
If there is anything else you would like to change?

rwols:
> If there is anything else you would like to change?

Still missing:
- It looks like the "complete_all_chars" option is still ignored? If "complete_all_chars" is true you want the behavior that the code is now, otherwise *only* do completions when we see a trigger character.
- The code pretends to support multiple cursors, but does it incorrectly. Please remove it.

plugin/completion.py
predragnikolic: Sorry, this code here `for sel in self.view.sel():` 
is used to insert back the removed prefix, 
so the text edit range for the completion item is valid again.
it is not used to to support multicursor text edits.

rwols: Thanks for the explanation :)

predragnikolic:
As for the "complete_all_chars". I would like to deprecate that setting in favor of the default setting that comes with sublime.

```
// Enable auto complete to be triggered automatically when typing.
"auto_complete": true,
```
I was going to implement it and than I stumbled upon this setting and then I realize that they maybe we don't need `complete_all_chars` after all. 

IMO "complete_all_chars" in ST3 gave bad user experience(and that would be the same for st4). 
Because the request would only be send if the previous char is in trigger_chars. What if I typed `document.query|` and triggered autocomplete manually. The request would not be send, because the prev char is not a trigger char... To fix that I would need to find a way to distinguish when the autocomplete was triggered manually to send a request. Which is not that simple. 

Having `"auto_complete": false` will not trigger autocomplete popup to show up, until explicitly triggered. And I thing that that behavior is a nice replacement for people who would set `complete_all_chars: false` in the first place.

@rwols what do you think?

predragnikolic:
And I will probably revert 44a3b2a.
Because it leads to this behavior.

2020-02-19-230022_569x238_scrot

When selecting a completion item, the replaced text will be selected.

One way to fix is is to revert 44a3b2a, (because when calling `view.insert`, the cursor will move to the end)
Second way is to leave it like it is, but go through `sel in view.sel()` and move the cursor at the end. (which requires a couple lines of code more)

I think it is cleaner just to revert?

rwols:
I'm okay with this. It sounds as if you've put a lot of thought into it :) I think we can merge it.
It's still unclear to me why we need to loop over the selections, especially because we don't loop over selections when we're dealing with a snippet.
Can you write a unit-test that *proves* that this looping over the selections is necessary? It doesn't have to be in this PR.

predragnikolic:
First to say why I would like to revert to use erase and insert instead of replace. 

The behavior I am going to describe is not related to InsertTextFormat.Snippet (they work fine). I will be talking to in regard to completion items that have  InsertTextFormat.PlainText and a text edit.

here is the behavior with replace:
replace
- notice that the completion item inserted text got selected, that means if you continue typing you will delete that inserted text
- also note that it only got selected after typing `>` and choosing a completion item.(I think that sublime is inconsistent here)

here is the behavior with erase and insert:
erase_insert
- the completion item inserted text doesn't get selected

predragnikolic:
I would not merge this yet. I added a test 6fed5d3 (which is currently failing, but I have a test for it). :) 

good

The problem is related to the prefix.
Here we have a view:
```
$he|
?>
```
if we trigger the autocomplete 
the prefix is `he`(and not `$he`, if the prefix was `$he` we would not have this problem)


Before we select the command completion the view is:
```
$he|
?>
```
After we select the command completion the view is:
```
|
?>
```
Than we inset back the prefix:
```
he|
?>
```
here is the code for that
https://github.com/sublimelsp/LSP/blob/dae5eba5bc9a1ea8f09125867b7781a4debd68ab/plugin/completion.py#L23-L26

than we apply the text edit, but because the prefix was not `$he`, but `he` the range in the text edit is not correct.

so we end up with this
```
$hello|?>
```

predragnikolic:
Sorry I had to revert 44a3b2a, because the autocomplete should not behave like this: :)
replace

instead it will behave like this (the inserted text wont get a selection)
erase_insert

predragnikolic:
I also fixed the problem with the $ sign being removed, which lead to incorrectly applying text edits. I was thinking of different solutions. But I ended up with the most simple I could think of 1ccb7ab. 

If you have a better suggestion, I am open for discussion. :)

plugin/completion.py
```diff
@@ -147,12 +147,34 @@ def on_query_completions(self, prefix: str, locations: List[int]) -> Optional[su
...
 
         self.do_request(completion_list, prefix, locations)
 
         return completion_list
 
+    def include_special_chars(self, prefix: str, point: int) -> str:
+        """
+        Sanatize the prefix, sublime trims of special characters from the prefix
+        """
+        special_chars = ['$']
```
rwols: I'm guessing the special chars are exactly the `view.settings().get("word_separators")`

rwols: Perhaps you can test this by removing `$` from the word separators and then testing if sublime still trims `$`.

rwols: If `$` is troublesome for PHP, I'm also predicting that Powershell will give us troubles. Prefixes for Powershell are `-`.

predragnikolic: I thought that word_separators are the cause of this. But they are not.
I removed the $ from the list. But I experience the same problem.
"word_separators": "./\\()\"'-:,.;<>~!@#%^&*|+=[]{}`~?",

I don't understand why sublime removes the $ when selecting a command completion item in the first place. Because the $ is by default a word separator. And by my logic it should not be removed when selecting a command completion because it is not part of the word.

rwols: Consider trying out Powershell with PowershellEditorServices.

Try the syntax from the master branch at https://github.com/SublimeText/PowerShell
Then try the syntax from the ongoing rewrite: https://github.com/SublimeText/PowerShell/pull/165

Notably, the PR author decided to remove `-` from the word separators. I'm wondering out loud what that would entail for our logic.

rwols: > I thought that word_separators are the cause of this. But they are not. I removed the $ from the list. But I experience the same problem.

Bummer. I had hoped we would find the solution through word_separators :)

predragnikolic: I will try to setup the PowershellEditorServices. 
What would you like me to test exactly? :)

rwols: > What would you like me to test exactly?

Create a test file test.ps1 and start writing `get-` or `$env:`, completions should appear and e.g. `-` symbols should not disappear, and `$` symbols should also not disappear.

predragnikolic: I have no luck setting up PowershellEditorServices on linux right now :( 
Do you maybe have it setup?

plugin/completion.py
```diff
@@ -1,36 +1,83 @@
 import sublime
 import sublime_plugin
-from .core.completion import parse_completion_response, format_completion
-from .core.configurations import is_supported_syntax
-from .core.documents import position_is_word
-from .core.edit import parse_text_edit
+
+from .core.protocol import Request, Range, InsertTextFormat
```
rchl: I've tried this branch with `LSP-json` as completions have issues in old code.

With new code it's better, I don't get duplicate quotes when completing values in `*.sublime-settings` file. For example in:
```js
	"default_line_ending": "unix",
```
Changing "unix" to something else with help of completions insert correct amount of quotes.
But I found another problem:
1. Select `unix` value and delete it
2. Trigger completions and pick some completion

Problem: The value and quotes are inserted correctly but trailing comma gets removed.

predragnikolic: I am aware of that issue. 🙂 I know the cause but haven't found a solid solution.

deathaxe:
Playing around with this branch and LSP-json, I found the following issues:

Without this branch duplicated quotation marks are added - a know issue.

With this branch merged - trailing commas or closing brackets are removed, which is not ok as well.

Animation


The completion was done with a dump sublime-theme jsonschema

```
{
	"$schema": "http://json-schema.org/draft-07/schema#",
	"$id": "sublime://sublime-theme",
	"title": "Sublime Text Theme",
	"allowComments": true,
	"allowTrailingCommas": true,
	"type": ["array", "object"],
	"definitions": {

		"byte": {
			"type": "integer",
			"minimum": 0,
			"maximum": 255
		},

		"colorValue": {
			"oneOf": [
				{
					"type": "array",
					"items": [
						{ "$ref": "#/definitions/byte" },
						{ "$ref": "#/definitions/byte" },
						{ "$ref": "#/definitions/byte" }
					]
				},
				{
					"type": "array",
					"items": [
						{ "$ref": "#/definitions/byte" },
						{ "$ref": "#/definitions/byte" },
						{ "$ref": "#/definitions/byte" },
						{ "type": "number" }
					]
				},
				{
					"type": "array",
					"items": [
						{
							"type": "string",
							"enum": [
								"background",
								"foreground"
							]
						},
						{ "type": "number" }
					]
				},
				{
					"type": "string"
				},
				{
					"type": "null"
				}
			]
		},
		"commonAttributeValue": {
			"type": "string",
			"enum": [
				"hover"
			],
			"uniqueItems": true
		},
		"luminosityAttrubuteValue": {
			"description": "Although not available on all elements, many have attributes set based on the approximate luminosity of the current color scheme. Most elements have the attributes set based on the global color scheme. Tabs and the tab background, however, have the attributes based on the color scheme specific to the selected view.",
			"type": "string",
			"enum": [
				"file_light",
				"file_medium_light",
				"file_medium_dark",
				"file_dark"
			],
			"uniqueItems": true
		},
		"opacityValue": {
			"anyOf": [
				{
					"type": "number"
				},
				{
					"type": "object",
					"properties": {
						"target": { "type": "number" },
						"speed": { "type": "number" },
						"interpolation": {
							"enum": [
								"linear",
								"smoothstep"
							]
						}
					}
				}
			]
		}
	},
	"properties": {
		"name": {
			"description": "A descriptive name.",
			"type": "string",
			"uniqueItems": true
		},
		"author": {
			"description": "The author of the syntax definition.",
			"type": "string",
			"uniqueItems": true
		},
		"extends": {
			"description": "The Sublime Syntax which is extended by this file.",
			"type": "string",
			"uniqueItems": true
		},
		"variables": {
			"description": "Define variables to re-use in the rules.",
			"type": "object",
			"additionalProperties": {
				"anyOf": [
					{ "type": "boolean" },
					{ "$ref": "#/definitions/colorValue" },
					{ "$ref": "#/definitions/opacityValue" }
				]
			}
		},
		"rules": {
			"description": "The list of rules, which describe the visual style of the control elements.",
			"type": "array",
			"items": {
				"anyOf": [
					{
						"type": "object",
						"properties": {
							"class": {
								"description": "The MacOS title bar",
								"const": "title_bar"
							},
							"attributes": {
								"type": "array",
								"items": [
									{ "$ref": "#/definitions/commonAttributeValue" },
									{ "$ref": "#/definitions/luminosityAttrubuteValue" }
								],
								"minItems": 1,
								"uniqueItems": true
							},
							"parents": {
								"type": "array",
								"$ref": "#/properties/rules/items"
							},
							"bg": {
								"$ref": "#/definitions/colorValue"
							},
							"fg": {
								"$ref": "#/definitions/colorValue"
							},
							"background": {
								"$ref": "#/definitions/colorValue"
							},
							"foreground": {
								"$ref": "#/definitions/colorValue"
							}

						},
						"additionalItems": false,
						"uniqueItems": true
					},

					{
						"type": "object",
						"properties": {
							"class": {
								"description": "A text label control",
								"const": "label_control"
							},
							"attributes": {
								"type": "array",
								"items": [
									{ "$ref": "#/definitions/commonAttributeValue" },
									{ "$ref": "#/definitions/luminosityAttrubuteValue" }
								],
								"minItems": 1,
								"uniqueItems": true
							},
							"bg": {
								"$ref": "#/definitions/colorValue"
							},
							"fg": {
								"$ref": "#/definitions/colorValue"
							},
							"background": {
								"$ref": "#/definitions/colorValue"
							},
							"foreground": {
								"$ref": "#/definitions/colorValue"
							}

						},
						"additionalItems": false,
						"uniqueItems": true
					},

					{
						"type": "object",
						"properties": {
							"class": {
								"description": "Unknown rule type",
								"type": "string"
							}
						},
						"additionalItems": false,
						"uniqueItems": true
					}
				],
				"additionalItems": false
			}
		}
	}
}
```

predragnikolic:
I tried my last attempt at finding a nice fix for the issue @deathaxe and @rchl are having. 
Because I am out of ideas, I opened up a feature request at ST
 and hope I will soon get a reply back.

predragnikolic:
> would optimization like only saving from the start of the line to the cursor position work also?

Yes, I thought about it, but I didn't wanted to complicate things. 
Should I do it?

predragnikolic:
I will do it, and than we can compare. (although I am always for things that are easier to understand)

predragnikolic:
Also commit e3f580c brings better multicursor behavior for text edits.
before:
output

 
after:
after

rchl:
> I will do it, and than we can compare. (although I am always for things that are easier to understand)

I don't want you to spend time on it if it complicates thing and there is no clear benefit. I think super long lines are unlikely when coding.

predragnikolic:
Ok, then I think this PR can be consider ready for a final review. :-)

plugin/completion.py
```diff
@@ -200,9 +178,11 @@ def handle_response(self, response: Optional[Union[dict, List]],
 
         if response_incomplete:
             flags |= sublime.DYNAMIC_COMPLETIONS
-
         completion_list.set_completions(items, flags)
-        CompletionHandler.last_prefix = prefix
+
+        for point in locations:
+            # save the line to restore later (only when selecting a completion item with a TextEdit)
+            RestoreLines.save_line(point, self.view)
```
jskinner: You'd want to do this in `on_query_completions`, as (assuming `client.send_request` is non-blocking) the user can type before the response comes back

predragnikolic: For some reason I thought on_query_completions was too early. 
But you are right.

predragnikolic: Moved it
Thanks :)

deathaxe:

Animation

Settings are

```
	"auto_complete": true,
```

rchl:
I can reproduce issues related to completions being returned with some delay.
I've reproduced those with artificially delayed `textDocument/completion` response by 2 seconds on both `LSP-json` and `LSP-css` servers.

<hr>
First issue.
Assume that one of the completion items will be `"url"`.

<details>
  <summary>Here is an example completion result with such item</summary>
  
  ```js
{
  'isIncomplete' : False, 'items' : [
    {
      'kind' : 10,
      'insertTextFormat' : 2,
      'documentation' : '',
      'filterText' : '"url"',
      'label' : 'url',
      'textEdit' : {
        'range' : {
          'start' : {'line' : 30, 'character' : 4},
          'end' : {'line' : 31, 'character' : 0}
        },
        'newText' : '"url": "$1",'
      },
      'insertText' : '"url": "$1",'
    },
    {
      'kind' : 10,
      'insertTextFormat' : 2,
      'documentation' : '',
      'filterText' : '"email"',
      'label' : 'email',
      'textEdit' : {
        'range' : {
          'start' : {'line' : 30, 'character' : 4},
          'end' : {'line' : 31, 'character' : 0}
        },
        'newText' : '"email": "$1",'
      },
      'insertText' : '"email": "$1",'
    }
  ]
}
  ```
</details>

1.
```json
{
  "|"
}
```
2.
```json
{
  "u|"
}
```
3. Wait for completion popup to appear
4. Type `r` and complete before new completions return

Expected:
```json
{
  "url": "|"
}
```

Actual:
```json
{
  "url": "","
}
```

<hr>

Second issue. In css

1. With a document like so:
```css
.foo {
  margin-inl|
}
```
2. Type `i` and immediately backspace to delete it
```css
.foo {
  margin-inli|
}
```
```css
.foo {
  margin-inl|
}
```
3. Wait for completions popup to show up
4. Trigger `margin-inline-block` completion

Expected:
```css
{
    margin-inline-start: ;
}
```

Actual (this can get widely different):
```css
{
     margin-inlimargin-inline-start: ;
}
```

predragnikolic:
@deathaxe 
The `"` is a trigger character in JSON language server. Maybe that is the reason why it is constantly showing the autocomplete popup? I will have to check.

The second strange thing I noticed. 
Although the `"` is a trigger character.
When I type `"`
it automatically expands to `"|"`
but I get no autocomplete popup.

rchl:
> The second strange thing I noticed.
> Although the `"` is a trigger character.

Yes, annoying. Could be related to default auto-pairing keybinding that triggers snippet:
```
	{ "keys": ["\""], "command": "insert_snippet", "args": {"contents": "\"$0\""}, "context":
```

rwols:
We learned from discussions on the Discord channel that ST4 will most likely get additional API features to fix this problem. An example usage snippet is posted here: https://gist.github.com/jskinner/4dfa6d86f848e1e1e583883507369b67

Let's wait until that is released.

Also, like I mentioned before let's not pretend we support multiple cursors. Please just bail out early in `on_query_completions` by checking if `len(locations) == 1`.

deathaxe:
> Please just bail out early in on_query_completions by checking if len(locations) == 1.

:-(

Often do multiple edits - maybe in themes or color-schemes for instance. Completions are computed using the first caret position and committed to all. Not supporting something like that was another downside for using LSP-json over PackageDev for theming.

rwols:
It happens to be a coincidence that your completions are valid for all your cursor positions. Your knowledge of the schema allows you to utilize multiple cursors in the correct positions.

I suppose we can allow this behavior of requesting only completions for the first caret, and then applying them to all carets. But I predict some users will create bug issues about it.

<hr>
Just so that everyone is on the same page, the problem with querying for completions for the first caret position and then applying them to all caret positions is this: let's say I have these three python classes:

```python
from abc import ABCMeta, abstractmethod

class A(metaclass=ABCMeta):
    @abstractmethod
    def f(self) -> None:
        pass

class B(A):
    def f(self) -> None:
        pass

    def g(self) -> None:
        pass

class C(A):
    def f(self) -> None:
        pass

    def h(self) -> None:
        pass


b = B()
c = C()

# caret positions (|) just after the dot.
# The returned completions will contain `f` and `g`.
# But, the only valid method call is `f`.
b.|
c.|
```

So, for `b.` we would get completions {f, g}. For `c.` we would get completions {f, h}. Their intersection is {f}, which is the only valid completion for both caret positions. The current diff will also allow g as a completion.

deathaxe:
> But I predict some users will create bug issues about it.

Possible, but 

1. applying the same competion to all carets is, what ST does when using sublime-completions as well and thus should be expected behavior.
2. all?! operations including basic writing apply the same change to all caret positions.
3. This behavior could be described in documentation to make it clear for everyone it to be designed behavior.

Note: I used to have the same oppintion about just aborting completions if multiple carets are active, but found it very useful to have after a while.

predragnikolic:
@rwols  here is how VS code handles your case:

It behaves differently depending where the cursor is first placed.

If the cursor is first on `b` than on `c`:
`f` and `g` are suggested.

If the cursor is first on `c`, than on `b`, 
because we have a syntax error, the server wont be able to compute them.
output

But if I write it differently so there is no syntax error.
`f` and `h` are suggested.
Screenshot from 2020-03-26 17-39-06

predragnikolic:
@deathaxe I fixed the issue where the autocomplete was triggering "forever".
@rchl I also fixed the issue where you added a delay :)

All in all I am still excited to see how the new API will affect this PR :)

@rwols As for as multi cursors. 
The only problem with multi cursors can be with completions items that have a TextEdit.
While I agree that e3f580c32 doesn't handle all the cases for text edits.
it should work for 90% of the cases. It is still better than not handling multi-cursors at all. :)

We can always remove the code introduced with e3f580c32 and replaced it with a better one.

FichteFoll:
If the completion is just inserted with `commit_completion` and the last character happens to be a trigger char, I believe it makes sense in some occasions to open another popup and disabling this behavior globally (which it seemed like from the diff) is not desireable. This probably needs some more investigation and might lead to a ST code ticket depending on the results.

rwols:
@predragnikolic @deathaxe OK, you guys convinced me doing a completion for the first cursor only has benefits.

Agree with @FichteFoll that the commit_completion behavior feels more like an ST4 issue.

predragnikolic:
>Disabling this behavior globally (which it seemed like from the diff) would not be desireable

Now when I think about it more, I agree that it wouldn't be desirable. Because I disabled it globally. I will revert the commit.

jskinner:
This is probably off topic for this ticket, so I'm more than happy to talk further about this elsewhere, but having `"` as a trigger character without a scope limiting it to `punctuation.definition.string.begin` is bananas. Personally, I'd say `"` must not trigger AC in comments, in strings (e.g., when the user types an escaped quote), or for a quote at the end of a string.

In general, I think it's best to either:

A) Ignore server provided trigger characters entirely. ST4 in general does a good job at determining when to show AC automatically: it knows to show AC when the user types "foo.", but not when the user types in "1.0", or in the middle of a comment when the user is typing "this is the end of the sentence.". In a more general sense, ST4 knows when a completion is likely to be inserted (by examining similar text to what's been entered on the current line), and will show AC automatically then. 
B) Add scope contexts to each trigger character to make them sane. For example, if `"` is a trigger character, implicitly add the context `punctuation.definition.string.begin`

In terms of showing AC if the inserted completion contains a trigger character, that in general is desirable IMO: if a completion is "foo.", then most users would expect AC to be shown immediately afterwards.

Also, there is a bug in in 4068 that causes AC to automatically show when it shouldn't, but I don't think it's related to the items in this thread. It'll be fixed in 4069.

plugin/core/restore_lines.py
```diff
@@ -7,15 +7,16 @@ def __init__(self):
         self.saved_lines = []  # type: List[dict]
```
predragnikolic: Feel free to correct me, I don't know if I used the new API correctly :)

Here is my feedback:
I want to highlight the differences between the new api, and the code before it:

The code before it did this:
1. It copied the `row`, `text`, `cursor` position of each line, right before sending the request.
2. On selecting a completion item, it inserted back the text on each row. 
END RESULT: It brought the lines in a state like they were when the completions were requested.

The code in this commit does this.
1. It copies the `change_id`, `change_region`(the region passed to the `transform_region_from`),  `text` and `cursor` position for each line, right before sending the request.
2. On selecting a completion item, the transform_region_from returns a `transform_region`.  `transform_region` is used for removing and insert back the previous text for each line. 
```
 view.erase(edit, transform_region)
 view.insert(edit, transform_region.begin(), saved_line['text'])

// I could probably use view.replace instead of erase and insert
```
END RESULT: It brings the lines in a state like they were when the completions were requested.

You may have noticed that I did this a bit differently, than what John did in his gist.
https://gist.github.com/jskinner/4dfa6d86f848e1e1e583883507369b67#file-async_edits-py-L238-L244
```
    def fill_completions(self, edits):
        # The server has sent us a series of edits it wants to be made when
        # the completion command is run, however the user may have edited the
        # buffer since the request want sent. Adjust the response from the
        # server to account for these edits
        for c in self.changes_since_sent:
            edits = [(transform_region(r, c), text) for r, text in edits]
```
If I done the same thing as John did in his gist:
I would have to adjust the response from the server to account for all view edits that occurred,
and apply all TextChange-s to all completion items.
But that to me looks like a lot of work.

For me it looks easier to adjust the view state to the state it was when the request is send, 
than to adjust the response from the server to account all the possible edits that may have had occurred. 

Either way, I really want to hear the feedback from you all
And don't take me wrong with what I just said :)

FichteFoll:
> This is probably off topic for this ticket, so I'm more than happy to talk further about this elsewhere, but having `"` as a trigger character without a scope limiting it to `punctuation.definition.string.begin` is bananas. Personally, I'd say `"` must not trigger AC in comments, in strings (e.g., when the user types an escaped quote), or for a quote at the end of a string.

@jskinner, just for confirmation, the trigger scope is matched against the inserted char *after* it has been inserted? Because that would be different from the `auto_complete_selector` setting, which makes AC appear if the line break character at the end of the comment is not included in the `comment` scope.

plugin/core/restore_lines.py
```diff
@@ -7,15 +7,16 @@ def __init__(self):
...
-        # Clear previously saved lines
-        self.clear()
+        change_id = view.change_id()
 
         for point in locations:
-            text = view.substr(view.line(point))
-            row, _col = view.rowcol(point)
+            line = view.line(point)
+            change_region = (line.begin(), line.end())
+            text = view.substr(line)
```
deathaxe: Anything which prevents the `line` to be stored directly? Doing so would avoid converting `region->tuple->region`.

FichteFoll: Regions can't be serialized for transmission in `run_command`, which is required for the commit completion command.

rwols:
When I supplement the completion by typing more characters, the completion widget goes away.

Peek 2020-03-28 13-19
<hr>
I don't get a completion widget at all sometimes. In the gif below I expect to at least see a completion for the "time" variable:

Peek 2020-03-28 13-31

I'm guessing this is due to incomplete completions. I'm not sure what the dynamic is here.

plugin/completion.py
```diff
@@ -72,274 +133,79 @@ def initialize(self) -> None:
...
-                prefix, locations = self.next_request
-                self.do_request(prefix, locations)
-        else:
-            debug('Got unexpected response while in state {}'.format(self.state))
+        if response_incomplete:
+            flags |= sublime.DYNAMIC_COMPLETIONS
+        completion_list.set_completions(items, flags)
 
     def handle_error(self, error: dict) -> None:
         sublime.status_message('Completion error: ' + str(error.get('message')))
```
rwols: You should resolve the promise object during an error too. Return zero completions in this case.

plugin/core/restore_lines.py
deathaxe: Guess the python 3.3 version of `sublime.Region` does not support iteration. This will be available as soon as we can move to python 3.8

plugin/core/completion.py
```diff
@@ -1,76 +1,61 @@
...
+    1: (sublime.KIND_ID_MARKUP, "Ξ", "Text"),
+    2: (sublime.KIND_ID_FUNCTION, "λ", "Method"),
+    3: (sublime.KIND_ID_FUNCTION, "λ", "Function"),
+    4: (sublime.KIND_ID_FUNCTION, "c", "Constructor"),
+    5: (sublime.KIND_ID_VARIABLE, "f", "Field"),
+    6: (sublime.KIND_ID_VARIABLE, "v", "Variable"),
+    7: (sublime.KIND_ID_TYPE, "c", "Class"),
+    8: (sublime.KIND_ID_TYPE, "i", "Interface"),
+    9: (sublime.KIND_ID_NAMESPACE, "◪", "Module"),
```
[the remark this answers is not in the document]
predragnikolic: The symbol looks nice ❒,
but I would say that it doesn't fill well with other symbols. (it looks too thin, and it looks like it is not centered)

Screenshot from 2020-03-28 20-02-28

plugin/core/restore_lines.py
predragnikolic: I didn't know you could do that `tuple(region)` with a Region. Cool :)

predragnikolic: 'Region' object is not iterable. 
```
    saved_line['change_region'] = tuple(saved_line['change_region'])
TypeError: 'Region' object is not iterable
```
Looks like the LSP plugin is still using python 3.3.

plugin/completion.py
predragnikolic: Done

jskinner:
@FichteFoll auto_complete_triggers is tested against the character just inserted, and auto_complete_selector is tested at the position before the inserted character (a change from ST3), but after it has been inserted

plugin/core/completion.py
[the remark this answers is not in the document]
AmjadHD: emm, i thought it would be more suitable as it looked like a package, but you're right.

rchl:
Noticed a bug with latest changes:

When, in Python, I'm typing `False` and get an (only) suggestion with it, the suggestion popup doesn't disappear once I've typed whole word. It did before those changes.

Screenshot 2020-03-28 at 21 49 09

predragnikolic:
> It did before those changes.

@rchl are you sure?
I just installed build 4068 and checkout commit(the one before adding the new API) b5c0de6ce9710bbcb21ad992bec7a2e682a3de1f

The autocomplete doesn't disapear there too when I finish typing `False`

rchl:
Sorry, I didn't mean to say the latest changes specifically, just that I've noticed it only with the latest changes included. :)

I've compared to master branch.

deathaxe:
See the bug, too, using the LSP-json completions.

Animation

With regards to @jskinner's statements the triggers and selectors of the view are set to:

```
>>> view.settings().get("auto_complete_selector")
'- string - comment - constant | meta.main-key | meta.rule-key | meta.interpolation-key | meta.animation-key | meta.attributes-sequence - comment | meta.class-selector | meta.function-call.var meta.group'
>>> view.settings().get("auto_complete_character")
Unable to open /C/Apps/Sublime Text/Data/Packages/Default/Preferences.sublime-settings
[{'characters': '<', 'selector': 'text.html'}, {'selector': 'punctuation.accessor'}, {'characters': '":', 'selector': 'source.json'}]
```

The `string` scope is explicitly excluded from completitions. The trigger `":` is defined.

It looks like ST only uses the first character of the trigger, thus retriggering AC after committing a value which ends with `"`. Same happens after completing the `"class"` or just removing and retyping the final `"`. Both the json-key and json-value are `string.quoted.double`.

Not sure whether the trigger is intended to be of one character only by design or whether it is a bug.

predragnikolic:
@rwols https://github.com/sublimelsp/LSP/pull/866#issuecomment-605441312
The LSP plugin doesn't trigger or hide the autocomplete programmatically, anymore.
It just registers the autocomplete triggers and then let sublime decide whether to show the AC or not.
I could try to reproduce it with a language server that returns incomplete completions items (LSP-intelephense supports that). To see if I can reproduce it.

I want to say that this PR is getting quiet large :)

@rwols @rchl If you guys are happy with the changes till now,
can we merge this and create separate issues that should be addressed (like for the autocomplete triggers "forever" and other things) :)

rwols:
Does tests/test_completion.py not crash for you? It segfaults SublimeText for me.

rwols:
I'm okay with merging to st4000-exploration branch provided that the tests pass.

Outstanding issues:
- commit_completion infinite loop (https://github.com/sublimelsp/LSP/pull/866#issuecomment-602243797)
- When, in Python, I'm typing False and getting an (only) suggestion with it, the suggestion popup doesn't disappear once I've typed whole word. It did before those changes. (https://github.com/sublimelsp/LSP/pull/866#issuecomment-605518024) (maybe fixed?)
- Review auto_complete_selector and auto_complete_triggers (https://github.com/sublimelsp/LSP/pull/866#issuecomment-605629210)
- I'm not convinced the restore_lines.py file is needed now that we have `view.transform_from_region`. Think about how to use that method.
- When I supplement the completion by typing more characters, the completion widget goes away. (https://github.com/sublimelsp/LSP/pull/866#issuecomment-605441312)
- I don't get a completion widget at all sometimes. In the gif below I expect to at least see a completion for the "time" variable (https://github.com/sublimelsp/LSP/pull/866#issuecomment-605441312)

predragnikolic:
No, they run fine:
```
Ran 19 tests in 5.981s

OK
```

rwols:
Strange, ST4 segfaults here when running `UnitTesting: Test Current File` :confused:

rwols:
OK, when I run *all* tests with `UnitTesting: Test Current Package` then ST4 doesn't segfault and all tests pass, so I'm okay with merging this :)

rchl:
I won't claim that I've reviewed it but at least with my testing, I could not find more issues (than the one still remaining)

predragnikolic:
Thanks for the response. 
For me, it looks easier to split each issue,
and discuss each separately, 
and make smaller PR that will fix those. 
Than to do it all in this PR here :)

predragnikolic:
@rwols I created issues for most of the things you wrote, but skipped these ones because I don't have enough information on what exactly needs to be done:
* Review auto_complete_selector and auto_complete_triggers
* I'm not convinced the restore_lines.py file is needed now that we have view.transform_from_region. Think about how to use that method.

@deathaxe if you don't have nothing against, 
can you continue the topic you started about triggers and selectors in #933 :)

PR #876, opened 2020-02-12
Fix accumulation in reference quick panel.

Anaminus:
When the `show_references_in_quick_panel` setting is true, references are shown in the quick panel. In `LspSymbolReferencesCommand`, references are stored in the `reflist` variable, which are then passed to `window.show_quick_panel` to display the results. However, at no point is `reflist` emptied. So, as more references are displayed, previous results accumulate in the panel, which gets very noisy.

This PR simply clears `reflist` before it is repopulated.


rwols:
Thanks for your contribution!

PR #882, opened 2020-02-19
core: Send clientInfo in initialize request

radeksimko:
The LSP doesn't seem to be defining any guidelines on what goes into each field, but it seems that most clients just put the editor's name and editor's version.

I can imagine some useful additional information such as OS/arch and version of the plugin here, but I'm being intentionally conservative for now and I feel this could this deliver some value to Language Servers that wish to collect some usage data and understand their user base.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -14,6 +14,9 @@ def get_initialize_params(workspace_folders: List[WorkspaceFolder], config: Clie
     first_folder = workspace_folders[0] if workspace_folders else None
     initializeParams = {
         "processId": os.getpid(),
+        "clientInfo": {
+            "name": "Sublime Text LSP",
+        },
         "rootUri": first_folder.uri() if first_folder else None,
         "rootPath": first_folder.path if first_folder else None,
         "workspaceFolders": [folder.to_lsp() for folder in workspace_folders] if workspace_folders else None,


plugin/core/sessions.py
```diff
@@ -9,13 +9,18 @@
...
 import threading
 
 
 def get_initialize_params(workspace_folders: List[WorkspaceFolder], designated_folder: Optional[WorkspaceFolder],
                           config: ClientConfig) -> dict:
     initializeParams = {
         "processId": os.getpid(),
+        "clientInfo": {
+            "name": "Sublime Text",
+            "version": sublime.version(),
```
predragnikolic: Are you sure we want to put sublime version here? :)
By looking at the `coc` lsp plugin for vim, I think we should put the current version of LSP plugin.
https://github.com/neoclide/coc.nvim/blob/9bfb04d96793aceb0bd6a57df95229596babaa6c/src/attach.ts#L83-L84

radeksimko: As I mentioned above there (sadly) aren't any guidelines on the format of these fields.
Arguably having both editor & plugin names and versions would be useful.

I don't have a strong opinion on the format, I just feel that any identification is better than nothing 😄 

`emacs-lsp` plugin has emacs version there: https://github.com/emacs-lsp/lsp-mode/blob/1b5963dc04c6aa2f29c508ed3863f99b932974ac/lsp-mode.el#L6191-L6192

rwols: I agree with @predragnikolic. It's probably better to have the name be "Sublime Text LSP Plugin" and the version be the version of this plugin. We could hardcode the version number but I'm not a fan of that as those strings always get out-of-date fast (see: tests/server.py version being out of date). We should not claim the string "Sublime Text" in case SublimeHQ ever decides to create a native LSP mechanism.

So these ideas require us to answer the question: how do we maintain a global version number in the plugin code? Are there Python best-practises for this? Should we derive the version number from the files in messages/*.txt or from a git tag?

AmjadHD: You can ask the sublime folks how they manage the version ?

radeksimko: The pattern I have seen in the wild is version being encoded/bumped at release time, e.g.
https://github.com/emmetio/sublime-text-plugin/commit/a77fd929fc99740ca2c682480bb06b41c278e55f
(which I suppose uses [`bumpversion`](https://github.com/peritus/bumpversion))

https://github.com/DisposaBoy/GoSublime/pull/960/files#diff-25d17236be18aa627b2a33c84719698f

I don't know how this extension gets released currently, but I suspect you should be able to incorporate similar mechanism which bumps the version and make a script/makefile a single point of entry for releases to make sure that it's not forgotten.

rwols:
@radeksimko, can you change the `name` to `Sublime Text LSP Plugin`, and leave the `version` out? The version is optional, so let's worry about that in another PR.

radeksimko:
@rwols PR updated, PTAL

predragnikolic:
Thank you :)

PR #897, opened 2020-03-02
Support v2 of LanguageHandler that gets Session in on_initialized

rchl:
Added possibility to opt-in to LanguageHandler v2 which receives
`Session` instance in `on_initialized` handler instead of `Client`.
This allows to get access to WorkspaceFolder's and provide
`workspaceFolder` property in response to workspace/configuration
request.

Related: https://github.com/sublimelsp/LSP-eslint/pull/6

### DIFF
--- a/plugin/core/handlers.py
+++ b/plugin/core/handlers.py
@@ -4,6 +4,7 @@ import abc
 
 
 class LanguageHandler(metaclass=abc.ABCMeta):
+    api_version = 1
     on_start = None  # type: Optional[Callable]
     on_initialized = None  # type: Optional[Callable]
 
--- a/plugin/core/registry.py
+++ b/plugin/core/registry.py
@@ -14,6 +14,7 @@ from .typing import Optional, Callable, Dict, Any, Iterable
 
 client_start_listeners = {}  # type: Dict[str, Callable]
 client_initialization_listeners = {}  # type: Dict[str, Callable]
+client_initialization_listeners_v2 = {}  # type: Dict[str, Callable]
 
 
 class LSPViewEventListener(sublime_plugin.ViewEventListener):
@@ -49,9 +50,12 @@ class LanguageHandlerDispatcher(object):
         else:
             return True
 
-    def on_initialized(self, config_name: str, window: WindowLike, client: Client) -> None:
-        if config_name in client_initialization_listeners:
-            client_initialization_listeners[config_name](client)
+    def on_initialized(self, session: Session, window: WindowLike) -> None:
+        config_name = session.config.name
+        if config_name in client_initialization_listeners_v2:
+            client_initialization_listeners_v2[config_name](session, window)
+        elif config_name in client_initialization_listeners:
+            client_initialization_listeners[config_name](session.client)
 
 
 def load_handlers() -> None:
@@ -66,7 +70,10 @@ def register_language_handler(handler: LanguageHandler) -> None:
     if handler.on_start:
         client_start_listeners[handler.name] = handler.on_start
     if handler.on_initialized:
-        client_initialization_listeners[handler.name] = handler.on_initialized
+        if handler.api_version >= 2:
+            client_initialization_listeners_v2[handler.name] = handler.on_initialized
+        else:
+            client_initialization_listeners[handler.name] = handler.on_initialized
 
 
 def client_from_session(session: Optional[Session]) -> Optional[Client]:
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -200,10 +200,15 @@ class Session(object):
             if not self._workspace_folders or self._unsafe_supports_workspace_folders():
                 return True
         # We're in a window with folders, and we're a single-folder session.
+        return self.workspace_folder_from_path(file_path) is not None
+
+    def workspace_folder_from_path(self, file_path: str) -> Optional[WorkspaceFolder]:
+        candidate = None  # type: Optional[WorkspaceFolder]
         for folder in self._workspace_folders:
             if file_path.startswith(folder.path):
-                return True
-        return False
+                if candidate is None or len(folder.path) > len(candidate.path):
+                    candidate = folder
+        return candidate
 
     def update_folders(self, folders: List[WorkspaceFolder]) -> None:
         with self.acquire_timeout():
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -35,7 +35,7 @@ class LanguageHandlerListener(Protocol):
     def on_start(self, config_name: str, window: WindowLike) -> bool:
         ...
 
-    def on_initialized(self, config_name: str, window: WindowLike, client: Client) -> None:
+    def on_initialized(self, session: Session, window: WindowLike) -> None:
         ...
 
 
@@ -572,7 +572,7 @@ class WindowManager(object):
             "textDocument/publishDiagnostics",
             lambda params: self.diagnostics.receive(session.config.name, params))
 
-        self._handlers.on_initialized(session.config.name, self._window, session.client)
+        self._handlers.on_initialized(session, self._window)
 
         session.client.send_notification(Notification.initialized())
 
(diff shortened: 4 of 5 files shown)


predragnikolic:
This is a breaking change, but not a necessary one :)

I haven't look much at the code, but i noted that there is a session has a client?
```
class Session(object):
    def __init__(self,...client,...) -> None:
        ...
        self.client = client

```
That means that a client belongs to a session. 

But can we have a one-to-one relation between them.

Why not also add a session property to the client?
So we know what session belongs to a client:

```
class Client(object):
    def __init__(self, transport: Transport, settings: Settings) -> None:
        # can't pass `Session` in the constructror
        # because a Client is first created, than a Session 
        # the session can be added/retrieved with getter/setter methods
        ...
        self._session = None

    def set_session(self, session):
        self._session = session

    def get_session(self):
        return self._session
```

than in `sessions.py` we just find a nice place to attach the session to a client.
```
    def with_client(client: Client) -> Session:
        session = Session(
            config=config,
            workspace_folders=workspace_folders,
            designated_folder=designated_folder,
            client=client,
            on_pre_initialize=on_pre_initialize,
            on_post_initialize=on_post_initialize,
            on_post_exit=on_post_exit)

        client.set_session(session)

        return session
```

This way we avoid making a breaking change.

rchl:
I think it's generally a bad idea to have a circular dependency like that (session owns client and client owns session). With a language that uses garbage collection, it would basically mean that those could never be garbage collected (I believe, I'm not a python expert).

But to clarify, this is not a breaking change for existing clients. `Session` is only passed to `on_initialized` if LanguageHandler overrides `api_version` to `2`, otherwise the same arguments as before are passed and nothing breaks.

predragnikolic:
I haven't thought about circular dependency in classes and how it affects garbage collection :). 
You gave me an interesting topic to google tomorrow.

rwols:
Perhaps we can use the `weakref` module? This way you can weakly reference a session from a client, without a circular reference.

By the way, the client object already has many strong references to a Session object and to a WindowManager object. Namely, the `Client.on_notification` and `Client.on_request` methods store a strong reference to a callable, and those callables probably have a `self` captured in their closure.

rchl:
But do you think it's a good design to have those reference each other or are you just choosing a lesser evil? :)

It's much easier to do what you propose but it feels like a smelly thing to do to me. :)

rchl:
BTW. There is also no access to `Window` from on_initialized` currently and I don't think there is a way to get it from client or session without some hacking. I think it might be needed at some point. For example when implementing progress indicators.

predragnikolic:
> For example when implementing progress indicators.

You could use a window, but I am pretty certain that you will soon hit its limitation. Window has a 'status_message(string)' on it. But the limitation is that it disappears after a current amount of time, and you cannot make it stop when you want to 🙂. 

Another option is to use a view.set_status(key, value). This way you are in control when to show/hide it.

tomv564:
There is a standard api for progress which this LSP package could implement, instead of each language adding having their own flavour. I did a quick hack for it at https://github.com/sublimelsp/LSP/tree/show-progress

rchl:
Yes, you are right. I was thinking about that one but thought it's sever-specific which it isn't.

tomv564:
Would prefer additive api changes instead of going to a "v2" already now.
Or feature detection.
But in the past I've limited exposing large chunks of LSP internals because it encourages ugly hacks - most of Session is not language handler business and I think the eslint server has no compelling reason to need it.

rchl:
- I only need access to WorkspaceFolder(s) for eslint's sake.
- Adding new arguments to `LanguageHandler.on_initialized` of course doesn't work as that wouldn't be backward-compatible.
- Adding a new class method to `LanguageHandler` is an option but I'm not sure what it would be called, as `on_initialized` seems like the appropriate one already.

rchl:
Since Python is generally quite flexible, maybe there is a way to pass extra argument depending on the signature of the called method? Maybe try/catch would work?
Or do you have some idea that would be least invasive and specific?

tomv564:
Let's not use try-catch for specific reasons we can detect beforehand.

In python, we can add the arguments to `on_start` and `on_initialized` in a backwards-compatible way by:
* providing defaults for new arguments (probably `new_arg=None` for our cases)
* adding keyword arguments (`**on_start_args`). 

Keyword arguments have a benefit over optional arguments that the signature of the APIs stays short.

Note that I also wanted to change the return value of the `on_start` function - this change must be caught in some type-checking logic on the LSP side.

It also is a good idea to wrap invocations from the dispatcher in a try-catch (for anything) so broken LanguageHandlers can't break LSP. But if that's not done now, it can be added in another PR.

Perhaps some of these ideas need to be moved to the issue #972.

I vote we do the work for `on_start` in #926 in parallel but make the decisions here first.

rchl:
> * providing defaults for new arguments (probably `new_arg=None` for our cases)

I believe that's something I've investigated and it wasn't viable because even though `LanguageHandler` base class can be changed to have have `on_start` method (for example) with default argument, the existing implementations of `LanguageHandler` that are out there, don't specify that optional argument so calling them with optional arguments will crash.

rchl:
Unless you are talking about making a breaking change now with those added. Then I agree it will be future proof solution. But we need to break API at least once (possibly in an opt-in way though).

rwols:
The name "LanguageHandler" is bad anyway. It doesn't handle a language. It handles a language server. e.g. a hypothetical LSP-clangd and LSP-ccls would both handle the same four languages.

rwols:
Personally I think it'd be best if we can make it work so that LSP-foo packages can inherit from `Session` instead. Have a `SessionFactory` in the base LSP package and allow LSP-foo packages to register themselves to the `SessionFactory`. When the `WindowManager` is about to start a session for a view, it asks the `SessionFactory` for a suitable session (which could then return a derived session from an LSP-foo package).

rwols:
And we shouldn't be afraid of breaking things for now if that makes the ecosystem better in the long run. We're not even at version 1.0.0 for LSP yet. And nobody said anything about API stability promises.

rchl:
> When the `WindowManager` is about to start a session for a view, it asks the `SessionFactory` for a suitable session

I guess we can figure out details later but right now `on_start` can prevent a session from being created so it wouldn't be possible if handler would inherit Session.

> And we shouldn't be afraid of breaking things for now if that makes the ecosystem better in the long run. We're not even at version 1.0.0 for LSP yet. And nobody said anything about API stability promises.

We probably control 99% of LanguageHandler implementations right now anyway so we can update those in controlled way.

rwols:
> I guess we can figure out details later but right now on_start can prevent a session from being created so it wouldn't be possible if handler would inherit Session.

Well, that's easily solved by adding a classmethod of sorts:

```python
# LSP package
class Session:

    @classmethod
    def should_start(cls, config: ClientConfig, folders: Optional[List[WorkspaceFolder]]) -> Optional[str]:
        # Return the *REASON* for why we should NOT start.
        # Returning None means we can start.
        return None
```

```python
# LSP-foo package
class FooSession(Session):

    @classmethod
    def should_start(cls, config: ClientConfig, folders: Optional[List[WorkspaceFolder]]) -> Optional[str]:
        return "I want a .flowconfig present >:("

tomv564:
@rchl you're right, I completely assumed Python didn't care about extra defaulted arguments but it does!

I see two actionable conclusions for the two pull requests
- Any API updates should include `*`,  `**kwargs`, or take a dict so we don't have this problem in the future.
- Other than checking `callback.__code__.co_argcount`, there is no saving `on_start` and `on_initialized`, so we can only decide how much new API we want to create.

Since @rwols opened another can of worms about totally reimplementing the api:
* I don't think exposing `Session` is a good idea - we only need workspace folders, the language handlers don't care about capabilities or the session lifecycle management stuff in there.
* Using inheritance to Session would introduce even tighter coupling between handlers and LSP.

Let's take the rest of the discussion about a future API to #927

rwols:
Superseded by #979

PR #902, opened 2020-03-06
Handle $/progress

tomv564:
Super naïve mini-implementation, but unsure if LSP wants to implement spinners etc for itself.

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -607,6 +607,10 @@ class WindowManager(object):
             "textDocument/publishDiagnostics",
             lambda params: self.diagnostics.receive(session.config.name, params))
 
+        client.on_notification(
+            "$/progress",
+            lambda params: self._show_progress(params))
+
         self._handlers.on_initialized(session.config.name, self._window, client)
 
         client.send_notification(Notification.initialized())
@@ -636,6 +640,13 @@ class WindowManager(object):
     def _check_window_closed(self) -> None:
         if not self._is_closing and not self._window.is_valid():
             self._handle_window_closed()
+
+    def _show_progress(self, params) -> None:
+        # $/progress: {'token': 'rustAnalyzer/cargoWatcher', 'value': {'kind': 'begin', 'cancellable': False, 'title': "Running 'cargo check'"}}
+        value = params.get('value', {})
+        if value:
+            if value.get('kind') == 'begin':
+                self._sublime.status_message(value.get('title'))
 
     def _handle_window_closed(self) -> None:
         debug('window {} closed, ending sessions'.format(self._window.id()))


rchl:
Contrary to what I thought before, shouldn't the implementation be server specific since the type of Value can be anything? At least based on documentation at https://microsoft.github.io/language-server-protocol/specification#progress
What is this implementation based on?

tomv564:
Right, the spec is really unspecific!
I’m not sure how adoption is across all servers but rust-analyzer uses some of the types described in https://github.com/microsoft/language-server-protocol/issues/786

rwols:
From what I understand there are two types of "partial" messages:

1) progress notifications from the server to the client:
  a) "begin" work notification
  b) "busy" work notification
  c) "end" work notification
  To enable these notifications we:
    1) advertise it in our client capabilities
    2) put a `progressToken` key-value in our request to a server.
  e.g. we put `"progressToken": "foo"` in our `InitializeParams`, and then a server can send a bunch of `$/progress` notifications, the first one has `"kind": "begin`", intermediate ones have `"kind": "report"`, and the final one has `"kind": "end"`. All of these `$/progress` notifications are expected to have `"progressToken": "foo"` in them. Finally the server then sends a reply to the `initialize` request.

2) partial result from the server to the client:
  These are also reported through `$/progress`, but the `"value"` key's value has a type that depends on our request. For instance, if we send a `textDocument/references` request with a `"partialResultToken": "foo"`, then a server can send one `Location` per `$/progress` notification. e.g. it could notify three partial results:
```js
{"token": "foo", "value": [{"uri": "a.js", "range": {...}}]}
{"token": "foo", "value": [{"uri": "b.js", "range": {...}}]}
{"token": "foo", "value": [{"uri": "c.js", "range": {...}}]}
```
  followed by an empty response to our `textDocument/references` request, signaling that these are all the results:
```js
[]  // would have been the three lists combined above if it weren't for partial results
```

So the type of `"value"` depends on whether it came from a progressToken or from a partialResultToken. And in the case of a partialResultToken, the type depends on what the client request was.

The advantage of partial results is of course that the user gets more immediate feedback on what is happening. For extremely large codebases where doing a textDocument/references could take 10 seconds this is a true improvement.

I think handling partial results would require us to have much more state in our request handling logic. With asyncio this would certainly be doable, but I wouldn't have a go at it with Python 3.3.

Handling progress reports would be doable with Python 3.3. Just eat those notifications and report them to the status bar.

It seems odd that `rust-analyzer` would report progress since we don't send a `progressToken` in our requests. Seems like an oversight on `rust-analyzer`'s side?

In any case, if a "begin" progress notification is sent, I would expect to see an "end" progress notification too.

jwortmann:
>        2. put a `progressToken` key-value in our _request_ to a server.
>           e.g. we put `"progressToken": "foo"` in our `InitializeParams`, and then a server can send a bunch of `$/progress` _notifications_

It seems like the request and token can also be initiated by the server: [Initiating Work Done Progress](https://microsoft.github.io/language-server-protocol/specification#initiatingWorkDoneProgress)

Here is an extract of the payload from the Julia server in VS Code:
```
[Trace - 20:58:29] Received request 'window/workDoneProgress/create - (05459229-f517-4073-8f7c-8e3eb7def59e)'.
Params: {
    "token": "e397466d-9a6c-478d-968f-792a3955966b"
}

[Trace - 20:58:29] Sending response 'window/workDoneProgress/create - (05459229-f517-4073-8f7c-8e3eb7def59e)'. Processing request took 0ms
No result returned.

[Trace - 20:58:29] Received notification '$/progress'.
Params: {
    "token": "e397466d-9a6c-478d-968f-792a3955966b",
    "value": {
        "kind": "begin",
        "message": "Indexing packages...",
        "title": "Julia Language Server"
    }
}
```

Then the server indeed sends a lot of `$/progress` notifications with `"kind": "report"` and in the end one notification with

```
[Trace - 21:00:17] Received notification '$/progress'.
Params: {
    "token": "e397466d-9a6c-478d-968f-792a3955966b",
    "value": {
        "kind": "report",
        "message": "Indexing StaticArrays..."
    }
}

.....

[Trace - 21:12:25] Received notification '$/progress'.
Params: {
    "token": "e397466d-9a6c-478d-968f-792a3955966b",
    "value": {
        "kind": "end"
    }
}
```

This is how it is displayed in the VS Code status bar:
screenshot
 
I tried to recreate this functionality based on the payload log and this PR in https://github.com/jwortmann/LSP/commit/18cc965feaf8ba8d9115d54278ef5b9188f0b1f3, but it doesn't work yet. I can write an update here in case I get it to work somehow later.

rwols:
A very complete implementation is now in master https://github.com/sublimelsp/LSP/pull/1014

PR #904, opened 2020-03-10
Fix point_to_offset() when Position.character > line length

jfcherng:
# Problem

image

<details><summary>YAML codes for reproduction</summary>

```yaml
# This file is the entry point to configure your own services.
parameters:
    locale: 'en'
j: ""x
services:
    # default configuration for services in *this* file
    _defaults:
        autowire: true      # Automatically injects dependencies in your services.
        autoconfigure: true # Automatically registers your services as commands, event subscribers, etc.
        public: false       # Allows optimizing the container by removing unused services; this also means
                            # fetching services directly from the container via $container->get() won't work.
                            # The best practice is to be explicit about your dependencies anyway.
```

</details>

<details><summary>LSP diagnostic panel</summary>

```text
◌ C:\Users\Clover\Desktop\test.yaml:
        4:115 	None        	error     	can not read a block mapping entry; a multiline key may not be an implicit key
```

</details>

<details><summary>LSP payload log</summary>

```text
:: <-- yaml-lsp textDocument/publishDiagnostics: {'diagnostics': [{'message': 'can not read a block mapping entry; a multiline key may not be an implicit key', 'range': {'end': {'character': 114, 'line': 3}, 'start': {'character': 114, 'line': 3}}, 'severity': 1}], 'uri': 'file:///C:/Users/Clover/Desktop/test.yaml'}
```

</details>

I have [yaml-language-server](https://www.npmjs.com/package/yaml-language-server) installed and sometimes I see weird red underlines (like in line 8) and I don't know why. Today, I open the diagnostic panel. It shows `line 4, character 115`. The `line 4` makes sense since there is an invalid trailing `x`. `character 115` makes me feel it's counted from the 0th line, which is incorrect according to LSP specifications, but even in that case, 115th char is actually on line 6 (not line 4).

# Fix

While ST's `view.text_point(row, col)` says 

> Note that col is interpreted as the number of characters to advance past the beginning of the row.

I checked the [Position](https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#position) definition in LSP specifications. It says 

> If the character value is greater than the line length it defaults back to the line length.

After this patch, it looks like

image

# References

- https://www.sublimetext.com/docs/3/api_reference.html
- https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#position

```ts
interface Position {
	/**
	 * Line position in a document (zero-based).
	 */
	line: number;

	/**
	 * Character offset on a line in a document (zero-based). Assuming that the line is
	 * represented as a string, the `character` value represents the gap between the
	 * `character` and `character + 1`.
	 *
	 * If the character value is greater than the line length it defaults back to the
	 * line length.
	 */
	character: number;
}
```

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -25,7 +25,12 @@ def get_line(window: Optional[sublime.Window], file_name: str, row: int) -> str:
 
 
 def point_to_offset(point: Point, view: sublime.View) -> int:
-    return view.text_point(point.row, point.col)
+    return view.text_point(
+        point.row,
+        # @see https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#position
+        # If the character value is greater than the line length it defaults back to the line length.
+        min(point.col, len(view.line(view.text_point(point.row, 0))))
+    )
 
 
 def offset_to_point(view: sublime.View, offset: int) -> Point:
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -1,9 +1,11 @@
+from LSP.plugin.core.protocol import Point
 from LSP.plugin.core.typing import Generator
 from LSP.plugin.core.url import filename_to_uri
 from LSP.plugin.core.views import did_change
 from LSP.plugin.core.views import did_open
 from LSP.plugin.core.views import did_save
 from LSP.plugin.core.views import MissingFilenameError
+from LSP.plugin.core.views import point_to_offset
 from LSP.plugin.core.views import text_document_formatting
 from LSP.plugin.core.views import text_document_position_params
 from LSP.plugin.core.views import text_document_range_formatting
@@ -24,7 +26,7 @@ class ViewsTest(DeferrableTestCase):
         self.view.set_scratch(True)
         self.mock_file_name = "C:/Windows" if sublime.platform() == "windows" else "/etc"
         self.view.file_name = MagicMock(return_value=self.mock_file_name)
-        self.view.run_command("insert", {"characters": "hello world"})
+        self.view.run_command("insert", {"characters": "hello world\nfoo bar baz"})
 
     def tearDown(self) -> None:
         self.view.close()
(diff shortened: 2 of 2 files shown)


tomv564:
I wonder if the output from the yaml server is intentional? Why 115?
Too bad there's no real way to write the logic optimising the valid-column case. Validating is half the cost/complexity :(

jfcherng:
> I wonder if the output from the yaml server is intentional? Why 115?

https://github.com/redhat-developer/yaml-language-server/blob/07ca6cadfd3220ad40020207686236b2a857407a/src/languageservice/parser/yamlParser07.ts#L163

My first guess is that it's counted from the begin of the file (which is wrong according to the spec) but somehow misaligned a little with the actual position in ST.

Yet shall we "fix" the conversion according to the LSP spec? It looks like making the conversion heavy indeed (without actually doing a profiling or playing with a large project).

rwols:
Be weary that changes to point_to_offset might break the workaround for gopls returning out-of-bound ranges for the textDocument/formatting request. There's a test for that in test_single_document.py:

https://github.com/sublimelsp/LSP/blob/cee7ff9a083545413231f21bed5d37c29627d191/tests/test_single_document.py#L109-L130

plugin/core/views.py
```diff
@@ -25,7 +25,12 @@ def get_line(window: Optional[sublime.Window], file_name: str, row: int) -> str:
 
 
 def point_to_offset(point: Point, view: sublime.View) -> int:
-    return view.text_point(point.row, point.col)
+    return view.text_point(
+        point.row,
+        # @see https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#position
+        # If the character value is greater than the line length it defaults back to the line length.
+        min(point.col, len(view.line(view.text_point(point.row, 0))))
+    )
```
rwols: Can I poke you to write a test for this? tests/test_views.py seems like a good place to add some tests. `UnitTesting: Test Current File` is your friend :)

jfcherng: I've added a test in https://github.com/sublimelsp/LSP/pull/904/commits/f45f90e5d0067ff33c3e7cc4f66d4a24f70d86a5. Thanks for the hint about `UnitTesting: Test Current File`.

PR #910, opened 2020-03-14
Fix workspace/didChangeConfiguration capability type

rchl:
Capability is defined to be a boolean so shouldn't be a dict.
https://microsoft.github.io/language-server-protocol/specification#workspace_didChangeConfiguration


rchl:
Nope. the `dynamicRegistration` is a boolean. Sorry.

PR #913, opened 2020-03-16
remove outdated eslint config

Narretz:
The recommended way is to use LSP-eslint.
That said, the default config was *very*
specific, so I doubt it worked for many people ;)

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -365,16 +365,8 @@
         "syntaxes": ["Packages/PHP/PHP.sublime-syntax"],
         "languageId": "php"
     },
-    "eslint":
-    {
-      "command": ["node", "/Users/tomv/Projects/tomv564/vscode-eslint/eslint-server/src/server"],
-      "scopes": ["source.js"],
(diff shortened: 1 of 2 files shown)


LSP.sublime-settings
```diff
@@ -365,16 +365,8 @@
...
-    {
-      "command": ["node", "/Users/tomv/Projects/tomv564/vscode-eslint/eslint-server/src/server"],
-      "scopes": ["source.js"],
-      "syntaxes": ["Packages/Babel/JavaScript (Babel).sublime-syntax", "Packages/JavaScript/JavaScript.sublime-syntax"],
-      "languageId": "javascript",
-      "initializationOptions": {
-        "nodePath": "/usr/local/bin/node"
-      }
-    },
+    //"eslint": { },
```
rchl: This is potentially a breaking change if someone was using existing config as a base and just changed command & nodePath but oh well, I think it's fine.

rchl:
Could you also add LSP-eslint to docs/index.md, where it mentions other LSP-* packages?

Narretz:
I've added it to index.md

Issue #891, opened 2020-02-24
Missing highlighting for severity= hint


Narretz:
I have a situation where I only get highlighting for severity = error.
To illustrate, in typescript, the red outlined diagnostic is an error, and the one above is a hint (unused variable).
<img width="146" alt="2020-02-25 004210" src="https://user-images.githubusercontent.com/1153097/75201157-fb71ab00-5767-11ea-9a09-d30e50824f93.png">

But I can't figure out why I only get a red outline for the error severity.
The relevant config entries:
```
{
	"auto_show_diagnostics_panel": false,
	"show_diagnostics_severity_level": 4,
	"diagnostics_highlight_style": "box",
}
```

The hint shows up in the diagnostic panel, so I know it's there

Narretz:
Oh I think I found the problem. In the following line
https://github.com/sublimelsp/LSP/blob/6a0afdfa99965326d12d26d06a72b54c09962c1d/plugin/diagnostics.py#L249

the call excludes `DiagnosticSeverity.Hint` because range() max is not inclusive. Essentially it stops at severity 3. Does that make sense?

rwols:
I think this was intentional. The idea was, from memory, that the user doesn't want to get flooded with squigly linter lines that signify hints. See: https://github.com/sublimelsp/LSP/pull/800

Narretz:
Oh, I see. I disagree with the decision to take this out of the user's hand completely. There should be an option that controls the severity highlight visibility, too. We can't just assume that no developer ever wants to see hint highlights.
Would a PR with an option be considered?

rwols:
> Would a PR with an option be considered?

Sounds good to me!

I'm starting to think it *is* actually an off-by-one mistake due to the wording in this commit: https://github.com/sublimelsp/LSP/pull/800/commits/3c267eab898cf43778a1d05e6b582045b8812b4b

Narretz:
Yeah "all regions" sounds like it's a bug.

rchl:
Resolved by #941

PR #914, opened 2020-03-16
feat: make diagnostics highlights configurable

Narretz:
... which allows to display hint diagnostics, which
were previoulsy never displayed.

The default setting is 3 to reflect the current status quo.

Fixes https://github.com/sublimelsp/LSP/issues/891

Are there tests for the highlighting? A quick search hasn't shown anything. I can try to add them, but from scratch is a bit difficult as I've never written tests with python.

### DIFF
--- a/plugin/diagnostics.py
+++ b/plugin/diagnostics.py
@@ -247,7 +247,7 @@ class DiagnosticViewRegions(DiagnosticsUpdateWalk):
         self._relevant_file = False
 
     def end(self) -> None:
-        for severity in range(DiagnosticSeverity.Error, DiagnosticSeverity.Hint):
+        for severity in range(DiagnosticSeverity.Error, settings.show_highlights_severity_level + 1):
             region_name = "lsp_" + format_severity(severity)
             if severity in self._regions:
                 regions = self._regions[severity]
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -23,13 +23,21 @@
   // under the cursor in status bar if available.
   "show_diagnostics_in_view_status": true,
 
-  // Show the diagnostics with level less than or equal to
+  // Show the diagnostics in the panel with level less than or equal to
   // the given value.
   // error: 1
   // warning: 2
   // info: 3
   // hint: 4
   "show_diagnostics_severity_level": 2,
+
+  // Highlights the diagnostics in the document with level less than or equal to
+  // the given value.
+  // error: 1
+  // warning: 2
+  // info: 3
+  // hint: 4
+  "show_highlights_severity_level": 3,
 
   // Highlighting style of code diagnostics.
   // Valid values are "underline" or "box"
@@ -533,15 +541,15 @@
     },
     "vscode-css":
     {
-      "command": 
+      "command":
       [
         "css-languageserver", "--stdio"
       ],
-      "scopes": 
+      "scopes":
       [
         "source.css"
       ],
-      "syntaxes": 
+      "syntaxes":
       [
         "Packages/CSS/CSS.sublime-syntax"
       ],
--- a/plugin/core/settings.py
+++ b/plugin/core/settings.py
@@ -67,6 +67,7 @@ def update_settings(settings: Settings, settings_obj: sublime.Settings) -> None:
                                                                        "show_diagnostics_count_in_view_status", False)
     settings.show_diagnostics_in_view_status = read_bool_setting(settings_obj, "show_diagnostics_in_view_status", True)
     settings.show_diagnostics_severity_level = read_int_setting(settings_obj, "show_diagnostics_severity_level", 2)
+    settings.show_highlights_severity_level = read_int_setting(settings_obj, "show_highlights_severity_level", 3)
     settings.diagnostics_highlight_style = read_str_setting(settings_obj, "diagnostics_highlight_style", "underline")
     settings.document_highlight_style = read_str_setting(settings_obj, "document_highlight_style", "stippled")
     settings.document_highlight_scopes = read_dict_setting(settings_obj, "document_highlight_scopes",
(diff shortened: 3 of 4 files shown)


plugin/diagnostics.py
```diff
@@ -247,7 +247,7 @@ def end_file(self, file_name: str) -> None:
         self._relevant_file = False
 
     def end(self) -> None:
-        for severity in range(DiagnosticSeverity.Error, DiagnosticSeverity.Hint):
+        for severity in range(DiagnosticSeverity.Error, settings.show_highlights_severity_level + 1):
```
predragnikolic, line 250: First, thanks for fixing this :) 

I noted that you added a new setting `show_highlights_severity_level`.
I would not increase the number of settings.

Can we instead just use the `show_diagnostics_severity_level`?
It makes sense that `show_diagnostics_severity_level` setting
should affect all the diagnostics in the panel,
but that it should also affect all the diagnostics highlights in the document.

roblav96: @predragnikolic @rchl

This should definitely be merged instead of #941 using `settings.show_diagnostics_severity_level`
```suggestion
        for severity in range(DiagnosticSeverity.Error, settings.show_diagnostics_severity_level + 1):
```
TypeScript language server likes to over report diagnostics, obnoxiously drawing squiggly lines all over the code I'm trying to write.

For example, the `unused variable` diagnostic draws a squiggly line under the variable I _just_ wrote.

rchl: Introducing new settings is something I'd personally rather avoid.
Another option could be to let the diagnostics panel show all severities always and rename existing `show_diagnostics_severity_level` to `show_highlights_severity_level`.

roblav96: I'm proposing to use the existing setting `show_diagnostics_severity_level` so there's no need for an additional setting or renaming.

roblav96: I personally would assume the `highlighting` is a sub-feature of the `diagnostics`.

rchl: `show_diagnostics_severity_level` currently affects diagnostics shown in the panel and diagnostic highlights. Keeping the existing name and not make it affect diagnostics in the panel would make the meaning of that setting inaccurate. That's why I think it should be renamed if we don't want it to apply to the panel anymore.

Because I guess the whole point of raising this issue is that you want to use different levels for the panel and for the highlights.

roblav96: Then the `show_diagnostics_severity_level` does not correctly match the diagnostics shown in the panel with the document diagnostic highlights.

For example,

## `"show_diagnostics_severity_level": 2,`

Screen Shot 2020-05-21 at 6 02 46 AM

Notice how the diagnostics panel **only** shows `errors`, yet the document still renders `hint` highlights.

## `"show_diagnostics_severity_level": 4,`

Screen Shot 2020-05-21 at 6 03 37 AM

Notice how the diagnostics panel shows `errors` and `hints`, and the document correctly renders both `errors` and `hints`.

rchl: This is not how it works for me.
Make sure that you restart after changing the `show_diagnostics_severity_level` setting for good measure.
Also, are you sure those highlights are from typescript server? Should be clear when hovering them.

roblav96: Yes I'm very familiar with the `Quit Sublime Text && Open Sublime Text` workflow after making _any_ settings.

Yes the `typescript-language-server` is sending `:: <-  ts-lsp textDocument/publishDiagnostics:`
```json
{
	"diagnostics": [
		{
			"source": "typescript",
			"range": {
				"end": {
					"character": 66,
					"line": 10
				},
				"start": {
					"character": 41,
					"line": 10
				}
			},
			"relatedInformation": [
				{
					"location": {
						"uri": "file:///Users/user/Projects/futon-media/node_modules/playwright-core/types/types.d.ts",
						"range": {
							"end": {
								"character": 45,
								"line": 2293
							},
							"start": {
								"character": 26,
								"line": 2293
							}
						}
					},
					"message": "An argument for 'userDataDir' was not provided."
				}
			],
			"severity": 1,
			"message": "Expected 1-2 arguments, but got 0.",
			"code": 2554
		},
		{
			"severity": 4,
			"source": "typescript",
			"range": {
				"end": {
					"character": 27,
					"line": 1
				},
				"start": {
					"character": 0,
					"line": 1
				}
			},
			"code": 6133,
			"message": "'R' is declared but its value is never read."
		},
		{
			"severity": 4,
			"source": "typescript",
			"range": {
				"end": {
					"character": 38,
					"line": 2
				},
				"start": {
					"character": 0,
					"line": 2
				}
			},
			"code": 6133,
			"message": "'utils' is declared but its value is never read."
		},
		{
			"severity": 4,
			"source": "typescript",
			"range": {
				"end": {
					"character": 8,
					"line": 6
				},
				"start": {
					"character": 6,
					"line": 6
				}
			},
			"code": 6133,
			"message": "'db' is declared but its value is never read."
		},
		{
			"severity": 4,
			"source": "typescript",
			"range": {
				"end": {
					"character": 12,
					"line": 10
				},
				"start": {
					"character": 5,
					"line": 10
				}
			},
			"code": 6133,
			"message": "'browser' is declared but its value is never read."
		}
	],
	"uri": "file:///Users/user/Projects/futon-media/src/adapters/chromium.ts"
}
```

Which is why I'm saying the change I've proposed (https://github.com/sublimelsp/LSP/pull/914#discussion_r428477992) will resolve this problem. I've even personally created an override of that `diagnostics.py` file to fix this problem.

rchl: This is already fixed on both st3 and st4 branches (that's why I got confused) but fix is not released yet. We should release more often.

predragnikolic: Yes is already in the master,
We should also probably close this PR.

rwols: I’ll make a release soon-ish.

rwols: Version 0.11.0 is pushed which should contain the desired behavior.

predragnikolic:
I will close this PR, because this has been addressed.

PR #916, opened 2020-03-17
[erlang-ls/erlang_ls#401] Add Erlang to Sublme LSP Client documentation

jfacorro:
Add Erlang to Sublme LSP Client documentation (... and also removes some trailing whitespaces).

Fixes erlang-ls/erlang_ls#401.

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -314,6 +314,13 @@
         "Packages/Elixir/Syntaxes/Elixir.tmLanguage",
       ]
     },
+    "erlang-ls": {
+      "command"   : [ "/path/to/my/erlang_ls", "--transport", "stdio" ],
+      "enabled"   : true,
+      "languageId": "erlang",
+      "scopes"    : [ "source.erlang" ],
+      "syntaxes"  : ["Packages/Erlang/Erlang.sublime-syntax"]
+    },
     "flow":
     {
       "command": ["flow", "lsp"],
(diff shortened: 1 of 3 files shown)


LSP.sublime-settings
```diff
@@ -314,6 +314,13 @@
         "Packages/Elixir/Syntaxes/Elixir.tmLanguage",
       ]
     },
+    "erlang-ls": {
+      "command"   : [ "/path/to/my/erlang_ls", "--transport", "stdio" ],
+      "enabled"   : true,
```
rwols: Please remove `"enabled": true` here.

Other than that, looks mergeable to me!

jwortmann:
I just saw this PR. The double trailing whitespaces at the Fortran and Julia sections in the documentation were intentional, because 2 trailing spaces indicates a linebreak in Markdown. Now the correct formatting is gone. Also fenced code blocks aren't allowed inside lists, therefore the third item in the Erlang section wrongly has the number "1." again. https://lsp.readthedocs.io/en/latest/#erlang

rchl:
Don't want to defend the change but relying on trailing whitespace sounds bad as many have settings to remove those automatically (or at least highlight as errors). Is there maybe a better way to handle the same (for example an extra blank line between items)?

jfacorro:
@jwortmann Thank you for the heads up. PR addressing the issues mentioned [here](https://github.com/sublimelsp/LSP/pull/924).

PR #919, opened 2020-03-18
Narrow down imports

rwols:
This should be everything. Please check if I forgot something!

### DIFF
--- a/boot.py
+++ b/boot.py
@@ -1,30 +1,42 @@
-from .plugin.core.main import startup, shutdown
-
-# TODO: narrow down imports
-from .plugin.core.panels import *
+# Please keep this list sorted (Edit -> Sort Lines)
+from .plugin.code_actions import LspCodeActionBulbListener
+from .plugin.code_actions import LspCodeActionsCommand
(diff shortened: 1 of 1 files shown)


boot.py
```diff
@@ -1,30 +1,42 @@
-from .plugin.core.main import startup, shutdown
-
-# TODO: narrow down imports
-from .plugin.core.panels import *
+# Please keep this list sorted (Edit -> Sort Lines)
+from .plugin.code_actions import LspCodeActionBulbListener
+from .plugin.code_actions import LspCodeActionsCommand
```
jfcherng: I assume 

```py
from .plugin.code_actions import LspCodeActionBulbListener, LspCodeActionsCommand
```

is preferred in this github organization?

rwols: It doesn't really matter that much. But I prefer having each import on a separate line and alphabetically sorted. This makes diffs easier to review, and makes merge conflicts much easier to resolve. If I'm adding a new import and the other imports have a certain style, I'll try to keep to that style. It depends on the context.

PR #920, opened 2020-03-18
Differentiate between `null` init options and empty dict

### LINKED ISSUE #917: Empty initializationOptions is not sent
# Problem

If the `initializationOptions` is an empty dict, it won't be sent to the server.
```js
// this is not sent
"initializationOptions": {},
```

Some servers (such as [vscode-css-languageserver](https://github.com/vscode-langservers/vscode-css-languageserver)) need `initializationOptions` to be set even if a empty one or a useless one (as a workaround) such as the following.

```js
// this will be sent
"initializationOptions": {"just_an_useless_key": 1},
```

# Reference

- https://github.com/sublimelsp/LSP-css/pull/2#discussion_r393881421

rwols:
close #917

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -85,7 +85,7 @@ def get_initialize_params(workspace_folders: List[WorkspaceFolder], designated_f
             }
         }
     }
-    if config.init_options:
+    if config.init_options is not None:
         initializeParams['initializationOptions'] = config.init_options
 
     return initializeParams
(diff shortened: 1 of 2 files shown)


plugin/core/sessions.py
```diff
@@ -85,7 +85,7 @@ def get_initialize_params(workspace_folders: List[WorkspaceFolder], designated_f
             }
         }
     }
-    if config.init_options:
+    if config.init_options is not None:
```
rchl: `ClientConfig` defaults to `dict` though, right? So this will in practice never be None.

I think the most spec-compliant way would be to not default to `dict` in `ClientSettings` since per-spec the value can be `any`thing.

But this way is more compatible probably, given the wild west out there.

rwols: You're right, I hadn't thought of ClientConfig having a default value of `{}` for `init_options`.

So the behavior previously was that we don't send an initializationOptions if the dict was empty.
The resulting behavior due to this diff now is that we *always* send initializationOptions, except when the user puts `"initializationOptions": null` in LSP.sublime-settings. If the user leaves out that key LSP will send `"initializationOptions": {}`.

It would be a little neater if init_options were None by default. But like you mentioned, changes to these parts of the code are very sensitive, so let's do that in another PR.

PR #924, opened 2020-03-21
Fix formatting and numbering for items

jfacorro:
It was pointed out [here](https://github.com/sublimelsp/LSP/pull/916#issuecomment-602043968) that after having deleted trailing whitespaces which signaled newlines, the format of the Julia and Fortran sections was broken. Also the numbering in the Erlang section was off.

This PR removes the reliance on trailing whitespaces and fixes the numbering.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -281,21 +281,19 @@ dub run dls:bootstrap
 1. See instructions for installing the [Erlang Language Server](https://github.com/erlang-ls/erlang_ls).
 2. Add to LSP settings' clients:
 
-```json
-"erlang-ls": {
-  "command"   : [ "/path/to/my/erlang_ls", "--transport", "stdio" ],
-  "enabled"   : true,
-  "languageId": "erlang",
-  "scopes"    : [ "source.erlang" ],
-  "syntaxes"  : ["Packages/Erlang/Erlang.sublime-syntax"]
-}
-```
+         "erlang-ls": {
+           "command"   : [ "/path/to/my/erlang_ls", "--transport", "stdio" ],
+           "enabled"   : true,
+           "languageId": "erlang",
+           "scopes"    : [ "source.erlang" ],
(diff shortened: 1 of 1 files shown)


rwols:
This removes the syntax highlighting in the code blocks. I'm okay with that. What do you think @jwortmann ?

jfcherng:
> This removes the syntax highlighting in the code blocks.

But why we have to? `readthedocs` doesn't support it (I have never used it before)?


# Source

````
1. 1st
1. 2nd

   ```json
   "erlang-ls": {
     "test": 1
   }
   ```

1. 3rd
````


# Github Rendered

1. 1st
1. 2nd

   ```json
   "erlang-ls": {
     "test": 1
   }
   ```

1. 3rd

jwortmann:
Readthedocs uses Mkdocs to generate the html pages from Markdown. GitHub apparently uses another Markdown engine, here is how Mkdocs renders the code from @jfcherng for me locally:
screenshot

The reason I used escaped dots `1\.` after the numbers instead of lists was that I found no other way to keep the syntax highlighting for the code blocks. Empty lines between items instead of the trailing double space will actually add empty lines in the rendered html.
  
I would at least suggest to keep the syntax highlighting for the json and js fenced code blocks after the third list item.

jfcherng:
For mkdocs, the following is good in my test.

````
1. 1st
1. 2nd

    ```json
    "erlang-ls": {
      "test": 1
    }
    ```

1. 3rd
````

But generally I don't like dedicated tweaks if I don't know why it works.

rwols:
Hmm, I'm not sure what to do with this PR. Personally I don't really mind the missing highlighting. If nobody speaks up I'll merge this tomorrow.

jwortmann:
I've tested how the generated html from Mkdocs will look like, and interestingly we actually don't lose the highlighting with this change. Seems like the JSON/JavaScript language is automatically recognized for indented code blocks and there is only a different line spread. With that in mind, I'm fine with this change.

screenshot

PR #925, opened 2020-03-24
Use line-wrapping algorithm provided by mdpopups

### LINKED ISSUE #859: Forced-break in hover popup can break syntax highlighting
Using: OSX / typescript-language-server

Line-break, that LSP forces in the popup, can cause syntax highlighting to break. For example, if breaking a plain string in JS syntax.

line-break

Notice that in the screenshot the whole string should have a yellow color. Syntax highlighting breaks because line break within a string is not a valid syntax.

rchl:
It seems to only break lines if it finds opportunity to
do so (whitespace). That means that in some cases lines won't
be wrapped where old code would aggressively wrap.

Resolves #859

### DIFF
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -196,7 +196,7 @@ class LspHoverCommand(LspTextCommand):
                 formatted.append(value)
 
         if formatted:
-            frontmatter_config = "---\nallow_code_wrap: true\n---\n"
+            frontmatter_config = mdpopups.format_frontmatter({'allow_code_wrap': True})
             return mdpopups.md2html(self.view, frontmatter_config + "\n".join(formatted))
 
         return ""
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -3,7 +3,6 @@ import sublime
 import sublime_plugin
 import webbrowser
 import os
-import textwrap
 from html import escape
 from .code_actions import actions_manager, run_code_action_or_command
 from .code_actions import CodeActionOrCommand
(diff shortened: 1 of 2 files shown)


tomv564:
I think I invested some time in documenting the old issues in my pull request https://github.com/sublimelsp/LSP/pull/776 - would be awesome if you can show what those cases look like now!

rchl:
Honestly, for those cases in your first comment there, there doesn't seem to be any difference.

These look the same both before and after this change:
long-signature
long-description

Which is interesting because the long description of `font-family: system` should have been line wrapped before my change per your screenshot there.

rchl:
Are there still some concerns?
To summarize:
1. This change only affects hover contents (`textDocument/hover` responses), not signatures or diagnostics.
2. The cases that are improved are:
- when there is content without newline or whitespace -- it won't be wrapped now while previously it was wrapped at a hard-coded column and sometimes broke syntax
- when there is content without newline but with whitespace -- the content will break at a more reasonable place now than before

plugin/hover.py
```diff
@@ -191,16 +190,14 @@ def hover_content(self) -> str:
...
-
             if language:
                 formatted.append("```{}\n{}\n```\n".format(language, value))
             else:
                 formatted.append(value)
 
         if formatted:
-            return mdpopups.md2html(self.view, "\n".join(formatted))
+            frontmatter_config = "---\nallow_code_wrap: true\n---\n"
+            return mdpopups.md2html(self.view, frontmatter_config + "\n".join(formatted))
```
rwols: I think it'd be just a tad bit nicer if this was using the `mdpopups.format_frontmatter` function.

rchl: Done

PR #926, opened 2020-03-24
Startup folder selection

tomv564:
Adds language handler api for verifying a given folder can be started.
Example usage:

```python
from LSP.plugin.core.handlers import LanguageHandler
from LSP.plugin.core.settings import ClientConfig, LanguageConfig
import shutil
import os

config_name = 'flow'
server_name = 'flow'
flow_config = ClientConfig(
    name=config_name,
    binary_args=[
        "flow", "lsp"
    ],
    tcp_port=None,
    languages=[
        LanguageConfig("javascript", ["source.js"], ["Packages/Babel/JavaScript (Babel).sublime-syntax", "Packages/JavaScript/JavaScript.sublime-syntax"])
    ],
    enabled=True,
    init_options=dict(),
    settings=dict(),
    env=dict())


def flow_is_installed() -> bool:
    return shutil.which("flow") is not None


def has_flow_config(path) -> bool:
    has_config = os.path.exists(os.path.join(path, ".flowconfig"))
    if not has_config:
        print("No .flowconfig found, skipping folder: ", path)
    return has_config

class LspFlowPlugin(LanguageHandler):
    def __init__(self):
        self._name = config_name
        self._config = flow_config

    @property
    def name(self) -> str:
        return self._name

    @property
    def config(self) -> ClientConfig:
        return self._config

    def can_start(self, window, folders, file_path): -> 'List[str]':
        if not flow_is_installed():
            window.status_message(
                "Flow must be installed to run {}".format(server_name))
            return []
        configured_folders = [folder for folder in folders if has_flow_config(folder)]
        if configured_folders:
            if not file_path.startswith(configured_folders[0]):
                print("Folder for file not configured:" + file_path)
                return []
        return configured_folders
```

### DIFF
--- a/plugin/core/registry.py
+++ b/plugin/core/registry.py
@@ -9,10 +9,11 @@ from .sessions import Session
 from .settings import settings, client_configs
 from .types import ClientConfig, ClientStates, WindowLike
 from .windows import WindowRegistry, DocumentHandlerFactory, WindowManager
-from .typing import Optional, Callable, Dict, Any, Iterable
+from .typing import Optional, Callable, Dict, Any, Iterable, List
 
 
 client_start_listeners = {}  # type: Dict[str, Callable]
+client_can_start_listeners = {}  # type: Dict[str, Callable]
 client_initialization_listeners = {}  # type: Dict[str, Callable]
 
 
@@ -43,11 +44,13 @@ class LSPViewEventListener(sublime_plugin.ViewEventListener):
 
 class LanguageHandlerDispatcher(object):
 
-    def on_start(self, config_name: str, window: WindowLike) -> bool:
+    def on_start(self, config_name: str, window: WindowLike, folders: List[str], file_path: str) -> List[str]:
         if config_name in client_start_listeners:
-            return client_start_listeners[config_name](window)
+            return folders if client_start_listeners[config_name](window) else []
+        elif config_name in client_can_start_listeners:
+            return client_can_start_listeners[config_name](window, folders, file_path)
         else:
-            return True
+            return folders
 
     def on_initialized(self, config_name: str, window: WindowLike, client: Client) -> None:
         if config_name in client_initialization_listeners:
@@ -64,9 +67,11 @@ def register_language_handler(handler: LanguageHandler) -> None:
     debug("received config {} from {}".format(handler.name, handler.__class__.__name__))
     client_configs.add_external_config(handler.config)
     if handler.on_start:
-        client_start_listeners[handler.name] = handler.on_start
+        client_start_listeners[handler.config.name] = handler.on_start
+    if handler.can_start:
+        client_can_start_listeners[handler.config.name] = handler.can_start
     if handler.on_initialized:
-        client_initialization_listeners[handler.name] = handler.on_initialized
+        client_initialization_listeners[handler.config.name] = handler.on_initialized
 
 
 def client_from_session(session: Optional[Session]) -> Optional[Client]:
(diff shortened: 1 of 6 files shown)


plugin/core/registry.py
```diff
@@ -43,11 +44,13 @@ def has_manager(self) -> bool:
 
 class LanguageHandlerDispatcher(object):
 
-    def on_start(self, config_name: str, window: WindowLike) -> bool:
+    def on_start(self, config_name: str, window: WindowLike, folders: List[str], file_path: str) -> List[str]:
         if config_name in client_start_listeners:
-            return client_start_listeners[config_name](window)
+            return folders if client_start_listeners[config_name](window) else []
+        elif config_name in client_can_start_listeners:
+            return client_can_start_listeners[config_name](window, folders, file_path)
```
rchl: I have flashbacks to #897.

I think it's just not great to have to create a new API that is basically a copy of existing one, just with different arguments. There should be a way to add arguments without breaking the API or some other clean solution for that.

I would also like to have some good way of fixing #897 without introducing a new API that duplicates existing functionality so I'm interested in fixing this properly and in a generic way.

tomv564: Yes, solving the same problem.
I'll comment on #897 to see how far we can get there.

plugin/core/windows.py
```diff
@@ -430,16 +430,18 @@ def _start_client(self, config: ClientConfig, file_path: str) -> None:
             debug('Already starting on this window:', config.name)
             return
 
-        if not self._handlers.on_start(config.name, self._window):
+        workspace_paths = sorted_workspace_paths(self._workspace.folders, file_path)
+        startable_folders = self._handlers.on_start(config.name, self._window, workspace_paths, file_path)
+
+        if not startable_folders:
```
rchl: It doesn't quite feel right to return a list of folders from can_start.

What would the meaning be when the server does and doesn't support workspaceFolders?
 - if it doesn't then we start one server per window, even with multiple folders, right?
With this here, we would start it if at least one folder has flow config.
Would that be a problem for the other folder(s) since those wouldn't have a config?
 - If it does support workspaces then I guess can_start will receive a list with just one folder right? If yes then that case should be fine.

But then it appears that the returned list of folders is not used anyway so why not just return True/False?

rchl: Or are we gonna get multiple folders when the server supports workspaceFolders? If so then we would probably have to answer yes/no per-folder.

tomv564: We don't know if the server supports workspace folders at this point (before it has replied to `initialize`), so here the handler will always receive all the folders in the project.
The possible responses with this use are:
* All `workspace_paths` returned (server doesn't care what folders are opened)
* Some `workspace_paths` returned (server is picky about what folders it supports, eg flow lsp)
* No `workspace_paths` are returned (server can't start at all - eg. is not installed)

rwols: > What would the meaning be when the server does and doesn't support workspaceFolders?

> Or are we gonna get multiple folders when the server supports workspaceFolders? If so then we would probably have to answer yes/no per-folder.

In an LSP-foo package, you know exactly which server you're dealing with, so you know whether that server supports workspace folders or not.

rchl: And what about that part?

> But then it appears that the returned list of folders is not used anyway so why not just return True/False?

What happens with Julia server if one of the folders doesn't have configuration file. Will it crash anyway?
Does its behavior depend on which file is opened first (from which folder)?

tomv564:
Hope it's okay I close this, if the problem still exists and someone cares about this issue in the future it's probably best rewritten as a totally new pull request ;)

PR #928, opened 2020-03-25
Add support for Swift's sourcekit-lsp

saagarjha:
Just a little thing I whipped up using the SourceKit LSP server that ships with Xcode 11.4. This seems to work, but is this the right way to add support for a new server?

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -570,7 +570,12 @@
         "Packages/R/R.sublime-syntax"
       ],
       "languageId": "r"
-    }
-
+    },
+    "sourcekit-lsp": {
+      "command": ["xcrun", "sourcekit-lsp"],
+      "scopes": ["source.swift"],
+      "syntaxes": ["Packages/Swift/Syntaxes/Swift.tmLanguage"],
+      "languageId": "swift"
+    },
   }
 }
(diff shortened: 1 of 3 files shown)


LSP.sublime-settings
```diff
@@ -570,7 +570,12 @@
...
       "languageId": "r"
-    }
-
+    },
+    "sourcekit-lsp": {
+      "command": ["xcrun", "sourcekit-lsp"],
+      "scopes": ["source.swift"],
+      "syntaxes": ["Packages/Swift/Syntaxes/Swift.tmLanguage"],
+      "languageId": "swift"
+    },
```
rwols: Looking at sourcekit-lsp's github page, it seems that this server can also handle the C-family of languages (C, C++, ObjC, ObjC++). This makes sense as it's based on clangd. Please add those languages. (See the clangd entry for guidance on how to add more than one language)

https://github.com/apple/sourcekit-lsp/tree/master/Editors

saagarjha: sourcekit-lsp seems to directly call clangd, which also now ships with Xcode. Do you think I should add it anyways (knowing that if you want clangd, it's already available under clangd)? For example, I personally do not have clangd enabled because I cannot figure out a way to pass it compiler flags other than compile_flags.txt, which does not care for the project's language. So I would need to add this to my settings to disable all the "C" features because I can only enable all of the languages together:

```json
"sourcekit-lsp": {
	"enabled": true,
	"languages": [
		{
			"scopes": ["source.swift"],
			"syntaxes": ["Packages/Swift/Syntaxes/Swift.tmLanguage"],
			"languageId": "swift"
		}
	]
}
```

rwols: OK sounds good to me. There are already at least two langservers handling the C-family anyway.

PR #929, opened 2020-03-27
Add 'actions' class to hover popup action links.

FHTheron:
The popups.css selector uses ".lsp_popups .actions a" which does not
match the generated html. Without this, the links are solid blue without
framing.

### DIFF
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -127,7 +127,7 @@ class LspHoverCommand(LspTextCommand):
             actions.append("<a href='{}'>{}</a>".format('references', 'References'))
         if self.has_client_with_capability('renameProvider'):
             actions.append("<a href='{}'>{}</a>".format('rename', 'Rename'))
-        return "<p>" + " | ".join(actions) + "</p>"
+        return "<p class='actions'>" + " | ".join(actions) + "</p>"
 
     def format_diagnostic_related_info(self, info: DiagnosticRelatedInformation) -> str:
         file_path = info.location.file_path


rwols:
Do you have a screenshot of the before and after styling?

FHTheron:
With my local custom theme.
Before
CustomthemeBefore
After
CustomthemeAfter

With a clean ST 4096 installation.
Before
CleaninstallBefore
After
CleaninstallAfter

So, I can update popups.css to make the links look like the default style in Before but I'm not entirely sure what CSS would be correct for that. Or my approach to this is incorrect.

jwortmann:
This rule in the `popup.css` is the reason why the links are not underlined blue anymore with this change:
```css
.lsp_popup .actions a {
    text-decoration: none;
    color: var(--foreground);
}
```

I guess the `action` class was just accidentally omitted in the popup html. I personally prefer the style as it will be with this PR, and therefore don't think the CSS necessarily needs to be updated.

<hr>

@FHTheron Do you mind sharing the `popup_css` and `phantom_css` values that you use in your color scheme? I was experimenting with these settings for my color scheme a while ago, but I refrained from using it because ST applies some undocumented CSS rules if those settings are not used and because there is a bug in ST to evaluate CSS specificity correctly in some cases.

FHTheron:
The rule above it is also important, it provides the framing/background:
`.lsp_popup .actions {`

As for my colour schemes, I have this in global scheme settings:
`"popup_css": "body{background-color:#444;border:1px solid #ffffff55;color:white;padding:12px}",`

But I also messed with mdpopups:
```diff
--- ../mdpopups/mdpopups_css/default.css        2020-02-29 14:37:56.721366200 +0200
+++ mdpopups.css        2020-02-05 15:28:10.000000000 +0200
@@ -1,6 +1,6 @@
 {%- if var.default_style %}
 html {
-  --mdpopups-font-mono: "sf mono", Consolas, "Liberation Mono", Menlo, Courier, monospace;
+  --mdpopups-font-mono: "Input Mono Narrow";
   {%- if var.sublime_version < 3127 %}
   --mdpopups-admon-fg: var(--background);
   --mdpopups-admon-info-fg: var(--background);
@@ -40,13 +40,13 @@
 html.light {
   --mdpopups-hr: color(var(--foreground) blend(white 80%));
   --mdpopups-kbd-border: color(var(--foreground) blend(white 80%));
-  --mdpopups-hl-border: color(var(--background) blend(black 90%));
+  --mdpopups-hl-border: white;
   --mdpopups-hl-bg: color(var(--background) blend(black 98%));
 }
 html.dark {
   --mdpopups-hr: color(var(--foreground) blend(black 80%));
   --mdpopups-kbd-border: color(var(--foreground) blend(black 80%));
-  --mdpopups-hl-border: color(var(--background) blend(white 90%));
+  --mdpopups-hl-border: #fff5;
 {%- if var.sublime_version < 3127 %}
   /* 3127 borders allow us more flexibility with code backgrounds. Older versions are going to stick to darkening. */
   --mdpopups-hl-bg: color(var(--background) blend(black 98%));
@@ -59,7 +59,7 @@
   display: block;
   margin: 0;
   padding: 0;
-  font-size: 1rem;
+  font-size: 0.9rem;
   line-height: 1.1rem;
 }
 .mdpopups .highlight,
@@ -123,7 +123,8 @@
 /* kbd */
 .mdpopups kbd {
   display: inline;
-  font-size: 0.9rem;
+  overflow-wrap: all;
+  font-size: 0.6rem;
   padding: 0.05rem 0.25rem;
   border-radius: 0.25rem;
   background-color: var(--mdpopups-kbd-bg);
```

I just realised I kind of want `mdpopups-font-mono` to stay in sync with my normal font choice for the week. Kind of wondered why the popup looks different.

For phantoms (now annotations) I edit Default/exec.py directly.

Issue #867, opened 2020-01-30
Creaete separate tags for package control for ST4?


predragnikolic:
How about adding a new tag for ST4 on package control? 

That would allow people who use ST4 to test the `st4000-exploration` branch :)

Here is an example I found on how to achieve this:
```

		// If your package is only compatible with specific builds of
		// Sublime Text, this will cause the package to be hidden from
		// users with incompatible versions.
		//
		// The "tags" key can be true for all valid semantic version tags, or
		// can be a prefix string. Only tags in the form
		// {prefix}{semantic_version} will be selected. In the example below,
		// the entry with "sublime_text": "<3000" will match tags like:
		//
		// "st2-1.0.0"
		// "st2-1.1.0"
		//
		// The release with "sublime_text": ">=3000" will match tags like:
		//
		// "st3-1.0.0"
		// "st3-1.1.0"
		{
			"details": "https://github.com/wbond/sublime_alignment",
			"releases": [
				{
					"sublime_text": "<3000",
					"tags": "st2-"
				},
				{
					"sublime_text": ">=3000",
					"tags": "st3-"
				}
			
```

rwols:
I've made the appropriate changes here https://github.com/sublimelsp/package_control_channel/tree/lsp-tag-split

One question I have is what happens when we don't make a new tag in the meantime and this is merged in the upstream package_control_channel repository?

rwols:
Perhaps it's a better strategy to use the minimum-supported ST build number as the major version. e.g.

3092.0.0
3092.0.1
3092.1.0
3092.1.1
4065.0.0
4065.0.1
4065.1.1

This way, we can start using the latest features as they arrive for ST4. We then specify the minimum supported ST build in the package control entry:

```js
{
  "details": "https://github.com/sublimelsp/lsp",
  "releases": [{
      "sublime_text": "3092 - 4065",
      "tags": "3092."
    }, {
      "sublime_text": "4065 - 4111",  // imaginary 4111 build with new features
      "tags": "4065."
    }, {
      "sublime_text": ">=4111",
      "tags": "4111."
    }
  ]
}
```

Users not using those latest builds will be left with older releases.

rwols:
Note: sublime.ListInputHandler first appeared in build 3154. I believe that's the minimum ST build version required for this plugin to work correctly right now.

predragnikolic:
[...]

PR #931, opened 2020-03-28
Road to selectors

### LINKED ISSUE #540: Is the "syntaxes" key really needed?

rwols:
Remove the syntaxes key, thanks to the new `sublime.list_syntaxes()` API function. This function is available only in ST4.

This *will* break some LSP-* packages. Also some LSP-* that are not in the sublimelsp org, e.g. metals-sublime. In any case I have prepared diffs for all LSP-* packages, even metals-sublime.

<hr>

In many cases, a single scope of "source.${languageId}" is sufficient to match against a file with ${languageId}.

Examples are:

ruby -> source.ruby
c -> source.c
erlang -> source.erlang
json -> source.json
dart -> source.dart
d -> souce.d
(many more)

Non-examples are:

javascript -> source.js
cpp -> source.c++
html -> text.html.basic
vue -> text.html.vue

This cleans up LSP.sublime-settings even more.

Questions:
- When running two langservers in one file, is the right one picked for certain features like completion/signatureHelp?

close #540


rchl:
> This _will_ break some LSP-* packages.

How so?

plugin/core/configurations.py
```diff
@@ -75,10 +34,14 @@ def apply_project_overrides(client_config: ClientConfig, lsp_project_settings: d
...
-        if config_supports_syntax(config, syntax):
-            return True
+def is_supported_syntax(syntax: str, configs: Iterable[ClientConfig]) -> bool:
+    try:
+        scope = syntax2scope(syntax)
+        for config in configs:
+            if config.supports(scope):
+                return True
+    except StopIteration:
+        debug(syntax, "is not in sublime.list_syntaxes()")
```
rchl: I'm getting "million" of those:
```
LSP: Packages/Default/Find Results.hidden-tmLanguage is not in sublime.list_syntaxes()
```
Is it really useful to print those?
If you think so then I guess it might work to exclude the `hidden-*` ones.

rwols: I've removed the message.

rwols:
> How so?

I have modified the constructors of ClientConfig and LanguageConfig. It would break e.g.

https://github.com/scalameta/metals-sublime/blob/e7b17f66740bd7793020bd23788eb171b6df83cd/plugin.py#L52-L55

The fix is simple though. Use `LanguageConfig("scala")` instead. `source.scala` will be auto-deduced if you don't provide scopes.

The LSP-* packages in this organisation don't use LanguageConfig directly, so they won't break.

rchl:
Since it will take a while until ST3 can be deprecated, I have to ask: would it be a lot of work to retain support for syntaxes but use the new system if syntaxes are not provided?

I know that nobody likes maintaining two codebases and having to deal with old code but maybe it's worth doing it for this case? If the logic could be separated into its own file maybe, so it wouldn't get in the way?

ayoub-benali:
Just to make sure I understood, once this PR is in Master then LSP will be working only with ST4 ?

jwortmann:
If I understand it correctly, everything that requires ST4 (like this PR) will be merged into the st4000-exploration branch only. Then different tags will be used for master and this branch, so that new releases of the LSP package can be made for ST3 and ST4 separately, see #867.

rwols:
I went one step further and replaced the notion of `scopes` with the notion of `selector`.
It is a standard concept and [has documentation](https://www.sublimetext.com/docs/3/selectors.html).

What you can describe with selectors is a superset of what you can describe with an array of `scopes`, since if `["a", "b"]` is the array of scopes, then the equivalent selector is `"a | b"`.

The idea about leaving out the selector still applies: if the languageId happens to coincide with the selector you can leave out the selector.

An example:

We want to start the Vetur language server when Vue files are opened.
LSP languageId: `vue`
selector: `text.html.vue`

We want to start erlang_ls when Erlang files are opened.
LSP languageId: `erlang`
selector: not needed, we're targeting `source.erlang` if we leave it out.

rwols:
> Just to make sure I understood, once this PR is in Master then LSP will be working only with ST4 ?

Yes.

> If I understand it correctly, everything that requires ST4 (like this PR) will be merged into the st4000-exploration branch only. Then different tags will be used for master and this branch, so that new releases of the LSP package can be made for ST3 and ST4 separately, see #867.

The tags are both from `master`. Both tags point to the same commit.
Once st4000-exploration is in master, we'll stop supporting the `3154` tags and continue only with the `4070` tags (these numbers are the ST build numbers).

jwortmann:
> The tags are both from `master`. Both tags point to the same commit.
> Once st4000-exploration is in master, we'll stop supporting the `3154` tags and continue only with the `4070` tags (these numbers are the ST build numbers).

What is the reason for the `3154-` and `4070-` prefixes then? I thought the purpose in #867 was this:
> How about adding a new tag for ST4 on package control?
> That would allow people who use ST4 to test the st4000-exploration branch :)

The `st4000-exploration` branch cannot be merged into `master` until there is a stable release of ST4, and when dropping support for ST3 completely, wouldn't it be easier to just create a `st3-` tag for the lastest release compatible with ST3, instead of having to create two tags for each new release until a stable ST4 is publicly available?

rchl:
I also thought that `4070-` tags were meant to be pointing at `st4000-exploration` branch.

And I also thought that `st3-` and `st4-` tags would be sufficient.

Current tags, if used the way I thought they would, would be a little bit more flexible because you could "up" the tag when new version of ST4 introduces new feature or breaking change but in that case we could we shouldn't need to care for older versions of ST4 so we could just keep using `st4-` tags.

rwols:
> I also thought that 4070- tags were meant to be pointing at st4000-exploration branch.

OK, I'll point the next release for `4070` at the st4000-exploration branch and we'll see what happens.

> And I also thought that st3- and st4- tags would be sufficient.

My prediction is that ST4 will receive more useful API upgrades in the future. We want to make use those features as soon as possible. So if build 4132 has an important API update, we'll update package control again by adding tag a prefix `4132-` and users can enjoy the new features asap.

The reason I'm doing this is to not break users that stay with old ST versions.

rchl:
> we'll update package control again by adding tag a prefix `4132-` and users can enjoy the new features asap.

Fine :)

plugin/core/types.py
```diff
@@ -65,25 +77,30 @@ def __init__(self,
...
         self.init_options = init_options
         self.settings = settings
         self.env = env
         self.experimental_capabilities = experimental_capabilities
 
+    def supports(self, base_scope: str) -> bool:
+        for language in self.languages:
+            if language.match(base_scope):
+                return True
+        return False
```
rchl, line 92: I'm all for reducing vertical space usage. Can we use `any`?
```py
return any([language.match(base_scope) for language in self.languages])
```

rchl: I suppose that's probably less efficient though as it needs to build the list first.

TheSecEng: @rchl in this case, I don't think efficiency would be an issue, I can't imagine that `self.languages` would ever be big enough to cause a performance impact.

predragnikolic: `return any([language.match(base_scope) for language in self.languages])`

such a wonderful one __line__ of code :)

rwols: I believe one can simply omit the `[` and `]` to make it an anonymous lazy generator and not create a list, making it as "fast" as a for loop.

rwols:
The user will see lines like this in the console now:

```
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/OpenGL Shading Language (GLSL)/GLSL.sublime-syntax'] was deduced to "source.glsl"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/Clojure/Clojure.sublime-syntax'] was deduced to "source.clojure"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/YAML/YAML.sublime-syntax'] was deduced to "source.yaml"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/C++/C.sublime-syntax'] was deduced to "source.c"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/C++/C++.sublime-syntax'] was deduced to "source.c++"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/Objective-C/Objective-C.sublime-syntax'] was deduced to "source.objc"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/Objective-C/Objective-C++.sublime-syntax'] was deduced to "source.objc++"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/C++/C++.sublime-syntax'] was deduced to "source.c++"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/C#/C#.sublime-syntax'] was deduced to "source.cs"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/Lua/Lua.sublime-syntax'] was deduced to "source.lua"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/Python/Python.sublime-syntax'] was deduced to "source.python"
LSP: LanguageHandler instantiation crashed! __init__() takes from 2 to 3 positional arguments but 4 were given
LSP: received config lsp-css from LspCssPlugin
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/CSS/CSS.sublime-syntax', 'Packages/CSS3/CSS3.sublime-syntax'] was deduced to "source.css"
LSP: received config lsp-dockerfile from LspDockerfilePlugin
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/Dockerfile Syntax Highlighting/Syntaxes/Dockerfile.sublime-syntax'] was deduced to "source.dockerfile"
LSP: received config lsp-elm from LspElmPlugin
LSP: received config lsp-eslint from LspEslintPlugin
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/Babel/JavaScript (Babel).sublime-syntax', 'Packages/JavaScript/JavaScript.sublime-syntax', 'Packages/JavaScript/JavaScript.sublime-syntax', 'Packages/User/JS Custom/Syntaxes/React.sublime-syntax', 'Packages/Vue Syntax Highlight/Vue Component.sublime-syntax'] was deduced to "text.html.vue|source.js"
LSP: received config lsp-html from LspHtmlPlugin
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/HTML/HTML.sublime-syntax', 'Packages/PHP/PHP.sublime-syntax'] was deduced to "text.html.basic|embedding.php"
LSP: received config lsp-intelephense from LspIntelephensePlugin
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/PHP/PHP.sublime-syntax'] was deduced to "embedding.php"
LSP: received config lsp-json from LspJSONPlugin
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/JavaScript/JSON.sublime-syntax', 'Packages/JSON/JSON.sublime-syntax', 'Packages/PackageDev/Package/Sublime Text Build System/Sublime Text Build System.sublime-syntax', 'Packages/PackageDev/Package/Sublime Text Color Scheme/Sublime Text Color Scheme.sublime-syntax', 'Packages/PackageDev/Package/Sublime Text Commands/Sublime Text Commands.sublime-syntax', 'Packages/PackageDev/Package/Sublime Text Completions/Sublime Text Completions.sublime-syntax', 'Packages/PackageDev/Package/Sublime Text Keymap/Sublime Text Keymap.sublime-syntax', 'Packages/PackageDev/Package/Sublime Text Macro/Sublime Text Macro.sublime-syntax', 'Packages/PackageDev/Package/Sublime Text Menu/Sublime Text Menu.sublime-syntax', 'Packages/PackageDev/Package/Sublime Text Mousemap/Sublime Text Mousemap.sublime-syntax', 'Packages/PackageDev/Package/Sublime Text Project/Sublime Text Project.sublime-syntax', 'Packages/PackageDev/Package/Sublime Text Settings/Sublime Text Settings.sublime-syntax', 'Packages/PackageDev/Package/Sublime Text Theme/Sublime Text Theme.sublime-syntax'] was deduced to "source.json"
LSP: received config lsp-lemminx from LspXMLPlugin
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/PackageDev/Package/Property List/Property List.sublime-syntax', 'Packages/PackageDev/Package/Sublime Text Snippet/Sublime Text Snippet.sublime-syntax', 'Packages/PackageDev/Package/TextMate Preferences/TextMate Preferences.sublime-syntax', 'Packages/XML/XML.sublime-syntax'] was deduced to "text.xml"
LSP: received config lsp-typescript from LspTypeScriptPlugin
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/JavaScript/JavaScript.sublime-syntax', 'Packages/Babel/JavaScript (Babel).sublime-syntax'] was deduced to "source.js"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/TypeScript Syntax/TypeScript.tmLanguage'] was deduced to "source.ts"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/TypeScript Syntax/TypeScriptReact.tmLanguage'] was deduced to "source.tsx"
LSP: received config lsp-vue from LspVuePlugin
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/Vue Syntax Highlight/Vue Component.sublime-syntax'] was deduced to "text.html.vue"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/OpenGL Shading Language (GLSL)/GLSL.sublime-syntax'] was deduced to "source.glsl"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/Clojure/Clojure.sublime-syntax'] was deduced to "source.clojure"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/YAML/YAML.sublime-syntax'] was deduced to "source.yaml"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/C++/C.sublime-syntax'] was deduced to "source.c"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/C++/C++.sublime-syntax'] was deduced to "source.c++"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/Objective-C/Objective-C.sublime-syntax'] was deduced to "source.objc"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/Objective-C/Objective-C++.sublime-syntax'] was deduced to "source.objc++"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/C++/C++.sublime-syntax'] was deduced to "source.c++"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/C#/C#.sublime-syntax'] was deduced to "source.cs"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/Lua/Lua.sublime-syntax'] was deduced to "source.lua"
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/Python/Python.sublime-syntax'] was deduced to "source.python"
```

Among them:
```
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/Babel/JavaScript (Babel).sublime-syntax', 'Packages/JavaScript/JavaScript.sublime-syntax', 'Packages/JavaScript/JavaScript.sublime-syntax', 'Packages/User/JS Custom/Syntaxes/React.sublime-syntax', 'Packages/Vue Syntax Highlight/Vue Component.sublime-syntax'] was deduced to "text.html.vue|source.js"
```
This config comes from LSP-eslint. It apparently wants to start in `text.html.vue` files and `source.js` files. This is correct I think.

rwols:
Also, PHP.sublime-syntax apparently uses `embedding.php` as base scope, a bit of a weird choice.

rchl:
I don't fully understand those changes as I haven't had a proper look yet.

That said maybe the new `NpmClientHandler` abstraction will make the transition smoother? We can have a code there that automatically removes syntaxes from config when using certain build (Sublime or LSP). And if necessary, have other logic to specific to new code since things are quite abstracted now.

plugin/core/configurations.py
```diff
@@ -112,26 +73,29 @@ def is_supported(self, view: Any) -> bool:
...
+    def syntax_config_languages(self, view: sublime.View) -> Dict[str, LanguageConfig]:
+        scope = view2scope(view)
         config_languages = {}
         for config in self.all:
             if config.enabled:
-                language = syntax_language(config, syntax)
-                if language:
-                    config_languages[config.name] = language
+                for language in config.languages:
+                    if language.score(scope) > 0:
```
rchl, line 97: Can we use `match` here? The difference would be that this can return True also for `1` but I don't think that case is possible because we can not have empty selector due to this logic:
```
self.selector = selector if selector else "source.{}".format(self.id)
```

(We could then even make `score` method private as it wouldn't be needed anymore, I think.)

plugin/core/windows.py
```diff
@@ -96,7 +96,7 @@ def __init__(self, sublime: Any, settings: Settings) -> None:
         self._settings = settings
 
     def for_window(self, window: WindowLike, workspace: ProjectFolders,
-                   configs: ConfigRegistry) -> DocumentHandler:
+                   configs: WindowConfigManager) -> DocumentHandler:
         return WindowDocumentHandler(self._sublime, self._settings, window, workspace, configs)
```
rchl, line 100: Wrong type here.

rchl:
Another tricky example:
`LSP-html` languages:
```js
	"languages": [
		{
			"languageId": "html",
			"scopes": [
				"text.html.basic",
			],
			"syntaxes": [
				"Packages/HTML/HTML.sublime-syntax",
				"Packages/PHP/PHP.sublime-syntax",
			],
		},
	],
```
And we get this warning:
```
DEPRECATION: "syntaxes" is deprecated, use "selector" instead. The selector for ['Packages/HTML/HTML.sublime-syntax', 'Packages/PHP/PHP.sublime-syntax'] was deduced to "embedding.php|text.html.basic"
```
So the selector becomes `embedding.php|text.html.basic`. While we want the server to start for both PHP and HTML, we don't want to have PHP's scope trigger completions or other LSP features.

rwols:
> While we want the server to start for both PHP and HTML, we don't want to have PHP's scope trigger completions or other LSP features.

If you don't want vscode-html to "trigger completions" or "other LSP features", exactly what should it do in PHP files, then?

rchl:
> > While we want the server to start for both PHP and HTML, we don't want to have PHP's scope trigger completions or other LSP features.
> 
> If you don't want vscode-html to "trigger completions" or "other LSP features", exactly what should it do in PHP files, then?

It's supposed to trigger in HTML blocks embedded in PHP:

Screenshot 2020-05-03 at 15 19 23

rwols:
Doesn't intelephense already handle this?

rchl:
I think that Intelephense doesn't do anything with HTML. But that's not the point. I'm just reporting a potential issue with implementation. :)

rwols:
OK, I'll figure out what should be done. This PR needs more work :)

rwols:
@rchl I believe commit e43b1c1 solves your problem.

`"scopes"` key is replaced with `"feature_selector"`. Using `text.html` would let an html language server "win" over a PHP language in embedded HTML in PHP.

So now we have:

- `document_selector` (replacement for `syntaxes`)
- `feature_selector` (replacement for `scopes`)

rchl:
Just a note that casing is a bit inconsistent. We have many properties using underscores but then we also have `languageId` and `initializationOptions`.

rchl:
I believe latest changes are fixing my issue as it looks like we trigger completions using lsp-html now. But I've also noticed that with this configuration:
```js
	"languages": [
		{
			"languageId": "html",
			"document_selector": "text.html",
			// "feature_selector": "text.html",
		},
	],
```
we are initializating lsp-html on *.php file...

rwols:
Are you sure you don't have a `syntaxes` in your User/LSP.sublime-settings for this server?

rchl:
> Are you sure you don't have a `syntaxes` in your User/LSP.sublime-settings for this server?

I don't.
Both `document_selector` and `feature_selector` are set to `text.html` (when logging in `ClientConfig`.

I think it's due to one of the calls to `match_document` that pass view scope instead of syntax scope.

rwols:
> I think it's due to one of the calls to match_document that pass view scope instead of syntax scope.

Ah yes. How stupid of me :)

rwols:
I consider this pretty much ready. Any comments? If you switch back and forth between `master` and `st4000-exploration`, you will have to live with the deprecation warnings for a while until ST4 gets a dev release.

rchl:
I'll have a deeper look shortly.

plugin/code_actions.py
```diff
@@ -75,7 +75,7 @@ def request_code_actions_with_diagnostics(view: sublime.View, diagnostics_by_con
 
     actions_at_location = CodeActionsAtLocation(actions_handler)
 
-    for session in sessions_for_view(view, point):
+    for session in sessions_for_view(view):
```
rchl, line 78: Why this change?

plugin/color.py
```diff
@@ -74,7 +73,7 @@ def send_color_request(self) -> None:
         if is_transient_view(self.view):
             return
 
-        client = client_from_session(session_for_view(self.view, 'colorProvider'))
+        client = client_from_session(session_for_view(self.view, 'colorProvider', self.view.sel()[0].b))
```
rchl, line 76: Is this a correct change?
I would think that the `documentColor` request should be sent regardless of where the cursor is.

plugin/completion.py
```diff
@@ -123,7 +123,7 @@ def is_applicable(cls, view_settings: dict) -> bool:
 
     def initialize(self) -> None:
         self.initialized = True
-        session = session_for_view(self.view, 'completionProvider')
+        session = session_for_view(self.view, 'completionProvider', self.view.sel()[0].b)
```
rchl, line 126: Same as in color.py - is this a correct change?
`initialize` is called once per view. Is it correct to base the result here on where the cursor is?

plugin/code_actions.py
rwols: Can you give an example where it matters that the sessions are sorted by their feature_selector / scopes?

plugin/color.py
rwols: This always happened behind the curtains. I just made it explicit. Read this carefully:

https://github.com/sublimelsp/LSP/blob/10c4004be55dd54ff61473093f03e2a90ca867ca/plugin/core/configurations.py#L15

rchl: Oh, OK, I see now.

Previous logic used `region.begin()` rather than `region.b` though. I'm not sure which one would be more correct to honest.

plugin/core/settings.py
```diff
@@ -157,53 +157,78 @@ def unload_settings() -> None:
...
+        return list(map(read_language_config, client_config.get("languages", [])))
+    if "languageId" in client_config:
+        return [read_language_config(client_config)]
+    return []
 
 
-def read_client_config(name: str, client_config: Dict) -> ClientConfig:
-    languages = read_language_configs(client_config)
-
+def read_client_config(name: str, d: Dict[str, Any]) -> ClientConfig:
```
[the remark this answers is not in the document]
rwols: It's used in LSP-* packages so we can't :)

plugin/core/settings.py
```diff
@@ -157,53 +157,78 @@ def unload_settings() -> None:
         _settings_obj.clear_on_change("_on_new_client_settings")
 
 
-def read_language_config(config: dict) -> LanguageConfig:
-    language_id = config.get("languageId", "")
-    scopes = config.get("scopes", [])
-    syntaxes = config.get("syntaxes", [])
-    return LanguageConfig(language_id, scopes, syntaxes)
+def convert_syntaxes_to_selector(d: Dict[str, Any]) -> Optional[str]:
```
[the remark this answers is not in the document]
rwols: Because there's two ways to specify languages in JSON. One way is to put languageId/syntaxes/document_selector at the same level as command/enabled/env. Other way is to have languages array.

plugin/core/windows.py
```diff
@@ -119,7 +118,7 @@ def update(self, version: int, changes: Iterable[sublime.TextChange]) -> None:
 
 class WindowDocumentHandler(object):
```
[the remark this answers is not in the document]
rwols: I have plans for this :)

plugin/core/registry.py
```diff
@@ -138,10 +133,10 @@ def is_visible(self, event: Optional[dict] = None) -> bool:
         return is_supported_view(self.view)
 
     def has_client_with_capability(self, capability: str) -> bool:
-        return session_for_view(self.view, capability) is not None
+        return session_for_view(self.view, capability, self.view.sel()[0].b) is not None
```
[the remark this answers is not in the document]
rwols: See comment about curtains.

plugin/completion.py
rwols: This is deliberate to make it obvious that we are actually doing this behind the curtains.

plugin/code_actions.py
rwols: From a higher POV:

session_for_view and sessions_for_view are god-functions. I would like to eventually get rid of them. They won't be necessary once we have moved to the one true ViewEventListener design (if that ever happens). I'll make a discussion issue about my thoughts and future lookahead.

rchl: I thought that this will exclude sessions that don't match scope at point (so, for example, HTML server running in PHP file where HTML server would only match in HTML blocks). But that's not how it worked so I think I'm fine with this change.

plugin/completion.py
rchl: But then this branch is currently confused about that as the point argument was removed in session(s)_for_view. What's the ultimate goal then?

rwols: - session_for_view: Give me the best matching session for the current point and view
- sessions_for_view: Give me all sessions for the view

rchl: No, I mean that the `point` argument is no longer in those but then you are adding explicit `point` argument to the callers.

So I'm not sure what your intention is. To remove argument and make point implicit or add it everywhere.

BTW. If we are adding `sel()[0].being()` everywhere then please have a shared helper for it somewhere as it's a lot of repetition and also this should be null-checking selection.

plugin/color.py
rwols: `region.b` is better because that's the cursor position. I made the `point` arg optional.

plugin/completion.py
rwols: I made the `point` arg optional.

plugin/core/configurations.py
```diff
@@ -58,28 +18,26 @@ def apply_project_overrides(client_config: ClientConfig, lsp_project_settings: d
...
 
 
-def is_supported_syntax(syntax: str, configs: List[ClientConfig]) -> bool:
-    for config in configs:
-        if config_supports_syntax(config, syntax):
-            return True
+def is_supported_syntax(syntax: str, configs: Iterable[ClientConfig]) -> bool:
+    scope = syntax2scope(syntax)
+    if scope is not None:
+        return any(config.match_document(scope) for config in configs)
```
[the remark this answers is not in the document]
rwols: This would leak optional handling into primitive functions, something I want to avoid.

plugin/core/configurations.py
```diff
@@ -1,51 +1,11 @@
...
-
-    return (config_score[0] for config_score in sorted(
-        scope_configs, key=lambda config_score: config_score[1], reverse=True))
-
-
-def get_global_client_config(view: sublime.View, global_configs: List[ClientConfig]) -> Optional[ClientConfig]:
-    return get_scope_client_config(view, global_configs)
-
-
 def create_window_configs(window: WindowLike, global_configs: List[ClientConfig]) -> List[ClientConfig]:
```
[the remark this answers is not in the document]
rwols: Made a PR for it https://github.com/sublimelsp/LSP/pull/1036

plugin/core/windows.py
rwols: The DocumentHandler is a Protocol. The WindowDocumentHandler fits that Protocol.

plugin/color.py
predragnikolic: > region.b is better because that's the cursor position

I always wondered what the exact cursor position is. And now I know

plugin/code_actions.py
```diff
@@ -72,10 +72,8 @@ def request_code_actions(view: sublime.View, point: int,
 def request_code_actions_with_diagnostics(view: sublime.View, diagnostics_by_config: Dict[str, List[Diagnostic]],
                                           point: int, actions_handler: Callable[[CodeActionsByConfigName], None]
```
rchl, line 73: `point` is unused now

PR #932, opened 2020-03-28
Implement incremental text sync

### LINKED ISSUE #437: Feature request: Support incremental mode
Here you're passing the whole document in one single change:
https://github.com/tomv564/LSP/blob/f5a4dcfbe117ddf1a07e1c971e8b14ca7e311c00/plugin/core/windows.py#L292-L294

It would be awesome to support incremental changes (if the language server also supports it).

Is grabbing those diffs supported by Sublime's API the way it is in VS Code?

rwols:
This can break pyls in subtle ways.
clangd seems to be handling the incremental changes correctly.
eslint and vetur can only do full text sync.

The reason pyls breaks is that it doesn't seem to handle UTF-16 character offsets correctly: https://github.com/palantir/python-language-server/blob/21833eafd4e254a79d35ca539967f3a010e8c2a1/pyls/workspace.py#L150-L191

It is stated in the LSP spec that character offsets are UTF-16: https://microsoft.github.io/language-server-protocol/specification#textDocuments

This diff uses the UTF-16 character offsets provided in sublime.HistoricPosition.

That means that if you have an emoji in your source code the incremental sync can break (if the langserver assumes character offsets are in terms of unicode code points)

close #437


tests/test_single_document.py
```diff
@@ -58,13 +58,28 @@ def test_did_close(self) -> 'Generator':
...
+                {'range': {'start': {'line': 0, 'character': 2}, 'end': {'line': 0, 'character': 2}}, 'text': '\n'},
+                {'range': {'start': {'line': 1, 'character': 0}, 'end': {'line': 1, 'character': 0}}, 'text': '🙂'},
+                # Note that this is character offset (2) is correct (UTF-16).
+                {'range': {'start': {'line': 1, 'character': 2}, 'end': {'line': 1, 'character': 2}}, 'text': '\n'},
+                {'range': {'start': {'line': 2, 'character': 0}, 'end': {'line': 2, 'character': 0}}, 'text': 'D'}],
+            'textDocument': {
+                'version': self.view.change_count(),
+                'uri': filename_to_uri(TEST_FILE_PATH)
+            }
+        })
```
rwols: This is an important test: it verifies that the character offset for the second newline char is **two**, not **one**.

AmjadHD:
> The reason pyls breaks is that it doesn't seem to handle UTF-16 character offsets correctly:

you should open an issue there.

rwols:
I tested intelephense. Intelephense supports incremental text sync. As it turns out, the textDocument/documentHighlight feature is a good way to test whether things are working as intended :)

In the screenshot below I'm using intelephense and I've added a line with a beer emoji in it. This emoji encompasses **two** UTF-16 points. The textDocument/documentHighlight feature shows that the line under `$uri` is shifted by 1 to the right, compared to the other `$uri` underlines.

Screenshot from 2020-03-28 22-39-07

As it turns out we have to use UTF-16 in the `point_to_offset` and `offset_to_point` functions too. This puts the line under `$uri` in the correct spot.

Screenshot from 2020-03-28 22-43-50

Knowing this, I'm pretty sure we're on the right track using UTF-16.

plugin/core/views.py
```diff
@@ -25,16 +25,14 @@ def get_line(window: Optional[sublime.Window], file_name: str, row: int) -> str:
...
-    return view.text_point(
-        point.row,
-        # @see https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#position
-        # If the character value is greater than the line length it defaults back to the line length.
-        min(point.col, len(view.line(view.text_point(point.row, 0))))
-    )
+    # TODO: How can we clamp the column in UTF-16?
+    # @see https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#position
+    # If the character value is greater than the line length it defaults back to the line length.
+    return view.text_point_utf16(point.row, point.col)
```
rwols: @jfcherng @tomv564 Unfortunately I couldn't find a way to clamp the column anymore in UTF-16. Perhaps we need to request a new ST API function for this.

rwols:
clangd works with UTF-16 too.

Screenshot from 2020-03-29 13-58-39

Issue #927, opened 2020-03-24
Inflexible API for language handlers


rchl:
Related to #897 and #926

`LanguageHandler` classes can implement `on_start` and `on_initialized` methods and it is enough for most servers but some need access to extra things which are not currently available:
 - `LSP-intelephense` needs access to Sublime window as it uses `window.extract_variables()` to get window variables. Currently it's getting window using `sublime.active_window()` but that is error-prone as there is no guarantee that active_window is the correct one.
 - `LSP-eslint` would need access to `WorkspaceFolders` in `on_initialized` as it needs to reply to `workspace/configuration` with workspace that matches the current file .
 - Tom is adding `can_start` method which is basically like `on_start` (both can prevent server from starting) but with different arguments 

I'd like us to have some flexible and extensible solution for these issues where we don't have to basically duplicate methods to add new arguments to make API backward-compatible.

tomv564:
As we discussed in #897:
* Duplicating methods to add arguments is pretty ugly
* We can break the APIs once to create future-compatible APIs now.
* In practice people create ServerHandlers, not LanguageHandlers.

**If we create a new ServerHandler API, what do we take from LanguageHandler?**

* `on_start` sounds like an event handler but it actually decides if sessions can be started. I think `can_start` better reflects what the call is capable of.
  * Should we move all arguments into a CanStartArgs dict?
  * Is a list of startable folders a good response? Or move this into a dict as well? How can we easily represent yes/no/dontcare in this dict-form?
* `on_initialized` is still an event handler, is it a good name?
  * What do we want to add to an InitializedArgs dict?

**How did configuration management work out?**

This is in relation to the `config` member. I see packages like lsp-intelephense invented their own settings structure. It's unclear to me if those packages handled eg. project-specific overrides correctly and if it is all worth it?

Also there are requests to remove default configs for eg. eslint (even though endusers still may have references to that config in user settings).

Is there something about the way `config` is provided via LanguageHandler that's causing pain? Now is a good time to fix that as well.

rchl:
> We can break the APIs once to create future-compatible APIs now.

If we could keep both APIs alive for a while, that would make the transition smoother.

> * I think `can_start` better reflects what the call is capable of.

That name sounds good to me.

> * Is a list of startable folders a good response?

Can we move that discussion to #926? I'd like to understand better the problem it's solving and how it relates to workspaces support. I'll raise a question there.

> `on_initialized` is still an event handler, is it a good name?
[...]

PR #937, opened 2020-03-31
Add more specific selector for some trigger characters.

### LINKED ISSUE #933: Keep using trigger characters, or start using selectors?

Animation

Settings are

```
	"auto_complete": true,
```

_Originally posted by @deathaxe in https://github.com/sublimelsp/LSP/pull/866#issuecomment-603466761_

predragnikolic:
closes #933 
For now this will only be added for the `"` char, with the selector punctuation.definition.string.begin.

If we notice more of this cases where the AC is triggering forever,
we could either assign a special selector like we did here,
or fill an issue to the language server that it shouldn't return completions in that case.

That is also an option worth considering.

The server could specify not to return completions when inside comments. (just giving elm language server as an example :) ) https://github.com/elm-tooling/elm-language-server/issues/238

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -171,29 +171,15 @@ class CompletionHandler(LSPViewEventListener):
             # This is to make ST match with labels that have a weird prefix like a space character.
             self.view.settings().set("auto_complete_preserve_order", "none")
 
-    def _view_language(self, config_name: str) -> Optional[str]:
-        languages = self.view.settings().get('lsp_language')
-        return languages.get(config_name) if languages else None
-
     def register_trigger_chars(self, session: Session, trigger_chars: List[str]) -> None:
-        completion_triggers = self.view.settings().get('auto_complete_triggers', []) or []  # type: List[Dict[str, str]]
-        view_language = self._view_language(session.config.name)
-        if view_language:
-            for language in session.config.languages:
-                if language.id == view_language:
-                    for scope in language.scopes:
-                        # debug("registering", trigger_chars, "for", scope)
-                        scope_trigger = next(
-                            (trigger for trigger in completion_triggers if trigger.get('selector', None) == scope),
-                            None
-                        )
-                        if not scope_trigger:  # do not override user's trigger settings.
-                            completion_triggers.append({
-                                'characters': "".join(trigger_chars),
-                                'selector': scope
-                            })
+        completion_triggers = self.view.settings().get('auto_complete_triggers') or []  # type: List[Dict[str, str]]
 
-            self.view.settings().set('auto_complete_triggers', completion_triggers)
+        completion_triggers.append({
+            'characters': "".join(trigger_chars),
+            'selector': "- comment - punctuation.definition.string.end"
+        })
+
+        self.view.settings().set('auto_complete_triggers', completion_triggers)
 
     def on_query_completions(self, prefix: str, locations: List[int]) -> Optional[sublime.CompletionList]:
         if not self.initialized:


plugin/completion.py
```diff
@@ -136,24 +136,35 @@ def _view_language(self, config_name: str) -> Optional[str]:
...
+        )
+
+        if not existing_scope_trigger:  # do not override user's trigger settings.
+
+            def selector_for_char(char, default_scope):
+                # to avoid tiggering AC forever on commit_completion
+                # some trigger chars need to be provided a better selector
+                char_scopes = {
+                    # fixes https://github.com/sublimelsp/LSP/issues/933
+                    "\"": "punctuation.definition.string.begin"
```
rwols, line 156: While I might agree that using selectors is the way forward, this particular implementation looks hastily written. I don't want a repeat of the previous situation where we added many workarounds for text edits. It seems like less functionality, but somehow we have more lines of code.

For a starter, I don't understand why `selector_for_char` needs to be a lambda. And it has no type hints.

Consider discussing this change in #927.

Ideally a `ServerHandler` (`LanguageHandler`?) would provide this base package with the necessary selectors.

predragnikolic: > I don't want a repeat of the previous situation where we added many workarounds for text edi

I agree. This code here can easily get out of control an become something like this:
https://github.com/sublimelsp/LSP/blob/24cee140fd7d02a29e82a139b918bef77d89f6fb/plugin/core/completion.py#L39-L50

As for the type hints, I haven't installed pyls-mypy correctly (I installed it, but for some reason it doesn't work, that is the reason I missed to add type hints on some places).

>Ideally a ServerHandler (LanguageHandler?) would provide this base package with the necessary selectors.

Moving this logic to a LanguageHandler/ServerHandler sounds like a good way to go. :)

I am gonna close this PR until we first discuss more about the possible solutions.

rwols:
Can we re-open this?

Perhaps we can try removing the trigger chars altogether, and maintaining a dict for now.

predragnikolic:
Sure,
can you be just more specific about this " try removing the trigger chars altogether, and maintaining a dict for now"

What are your intentions. To not use the trigger chars returned from the server?

predragnikolic:
I understand that relaying 100% on trigger_chars can leads to bad UX in some cases #933,
but nor relaying on trigger_chars at all and manually adding scopes for each(n) language seems like something I would like to avoid :)

Can we have a balance of both?
The intention of this PR was to use trigger chars, 
but assign special scopes to those chars that can cause bad UX.

I just think that is too much work to write a selector for each scope.

plugin/completion.py
```diff
@@ -171,29 +171,15 @@ def initialize(self) -> None:
...
-                                'selector': scope
-                            })
-
-            self.view.settings().set('auto_complete_triggers', completion_triggers)
+        completion_triggers = self.view.settings().get('auto_complete_triggers') or []  # type: List[Dict[str, str]]
+
+        completion_triggers.append({
+            'characters': "".join(trigger_chars),
+            'selector': "- comment - punctuation.definition.string.end"
+        })
```
rwols, line 180: This is starting to look good :)
The `auto_complete_triggers` list is stored in the .sublime-workspace file in case the user is using a .sublime-project, and otherwise in the .sublime-session file in $DATA/Local/session.sublime-session. Did you verify whether these changes are not adding an infinite amount of entries to the `auto_complete_triggers` list every time ST is restarted? The old code made sure that that didn't happen.

On that note, it's maybe good to rethink this too. ST ignores unknown keys in an auto_complete_trigger dict, so what we can do is add a `server` key like this:

```python
{
    'characters': "".join(trigger_chars),
    'selector': '- comment - punctuation.definition.string.end'
    'server': session.config.name
}
```

We can then remove our enrties in the `__del__` method of the listener:

```python

    def __del__(self) -> None:
        triggers = self.view.settings().get("auto_complete_triggers")
        triggers = [trigger for trigger in triggers if 'server' not in trigger]
        # put triggers back into auto_complete_triggers ...
```

predragnikolic: > adding an infinite amount of entries to the auto_complete_triggers list 

Short anwser:
To answer your question, it won't append the list to infinity,
so we don't need a `server` key in the dict.

Long answer:
That was one of the first things I checked. 
Each time you open for example a index.js file.
And you did not trigger the AC.
if you run in the sublime console this:
```
>>> view.settings().get('auto_complete_triggers')
[{'characters': '<', 'selector': 'text.html, text.xml'}, {'selector': 'punctuation.accessor'}]
```

If you trigger the AC, the output will be different
```
>>> view.settings().get('auto_complete_triggers')
[{'characters': '<', 'selector': 'text.html, text.xml'}, {'selector': 'punctuation.accessor'}, {'characters': '."\'/@<', 'selector': '- comment - punctuation.definition.string.end'}]
```
if you trigger the AC again you will get the same output as before.

If you open a new file `main.js`.
And you did not trigger the AC.
if you run in the sublime console this:
```
>>> view.settings().get('auto_complete_triggers')
[{'characters': '<', 'selector': 'text.html, text.xml'}, {'selector': 'punctuation.accessor'}]
```

If you trigger the AC, the output will be different
```
>>> view.settings().get('auto_complete_triggers')
[{'characters': '<', 'selector': 'text.html, text.xml'}, {'selector': 'punctuation.accessor'}, {'characters': '."\'/@<', 'selector': '- comment - punctuation.definition.string.end'}]
```

PR #938, opened 2020-04-01
Maintenance/merge master into st4000-exploration

rchl:

PR #939, opened 2020-04-01
Fix imports

predragnikolic:
Merged master and fix imports that caused problem AC.
When committing the completion, the the LspSelectCompletionItemCommand wasn't run
because it wasn't imported.


predragnikolic:
Fixed with #938

PR #941, opened 2020-04-02
Fix off-by-one error

KevinSjoberg:
In commit https://github.com/sublimelsp/LSP/commit/3c267eab898cf43778a1d05e6b582045b8812b4b a change was made to underline all regions for all severities of diagnostics. However `range` isn't inclusive at the upper limit and thus, hints are currently excluded.

### DIFF
--- a/plugin/diagnostics.py
+++ b/plugin/diagnostics.py
@@ -247,7 +247,7 @@ class DiagnosticViewRegions(DiagnosticsUpdateWalk):
         self._relevant_file = False
 
     def end(self) -> None:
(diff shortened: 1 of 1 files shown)


rchl:
There is existing PR for fixing this (in a bit more extensive way).

KevinSjoberg:
@rchl well, it introduces a setting that was previously removed. Further, the original commit explicitly set this to all since we already get hover diagnostics for all levels. The proposed change in #914 would need to take hover into account as well then to make the experience consistent.

Perhaps this change is enough for now?

rwols:
This diff is a bug fix instead of a bug fix + feature, so I'm inclined to merge this first. We can then discuss the feature in #914.

PR #943, opened 2020-04-04
Match behavior of VS Code when inserting TextEdit 

predragnikolic:
Match behavior of VS Code when inserting TextEdit on lines where the range text that is going to be deleted is not the same on each line.

Before this PR:
output

Not that the :red; got removed, because I tried to apply an edit region to a line that doesn't have the same text as the line the completion was requested for.

After:
st-mimic-vs-code

This PR matches how VS code handles this case:
code-bad

For more info, on why I had done this way,
you can read:
https://github.com/microsoft/language-server-protocol/issues/950#issuecomment-608814427

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -132,6 +132,7 @@ class LspSelectCompletionItemCommand(sublime_plugin.TextCommand):
 
             range = Range.from_lsp(text_edit['range'])
             edit_region = range_to_region(range, self.view)
+            edit_text = self.view.substr(edit_region)
 
             # calculate offset by comparing cursor position with edit_region.begin.
             # by applying the offset to all selections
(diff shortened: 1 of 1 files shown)


jwortmann:
I wonder whether VS Code's behaviour is optimal. Wouldn't it be more usefull to select the whole word with `self.view.word(selection_cursor)`, i.e. in your example to erase `border` instead of just `bor` for the second cursor?

One problem of that in CSS could be, that a dash is considered as a word-break (if I remember correctly).

predragnikolic:
> I wonder whether VS Code's behavior is optimal

I agree that is not optimal, 
but just keep in mind that this edge case should not happen vary often. 

>Wouldn't it be more usefull to select the whole word?

While it would work in most cases,
It wont work in cases like you said:
- CSS, where a dash is considered as a word-break
- in LSP-json:

| - cursor
```
"a|uto_complete_preserve_order": "none",
"a|uto_complete_include_snippets_when_typing": true,
```

if we select "animation_enabled" completion item.

with the current PR we get: (that is how VS Code does it)
a)
```
"animation_enabled": "none",
"animation_enabled"uto_complete_include_snippets_when_typing": true,
```

b)
with the approach you are proposing we will get:
```
"animation_enabled": "none",
"animation_enabled"": true, // note that the " is left from the previous line
```

I really don't know what is better, I would maybe prefer b) over a)
But I would like to ask for feedback from you and other people :)

@jfcherng @rchl @rwols

rwols:
In general, think about what is best for SublimeText. Don't copy an application that is trying to be a copy of SublimeText ;)

I would suggest to park this PR until we get the basics right.

predragnikolic:
I will close this for now.

PR #944, opened 2020-04-04
Move around some code in completion.py

rwols:
Because we have so little code now for the completion logic, it doesn't
make much sense anymore to keep around relevant code in various files.

- Move all the stuff in plugin/core/completion.py to plugin/completion.py
- Move all the stuff in restore_lines.py to completion.py
- Remove the tests in test_completion_core.py, I am confident in the
  functionality working as intended
- Streamline the tests in test_completion.py

So everything relevant to completion lives in completion.py now.
This diff changes nothing functionally.


tests/test_completion.py
```diff
@@ -236,172 +78,223 @@ def move_cursor(self, row: int, col: int) -> None:
...
+        yield commit_completion
 
     def read_file(self) -> str:
         return self.view.substr(sublime.Region(0, self.view.size()))
 
-    def test_simple_label(self) -> 'Generator':
-        self.set_response("textDocument/completion", completions_with_label)
-
-        self.type("a")
+    def verify(self, response: List[Dict[str, Any]], expected_text: str,
```
predragnikolic, line 99: What do you think about reordering the arguments?
instead of 
`response, expected_text, pre_insert_text`
to be: 
`response, insert_text, expected_text`?

And do you think that this:
```
self.verify([{'label': 'asdf', 'insertText': 'asdf()'}], 
            insert_text='a', 
            expected_text='asdf()')
```
is more easier to understand on first glance, than when you see this?
`self.verify([{'label': 'asdf', 'insertText': 'asdf()'}], 'asdf()', "a")`

plugin/completion.py
```diff
@@ -1,22 +1,124 @@
...
+    20: (sublime.KIND_ID_TYPE, "ε", "EnumMember"),
+    21: (sublime.KIND_ID_VARIABLE, "π", "Constant"),
+    22: (sublime.KIND_ID_TYPE, "s", "Struct"),
+    23: (sublime.KIND_ID_FUNCTION, "e", "Event"),
+    24: (sublime.KIND_ID_KEYWORD, "ο", "Operator"),
+    25: (sublime.KIND_ID_TYPE, "τ", "Type Parameter")
+}
+
+
+class RestoreLines:
```
predragnikolic, line 45: Great that you moved all the code related completions to completions.py.
 
But can you move back only the RestoreLines in the separate file,
just to keep the completions file a bit smaller, 
so other people don't get overwhelmed  by all the code in here?

rwols: Meh. 300 lines of Python code is not overwhelming.

PR #945, opened 2020-04-05
Set auto_complete_preserve_order to none

### LINKED ISSUE #935: [AC] AC not showing up with a "decorated" trigger
Note: check if only DYNAMIC_COMPLETIONS are affected.

When I supplement the completion by typing more characters, the completion widget goes away.

Peek 2020-03-28 13-19
<hr>
I don't get a completion widget at all sometimes. In the gif below I expect to at least see a completion for the "time" variable:

Peek 2020-03-28 13-31

I'm guessing this is due to incomplete completions. I'm not sure what the dynamic is here.

_Originally posted by @rwols in https://github.com/sublimelsp/LSP/pull/866#issuecomment-605441312_

rwols:
This makes the AC show up for labels that have an odd prefix like a space character (e.g. clangd).

close #935

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -225,8 +225,6 @@ class CompletionHandler(LSPViewEventListener):
             if trigger_chars:
                 self.register_trigger_chars(session, trigger_chars)
             # This is to make ST match with labels that have a weird prefix like a space character.
-            # It is unfortunately untestable as this problem will only show up when typing characters, not when
-            # running view.run_command('auto_complete') in a test. See: https://github.com/sublimelsp/LSP/issues/935
             self.view.settings().set("auto_complete_preserve_order", "none")
 
     def _view_language(self, config_name: str) -> Optional[str]:
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -224,6 +224,10 @@ class CompletionHandler(LSPViewEventListener):
                 'triggerCharacters') or []
             if trigger_chars:
                 self.register_trigger_chars(session, trigger_chars)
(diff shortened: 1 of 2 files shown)


plugin/completion.py
```diff
@@ -224,6 +224,10 @@ def initialize(self) -> None:
                 'triggerCharacters') or []
             if trigger_chars:
                 self.register_trigger_chars(session, trigger_chars)
+            # This is to make ST match with labels that have a weird prefix like a space character.
+            # It is unfortunately untestable as this problem will only show up when typing characters, not when
```
predragnikolic, line 228: Just wanted to say that we already have a test for the case where the label contains spaces.

https://github.com/sublimelsp/LSP/blob/87e98cf9d0a7d80f315dfb74d22b6c568a2a34b3/tests/test_completion.py#L237-L245

If the label is `'override def myFunction(): Unit'`.

You can type `def myF`.

1. If you trigger AC and auto_complete_preserve_order is "some":
You wont see the AC, because filtered out that completion item.

2. If you trigger AC and auto_complete_preserve_order is "none",
You will see the AC

Here is an example for that.
https://github.com/sublimelsp/LSP/blob/87e98cf9d0a7d80f315dfb74d22b6c568a2a34b3/tests/test_completion.py#L260-L261

Just a reminder,
I set for all completion tests auto_complete_preserve_order to "none"
https://github.com/sublimelsp/LSP/blob/87e98cf9d0a7d80f315dfb74d22b6c568a2a34b3/tests/setup.py#L175

rwols: Thanks, I had not noticed this was set up in the tests :)

PR #948, opened 2020-04-05
Resolve the promise on the main thread

rwols:
It is unclear to me whether sublime.CompletionList is thread-safe. It doesn't seem to be thread-safe when looking at the details in sublime_plugin.py and sublime.py. So just to be on the safe side, let us resolve the promise from the main thread.

By the way, I noticed ST freezing when using clangd, I don't experience freezes anymore with this change. I'm not 100% sure if it's related, but since there's no threading locks being used in sublime_plugin.py I'm pretty sure it's not thread safe.

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -196,6 +196,11 @@ class LspSelectCompletionItemCommand(sublime_plugin.TextCommand):
         sublime.status_message('Applied additional edits for completion')
 
 
+def resolve(completion_list: sublime.CompletionList, items: List[sublime.CompletionItem], flags: int = 0) -> None:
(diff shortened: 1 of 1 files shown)


predragnikolic:
I don't have a good knowledge about threads or how sublime handles set_completions internally.
so my review on this PR should not be valid.

The changes to me don't look complicated,
you only delay calling set_completions.

I also can't test if this change also fixes that freezing.
I don't use clangd and I hadn't experienced freezes with other language servers.

That said, 
I have nothing against merging this, if you think that this will improve the experience for clangd users

rwols:
> The changes to me don't look complicated, you only delay calling set_completions.

No, the callback handlers are ran on a py thread somewhere. But calling `sublime.set_timeout` makes the lambda run on the main thread. It is not as simple as a delay.

PR #949, opened 2020-04-05
WIP: Make clangd work for single cursor

rwols:
This seems to be working for pyls, clangd, intelephense and vsode-json-languageserver.

The multi-cursor support is broken at the moment. There is no need to store the line from the past. Bringing the edit region from the past to the present seems to be working okay.

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -42,56 +42,7 @@ completion_kinds = {
 }
 
 
-class RestoreLines:
-    def __init__(self) -> None:
-        self.saved_lines = []  # type: List[dict]
-
-    def save_lines(self, locations: List[int], view: sublime.View) -> None:
-        change_id = view.change_id()
-
-        for point in locations:
-            line = view.line(point)
-            change_region = (line.begin(), line.end())
-            text = view.substr(line)
-
-            self.saved_lines.append({
-                "change_id": change_id,
-                "change_region": change_region,
-                "text": text,
-                # cursor will be use retore the cursor the te exact position
-                "cursor": point
-            })
-
-    def to_dict(self) -> Dict[str, Any]:
-        return {
-            "saved_lines": self.saved_lines
-        }
-
-    @staticmethod
-    def from_dict(dictionary: Dict[str, Any]) -> 'RestoreLines':
-        restore_lines = RestoreLines()
-        restore_lines.saved_lines = dictionary["saved_lines"]
-        return restore_lines
-
-    def restore_lines(self, edit: sublime.Edit, view: sublime.View) -> None:
-        # restore lines contents
-        # insert back lines from the bottom to top
-        for saved_line in reversed(self.saved_lines):
-            change_id = saved_line['change_id']
-            begin, end = saved_line['change_region']
-            change_region = sublime.Region(begin, end)
-
-            transform_region = view.transform_region_from(change_region, change_id)
-            view.erase(edit, transform_region)
-            view.insert(edit, transform_region.begin(), saved_line['text'])
-
-        # restore old cursor position
-        view.sel().clear()
-        for saved_line in self.saved_lines:
-            view.sel().add(saved_line["cursor"])
-
-
-def format_completion(item: dict, restore_lines: RestoreLines) -> sublime.CompletionItem:
+def format_completion(item: dict, change_id: Any) -> sublime.CompletionItem:
     kind = sublime.KIND_AMBIGUOUS
 
     item_kind = item.get("kind")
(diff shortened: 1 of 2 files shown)


rwols:
Does anyone know an example of language server that sends snippets in `insertText` form instead of `textEdit` form?

predragnikolic:
No 🤔 but you can always hardcore the response from the server to add a Completion Item that is a snippet and has insertText

rwols:
This is done. I hope this'll be a good basis for more experimentation like #943.

plugin/completion.py
```diff
@@ -105,62 +56,52 @@ def format_completion(item: dict, restore_lines: RestoreLines) -> sublime.Comple
...
+            new_text = text_edit['newText']
+            # this region was valid a few view.change_count() moments back ...
+            edit_region = range_to_region(Range.from_lsp(text_edit['range']), self.view)
+            # ... but this brings it to the present.
+            edit_region = self.view.transform_region_from(edit_region, item["change_id"])
+            selection = self.view.sel()
+            primary_cursor_position = selection[0].b
+            for region in reversed(selection):
+                # This seems questionable...
+                translation = region.b - primary_cursor_position
```
predragnikolic, line 91: It took some time to get a mental image of how this works.

This calculates the __distance__ between a selection cursor and the primary cursor(for which the request was send for),
and apply that __distance__ to the TextEdit region so it becomes valid for that selection cursor.

plugin/completion.py
```diff
@@ -105,62 +56,52 @@ def format_completion(item: dict, restore_lines: RestoreLines) -> sublime.Comple
...
-                self.view.run_command("insert_snippet", {"contents": new_text})
-            else:
-                # insert text from bottom to top
-                for sel in reversed(self.view.sel()):
-                    self.view.insert(edit, sel.begin(), new_text)
+            new_text = text_edit['newText']
+            # this region was valid a few view.change_count() moments back ...
+            edit_region = range_to_region(Range.from_lsp(text_edit['range']), self.view)
+            # ... but this brings it to the present.
+            edit_region = self.view.transform_region_from(edit_region, item["change_id"])
```
predragnikolic, line 86: I am still confused what region should be passed to transform_region_from, 
but I tested this change and it works. You also deleted restore lines class... All I can say is nice job! :)

plugin/completion.py
```diff
@@ -101,62 +52,52 @@ def format_completion(item: dict, restore_lines: RestoreLines) -> sublime.Comple
...
-            if insert_text_format == InsertTextFormat.Snippet:
-                self.view.run_command("insert_snippet", {"contents": new_text})
-            else:
-                # insert text from bottom to top
-                for sel in reversed(self.view.sel()):
-                    self.view.insert(edit, sel.begin(), new_text)
+            new_text = text_edit['newText']
+            # this region was valid a few view.change_count() moments back ...
+            edit_region = range_to_region(Range.from_lsp(text_edit['range']), self.view)
+            # ... but this brings it to the present.
```
predragnikolic, line 85: I would maybe just add a better comment here, 
I would not change the logic

plugin/completion.py
rwols: Indeed :) Note that once the reverse iteration reaches that same primary cursor, `region.b == primary_cursor_position`. But for the other cursors we have `region.b > primary_cursor_position`.

plugin/completion.py
predragnikolic: I commented on the wrong place,
ignore this :D

plugin/completion.py
rwols: From what I can tell, `transform_region_from` expects a region from the past.
It uses the change_id and the internal undo-stack of ST to bring the region to the present.

I noticed that in order to get this "region from the past", we need to translate the "LSP range from the past" into a "sublime region from the past", but in order to do so we need to use `range_from_region`, which uses `point_to_offset`, which uses `view.text_point_utf16`, which doesn't use the change_id. So I'm not 100% this will always work.

plugin/completion.py
```diff
@@ -101,62 +52,52 @@ def format_completion(item: dict, restore_lines: RestoreLines) -> sublime.Comple
...
+            edit_region = self.view.transform_region_from(edit_region, item["change_id"])
+            selection = self.view.sel()
+            primary_cursor_position = selection[0].b
+            for region in reversed(selection):
+                # This seems questionable...
```
predragnikolic, line 90: I would just write a more descriptive comment here :)
instead of  # This seems questionable...

PR #952, opened 2020-04-07
Travel up the dir tree to find .flowconfig

quangbuule:
This is special treat for flow, I am not sure about other language.

Usecase: My sublime workspace contains only a sub-folder of a big project, there is no `.flowconfig` in that directory, but there is one in root big project.

LSP originally doesn't work for that case. This PR is to find and use the `.flowconfig` on the ancestor folders.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -9,8 +9,8 @@ from .typing import Callable, Dict, Any, Optional, List, Tuple
 from .workspace import is_subpath_of
 import os
 
-def find_root_folder(workspace_folder: Optional[WorkspaceFolder], config: ClientConfig) -> Optional[WorkspaceFolder]:
-    def find_flow_config(dirname):
+def find_root_folder(workspace_folder: WorkspaceFolder, config: ClientConfig) -> WorkspaceFolder:
+    def find_flow_config(dirname: str) -> str:
         if not dirname or dirname is '/':
             return '/'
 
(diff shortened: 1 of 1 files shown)


jfcherng:
This should be in something like `lsp-flow` plugin imo.

Disclaimer: I don't write `flow`

`find_root_folder()` seems to go all the way top until it meets system root (i.e., `/`). 
From https://flow.org/en/docs/config/, shouldn't it stop at the project root?

quangbuule:
> This PR should reside in something like `lsp-flow` plugin imo.
> 
> Disclaimer: I don't write `flow`
> 
> `find_root_folder()` seems to go all the way top until it meets system root (i.e., `/`).
> From https://flow.org/en/docs/config/, shouldn't it stop at the project root?

@jfcherng I described the use case in the first comment.
There is similar approach done in JSPrettier plugin: https://github.com/jonlabelle/SublimeJsPrettier/pull/207/commits

jfcherng:
> There is similar approach done in JSPrettier plugin: https://github.com/jonlabelle/SublimeJsPrettier/pull/207/commits

From https://prettier.io/docs/en/configuration.html

> The configuration file will be resolved starting from the location of the file being formatted, and searching up the file tree until a config file is (or isn't) found.

I think in `Prettier`'s doc, it explicitly says "searching up the file tree until a config file is (or isn't) found." I don't know about `flow` though.

quangbuule:
You are right, this should be done in different approach.

quangbuule:
@jfcherng the approach I was talking about is having a config for root path in .sublime-project. and LSP should be able to read that somehow. Could be done in LSP-flow? I dunno, I am not quite used to Python though :D.

PR #955, opened 2020-04-08
Compensate for View#set_status eating newlines of diagnostic messages

frou:
Say a diagnostic with the following message has been published for a source line

```python
'This expression has type bool\nThis is not a function; it cannot be applied.'
```

When the caret is on the line, the diagnostic message is shown in the Status Bar:

78336669-ac4d4000-7587-11ea-90d9-fbe192f16c8d

...but the Sublime API function `View#set_status` simply eats the newline that's part of the string, meaning those two words become conjoined.

This PR inserts a space into the string passed to `set_status` where a line in a diagnostic message ends without trailing whitespace and the next line starts without leading whitespace.

### DIFF
--- a/plugin/diagnostics.py
+++ b/plugin/diagnostics.py
@@ -1,5 +1,6 @@
 import html
 import os
+import re
 import sublime
 import sublime_plugin
 
@@ -96,7 +97,8 @@ class DiagnosticsCursorListener(LSPViewEventListener):
 
     def show_diagnostics_status(self, diagnostic: Diagnostic) -> None:
         self.has_status = True
-        self.view.set_status('lsp_diagnostics', diagnostic.message)
+        spaced_message = re.sub(r'(\S\n)(\S)', r'\1 \2', diagnostic.message)
+        self.view.set_status('lsp_diagnostics', spaced_message)
 
     def clear_diagnostics_status(self) -> None:
         self.view.erase_status('lsp_diagnostics')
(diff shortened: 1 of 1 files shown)


plugin/diagnostics.py
```diff
@@ -96,7 +97,8 @@ def on_selection_modified_async(self) -> None:
 
     def show_diagnostics_status(self, diagnostic: Diagnostic) -> None:
         self.has_status = True
-        self.view.set_status('lsp_diagnostics', diagnostic.message)
+        spaced_message = re.sub(r'(\S\n)(\S)', r'\1 \2', diagnostic.message)
```
rchl, line 100: Wouldn't it make sense to just remove the newline: `r'(\S)\n(\S)'`?
I guess it wouldn't make a difference in practice but I feel it's "more correct".

predragnikolic: `diagnostic.message.replace("\n", " ")` would this be the equivalent of the regex you typed?

rchl: Not if you want this behavior:
> This PR inserts a space into the string passed to set_status where a line in a diagnostic message ends without trailing whitespace and the next line starts without leading whitespace.

predragnikolic: I would keep it simple and suggest using a regular replace,
unless @frou insist on using a regex solution.

In that case , just put a comment above the regex

frou: If the message already has a space before and/or after the newline, doing a simple string replace would result in there being >1 spaces between the words.

rwols: I feel like this extra text after `\n` pollutes the status bar too much anyway. Perhaps strip away everything after the `\n`?

frou: Why make an assumption about how servers arrange the useful content in their message? It no doubt varies widely. Even in my example, the latter part is definitely helpful.

rwols: OK, makes sense.

PR #961, opened 2020-04-10
Use JSON format for server logs

rchl:
Communication is done in JSON so it doesn't make sense to have
python representation of it in logs.

### DIFF
--- a/plugin/core/rpc.py
+++ b/plugin/core/rpc.py
@@ -12,9 +12,9 @@ TCP_CONNECT_TIMEOUT = 5
 DEFAULT_SYNC_REQUEST_TIMEOUT = 1.0
 
 
-def format_request(payload: Dict[str, Any]) -> str:
+def json_format_request(payload: Dict[str, Any], compact: bool = True) -> str:
     """Converts the request into json"""
-    return json.dumps(payload, sort_keys=False, check_circular=False, separators=(',', ':'))
+    return json.dumps(payload, sort_keys=False, check_circular=False, separators=(',', ':') if compact else None)
 
 
 def try_terminate_process(process: subprocess.Popen) -> None:
@@ -33,7 +33,7 @@ class PreformattedPayloadLogger:
 
     def log(self, message: str, params: Any, log_payload: bool) -> None:
         if log_payload:
-            message = "{}: {}".format(message, params)
+            message = "{}: {}".format(message, json_format_request(params, compact=False))
         self.sink(message)
 
     def format_response(self, direction: str, request_id: Any) -> str:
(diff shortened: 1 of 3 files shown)


plugin/core/rpc.py
```diff
@@ -12,9 +12,9 @@
 DEFAULT_SYNC_REQUEST_TIMEOUT = 1.0
 
 
-def format_request(payload: Dict[str, Any]) -> str:
+def json_format_request(payload: Dict[str, Any], compact=True) -> str:
```
rchl, line 15: Renamed it for clarity and also as it had the same name as one method in this file, making it confusing when searching for that name across the project.

rchl:
This change also makes  it easy to parse and prettify logs in external tools (typically written in JS).

rchl:
This change alone would not make logs compatible with inspector and that was not really the intention of the change.

> I don't think many people demand to see JSON payloads. Python dicts are an OK data structure to debug.

It's not a necessity for anyone, I don't think. But original communication is done in JSON so it's not entirely truthful to show it in Python. Especially if you want to compare it with the spec and you are seeing None in the logs and null in the spec.

That's the main motivation for this change - to show things as they really are supposed to be shown. It also makes it easier for the log-parser I was making to pretty-print the payloads but I don't think I'll be doing much with it anyway, now that I found inspector.

I think the arguments about impacting performance are not very relevant as the number of messages per second is so low that a tiny performance hit this would incur wouldn't really matter in practice IMO.

It's common for code to be less performant when debugging after all.

And also normal people don't run with logging enabled so I would expect this to only affect developers of LSP or LSP packages.

I agree with the steps needed to create an inspector-compatible logger but that's a separate feature in my mind.

rchl:
BTW. I could change the code to only do JSON parsing once but then I would prefer 'non-compact' representation to be used (so would have to use that one also for sending over the line).

rchl:
> BTW. I could change the code to only do JSON parsing once but then I would prefer 'non-compact' representation to be used (so would have to use that one also for sending over the line).

Well, that wouldn't be trivial so maybe I won't...

rwols:
> It's not a necessity for anyone, I don't think. But original communication is done in JSON so it's not entirely truthful to show it in Python. Especially if you want to compare it with the spec and you are seeing None in the logs and null in the spec.

It's exactly the same data structure, only presented differently.

> I think the arguments about impacting performance are not very relevant as the number of messages per second is so low that a tiny performance hit this would incur wouldn't really matter in practice IMO.

But this now introduces a performance penalty with no real gain. We would still have to write an external tool to massage the messages into something that language-server-protocol-inspector would understand and parse.

So we do 2x the work of serializing to JSON, while being 20% there. External tool needed.

I'm proposing a path of changes that will get us to doing 2x the work of serializing to JSON (but only when the user wants it), and getting 100% there, no external tool needed.

Issue #957, opened 2020-04-08
TextEdit is not applied correctly 


predragnikolic:
![output](https://user-images.githubusercontent.com/22029477/78840351-c8733600-79fa-11ea-82c1-8bb265d2299c.gif)

```
            # this region was valid a few view.change_count() moments back ...
            edit_region = range_to_region(Range.from_lsp(text_edit['range']), self.view)
            # ... but this brings it to the present.
  mutate >> edit_region = self.view.transform_region_from(edit_region, item["change_id"])
```

We are probably not passing the correct region to the transform_region_from.
We should also not mutate the TextEdit from the server. (the region returned from `transform_region_from` is not the same as the text edit region returned from the server)

I think that we need to grab a snapshot(something like restore_lines, or something different).

rwols:
It's possibly a bug in `transform_region_from`, or a misunderstanding on how that function is supposed to be used.

This is using `transform_region_from` (it should remove `u>`, but doesn't):
![Peek 2020-04-09 18-55](https://user-images.githubusercontent.com/2431823/78925357-9665e080-7a9b-11ea-8211-56565f3ae458.gif)

This skips `transform_region_from` and uses the historic region as-is:
![Peek 2020-04-09 18-57](https://user-images.githubusercontent.com/2431823/78925418-ada4ce00-7a9b-11ea-9ffe-527408efef09.gif)

rwols:
I believe the changes in #962 inadvertently fixed your issue too:

![Peek 2020-04-11 15-17](https://user-images.githubusercontent.com/2431823/79044811-a39fde00-7c07-11ea-9be6-738df8a1f8cb.gif)

The last remaining thing to do is to fix up the auto complete triggers.

rwols:
Fixed by #962

PR #962, opened 2020-04-11
Fix buffer modifications while completing

### LINKED ISSUE #956: clangd: duplicate bracket on completion
When clangd offers a completion option, it will include (closing?) brackets as well.
The problem is that if the closing bracket is already in place, the auto-completion will cause a duplicate character.
Is there a way to  tweak/disable this behaviour?

Here are some screenshots:

Screenshot 2020-04-08 at 23 35 47
Screenshot 2020-04-08 at 23 35 59
Screenshot 2020-04-08 at 23 35 35

rwols:
The buffer can be modified by the user, e.g. by typing more characters
while the AC widget is visible. This invalidates all text edits returned
by the language server. We now account for this by keeping an eye on the
buffer modifications via on_text_changed and adjusting the text edits
accordingly.

Another problem was that we called view.change_id() too early. It should be
called just before running the completion command. This accounts for the
prefix removal that ST does.

close #956

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -73,11 +73,8 @@ class LspSelectCompletionItemCommand(sublime_plugin.TextCommand):
                 # To do that, translate, or offset, the LSP edit region into the non-"primary" regions.
                 # The concept of "primary" is our own, and there is no mention of it in the LSP spec.
                 translation = region.b - primary_cursor_position
-                a = edit_region.a + translation
-                b = edit_region.b + translation
-                r = sublime.Region(a, b)
                 # We will assume that the user was typing the same characters.
-                self.view.erase(edit, r)
+                self.view.erase(edit, sublime.Region(edit_region.a + translation, edit_region.b + translation))
         else:
             new_text = item.get('insertText') or item['label']
 
@@ -95,8 +92,7 @@ class LspSelectCompletionItemCommand(sublime_plugin.TextCommand):
             self.do_resolve(item)
 
     def do_resolve(self, item: dict) -> None:
-        session = session_for_view(
-            self.view, 'completionProvider', self.view.sel()[0].begin())
+        session = session_for_view(self.view, 'completionProvider', self.view.sel()[0].begin())
         if not session:
             return
 
@@ -107,8 +103,7 @@ class LspSelectCompletionItemCommand(sublime_plugin.TextCommand):
         completion_provider = session.get_capability('completionProvider')
         has_resolve_provider = completion_provider and completion_provider.get('resolveProvider', False)
         if has_resolve_provider:
-            client.send_request(Request.resolveCompletionItem(
-                item), self.handle_resolve_response)
+            client.send_request(Request.resolveCompletionItem(item), self.handle_resolve_response)
 
     def handle_resolve_response(self, response: Optional[dict]) -> None:
         if response:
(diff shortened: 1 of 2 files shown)


plugin/completion.py
```diff
@@ -205,32 +199,94 @@ def on_query_completions(self, prefix: str, locations: List[int]) -> Optional[su
...
+            # From the LSP spec: the range of the edit must be a single line range and it must contain the position at
+            # which completion has been requested.
+            row, start_col_utf16, end_col_utf16 = start["line"], start["character"], end["character"]
+            # The TextEdit from the language server might apply to an old version of the buffer. The user may have
+            # entered more characters in the meantime.
+            global _changes_since_sent
+            for change in _changes_since_sent:
+                # Thus, adjust the range for each buffer change.
+                start_col_utf16, end_col_utf16 = transform_region(row, start_col_utf16, end_col_utf16, change)
+            # We don't have to use region_to_range. That function clamps the unreliable input, but having unreliable
```
predragnikolic, line 232: Can we somehow apply _changes_since_sent to only the selected completion item (in the LspSelectCompletionItemCommand), instead of all completion items. 

I don't see a point in applying the transform_region to all items, because only one will be chosen.

predragnikolic: To sum up the problem that this PR solves:

If the user types:

```
1| h|
```
-> completions `request_1` is sent.

User continues typing while, but the request_1 is not finished

```
1| he|
```

<- responses from request_1 arrived, users sees completion items in AC.

-> request_2 has been send (lets say it will take 5sec for the second request to complete, it is not a real case but lets imagine that this happens)

The user sees completion items from request_1
that were calculated when the view state was:
```
1| h|
```

but the current view state is:
```
1| he|
```

When the select a completion item,
because the view state has changed. 
The text edit wont be applied correctly.

<hr>

There are two solutions to this problem
one of them being your solution and that is:

When a request is send.
You record each edit that happened (with on_text_changed).
This way you pick that the letter `e` was typed.
And call transform_region on **all completion items** returned from the server
to change the TextEdit range to include all text changes. 
Keep in mind that there could be 1000 of completion items, but the user will only select one. 
So you did 999 of unnecessary calls to transform_region. (which I personally don't find that bad, but I want to point out that)

But another solution is to just bring views state to the state it was when the request was made.

So we need the current view state: 
```
1| he|
```

to bring to the state it was when the request was made ant that is:
```
1| h|
```

That way we can avoid transforming TextEdit-

This was all I had to say.
Functionality wise this PR does the job
and it fixes #957

rwols: If I move the adjustment of the text edit range to the LspInsertCompletionItem command, then my test fails:

```
FAIL: test_nontrivial_text_edit_removal (test_completion.QueryCompletionsTests)
  File "/home/raoul/.config/sublime-text-3/Packages/UnitTesting/unittesting/core/py33/case.py", line 28, in _executeTestPart
    yield from deferred
  File "/home/raoul/.config/sublime-text-3/Packages/LSP/tests/test_completion.py", line 509, in test_nontrivial_text_edit_removal
    self.assertEqual(self.read_file(), '#include <uchar.h>')
AssertionError: '#include <uchar.h>>' != '#include <uchar.h>'
- #include <uchar.h>>
?                   -
+ #include <uchar.h>
```

I am guessing this is because `transform_region` won't work if the prefix string is already removed (which it is if we're inside LspSelectCompletionCommand).

> Keep in mind that there could be 1000 of completion items, but the user will only select one. So you did 999 of unnecessary calls to transform_region.

Yes :/ no way around it.

Well, with the new `details` field at the bottom we're going to parse 1000 markdown strings into HTML anyway ;)

rwols: Besides a quality lang server will stop at ~200 items and set `isIncomplete` to `True`.

rwols: I invite you to try moving the logic to LspSelectCompletionCommand, and convince yourself that it will be impossible to do so.

predragnikolic: >  1000 markdown strings into HTML anyway

True :D

PR #964, opened 2020-04-11
Make tests less likely to fail on Mac

rchl:
With default value of 1, tests fails pretty much consistently on every
run on Mac.

10 seems to give stable results as long as there is no other activity
on the machine.


rchl:
Let me know if you want this to be Mac-specific but I think unifying can be beneficial to ensure that the testing environment is equal on all platforms (even if the platforms themselves behave differently).

PR #968, opened 2020-04-13
Fix: too many auto-complete registrations

rwols:
The auto-complete triggers do not de-register themselves when restarting ST:

    [{'characters': '<', 'selector': 'text.html, text.xml'},
     {'selector': 'punctuation.accessor'},
     {'characters': '.', 'selector': '- comment - punctuation.definition.string.end'},
     {'characters': '.', 'selector': '- comment - punctuation.definition.string.end'},
     {'characters': '.', 'selector': '- comment - punctuation.definition.string.end'},
     {'characters': '.', 'selector': '- comment - punctuation.definition.string.end'},
     {'characters': '.', 'selector': '- comment - punctuation.definition.string.end'}]

This commit adds a token to the registration dict that we re-use in a new
`__del__` method to remove our registration. The token is the name of the
language server.

After this change, when you look in the console *before typing a character*:

    [{'characters': '<', 'selector': 'text.html, text.xml'},
     {'selector': 'punctuation.accessor'}]

When you typed a character and then look in the console again:

    [{'characters': '<', 'selector': 'text.html, text.xml'},
     {'selector': 'punctuation.accessor'},
     {'characters': '.', 'selector': '- comment - punctuation.definition.string.end', 'server': 'pyls'}]

When you restart ST, the registration is gone.

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -6,7 +6,6 @@ from .core.edit import parse_text_edit
 from .core.logging import debug
 from .core.protocol import Request, InsertTextFormat
 from .core.registry import session_for_view, client_from_session, LSPViewEventListener
-from .core.sessions import Session
 from .core.settings import settings, client_configs
 from .core.typing import Any, List, Dict, Optional, Union, Iterable, Tuple
 from .core.views import text_document_position_params
@@ -130,6 +129,12 @@ class CompletionHandler(LSPViewEventListener):
         self.enabled = False
         self._request_in_flight = False
 
+    def __del__(self) -> None:
+        settings = self.view.settings()
+        triggers = settings.get("auto_complete_triggers")  # type: List[Dict[str, str]]
+        triggers = [trigger for trigger in triggers if 'server' not in trigger]
+        settings.set("auto_complete_triggers", triggers)
+
     @classmethod
     def is_applicable(cls, view_settings: dict) -> bool:
         if 'completion' in settings.disabled_capabilities:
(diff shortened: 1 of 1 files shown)


plugin/completion.py
```diff
@@ -130,6 +129,12 @@ def __init__(self, view: sublime.View) -> None:
         self.enabled = False
         self._request_in_flight = False
 
+    def __del__(self) -> None:
```
rchl, line 132: This most likely won't be called on restarting ST but I guess that's acceptable?

rwols: It does, that's the trick :)

rchl: It didn't seem like it here.
Trying to write settings from __del__ might be frowned upon as some things are already destroyed.

rwols: It only removes entries that have a `server` key. If you have multiple entries already you need to remove your `sublime.workspace` cache file if you're using a .sublime-project and otherwise you'll need to remove your global cache file in $DATA/Local.Session.sublime_session next to your license file.

rchl: Now that I think of it, maybe I was just having wrong expectations. On restarting ST the view is not necessarily destroyed (from user's POV) so maybe it makes sense to preserve that in the session.

What I was testing was:
1) Open JS file with lsp-typescript server enabled
2) Verify that `view.settings().get("auto_complete_triggers")` returns only default triggers
3) Type dot and verify that new entry with `'server': 'lsp-typescript'` was added
4) Restart ST
5) Check `view.settings().get("auto_complete_triggers")`

I'm getting triggers with new item added while I was expecting only default ones.
Closing the view and re-opening shows only default triggers again.

PR #971, opened 2020-04-18
Add experimental_capabilities setting

ayoub-benali:
This would allow language handlers to define client experimental capabilities
A fallow up on https://github.com/sublimelsp/LSP/issues/927#issuecomment-612252151

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -94,7 +94,7 @@ Some language servers support multiple languages, which can be specified in the 
 | tcp_port | see instructions below |
 | tcp_host | see instructions below |
 | tcp_mode | see instructions below |
-| experimental_capabilities | options to send to the server at startup (rarely used) |
+| experimental_capabilities | Turn on experimental capabilities of a language server. This is a dictionary and differs per language server |
 
 The default transport is stdio, but TCP is also supported.
 The port number can be inserted into the server's arguments by adding a `{port}` placeholder in `command`.
--- a/docs/index.md
+++ b/docs/index.md
@@ -94,6 +94,7 @@ Some language servers support multiple languages, which can be specified in the 
 | tcp_port | see instructions below |
 | tcp_host | see instructions below |
 | tcp_mode | see instructions below |
+| experimental_capabilities | options to send to the server at startup (rarely used) |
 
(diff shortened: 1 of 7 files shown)


rwols:
Could you add some tests for this in the `ConfigParsingTests` suite in `tests/test_configs.py`?

You can run the tests with the `UnitTesting: Test Current File` command from the command palette.

plugin/core/settings.py
```diff
@@ -187,7 +187,7 @@ def read_client_config(name: str, client_config: Dict) -> ClientConfig:
         client_config.get("env", dict()),
         client_config.get("tcp_host", None),
         client_config.get("tcp_mode", None),
-        client_config.get("experimental_capabilities", config.experimental_capabilities)
+        client_config.get("experimental_capabilities", dict())
```
rwols, line 190: Should this be `None` instead of `dict()`? We're checking for `is not None` in `get_initialize_params`.

ayoub-benali: In theory yes but then `initializationOptions` should be None as well [above](https://github.com/sublimelsp/LSP/pull/971/files#diff-c7b7ba6d8b0930a59af07dc6cdb03159R185) ?

ayoub-benali: I mean we also do `if config.init_options is not None` in `get_initialize_params` and `init_options` has `dict()` as default value in `read_client_config`. Should they then have the same default ?

rwols: Good observation. Well, I guess it doesn't really matter for this PR.

docs/index.md
```diff
@@ -94,6 +94,7 @@ Some language servers support multiple languages, which can be specified in the
 | tcp_port | see instructions below |
 | tcp_host | see instructions below |
 | tcp_mode | see instructions below |
+| experimental_capabilities | options to send to the server at startup (rarely used) |
```
rwols, line 97: This now has the exact same description as `initializationOptions`. This might confuse new users. Is there a better description for experimental_capabilities? Perhaps simply "Turn on experimental capabilities of a language server. This is a dictionary and differs per language server".

ayoub-benali:
Thanks @rwols

PR #972, opened 2020-04-18
Extract an abstract base class from the logger

rwols:
This is a first step towards different kinds of loggers, for usage in https://github.com/rchl/lsp-log-parser

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -2,7 +2,7 @@ from .diagnostics import DiagnosticsStorage
 from .edit import parse_workspace_edit
 from .logging import debug
 from .protocol import Notification, Response
-from .rpc import Client
+from .rpc import Client, SublimeLogger
 from .sessions import Session
 from .types import ClientConfig
 from .types import ClientStates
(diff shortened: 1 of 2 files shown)


plugin/core/windows.py
```diff
@@ -515,7 +515,7 @@ def _handle_pre_initialize(self, session: Session) -> None:
         client.set_crash_handler(lambda: self._handle_server_crash(session.config))
         client.set_error_display_handler(self._window.status_message)
 
-        if self.server_panel_factory:
+        if self.server_panel_factory and isinstance(client.logger, SublimeLogger):
             client.logger.server_name = session.config.name
             client.logger.sink = self._payload_log_sink
```
rchl, line 520: Just a note since this is not a new "problem" but I find it hacky to just "deep-update" properties of some other instance (even if those properties are not denoted with a private modifier).

That just made me scratch my head, trying to figure out how things end up in the server panel rather than the initial "print" sink.

I think there should rather be a method exposed on the client (or logger, or both) to do that.

rwols: You are correct in pointing this out. The Client should ideally receive an abstract `Logger` type just like it receives an abstract `Transport` type, but to do that we'd have to add yet another arg to the many, many layers of callbacks and factory functions. I'm playing around with ideas in my head for this though.

PR #974, opened 2020-04-19
migrate window/showMessageRequest handling to mdpopups.show_popup

ayoub-benali:
[Previously](https://github.com/sublimelsp/LSP/pull/470) it was handled with show_quick_panel but we couldn't show
the request message to the user. Until ST4 would allow such a feature
we resort to mdpopups.show_popup

popup

 - [x] add tests

### DIFF
--- a/plugin/core/message_request_handler.py
+++ b/plugin/core/message_request_handler.py
@@ -6,7 +6,7 @@ import sublime
 
 
 class MessageRequestHandler():
-    def __init__(self, view: sublime.View, client: Client, request_id: Any, params: dict) -> None:
+    def __init__(self, view: sublime.View, client: Client, request_id: Any, params: dict, source: str) -> None:
         self.client = client
         self.request_id = request_id
         self.request_sent = False
@@ -15,6 +15,7 @@ class MessageRequestHandler():
         self.titles = list(action.get("title") for action in actions)
         self.message = params.get('message', '')
         self.message_type = params.get('type', 4)
+        self.source = source
 
     def _send_user_choice(self, href: int = -1) -> None:
         if not self.request_sent:
@@ -31,6 +32,7 @@ class MessageRequestHandler():
     def show(self) -> None:
         show_notification(
             self.view,
+            self.source,
             self.message_type,
             self.message,
             self.titles,
@@ -39,7 +41,7 @@ class MessageRequestHandler():
         )
 
 
-def message_content(message_type: int, message: str, titles: List[str]) -> str:
+def message_content(source: str, message_type: int, message: str, titles: List[str]) -> str:
     formatted = []
     icons = {
         1: '❗',
@@ -48,6 +50,7 @@ def message_content(message_type: int, message: str, titles: List[str]) -> str:
         4: '📝'
     }
     icon = icons.get(message_type, '')
+    formatted.append("<h2>{}</h2>".format(source))
     formatted.append("<p class='message'>{} {}</p>".format(icon, message))
 
     buttons = []
@@ -59,10 +62,10 @@ def message_content(message_type: int, message: str, titles: List[str]) -> str:
     return "".join(formatted)
 
 
-def show_notification(view: sublime.View, message_type: int, message: str, titles: List[str],
+def show_notification(view: sublime.View, source: str, message_type: int, message: str, titles: List[str],
                       on_navigate: Callable, on_hide: Callable) -> None:
     stylesheet = sublime.load_resource("Packages/LSP/notification.css")
-    contents = message_content(message_type, message, titles)
+    contents = message_content(source, message_type, message, titles)
     mdpopups.show_popup(
         view,
         contents,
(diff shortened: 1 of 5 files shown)


ayoub-benali:
@rwols currently location is left to default (-1) which is the current cursor but I think notifications should be shown at the same position. 
I believe vscode shows notifications a bottom right. I think we can't do that with sublime API, or do you know a way to do it ?
Otherwise we can show it always at location (0, 0) top left ?

plugin/core/message_request_handler.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, view: Optional[ViewLike], client: Client, request_id: Any, pa
     def _send_user_choice(self, href: int = -1) -> None:
         if not self.request_sent and self.view:
             self.request_sent = True
-            mdpopups.hide_popup(self.view)
+            mdpopups.hide_popup(self.view)  # type: ignore
```
rwols, line 22: There's a file `mdpopups.pyi` which provides the type hinting. If you want to fix the mypy error you should add `hide_popup` to `mdpopups.pyi` in this repo. But, you could also simply use `self.view.hide_popup()`, it does the same thing.

rwols:
ping @rchl, you can see the CI failing here.

rchl:
If you will get `test_apply_formatting_and_preserve_order` failing as the first test then it's most likely not related to your changes. I'm looking into that issue now in #977.

plugin/core/message_request_handler.py
```diff
@@ -0,0 +1,93 @@
+import mdpopups
+from .rpc import Client
+from .typing import Any, List, Callable
+from .protocol import Response
+from sublime import View
+
+
+class MessageRequestHandler(object):
```
rwols, line 8: You can leave out the fact that `MessageRequestHandler` inherits from `object` -- that is old-style python2 syntax https://stackoverflow.com/questions/4015417/python-class-inherits-object

plugin/core/message_request_handler.py
```diff
@@ -0,0 +1,93 @@
...
+        )
+
+
+def message_content(message_type: int, message: str, titles: List[str]) -> str:
+    formatted = []
+    icons = {
+      1: '❗',
+      2: '⚠️',
+      3: 'ℹ️',
+      4: '📝'
```
rwols, line 48: The indentation is 6 spaces here, can it be 8 instead?

plugin/core/windows.py
```diff
@@ -21,6 +21,7 @@
 from .workspace import ProjectFolders
 from .workspace import sorted_workspace_folders
 import threading
+from .message_request_handler import MessageRequestHandler
```
rwols, line 24: Please keep this list of imports sorted (select all import lines -> Edit -> Sort Lines)

plugin/core/message_request_handler.py
```diff
@@ -0,0 +1,93 @@
...
+        buttons.append("<a href='{}'>{}</a>".format(idx, title))
+
+    formatted.append("<p class='actions'>" + " ".join(buttons) + "</p>")
+
+    return "".join(formatted)
+
+
+def show_notification(view: View, message_type: int, message: str, titles: List[str],
+                      on_navigate: Callable, on_hide: Callable) -> None:
+    myStyle = """
```
rwols, line 64: Can we name this variable `css` instead of `myStyle`? Try to avoid camelCase naming and instead go for snake_case naming.

plugin/core/message_request_handler.py
```diff
@@ -0,0 +1,93 @@
...
+        color: var(--foreground);
+    }
+    """
+
+    contents = message_content(message_type, message, titles)
+    mdpopups.show_popup(
+        view,
+        contents,
+        css=myStyle,
+        md=False,
```
rwols, line 86: If `md=False`, do we use `mdpopups.show_popup` because it injects additional html and css? And the user can modify that to his/her liking?

ayoub-benali: I made the css loaded as ressource so user can modified it.

rwols:
I noticed that the popup doesn't have a title/header.

Because this popup may... pop up... out of nowhere, I predict users will find it confusing where it came from.

Can the name of the server (`session.config.name`) be incorporated into a big header? That way, users will know where the pop up came from.

ayoub-benali:
Good point, I will add that

ayoub-benali:
This is how it looks like now

Peek 2020-04-22 16-59

themilkman:
Sorry for digging up such an old PR, but I recently saw this the first time in action and thought it was very cool. @predragnikolic pointed out that it was contributed by you, @ayoub-benali, so I just wanted to say thanks! :-)

rchl:
Not to take away from the moment but you should probably be aware of its drawbacks - it is very easy to dismiss accidentally by moving the cursor and if the cursor is out of view then the dialog won't even appear.

ST doesn't have a good UX for this kind of functionality.

jwortmann:
> and if the cursor is out of view then the dialog won't even appear

I fixed that in a pull request a while ago.

But yeah, ideally ST would provide a "notification" UI with a floating window in a corner like in VS Code, that stays around while you can still type in the file.

notification

rchl:
> and if the cursor is out of view then the dialog won't even appear

Then I guess if you scroll it out of view after it appears since that's how ST popups behave.

PR #975, opened 2020-04-19
Handle incoming notifications/responses during a sync request (#787)

rchl:
Try re-applying the sync fixes.


rwols:
OK, let's see what it does on the master branch.

PR #977, opened 2020-04-21
Various fixes / improvements

rchl:
At least locally the window was closed when running `close_all` command
and that made the tests fail due to missing window.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,7 @@ jobs:
     - run: |
         sh ci.sh bootstrap
         sh ci.sh install_package_control
-        subl &
+        subl --stay &
         sleep 8
         sh ci.sh run_tests --coverage
       name: Run UnitTesting tests
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,3 +37,33 @@ jobs:
     - run: |
         pip3 install coverage==4.5.4 codecov==2.0.15
         codecov
+  Linux:
+    runs-on: ubuntu-latest
+    container: sublimetext/unittesting
+    env:
+      CODECOV_TOKEN: ${{secrets.CODECOV_TOKEN}}
+      SUBLIME_TEXT_VERSION: "3"
+      SUBLIME_TEXT_ARCH: x64
+      PACKAGE: LSP
+    steps:
+      - uses: actions/checkout@v1
+      - run: echo ::add-path::$HOME/.local/bin
+      - run: sudo apt update
+      - run: sudo apt install --no-install-recommends -y x11-xserver-utils imagemagick python3-pip
+      - run: sh -e /etc/init.d/xvfb start
+      - run: curl -OL https://raw.githubusercontent.com/SublimeText/UnitTesting/master/sbin/github.sh
+      - run: python3 tests/server.py --version
+      - run: python3 tests/adjust_unittesting_config_for_ci.py && cat unittesting.json
+      - run: |
+          PATH="$HOME/.local/bin:$PATH"
+          sh github.sh bootstrap
+          sh github.sh install_package_control
+          sh github.sh run_tests --coverage
+        name: Run UnitTesting tests
+      - run: pip3 install mypy==0.750 flake8==3.7.9 yapf==0.29.0 --user
+      - run: mypy -p plugin
+      - run: flake8 plugin tests
+      - run: yapf --diff plugin/core/main.py
+      - run: |
+          pip3 install coverage==4.5.4 codecov==2.0.15
+          codecov
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -262,11 +262,11 @@ class CompletionHandler(LSPViewEventListener):
         if settings.complete_all_chars or self.is_after_trigger_character(locations[0]):
             self.manager.documents.purge_changes(self.view)
             document_position = text_document_position_params(self.view, locations[0])
+            self.state = CompletionState.REQUESTING
             client.send_request(
                 Request.complete(document_position),
                 self.handle_response,
                 self.handle_error)
-            self.state = CompletionState.REQUESTING
 
     def do_resolve(self, item: dict) -> None:
         view = self.view
--- a/plugin/core/rpc.py
+++ b/plugin/core/rpc.py
@@ -153,9 +153,10 @@ class Client(object):
         if self.transport is not None:
             with self._sync_request_cvar:
                 self.request_id += 1
-                self.logger.outgoing_request(self.request_id, request.method, request.params, blocking=False)
-                self._response_handlers[self.request_id] = (handler, error_handler)
-                self.send_payload(request.to_payload(self.request_id))
+                request_id = self.request_id
+                self._response_handlers[request_id] = (handler, error_handler)
+            self.logger.outgoing_request(request_id, request.method, request.params, blocking=False)
+            self.send_payload(request.to_payload(request_id))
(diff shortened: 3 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -30,6 +30,10 @@ jobs:
         sleep 8
         sh ci.sh run_tests --coverage
       name: Run UnitTesting tests
+    - name: Create a desktop screenshot
+      uses: OrbitalOwen/desktop-screenshot-action@0.1
+      with:
+        file-name: 'desktop.jpg'
```
rwols, line 36: I just wanted to say that this is some amazing work you're doing :)

rchl: I have no idea if it's even gonna work and where would the screenshot end up. I'm running out of ideas how to debug that. :)

I'm pretty sure focus is getting lost after first `TextDocumentTestCase` test. Either due to UnitTesting issue I've mentioned elsewhere or maybe some (PackageControl?) dialog popping up...

rchl:
I might have found the root issue. Now only the tests in test_rpc are failing and probably will have to be adjusted.

The issue was that  `rpc.send_request` could end up returning after the result handler was called and some calling code didn't like that.

From the logs, this is what could happen:
```
[MainThread] [21:57:20.246]  [RPC] (send_request) Locking...
[ Thread-62] [21:57:20.246]  [RPC] (receive_payload) Locking...
[ Thread-62] [21:57:20.246]  [RPC] (receive_payload) Unlocked! None
[MainThread] [21:57:20.246]  [RPC] (send_request Unlocked)
```
So async thread manages to return results before the `send_request` block that is holding the lock is exited. I'm surprised that it's possible since it required sending payload to separate thread but apparently Linux is very fast. ;) Or the locks slow down things enough. I don't know, I'm not a thread expert. :)

rchl:
After sleeping on it, I'm not sure what I said makes sense... The main thread is holding the lock so it should block the async thread when that one attempts to acquire it.

Maybe the logs are confusing me with wrong order since those are from different threads and tend to be out of sync in relation to each other.

But that change **does** fix the Linux issues for me at least.
I will look again to try to understand that better.

rchl:
So here is the relevant code that fails:
```py
xdebug('[completions] sending request')
client.send_request(
    Request.complete(document_position),
    self.handle_response,
    self.handle_error)
xdebug('[completions] request sent')
self.state = CompletionState.REQUESTING
xdebug('[completions] state changed')
```

And here is the log from execution:
```
[MainThread] [06:43:24.415]  [completions] sending request
[MainThread] [06:43:24.415]  [RPC] -> send_request: textDocument/completion
[MainThread] [06:43:24.416]  [RPC] (send_request) Locking...
[MainThread] [06:43:24.416]  --> TEST textDocument/completion(5)
[ Thread-62] [06:43:24.428]  [RPC] (receive_payload) Locking...
[ Thread-62] [06:43:24.428]  <<< TEST 5
[ Thread-62] [06:43:24.428]  [RPC] (receive_payload) Unlocked
LSP: Got unexpected response while in state 0
[MainThread] [06:43:24.417]  [RPC] (send_request Unlocked)
[MainThread] [06:43:24.428]  [completions] request sent
[MainThread] [06:43:24.428]  [completions] state changed
```

And for completness, here is `send_request` with above logs added:
```py
def send_request(
        self,
        request: Request,
        handler: Callable[[Optional[Any]], None],
        error_handler: Optional[Callable[[Any], None]] = None,
) -> None:
    xdebug('[RPC] -> send_request:', request.method, request.params, ex=True)
    if self.transport is not None:
        xdebug('[RPC] (send_request) Locking...', ex=False)
        with self._sync_request_cvar:
            self.request_id += 1
            self.logger.outgoing_request(self.request_id, request.method, request.params, blocking=False)
            self._response_handlers[self.request_id] = (handler, error_handler)
            self.send_payload(request.to_payload(self.request_id))
        xdebug('[RPC] (send_request Unlocked)')
    else:
        debug('unable to send', request.method)
        if error_handler is not None:
            error_handler(None)
        return None
```

So clearly, the response handler is processed before we manage to set `self.state` to `REQUESTING`. While the order or lines is confusing, the timestamps show better what happened.

rwols:
Wow, the linux build succeeds!

rchl:
So there is still possibility of this triggering on either Mac or Linux:
```
ERROR: test_additional_edits (test_completion.QueryCompletionsTests)
  File "/Users/runner/Library/Application Support/Sublime Text 3/Packages/UnitTesting/unittesting/core/py33/case.py", line 28, in _executeTestPart
    yield from deferred
  File "/Users/runner/Library/Application Support/Sublime Text 3/Packages/LSP/tests/test_completion.py", line 383, in test_additional_edits
    yield from self.await_view_change(original_change_count + 2)
  File "/Users/runner/Library/Application Support/Sublime Text 3/Packages/LSP/tests/setup.py", line 254, in await_view_change
    yield {"condition": condition, "timeout": TIMEOUT_TIME}

```
also seen this:
```
ERROR: test_remove_line_and_then_insert_at_that_line_at_end (test_single_document.SingleDocumentTestCase)
  File "/Users/runner/Library/Application Support/Sublime Text 3/Packages/UnitTesting/unittesting/core/py33/case.py", line 28, in _executeTestPart
    yield from deferred
  File "/Users/runner/Library/Application Support/Sublime Text 3/Packages/LSP/tests/test_single_document.py", line 133, in test_remove_line_and_then_insert_at_that_line_at_end
    yield from self.__run_formatting_test(original, expected, file_changes)
  File "/Users/runner/Library/Application Support/Sublime Text 3/Packages/LSP/tests/test_single_document.py", line 193, in __run_formatting_test
    yield from self.await_view_change(original_change_count + len(file_changes))
  File "/Users/runner/Library/Application Support/Sublime Text 3/Packages/LSP/tests/setup.py", line 234, in await_view_change
    yield {"condition": condition, "timeout": TIMEOUT_TIME}

```
But so far I haven't been able to reproduce it with extra logging.
I would say we can merge this and look at the random failures later.

The same failure was randomly triggering even before the sync changes were integrated. One example here: https://github.com/sublimelsp/LSP/runs/531458569

rchl:
Well, it is not shy to fail without logging though.
Some sort of timing issue...

predragnikolic:
The Open-Closed Principle 😂

rchl:
Subscribing not recommended :)
Still can't get it to fail with any meaningful logging included. Seems like logging just changes timing to make it more likely to pass.

rchl:
As soon as I revert logging it starts failing...

rwols:
Perhaps we can remove logging in little steps? That way we can pinpoint what parts are "influenced" by the added logging.

rchl:
Yes, but latest logs were pretty minimal (https://github.com/sublimelsp/LSP/pull/977/commits/d66cb7361afeaaf02c52b7e1bb1ec88a128f7a1b). I didn't even log RPC.

If I remove more, the usability of logs kinda goes away :)

But I will experiment with it more.

rwols:
At this point it's not really about the actual logging anymore, it's more about which line of code added makes the tests pass and why?

rchl:
The only potential effect of those logs is slowing down execution a bit so with multiple threads that can make a difference.

rchl:
And yes, we want to figure out why that makes tests pass. It could be something similar to the issue in completions where state was set to late maybe.

rwols:
We're going to want this figured out :)

rwols:
Excellent work, many thanks.

PR #978, opened 2020-04-21
Fix incremental didChange notification

deathaxe:
The `rangeLength` is marked deprecated in the specs.

see: https://microsoft.github.io/language-server-protocol/specification.html#textDocument_didChange

Nevertheless it is still required by some language servers such as
RedHat's LemMinX and needs to be passed to make them happy.


rwols:
This broke the `test_did_change` test in test_single_document.py. The fix is straightforward though.

Issue #899, opened 2020-03-03
plugin_loaded initializes LSP-* plugins before they have a chance to run plugin_loaded


rchl:
Using `LSP` and `LSP-eslint`.

When `plugin_loaded` function in LSP is called, it initializes all instances of `LanguageHandler` (so one added by `LSP-eslint` for example). The problem is that this runs before `LSP-eslint`'s `plugin_loaded` has a chance to run (the order which one runs first is probably undefined but in my case LSP's runs first).

I want to initialize some paths from `plugin_loaded` (`sublime.cache_path()` that is only valid when called after `plugin_loaded` runs) and so current behavior causes issues.

I was wondering if it would be possible for LSP to trigger a short timeout (for the purpose of running from "next tick" or "next message loop cycle") from `plugin_loaded` and only start initializing after that. It would potentially cause no harm and make sure that all other `plugin_loaded` handlers fully executed (at least that the synchronous code ran there).

rwols:
Perhaps we can lazily instantiate these handlers when they are needed (so when starting a potential Session). This way, I hope we can do without a “next tick” workaround. In fact, I don’t think there is any guarantee that an LSP-* will load in the next cycle of the sublime message loop. Perhaps we can take a peek at how SublimeLinter handles plugins?

rchl:
> I don’t think there is any guarantee that an LSP-* will load in the next cycle of the sublime message loop

rchl:
The code is even "public" so could be checked manually but I don't have time / patience / knowledge to do it ;). But this is small part of the code:
```py

    if len(module_plugins) > 0:
        m.__plugins__ = module_plugins

    if api_ready:
        if "plugin_loaded" in m.__dict__:
            try:
                m.plugin_loaded()
            except:
                traceback.print_exc()
```
(whether `m.__plugins__` contains all plugins initially or there are other cases, I don't know. :))

rchl:
As for initializing lazily, we probably need to get config right away to know for which files/syntaxes it should initialize. So I'm not sure we can do it more lazily than just from short timeout (which could probably also cause timing issues with config not being available on time).

deathaxe:
A LSP-client plugin should propably register and unregister itself by a certain API call within the `plugin_loaded()` and `plugin_unloaded()` functions.

This way LSP could stop all running server instances and unregister its handlers once a client plugin is unloaded. Unloading may be triggered by plugin updates by package control or by disabling the client package.

[...]

Issue #641, opened 2019-07-03
API reference and docs on how to implement a LSP plugin


AmjadHD:
Related to #628, this project should provide an API reference and docs on how to implement a LSP plugin, some of the questions that I have in mind are:
* How to add custom configurations (settings file), and handle updates, should I implement a `ConfigManager` subclass ?
* what's a `DocumentSyncListener` ?
* ...
as I said in #628, this project should focus on being a framework, and leave out language servers support.

tomv564:
The easy question:
DocumentSyncListener listens to view lifecycle events (load, activate, change, save, close) to notify language servers about changes to document state.

About writing docs for LanguageHandlers:
I think working code can be helpful - Raoul gave a link of some implementations in the earlier issue, and there's also https://github.com/tomv564/LSP-rust/ 

If the functionality your server needs is not in the examples, perhaps you can discuss a specific example here and we can come to needed documentation or API changes?

AmjadHD:
how can I get a reference to the view in which a language server is running ?
My use case is: I want to handle RLS's `"window/progress"` notification so I want to show progress details in the statusbar.

predragnikolic:
Hello @AmjadHD,
You can have a look at what @Kronuz did in his fork https://github.com/Kronuz/Rust-CodeIntel/blob/master/plugin.py

tomv564:
Yeah, that's a good example. To save a bit of time, he uses this spinner class https://github.com/Kronuz/SublimeCodeIntel/blob/master/plugin/core/spinner.py which uses `sublime.status_message` and emoji chars to show a spinner.

Since you're not getting progress for a specific document, you could also `set_status` on `sublime.active_window().active_view()`, but I believe that would disappear when the user switches views?

AmjadHD:
Thanks @predragnikolic

> but I believe that would disappear when the user switches views?

I want the progress to only be visible in the views where the server is running.
The problem with using `sublime.active_window().active_view()` (i guess) is when the user switches to a non supported view (where no server is running) in between the notification and it's handling.

ayoub-benali:
@AmjadHD do you have any WIP branch for this issue ? I also want to provide Metals with compilation status bar.

dinvlad:
@AmjadHD could you share a code example on how to enable a new language server? We've got some interest in enabling our [wdl-lsp](https://pypi.org/project/wdl-lsp) for Sublime users..

AmjadHD:
[...]

PR #979, opened 2020-04-22
Shut down subprocesses, more flexible LSP-* API

rwols:
This shuts down the (primary) subprocess in windows on application exit in Sublime Text.

Main ideas:
- Use `weakref.ref` to break cyclic references.
- Wait for threads to stop in `__del__`.
- Wait for the process to shutdown gracefully for one second, otherwise `.terminate()` it.
- Don't use a dict in Client to register notification and request handlers. This creates a lot of cyclic references. Instead, do `getattr` on yourself.
- The above breaks LanguageHandler entirely. I'm figuring out how to fix this.

- ~~Repurpose LanguageHandler as an API transition layer :)~~
- [x] Make at least the npm-based plugins work smoothly
- [x] Test TCP!
- [x] Project switching
- [x] Shutdown at exit
- [x] Major cleanups
- [x] Use `can_start` to show off the new API
- [x] Use `needs_update_or_installation`
- [x] Use `install_or_update` (works wonderfully)
- [x] Try re-writing LanguageHandler as something derived from AbstractPlugin


rwols:
I'm rebasing this onto the st4000-exploration branch. There's many changes and broken tests. Let's take ST4 as the point in time where we introduce all changes at once.

plugin/core/main.py
```diff
@@ -6,9 +6,11 @@
 from .logging import set_debug_logging, set_exception_logging
 from .panels import destroy_output_panels, ensure_panel, PanelName
 from .popups import popups
-from .registry import windows, load_handlers, unload_sessions
+from .registry import windows, unload_sessions
 from .settings import settings, load_settings, unload_settings
 from .typing import Optional
+from .sessions import Session
```
rchl, line 12: Sort by name? :)

plugin/core/main.py
```diff
@@ -6,24 +6,42 @@
...
     set_debug_logging(settings.log_debug)
     set_exception_logging(True)
+    final_subclasses = []  # type: List[Type[Session]]
+    get_final_subclasses(Session.__subclasses__(), final_subclasses)
+    for session_type in final_subclasses:
+        # Unfortunately we need to do this ourselves instead of helper packages registering themselves in plugin_loaded
+        # and unregistering themselves in plugin_unloaded. This is because this base LSP package is loaded before
+        # the helper packages. If the helper packages were loaded before this package, we would not have to do this.
+        # Perhaps we can come up with a scheme to load the helper packages before this package?
+        register_session_type(session_type)
```
rchl, line 41: I'm not sure I understand the reasoning?

When saying one package is loaded before another, you mean the `plugin_loaded` handler, not the importing of the python files itself?

Why would package not be able to register itself regardless if it's called before or after LSP?
It would import some global `register` function from LSP and call it. If LSP hasn't loaded yet, it would be put in the queue and registered later.

rwols: I admit my comments don't explain the problem.

The problem is that at startup we also check which sessions to start for the open views. So we fetch relevant session configs and then start some sessions.

The problem is that when that happens, LSP-* are too late for registering their custom session.

rchl: So the problem is related to current architecture but should be fixable right?
Another matter whether it's fixable within this PR.

rwols: I'm keeping the behavior the same for now. I'm not sure what the solution should be.

plugin/core/rpc.py
```diff
@@ -246,48 +216,36 @@ def send_error_response(self, request_id: Any, error: Error) -> None:
...
+            pass
 
     def send_payload(self, payload: Dict[str, Any]) -> None:
-        if self.transport:
-            message = format_request(payload)
-            self.transport.send(message)
+        try:
+            self.transport.send(payload)  # type: ignore
+        except AttributeError:
+            pass
```
rchl, line 228: Why do we want to silently ignore those?
Maybe at least print debug message?

plugin/core/rpc.py
```diff
@@ -333,26 +282,31 @@ def receive_payload(self, message: str) -> None:
...
-            self.handle_transport_failure()
+    def on_stderr_message(self, message: str) -> None:
+        pass
+
+    def on_transport_close(self, exit_code: int, exception: Optional[Exception]) -> None:
+        self.transport = None
 
     def response_handler(self, response_id: int, response: Dict[str, Any]) -> Tuple[Optional[Callable], Any]:
         handler, error_handler = self._response_handlers.pop(response_id, (None, None))
+        if not handler:
```
rchl, line 293: Is everything still working correctly with sync requests after this change? Those don't add anything to `_response_handlers`, I believe.

rwols: Ah! So that's why the formatting tests failed in test_single_document :)

rwols: Nicely spotted

plugin/core/rpc.py
rwols: `self.transport` is set to `None` when the reader writer pipes stop. It means the langserver crashed or we sent a `shutdown` request followed by an `exit` notification. It may happen that we try to send stuff like debounced `textDocument/didChange` or `textDocument/documentHighlight`.

plugin/core/sessions.py
```diff
@@ -220,24 +275,49 @@ def clear_dotted_value(current: dict, dotted: str) -> None:
...
-                 on_post_initialize: 'Optional[Callable[[Session], None]]' = None,
-                 on_post_exit: Optional[Callable[[str], None]] = None) -> None:
+class Session(Client):
+
+    @classmethod
+    def name(cls) -> str:
+        raise NotImplementedError()
+
+    @classmethod
+    def needs_update_or_installation(cls) -> bool:
```
rchl, line 285: This is not called by the code. As is `install_or_update`.
This will make it impossible to implement NpmClientHandler for new API right?

rchl: Well, `NpmClientHandler` can of course keep doing things how it's doing them currently and handle that itself. So is the plan to implement that later?

plugin/core/sessions.py
```diff
@@ -220,24 +275,49 @@ def clear_dotted_value(current: dict, dotted: str) -> None:
...
+        return False
+
+    @classmethod
+    def install_or_update(cls) -> None:
+        pass
+
+    @classmethod
+    def standard_configuration(cls) -> ClientConfig:
```
rchl, line 293: I'd like to be able to softly transition to this new API by switching out `NpmClientHandler` without changing subclasses of it. The fact that this is a class method rather than an instance method might make it hard, messy, or even impossible.

Previously `config` was an instance property. What do you think about trying to make this an instance class?

plugin/core/sessions.py
rwols: > So is the plan to implement that later?

Yes.

plugin/core/sessions.py
```diff
@@ -220,24 +275,49 @@ def clear_dotted_value(current: dict, dotted: str) -> None:
...
+        raise NotImplementedError()
+
+    @classmethod
+    def adjust_configuration(cls, configuration: ClientConfig) -> ClientConfig:
+        return configuration
+
+    @classmethod
+    def can_start(cls, window: sublime.Window, initiating_view: sublime.View,
```
rchl, line 301: Also not called currently.

plugin/core/sessions.py
rchl: Hmm, we'll need config to create session...

plugin/core/sessions.py
```diff
@@ -220,24 +275,49 @@ def clear_dotted_value(current: dict, dotted: str) -> None:
...
-        self._on_post_initialize = on_post_initialize
-        self._on_post_exit = on_post_exit
         self.capabilities = dict()  # type: Dict[str, Any]
-        self.client = client
         self._workspace_folders = workspace_folders
-        if on_pre_initialize:
-            on_pre_initialize(self)
-        self._initialize()
+        self._progress = {}  # type: Dict[Any, Dict[str, str]]
+        super().__init__(config.name, settings)
```
rchl, line 320: Why is that called? This class doesn't subclass anything (apart from `object`).

plugin/core/sessions.py
```diff
@@ -371,21 +447,31 @@ def _handle_initialize_result(self, result: Any) -> None:
...
-        self.on_request("workspace/configuration", self._handle_request_workspace_configuration)
-        self.on_request("client/registerCapability", self._handle_register_capability)
-        self.on_request("client/unregisterCapability", self._handle_unregister_capability)
+        self.on_initialized()
         if self.config.settings:
-            self.client.send_notification(Notification.didChangeConfiguration({'settings': self.config.settings}))
+            self.send_notification(Notification.didChangeConfiguration({'settings': self.config.settings}))
+        mgr = self.manager()
+        if mgr:
+            mgr.on_post_initialize(self)
```
rchl, line 455: Can use `call_manager`?

plugin/core/sessions.py
```diff
@@ -397,96 +483,93 @@ def _handle_request_workspace_configuration(self, params: Dict[str, Any], reques
...
+
+    def _handle_shutdown_result(self, _: Any) -> None:
+        self.send_notification(Notification.exit())
+
+    def on_transport_close(self, exit_code: int, exception: Optional[Exception]) -> None:
+        super().on_transport_close(exit_code, exception)
+        debug("stopped", self.config.name, "exit code", exit_code)
+        mgr = self.manager()
+        if mgr:
+            mgr.on_post_exit(self, exit_code, exception)
```
rchl, line 575: `call_manager`?

plugin/core/sessions.py
```diff
@@ -397,96 +483,93 @@ def _handle_request_workspace_configuration(self, params: Dict[str, Any], reques
...
-    return session
+        self.state = ClientStates.STOPPING
+        self.on_shutdown()
+        self.send_request(Request.shutdown(), self._handle_shutdown_result, self._handle_shutdown_result)
+
+    def _handle_shutdown_result(self, _: Any) -> None:
+        self.send_notification(Notification.exit())
+
+    def on_transport_close(self, exit_code: int, exception: Optional[Exception]) -> None:
+        super().on_transport_close(exit_code, exception)
```
rchl, line 571: Again, don't know why are you calling `super()`. Must be missing something. :)

rchl:
`__version__` needs to change.

plugin/core/sessions.py
rchl, line 293: **standard**_configuration as opposed to what? Wouldn't it be enough to call it `configuration`?

rchl: I was able to implement everything. Even if it was a bit hacky due to needing to initialize server resources from `standard_configuration`.

plugin/core/sessions.py
rwols: The base class is `Client`.

plugin/core/sessions.py
rwols: The base class is `Client`.

plugin/core/sessions.py
rwols: I actually want to get rid of `call_manager`. It's a temporary solution. Any calls to the `Manager` should use the abstract base class.

plugin/core/sessions.py
rwols: I actually want to get rid of `call_manager`. It's a temporary solution. Any calls to the `Manager` should use the abstract base class.

plugin/core/sessions.py
rwols: I was also playing around with lsp_utils, see my draft pull request there. I was thinking to re-purpose the name LanguageHandler to be a temporary API wrapper class.

plugin/core/sessions.py
rchl: Wow. `Client` nust have looked like `object` to me when I looked at it. :)

plugin/__init__.py
```diff
@@ -1 +1,22 @@
...
+    '__version__',
+    'ClientConfig',
+    'LanguageConfig',
+    'Notification',
+    'read_client_config',
+    'Request',
+    'Response',
+    'Session',
+    'WorkspaceFolder',
+]
```
rchl, line 22: If `NpmClientHandler` would override `__init__`, it would also need to have access to extra types that are arguments of `Session.__init__` (assuming that it would want to properly annotate types). And I guess that would be kinda bad because __init__ wouldn't be necessarily stable. So I think we might actually want public class that we would override, as you've noted in https://github.com/sublimelsp/lsp_utils/pull/8. That would make thing a lot more proper in terms of having a stable API.

rwols: It can be avoided like this:

```python
class MySession(Session):

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        # do init stuff here
```

plugin/core/sessions.py
rwols: I need to implement that still, it’s a draft after all :)

rchl:
I've also just found this comment of mine in the issue about new API:

> LSP-intelephense wants to return some values from `config()` that are dependant on Sublime's `window` for a given file. At that point, there is no window yet though. At least not one that is specific to the project that the server will run on.
> 
> So it might be a good idea to split configuration into two parts. First that returns just `languages` configuration (`syntaxes`, `scopes` and `languageId`) for the LSP to know where it should initialize server and second that is retrieved when server is about to initialize and `window` is known already.
> This would be also good as it would allow delaying some potentially expensive operations until server is about to initialize.
> For example we could have `def languages()` and `def config(window)`

Should it be addressed too?

rchl:
And I'll bring up this again as I think it might still be worth to investigate that:

> > Also, needs_update_or_installation should probably also run in a separate thread as the process of checking if update is needed can be expensive and trigger file access.
> 
> Hmm. This would be somewhat annoying. We would have to start a py thread just for checking whether an update is needed. That means starting an extra short-lived py thread every time the user opens ST. This is IMO bad design.

We could just spawn one extra thread for checking all clients on it. We could keep it open for 30 seconds after starting for example.

It's because I think that doing all that IO can still cause noticeable jerkiness on ST start. Especially when someone, like me, has dozen of them.

rwols:
> And also LSP-vue wants to set tab/indent settings based on the view. And even if view is not feasible, it would be better to at least base those on correct window.

IMO this is the wrong approach. It should iterate the workspace folders (or initiating workspace folder if it doesn't support workspaces) and find:

1) .editorconfig and read the settings from that, or
2) read a potential .sublime-project file and read the settings from that

The fact that tabSize and useTabs are part of the initializationOptions is a problem of Vetur. Those should be coming from `workspace/didChangeConfiguration`. That is because a user can freely change the tabSize while working on a project. The `workspace/didChangeConfiguration` notification should be sent whenever settings change. initializationOptions can only be sent once.
<hr>

I do agree that a split between *applicable languages* (`def languages()`) (a.k.a. document selectors) and *process startup arguments* (`command()`, `env()`, `init_options()`, `experimental_capabilities()`) is useful.

> Should it be addressed too?

Let's do that in another PR :( this is getting pretty big and complicated.

> We could just spawn one extra thread for checking all clients on it. We could keep it alive for 30 seconds after starting for example.
> It's because I think that doing all that IO can still cause noticeable jerkiness on ST start. Especially when someone, like me, has dozen of them.

I understand, but I would still like to *not* spawn a thread for this. Once we can use python 3.8 and asyncio anyone can do any IO they want.

rchl:
> The fact that tabSize and useTabs are part of the initializationOptions is a problem of Vetur. Those should be coming from `workspace/didChangeConfiguration`. That is because a user can freely change the tabSize while working on a project. The `workspace/didChangeConfiguration` notification should be sent whenever settings change. initializationOptions can only be sent once.

vetur supports both so I guess it's not an issue for it. We can initially send wrong values and then from didChangeConfiguration guess something better.

rwols:
More trouble. Some LanguageHandlers have a `config` property, but the `Session` base class already has a `config` instance variable.

This means that sometimes `self.config` is a `property` object, and sometimes `self.config` is a `ClientConfig`.

I'm not sure yet how to solve this.

rchl:
It would kinda make sense to not have to extend the Session directly as then there is always a chance of conflicts in names. Having a separate class interface that would communicate with Session would probably be better for API purposes. And we can then more easily and safely control what is exposed in the API.

But how to do it...?

I can only think right now of something that feels a bit wrong:
 - the new API that you've created would be defined on a new `PluginInterface` (WIP name) class.
 - Every `PluginInstance` would be instantiated and have corresponding `Session` (Session would own `PluginInstance`.
 - `Session` would have to ask `PluginInterface` (getting config, etc.) and `PluginInterface` would have to ask `Session` stuff (setting up listeners, etc.).

So the really bad part is that it's a cyclic dependency.
There must be a better solution to this...

rwols:
I implemented your ideas in 75f9000

plugin/core/sessions.py
```diff
@@ -278,50 +279,189 @@ def clear_dotted_value(current: dict, dotted: str) -> None:
...
+    global _plugins
+    global client_configs
+    try:
+        name = plugin.name()
+        client_configs.remove_external_config(name)
+        _plugins.pop(name, None)
+    except Exception as ex:
+        exception_log("Failed to unregister plugin", ex)
+    finally:
+        client_configs.update_configs()
```
rchl, line 432: Is it easy to let plugin unregister itself now?
That would be needed as when plugin wants to update the server, sessions should be destroyed first so that file handles are freed (important on Windows).

plugin/core/windows.py
```diff
@@ -497,8 +466,17 @@ def start(self, config: ClientConfig, initiating_view: sublime.View) -> None:
         self._window.status_message("Starting " + config.name + "...")
         try:
             workspace_folders = sorted_workspace_folders(self._workspace.folders, file_path)
-            session_type = get_session_type(config.name)
-            session = session_type(self, self._settings, workspace_folders, config)
+            plugin_class = get_plugin(config.name)
+            if plugin_class is not None:
+                if plugin_class.needs_update_or_installation():
+                    # TODO: Run on separate thread somewhere
+                    plugin_class.install_or_update()
```
rchl, line 473: I'll just mark that for myself.
Will be great when we won't even attempt to start the session when server is installing.

plugin/core/windows.py
```diff
@@ -517,6 +495,7 @@ def start(self, config: ClientConfig, initiating_view: sublime.View) -> None:
                 "{}",
                 "Server will be disabled for this window"
             ]).format(config.name, str(e))
+            exception_log("Unable to start {}".format(config.name), e)
```
rchl, line 498: Finally. Was always annoyed that the alert box doesn't show any stack.

plugin/core/windows.py
```diff
@@ -312,9 +306,7 @@ def __init__(
         configs: WindowConfigManager,
         documents: DocumentHandler,
         diagnostics: DiagnosticsStorage,
-        session_starter: Callable,
         sublime: Any,
```
predragnikolic, line 309: Can we remove `sublime` from the constructor? :)

plugin/core/registry.py
```diff
@@ -41,36 +38,8 @@ def has_manager(self) -> bool:
...
-def register_language_handler(handler: LanguageHandler) -> None:
-    debug("received config {} from {}".format(handler.name, handler.__class__.__name__))
-    client_configs.add_external_config(handler.config)
-    if handler.on_start:
-        client_start_listeners[handler.name] = handler.on_start
-    if handler.on_initialized:
-        client_initialization_listeners[handler.name] = handler.on_initialized
-
-
 def client_from_session(session: Optional[Session]) -> Optional[Client]:
```
predragnikolic, line 41: Maybe correct the return type to be `-> Optional[Type[Client]]` if you plan to keep this function.

One question, you probably plan to remove this function in the future?

plugin/core/registry.py
```diff
@@ -122,8 +91,7 @@ def unload_sessions(window: sublime.Window) -> None:
 configs = ConfigManager(client_configs.all)
 client_configs.set_listener(configs.update)
 documents = DocumentHandlerFactory(sublime, settings)
-handlers_dispatcher = LanguageHandlerDispatcher()
-windows = WindowRegistry(configs, documents, start_window_config, sublime, handlers_dispatcher)
+windows = WindowRegistry(configs, documents, sublime)
```
predragnikolic, line 94: :clap: less arguments :)

plugin/core/rpc.py
```diff
@@ -126,42 +111,42 @@ def reset(self) -> None:
...
-class Client(object):
-    def __init__(self, transport: Transport, settings: Settings) -> None:
-        self.transport = transport  # type: Optional[Transport]
-        self.transport.start(self.receive_payload, self.on_transport_closed)
+def print_to_status_bar(error: Dict[str, Any]) -> None:
+    sublime.status_message(error["message"])
+
+
+def method2attr(method: str) -> str:
+    # window/messageRequest -> m_window_messageRequest
```
predragnikolic, line 119: `m_window_messageRequest` is nice, but it would be nicer if it converted `window/messageRequest -> m_window_message_request` instead  the capital `R` :) 

```
>>> s = "m_window_messageRequest"
>>> ''.join('_' + char.lower() if char.isupper() else char
...               for char in s)
'm_window_message_request'
```

plugin/core/sessions.py
```diff
@@ -278,50 +279,189 @@ def clear_dotted_value(current: dict, dotted: str) -> None:
...
+
+    @classmethod
+    def initialization_options(cls) -> Optional[Dict[str, Any]]:
+        """
+        initializationOptions for the initialize request
+        """
+        return None
+
+    @classmethod
+    def default_settings(cls) -> Optional[Dict[str, Any]]:
```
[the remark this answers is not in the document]
predragnikolic: workspace_settings or settings? :)

plugin/core/rpc.py
rwols: I know there's not a request method like this at the moment, but maybe in the future :)

```
>>> s = 'm_window_getHHTPHeader'; ''.join('_' + char.lower() if char.isupper() else char for char in s)
'm_window_get_h_h_t_p_header'
```

plugin/core/registry.py
rwols: Do you mean `-> Session` ?

> One question, you probably plan to remove this function in the future?

Yes :)

plugin/core/windows.py
rwols: I would like to, but not for this PR.

plugin/core/sessions.py
rwols: This is a good point. I will have to investigate this.

plugin/core/rpc.py
rchl: Can we store a map of request to handler (just a string map so it doesn't affect GC)?
This automatic mapping is not that clear how it works from someone looking at implementation of the plugin.

rwols: I don't understand what is complicated about it. I'll add more docs in the AbstractPlugin class... For a request/notification foo/bar, replace non-alphabetic characters with _, and prepend m_.

```python
def m_vscode_content(self, params, request_id) -> None:
    """handles vscode/content request"""
    session = self.weaksession()
    if session:
        session.send_response(request_id, ...)

def m_intelephense_analysisStarted(self, params) -> None:
    """handles intelephense/analysisStarted notification"""
    pass

def m_intelephense_analysisStopped(self, params) -> None:
    """handles intelephense/analysisStopped notification"""
    pass

def m_eslint_openDoc(self, params, request_id) -> None:
    """handles eslint/openDoc request"""
    session = self.weaksession()
    if session:
        session.send_response(request_id, ...)

def m_eslint_status(self, params) -> None:
    """handles eslint/status notification"""
    pass
```

rchl: There is something about it that makes it kinda magic. It would feel better to me to have an imperative API, I think. But I'll resolve it as I don't have any good arguments as to why. :)

plugin/core/registry.py
predragnikolic: `Optional[Session]` is valid. I would go with that option. 

But `Optional[Type[Client]]` is also valid, because that way we are saying this function will return the base class or any class that extends base class (in our case Session extends Client, that is the reason why it is valid). I have read that in the mypy docs for the `Type` type...

But I would just annotate it with Optional[Session]

predragnikolic: `Optional[Session]` is valid. I would go with that option. 

But `Optional[Type[Client]]` is also valid, because that way we are saying this function will return the base class or any class that extends base class (in our case Session extends Client, that is the reason why it is valid). I have read that in the mypy docs for the `Type` type...

But I would just annotate it with Optional[Session]

rwols:
I made this work with eslint, see https://github.com/sublimelsp/lsp_utils/pull/13/files and https://github.com/sublimelsp/LSP-eslint/pull/21 (man, it's so annoying to have so many separate git trees)

jwortmann:
Maybe a bit offtopic here, but could you tell me whether custom requests like `myserver/someMethod` are possible with the current API (master branch)? From a look over some of the LSP-* packages I couldn't find anything related. Thanks.

rchl:
> Maybe a bit offtopic here, but could you tell me whether custom requests like `myserver/someMethod` are possible with the current API (master branch)? From a look over some of the LSP-* packages I couldn't find anything related. Thanks.

Current packages don't initiate any requests as they don't have a need to do that. Due to this, the abstraction I've created in `lsp_utils` doesn't support that. But in general, it's possible as we have access to `Client` and can call `send_request` on it.

plugin/core/rpc.py
rwols: The magic is in the `__getattr__` method of `Session` :)

plugin/core/sessions.py
rwols: This is https://github.com/sublimelsp/LSP/issues/899 and I'm not fixing it in this PR.

rwols: OK, I'll see what I can do in this PR. It doesn't seem that much more work after all. And we might as well break all the things in one go.

jwortmann:
I'm experimenting with how to implement something like a LSP-* package to enable more features of the Julia server. After some time looking at the other LSP-* packages and digging through various files of this repo here, I've more or less got everything to work now, but I'm not sure if I did everything as intended. Therefore I have some suggestions for an improved API and how to make the plugin development easier. I apologize in case I missunderstood how some parts should work, and I'm probably not quite up to date with the st4 branch and the changes from this PR here. And perhaps it's out of scope of this PR, I just write it here because it has "LSP-* API" in its title. 😉 

* Add a section on readthedocs about how to write a LSP-* plugin, or reopen #641 in the meanwhile
* A method `generate_client_config(settings: sublime.Settings) -> ClientConfig`, so that the `config(self) -> ClientConfig` method in LSP-* packages' `LanguageHandler` subclass can be a single line: `return generate_client_config(sublime.load_settings("LSP-*.sublime-settings")`.
* Simple methods to send notifications and requests to the server from a `sublime.TextCommand`, e.g. something like `API.send_notification(view: sublime.View, server_name: str, method_name: str, params: Any) -> bool`. The return value would indicate whether the notification or request was able to be sent (i.e. `False` if view has no session with the given server name).
* Methods `set_status(window: sublime.Window, server_name: str, status_key: str, status_msg: str)` and `erase_status` that sets or updates an additional permanent status message for all views of a session (especially for newly opened files too), next to the already existing server name.
* (Maybe: Various common functions from `views.py`, could be moved to a new file `utils.py`, so that package authors can find them easier. Or just rename the whole file, not all of the functions currently contained in that file are bound to a view anyway.)

<details><summary>plugin.py for reference</summary>
<p>

```python
import os
import sublime
import sublime_plugin

from LSP.plugin.core.handlers import LanguageHandler
from LSP.plugin.core.settings import ClientConfig, LanguageConfig

from LSP.plugin.core.protocol import Request, Notification, Point
from LSP.plugin.core.registry import session_for_view, client_from_session
from LSP.plugin.core.rpc import Client
from LSP.plugin.core.sessions import Session
from LSP.plugin.core.views import text_document_position_params, point_to_offset


def update_env_status(view: sublime.View, env_name: str) -> None:
    if view.match_selector(0, "source.julia"):
        view.set_status("lsp_clients_julia", "Julia env: {}".format(env_name))


class JuliaFileListener(sublime_plugin.EventListener):
    def on_load(self, view: sublime.View) -> None:
        if view.match_selector(0, "source.julia"):
            settings = sublime.load_settings("LSP-julia.sublime-settings")
            env_name = settings.get("env_name")
            if settings.get("show_env_status") and env_name and settings.get("enabled", True):
                update_env_status(view, env_name)


class LspJuliaPlugin(LanguageHandler):
    _window = None

    def __init__(self) -> None:
        super().__init__()

    @property
    def name(self) -> str:
        return "lsp-julia"

    @property
    def config(self) -> ClientConfig:
        settings = sublime.load_settings("LSP-julia.sublime-settings")
        language_config = LanguageConfig(
            settings.get("languageId", ""),
            settings.get("scopes", []),
            settings.get("syntaxes", [])
        )
        client_config = ClientConfig(
            self.name,
            settings.get("command", []),
            settings.get("tcp_port", None),
            settings.get("scopes", []),
            settings.get("syntaxes", []),
            settings.get("languageId", ""),
            [language_config],
            settings.get("enabled", True),
            settings.get("initializationOptions", dict()),
            settings.get("settings", dict()),
            settings.get("env", dict()),
            settings.get("tcp_host", None),
            settings.get("tcp_mode", None),
            settings.get("experimental_capabilities", dict())
        )
        return client_config

    def on_start(self, window) -> bool:
        self._window = window
        return True

    def on_initialized(self, client) -> None:
        settings = sublime.load_settings("LSP-julia.sublime-settings")
        env_name = settings.get("env_name")
        if settings.get("show_env_status") and env_name:
            for view in self._window.views():
                update_env_status(view, env_name)

        # client.on_request("julia/request_method", self._handle_julia_request_method)
        # client.on_notification("julia/notification_method", self._handle_julia_notification_method)


class JuliaActivateEnvironmentCommand(sublime_plugin.TextCommand):
    def is_enabled(self) -> bool:
        return self.view.match_selector(0, "source.julia")

    def run(self) -> None:
        lsp_session = session_for_view(self.view, None)
        if not lsp_session:
            sublime.error_message("The Language Server has not initialized yet.")
            return

        home = os.path.expanduser("~")
        julia_env_home = os.path.join(home, ".julia", "environments")
        julia_env_labels = [f for f in os.listdir(julia_env_home) if os.path.isdir(os.path.join(julia_env_home, f))]
        julia_env_paths = [os.path.join(julia_env_home, f) for f in julia_env_labels]
        self.view.window().show_quick_panel(julia_env_labels, lambda index: self.activate_environment(index, julia_env_labels, julia_env_paths, lsp_session))

    def activate_environment(self, index: int, julia_env_labels: list, julia_env_paths: list, lsp_session: Session) -> None:
        if index == -1:
            return

        env_name = julia_env_labels[index]
        env_path = julia_env_paths[index]

        # send julia/activateenvironment notification
        lsp_client = client_from_session(lsp_session)
        lsp_client.send_notification(Notification("julia/activateenvironment", env_path))

        # update settings
        settings = sublime.load_settings("LSP-julia.sublime-settings")
        julia_bin = settings.get("command", [None])[0] or "julia"
        command = [julia_bin, "--startup-file=no", "--history-file=no", "-e", "using LanguageServer; using LanguageServer.SymbolServer; server=LanguageServer.LanguageServerInstance(stdin,stdout,raw\"{}\",first(Base.DEPOT_PATH)); run(server)".format(env_path)]
        settings.set("command", command)
        settings.set("env_name", env_name)
        sublime.save_settings("LSP-julia.sublime-settings")

        # update status bar
        if settings.get("show_env_status"):
            for view in self._window.views():
                update_env_status(view, env_name)


class JuliaSelectCodeBlockCommand(sublime_plugin.TextCommand):
    def is_enabled(self) -> bool:
        return self.view.match_selector(0, "source.julia")

    def run(self, edit) -> None:
        lsp_session = session_for_view(self.view, None)
        if not lsp_session:
            sublime.error_message("The Language Server has not initialized yet.")
            return

        params = text_document_position_params(self.view, self.view.sel()[0].b)

        # send julia/getCurrentBlockRange request
        lsp_client = client_from_session(lsp_session)
        lsp_client.send_request(Request("julia/getCurrentBlockRange", params), self.handle_response)

    def handle_response(self, response: list) -> None:
        if len(response) > 1:
            a = point_to_offset(Point.from_lsp(response[0]), self.view)
            b = point_to_offset(Point.from_lsp(response[1]), self.view)
            self.view.run_command("lsp_selection_add", {"regions": [(a, b)]})
```
</p>
</details>

rwols:
This is good feedback @jwortmann!

> A method generate_client_config(settings: sublime.Settings) -> ClientConfig, so that the config(self) -> ClientConfig method in LSP-* packages' LanguageHandler subclass can be a single line: return generate_client_config(sublime.load_settings("LSP-*.sublime-settings").

The current PR actually splits it up into smaller functions:

```python
def command(cls) -> List[str]: ...
def languages(cls) -> List[LanguageConfig]: ...
def initialization_options(cls) -> Dict[str, Any]: ...
def settings(cls) -> Dict[str, Any]: ...
```
But so far I haven't really seen an advantage to this. The old (current) approach of returning a `ClientConfig` was okay. In fact, perhaps we can go one step further into that direction and say: return a `sublime.Settings` object and we'll deal with that.

However that cannot currently work because binaries are put into `sublime.cache_path()` usually, which doesn't have a template in `window.extract_variables()`. But we can add that.

There's also LSP-intelephense which uses template strings for `$temp_dir` and `$home` and expands them for the initializationOptions.

I think we can solve these missing template strings easily by adding them...

```python
variables = window.extract_variables()
variables['cache_path'] = sublime.cache_path()
variables['temp_dir'] = tempfile.gettempdir()
variables['home'] = os.path.expanduser('~')
```

And apply `sublime.expand_variables` on `command`, `initializationOptions`, `env`, even `settings`, why not?

> Simple methods to send notifications and requests to the server from a sublime.TextCommand, e.g. something like API.send_notification(view: sublime.View, server_name: str, method_name: str, params: Any) -> bool. The return value would indicate whether the notification or request was able to be sent (i.e. False if view has no session with the given server name).

It's on my todo list. Basically `LspTextCommand` should have the methods 

```python
def sessions(capability_path: str) -> Iterable[Session]: ...
def session(capability_path: str, point: Optional[int]) -> Optional[Session]: ...
```

> Methods set_status(window: sublime.Window, server_name: str, status_key: str, status_msg: str) and erase_status that sets or updates an additional permanent status message for all views of a session (especially for newly opened files too), next to the already existing server name.

Moving towards that in #1069.
It'll be easy to add once this PR and #1069 are merged, because AbstractPlugin (this PR) has access to its Session, and Session has access to all of its views by then.

> Various common functions from views.py, could be moved to a new file utils.py, so that package authors can find them easier. Or just rename the whole file, not all of the functions currently contained in that file are bound to a view anyway.

Yes, views.py has turned into conversions.py.

rchl:
> The current PR actually splits it up into smaller functions:
> 
> ```python
> def command(cls) -> List[str]: ...
> def languages(cls) -> List[LanguageConfig]: ...
> def initialization_options(cls) -> Dict[str, Any]: ...
> def settings(cls) -> Dict[str, Any]: ...
> ```
> 
> But so far I haven't really seen an advantage to this.

I'd remind that at least splitting out `languages` makes sense as LSP needs to know from the start for which languages server should be started. The rest can be delayed until needed in case we need to resolve some envs based on actual `window` that server is started for. We need reference to that actual window so that we can resolve envs properly.

rwols:
Yes delaying expansion is already taken care of (see the create_transport function in transports.py).

plugin/core/sessions.py
rwols: This turns out to be way harder than I thought. It's a gigantic mess of global state that isn't properly RAII-ed in the right places. Let's leave it for another PR.

rwols:
@rchl 5f0c5a5 uses `def configuration() -> sublime.Settings`, and expands many variables like `$cache_path`. It works really well I must say. Example usage here: https://github.com/sublimelsp/LSP-lemminx/pull/5

plugin/core/main.py
```diff
@@ -1,29 +1,131 @@
...
+                return n
+
+            @classmethod
+            def configuration(cls) -> sublime.Settings:
+                if cls.has_regular_name:
+                    settings_filename = "LSP-{}.sublime-settings".format(cls.name())
+                else:
+                    # e.g. "metals-sublime.sublime-settings"
+                    settings_filename = "{}.sublime-settings".format(cls.name())
+                settings = sublime.load_settings(settings_filename)
```
jwortmann, line 74: I think now that returning a `sublime.Settings` object might not be a good choice. The problem if this settings object is created with the `sublime.load_settings()` function, is, that keys from the user settings completely overwrite those which are predefined in the packages' settings file. This is especially problematic for the `"settings": {}` dict, if I want to only change a single value, without loosing all other values. Therefore we should recursively merge dicts from the default and user settings instead.

I adopted the code from LSP-lemminx [here](https://github.com/sublimelsp/LSP-lemminx/blob/master/plugin.py#L24-L47) and changed some details (for example I think lists should not be merged, since it could lead to problems with the "command" setting).

```python
import sublime


def merge_dicts(default: dict, user: dict):
    """ Recursively merges user into default """
    for key, value in default.items():
        if key in user:
            new_value = user[key]
            if isinstance(value, dict) and isinstance(new_value, dict):
                yield key, dict(merge_dicts(value, new_value))
            # elif isinstance(value, list) and isinstance(new_value, list):
            #     yield key, list(set(value) | set(new_value))
            else:
                # overwrite default value with user value
                yield key, new_value
        else:
            yield key, value
    # add all additional items from user
    for key, value in user.items():
        if key not in default:
            yield key, value


def load_settings(base_name: str) -> dict:
    default_settings_json = sublime.load_resource("Packages/{}/{}".format(__package__, base_name))
    user_settings_json = sublime.load_resource("Packages/User/{}".format(base_name))
    default_settings = sublime.decode_value(default_settings_json)
    user_settings = sublime.decode_value(user_settings_json)
    settings = {"enabled": True}
    for key, value in merge_dicts(default_settings, user_settings):
        settings[key] = value
    return settings



# LanguageHandler
def config(self):
    settings = load_settings("LSP-julia.sublime-settings")
    return read_client_config(self.name, settings)
```

We should add something like this here or put it into the lsp_utils, so that LSP-* packages don't have to reinvent the wheel everytime.

rwols: Well, one important property of sublime.Settings objects is that you can get notifications about them when they change. That's an important thing to have to be able to get to a state of this package where we can do hot-reloading when some setting somewhere changed. There's been a lot of users that get confused because they've changed some setting and nothing happens. Only when restarting ST.

Calling `sublime.load_resource` and `sublime.decode_value` is also a waste of CPU cycles because the settings are already in memory.

However I do understand the concern and I suppose we would have to have this merge_dicts function for a while.

The discussion in #1048 should lead to something better where we don't use `sublime.load_resource` / `sublime.decode_value`. I still say a separate `server-settings.sublime-settings` file in Packages/User works best, because:

1) We read the (dotted, since ST4) key-value pairs from `"settings"` from the sublime.Settings object backed by some LSP-foobar.sublime-settings file.
2) For each of those key-value pairs we do `server_settings.get(key, default=value)` where `server_settings` is also a sublime.Settings object but backed by Packages/User/server-settings.sublime-settings. This way the `"settings"` from LSP-foobar.sublime-settings files play the role of sane defaults and also provide the keys to look up in server-settings.sublime-settings. Hope this all makes sense.

rwols: Hmm I realize now that we would still have to use `sublime.load_resource("Packages/User/server-settings.sublime-settings")` because of dotted sections like `cpp.formatterOptions`. We can't do `settings.get("cpp.formatterOptions")` and expect it to match `cpp.formatterOptions.foo` and `cpp.formatterOptions.bar`.

rwols: Let's keep the api as-is and fix outstanding issues in another PR.

Now that dotted-keys notation works in ST4 I'm fairly confident that a separate `Packages/User/server-settings.sublime-settings` for all user overrides will work fine.

rwols:
Some things of note:
- dynamic registration of LSP-foo helper packges based on plugin_loaded / plugin_unloaded was and still is broken due to us iterating over the `__subclasses__()`.
- client_configs.update_configs() should be changed into something that isn't rebuilding the entire set of configs from scratch every time something changed (related to the previous point: we can't have helper packages register themselves due to the slowness of rebuilding the entire set of configs after every registration, a hidden O(N^2) algorithm).
- With the current api of AbstractPlugin, if a user overrides the `"settings"` dict, he/she must override all settings at once. (kind of the same problem as `"default_clients"` and `"clients"` in LSP.sublime-settings).

plugin/core/main.py
```diff
@@ -1,29 +1,132 @@
...
+            handler = language_handler_class()
+            has_regular_name = False
+
+            @classmethod
+            def name(cls) -> str:
+                n = cls.handler.name  # type: ignore
+                if n.lower().startswith('lsp-'):
+                    cls.has_regular_name = True
+                    return n[len('lsp-'):]
+                return n
```
rchl, line 66: That part breaks backward-compatibility with existing project-wide overrides. It will be expected to use `typescript` now, instead of `lsp-typescript`, for example.

It might be acceptable but since all the other changes strive to make everything backward-compatible, it would be shame to have that one thing break.

plugin/core/main.py
```diff
@@ -1,29 +1,122 @@
...
+
+            @classmethod
+            def name(cls) -> str:
+                return cls.handler.name  # type: ignore
+
+            @classmethod
+            def configuration(cls) -> sublime.Settings:
+                file_base_name = cls.name()
+                if file_base_name.startswith("lsp-"):
+                    file_base_name = "LSP-" + file_base_name[len("lsp-"):]
```
rchl, line 67: Maybe utilize `cls.package_name` when exists instead. Kinda feels less hacky and error-prone than special-casing `LSP-` prefix.

rchl: Actually that would be more error-prone, looking at first LSP-* example outside of sublime organization - https://packagecontrol.io/packages/LSP-tagml

PR #981, opened 2020-04-24
Add Dockerfile language server to docs

### LINKED ISSUE #980: Add Dockerfile support
I don't see Dockerfiles mentioned in the list of supported languages in the [`README.md`](https://github.com/sublimelsp/LSP/blob/2773758de44ad051f0ace93667cfdf33b3c4ad69/README.md) file. Can this be added?

My [language server](https://github.com/rcjsuen/dockerfile-language-server-nodejs) is written in TypeScript and is already used by some existing LSP clients.

jwortmann:
Close #980 

Also fix some whitespace and add back a few fenced code blocks after the last item in a list, because the version of Mkdocs that we use with readthedocs otherwise doesn't recognize comments in indented code blocks (in contrary to the latest Mkdocs release when tested locally).
And add some more LSP-xxx packages that are on Package Control now.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -16,11 +16,14 @@ Visit [Langserver.org](https://langserver.org/) or the [list of language server 
 For a few languages you can also find dedicated packages on Package Control, which can optionally be installed to simplify the configuration and installation process of a language server and might provide additional features such as automatic updates for the server:
 
 * [LSP-css](https://packagecontrol.io/packages/LSP-css)
+* [LSP-elm](https://packagecontrol.io/packages/LSP-elm)
 * [LSP-eslint](https://packagecontrol.io/packages/LSP-eslint)
 * [LSP-html](https://packagecontrol.io/packages/LSP-html)
 * [LSP-intelephense](https://packagecontrol.io/packages/LSP-intelephense)
 * [LSP-json](https://packagecontrol.io/packages/LSP-json)
 * [LSP-metals](https://packagecontrol.io/packages/LSP-metals)
+* [LSP-serenata](https://packagecontrol.io/packages/LSP-serenata)
+* [LSP-typescript](https://packagecontrol.io/packages/LSP-typescript)
 * [LSP-vue](https://packagecontrol.io/packages/LSP-vue)
 
 ### Server Configuration<a name="client-config"></a>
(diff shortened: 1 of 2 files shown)


rchl:
I'm planning to create LSP-dockerfile in LSP organization. Maybe better to wait for that?

jwortmann:
I mean it could be usefull to still have the neccessary configurations documented in case someone wants to manually install the server via npm rather than use the LSP-dockerfile package, and as soon as that package is available on Package Control, we can easily just add an additional link for it to the docs.

rchl:
I believe that could be considered duplication but that's my opinion.

docs/index.md
```diff
@@ -241,7 +238,43 @@ dub run dls:bootstrap
...
+  "command": ["docker-langserver", "--stdio"],
+  "enabled": true,
+  "languageId": "dockerfile",
+  "scopes": ["source.dockerfile"],
+  "settings": {
+    "docker": {
+      "languageserver": {
+        "diagnostics": {
+          // string values must be equal to "ignore", "warning", or "error"
+          "deprecatedMaintainer": "warning",
```
rchl, line 263: There are some mixed tabs and spaces in configuration object.

docs/index.md
```diff
@@ -321,49 +355,47 @@ npm install -g flow-language-server
...
-          "scopes": [
-            "source.modern-fortran",
-            "source.fixedform-fortran"
-          ],
-          "syntaxes": [
-            "Packages/Fortran/grammars/FortranModern.sublime-syntax",
-            "Packages/Fortran/grammars/FortranFixedForm.sublime-syntax"
-          ]
-        }
+```json
```
rwols, line 358: Shouldn't this be `js`?

jwortmann: I was using `js` for all code blocks which contain comments (because Mkdocs renders comments as invalid for JSON), and `json` for all code blocks without comments (because in that case the different highlighting colors for keys and values look a bit nicer in my opinion and `json` was used in the majority of cases beforehand). In this case here, there is no comment, so `js` wouldn't be necessary.
I could change all code blocks in the file to `js` for consistency, if you want though.

docs/index.md
jwortmann: Are there? I cannot see any tab here

Unbenannt

The code blocks use 2 spaces indentation per level, like the LSP settings file does, and so that the code blocks don't become too wide at the readthedocs page.
Is this maybe a git config issue?

rchl: hmm, sorry then. I've copied this object from github-rendered view so and pasted into sublime-settings file. Maybe something went wrong along the way.

docs/index.md
rwols: Well, I'm okay with either.

PR #983, opened 2020-04-24
Normalize server panel message newlines

deathaxe:
Some servers return OS specific newlines, which need to be normalized to `\n` for proper display. Otherwise ST prints `<0x0d>` at each eol.

### DIFF
--- a/plugin/core/panels.py
+++ b/plugin/core/panels.py
@@ -110,7 +110,6 @@ class LspUpdateServerPanelCommand(sublime_plugin.TextCommand):
     def run(self, edit: sublime.Edit, prefix: str, message: str) -> None:
         with mutable(self.view):
             message = message.replace("\r\n", "\n")  # normalize Windows eol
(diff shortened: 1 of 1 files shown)


plugin/core/panels.py
```diff
@@ -109,6 +109,8 @@ def run(self, edit: sublime.Edit, characters: Optional[str] = "") -> None:
 class LspUpdateServerPanelCommand(sublime_plugin.TextCommand):
     def run(self, edit: sublime.Edit, prefix: str, message: str) -> None:
         with mutable(self.view):
+            message = message.replace("\r\n", "\n")  # normalize Windows eol
+            message = message.replace("\r", "\n")  # normalize MacOS eol
```
rwols, line 113: These `\r` from macOS seem ancient. I don't think anybody is using macOS 9 anymore?

https://superuser.com/questions/439440/did-mac-os-lion-switch-to-using-line-feeds-lf-n-for-line-breaks-instead-of

PR #984, opened 2020-04-25
Flakiness

rwols:
Yet another attempt at fixing flakiness.

### DIFF
--- a/tests/setup.py
+++ b/tests/setup.py
@@ -1,3 +1,4 @@
+from LSP.plugin.core.documents import DocumentSyncListener
 from LSP.plugin.core.logging import debug
 from LSP.plugin.core.protocol import Notification, Request, WorkspaceFolder
 from LSP.plugin.core.registry import windows
@@ -7,6 +8,7 @@ from LSP.plugin.core.types import ClientConfig, LanguageConfig, ClientStates
 from os import environ
 from os.path import dirname
 from os.path import join
+from sublime_plugin import view_event_listeners
 from test_mocks import basic_responses
 from unittesting import DeferrableTestCase
 import sublime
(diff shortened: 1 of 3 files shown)


tests/setup.py
```diff
@@ -119,37 +112,28 @@ def setUp(self) -> 'Generator':
...
         self.assertTrue(self.wm._configs.syntax_supported(self.view))
         self.init_view_settings()
-        yield lambda: len(self.wm._sessions) > 0
-        sessions = self.wm._sessions.get(self.config.name, [])
-        self.assertEqual(len(sessions), 1)
-        self.session = sessions[0]
+        found_document_sync_listener = False
+        for listener in view_event_listeners[self.view.id()]:
+            if isinstance(listener, DocumentSyncListener):
+                # Bug in ST3?
```
rchl, line 125: We're probably gonna remember this problem well for the rest of our lives but some detailed comment might be nice anyway.

rchl: Basically my interpretation of the problem is that CI runs with ST window not in focus and that makes ST3 not trigger some events like on_load_async, on_activated, on_deactivated. That makes things not properly initialized on opening file (manager missing in DocumentSyncListener).

rwols: Thanks, I've added your sentences as a comment.

tests/setup.py
```diff
@@ -240,5 +221,7 @@ def tearDown(self) -> 'Generator':
 
     def doCleanups(self) -> 'Generator':
         # restore the user's configs
-        client_configs.update_configs()
+        close_test_view(self.view)
```
rchl, line 224: Now `close_test_view` is called both here and from `await_boilerplate_end()`. Maybe latter not needed?

rwols: Well, if the test fails then we still want to close the view, otherwise there will be a "dangling" view that was opened during a test.

rwols: Right, so in `await_boilerplate_end` it would not be needed :)

PR #986, opened 2020-04-25
Fix syntax highlighting (and maybe also logic?) in ST

rchl:


tests/setup.py
```diff
@@ -135,7 +135,7 @@ def setUp(self) -> 'Generator':
         self.session = self.wm.get_session(self.config.name, self.view.file_name())
         self.assertIsNotNone(self.session)
         self.assertEqual(self.session.config.name, self.config.name)
-        yield {"condition": lambda: self.session.state == ClientStates.READY}
+        yield {"condition": (lambda: self.session.state == ClientStates.READY)}
```
rwols, line 138: Because there's no `"timeout"` anymore, why not just write this as

```
yield lambda: self.session.state == ClientStates.READY`
```
(same for above)

rchl: Oh, didn't mean to remove timeout, just PERIOD. But maybe we are OK without TIMEOUT?

PR #989, opened 2020-04-27
Re-enable formatting tests that seems to work fine

rchl:
Seems like those work fine after latest changes.


rchl:
> tests/test_single_document.py:7:1: F401 'unittest' imported but unused

Can't get flake8 to run locally from LSP...

rwols:
Could you merge the latest master into this just to make sure that everything still passes? I removed the last remaining “yield 100” calls.

PR #992, opened 2020-04-28
Use sublime.COMPLETION_FLAG_KEEP_PREFIX for text edits

rwols:
This allows us to handle any text edit.

Also I split up the command completions:
- LspCompleteTextEdit for text edits (complex),
- LspCompleteInsertText for insert text that has additional edits,
- snippet completion for insert text that doesn't have additional edits.

There is also a separate command to handle resolving documentation,
but it's a work in progress. There is only some skeleton code.

- [x] Revisit completion tests

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -260,4 +260,5 @@ class CompletionHandler(LSPViewEventListener):
         resolve(completion_list, items, flags)
 
     def handle_error(self, error: dict, completion_list: sublime.CompletionList) -> None:
+        resolve(completion_list, [])
         sublime.status_message('Completion error: ' + str(error.get('message')))
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -61,7 +61,7 @@ class LspCompleteCommand(sublime_plugin.TextCommand):
 class LspCompleteInsertTextCommand(LspCompleteCommand):
 
     def run(self, edit: sublime.Edit, **item: Any) -> None:
-        insert_text = item["insertText"]
+        insert_text = item.get("insertText") or item["label"]
         if item.get("insertTextFormat", InsertTextFormat.PlainText) == InsertTextFormat.Snippet:
             self.view.run_command("insert_snippet", {"contents": insert_text})
         else:
@@ -228,7 +228,7 @@ class CompletionHandler(LSPViewEventListener):
                 details=st_details)
         else:
             # A snippet completion suffices for insertText with no additionalTextEdits.
-            snippet = item["insertText"]
+            snippet = item.get("insertText") or item["label"]
             if item.get("insertTextFormat", InsertTextFormat.PlainText) == InsertTextFormat.PlainText:
                 snippet = snippet.replace('$', '\\$')
             completion = sublime.CompletionItem.snippet_completion(
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -232,7 +232,7 @@ class CompletionHandler(LSPViewEventListener):
     def handle_response(self, response: Optional[Union[dict, List]], completion_list: sublime.CompletionList,
                         can_resolve_completion_items: bool) -> None:
         response_items = []  # type: List[Dict]
-        flags = sublime.INHIBIT_REORDER
+        flags = 0
         if settings.only_show_lsp_completions:
             flags |= sublime.INHIBIT_WORD_COMPLETIONS
             flags |= sublime.INHIBIT_EXPLICIT_COMPLETIONS
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -53,9 +53,7 @@ class LspCompleteCommand(sublime_plugin.TextCommand):
         additional_edits = item.get('additionalTextEdits')
         if additional_edits:
             edits = [parse_text_edit(additional_edit) for additional_edit in additional_edits]
-            debug('applying additional edits:', edits)
             self.view.run_command("lsp_apply_document_edit", {'changes': edits})
-            sublime.status_message('Applied additional edits for completion')
 
 
 class LspCompleteInsertTextCommand(LspCompleteCommand):
@@ -76,9 +74,12 @@ class LspCompleteTextEditCommand(LspCompleteCommand):
         new_text = text_edit['newText']
         edit_region = range_to_region(Range.from_lsp(text_edit['range']), self.view)
         if item.get("insertTextFormat", InsertTextFormat.PlainText) == InsertTextFormat.Snippet:
-            self.handle_text_edit_snippet(edit, edit_region, new_text)
+            for region in self.translated_regions(edit_region):
+                self.view.erase(edit, region)
+            self.view.run_command("insert_snippet", {"contents": new_text})
         else:
-            self.handle_text_edit_plaintext(edit, edit_region, new_text)
+            for region in self.translated_regions(edit_region):
+                self.view.replace(edit, region, new_text)
         self.handle_additional_edits(item)
 
     def translated_regions(self, edit_region: sublime.Region) -> Generator[sublime.Region, None, None]:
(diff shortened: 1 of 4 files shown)


rwols:
We previously resolved completions for additionalTextEdits, and I believe this was due to some language server in the past. Quoting the spec however:

> The request can only delay the computation of the detail and documentation properties. Other properties like sortText, filterText, insertText, textEdit and **additionalTextEdits must be provided in the textDocument/completion response and must not be changed during resolve**.

I removed this functionality.

plugin/completion.py
```diff
@@ -41,81 +41,68 @@
 }
 
 
-_changes_since_sent = []  # type: List[sublime.TextChange]
-_change_id = None  # type: Any
```
rchl, line 45: Removed globals! Yes please.

plugin/completion.py
```diff
@@ -41,81 +41,68 @@
...
-
-        # import statements, etc. some servers only return these after a resolve.
+    def handle_additional_edits(self, item: Dict[str, Any]) -> None:
         additional_edits = item.get('additionalTextEdits')
         if additional_edits:
-            self.apply_additional_edits(additional_edits)
-        else:
-            self.do_resolve(item)
+            edits = [parse_text_edit(additional_edit) for additional_edit in additional_edits]
+            debug('applying additional edits:', edits)
```
rchl, line 56: Are such logs worth keeping?

plugin/completion.py
```diff
@@ -41,81 +41,68 @@
...
+            self.view.run_command("lsp_apply_document_edit", {'changes': edits})
+            sublime.status_message('Applied additional edits for completion')
```
rchl, line 58: Is it really useful for user to know that?

plugin/completion.py
```diff
@@ -41,81 +41,68 @@
...
+    def translated_regions(self, edit_region: sublime.Region) -> Generator[sublime.Region, None, None]:
+        selection = self.view.sel()
+        primary_cursor_position = selection[0].b
+        for region in reversed(selection):
+            # For each selection region, apply the same removal as for the "primary" region.
+            # To do that, translate, or offset, the LSP edit region into the non-"primary" regions.
+            # The concept of "primary" is our own, and there is no mention of it in the LSP spec.
+            translation = region.b - primary_cursor_position
+            translated_edit_region = sublime.Region(edit_region.a + translation, edit_region.b + translation)
+            yield translated_edit_region
```
rchl, line 93: I wonder about performance differences with using generators as opposed to just return plain list.

I would assume it's less performant as it's basically a function call to get each item. And since we always calculate all items (we don't stop iteration early based on condition in the caller), I'm not sure it's really worth to use generators in such cases.

plugin/completion.py
```diff
@@ -203,48 +188,61 @@ def format_completion(self, item: dict) -> sublime.CompletionItem:
...
-            convert = self.view.text_point_utf16
-            item["native_region"] = (convert(row, start_col_utf16), convert(row, end_col_utf16))
-
         lsp_label = item["label"]
         lsp_filter_text = item.get("filterText")
         lsp_detail = item.get("detail") or ""
+        lsp_detail = html.escape(lsp_detail.replace('\n', ' '))
+
+        if can_resolve_completion_items or "documentation" in item:
+            doc_link = '<a href="subl:lsp_resolve_docs">Docs...</a>'
```
rchl, line 197: Maybe say `Documentation...`? A bit more formal.

plugin/completion.py
```diff
@@ -41,81 +41,68 @@
...
+
+    def handle_text_edit_plaintext(self, edit: sublime.Edit, edit_region: sublime.Region, new_text: str) -> None:
+        for region in self.translated_regions(edit_region):
+            if len(region) == 0:
+                self.view.insert(edit, region.b, new_text)
+            else:
+                self.view.replace(edit, region, new_text)
 
 
 def resolve(completion_list: sublime.CompletionList, items: List[sublime.CompletionItem], flags: int = 0) -> None:
```
rchl, line 108: Only called from one place now. Maybe inline it?

plugin/completion.py
```diff
@@ -170,29 +156,28 @@ def initialize(self) -> None:
...
             self.initialize()
         if not self.enabled:
             return None
-        client = client_from_session(session_for_view(self.view, 'completionProvider', locations[0]))
+        session = session_for_view(self.view, 'completionProvider', locations[0])
+        if not session:
+            return None
+        client = client_from_session(session)
         if not client:
             return None
```
rchl, line 169: Is it possible that there is session but no client?

rwols: Yes, when shutdown request is sent.

rwols: Sorry, I mean when the exit notification has been sent out.

rchl:
I've tried this in LSP-vue and the Docs link shows up but when I click it, the completion popup disappears and nothing else shows up.

rchl:
Actually, something showed up, it just blended into the code and I've missed it.

But apart from that, I see different behavior depending on the height of the completion list. When the list is bigger then it disappears while otherwise, it stays open.

plugin/completion.py
predragnikolic: I think that you dont need this check `if len(region) == 0:`
```diff
-            if len(region) == 0:
-                self.view.insert(edit, region.b, new_text)
-            else:
-                self.view.replace(edit, region, new_text)
+           self.view.replace(edit, region, new_text)
```
Because if you are do  view.replace(edit, Region(4,4), 'text'), it will behave like an insert.

predragnikolic: Also maybe a different approach to make the `handle_text_edit_plaintext` look like `handle_text_edit_snippet`?

```diff
-        for region in self.translated_regions(edit_region):
-            if len(region) == 0:
-                self.view.insert(edit, region.b, new_text)
-            else:
-                self.view.replace(edit, region, new_text)

+        for region in self.translated_regions(edit_region):
+            self.view.erase(edit, region)
+        self.view.insert("insert", {"characters": new_text})
```

predragnikolic: I would suggest inline the `handle_text_edit_snippet` and `handle_text_edit_plaintext` directly in the `run` method so that the LspCompleteTextEditCommand looks like this at the end.
```
class LspCompleteTextEditCommand(LspCompleteCommand):

    def run(self, edit: sublime.Edit, **item: Any) -> None:
        text_edit = item["textEdit"]
        new_text = text_edit['newText']
        edit_region = range_to_region(Range.from_lsp(text_edit['range']), self.view)

        for region in self.translated_regions(edit_region):
            self.view.erase(edit, region)

        if item.get("insertTextFormat", InsertTextFormat.PlainText) == InsertTextFormat.Snippet:
            self.view.run_command("insert_snippet", {"contents": new_text})
        else:
            self.view.run_command("insert", {"characters": new_text})

        self.handle_additional_edits(item)
```

rchl: > Only called from one place now. Maybe inline it?

Just to be clear, I meant the `resolve` function :)

plugin/completion.py
```diff
@@ -41,81 +41,68 @@
 }
 
 
-_changes_since_sent = []  # type: List[sublime.TextChange]
-_change_id = None  # type: Any
+class LspResolveDocsCommand(sublime_plugin.TextCommand):
 
+    def run(self, edit: sublime.Edit) -> None:
+        self.view.show_popup('coming soon (╯°□°)╯︵ ┻━┻', sublime.COOPERATE_WITH_AUTO_COMPLETE)
```
predragnikolic, line 47: @rwols, you cannot just use text like that, you have to style it `'<div style="padding: 10px;">coming soon (╯°□°)╯︵ ┻━┻</div>'` :)

plugin/completion.py
rwols: I accidentally removed the `resolve` call. I inlined the handle_text_edit_snippet and handle_text_edit_plaintext functions. I'm not too keen on calling `self.view.erase` for both plaintext and snippets. A `self.view.replace` for plaintext is much faster than calling an ST command.

plugin/completion.py
rwols: Turned it into `Documentation`. I hope it doesn't take up too much space.

plugin/completion.py
rwols: I *love* generators. The user isn't writing into the text buffer with 1000's of cursors, no?

PR #993, opened 2020-04-28
Remove diagnostic from status when moving away from region

rchl:
Status wasn't cleared when cursor was moved away from region with
diagnostic region so it appeared stuck.

There are still cases where it might not be removed like removing line
with diagnostic region without changing cursor position. It's can be
fixed easily by moving cursor though.

We could also handle on_text_changed_async (ST4) or on_modified_async (ST3)
to make this more reliable but I'm not sure it's worth the gain.


rwols:
How is it possible that this increases the coverage so much?

rchl:
> How is it possible that this increases the coverage so much?

Probably unrelated. Just codeconv being crazy ;). But seriously, maybe code paths executed during testing differ a bit depending on timing.

PR #994, opened 2020-04-28
Use `INHIBIT_REORDER` flag when only showing lsp completions

rchl:
This gives us order of items as defined by the server (alphabetical
or otherwise relevant) instead of seemingly random order from
ST's algorithm.


rwols:
This is probably the more well though-out solution. I had accidentally put that flag into the completion item flags. Then I put it into the completion list flags, but I always enabled it.

PR #995, opened 2020-04-28
Fix hover popup not triggering in some cases when hovering diagnostics

rchl:
With content like:
```
const x = 1

```

and error diagnostic at the end of the first line (missing semicolon),
it was impossible to trigger hover popup with code actions when hovering
end of the line due to `Range.contains`' broken logic.

The `Range.contains` logic was broken when range spanned more than one
line and point's column was larger than range's `end` column.

### DIFF
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -356,8 +356,8 @@ class Range(object):
             (self.end.row > point.row or self.start.col <= point.col <= self.end.col)
 
     def intersects(self, rge: 'Range') -> bool:
-        return rge.start.row <= self.end.row and rge.start.col <= self.end.col and \
-            rge.end.row >= self.start.row and rge.end.col >= self.start.col
+        return self.contains(rge.start) or self.contains(rge.end) or \
+            rge.contains(self.start) or rge.contains(self.end)
 
 
 class ContentChange(object):
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -352,7 +352,7 @@ class Range(object):
         }
 
     def contains(self, point: Point) -> bool:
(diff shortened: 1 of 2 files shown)


plugin/core/protocol.py
```diff
@@ -352,7 +352,8 @@ def to_lsp(self) -> Dict[str, Any]:
         }
 
     def contains(self, point: Point) -> bool:
-        return self.start.row <= point.row <= self.end.row and self.start.col <= point.col <= self.end.col
+        return self.start.row <= point.row and \
+            (self.end.row > point.row or self.start.col <= point.col <= self.end.col)
```
rchl, line 356: Ops, still a bug here. Let me fix.

plugin/core/protocol.py
```diff
@@ -352,7 +352,8 @@ def to_lsp(self) -> Dict[str, Any]:
         }
 
     def contains(self, point: Point) -> bool:
-        return self.start.row <= point.row <= self.end.row and self.start.col <= point.col <= self.end.col
+        return self.start.row <= point.row <= self.end.row and \
+            (self.end.row > point.row or self.start.col <= point.col <= self.end.col)
 
     def intersects(self, rge: 'Range') -> bool:
         return rge.start.row <= self.end.row and rge.start.col <= self.end.col and \
```
rchl, line 359: `intersects` is also buggy...

rchl: Fixed now. Math is hard.

rwols:
Ideally we want to jump into ST-space and get out of LSP-space ASAP for these kinds of calculations. These kinds of checks are trivial with a `sublime.Region`.

PR #1000, opened 2020-04-29
Initialize final subclass for each LanguageHandler

rchl:
Needed to address cases when there is more than one level of
sub-classing. Required for https://github.com/sublimelsp/lsp_utils/pull/5

### DIFF
--- a/plugin/core/handlers.py
+++ b/plugin/core/handlers.py
@@ -1,3 +1,4 @@
+from .logging import debug
 from .types import ClientConfig
 from .typing import List, Callable, Optional, Type
 import abc
@@ -17,9 +18,24 @@ class LanguageHandler(metaclass=abc.ABCMeta):
 
     @classmethod
     def instantiate_all(cls) -> 'List[LanguageHandler]':
-        return list(
-            instantiate(c) for c in cls.__subclasses__()
-            if issubclass(c, LanguageHandler))
+        def get_final_subclasses(derived: 'List[Type[LanguageHandler]]',
+                                 results: 'List[Type[LanguageHandler]]') -> None:
+            for d in derived:
+                d_subclasses = d.__subclasses__()
+                if len(d_subclasses) > 0:
+                    get_final_subclasses(d_subclasses, results)
+                else:
+                    results.append(d)
+
+        subclasses = []  # type: List[Type[LanguageHandler]]
+        get_final_subclasses(cls.__subclasses__(), subclasses)
+        instantiated = []
+        for c in subclasses:
+            try:
+                instantiated.append(instantiate(c))
+            except Exception as e:
+                debug('LanguageHandler instantiation crashed!', e)
+        return instantiated
 
 
 def instantiate(c: Type[LanguageHandler]) -> LanguageHandler:


plugin/core/handlers.py
```diff
@@ -17,9 +17,19 @@ def config(self) -> ClientConfig:
...
-        return list(
-            instantiate(c) for c in cls.__subclasses__()
-            if issubclass(c, LanguageHandler))
+        subclasses = []  # type: List[Type[LanguageHandler]]
+        cls.get_final_subclasses(cls.__subclasses__(), subclasses)
+        return list(instantiate(c) for c in subclasses)
+
+    @classmethod
+    def get_final_subclasses(cls, superclasses: 'List[Type[LanguageHandler]]',
+                             results: 'List[Type[LanguageHandler]]') -> None:
```
rwols, line 26: You don't have to quote types anymore.

plugin/core/handlers.py
```diff
@@ -17,9 +17,19 @@ def config(self) -> ClientConfig:
...
     def instantiate_all(cls) -> 'List[LanguageHandler]':
-        return list(
-            instantiate(c) for c in cls.__subclasses__()
-            if issubclass(c, LanguageHandler))
+        subclasses = []  # type: List[Type[LanguageHandler]]
+        cls.get_final_subclasses(cls.__subclasses__(), subclasses)
+        return list(instantiate(c) for c in subclasses)
+
+    @classmethod
+    def get_final_subclasses(cls, superclasses: 'List[Type[LanguageHandler]]',
```
rwols, line 25: The name of this variable is confusing. You're passing the subcasses / derived_classes, not the superclasses / base_classes.

plugin/core/handlers.py
rchl: Screenshot 2020-04-29 at 20 46 20

plugin/core/handlers.py
```diff
@@ -17,9 +17,19 @@ def config(self) -> ClientConfig:
...
+                             results: 'List[Type[LanguageHandler]]') -> None:
+        for c in superclasses:
+            c_super = c.__subclasses__()
+            if len(c_super) > 0:
+                cls.get_final_subclasses(c_super, results)
+            else:
+                results.append(c)
```
rwols, line 32: Interesting approach, nice idea.

plugin/core/handlers.py
```diff
@@ -17,9 +17,19 @@ def config(self) -> ClientConfig:
 
     @classmethod
     def instantiate_all(cls) -> 'List[LanguageHandler]':
-        return list(
-            instantiate(c) for c in cls.__subclasses__()
-            if issubclass(c, LanguageHandler))
+        subclasses = []  # type: List[Type[LanguageHandler]]
+        cls.get_final_subclasses(cls.__subclasses__(), subclasses)
+        return list(instantiate(c) for c in subclasses)
```
rwols, line 22: Perhaps this is also a good time to put this under a try ... except block because right now, if one instantiation of a faulty language handler fails, *all* language handlers fail :)

plugin/core/handlers.py
rwols: Oh right, I missed that we're in the LanguageHandler class itself. So we have to use a forward-declaration.

plugin/core/handlers.py
rchl: Only that currently derived classes don't really do anything in the constructor (I've started doing a bit more in sublimelsp/lsp_utils#5 but still not so much that it would be likely to crash). Is it worth it really?

rchl: Well OK, I've done a bit more that could crash.

PR #1003, opened 2020-04-29
Optimize incremental text sync

rwols:
Subsequent simple insertions are stitched together.
Let's try handling simple erasures at a later date.

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -1,7 +1,7 @@
 import sublime
 import linecache
 from .protocol import Point, Range, Notification, Request
-from .typing import Optional, Dict, Any, Iterable, List
+from .typing import Optional, Dict, Any, Iterable, List, Sequence
 from .url import filename_to_uri
 
 
@@ -100,15 +100,26 @@ def render_text_change(change: sublime.TextChange) -> Dict[str, Any]:
 
 
 def did_change_text_document_params(view: sublime.View,
-                                    changes: Optional[Iterable[sublime.TextChange]] = None) -> Dict[str, Any]:
+                                    changes: Optional[Sequence[sublime.TextChange]] = None) -> Dict[str, Any]:
     content_changes = []  # type: List[Dict[str, Any]]
     result = {"textDocument": versioned_text_document_identifier(view), "contentChanges": content_changes}
     if changes is None:
         # TextDocumentSyncKindFull
         content_changes.append({"text": entire_content(view)})
-    else:
+    elif len(changes) > 0:
         # TextDocumentSyncKindIncremental
-        for change in changes:
+        stitched_changes = []
+        current = changes[0]
+        for i in range(1, len(changes)):
+            candidate = changes[i]
+            length = len(current.str)
+            if current.b.pt + length == candidate.a.pt and candidate.a.pt == candidate.b.pt:
+                current.str += candidate.str
+            else:
+                stitched_changes.append(current)
+                current = candidate
+        stitched_changes.append(current)
+        for change in stitched_changes:
             content_changes.append(render_text_change(change))
     return result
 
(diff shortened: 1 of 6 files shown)


plugin/core/views.py
```diff
@@ -100,15 +100,26 @@ def render_text_change(change: sublime.TextChange) -> Dict[str, Any]:
...
+    elif len(changes) > 0:
         # TextDocumentSyncKindIncremental
-        for change in changes:
+        stitched_changes = []
+        current = changes[0]
+        for i in range(1, len(changes)):
+            candidate = changes[i]
+            length = len(current.str)
+            if current.b.pt + length == candidate.a.pt and candidate.a.pt == candidate.b.pt:
+                current.str += candidate.str
```
rchl, line 117: Don't we need to update `current.b.pt` after doing that?

plugin/core/views.py
```diff
@@ -100,15 +100,26 @@ def render_text_change(change: sublime.TextChange) -> Dict[str, Any]:
...
-    else:
+    elif len(changes) > 0:
         # TextDocumentSyncKindIncremental
-        for change in changes:
+        stitched_changes = []
+        current = changes[0]
+        for i in range(1, len(changes)):
+            candidate = changes[i]
+            length = len(current.str)
+            if current.b.pt + length == candidate.a.pt and candidate.a.pt == candidate.b.pt:
```
rchl, line 116: I guess doing the latter check would be a bit less expensive.
And also, the `len(current.str)` expression could be inlined in the `if` so that we don't compute length unless `candidate.a.pt == candidate.b.pt`.

tests/test_incremental_sync.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def test_single_insert(self) -> Generator:
+        self.insert_characters("a")
+        yield from self.verify(
+            (0, 0, 0, 0, 0, 'a')
+        )
+
+    def test_neighboring_inserts(self) -> Generator:
+        self.insert_characters("a")
+        yield 40
```
rchl, line 38: I suppose just `yield` would be as effective?
That's assuming that it's enough to break out of the main thread for a moment.

rwols: Turns out no yield is needed at all.

plugin/core/views.py
rwols: No, we don't update `current.b` :) Convince yourself that this is correct.

rwols: Simplifying the TextChange object let's denote it by `(a, b, str)` where `a` and `b` are sublime unicode codepoint offsets into the buffer. If the user types two characters then it'll look like this:

(0, 0, 'a')
(1, 1, 'b')

Stitching this together we should get

(0, 0, 'ab')

rchl: Oh, I was interpreting them as ranges I guess. :)

rwols: Hmm, perhaps my code isn't entirely correct after all

```
(0, 10, 'hel')  # replace (0, 10) with 'hel'
(10, 10, 'l')  # insert 'l'
(11, 11, 'o')    # insert 'o'
```

rwols: I guess the only requirement is

```
current.b.pt == candidate.a.pt and candidate.a.pt == candidate.b.pt
```

rwols:
This trick doesn't work when `current` is a replacement, but we'll get to that. Let's take small steps here.

plugin/core/views.py
rchl: So you've integrated with this bug?

rwols: It's not a bug, it just won't stitch it together.

rwols: So the final changes will look like

```
(0, 10, 'hel')
(10, 10, 'lo')
```
which is OK.

rwols: Ideally it would look like `(0, 10, 'hello')` though

rchl: OK, I thought you wanted to change to a different condition anyway. But you've removed that part of your comment so maybe not. :)

PR #1004, opened 2020-05-01
Fix error on diagnostics 'relatedInformation': null

jwortmann:
Avoid `TypeError: 'NoneType' object is not iterable` in case of `'relatedInformation': None` (Julia server).


rwols:
Thanks!

PR #1005, opened 2020-05-01
Handle all dynamic capabilities

rwols:
It might need more exceptions, but it should be possible to add those in the "exceptions" dict and that's really all there should be to it.

Note that once a user has typed a char, and then the server registers textDocument/completion, then we still won't do completion because of the `self.initialize()` strategy.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -293,6 +293,9 @@ class Session(object):
     def should_notify_did_change_workspace_folders(self) -> bool:
         return bool(self.capabilities.get("workspace", {}).get("workspaceFolders", {}).get("changeNotifications"))
 
+    def should_notify_did_change_configuration(self) -> bool:
+        return "didChangeConfigurationProvider" in self.capabilities
+
     def handles_path(self, file_path: Optional[str]) -> bool:
         if not file_path:
             return False
@@ -383,6 +386,7 @@ class Session(object):
         self.client.send_response(Response(request_id, items))
 
     def _handle_register_capability(self, params: Any, request_id: Any) -> None:
+        send_did_change_configuration = False
         registrations = params["registrations"]
         for registration in registrations:
             method = registration["method"]
@@ -391,9 +395,12 @@ class Session(object):
             set_dotted_value(self.capabilities, capability_path, registration["registerOptions"])
             set_dotted_value(self.capabilities, registration_path, registration["id"])
             if method == "workspace/didChangeConfiguration":
-                # The mega-exceptional case.
-                self.client.send_notification(Notification.didChangeConfiguration({'settings': self.config.settings}))
+                send_did_change_configuration = True  # The mega-exceptional case.
         self.client.send_response(Response(request_id, None))
+        if send_did_change_configuration:
+            # We send the notification after resolving the server request.
+            # This avoids risking confusing the server.
+            self.client.send_notification(Notification.didChangeConfiguration({'settings': self.config.settings}))
 
     def _handle_unregister_capability(self, params: Any, request_id: Any) -> None:
         unregistrations = params["unregisterations"]  # typo in the official specification
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -369,8 +369,7 @@ class Session(object):
         self.on_request("workspace/configuration", self._handle_request_workspace_configuration)
         self.on_request("client/registerCapability", self._handle_register_capability)
         self.on_request("client/unregisterCapability", self._handle_unregister_capability)
-
-        if self.config.settings is not None:
+        if self.config.settings:
             self.client.send_notification(Notification.didChangeConfiguration({'settings': self.config.settings}))
 
         if self._on_post_initialize:
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -129,7 +129,7 @@ def get_initialize_params(workspace_folders: List[WorkspaceFolder], config: Clie
 
 
 # method -> (capability dotted path, optional registration dotted path)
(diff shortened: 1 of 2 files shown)


rwols:
Makes jedi workable:

Screenshot from 2020-05-01 23-00-28

It still doesn't have good diagnostics though.

plugin/core/sessions.py
```diff
@@ -387,10 +385,14 @@ def _handle_request_workspace_configuration(self, params: Dict[str, Any], reques
...
         for registration in registrations:
-            capability_path, registration_path = method_to_capability(registration["method"])
+            method = registration["method"]
+            capability_path, registration_path = method_to_capability(method)
             debug("{}: registering capability:".format(self.config.name), capability_path)
             set_dotted_value(self.capabilities, capability_path, registration["registerOptions"])
             set_dotted_value(self.capabilities, registration_path, registration["id"])
+            if method == "workspace/didChangeConfiguration":
+                # The mega-exceptional case.
+                self.client.send_notification(Notification.didChangeConfiguration({'settings': self.config.settings}))
```
rchl: I think it might be wrong to send it before sending a response to registration.
The server might only register a handler for it after it receives the response to registration.

rchl: BTW. VSCode doesn't send it in this case. Even though eslint registers this notification, vscode doesn't send it at all in the session (without changing VSCode settings at least).

rwols: Fixed in https://github.com/sublimelsp/LSP/pull/1005/commits/c087319b7ef27599cc6de01bec31317587b5ad50

rwols: > BTW. VSCode doesn't send it in this case. Even though eslint registers this notification, vscode doesn't send it at all in the session (without changing VSCode settings at least).

If we apply the same behavior, then we never actually send `didChangeConfiguration`. This is because settings changes from LSP.sublime-settings, view settings, project settings are not propagated into the plugin.

When we get to a point that we do propagate settings changes, we can perhaps change the above behavior to only notify server when settings actually change.

rchl: I'm fine with current behavior but chances are we don't need to send it then.

plugin/core/sessions.py
```diff
@@ -97,6 +128,43 @@ def get_initialize_params(workspace_folders: List[WorkspaceFolder], config: Clie
     return initializeParams
 
 
+# method -> (capability dotted path, optional registration dotted path)
+# these are the EXCEPTIONS. The general rule is: method foo/bar --> (barProvider, barProvider.id)
+_method_to_capability_exceptions = {
```
rchl, line 133: I would rather make it use a constant convention

rwols:
Codecov is a bit weird. Simply uppercasing a global variable makes the coverage go up?

rchl:
My previous theory was that the code is not always the same when testing so some code parts might not always be triggered. If we can force-retest then we could see if that's the case.

plugin/core/sessions.py
```diff
@@ -368,6 +370,9 @@ def _handle_initialize_result(self, result: Any) -> None:
         self.on_request("client/registerCapability", self._handle_register_capability)
         self.on_request("client/unregisterCapability", self._handle_unregister_capability)
 
+        if self.config.settings is not None:
```
rchl, line 373: `settings` are never none because we default to empty Dict. Can you just do an `if` check without explicitly comparing with `None`? Then even Jedi should be happy (maybe).

rwols: Well, jedi won't be happy because it also wants `settings`. It actively requests them with `workspace/configuration` request instead.

Anyway, it doesn't matter much because it keeps on jugging along.

The spec is perhaps too vague about this:
- Must editor notify `workspace/didChangeConfiguration`, always at startup?
- If so, what is the point of registering for `workspace/didChangeConfiguration`?
- If editor always notifies `workspace/didChangeConfiguration`, what is the point of `workspace/configuration` request?

rchl: Current logic also breaks `LSP-vue`. It expects not the get that call at that point apparently.

rchl: Maybe we should:
 - notify at starting if there are non-empty settings
 - notify on configuration change only if server did dynamic registration (of course we won't do that right now anyway).

plugin/core/sessions.py
```diff
@@ -284,27 +393,23 @@ def _handle_request_workspace_configuration(self, params: Dict[str, Any], reques
         self.client.send_response(Response(request_id, items))
 
     def _handle_register_capability(self, params: Any, request_id: Any) -> None:
+        send_did_change_configuration = False
```
rchl, line 396: Remvoe :)

plugin/core/sessions.py
```diff
@@ -148,15 +241,22 @@ def get_capability(self, capability: str) -> Optional[Any]:
...
         return False
 
     def text_sync_kind(self) -> int:
         textsync = self.capabilities.get('textDocumentSync')
         if isinstance(textsync, dict):
-            return int(textsync.get('change', TextDocumentSyncKindNone))
+            change = textsync.get('change', TextDocumentSyncKindNone)
+            if isinstance(change, dict):
+                # dynamic registration
+                return TextDocumentSyncKindIncremental  # or TextDocumentSyncKindFull?
```
rchl, line 258: I wonder about that...

rwols: I would like to test this in the st4000-exploration branch with both jedi and omnisharp and see what happens.

rwols: If it turns out that a dynamic registration of `textDocument/didChange` notifications only means "full sync", then we'll just put full sync here.

rwols:
Thanks for excellent review, @rchl

PR #1006, opened 2020-05-01
Fix logging of sync payload

rchl:
The payload wasn't logged at all (it was always "None") as result was
not returned from handle_response() (because frankly it's not needed
apart from logging).

Also marking sync responses with a symbol that matches sync requests
to be able to spot them better.

### DIFF
--- a/plugin/core/rpc.py
+++ b/plugin/core/rpc.py
@@ -362,10 +362,10 @@ class Client(object):
                 self._sync_request_cvar.notify()
             else:
                 self._deferred_responses.append((handler, result))
-            return (None, None)
+            return (None, result)
         else:  # self._sync_request_result.is_ready()
             self._deferred_responses.append((handler, result))
-            return (None, result)
+            return (None, None)
         if handler:
             return (handler, result)
         elif is_error:
--- a/plugin/core/rpc.py
+++ b/plugin/core/rpc.py
@@ -44,7 +44,7 @@ class Logger(metaclass=ABCMeta):
         pass
 
     @abstractmethod
-    def incoming_response(self, request_id: int, params: Any, blocking: Optional[bool] = False) -> None:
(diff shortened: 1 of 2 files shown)


plugin/core/rpc.py
```diff
@@ -296,7 +296,8 @@ def deduce_payload(
             response_id = int(payload["id"])
             handler, result = self.response_handler(response_id, payload)
             response_tuple = (handler, result, None, None, None)
-            self.logger.incoming_response(response_id, result)
+            blocking = self._sync_request_result.is_ready()
```
rchl, line 299: This condition might be unclear on first sight. I could also compare `response_id` with `request_id` if that would make it clearer.

rwols: You are returning `result` when `is_requesting()` in `response_handler`, but here we check for `is_ready()`. I don't understand this.

rchl: `set()` or `set_error()` is called after `is_requesting()`, before returning. So that makes it `is_ready()`.

rwols: I see, good point!

plugin/core/rpc.py
```diff
@@ -44,7 +44,7 @@ def outgoing_notification(self, method: str, params: Any) -> None:
         pass
 
     @abstractmethod
-    def incoming_response(self, request_id: int, params: Any) -> None:
+    def incoming_response(self, request_id: int, params: Any, blocking: Optional[bool] = False) -> None:
```
rwols, line 47: Since the default is already `False`, it's IMO better for this argument to be of type `bool` than of type `Optional[bool]`.

rwols: In fact there should not be a default argument, because `outgoing_request` doesn't have a default argument either. It's more symmetric if the signatures stay somewhat the same.

PR #1010, opened 2020-05-02
Handle case when registerOptions doesn't exist

rchl:
The code has triggered exception with LSP-lemminx as "registerOptions"
is optional.


rwols:
It triggered an exception, but turns that into a JSON-RPC error response. If that isn’t elegant I don’t know what is ;)

PR #1011, opened 2020-05-03
Add a version number

rwols:
- Send it over the wire for language servers
- Helper packages can use this too (main reason to implement this)

e.g. helper package would write

```
from LSP.plugin import __version__
```


rchl:
Mypy is not entirely happy though:
Screenshot 2020-05-03 at 15 02 48

rwols:
That is pyls-mypy, not mypy.

PR #1012, opened 2020-05-03
Ignore edit if textDocument.version doesn't match current view

rchl:
If version is provided use that to verify that edit can be applied to
current version of the document.

### DIFF
--- a/plugin/core/edit.py
+++ b/plugin/core/edit.py
@@ -1,3 +1,4 @@
+from .logging import debug
 from .typing import List, Dict, Any, Iterable, Optional, Tuple
 from .url import uri_to_filename
 import operator
@@ -13,6 +14,9 @@ def parse_workspace_edit(workspace_edit: Dict[str, Any]) -> Dict[str, List[TextE
             changes[uri_to_filename(uri)] = list(parse_text_edit(change) for change in file_changes)
     if 'documentChanges' in workspace_edit:
         for document_change in workspace_edit.get('documentChanges', []):
+            if 'kind' in document_change:
+                debug('Ignoring unsupported "resourceOperations" edit type')
+                continue
             uri = document_change.get('textDocument').get('uri')
             version = document_change.get('textDocument').get('version')
             text_edit = list(parse_text_edit(change, version) for change in document_change.get('edits'))
--- a/plugin/core/edit.py
+++ b/plugin/core/edit.py
@@ -1,8 +1,9 @@
-from .typing import List, Dict, Any, Iterable, Tuple
+from .typing import List, Dict, Any, Iterable, Optional, Tuple
 from .url import uri_to_filename
 import operator
 
-TextEdit = Tuple[Tuple[int, int], Tuple[int, int], str]
+# tuple of start, end, newText, version
+TextEdit = Tuple[Tuple[int, int], Tuple[int, int], str, Optional[int]]
 
 
 def parse_workspace_edit(workspace_edit: Dict[str, Any]) -> Dict[str, List[TextEdit]]:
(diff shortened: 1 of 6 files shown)


rchl:
Small improvement related to code actions on save feature I'm working on.

plugin/core/edit.py
```diff
@@ -13,19 +14,22 @@ def parse_workspace_edit(workspace_edit: Dict[str, Any]) -> Dict[str, List[TextE
     if 'documentChanges' in workspace_edit:
         for document_change in workspace_edit.get('documentChanges', []):
             uri = document_change.get('textDocument').get('uri')
-            changes[uri_to_filename(uri)] = list(parse_text_edit(change) for change in document_change.get('edits'))
+            version = document_change.get('textDocument').get('version')
```
predragnikolic, line 17: While you are at it, can you add `workspace.workspaceEdit.documentChanges` to the `session.py capabilities`
[Workspace Edit](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspaceEdit)
```
export interface WorkspaceEdit {
	 ...

	/**
	 * Depending on the client capability `workspace.workspaceEdit.resourceOperations` document changes
	 * are either an array of `TextDocumentEdit`s to express changes to n different text documents
	 * where each text document edit addresses a specific version of a text document. Or it can contain
	 * above `TextDocumentEdit`s mixed with create, rename and delete file / folder operations.
	 *
	 * Whether a client supports versioned document edits is expressed via
	 * `workspace.workspaceEdit.documentChanges` client capability.
	 *
	 * If a client neither supports `documentChanges` nor `workspace.workspaceEdit.resourceOperations` then
	 * only plain `TextEdit`s using the `changes` property are supported.
	 */
	documentChanges?: (TextDocumentEdit[] | (TextDocumentEdit | CreateFile | RenameFile | DeleteFile)[]);
}

```

rchl: Done. Thanks.

predragnikolic: And one small note, if we ever in the future add `resourceOperations`,
that means that potentially this code will be partially broken. 

Because then, `document_change` is not guaranteed to be an instance of `TextDocumentEdit`, it can also be  `CreateFile | RenameFile | DeleteFile`.

These types don't have a `textDocument` on them, so it will throw an error when trying to get the version. `.get('version')`

So I think it is better to worry about those things right now than later. :)

predragnikolic: Small typo in 45c4096dff87632f635e7d8a619eaa5c2b2f4a37 :D 
instead of: 
```
   "workspace": {
            ...
            "documentChanges: True
```
should be:
```
"workspace" {
  "workspaceEdit: {
    "documentChanges": True
--

File: plugin/edit.py [hunk 2]
```diff
@@ -45,7 +43,10 @@ def run(self, edit: Any, changes: Optional[List[TextEdit]] = None) -> None:
         if changes:
             last_row, last_col = self.view.rowcol(self.view.size())
             for change in reversed(sort_by_application_order(changes)):
-                start, end, newText = change
+                start, end, newText, version = change
+                if version is not None and version != self.view.change_count():
+                    debug('ignoring edit due to non-matching document version')
+                    continue
                 region = sublime.Region(self.view.text_point(*start), self.view.text_point(*end))
```
rwols, line 50: While you're at it, these should be `text_point_utf16`.

rwols: Ah, this is the master branch, never mind :)

rchl: This is ST3 branch. :)

plugin/core/edit.py
rchl: Ops

plugin/core/sessions.py
```diff
@@ -101,6 +101,9 @@ def get_initialize_params(workspace_folders: List[WorkspaceFolder], config: Clie
                 "dynamicRegistration": True
             },
             "executeCommand": {},
+            "workspaceEdit": {
+                "documentChanges": True,
```
rwols, line 105: Might as well add `"failtureHandling": "abort"` here?

PR #1014, opened 2020-05-03
Support window/workDoneProgress and $/progress notifications

jwortmann:
This is an enhanced version of PR #902 with support for `window/workDoneProgress/create` from protocol version 3.15.0: [Work Done Progress](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workDoneProgress).

* `$/progress` notifications are displayed for a short period with Sublime's `status_message(string)` method. Alternatively we could use `set_status(key, value)` for permanent status messages and `erase_status(key)` on notifications with `"kind": "end"`, but i think the former option might be better for very long background processes from the server to avoid unnecessary distraction.
* `cancellable?: boolean;` in `$/progress` notification is ignored because Sublime's UI has no simple cancel button. `window/workDoneProgress/cancel` would be used for that.

Tested with the Julia server:

screenshot

(This requires to start the server like the following, to make the background indexing and corresponding progress notifications work:)
```json
"command": ["julia", "--startup-file=no", "--history-file=no", "-e", "using Pkg; using LanguageServer; using LanguageServer.SymbolServer; env_path=dirname(Pkg.Types.Context().env.project_file); server=LanguageServer.LanguageServerInstance(stdin,stdout,false,env_path); run(server)"],
```

I've also tried with `rust-analyzer`, but it only sends a single notification with `{'token': 'rustAnalyzer/startup', 'value': {'kind': 'end', 'message': 'rust-analyzer loaded, 1 packages'}}`, which isn't really usefull. Probably because I'm no Rust user and only tried with the syntax_test file from Sublime's Rust package, and the server complains with `rust-analyzer failed to discover workspace, no Cargo.toml found` beforehand.
Perhaps someone knows further servers which support `workDoneProgress` and `$/progress` notifications for testing purposes?

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -316,6 +316,7 @@ class WindowManager(object):
         self._workspace = workspace
         self._workspace.on_changed = self._on_project_changed
         self._workspace.on_switched = self._on_project_switched
+        self._progress_title = str()
 
     def _on_project_changed(self, folders: List[str]) -> None:
         workspace_folders = get_workspace_folders(self._workspace.folders)
@@ -528,9 +529,17 @@ class WindowManager(object):
             "workspace/applyEdit",
             lambda params, request_id: self._apply_workspace_edit(params, session.client, request_id))
 
+        session.on_request(
+            "window/workDoneProgress/create",
+            lambda params, request_id: self._receive_progress_token(params, session.client, request_id))
+
         session.on_notification(
             "textDocument/publishDiagnostics",
             lambda params: self.diagnostics.receive(session.config.name, params))
+
+        session.on_notification(
+            "$/progress",
+            lambda params: self._show_progress(params))
 
         self._handlers.on_initialized(session.config.name, self._window, session.client)
 
@@ -556,6 +565,38 @@ class WindowManager(object):
     def _check_window_closed(self) -> None:
         if not self._is_closing and not self._window.is_valid():
             self._handle_window_closed()
+
+    def _receive_progress_token(self, params: Dict[str, Any], client: Client, request_id: int) -> None:
+        # we ignore progress token for now
+        # token = params.get('token')
+        client.send_response(Response(request_id, None))
+
+    def _show_progress(self, params) -> None:
+        # $/progress: {'token': 'rustAnalyzer/cargoWatcher', 'value': {'kind': 'begin', 'cancellable': False, 'title': "Running 'cargo check'"}}
+        value = params.get('value', {})
+        if value:
+            if value.get('kind') == 'begin':
+                status_msg = value.get('title')  # mandatory
+                self._progress_title = status_msg
+                progress_message = value.get('message')  # optional
+                progress_percentage = value.get('percentage')  # optional
+                if progress_message:
+                    status_msg += ": " + progress_message
+                if progress_percentage:
+                    status_msg += "(" + str(progress_percentage) + "%)"
+                self._sublime.status_message(status_msg)
+            elif value.get('kind') == 'report':
+                progress_message = value.get('message')  # optional
+                progress_percentage = value.get('percentage')  # optional
+                status_msg = self._progress_title
+                if progress_message:
+                    status_msg += ": " + progress_message
+                if progress_percentage:
+                    status_msg += "(" + str(progress_percentage) + "%)"
+                self._sublime.status_message(status_msg)
+            elif value.get('kind') == 'done':
+                # reset progress title
+                self._progress_title = str()
 
     def _handle_window_closed(self) -> None:
         debug('window {} closed, ending sessions'.format(self._window.id()))
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -591,7 +591,7 @@ class WindowManager(object):
             elif value.get('kind') == 'end':
                 self._progress.pop(token, None)  # erase stored title and message for token
 
-    def _progress_string(self, token: str, value: Dict[str, Any]) -> str:
+    def _progress_string(self, token: Any, value: Dict[str, Any]) -> str:
         progress = self._progress.get(token)
         if not progress:
             debug('unknown $/progress token: {}'.format(token))
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -571,22 +571,22 @@ class WindowManager(object):
         client.send_response(Response(request_id, None))
 
     def _handle_progress_notification(self, params: Dict[str, Any]) -> None:
-        token = params.get('token')
+        token = params['token']
         if token not in self._progress:
             debug('unknown $/progress token: {}'.format(token))
             return
-        value = params.get('value')
+        value = params['value']
         if not value:
             return
-        if value.get('kind') == 'begin':
-            self._progress[token]['title'] = value.get('title')  # mandatory
+        if value['kind'] == 'begin':
+            self._progress[token]['title'] = value['title']  # mandatory
             self._progress[token]['message'] = value.get('message')  # optional
             self._sublime.status_message(self._progress_string(token, value))
-        elif value.get('kind') == 'report':
+        elif value['kind'] == 'report':
             self._sublime.status_message(self._progress_string(token, value))
-        elif value.get('kind') == 'end':
+        elif value['kind'] == 'end':
             if value.get('message'):
-                status_msg = self._progress[token]['title'] + ': ' + value.get('message')
+                status_msg = self._progress[token]['title'] + ': ' + value['message']
                 self._sublime.status_message(status_msg)
             self._progress.pop(token, None)
 
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -587,6 +587,10 @@ class WindowManager(object):
                 if status_msg:
                     self._sublime.status_message(status_msg)
             elif value.get('kind') == 'end':
+                if value.get('message'):
+                    status_msg = self._progress_string(token, value)
+                    if status_msg:
+                        self._sublime.status_message(status_msg)
                 self._progress.pop(token, None)  # erase stored title and message for token
 
     def _progress_string(self, token: Any, value: Dict[str, Any]) -> str:
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -316,7 +316,7 @@ class WindowManager(object):
         self._workspace = workspace
         self._workspace.on_changed = self._on_project_changed
         self._workspace.on_switched = self._on_project_switched
-        self._progress_title = dict()  # type: Dict[str, str]
+        self._progress = dict()
 
     def _on_project_changed(self, folders: List[str]) -> None:
         workspace_folders = get_workspace_folders(self._workspace.folders)
@@ -566,32 +566,44 @@ class WindowManager(object):
         if not self._is_closing and not self._window.is_valid():
             self._handle_window_closed()
 
-    def _receive_progress_token(self, params: Dict[str, Any], client: Client, request_id: int) -> None:
-        token = params.get('token')
+    def _receive_progress_token(self, params: Dict[str, Any], client: Client, request_id: Any) -> None:
+        token = params.get('token')  # number | string
         if token:
-            self._progress_title[token] = None
-            client.send_response(Response(request_id, None))
+            self._progress[token] = dict()
+        client.send_response(Response(request_id, None))
 
     def _show_progress(self, params: Dict[str, Any]) -> None:
         token = params.get('token')
         value = params.get('value')
         if token and value:
             if value.get('kind') == 'begin':
-                self._progress_title[token] = value.get('title')  # mandatory
-                self._sublime.status_message(self._progress_string(token, value))
+                if not token in self._progress:
+                    self._progress[token] = dict()
+                self._progress[token]['title'] = value.get('title')  # mandatory
+                self._progress[token]['message'] = value.get('message')  # optional
+                status_msg = self._progress_string(token, value)
+                if status_msg:
+                    self._sublime.status_message(status_msg)
             elif value.get('kind') == 'report':
-                self._sublime.status_message(self._progress_string(token, value))
+                status_msg = self._progress_string(token, value)
+                if status_msg:
+                    self._sublime.status_message(status_msg)
             elif value.get('kind') == 'end':
-                self._progress_title[token] = None
+                self._progress[token] = dict()  # erase stored title and message for token
 
     def _progress_string(self, token: str, value: Dict[str, Any]) -> str:
-        status_msg = self._progress_title.get(token)
-        if not status_msg:
-            return 'unknown $/progress token: {}'.format(token)
+        progress = self._progress.get(token)
+        if not progress:
+            debug('unknown $/progress token: {}'.format(token))
+            return ''
+        status_msg = progress.get('title')
         progress_message = value.get('message')  # optional
         progress_percentage = value.get('percentage')  # optional
         if progress_message:
+            self._progress[token]['message'] = progress_message
             status_msg += ': ' + progress_message
+        elif self._progress[token]['message']:  # reuse last known message if not present
+            status_msg += ': ' + self._progress[token]['message']
         if progress_percentage:
             status_msg += ' (' + str(progress_percentage) + '%)'
         return status_msg
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -577,7 +577,7 @@ class WindowManager(object):
         value = params.get('value')
         if token and value:
             if value.get('kind') == 'begin':
-                if not token in self._progress:
+                if token not in self._progress:
                     self._progress[token] = dict()
                 self._progress[token]['title'] = value.get('title')  # mandatory
                 self._progress[token]['message'] = value.get('message')  # optional
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -539,7 +539,7 @@ class WindowManager(object):
 
         session.on_notification(
             "$/progress",
-            lambda params: self._show_progress(params))
+            lambda params: self._handle_progress_notification(params))
 
         self._handlers.on_initialized(session.config.name, self._window, session.client)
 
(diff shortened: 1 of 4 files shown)


plugin/core/windows.py
```diff
@@ -557,6 +566,48 @@ def _check_window_closed(self) -> None:
...
+                self._progress[token]['message'] = value.get('message')  # optional
+                status_msg = self._progress_string(token, value)
+                if status_msg:
+                    self._sublime.status_message(status_msg)
+            elif value.get('kind') == 'report':
+                status_msg = self._progress_string(token, value)
+                if status_msg:
+                    self._sublime.status_message(status_msg)
+            elif value.get('kind') == 'end':
+                self._progress[token] = dict()  # erase stored title and message for token
```
rwols, line 592: Please use `self._progress.pop(token, None)` here instead, otherwise `len(self._progress)` will grow to infinity.

jwortmann: Thanks, fixed.

plugin/core/windows.py
```diff
@@ -557,6 +566,48 @@ def _check_window_closed(self) -> None:
         if not self._is_closing and not self._window.is_valid():
             self._handle_window_closed()
 
+    def _receive_progress_token(self, params: Dict[str, Any], client: Client, request_id: Any) -> None:
+        token = params.get('token')  # number | string
```
rwols, line 570: For mandatory keys in the spec, I would use `params['token']`. This frees up a lot of clutter and error checking.
If a `KeyError` is raised because of this, then it is the language server's fault. Any exception in any notification handler is caught and handled, so the plugin won't break because of faulty language servers.

For instance, below would become simply `self._progress[params['token']] = dict()`.

plugin/core/windows.py
```diff
@@ -316,6 +316,7 @@ def __init__(
         self._workspace = workspace
         self._workspace.on_changed = self._on_project_changed
         self._workspace.on_switched = self._on_project_switched
+        self._progress = dict()  # type: Dict[Any, Any]
```
rchl, line 319: There is also this case to handle:
> In case the client sets up a progress token in the initialize params (e.g. property workDoneToken) the server is also allowed to use that token (and only that token) using the $/progress notification sent from the server to the client.

plugin/core/windows.py
```diff
@@ -557,6 +566,48 @@ def _check_window_closed(self) -> None:
...
+        client.send_response(Response(request_id, None))
+
+    def _show_progress(self, params: Dict[str, Any]) -> None:
+        token = params.get('token')
+        value = params.get('value')
+        if token and value:
+            if value.get('kind') == 'begin':
+                if token not in self._progress:
+                    self._progress[token] = dict()
+                self._progress[token]['title'] = value.get('title')  # mandatory
```
rchl, line 582: `$/progress` notification is very generic and nothing is mandatory in the `value`.

You are assuming here that progress notifications are all workDoneProgress related.

The proper solution is to only handle progress notifications for tokens that were created through `window/workDoneProgress/create` and ignore any other (as we don't know how to handle them - that would be something for server-specific packages to implement).

Also, since most (if not all) requests inherit from `WorkDoneProgressParams`, that means all of them can provide token for that particular request. That is also not handled (but might not need to be if it's not needed for Julia right now).

rwols: The type of the `value` field depends on what the original request was. In the case of `window/workDoneProgress/create` it is server-initiated. This is, from what I understand, the exception to the rule.

What would generally happen is that the editor does a request to the server, say `textDocument/references`, and passes along a `workDoneToken`.

Then, the server sends a bunch of `$/progress` notifications using the `workDoneToken` supplied by editor. The type of `value` is then always what is assumed here in this diff.

But editor can *also* send a `partialResultToken`. Then, the server is allowed to send partial results via the `$/progress` notification. The `token` would then correspond to the `partialResultToken`, and the type of `value` is whatever the type is of the original request from editor (in this example it would be partial reference results from `textDocument/references`).

So the `$/progress` notification serves two purposes.

But it doesn't matter for this PR.

rchl: Yes, so I'm saying the PR should only handle the `$/progress` requests related to the tokens that were either sent with the request (we don't send those right now) or through  `window/workDoneProgress/create`. Ignore all other progress notifications.

And rename `_show_progress` to, for example, `_handle_work_progress` to make the logic more explicit.

(I wonder what Julia is doing and if that would be enough to support it...)

jwortmann: Yes, this PR can only handle notifications for `workDoneProgress`. As we don't send a `partialResultToken`, this assumption is valid for now. I've added the lines
```python
if token not in self._progress:
    self._progress[token] = dict()
```
for notifications with `"kind": "begin"` so that the following two lines don't throw an error for unknown tokens and this way we could still show notifications if the server somehow forgets to send a `window/workDoneProgress/create` request first. Actually I thought this was the case with `rust-analyzer`, but after checking again I get the following log:
```
:: <-- rust-analyzer window/workDoneProgress/create(1): {'token': 'rustAnalyzer/startup'}
:: >>> rust-analyzer 1: None
:: <-  rust-analyzer $/progress: {'token': 'rustAnalyzer/startup', 'value': {'message': '0/1 packages', 'percentage': 0.0, 'title': 'rust-analyzer', 'kind': 'begin'}}
:: <-  rust-analyzer $/progress: {'token': 'rustAnalyzer/startup', 'value': {'message': 'rust-analyzer loaded, 1 packages', 'kind': 'end'}}
```
Also it still works with the suggestion to use `pop(token, None)` for `"end"` messages: the spec seems a bit unlear, but if I understand it right, it doesn't forbid to send the `window/workDoneProgress/create` request only once, and then reuse that token for multiple sets of notifications with `begin/report/end`.
Should we be more strict and ignore all tokens not submitted via `window/workDoneProgress/create`?

Julia also sends the request first:
```
:: <-- julials window/workDoneProgress/create(faa26dc2-c47f-402d-8490-2fb5d944f63d): {'token': '61cc1844-6988-4178-8aa5-912dbfed892f'}
:: >>> julials faa26dc2-c47f-402d-8490-2fb5d944f63d: None
:: <-  julials $/progress: {'token': '61cc1844-6988-4178-8aa5-912dbfed892f', 'value': {'message': 'Indexing packages...', 'title': 'Julia Language Server', 'kind': 'begin'}}
```

rchl: > Yes, this PR can only handle notifications for `workDoneProgress`. As we don't send a `partialResultToken`, this assumption is valid for now.

I'm arguing that technically `$/progress` can be sent from the server for whatever reason and with whatever value, unrelated to either work progress or partial results. Not sure why it would do that and probably it's not likely that it would that as how would client know how to handle it, but it's possible.

plugin/core/windows.py
```diff
@@ -587,6 +587,10 @@ def _show_progress(self, params: Dict[str, Any]) -> None:
                 if status_msg:
                     self._sublime.status_message(status_msg)
             elif value.get('kind') == 'end':
+                if value.get('message'):
+                    status_msg = self._progress_string(token, value)
+                    if status_msg:
+                        self._sublime.status_message(status_msg)
                 self._progress.pop(token, None)  # erase stored title and message for token
```
rchl, line 594: My thinking is:
 - If it's a token provided through `initialize` then it should live forever
 - If it's a token provided through `window/workDoneProgress/create` then it also should forever (as it's not bound to any specific message)
 - If it's a token bound to specific request then should be removed when message is handled.

rchl: Does Julia or Rust re-create token every time they want to report progress?

I suppose it make sense since you might want to have two different progress states for two long-standing operations. Then It would probably make sense to remove on `end` (for the `window/workDoneProgress/create` case).

rwols: Spec says:

> In case **the client sets up a progress token in the initialize params** (e.g. property workDoneToken) the server is **also allowed** to use that token (and only that token) using the $/progress notification sent from the server to the client.

Meaning that, along with `window/logMessage`, `window/showMessage` and `window/showMessageRequest`, it is also allowed to notify `$/progress` during the initialize phase.

> If it's a token provided through initialize then it should live forever

No, we initiate the allowance of `$/progress` during the special `initialize` phase.

> If it's a token provided through window/workDoneProgress/create then it also should live forever (as it's not bound to any specific message)

No, the server initiates this and ends it with an `end` kind.

IMO this diff is correct. We don't put a workDoneToken in the `initialize` params so server is not allowed to notify `$/progress` during initialize.

jwortmann: Afaik Julia uses progress notification only for background package indexing, which probably is initiated only once after the server starts. I agree with all your statements here. However, in case we want to preserve the tokens from `window/workDoneProgress/create` even after `end` notifications, we should at least erase the stored title and message, so that they don't get reused for new notifications which could omit `"message"` (the spec suggests to reuse the last known `"message"` for `report` notifications if omitted).

rwols: There seems to be some confusion around requests from editor -> server and server initiated progress.

What is implemented here is *server initiated progress*.

A whole other can of worms is *editor initiated progress*.

rwols: The `initialize` request with its workDoneProgress token is not special. You can put a workDoneProgress token in any request and the server can report progress on that request.

rchl: I stand corrected.

But IMO we still should reject all progress notifications that don't match registered token.

rwols: > But IMO we still should reject all progress notifications that don't match registered token.

Agreed.

jwortmann: > But IMO we still should reject all progress notifications that don't match registered token.

OK, I will update this PR later today. But we cannot use `pop()` then on `end` notifications, correct? How would it have to be implemented then, considering @rwols assertion that `len(self._progress)` will grow to infinity with my first implementation?

plugin/core/windows.py
predragnikolic: I agree with @rchl,
The `$/progress` notification is to generic.
```
type ProgressToken = number | string;
interface ProgressParams<T> {
	/**
	 * The progress token provided by the client or server.
	 */
	token: ProgressToken;

	/**
	 * The progress data.
	 */
	value: T;
}
```

We can't assume that the value, will have a `title` or a `message`.
For me the `$/progress` notification should be implemented in `LanguageHandlers` (take [LSP-rust](https://github.com/tomv564/LSP-rust/blob/ecad648dfcebaba6e375338e1848aa3079481e0f/plugin.py#L124-L134) for example, although not the $progress notification, but it is similar).

There are some request/notifications that are to generic to be in the core LSP plugin (code lens is another example, it is also to generic),
and should be in a language handler.

rchl: We've agreed in another thread to handle only "work progress" related notifications which are specified exactly.

plugin/core/windows.py
rchl: > > If it's a token provided through window/workDoneProgress/create then it also should live forever (as it's not bound to any specific message)
> 
> No, the server initiates this and ends it with an `end` kind.

I think we've agreed that we should pop the token on `end`

plugin/core/windows.py
predragnikolic: Talking about this one:
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workDoneProgressBegin

Missed that part...

plugin/core/windows.py
rwols: > But we cannot use pop() then on end notifications, correct?

No it's fine as it is. The entire diff is correct in terms of understanding the spec.

jwortmann: Alright, that assumes that all `begin` notifications must be preceded by their own `workDoneProgress/create` request first (or otherwise use a token that we could send as `workDoneToken` parameter in requests). That makes sense, but it was not super clear for me from the specs. Let's hope all server implementations assume the same, instead of sending the request only once and reusing that token for potential multiple progresses.

The PR is updated now to ignore unknown tokens.

rwols: It's the only correct way to interpret the spec because if servers were allowed to both re-use tokens and create new tokens, there would be no way for an editor to know when to delete data associated with tokens, and thus memory usage would grow indefinitely.

plugin/core/windows.py
jwortmann: Yes, we ignore all notifications with unknown `"token"` values now. Tokens can (for now) only be added by the server via an `window/workDoneProgress/create` request, so that we can assume that all notifications with corresponding tokens follow the implemented conventions for the `"value"` object.

plugin/core/windows.py
```diff
@@ -557,6 +566,50 @@ def _check_window_closed(self) -> None:
...
+                self._progress[token]['title'] = value.get('title')  # mandatory
+                self._progress[token]['message'] = value.get('message')  # optional
+                status_msg = self._progress_string(token, value)
+                if status_msg:
+                    self._sublime.status_message(status_msg)
+            elif value.get('kind') == 'report':
+                status_msg = self._progress_string(token, value)
+                if status_msg:
+                    self._sublime.status_message(status_msg)
+            elif value.get('kind') == 'end':
```
rwols: This is still overly cautious IMO. The `kind` **must** be present, or an editor does not know when to delete data associated with the token. If it's not present then a language server can arbitrarily increase the memory usage of an editor. Just use `value['kind']` and `value['title']`, `params['token']` and `params['value']`, it makes things easier to read.

jwortmann: Ok.

plugin/core/windows.py
```diff
@@ -557,6 +566,43 @@ def _check_window_closed(self) -> None:
...
+    def _receive_progress_token(self, params: Dict[str, Any], client: Client, request_id: Any) -> None:
+        self._progress[params['token']] = dict()
+        client.send_response(Response(request_id, None))
+
+    def _handle_progress_notification(self, params: Dict[str, Any]) -> None:
+        token = params['token']
+        if token not in self._progress:
+            debug('unknown $/progress token: {}'.format(token))
+            return
+        value = params['value']
```
rchl, line 578: `value` is required in the spec.

plugin/core/windows.py
```diff
@@ -557,6 +566,43 @@ def _check_window_closed(self) -> None:
...
+        if not value:
+            return
```
rchl, line 580: What would be the point of some server sending no (or null) value? That sounds like a server issue and probably shouldn't just be ignored?

rchl: Also, non-false'y value doesn't guarantee that it's a Dict so it would crash later anyway.

jwortmann: Well, this is just to avoid errors if the server doesn't follow the spec. Actually I just was following the style of the PR #902 here (https://github.com/sublimelsp/LSP/blob/06bdc86a980d27ac24989ab7daff9398493f5a9a/plugin/core/windows.py#L647).

I'm not sure if we can assume all servers to follow the specs everywhere and just ignore potential error sources. In fact, I recently submitted a small change here in #1004 because the Julia server sends a null value for a diagnostics related item, which resulted in an error in the Sublime console and broke the LSP diagnostics functionality completely. Actually, I think null wasn't even allowed in the specs for that value, so it was the servers fault, but now we are on the safe side for all servers.

rwols: For diagnostics, we should be somewhat careful. But for straightforward notification handlers, these things are wrapped in a giant try ... except block, so any error checking is sort of redundant.

rchl: I was thinking in terms of server developers (having in mind today's discussion on discord) which might want to have some indication if something goes wrong rather than silently ignoring the issue.

But in the end it's a design decision and for project owners to decide. But then @rwols has good points that it should be needed here as it won't break LSP.

plugin/core/windows.py
```diff
@@ -557,6 +566,43 @@ def _check_window_closed(self) -> None:
...
+        if value['kind'] == 'begin':
+            self._progress[token]['title'] = value['title']  # mandatory
+            self._progress[token]['message'] = value.get('message')  # optional
+            self._sublime.status_message(self._progress_string(token, value))
```
rchl, line 584: I'm chiming in late on that, but that status message should probably be bound to the view.
window-bound statuses suck as they override each other if there is more than one.

Those notifications can be used both for file-bound (getting diagnostics) or project-bound (indexing) operations but still I would think view-bound status would be better.

rwols: > window-bound statuses suck as they override each other if there is more than one.

That's a limitation of ST. I think we do the right thing here.

rchl: Yes, but we are catering for ST, not the general population of editors, so we can choose whatever works best rather than ideal solution. :)

rwols: That's true, but server-initiated progress is explicitly per-window as indicated by the `window/` prefix of the `window/workDoneProgress/create` request. If the user closes the view while the server is indexing a massive project for 10 minutes then he/she won't see any more progress.

rwols:
One thing that's possibly problematic is that if you run 2 sessions in a view, and both of them happen to start progress reporting with the same token, then they would overwrite eachother's token in the `self._progress` dict. Ideally all this logic should move to the `Session`. The `WindowManager` does too much and has become a god-class. This is a welcome addition in the right direction nonetheless.

rwols:
clangd sends a `float` as percentage :) @jwortmann can you round floats to at most 1 digit after the comma?

Peek 2020-05-04 21-44

rwols:
I'm wondering whether `number` in the spec means `int` or `int | float` now...

rchl:
> I'm wondering whether `number` in the spec means `int` or `int | float` now...

Since it's JSON (although spec is described in TS but it's still the same for Number), Number represents both.

rwols:
OK. Made a PR with the simple fix here: #1019

jwortmann:
> One thing that's possibly problematic is that if you run 2 sessions in a view, and both of them happen to start progress reporting with the same token, then they would overwrite eachother's token in the `self._progress` dict.

I would think that the purpose of the token is to be a unique identifier for a specific progress, right? That is why each (server initiated) progress needs to send a `window/workDoneProgress/create` request with a token first. Julia indeed dynamically generates a new UUID for each progress. On the other hand, rust-analyzer always sends `rustAnalyzer/startup` which is certainly not the best implementation on the server's side.

If/when we want to implement to send a `workDoneToken` for possible progress reports within some of our requests, we should also dynamically generate new tokens for each request to avoid interference between concurrent progress reports.

PR #1016, opened 2020-05-04
Extend test server with $test/sendNotification request handler

rchl:
Part of my work on code actions on save where I will need that functionality to popular document diagnostics. Also might be useful for testing $/progress notifications, I imagine.

Allows to trigger any notification from tests.

For example if we want the server to send a 'textDocument/publishDiagnostics'
notification to populate view's diagnostics we'd do:

yield from self.await_client_notification('textDocument/publishDiagnostics', {'params': {}})

On the server side this is a request handler and not a notification handler
as I wanted to be able to await the result so that I would have guarantee
that the notification has been handled when "await_client_notification" call
returned.

### DIFF
--- a/tests/server.py
+++ b/tests/server.py
@@ -283,11 +283,23 @@ class Session:
         self._on_notification("exit", self._on_exit)
 
         self._on_request("$test/getReceived", self._get_received)
+        self._on_request("$test/sendNotification", self._on_send_notification)
         self._on_notification("$test/setResponse", self._on_set_response)
 
     async def _on_set_response(self, params: PayloadLike) -> None:
         if isinstance(params, dict):
             self._responses[params["method"]] = params["response"]
+
+    async def _on_send_notification(self, params: PayloadLike) -> None:
+        if not isinstance(params, dict):
+            raise Error(ErrorCode.InvalidParams, "expected params to be a dictionary")
+        if "method" not in params:
+            raise Error(ErrorCode.InvalidParams, 'expected "method" key')
+        method = params["method"]
+        if not isinstance(method, str):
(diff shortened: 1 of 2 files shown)


tests/server.py
```diff
@@ -283,12 +283,24 @@ def _install_handlers(self) -> None:
...
 
+    async def _on_send_notification(self, params: PayloadLike) -> None:
+        if not isinstance(params, dict):
+            raise Error(ErrorCode.InvalidParams, "expected params to be a dictionary")
+        if "method" not in params:
+            raise Error(ErrorCode.InvalidParams, 'expected "method" key')
+        method = params["method"]
+        if not isinstance(method, str):
+            raise Error(ErrorCode.InvalidParams, 'expected "method" key to be a string')
+        self._notify(method, params.get('params'))
```
rwols, line 301: This does not actually send out the notification before returning from the request. It schedules it on the runloop so it will be sent out later. If you want to be sure of this, you could make the `_log`, `_notify`, `_reply` and `_error` methods `async` and they can `await` on `_send_payload`. And then do `await self._notify(...)`. That would guarantee that the notification is sent before returning here.

If you don't care about this, then I'm okay with how it looks :)

Please update the comments at the top: remove the TODO about sending notifications and explain how to create fake notifications.

rchl: > This does not actually send out the notification before returning from the request. It schedules it on the runloop so it will be sent out later.

True, but the response to the request is also scheduled on the runloop. So given the order they are scheduled, they should arrive in that order (request response after notification response), I would think.

rwols: > True, but the response to the request is also scheduled on the runloop.

You're right.

PR #1017, opened 2020-05-04
Check changes to apply against initial view version (change count)

rchl:
With multiple changes to apply, first change would change view version
and later changes would not longer apply.

I've played myself here before. This really made code-actions-on-save unusable. :)

### DIFF
--- a/tests/test_code_actions.py
+++ b/tests/test_code_actions.py
@@ -1,0 +1,60 @@
+from LSP.plugin.core.protocol import Point, Range
+from LSP.plugin.core.url import filename_to_uri
+from LSP.plugin.core.views import entire_content
+from LSP.plugin.code_actions import run_code_action_or_command
+from setup import TextDocumentTestCase
+from test_single_document import TEST_FILE_PATH
+
+try:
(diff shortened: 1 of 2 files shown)


rchl:
I'm gonna try to add some tests for this (since I have done similar things already for CAOS*)

* code-actions-on-save

tests/test_code_actions.py
```diff
@@ -0,0 +1,60 @@
...
+from LSP.plugin.core.views import entire_content
+from LSP.plugin.code_actions import run_code_action_or_command
+from setup import TextDocumentTestCase
+from test_single_document import TEST_FILE_PATH
+
+try:
+    from typing import Generator, Optional, Iterable, Tuple, List, Dict
+    assert Generator and Optional and Iterable and Tuple and List and Dict
+except ImportError:
+    pass
```
rwols, line 12: Please don't use these `try import ...` idioms anymore. You can use `from LSP.plugin.core.typing import Generator, Optional, Iterable, Tuple, List, Dict` at the top of the file. You can also remove quotes from types.

Issue #1008, opened 2020-05-02
Workspace Symbol command should accept an empty query string, since 3.15


genericptr:
The "LSP: Workspace Symbol" command currently requires you enter a query string before receiving any results and thus the command does not function similar to the built-in "Find Symbol in Project" command. 

It is my understanding of the LSP that Workspace Symbol is basically "Goto Symbol in Project" so it should (at least by way of option) allow an empty query string so that it can be implemented in this way.

The spec also allows this according to the comment below:

```
/**
 * The parameters of a Workspace Symbol Request.
 */
interface WorkspaceSymbolParams extends WorkDoneProgressParams, PartialResultParams {
	/**
	 * A query string to filter symbols by. Clients may send an empty
	 * string here to request all symbols.
	 */
	query: string;
}
```

rwols:
Pretty much blocked on https://github.com/sublimehq/sublime_text/issues/3338

genericptr:
It's been so long I forget now, but completion panels now work with partial results so is there anything new that would allow symbols to be recalculated as you type? Workspace Symbols are basically dead in the water until that is possible. Sending empty strings is actually the smaller of the 2 issues.

PR #1018, opened 2020-05-04
Allow empty query in  workspace/symbol

ayoub-benali:
Since LSP 3.15 workspace/symbol should allow sending empty query

https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#workspace_symbol

fix https://github.com/sublimelsp/LSP/issues/1008

### DIFF
--- a/plugin/workspace_symbol.py
+++ b/plugin/workspace_symbol.py
@@ -64,9 +64,8 @@ class LspWorkspaceSymbolsCommand(LspTextCommand):
         return SymbolQueryInput()
 
     def run(self, edit: Any, symbol_query_input: str = "") -> None:
-        if symbol_query_input:
-            request = Request.workspaceSymbol({"query": symbol_query_input})
-            client = self.client_with_capability('workspaceSymbolProvider')
-            if client:
-                self.view.set_status("lsp_workspace_symbols", "Searching for '{}'...".format(symbol_query_input))
-                client.send_request(request, lambda r: self._handle_response(symbol_query_input, r), self._handle_error)
+        request = Request.workspaceSymbol({"query": symbol_query_input})
+        client = self.client_with_capability('workspaceSymbolProvider')
+        if client:
+            self.view.set_status("lsp_workspace_symbols", "Searching for '{}'...".format(symbol_query_input))
+            client.send_request(request, lambda r: self._handle_response(symbol_query_input, r), self._handle_error)
--- a/plugin/workspace_symbol.py
+++ b/plugin/workspace_symbol.py
@@ -11,7 +11,7 @@ from .symbols import format_symbol_kind
 class SymbolQueryInput(sublime_plugin.TextInputHandler):
 
     def placeholder(self) -> str:
-        return "Symbol"
+        return "Symbol (leave empty to request all symbols)"
 
 
 class LspWorkspaceSymbolsCommand(LspTextCommand):
--- a/plugin/workspace_symbol.py
+++ b/plugin/workspace_symbol.py
@@ -24,24 +24,30 @@ class LspWorkspaceSymbolsCommand(LspTextCommand):
         name = "{} ({}) - {} -- {}".format(s['name'], symbol_kind, s.get('containerName', ""), file_name)
         return name
 
-    def _open_file(self, symbols: List[Dict[str, Any]], index: int) -> None:
+    def _open_file(self, symbol: Dict[str, Any]) -> None:
+        start = symbol['location']['range']['start']
+        file_name = uri_to_filename(symbol['location']['uri'])
+        encoded_file_name = "{}:{}:{}".format(file_name, start['line'], start['character'])
+        window = self.view.window()
+        if window:
+            window.open_file(encoded_file_name, sublime.ENCODED_POSITION)
+
+    def _on_selected(self, symbols: List[Dict[str, Any]], index: int) -> None:
         if index != -1:
             symbol = symbols[index]
-            start = symbol['location']['range']['start']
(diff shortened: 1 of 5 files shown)


plugin/workspace_symbol.py
```diff
@@ -10,9 +10,6 @@
 
 class SymbolQueryInput(sublime_plugin.TextInputHandler):
 
-    def validate(self, txt: str) -> bool:
-        return txt != ""
-
     def placeholder(self) -> str:
         return "Symbol"
```
rchl, line 14: Maybe placeholder could indicate that empty string is allowed. Something like `Symbol (leave empty to request all symbols)`

rwols:
It's weird that there's no additional server capability for this. How can we detect servers that support v3.15 of the protocol versus servers that support < 3.15?

rwols:
Can we take this opportunity to start a discussion about workspace/symbol?

Perhaps we can do the request with an empty query, always. This way we utilize ST’s fuzzy string matching better, and it would feel more like “Goto Symbol In Project” (ctrl+shift+R)

Meaning: just like we “take over” Goto definition with F12, we may “take over” Goto Symbol In Project with ctrl+shift+R and it should act/present approximately the same.

rchl:
> Perhaps we can do the request with an empty query, always.

Just quick thoughts without having much experience with that functionality:
1) Would we be able to just directly to symbol, without showing the list, if there is only one matching symbol?
2) I understand that `workspace/symbol` is not supported by that many servers. So would it be feasible to use it instead of `textDocument/definition`? Or if not, would it be possible to fallback to it if `workspace/symbol` doesn't work?

ayoub-benali:
> Perhaps we can do the request with an empty query, always.

I don't know about the other servers but for Metals the search is done across the "classpath": project code, the dependencies source code, the dependencies of the dependencies etc. So you can see that sending an empty query can produce a big latency, a quite huge result list and I have no clue how servers and sublime would react to all project symbols.

> Would we be able to just directly jump to symbol, without showing the list, if there is only one matching symbol?

Good idea

> I understand that workspace/symbol is not supported by that many servers. So would it be feasible to use it instead of textDocument/definition? Or if not, would it be possible to fallback to it if workspace/symbol doesn't work?

`textDocument/definition` and `workspace/symbol` don't have the same cost and latency for servers and they don't mean the same thing semantically so I would keep distinct.

ayoub-benali:
> It's weird that there's no additional server capability for this. 

The 3.15 spec mentions `WorkspaceSymbolClientCapabilities`  but for me it looks the same as < 3.15.

> How can we detect servers that support v3.15 of the protocol versus servers that support < 3.15?

The only difference seems to be allowing empty query so no need to detect 3.15 servers, or an I missing something ?

rwols:
It might be possible that servers that support < 3.15 will assume the query-string is non-empty (and might crash if provided with an empty query string). But I might be splitting hairs here.

> textDocument/definition and workspace/symbol don't have the same cost and latency for servers and they don't mean the same thing semantically so I would keep distinct.

Indeed. `textDocument/definition` is a local (=easy) problem, `workspace/symbol` is a global (=hard) problem.

plugin/workspace_symbol.py
```diff
@@ -27,24 +24,30 @@ def _format(self, s: Dict[str, Any]) -> str:
...
         if response:
             matches = response
-            choices = list(map(lambda s: self._format(s), matches))
-            window = self.view.window()
-            if window:
-                window.show_quick_panel(choices, lambda i: self._open_file(matches, i))
+            if len(matches) == 1:
+                self._open_file(matches[0])
+            else:
+                choices = list(map(lambda s: self._format(s), matches))
```
rwols, line 47: You can omit the lambda: `list(map(self._format, matches))`

rwols:
Now that you're busy in this file anyway, perhaps you can return a `List[str]` from `_format` instead of a `str`. The `show_quick_panel` method can also take a `List[List[str]]` (instead of a `List[str]` as we pass it now). This makes the quick panel show items with three rows. The lower two rows have a small font. I think you can stash all of the semantic info there. That might be more informative. Because right now I don't quite understand what some of the results mean when I use this feature.

I skimmed through #501 but I did not find a reason for omitting this feature, perhaps we just forgot about it?

rwols:
I've tested this locally and it doesn't actually run with an empty query string :) That's because there's still an `if` check in the `run` method.

ayoub-benali:
Good catch thanks

rwols:
I'm doubting whether what we do here makes any sense at all. For instance, clangd may sends hundreds of megabytes of JSON data when doing an empty query request.

What probably needs to happen is to use a `partialResultToken` to enable partial results (and a `workDoneProgressToken` while we're at it). Then, the server starts streaming results to us, and the user starts typing at the same time. It would remove some of the enormous wait time.

This is a complicated request.

rwols:
I've been chatting in the clangd Discord and they're telling me that partial results would still kind of suck, because the editor cannot provide feedback on what the user is typing. So a server might stream useless chunks of data (e.g. the user has typed `foo::` while the server streams thousands of tokens that start with `bar::`). But it would be a step in the right direction.

ayoub-benali:
> I'm doubting whether what we do here makes any sense at all. For instance, clangd may send hundreds of megabytes of JSON data when doing an empty query request.

This what I meant earlier about sending the command always with empty query.
For my usage it doesn't make send `workspace/symbol` with empty param but this PR is about allowing following the LSP spec. 

If you think it doesn't make sense to have such a feature let's close this PR and the related issue.

rwols:
Let's revisit this problem when we have `workDoneProgress` and `partialResult` for requests.

Issue #991, opened 2020-04-28
Support for MarkupContent in hover


rchl:
[MarkupContent](https://microsoft.github.io/language-server-protocol/specification#markupContent) is an object with required `kind` and `value` properties where `kind` can be either `plaintext` or `markdown`.

It's allowed in:
 - `textDocument/completion` response `CompletionItem.documentation` ([link](https://microsoft.github.io/language-server-protocol/specification#textDocument_completion))
 - `textDocument/hover` response `Hover.contents` ([link](https://microsoft.github.io/language-server-protocol/specification#textDocument_hover))
 - `textDocument/signatureHelp` response: `SignatureHelp.signatures[].documentation?` ([link](https://microsoft.github.io/language-server-protocol/specification#textDocument_signatureHelp))

We are currently not supporting it in hover popups for sure and probably in other places too. And while we read the actual `value` from the contents, we don't make a proper choice regarding escaping, rendering as markdown due to not taking `kind` into account.

genericptr:
Here is an example hover response which shows the bug:

`{'contents': {'value': 'type TVec2i = specialize TGVec2<Integer>', 'kind': 'plaintext'}, 'range': {'end': {'character': 3, 'line': 339}, 'start': {'character': 3, 'line': 339}}}`

and a picture of how it is displayed:

![Screen Shot 2020-04-28 at 4 46 22 PM](https://user-images.githubusercontent.com/10937186/80486501-784a1c80-8985-11ea-83ed-5f530a150fa6.png)

rchl:
@rwols @predragnikolic do you know if this is now fully implemented in all contexts where `MarkupContent` is allowed?

predragnikolic:
We now have the `minihtml` function which accepts a string or a MarkupKind.

We can now use that function when parsing the content in hover or signature help popups.

rwols:
Fixed by #1046.

PR #1022, opened 2020-05-05
Show documentation in cooperate popup

predragnikolic:
Show detail(if not already shown in the AC) and documentation in cooperate popup.

Here is an example: 

output

Instead of the text link  `Documentation` in sublime details, 
there is a blue arrow pointing up. When clicked it will show the docs next to the AC. (Why arrow instead of `Documentation`... I thinks it is less distracting. Takes less space and leaves room to show details next to it).
You can also use `F12` to show the documentation
If they want to remap the keybinding, change this:
```
{ "keys": ["f12"], "command": "auto_complete_open_link", "context":
    [
        { "key": "auto_complete_visible", "operator": "equal", "operand": true },
    ]
},
```

Changes were made to whats gets shown in the annotation in the AC.
Before: the annotation could be either a `Label` or `Detail` or `Nothing`.
Now: the annotation can be either a  `Label or Nothing`.  `Detail` is removed because it can make the AC pretty wide and I think that Sublime details is a more suitable place for it.  

> note `detail` in the cooperate popup are visually different from `documentation`. They have a border and darker bg. 

If LSP `completionItem.detail` or `completionItem.documenation` is not present, 
and the Language Server supports resolving.
a completion `completionItem/resolve` request if fired 
and if the response contain `detail` `documentation` we will show them in the cooperate popup. Note that the details wont be shown in the cooperate popup if they are already shown in the sublime completion item details.

TODO:
Add documentation for the keybinding.
Listen for feedback.


predragnikolic:
I tired implementing that the documentation popups automatically when you trigger the AC, when you use the arrows, etc... but I couldn't make the UX right.  

Butttt, I was pleasantly surprised to discover that I now like more showing the the docs popup only when I trigger it.
I trigger it only when I really care and want to see documentation about a certain completion item.

At the end I can say that I prefer showing the documentation on demand, rather then shown it automatically all the time. (Not saying that there shouldn't be an option for that. Some people would find that option useful)

plugin/completion.py
```diff
@@ -40,11 +41,100 @@
 }
 
 
+class LspAutoCompleteListener(sublime_plugin.ViewEventListener):
```
rwols, line 44: This listener has no state. Just put the callbacks in `CompletionHandler`. Generally I would like to move towards a single ViewEventListener for the whole plugin.

plugin/completion.py
```diff
@@ -181,22 +282,23 @@ def format_completion(self, item: dict, can_resolve_completion_items: bool) -> s
...
-            st_annotation = lsp_detail
-            st_details = doc_link
+            st_annotation = ""
+
+        st_details = ""
+        if can_resolve_completion_items or "documentation" in item:
+            args = {
+                "item": deep_escape(item)
+            }
+            st_details += "<a href='subl:lsp_resolve_docs {}'>Δ</a>".format(sublime.encode_value(args))
```
rwols, line 299: How about this: Store the `response_items` from `def handle_response` in the listener (`self.response_items = ...`)

Then, encode a simple integer here as argument. I think you can use a simple string-format then:

```
'{{"i":{}}}'.format(index)
```
Then, in LspResolveDocsCommand, you can fish up the listener with 

```
for listener in sublime_plugin.view_event_listeners[self.view.id()]:
    if isinstance(listener, CompletionHandler):
        item = listener.response_items(index)
        # ... do stuff with item
```
This way we avoid calling `html.escape` for each completion item that has docs.

plugin/completion.py
```diff
@@ -181,22 +282,23 @@ def format_completion(self, item: dict, can_resolve_completion_items: bool) -> s
...
             st_annotation = lsp_label
-            st_details = '{} {}'.format(doc_link, lsp_detail) if doc_link else lsp_detail
         else:
             st_trigger = lsp_label
-            st_annotation = lsp_detail
-            st_details = doc_link
+            st_annotation = ""
+
+        st_details = ""
+        if can_resolve_completion_items or "documentation" in item:
```
rwols, line 295: It turns out that some language servers send completion items with `{..., "documentation": null, ...}`. We should check this instead:

```
if can_resolve_completion_items or item.get("documentation"):
```

plugin/completion.py
```diff
@@ -101,6 +191,17 @@ def resolve(completion_list: sublime.CompletionList, items: List[sublime.Complet
...
 
+def deep_escape(what):
+    if isinstance(what, dict):
+        return {key: deep_escape(value) for key, value in what.items()}
+    elif isinstance(what, list):
+        return [deep_escape(value) for value in what]
+    elif isinstance(what, str):
+        return html.escape(what)
+    else:
+        return what
```
rwols, line 202: I don't think this function is needed (see other comment).

plugin/completion.py
```diff
@@ -40,11 +41,100 @@
...
+            content += """<div style='margin: 10px'>
+                <p>{}</p>
+            </div>""".format(documentation)
+
+        return content
+
+    def normalized_documentation(self, item: Dict[str, Any]) -> str:
+        lsp_documentation = item.get("documentation")
+        if isinstance(lsp_documentation, str):
+            return mdpopups.md2html(self.view, lsp_documentation)
```
rwols, line 132: Should this not be `html.escape(lsp_documentation)`? Why do you asssume that it is markdown?

plugin/completion.py
```diff
@@ -40,11 +41,100 @@
...
+        elif isinstance(lsp_documentation, dict):
+            value = lsp_documentation.get("value", '')
+            return mdpopups.md2html(self.view, value)
```
rwols, line 135: You should check the `kind` here:

```
if lsp_documentation.get("kind") == "markdown":
    return mdpopups.md2html(self.view, value)
else:
    # must be plaintext
    return html.escape(value)
```

plugin/completion.py
```diff
@@ -40,11 +41,100 @@
...
+            </div>""".format(detail)
+
+        if documentation:
+            content += """<div style='margin: 10px'>
+                <p>{}</p>
+            </div>""".format(documentation)
+
+        return content
+
+    def normalized_documentation(self, item: Dict[str, Any]) -> str:
```
rwols, line 129: I think I used to call this function `normalized_documentation`, but I think the name `minihtml` suits better. It is a function that transform LSP-human-readable-text into ST-minihtml.

rwols: Perhaps move this function to plugin/core/views.py as we're going to need it in more places (see https://github.com/sublimelsp/LSP/issues/991).

plugin/completion.py
```diff
@@ -181,22 +282,23 @@ def format_completion(self, item: dict, can_resolve_completion_items: bool) -> s
...
+            st_annotation = ""
+
+        st_details = ""
+        if can_resolve_completion_items or "documentation" in item:
+            args = {
+                "item": deep_escape(item)
+            }
+            st_details += "<a href='subl:lsp_resolve_docs {}'>Δ</a>".format(sublime.encode_value(args))
+
+        st_details += "<p> {}</p>".format(lsp_detail)
```
rwols, line 301: If `st_details` is still an empty string, does the extra space get rendered or does it not get rendered due to the `<p>` tags? (I'm not good at html)

predragnikolic: I just put a space to space it from the link a bit. 
I should have probably put the space on a different place, for example after the closing a tag`"...</a> "`. :)

plugin/completion.py
predragnikolic: Aren't `if can_resolve_completion_items or "documentation" in item:` and `if can_resolve_completion_items or item.get("documentation"):` the same condition? :)

plugin/completion.py
predragnikolic: Hmm I agree, 
I should probably fill a but to the css language server. 

I did it this way, becasue the css ls return something like this
```
'documentation': {'value': "Specifies the width of the content area, padding area or border area (depending on 'box-sizing') of certain boxes.\n\nSyntax: <viewport-length>{1,2}\n\nMDN Reference: https://developer.mozilla.org/docs/Web/CSS/width", 'kind': 'plaintext'}
```
note that the `'kind'` is `'plaintext'`. 
But if I parse this `value` as markdown
the link gets clickable:
image

If i just escape it, it won't be clicable.

predragnikolic: I will change this to follow the spec

plugin/completion.py
predragnikolic: same reason, https://github.com/sublimelsp/LSP/pull/1022/files#r420379578

plugin/completion.py
rwols: It's plaintext, but we *could* still make links clickable :) I guess that's a next step after making sure it's plaintext.

rwols: Or we could ask the vscode-css guys to make their docs `markdown`. Then links will be automatically rendered as clickable.

plugin/completion.py
rchl: With the example you've linked:
```
'documentation': {'value': "Specifies the width of the content area, padding area or border area (depending on 'box-sizing') of certain boxes.\n\nSyntax: <viewport-length>{1,2}\n\nMDN Reference: https://developer.mozilla.org/docs/Web/CSS/width", 'kind': 'plaintext'}
```
The link will become clickable but also the `<viewport-length>` will disappear as it will be parsed as HTML element. And it's not supposed to be. It's supposed to be shown verbatim.

(CSS server is just stupid. It just use markdown formatting but escaping stuff that is supposed to be escaped.)

plugin/completion.py
rchl: With that example the link will become clickable but also the `<viewport-length>` will disappear as it will be parsed as HTML element. And it's not supposed to be. It's supposed to be shown verbatim.

(CSS server is just stupid. It should use markdown formatting but escape stuff that is supposed to be escaped.)

rchl: Hmm, now I've looked at your screenshot and it didn't disappear. Hmm. I remember recently someone reported a problem on Discord where an element like that from css server would mess up the text content.

rchl: I meant this bug https://github.com/sublimelsp/LSP/issues/991

plugin/completion.py
TheSecEng: This should likely be a conditional add, I hate to add "space" at the bottom of the popup if it is literally going to be blank.

```python
st_details += "<p> {}</p>".format(lsp_detail) if lsp_detail else ""
```

plugin/completion.py
rwols: It's not the same:

```
>>> f=lambda o: ('x' in o, bool(o.get('x')))
>>> f({})
(False, False)
>>> f({'x': None})
(True, False)
>>> f({'x': ''})
(True, False)
>>> f({'x': 'y'})
(True, True)
```

plugin/completion.py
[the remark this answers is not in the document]
rchl: Hmm, that book symbol might not be supported by many fonts.
📖 emoji works for me but probably won't work in ST on Windows.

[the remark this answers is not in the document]
rchl: We can use img tag with data URI.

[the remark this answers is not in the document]
rwols: Why not “More...”? It’s shorter than “Documentation”. I’m also honestly fine with the word “Documentation”.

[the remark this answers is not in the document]
rchl: If we would want to put more stuff in that field then either "Documentation" or "More" would be kinda confusing as it would blend with other text. Then the image/symbol makes no sense IMO.

But if we are not planning to put anything more in there then I'm fine with "More". It's more generic than "Documentation" and that is IMO good as the values shown after resolving item are not always strictly documentations.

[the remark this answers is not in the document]
predragnikolic: "up arrow" is actually the Delta Greek letter `Δ` (I changed it to `More`, so people can understand it easier on what it does) 

before:

Screenshot from 2020-05-07 22-08-32
(I prefer this visually)

after:

more
(I think this is more clearer)

plugin/completion.py
```diff
@@ -1,6 +1,7 @@
-import html
 import sublime
 import sublime_plugin
+import mdpopups
+import html
```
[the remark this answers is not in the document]
predragnikolic: I don't know why, but I have the tendency to put sublime imports at the top for no obvious reason :) 
I will sort them.

plugin/completion.py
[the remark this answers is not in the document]
rchl: I guess it's fine but it does kinda mix with the text (even though it has link style). Maybe have text in square brackets (`[More]`)?

plugin/completion.py
predragnikolic: Moved to CompletionHandler

plugin/completion.py
predragnikolic: Thanks for the explanation :)
I changed that to `.get('documentaion')`

plugin/completion.py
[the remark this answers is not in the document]
predragnikolic: this maybe (separated by a `|` pipe)?

image

plugin/completion.py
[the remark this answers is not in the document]
predragnikolic: done

plugin/completion.py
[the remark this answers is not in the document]
predragnikolic: here is how it looks with angle brackets:
image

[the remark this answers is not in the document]
rchl: Pipe WFM

predragnikolic: Should I wait for this sublime issue to be resolved or should I implement your suggested solution.
https://github.com/sublimehq/sublime_text/issues/3318

If 3318 gets fixed, that means that I don't have to deep escape the item.

Do you still want me to implement the index approach?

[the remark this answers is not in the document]
predragnikolic: I added the pipe,
@rwols feel free to unresolve the conversation if you think otherwise

plugin/completion.py
predragnikolic: Created a minihtml function that accepts a string or a MarkupContent as the first argument.

plugin/completion.py
predragnikolic: fixed with the minihtml function, https://github.com/sublimelsp/LSP/pull/1022/commits/8a803e0077b3eecaae56bf867155c3d9e7d96515#diff-f52730e9958cbd7da304f8aa67516d5cR179-R194

plugin/core/views.py
```diff
@@ -172,3 +174,22 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
         "options": formatting_options(view.settings()),
         "range": region_to_range(view, region).to_lsp()
     })
+
+
+def minihtml(content: Union[str, dict], view: sublime.View) -> str:
+    """ Content can be a string or MarkupContent """
+    if isinstance(content, str):
+        return html.escape(content)
```
rchl, line 182: In `pyls` I get this completion item for `json.JSONDecoreError`:
```
    {
      "kind": 7,
      "documentation": "JSONDecodeError(msg: str, doc: str, pos: int)\n\nSubclass of ValueError with the following additional properties:\n\nmsg: The unformatted error message\ndoc: The JSON document being parsed\npos: The start index of doc where parsing failed\nlineno: The line corresponding to pos\ncolno: The column corresponding to pos",
      "sortText": "aJSONDecodeError",
      "insertText": "JSONDecodeError",
      "detail": "json",
      "label": "JSONDecodeError"
    },
```
and the documentation is rendered without line breaks:
Screenshot 2020-05-08 at 10 28 13

Which makes it quite hard to parse.

Do we need to convert line breaks to HTML or something?
Should we do this instead:
```py
return mdpopups.md2html(view, html.escape(content))
```
 (and same for escape below)

plugin/core/views.py
```diff
@@ -172,3 +174,22 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
...
+
```
rchl, line 183: I understand that some use this style of have an empty line in if blocks but I find it weird :)

plugin/completion.py
```diff
@@ -41,10 +42,74 @@
...
+        self.is_detail_shown = bool(detail)
+
+        content = self.get_content(documentation, detail)
+        self.show_popup(content)
+
+        if not detail or not documentation:
+            # To make sure that the details or documentation fields don't exist
+            # we resove the completion item.
+            # If those fields appear after the item is resolved
+            # update the popup.
```
rchl, line 60: Style nit: No need to be so liberal with line breaks in comments. We have 120 column limit. :)
(Saving vertical space is good).
Also there are some typos in the text.

plugin/core/views.py
rwols: If it is plaintext, then:

1. Escape it with html.escape
2. The whitespace should be rendered verbatim. If we put the HTML-escaped plaintext string into ST's minihtml renderer it removes extraneous whitespace (like a browser does too).

We account for (1) with `html.escape`, but you're right that we don't account for (2).

Perhaps we can simply wrap the HTML-escaped plaintext in a `<code></code>` block?

rchl: Actually there is also another issue with using `html.escape` that needs to be investigated.
The `<http://json.org>` should be shown verbatim. The angle brackets turn into entities.
Screenshot 2020-05-08 at 10 45 24

plugin/completion.py
```diff
@@ -41,10 +42,74 @@
...
+    def get_content(self, documentation: str, detail: str) -> str:
+        content = ""
+        if detail and not self.is_detail_shown:
+            content += """<div class='highlight' style='margin: 10px'>
+                <p>{}</p>
+            </div>""".format(detail)
+
+        if documentation:
+            content += """<div style='margin: 10px'>
+                <p>{}</p>
```
rchl, line 109: I don't know if minihtml handles HTML per spec but having other block elements inside `p` tag is not allowed and will make the parser move the contents outside of it.

Whether minihtml follows that part of the spec or not, I'd rather follow it and not have `p` element since you don't know what elements rendered content will include. Let's change it to div (with a custom class if needed to style it).

plugin/core/views.py
rwols: Yes, so perhaps wrap it in `<code>` or `<tt>`. https://www.sublimetext.com/docs/3/minihtml.html

rwols: Meaning `"<code>{}</code>".format(html.escape(value))`

rchl: We can't `html.escape` as that will mess up the angle brackets from the example above.
But we need to experiment more and definitively add tests for that as it's way too easy to break.

rwols: But `&lt;` and `&gt;` are supposed to be rendered as `<` and `>` in a `<code>` tag?

```
<code>
&lgt;code&gt;
</code>
```
Is supposed to be rendered as
```
<code>
```

rchl: It doesn't seem to work but I don't have time to investigate more right now. Maybe I was testing wrong...

rwols: When I run this in the console:

```
>>> view.show_popup('<code>&lt;code&gt;</code>')
```
I get the correct popup. So I'm pretty sure this will give the correct output:

```
return '<code>{}</code>'.format(html.escape(content))
```

rwols: It would be nice if minihtml supported the `<pre>` tag, but oh well.

rchl: But line breaks are not preserved:
```
view.show_popup('<code>&lt;code&gt;\n\nEND</code>')
```

rwols: Oh, I didn't know that :) I guess we really do need a `<pre>` tag. Man, that is so annoying.

I guess you were right about using mdpopups.
```
Python 3.3>>> mdpopups.show_popup(view, '```\nfoo\nbar\n```', md=True)
```
It seems to take care of linebreaks and whitespace-preservation when you wrap the plaintext in three backticks:
```
return mdpopups.md2html(view, "```\n{}\n```".format(content))
```

rwols: Well, if the content happens to contain a line with three backticks, this could break. But I think that's a problem for later.

rwols: It's ironic that markdown is an easier problem to solve than plaintext...

rchl: It's not the problem that minihtml doesn't support `pre` tag but that it doesn't support `white-space: pre`. `code` tag should also have that by default per spec.

predragnikolic: https://forum.sublimetext.com/t/does-minihtml-support-the-pre-tag/46267/5
instead of the pre tag, we can replace the `\n` with `<br>` to **pre**serve the whitespace.

rwols: You would still need to account for extra space characters starting at a line (for indentation for example). Let’s not reinvent the wheel and just use mdpopups.

predragnikolic: I been thinking. We cannot apply md2html to markupKind "plaintext". 
Maybe the plaintext contain a `#` as the first char which would then accidentally get treated as h1  :)

Also I now encountered a bug, where sometimes the pyls return this:
{'insertText': 'Any', 'sortText': 'aAny', 'kind': 7, 'detail': 'plugin.completion', 'label': 'Any', 'documentation': 'Any()\n------------------------------\n'}

a

and sometimes it returns this:
{'detail': 'plugin.completion', 'documentation': 'Any()', 'kind': None, 'label': 'Any', 'sortText': 'aAny', 'insertText': 'Any'}

b

rchl: So we should also `strip()` the text?

rwols: My proposal above avoids all this.

rwols: I’m on a phone right now but it’s this:

```
return mdpopups.md2html(view, "```\n{}\n```".format(content))
```

Again no need to use HTML escape

plugin/completion.py
predragnikolic: I will go with the index approach, and see how it goes

predragnikolic: we can easily add the new api when sublime adds it

plugin/core/views.py
rwols: I agree. The liberal usage of empty lines is too much for me. The fact that PEP8 enforces empty lines between functions and classes provides enough visual separation.

If you believe your function is too hard to understand without empty lines, you should split up your function into smaller function. Enter PEP8 to enforce empty lines :)

plugin/core/views.py
predragnikolic: here is 
`html.escape(content).replace('\n', '<br>')`

image

`mdpopups.md2html(view, "```\n{}\n```".format(content))`

image

rwols: The last one is more faithful.

When there is whitespace like this:

```
some docs

    here is some indentation!
```

then that will be rendered correctly by your last example.

plugin/completion.py
predragnikolic: Thanks :) for each case I will use a div

plugin/completion.py
predragnikolic: done

plugin/completion.py
predragnikolic: deep_escepe is removed

plugin/core/views.py
predragnikolic: > BTW. ST closes completion list when it can't fit the "cooperate" popup...

Yes this happen to me now when I show docs for JSONDecodeError, but it doesn't happen always.

rchl:
> make rchlint happy :)
😆

plugin/core/views.py
```diff
@@ -172,3 +175,42 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
...
+            return mdpopups.md2html(view, value)
+        else:
+            # must be plaintext
+            return text2html(value)
+    else:
+        return ''
+
+
+def text2html(content: str) -> str:
+    content = html.escape(content).replace('\n', '<br>')
```
rwols, line 197: I did some experimentation. To get whitespace indentation all we have to do is this:

```
content = html.escape(content).replace('\n', '<br>').replace(' ', '&nbsp;')
```
minihtml seems to respect multiple `&nbsp;` escapes after each other, so this works.
Please add some tests for `text2html` in test_views.py because it has subtle behavior as we've learned.

rwols: This plaintext string:

```
first line\nsecond line\n    third line indented\n        more indent\nback on the first column again
```
is rendered as
Screenshot from 2020-05-09 12-05-05

plugin/core/views.py
```diff
@@ -172,3 +175,42 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
...
+
+    def repalce_url_with_link(matchObj: Any) -> str:
```
rwols, line 199: typo: repalce -> replace
style nitpick: matchObj -> match_obj (or just "match")

plugin/core/views.py
```diff
@@ -172,3 +175,42 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
...
+        href_tag, url = matchObj.groups()
+        if href_tag:
```
rwols, line 201: This can never happen right? HTML is already escaped, so we will never match `<a href="`. Please add a test for this in test_views.py if I'm wrong.

plugin/core/views.py
rwols: I forgot about tabs. We can do this for tabs:

```
html.escape(content).replace('\n', '<br>').replace(' ', '&nbsp;').replace('\t', '&nbsp;&nbsp;&nbsp;&nbsp;')
```

plugin/core/views.py
```diff
@@ -172,3 +175,42 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
...
+            # so return the whole match.
+            return matchObj.group(0)
+        else:
+            return "<a href='{}'>{}</a>".format(url, url)
+
+    pattern = re.compile(
+        r'((?:<a href[^>]+>)|(?:<a href="))?'
+        r'((?:https?):(?:(?://)|(?:\\\\))+'
+        r"(?:[\w\d:#@%/;$()~_?\+\-=\\\.&](?:#!)?)*)",
+        flags=re.IGNORECASE)
```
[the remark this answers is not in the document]
rwols: I'm also wondering whether it's even possible to match `<a href` because we `html.escape` the string.

predragnikolic:
Hey guys do you know how to fix this mypy error?
```
plugin/completion.py:86: error: Argument 2 to "send_request" of "Client" has incompatible type "Callable[[Dict[Any, Any]], None]"; expected "Callable[[Optional[Any]], None]"
```

```
86|            client.send_request(Request.resolveCompletionItem(item), self.handle_resolve_response)
87|
88|    def handle_resolve_response(self, item: dict) -> None:
```

rchl:
> Hey guys do you know how to fix this mypy error?
> 
> ```
> plugin/completion.py:86: error: Argument 2 to "send_request" of "Client" has incompatible type "Callable[[Dict[Any, Any]], None]"; expected "Callable[[Optional[Any]], None]"
> ```
> 
> ```
> 86|            client.send_request(Request.resolveCompletionItem(item), self.handle_resolve_response)
> 87|
> 88|    def handle_resolve_response(self, item: dict) -> None:
> ```

Make argument to `handle_resolve_response` `Optional`. And as a result of that, add error handling / type check of the result.

predragnikolic:
Thanks :)

plugin/core/views.py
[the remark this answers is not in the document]
predragnikolic: I must admit that I just found a thread on stack overflow on how to find a link and used it here.

But In our case there is no point if finding <a href="..."> because we already escaped characters .
Instead I will just look for any url starting with http till the first escaped " ' < > or \s (space) char.  
I will write tests for that. But If the regex turn out to bring more bad than good. I am for replacing the regex with something better. (or removing the link parsing entirely, till we get something more battle profed)

plugin/core/views.py
predragnikolic: I am now wondering if we should preserve space for plaintext content. 
Given we have content that is like this:
```
Specifies the width of the content area, padding area or border area (depending on 'box-sizing') of certain boxes.

Syntax: <viewport-length>{1,2}

MDN Reference: https://developer.mozilla.org/docs/Web/CSS/width
```

If we use `.replace(' ', '&nbsp;').replace('\t', '&nbsp;&nbsp;&nbsp;&nbsp;')`) we will see exactly the same formatting as in the code block above:

aa
But a horizontal scrollbar is shown.

Without it we will see:
ab
The text is wrapped, but a vertical scrollbar is shown.

I would rather see a vertical scrollbar than a horizontal one.
What do you guys think?

rchl: I don't like horizontal scrolling either.
So what was the problem with using triple backticks? Is it that it changes font look?

rchl: Since we are desperate and coming up with crazy workarounds, maybe we can try only **re**placing multiple **consecutive** spaces with `&nbsp;`. so:
```py
re.sub(r' {2,}', '&nbsp', str)
```

predragnikolic: @rchl correct. The link wont be clickable.

rwols: We can `split()` on newlines, then find the first non-whitespace char, replace the whitespace prefix with `&nbsp;` ?

And then `"<br>".join(split_lines)` again? :P

plugin/core/views.py
predragnikolic: I removed the regex for finding `<a href` because we already escaped the string. Instead I  think we should only find URLs and make them clickable.

plugin/core/views.py
rwols: I suppose the end-all-be-all answer can be deferred for a later PR...

predragnikolic: I can try something with `re.sub(r' {2,}', replace_npbs, str)`

```
def replace_npbs(match):
   return "&nbsp;" * len(match)
```

rchl: Replacing only leading spaces would be even easier and likely more efficient with regex. But IMO it's important to also preserve multiple spaces anywhere.
This is maybe crazy but this should work:
```py
re.sub(r' {2,}', lambda match: ''.join(['&nbsp;' for space in match.group(0)]), str)
```

plugin/core/views.py
```diff
@@ -196,21 +196,13 @@ def minihtml(content: Union[str, dict], view: sublime.View) -> str:
...
+        url = match.group(0)
+        return "<a href='{}'>{}</a>".format(url, url)
 
-    pattern = re.compile(
-        r'((?:<a href[^>]+>)|(?:<a href="))?'
-        r'((?:https?):(?:(?://)|(?:\\\\))+'
-        r"(?:[\w\d:#@%/;$()~_?\+\-=\\\.&](?:#!)?)*)",
+    FIND_URL = re.compile(
+        # match from `http` till the first escaped ", ', <, > or space character
+        r'(https?:(?://|\\\\).+?(?=(&quot;|&#x27;|&lt;|&gt;|\s)))',
```
rchl: `\\\\` dosn't seems right to me. IMO we could just remove it.

plugin/completion.py
```diff
@@ -41,10 +42,76 @@
...
+        content = ""
+        if detail and not self.is_detail_shown:
+            content += """<div class='highlight' style='margin: 4px'>
+                <div>{}</div>
+            </div>""".format(detail)
+
+        if documentation:
+            content += """<div style='margin: 10px'>
+                <div>{}</div>
+            </div>""".format(documentation)
```
rchl, line 71: We don't need extra child `<div>`. And we don't need whitespace in those strings (both can be on one line).

plugin/core/views.py
rwols: Don’t forget tab characters!

plugin/core/views.py
predragnikolic: that line was from an old commit.
I reverted it back to:
`r'(https?:(?://|\\\\)(?:[\w\d:#@%/;$()~_?\+\-=\\\.&](?:#!)?)*)',`

I just did `(?://|\\\\)` because I saw it in the stack overflow answer. What do you say about just `https?://(?:[\w\d:#@%/;$()~_?\+\-=\\\.&](?:#!)?)*)`?

rchl: I don't know parsing URLs by heart so it's good to base it on some popular and accepted answer. I can't validate that it's 100% correct (if it's even possible to do it 100% correct with RE...).

I only have issues with the `\\\\` part so your suggestion with it removed seems fine.

predragnikolic: bd1bb4587e8855525ba6470ac133fea9a4b334c4

plugin/completion.py
predragnikolic: ca0fbf72ef62b1fbe312e918d099832fa7f284d0

plugin/core/views.py
predragnikolic: 22d047ebaba5eff3dccade2e32130861d76e5727 and 22d047ebaba5eff3dccade2e32130861d76e5727 :)

predragnikolic: Tell me if you would like to change anything else. I
I am heading to write some test for plaintext2html.

plugin/core/views.py
predragnikolic: > if it's even possible to do it 100% correct with RE... 

it is probably possible, but I don't think that this regex is the one. :) 
but lets hope that this regex will suit our needs

plugin/completion.py
```diff
@@ -41,10 +42,72 @@
...
+                lambda res: self.handle_resolve_response(res))
+
+    def handle_resolve_response(self, item: Optional[dict]) -> None:
+        if not item:
+            return
+        detail = minihtml(item.get('detail') or "", self.view)
+        documentation = minihtml(item.get("documentation") or "", self.view)
+
+        content = self.get_content(documentation, detail)
+        if self.view.is_popup_visible():
```
rwols, line 104: This will most likely interfere with the signature help state machine if we decide to show sigHelp when the AC is open, but I suppose that is a story for another time.

plugin/completion.py
```diff
@@ -41,10 +42,72 @@
...
+        if has_resolve_provider:
+            client.send_request(
+                Request.resolveCompletionItem(item),
+                lambda res: self.handle_resolve_response(res))
+
+    def handle_resolve_response(self, item: Optional[dict]) -> None:
+        if not item:
+            return
+        detail = minihtml(item.get('detail') or "", self.view)
+        documentation = minihtml(item.get("documentation") or "", self.view)
```
rwols, line 101: I like how this turned out. It's clear what `minihtml` does.

plugin/completion.py
```diff
@@ -41,10 +42,72 @@
...
+            self.view,
+            content,
+            flags=sublime.COOPERATE_WITH_AUTO_COMPLETE,
+            max_width=480,
+            max_height=300,
+            allow_code_wrap=True
+        )
+
+    def do_resolve(self, item: dict) -> None:
+        session = session_for_view(self.view, 'completionProvider', self.view.sel()[0].begin())
```
rchl, line 82: Since #1030 was integrated once you rebase you can directly check capability `'completionProvider.resolveProvider'` and skip capability checks below.

plugin/core/views.py
```diff
@@ -172,3 +175,39 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
...
+            return mdpopups.md2html(view, value)
+        else:
+            # must be plaintext
+            return text2html(value)
+    else:
+        return ''
+
+
+def text2html(content: str) -> str:
+    def replace_npbs(match: Any) -> str:
```
rwols, line 197: Let's talk about visual whitespace separation.

In general you're adding *way too much* empty lines between trivial statements. This makes it hard to see where a function starts and ends.

And yet in this case where we define a named lambda, there is *not enough* whitespace separation. It's important to see where a (new) function/lambda/method starts/ends. It allows a reader to get a quick overview of the code.

Generally please do it like this:

```
def some_function():
    no
    new
    empty_lines
    for
    statements

    def some_named_lambda():
        no
        new
        empty_lines
        for_simple_statements

    continuing_on_in_this_function
    foobar_baz


def next_function():
    ...
```
I can easily see where functions/lambdas/methods start/end.

plugin/completion.py
```diff
@@ -147,6 +212,19 @@ def initialize(self) -> None:
             # This is to make ST match with labels that have a weird prefix like a space character.
             settings.set("auto_complete_preserve_order", "none")
 
+    def on_post_text_command(self, command: str, args: dict) -> None:
+        if not self.view.is_popup_visible():
+            return
+
```
rwols, line 218: I got confused thinking that the next line would be a named lambda / closure.

plugin/core/views.py
```diff
@@ -172,3 +175,39 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
...
+        if kind == "markdown":
+            return mdpopups.md2html(view, value)
+        else:
+            # must be plaintext
+            return text2html(value)
+    else:
+        return ''
+
+
+def text2html(content: str) -> str:
```
rwols, line 196: I want to see some tests for this function in test_views.py :)

plugin/core/views.py
```diff
@@ -172,3 +175,39 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
...
+        spaces = match.group(0)
+        return "&nbsp;" * len(spaces)
+
+    content = html.escape(content).replace('\n', '<br>').replace('\t', '&nbsp;&nbsp;&nbsp;&nbsp;')
+    # if there are 2 or more spaces, replace them with &nbsp;
+    content = re.sub(r"( {2,})", replace_npbs, content)
+
+    FIND_URL = re.compile(
+        r'(https?://(?:[\w\d:#@%/;$()~_?\+\-=\\\.&](?:#!)?)*)',
+        flags=re.IGNORECASE)
```
rwols, line 208: Please fit this on one line.

predragnikolic:
> This will probably be my last "Request changes" :) we're almost there :)

have you seen c6cc955? :)

plugin/core/views.py
predragnikolic: See if 1254173 is better.

plugin/completion.py
predragnikolic: 16741d06ce4b4b6c1da0073e743543882719bf94 is this how you imagined it?

rchl: I'm not gonna say this is wrong because I don't want to open another big exchange here but sending requests to all servers seems wrong. Sending it just to the first one also seems wrong. Ideally, we would send it to the one that provided the suggestion.

Maybe let's stick to sending the request just to the first server that supports capability?
After you rebase, the capability argument will also exist for `session_for_view` (singular).

predragnikolic: This is probably something we need to discuss in a separate issue and in more detail.

plugin/core/views.py
```diff
@@ -172,3 +176,39 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
...
+    else:
+        return ''
+
+
+def text2html(content: str) -> str:
+    content = "\n".join(textwrap.wrap(content, width=54,
+                                      replace_whitespace=False, break_long_words=False, break_on_hyphens=False))
+    content = html.escape(content).replace('\n', '<br>').replace('\t', '&nbsp;&nbsp;&nbsp;&nbsp;')
+
+    def replace_npbs(match: Any) -> str:
```
TheSecEng, line 202: Doesn't md2html already have this logic builtin? What am I missing?

Also should the method name be `replace_nbsp`?

plugin/core/views.py
```diff
@@ -172,3 +176,39 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
...
+        return "&nbsp;" * len(spaces)
+
+    # if there are 2 or more spaces, replace them with &nbsp;
+    content = re.sub(r"( {2,})", replace_npbs, content)
+
+    def replace_url_with_link(match: Any) -> str:
+        url = match.group(0)
+        return "<a href='{}'>{}</a>".format(url, url)
+
+    FIND_URL = re.compile(r'(https?://(?:[\w\d:#@%/;$()~_?\+\-=\\\.&](?:#!)?)*)', flags=re.IGNORECASE)
```
TheSecEng, line 213: If valid MD is being passed, shouldn't md2html cover this?

rwols: There are two content formats. One is markdown, the other is plaintext. When the server returns markdown, it is indeed handled by `mdpopups.md2html`. When the server returns plaintext, we have to put a little more thought into faithfully preserving the original plaintext when giving it to ST's minihtml.

plugin/core/views.py
TheSecEng: Same as below I presume

plugin/completion.py
```diff
@@ -41,10 +42,62 @@
...
+        if detail and not self.is_detail_shown:
+            content += "<div class='highlight' style='margin: 4px'>{}</div>".format(detail)
+
+        if documentation:
+            content += "<div style='margin: 10px'>{}</div>".format(documentation)
+
+        return content
+
+    def show_popup(self, content: str) -> None:
+        mdpopups.show_popup(
```
rwols, line 72: I don't think this is correct? Is it always interpreted as markdown now? Or is `md=True` needed to make that happen?

What happens when we pass minihtml in here?

plugin/completion.py
```diff
@@ -147,6 +202,18 @@ def initialize(self) -> None:
...
+        if not self.view.is_popup_visible():
+            return
+        if command in ["hide_auto_complete", "move"]:
+            # hide the popup when `esc` or arrows are pressed pressed
+            self.view.hide_popup()
+
+    def on_modified_async(self) -> None:
+        if self.view.is_popup_visible():
+            # hide the docs popup on each keystoke
+            self.view.hide_popup()
```
rwols, line 215: Does this logic interfere with the signature help logic? It also has an `on_modified_async` callback.

plugin/core/views.py
rwols: Much better :)

plugin/completion.py
predragnikolic: Yes, it interferes. 
I did it because when the AC is gone while for example deleting text, the cooperate popup is (sometimes) still visible. I should find a different way to handle this.

plugin/completion.py
predragnikolic: I don't think we need to set `md=True`. because we already converted the content if from markdown to html(if it was markdown) in minihtml with md2html. 

I will rename `content` to minihtml.

plugin/completion.py
predragnikolic: 73988f365dac94ba0fa25ccafea948c5d0dcaa3c

plugin/core/views.py
predragnikolic: Darn it :) https://github.com/sublimelsp/LSP/pull/1022/commits/6058b346788c2108466908592182952421cb9c32 
Thanks

plugin/completion.py
predragnikolic: @rchl I had to revert back to the old way of checking the capability.
I left a brief description in the commit message describing the problem.

Can you take a look?
6d7103e999c09defc4862416ad005bde469d72ea

plugin/core/views.py
predragnikolic: I will try to follow this style guide as much as I can

plugin/core/views.py
predragnikolic: 99983c36bdf7cbed005017f68b9ae57082c38928 :)
not a lot, but tests are tests.
Will add more test when people start reporting problems :)

predragnikolic:
I noticed that the links although wrapped in a link tag. 
and have a href, 
and are highlighted as links
When they are clicked, they wont open a browser... at least not for me.
fun. 

For me this peace of code doesn't open a browser :)
`view.show_popup('<a href="https://github.com/sublimelsp/LSP/pull/1022">My pull request</a>')`

rchl:
> Revert back logic for session_for_view.
> 
> @rchl can you take a look at this.

That's because you didn't have required changes from `st4000-exploration` branch at that point. You've only merged with it in later commit.

TheSecEng:
Would it be possible to provide a `shift` keybinding to detect the shift keypress and trigger showing documentation? I might be a nice solution. As it stands, moving the mouse down to click still isn't a perfect UX

rwols:
If you have F12 mapped to "lsp_symbol_definition", you can change the context:

```
	{"keys": ["f12"], "command": "lsp_symbol_definition", "context": [
		{"key": "setting.lsp_active"},
		{"key": "auto_complete_visible", "operator": "equal", "operand": false}
	]},
```

Then press F12 to open the docs/help in the AC widget.

This works because there's a new keybinding for F12 in the default keybindings:

```
	{ "keys": ["f12"], "command": "auto_complete_open_link", "context":
		[
			{ "key": "auto_complete_visible", "operator": "equal", "operand": true },
		]
	},
```

You can, of course, also just map `auto_complete_open_link` to a keybinding...

TheSecEng:
> If you have F12 mapped to "lsp_symbol_definition", you can change the context:
> 
> ```
> 	{"keys": ["f12"], "command": "lsp_symbol_definition", "context": [
> 		{"key": "setting.lsp_active"},
> 		{"key": "auto_complete_visible", "operator": "equal", "operand": false}
> 	]},
> ```
> 
> Then press F12 to open the docs/help in the AC widget.
> 
> This works because there's a new keybinding for F12 in the default keybindings:
> 
> ```
> 	{ "keys": ["f12"], "command": "auto_complete_open_link", "context":
> 		[
> 			{ "key": "auto_complete_visible", "operator": "equal", "operand": true },
> 		]
> 	},
> ```
> 
> You can, of course, also just map `auto_complete_open_link` to a keybinding...

I see, likely to change this to shift or something on my preferences. As on MacBooks with a Touch Bar. Function keys are hidden by default and this becomes a pain.

Thanks

predragnikolic:
I just want to add some documentation for the keybinding in the featues.md

plugin/core/views.py
```diff
@@ -1,9 +1,15 @@
...
+import mdpopups
+import re
+import sublime
+import textwrap
 from .protocol import Point, Range, Notification, Request
-from .typing import Optional, Dict, Any, Iterable, List
+from .typing import Optional, Dict, Any, Iterable, List, Union
 from .url import filename_to_uri
 
+DOCS_WRAP_WIDTH = 58
```
rchl, line 11: This seems rather aggressive...

Screenshot 2020-05-12 at 20 23 34

predragnikolic:
Added.

you can have a look 
https://github.com/sublimelsp/LSP/blob/a6cbbecea01e700e8a172c14bdcc4c7f0588abff/docs/features.md#show-autocomplete-documentation

plugin/core/views.py
predragnikolic: Should it be configurable?
For example,
have a setting `"documentation_popup_width"` 
and based on the given width, I somehow calculate the `DOCS_WRAP_WIDTH` ?

predragnikolic: That is the first thing that came to my mind.
Do you have any suggestions.

And yeah, I have forgotten about the JSONDecodeError which closes the AC for some reason...

rchl: Max width of popup is set to 480px and wrapping makes the content 300px. So there is still some room for having larger wrap threshold.

Of course we can't reliably determine what will fit in 480px but at least we should get closer to it.

rchl: I don't know how accurate it will be but we can use `view.em_width()` to get character width.

predragnikolic:
```
DOCS_POPUP_WIDTH = 480 # If have a width

# and if I know the width of the a single char, which I know with view.em_width()

# I can calculate the wrap_width.  (but I also should probably account for any margin left | right, but I need to test that) 

# wrap_width = MAX_WIDTH / char_width
#or maybe
# wrap_width = (MAX_WIDTH - margin-left - margin-right) / char_width

def text2html(view: sublime.View, content: str) -> str:
    content = html.escape(content).replace('\n', '<br>').replace('\t', '&nbsp;&nbsp;&nbsp;&nbsp;')
    x_margins = 12
    wrap_width = int((DOCS_POPUP_WIDTH - x_margins) / view.em_width())
    content = "<br>".join(textwrap.wrap(content, width=wrap_width,
                          replace_whitespace=False, break_long_words=False, break_on_hyphens=False))
```

rchl: Don't try to be pixel perfect as em_width won't give you perfectly accurate width. It's typically a wide of the widest letter so it depends on font style and what not. You probably have to test and eyeball it.

predragnikolic: I push the commit for calculating the wrap width.
Right now I am using a constant `DOCS_POPUP_WIDTH` which is hard-coded to `480`.

We can add a settings later if we notice that people want to customize the width of the popup

predragnikolic:
Running `tox .` locally gives me ` congratulations :)`, 
don't know why `Run sudo apt update` is failing int `main / Lint`

plugin/core/views.py
```diff
@@ -172,3 +178,42 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
...
+        return "&nbsp;" * len(spaces)
+
+    # if there are 2 or more spaces, replace them with &nbsp;
+    content = re.sub(r"( {2,})", replace_nbsp, content)
+
+    def replace_url_with_link(match: Any) -> str:
+        url = match.group(0)
+        return "<a href='{}'>{}</a>".format(url, url)
+
+    FIND_URL = re.compile(r'(https?://(?:[\w\d:#@%/;$()~_?\+\-=\\\.&](?:#!)?)*)', flags=re.IGNORECASE)
```
rchl, line 218: Sorry, one more thing:

The link is not "linkified" properly and clicking it opens blank tab:
Screenshot 2020-05-12 at 21 54 51

plugin/core/views.py
```diff
@@ -172,3 +178,42 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
...
+    else:
+        return ''
+
+
+def text2html(view: sublime.View, content: str, popup_width: Optional[int] = None) -> str:
+    content = html.escape(content).replace('\n', '<br>').replace('\t', '&nbsp;&nbsp;&nbsp;&nbsp;')
+    if popup_width is not None:
+        # if popup width is provided, calucalte how many caracters can fit on one line
+        wrap_width = int(popup_width / view.em_width())
+        content = "<br>".join(textwrap.wrap(content, width=wrap_width,
```
rwols, line 204: The newlines and tabs are already replaced in `content`. This will most likely influence what `textwrap.wrap` is doing?
Shouldn't we apply `textwrap.wrap` at the very beginning of this function instead?

plugin/core/views.py
rchl: Before and after parsing link:
```
Extensible JSON &lt;http://json.org&gt; encoder
=>
Extensible JSON &lt;<a href='http://json.org&gt;'>http://json.org&gt;</a> encoder
```
So the escaping is what causes the issue.
Can we replace links before escaping?

plugin/core/views.py
rwols: But if we apply this at the beginning of the function, before `html.escape`, then it might screw up `html.escape`.
I have my doubts about whether this `textwrap.wrap` function always does the right thing...

plugin/core/views.py
rchl: No, we can't because we'll escape html tags added by us...

predragnikolic: I can probably detect links before encoding 
```
detected_links = []

# before encoding

text = "Extensible JSON <http://json.org>; encoder"
#                ^              ^ link detected

detected_links = ['http://json.org']

# after encoding

text = "Extensible JSON &lt;http://json.org&gt; encoder"

for url in detected_links:
  text.replace(url, "<a href='{}'>{}</a>".format(url, url))

File: plugin/core/views.py (hunk 33 above)
Comment (predragnikolic) [reply plugin/core/views.py:204]: I can apply it at the beginning before the encoding. 
But **then** instead of `"<br>".join(textwrap.wrap...` I should use **"\n"**`.join(textwrap.wrap...` 
than in the next line the \n will be replaced by a <br> 
```
content = "\n".join(textwrap.wrap...
content = html.escape(content).replace('\n', '<br>')...
```

I inserted <br> tags after the encoding, because I don't wont to insert new lines, just so I could replace them with <br> in the next line :)

> This will most likely influence what textwrap.wrap is doing?

Can you be more specific on what you mean by *influence* what the text wrap is doing?
influence in what way? Are concerned about whitespace? What is your concern? :)

File: plugin/core/views.py (hunk 32 above)
Comment (rchl) [reply plugin/core/views.py:218]: Also the "table" is pretty misaligned in that same popup:
Screenshot 2020-05-12 at 22 34 59

File: plugin/core/views.py (hunk 33 above)
Comment (rwols) [reply plugin/core/views.py:204]: Well, if we replace one space character with six &nbsp; characters, the line has five more characters from the POV of textwrap.wrap

File: plugin/core/views.py [hunk 34]
```diff
@@ -172,3 +178,42 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
...
+        if kind == "markdown":
+            return mdpopups.md2html(view, value)
+        else:
+            # must be plaintext
+            return text2html(view, value, popup_width)
+    else:
+        return ''
+
+
+def text2html(view: sublime.View, content: str, popup_width: Optional[int] = None) -> str:
```
rchl, line 199: Here is my suggestion:
```py

RE = re.compile(r'(?P<entity>[&<>])|(?P<url>https?://(?:[\w\d:#@%/;$()~_?\+\-=\\\.&](?:#!)?)*)', flags=re.IGNORECASE)
ENTITY_MAP = {
    "&": "&amp;",
    "<": "&lt;",
    ">": "&gt;",
}


def text2html(view: sublime.View, content: str, popup_width: Optional[int] = None) -> str:
    # if popup_width is not None:
    #     # if popup width is provided, calucalte how many caracters can fit on one line
    #     wrap_width = int(popup_width / view.em_width())
    #     content = "<br>".join(textwrap.wrap(content, width=wrap_width,
    #                           replace_whitespace=False, break_long_words=False, break_on_hyphens=False))

    def replace_match(match: Any) -> str:
        entity_match = match.group('entity')
        if entity_match:
            return ENTITY_MAP[entity_match]
        url = match.group('url')
        return "<a href='{}'>{}</a>".format(url, url)

    content = re.sub(RE, replace_match, content).replace('\n', '<br>').replace('\t', '&nbsp;&nbsp;&nbsp;&nbsp;')

    # if there are 2 or more spaces, replace them with &nbsp;
    return re.sub(r"( {2,}|[\xc2\xa0])", lambda match: "&nbsp;" * len(match.group(0)), content)
```

- fixes link wrapping
- also replaces non-breaking spaces which helps with alignment in those "tables" I've mentioned before
- disabled text wrapping as I think it will do more harm than good and also it breaks formatting if we add '<be>s' before replacing.

plugin/core/views.py
rwols: And because \n is replaced by <br>, the function does not encounter new lines.

plugin/core/views.py
predragnikolic: The "table" is the reason on why I think I should remove the text wrap.
Although I don't like vertical scroll (and textwrap seemed to fix it), I would rather see a scroll than a broken "table" .

plugin/core/views.py
rchl: It could be massaged a bit more to have only one replace function instead of two separate ones.

predragnikolic: @rchl can you test what will happen if the content was
```
# I added " quotes which will be escaped to &qout; or something like that
content = 'Extensible JSON "http://json.org" encoder'
```
I think that the result will be `...://json.org&qout;</a>` or maybe not. 
But I have a feeling that it will

rchl: I'm not escaping quotes in my code.
The `html.escape()` does it by default (but you can disable it with an argument). I don't think we need to escape quotes but it's easy to do that if we want.

rchl: Screenshot 2020-05-12 at 22 54 38

predragnikolic: What if a url actually contains `&gt;`, `http://json.org&gt;`. Did we broke the link than?

TheSecEng:
Seeing as we know the language, could we not add syntax highlighting to code inside the documentation ?

TheSecEng:
Also, in instances where documentation is defined like such, we get unneeded leading whitespace.

image

image

plugin/core/views.py
rchl: That's an invalid URL in the first place. Where does that example come from?

We can't fix incorrect code ourselves. Or we shouldn't need to at least. That would go a bit too far.

rchl:
> Also, in instances where documentation is defined like such, we get unneeded leading whitespace.
> 
> image
> 
> image

It would be up to the server to trim that leading whitespace. We can't do that as then we would break other cases.

But I can't reproduce this with pyls:
Screenshot 2020-05-13 at 07 37 38
Screenshot 2020-05-13 at 07 38 38

predragnikolic:
@TheSecEng you probably checked out the commit, where I tried replacing all space with &nbsp;

PR #1023, opened 2020-05-05
LSP server restart specific

### LINKED ISSUE #1021: Restart specific LSP servers vs Restart All LSP servers
### What
**OS and language server:** All

LSP allows for a command `LSP: Restart Servers` which reloads all LSP servers currently in use via "sessions". Useful for debugging, however there is no need to force started all LSP servers, if you only need to restart a single server that is being buggy.

### Solution
 There should be a command that drops into a ListPanel that shows all running LSP servers and allows the user to select the specific server to restart.

TheSecEng:
**What**

Implements restart specific LSP servers.

Based off 4000-exploration, if that is an issue I can rebase off of master

fixes #1021


plugin/core/registry.py
```diff
@@ -144,11 +144,33 @@ def client_with_capability(self, capability: str) -> Optional[Client]:
         return client_from_session(session_for_view(self.view, capability))
 
 
-class LspRestartClientCommand(sublime_plugin.TextCommand):
+class LspRestartServersCommand(sublime_plugin.TextCommand):
```
jfcherng, line 147: Since this is gonna to be a BC anyway. I would like to propose naming it as `LspRestartAllServersCommand`.

TheSecEng: I agree, I think that would make more sense

plugin/core/windows.py
```diff
@@ -485,10 +488,21 @@ def _handle_message_request(self, params: dict, source: str, client: Client, req
...
 
+    def restart_session(self, config_name: str) -> None:
+        self._restarting = True
+        self.end_session(config_name=config_name)
+
     def restart_sessions(self) -> None:
         self._restarting = True
         self.end_sessions()
 
+    def end_session(self, config_name: str) -> None:
```
rwols, line 499: Why is `end_config_sessions` not sufficient?

TheSecEng: And overlook on my part. Will update

plugin/core/registry.py
```diff
@@ -144,11 +143,33 @@ def client_with_capability(self, capability: str) -> Optional[Client]:
...
             windows.lookup(window).restart_sessions()
+
+
+class LspRestartServerCommand(sublime_plugin.TextCommand):
+    def is_enabled(self) -> bool:
+        window = self.view.window()
+        lsp_servers = bool()
+        if window:
+            lsp_servers = True if windows.lookup(window).get_sessions() else False
+        return is_supported_view(self.view) and lsp_servers
```
TheSecEng, line 162: Was also thinking of creating a function in the WindowManager as such

```python
def has_active_sessions(self) -> bool:
    return True if list(self._sessions) else False
```

and updating this code to look like

```python
class LspRestartServerCommand(sublime_plugin.TextCommand):
    def is_enabled(self) -> bool:
        window = self.view.window()
        if not window:
            return False
        return is_supported_view(self.view) and windows.lookup(window).has_active_sessions()
```

TheSecEng: Thoughts

plugin/core/windows.py
```diff
@@ -4,7 +4,7 @@
 from .message_request_handler import MessageRequestHandler
 from .protocol import Notification, Response, TextDocumentSyncKindNone, TextDocumentSyncKindFull
 from .rpc import Client
-from .rpc import Client, SublimeLogger
+from .rpc import SublimeLogger
```
TheSecEng, line 7: There was a redundant import here for some reason. Just removing

plugin/core/registry.py
```diff
@@ -144,11 +143,32 @@ def client_with_capability(self, capability: str) -> Optional[Client]:
...
         if window:
             windows.lookup(window).restart_sessions()
+
+
+class LspRestartServerCommand(sublime_plugin.TextCommand):
+    def is_enabled(self) -> bool:
+        window = self.view.window()
+        if not window:
+            return False
+        return is_supported_view(self.view) and windows.lookup(window).has_active_sessions()
```
rchl, line 161: shouldn't need window check here as `is_supported_view` checks it

plugin/core/registry.py
```diff
@@ -144,11 +143,32 @@ def client_with_capability(self, capability: str) -> Optional[Client]:
...
+
+    def run(self, edit: Any) -> None:
+        self.window = self.view.window()
+        if not self.window:
+            return
+
+        self.lsp_sessions = windows.lookup(self.window).get_sessions()
+        if self.lsp_sessions:
+            self.window.show_quick_panel(self.lsp_sessions, self.restart_server)
```
rchl, line 170: Not a good practice to store stuff on the instance as it's per-view, it's not per command execution. You can create lambda:

```suggestion
        sessions = windows.lookup(self.window).get_sessions()
        if sessions:
            self.window.show_quick_panel(self.lsp_sessions, lambda index: self.restart_server(sessions[index]))
```
And update `restart_server` accordingly.

Although you'll get -1 for index when canceling quick panel so you will have to do a bit of error checking. Maybe pass `index` and `sessions` to `restart_servers`.

plugin/core/windows.py
```diff
@@ -351,6 +351,12 @@ def _on_project_switched(self, folders: List[str]) -> None:
     def get_session(self, config_name: str, file_path: str) -> Optional[Session]:
         return self._find_session(config_name, file_path)
 
+    def get_sessions(self):
+        return list(self._sessions)
+
+    def has_active_sessions(self) -> bool:
+        return True if self._sessions else False
```
rchl, line 358:
```suggestion
        return bool(self._sessions)
```

plugin/core/windows.py
```diff
@@ -485,6 +491,11 @@ def _handle_message_request(self, params: dict, source: str, client: Client, req
         handler = MessageRequestHandler(self._window.active_view(), client, request_id, params, source)  # type: ignore
         handler.show()
 
+    def restart_session(self, config_name: str) -> None:
+        self._restarting = True
+        self.documents.reset()
```
rchl, line 496: Does that guarantee that only the chosen server and not all servers for the given window are restarted?
(not sure what `reset()` does)

plugin/core/registry.py
```diff
@@ -144,11 +143,32 @@ def client_with_capability(self, capability: str) -> Optional[Client]:
...
+
+    def restart_server(self, index) -> None:
+        windows.lookup(self.window).restart_session(config_name=self.lsp_sessions[index])
```
rchl, line 173: Need to handle `-1` (cancel) case or this will crash.

plugin/core/windows.py
TheSecEng: I was actually looking at that at well,

https://github.com/sublimelsp/LSP/blob/8800b812a4b23a686059d2d9f11eeb8b5f5fb776/plugin/core/windows.py#L127-L130

https://github.com/sublimelsp/LSP/blob/8800b812a4b23a686059d2d9f11eeb8b5f5fb776/plugin/core/windows.py#L172-L174

`self._documents_states` is just a `set()`

rchl: So it will reset the `show_definitions` setting to default value and remove active server names from the status. That seems a bit wrong if we are not restarting all of the servers. Remember there can be two or more servers running for specific file.

TheSecEng: Ahh you are right. Will need to brainstorm on the right approach for this

plugin/core/registry.py
```diff
@@ -156,21 +156,19 @@ def run(self, edit: Any) -> None:
...
-        if window:
-            self.lsp_server_window = windows.lookup(window)
-            self.lsp_servers = self.lsp_server_window.get_sessions()
-
-            if self.lsp_servers:
-                window.show_quick_panel(self.lsp_servers, self.restart_server)
-
-    def restart_server(self, index) -> None:
-        self.lsp_server_window.restart_session(config_name=self.lsp_servers[index])
+        self.window = self.view.window()
```
rchl, line 161: There is really no need to store anything on the command instance. You can re-retrieve it easily.
And while it's unlikely that the window can change between invocation of the command and `restart_server`, it's still conceptually wrong to store state like that here.

plugin/core/windows.py
TheSecEng: >And re-attaching only happens on opening the file so the file would have to be re-opened for the status to show back.

Well, I am not sure that a re-open is required. `LSP: Restart Servers` does this same logic and the status shows back without a re-open of the file. 

I definitely think we can create a helper function that does something along the lines of

```python
def reset(self, config_name: str) -> None:
    for view in self._window.views():
        if view.settings().get('lsp_language').get(config_name):
            self.detach_view(view)
    self._document_states.clear()
```

So that we only detach the view if it is affected by that language server that is being restarted. Should we still use `self._document_states.clear()` in this case?

TheSecEng: `self._document_states.clear()` is a set that ends up looking like 

`{'/projects/LSP/plugin/core/windows.py', '/projects/LSP/plugin/core/registry.py'}`

What about changing how this is stored and making it similar to this, then we could clear document_states based on the language server.

```json
{
  "pyls": ["/projects/LSP/plugin/core/windows.py", "/projects/LSP/plugin/core/registry.py"],
  "gopls": ["/projects/LiveSessions/main.go", "/projects/LiveSessions/support.go"]
}
```

We could then call `self._document_states.clear()` or `self._document_states[config_name].clear()`

plugin/core/registry.py
```diff
@@ -144,11 +143,31 @@ def client_with_capability(self, capability: str) -> Optional[Client]:
...
+        return is_supported_view(self.view) and windows.lookup(self.view.window()).has_active_sessions()
+
+    def run(self, edit: Any) -> None:
+        window = self.view.window()
+        if not window:
+            return
+
+        sessions = windows.lookup(window).get_sessions()
+        if sessions:
+            window.show_quick_panel(sessions, lambda index: self.restart_server(index, sessions))
```
jwortmann, line 167: Maybe skip the quick panel if there is only one server running?

plugin/core/windows.py
rwols: It's a step in the right direction, but I already have plans for this involving the `Session` object. Please keep the changes to a single topic for a pull request :)

plugin/core/registry.py
TheSecEng: Possibly, it would make the UX a bit smoother.

plugin/core/windows.py
TheSecEng: This was in relation to erase_sessions calling self._document_states.clear()

Is that okay to do even if just restarting a single server?

TheSecEng:
> It's a step in the right direction, but I already have plans for this involving the `Session` object. Please keep the changes to a single topic for a pull request :)

On hold basically until this is done, as I don't want to implement a half-baked solution

plugin/core/windows.py
rwols: You can go ahead with this PR now.

TheSecEng: Awesome, will work this on the weekend. 3 days of freedom

rchl:
BTW. What's the reasoning for adding this?

I just personally don't see a benefit in restarting just a specific server. There is no state to loose when restarting a server really and performance shouldn't be an issue either.

TheSecEng:
> BTW. What's the reasoning for adding this?
> 
> I just personally don't see a benefit in restarting just a specific server. There is no state to loose when restarting all servers really and performance shouldn't be an issue either.

@rwols should I move forward still? Do you think there is still value to be had with this?

rchl:
> Do you think there is still value to be had with this?

I believe that whenever I have triggered "restart servers", I really meant to restart one specific server. So then maybe it would make sense to just have one command `LSP: Restart Server` that would show a list of servers.

I just can't see the point of having two commands for the restart functionality as it seems like too much.

TheSecEng:
> > Do you think there is still value to be had with this?
> 
> I believe that whenever I have triggered "restart servers", I really meant to restart one specific server. So then maybe it would make sense to just have one command `LSP: Restart Server` that would show a list of servers.
> 
> I just can't see the point of having two commands for the restart functionality as it seems like too much.

I agree, and will update this PR to reflect this

plugin/core/registry.py
```diff
@@ -104,31 +104,24 @@ def sessions(self, capability: Optional[str] = None) -> Generator[Session, None,
...
             window.show_quick_panel(sessions, lambda index: self.restart_server(index, sessions))
 
     def restart_server(self, index, sessions) -> None:
         if index == -1:
             return
-        windows.lookup(self.view.window()).restart_session(config_name=sessions[index])
+        windows.lookup(self.view.window()).restart_session_async(config_name=sessions[index])
         return
+
+    def sessions(self, capability: Optional[str] = None) -> Generator[Session, None, None]:
```
rchl, line 126: You can inherit from LspTextCommand which will give you `is_enabled` and `sessions` implementations already.

plugin/core/windows.py
```diff
@@ -270,6 +270,11 @@ def handle_message_request(self, session: Session, params: Any, request_id: Any)
         if view:
             MessageRequestHandler(view, session, request_id, params, session.config.name).show()
 
+    def restart_session_async(self, config_name: str) -> None:
```
rchl, line 273: Duplicated it on rebase I suppose.

plugin/core/registry.py
```diff
@@ -105,8 +104,24 @@ def sessions(self, capability: Optional[str] = None) -> Generator[Session, None,
...
+            return
+
+        sessions = [session.config.name for session in self.sessions()]
+        if sessions:
+            window.show_quick_panel(sessions, lambda index: self.restart_server(index, sessions))
+
+    def restart_server(self, index, sessions) -> None:
+        if index == -1:
+            return
+        windows.lookup(self.view.window()).restart_session_async(config_name=sessions[index])
```
TheSecEng, line 123: @rchl Any idea?

```
plugin/core/registry.py:123: error: Argument 1 to "lookup" of "WindowRegistry" has incompatible type "Optional[Window]"; expected "Window"
```

rchl: "if" check the window

TheSecEng: Oh wow, learn something new everyday

plugin/core/registry.py
```diff
@@ -105,8 +104,20 @@ def sessions(self, capability: Optional[str] = None) -> Generator[Session, None,
         yield from sessions_for_view(self.view, capability)
 
 
-class LspRestartClientCommand(sublime_plugin.TextCommand):
+class LspRestartServerCommand(LspTextCommand, sublime_plugin.TextCommand):
```
rchl, line 107: LspTextCommand already inherits the sublime text command so drop the second argument

plugin/core/registry.py
```diff
@@ -105,8 +104,20 @@ def sessions(self, capability: Optional[str] = None) -> Generator[Session, None,
...
+            window.show_quick_panel(sessions, lambda index: self.restart_server(index, sessions))
+
+    def restart_server(self, index: int, sessions: list) -> None:
+        if index == -1:
+            return
+        window = self.view.window()
         if window:
-            windows.lookup(window).restart_sessions_async()
+            windows.lookup(window).end_config_sessions_async(config_name=sessions[index])
+        return
```
rchl, line 123: Return at the end? Not needed

plugin/core/registry.py
```diff
@@ -105,8 +104,19 @@ def sessions(self, capability: Optional[str] = None) -> Generator[Session, None,
...
     def run(self, edit: Any) -> None:
         window = self.view.window()
+        if not window:
+            return
+
+        sessions = [session.config.name for session in self.sessions()]
+        if sessions:
+            window.show_quick_panel(sessions, lambda index: self.restart_server(index, sessions))
+
+    def restart_server(self, index: int, sessions: list) -> None:
```
rchl, line 117: `sessions: List[str]` to be more specific.

plugin/core/registry.py
```diff
@@ -105,8 +104,19 @@ def sessions(self, capability: Optional[str] = None) -> Generator[Session, None,
...
 
 
-class LspRestartClientCommand(sublime_plugin.TextCommand):
+class LspRestartServerCommand(LspTextCommand):
     def run(self, edit: Any) -> None:
         window = self.view.window()
+        if not window:
+            return
+
+        sessions = [session.config.name for session in self.sessions()]
```
rchl, line 113: Please call it something better. These are not sessions but config names rather.

plugin/core/registry.py
```diff
@@ -105,8 +104,19 @@ def sessions(self, capability: Optional[str] = None) -> Generator[Session, None,
...
+        if sessions:
+            window.show_quick_panel(sessions, lambda index: self.restart_server(index, sessions))
```
rchl, line 115: You could pass the window through the lambda to simplify the code (no need to lookup or null-check) but no big deal.

rchl:
You should also update `docs/features.md` which mentions now-removed command.

plugin/core/registry.py
TheSecEng: That means something along the lines of this? correct?

```python
window.show_quick_panel(config_names, lambda index: self.restart_server(window, index, config_names) if window else False)
```

rwols:
> I think it would be better to list all configs in the window rather than active file since the command restarts sessions for the window, not active file.

That's not correct because a config name doesn't uniquely identify a session (per window). A config name together with a set of workspace folders uniquely identifies a session (per window) (because of the workspace-unaware sessions).

This named argument `config_name=` is wrong as it might restart the wrong session with the same name (for example, gopls is workspace-unaware, so you can have two gopls instances running in a .sublime-project with two go root folders)

There's probably many ways to do it right, but one thing that you can do is to use the `view` that you have in the TextCommand, iterate over the sessions from the WindowManager, and ask `session.handles_path` for each of these.

rwols:
Oh well, since `end_config_sessions_async` already exists I think it's fine to use in this PR.

plugin/core/registry.py
```diff
@@ -105,8 +104,17 @@ def sessions(self, capability: Optional[str] = None) -> Generator[Session, None,
...
+            return
+
+        config_names = [session.config.name for session in self.sessions()]
+        if config_names:
+            window.show_quick_panel(config_names, lambda index: self.restart_server(window, index, config_names))
+
+    def restart_server(self, window: sublime.Window, index: int, config_names: List[str]) -> None:
+        if index == -1:
+            return
+        windows.lookup(window).end_config_sessions_async(config_name=config_names[index])
```
rwols, line 120: `end_config_sessions_async` ends with `_async`, so it should run in ST's worker thread. This snippet currently runs in the main thread. That means that WindowManager._sessions could be mutated while being iterated over, resulting in exceptions and angry users.

plugin/core/registry.py
jwortmann: Suggestion:
```python
        if len(config_names) == 1:
            self.restart_server(window, 0, config_names)
        elif config_names:
            window.show_quick_panel(config_names, lambda index: self.restart_server(window, index, config_names))
```

plugin/core/registry.py
TheSecEng: Sorry I am not more familiar with the codebase / python. Are you meaning it should run via a `sublime.set_timeout`

rwols: It should be run via `sublime.set_timeout_async`. That makes it run on the async thread. No need to be sorry! :) Also this is not really a python thing, it's just an ST peculiarity

TheSecEng:
> Oh well, since `end_config_sessions_async` already exists I think it's fine to use in this PR.

I see, well, I could leave it as is or implement

```python
    def end_view_session_async(self, view: sublime.View, session: Session) -> None:
        if session.can_handle(view):
            session.end_async()
            self._sessions.remove(session)
```

TheSecEng:
sorry for the delay, will get this updated this weekend

rchl:
Superseded by #1733

PR #1024, opened 2020-05-06
Run linting on CI for ST4

rchl:


rwols:
I'm just now noticing this is for the master branch. But given that I rebased st4000-exploration onto your commits I think if I rebase this as well git will not complain.

rchl:
It's the same commit (content-wise) so it should be smooth.

PR #1025, opened 2020-05-06
Run linting on CI for ST4

rchl:


rwols:
```
plugin/completion.py:112: error: Incompatible types in assignment (expression has type "Optional[Any]", variable has type "List[Dict[str, str]]")
```
I think this is my fault :) Fixing it as part of the pr is OK in my book.

rchl:
We'll have to update required checks since names have changes and required checks for st4000-exploration branch.

I have also created https://github.com/sublimelsp/LSP/pull/1024 for master but they will be the same. I guess it doesn't matter if we merge both st3 and st4 separately or just master and then merge.

rchl:
> I think this is my fault :) Fixing it as part of the pr is OK in my book.

I'm not sure how to handle that. The type comment seems correct so how to coerce mypy into accepting it? :)

.github/workflows/main.yml
```diff
@@ -3,48 +3,59 @@ name: main
...
+          sh ci.sh install_package_control
+          subl --stay &
+          sleep 8
+          sh ci.sh run_tests --coverage
+        name: Run UnitTesting tests
+      - run: |
+          pip3 install coverage==4.5.4 codecov==2.0.15
+          codecov
+
+  Mac-lint:
```
rchl, line 41: Should we even run lint on both mac and Linux? I can't imagine the lint issues to be platform-specific.

rchl: I've only left linting running on linux. Lemme know what you think.

PR #1026, opened 2020-05-06
Only underline diagnostics that are enabled with show_diagnostics_severity_level

predragnikolic:


predragnikolic:
Some useful feedback from discord regarding this PR.
 
> I does seem like the right thing to do but it's a "breaking" change as it changes default behavior. I'm currently happy with defaults - that is: show_diagnostics_severity_level being 2 and me still seeing all issues (including notices) in the view. So after your change I would need to change level to 4 to have the same. So then maybe the default should also change so that users don't have to experience a "breaking change".
That would change behavior of diagnostics panel but that might be less intrusive change

I think I am going to change the default for `auto_show_diagnostics_panel_level` from `2` to `4` so it doesn't break the current user will continue to see green underlines in the view. (but they will start seeing hints in the Diagnostic panel)

plugin/diagnostics.py
```diff
@@ -252,7 +252,7 @@ def end_file(self, file_name: str) -> None:
         self._relevant_file = False
 
     def end(self) -> None:
-        for severity in range(DiagnosticSeverity.Error, DiagnosticSeverity.Hint + 1):
+        for severity in range(settings.show_diagnostics_severity_level + 1):
```
rwols, line 255: This is `show_diagnostics_severity_level`, but in LSP.sublime-settings you changed `auto_show_diagnostics_panel_level`?

The `show_diagnostics_severity_level` is still at `2`. I think you meant to change that?

predragnikolic: I made a mistake, I didn't intend to change the level for the diagnostic panel. I will update that. 
Thanks :)

PR #1027, opened 2020-05-06
Code actions on save - WIP

rchl:
This is a very early work-in-progress. There are many issues to address.

## TODO
 - [x] 1. On triggering CAOS (Code Actions On Save), only request code actions from servers that report `codeActionProvider.codeActionKinds` and only request those specific kinds. ESlint reacts very weirdly if we send `source.fixAll.stylelint` to it, for example.
 - [x] 2. Implement more sophisticated `lsp_code_actions_on_save` setting handling. There are various combinations of that setting that can be applied by the user. There is a couple of examples in [vscode-stylelint readme](https://github.com/stylelint/vscode-stylelint#editorcodeactionsonsave). And while having `source.fixAll` set, we have to send that one to ESlint as its own, more specific kind, matches this more-generic `source.fixAll` one. When having both `source.fixAll: true` and `source.fixAll.eslint: false` we must not send the request.
 - [x] 3. Making a change and saving quickly fails to correct fixes. That's because, on a change, we clear diagnostics and send `didChange` notification, then find no diagnostics so we don't fire code action request. Only after that, we get diagnostics but it's too late.
 - [x] 4. Abandon sync handling in favor of overriding `save` (and `save_all`) command. Sync request blocks the main thread which is not an ideal experience. Especially if the fix takes a second or more.
 - [x] 5. Request and apply fixes from all servers on saving. That means we have to re-request code actions (which also means we have to wait for diagnostics) between each batch of fixes. And we need to request them until we exhaust all fixes or we time out...
 - [ ] 6. More that I forgot... Will add later.

## Questions/doubts
 - [ ] Should we only apply fixes with TextEdit. `Command`s need extra round-trip.
 - [ ] When we get fixes from multiple servers then what do we do? It's possible that fixes from one server will invalidate fixes from another server. It's fine if the fixes were versioned with `VersionedTextDocumentIdentifier` but otherwise, they might apply to a document that has already changed.

## Related links
[Code Action Request Spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_codeAction)

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -1,14 +1,17 @@
-import sublime
 from .core.edit import parse_workspace_edit
 from .core.protocol import Diagnostic
 from .core.protocol import Request, Point
-from .core.registry import LspTextCommand, LSPViewEventListener
+from .core.registry import LspTextCommand
+from .core.registry import LSPViewEventListener
 from .core.registry import sessions_for_view, client_from_session
+from .core.registry import windows
 from .core.settings import settings
 from .core.typing import Any, List, Dict, Callable, Optional, Union, Tuple, Mapping, TypedDict
 from .core.url import filename_to_uri
 from .core.views import entire_content_range, region_to_range
 from .diagnostics import filter_by_point, view_diagnostics
+import sublime
+import sublime_plugin
 
 CodeActionOrCommand = TypedDict('CodeActionOrCommand', {
     'title': str,
@@ -44,7 +47,8 @@ class CodeActionsCollector(object):
 
     def _notify_if_all_finished(self) -> None:
         if self._all_requested and len(self._requested_configs) == len(self._commands_by_config):
-            self._on_complete_handler(self._commands_by_config)
+            # Call back on the main thread
+            sublime.set_timeout(lambda: self._on_complete_handler(self._commands_by_config))
 
     def get_actions(self) -> CodeActionsByConfigName:
         return self._commands_by_config
@@ -125,7 +129,7 @@ def request_code_actions_on_save(
                 if matching_kinds:
                     params = _create_code_action_request_params(view, file_name, [], matching_kinds)
                     request = Request.codeAction(params)
-                    session.client.execute_request(
+                    session.client.send_request(
                         request, filtering_collector(session.config.name, matching_kinds, actions_collector))
     return actions_collector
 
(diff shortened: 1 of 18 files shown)


unittesting.json
```diff
@@ -1,11 +1,12 @@
...
-    "reload_package_on_testing": true,
+    "reload_package_on_testing": false,
     "start_coverage_after_reload": false,
     "show_reload_progress": false,
     "output": null,
-    "generate_html_report": true
-}
+    "pattern": "test_code_actions.py",
+    "generate_html_report": false
+}
```
predragnikolic, line 12: `"ensure_newline_at_eof_on_save": true` ;)

predragnikolic: just kidding :)
I am watching this PR, let the CAOS begin :popcorn:

rchl: But this file is auto-generated by `tests/adjust_unittesting_config_for_ci.py` when running tests the CI-way so not my fault. :)

predragnikolic: Tell that git blame ;)

rchl:
For point 3, I've realized that when asking for code actions on save, we don't need to send diagnostics with the request. Our code currently only sends code action request when it finds diagnostics but we don't have them when making modification and saving immediately.

rwols:
Please let this be the last thing before we put the ST3 branch in maintenance mode. The amount of changes between `master` and `st4000-exploration` is becoming large.

rchl:
> Please let this be the last thing before we put the ST3 branch in maintenance mode. The amount of changes between `master` and `st4000-exploration` is becoming large.

One could argue though that having this in both branches will just make the maintenance easier as there will be fewer differences between branches.

rchl:
Status:
- Sync approach has an issue that I'm not sure is fixable. When we are blocking while saving, we don't get `on_text_changed` notifications from ST so servers don't get notified about content changes. That's OK-ish if I want to apply just one round of fixes but if I want to apply fixes until all options are exhausted (point 5) then we're screwed.
- As I've just discussed on Discord, there is no way to override `save` command using command listeners. `save` command doesn't go through those (see https://discord.com/channels/280102180189634562/280157067396775936/716017902762983517). So the suggestion I've gotten is to add an own keybinding for `save` but with a context check (like `lsp_active`) so that it would trigger instead of the original one. That wouldn't be as generic as overriding `save` command so it wouldn't catch saving from main menu, for example (and possibly other cases). And I'm yet to investigate if that would even work for CAOS.

rchl:
The async approach works much nicer IMO - noticeably more responsive.

Since I'm working on this in parallel in ST3 & ST4, I'm starting to notice more and more differences in the code that is making it more difficult to maintain both versions. So I wonder if we should just do this for ST4 only...

One reason for this is that ST3 has this bug with emojis (or UTF16 chars in general) that can cause issues when auto-fixing (also applies to format-on-save). So I'm not sure it's worth to claim this functionality for ST3 when it can break the code.

predragnikolic:
Do you guys think that it is time to feature freeze LSP ST3? (I do)

rchl:
@rwols ?

rwols:
I agree, the changes between master and ST4 are getting so large that it’s a burden to maintain. I hope to see a public release of ST4 soon anyway.

plugin/code_actions.py
```diff
@@ -1,60 +1,74 @@
...
+    'kind': Optional[str]
 }, total=False)
 CodeActionsResponse = Optional[List[CodeActionOrCommand]]
 CodeActionsByConfigName = Dict[str, List[CodeActionOrCommand]]
 
 
-class CodeActionsAtLocation(object):
-
-    def __init__(self, on_complete_handler: Callable[[CodeActionsByConfigName], None]) -> None:
+class CodeActionsCollector(object):
```
rwols, line 26: Can you add a description here of what this class does?

plugin/code_actions.py
```diff
@@ -1,60 +1,74 @@
...
+        if self._all_requested and len(self._requested_configs) == len(self._commands_by_config):
+            # Call back on the main thread
+            sublime.set_timeout(lambda: self._on_complete_handler(self._commands_by_config))
+
+    def get_actions(self) -> CodeActionsByConfigName:
+        return self._commands_by_config
 
 
 class CodeActionsManager(object):
     """ Collects and caches code actions"""
```
rwols, line 58: The description here is strange, because there's already a "CodeActionsCollector" class above.

rchl:
Obsoleted by ST4 PR #1087

PR #1030, opened 2020-05-08
Support capability argument for sessions_for_view

rchl:
- Allow passed capability to be in 'dotted' format like 'textDocumentSync.change'
- Replace some checks to use get_dotted_value to make code more compact

The logic of changed code is not necessarily 100% identical to the old
one (empty dicts will return True now) but I think it shouldn't make
a difference in practice.

This change merges cleanly with st4000-exploration branch also.

### DIFF
--- a/plugin/core/registry.py
+++ b/plugin/core/registry.py
@@ -73,12 +73,14 @@ def client_from_session(session: Optional[Session]) -> Optional[Client]:
     return session.client if session else None
 
 
-def sessions_for_view(view: sublime.View, capability: str = None, point: int = None) -> Iterable[Session]:
+def sessions_for_view(view: sublime.View, capability: Optional[str] = None,
+                      point: Optional[int] = None) -> Iterable[Session]:
     return (session for session in _sessions_for_view_and_window(view, view.window(), point)
             if not capability or session.has_capability(capability))
 
 
-def session_for_view(view: sublime.View, capability: str, point: int = None) -> Optional[Session]:
+def session_for_view(view: sublime.View, capability: Optional[str],
+                     point: Optional[int] = None) -> Optional[Session]:
     return next((session for session in sessions_for_view(view, capability, point)), None)
 
 
--- a/plugin/core/registry.py
+++ b/plugin/core/registry.py
@@ -73,15 +73,13 @@ def client_from_session(session: Optional[Session]) -> Optional[Client]:
     return session.client if session else None
 
 
-def sessions_for_view(view: sublime.View, point: Optional[int] = None) -> Iterable[Session]:
-    return _sessions_for_view_and_window(view, view.window(), point)
+def sessions_for_view(view: sublime.View, capability: str = None, point: int = None) -> Iterable[Session]:
+    return (session for session in _sessions_for_view_and_window(view, view.window(), point)
+            if not capability or session.has_capability(capability))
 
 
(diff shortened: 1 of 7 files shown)


plugin/core/registry.py
```diff
@@ -73,15 +73,13 @@ def client_from_session(session: Optional[Session]) -> Optional[Client]:
...
+    return (session for session in _sessions_for_view_and_window(view, view.window(), point)
+            if not capability or session.has_capability(capability))
 
 
-def session_for_view(view: sublime.View,
-                     capability: str,
-                     point: Optional[int] = None) -> Optional[Session]:
-    return next((session for session in sessions_for_view(view, point)
-                 if session.has_capability(capability)), None)
+def session_for_view(view: sublime.View, capability: str, point: int = None) -> Optional[Session]:
```
rchl, line 81: So when we use default argument then we don't need `Optional` annotation? It works either way, not sure what's preferred though.

rwols: Odd. It might be a mypy bug. It should definitely be `point: Optional[int] = None`, if only for communication to the reader.

rchl: No, I think it works either way with mypy.

tests/test_session.py
```diff
@@ -245,7 +247,6 @@ def test_document_sync_capabilities(self) -> None:
         self.assertEqual(session.text_sync_kind(), TextDocumentSyncKindIncremental)
         self.assertTrue(session.should_notify_did_change())
         self.assertFalse(session.should_notify_will_save())  # old-style text sync will never send willSave
-        self.assertFalse(session.should_request_will_save_wait_until())
         self.assertEqual(session.should_notify_did_save(), (False, False))
```
rwols, line 250: If the lang server is "old" and has this in its server capabilities:

```
'capabilities': {
  'textDocumentSync': 2
}
```
Does `has_capability('textDocumentSync.change')` still return True?

rchl: Not in this case since it doesn't include `change` property.

rwols: Should it return True? The server can do incremental text sync, I think that's what one asks when querying `has_capability('textDocumentSync.change')`?

Does `has_capability` mean "Is this dotted path present" or does it mean "does the server have this capability" ?

rchl: I'd say it's just "is this dotted path present and not False or Null".
This is also how previous implementation of `has_capability` worked really. We have more specific methods for capabilities checks that need more fancy logic.

plugin/core/sessions.py
```diff
@@ -242,21 +242,17 @@ def __init__(self,
         self._initialize()
 
     def has_capability(self, capability: str) -> bool:
-        return capability in self.capabilities and self.capabilities[capability] is not False
+        value = self.get_capability(capability)
+        return value is not False and value is not None
```
rwols, line 246: I remember we had to account for `registerOptions` not being present. In that case I think we place a `None` value in `self.capabilities`?

```
 set_dotted_value(self.capabilities, capability_path, registration.get("registerOptions"))
```

Perhaps this should become:


```
 set_dotted_value(self.capabilities, capability_path, registration.get("registerOptions", {}))
```
to avoid scratching our heads in the future.

rchl: Good point. Will change.

PR #1031, opened 2020-05-09
Textdocument symbol

### LINKED ISSUE #1009: Document symbol panel should select ranges without committing
The built-in "Goto Symbol" panel selects the symbol range when the selection changes even without committing by pressing return. The LSP command "Document Symbols" should have the same functionality but currently it does not.

rwols:
- When the quick panel opens, do not jump to the first selected symbol the
  first time.
- Subsequent "on_selected" callbacks do jump to the symbol.
- The viewport is focused to the region's start point.
- add_regions is used. This is because selecting very large regions looks a bit
  weird. add_regions is more neutral.
- Restore selections when the user cancels.
- Proper distinction between DocumentSymbol and SymbolInformation LSP types.
- hierarchicalDocumentSymbolSupport added in ClientCapabilities
- We have to use separate TextCommands for clearing the selection and for
  adding regions to the selection, because these mutations must be under a
  sublime.Edit object.

close #1009


rwols:
A server returning SymbolInformation (pyls):

symbol-information

A server returning DocumentSymbol (clangd):

document-symbols

Keymaps/Default (Linux).sublime-keymap
```diff
@@ -54,8 +54,8 @@
     // Format Selection
     // {"keys": ["UNBOUND"], "command": "lsp_format_document_range", "context": [{"key": "setting.lsp_active"}]},
 
-    // Document Symbols
-    // {"keys": ["UNBOUND"], "command": "lsp_document_symbols", "context": [{"key": "setting.lsp_active"}]},
+    // Document Symbols (a replacement for ST's "Goto Symbol")
```
rchl, line 57: `ctrl+r` is `next_bookmark` though.

Keymaps/Default (Linux).sublime-keymap
```diff
@@ -54,8 +54,8 @@
...
+    // {"keys": ["ctrl+r"], "command": "lsp_document_symbols", "context": [{"key": "setting.lsp_active"}]},
```
rchl, line 58: `ctrl+r` is `next_bookmark` though.

plugin/symbols.py
```diff
@@ -42,53 +41,118 @@ def format_symbol_kind(kind: int) -> str:
...
+        return self.regions[index][0]
+
+    def selection_region(self, index: int) -> Optional[sublime.Region]:
+        return self.regions[index][1]
+
+    def on_symbol_selected(self, index: int) -> None:
+        if index == -1:
+            if len(self.old_regions) > 0:
+                self.view.run_command("lsp_selection_add", {"regions": [(r.a, r.b) for r in self.old_regions]})
+                self.view.show_at_center(self.old_regions[0].b)
```
rchl, line 106: Maybe use `begin()` instead of `b` to make sure that we center on the start of the region?

Keymaps/Default (Linux).sublime-keymap
rwols: I did some research. It is `super+r` on macOS, `ctrl+r` on Linux and Windows.

plugin/symbols.py
```diff
@@ -42,53 +41,118 @@ def format_symbol_kind(kind: int) -> str:
...
+        return self.regions[index][1]
+
+    def on_symbol_selected(self, index: int) -> None:
+        if index == -1:
+            if len(self.old_regions) > 0:
+                self.view.run_command("lsp_selection_add", {"regions": [(r.a, r.b) for r in self.old_regions]})
+                self.view.show_at_center(self.old_regions[0].begin())
+        else:
+            region = self.selection_region(index) or self.region(index)
+            self.view.run_command("lsp_selection_add", {"regions": [(region.a, region.a)]})
```
jwortmann, line 109: Just looking at this file here... is this a typo and should be `(region.a, region.b)`?

rwols: It's a matter of preference. With (region.a, region.b) it will select the whole entity. While the built-in Goto Symbol indeed does that, the behavior is weird for this feature. Because the region can be large (e.g. a body of a large function, or a namespace). This is also the region why `add_regions` and `erase_regions` is used instead of "highlighting" entities with `view.sel()`. The regions for the built-in Goto Symbol never span more than 1 line.

jwortmann: All fine then, thanks for explanation.

PR #1035, opened 2020-05-11
Add suggested and fairly complete configuration for pyls

rchl:


docs/index.md
```diff
@@ -757,7 +757,70 @@ If you use a virtualenv for your current project, add a path to it in your [proj
 }
 ```
 
-See: [github:palantir/python-language-server](https://github.com/palantir/python-language-server)
+A basic configuration below can be used for bootstraping your own:
```
rwols, line 760: bootstraping -> bootstrapping

docs/index.md
```diff
@@ -757,7 +757,70 @@ If you use a virtualenv for your current project, add a path to it in your [proj
...
+        // "flake8",  // discovered in ~/.config/flake8, setup.cfg, tox.ini and flake8.cfg
+      ],
+      "plugins": {
+        "jedi": {
+          "extra_paths": [
+            // The directory where the pip installation package is located
+          ],
+        },
+        "jedi_completion": {
+          "fuzzy": false,  // Enable fuzzy when requesting autocomplete
```
rwols, line 795: Set this to `true` :) It probably works well with ST's own fuzzy string matcher for completions.

rchl: Will do but it seems fuzzy to me even without it:

Screenshot 2020-05-11 at 10 56 56

I've triggered this completion on typing `m` in `fnm`.

jwortmann:
I had copied the settings with default values (I hope) from VS Code into my LSP settings before:
```js
      "settings": {
        "pyls": {
          "plugins": {
            "jedi": {
              "extra_paths": [], // Define extra paths for jedi.Script (type: Array[String])
              "environment": null // Define environment for jedi.Script and Jedi.names  (type: String)
            },
            "jedi_completion": {
              "enabled": true, // Enable or disable the plugin
              "include_params": true, // Auto-completes methods and classes with tabstops for each parameter
              "fuzzy": false // Enable fuzzy when requesting autocomplete
            },
            "jedi_definition": {
              "enabled": true, // Enable or disable the plugin
              "follow_imports": true, // The goto call will follow imports
              "follow_builtin_imports": true // If follow_imports is True will decide if it follow builtin imports
            },
            "jedi_hover": {
              "enabled": true // Enable or disable the plugin
            },
            "jedi_references": {
              "enabled": true // Enable or disable the plugin
            },
            "jedi_signature_help": {
              "enabled": true // Enable or disable the plugin
            },
            "jedi_symbols": {
              "enabled": true, // Enable or disable the plugin
              "all_scopes": true // If True lists the names of all scopes instead of only the module namespace
            },
            "mccabe": {
              "enabled": true, // Enable or disable the plugin
              "threshold": 15 // The minimum threshold that triggers warnings about cyclomatic complexity
            },
            "preload": {
              "enabled": true, // Enable or disable the plugin
              "modules": null // List of modules to import on startup (type: Array[String])
            },
            "pycodestyle": {
              "enabled": true, // Enable or disable the plugin
              "exclude": null, // Exclude files or directories which match these patterns (type: Array[String])
              "filename": null, // When parsing directories, only check filenames matching these patterns (type: Array[String])
              "select": null, // Select errors and warnings (type: Array[String])
              "ignore": null, // Ignore errors and warnings (type: Array[String])
              "hangClosing": null, // Hang closing bracket instead of matching indentation of opening bracket's line (type: Boolean)
              "maxLineLength": null // Set maximum allowed line length (type: Number)
            },
            "pydocstyle": {
              "enabled": false, // Enable or disable the plugin
              "convention": null, // Choose the basic list of checked errors by specifying an existing convention ("pep257" | "numpy")
              "addIgnore": null, // Ignore errors and warnings in addition to the specified convention (type: Array[String])
              "addSelect": null, // Select errors and warnings in addition to the specified convention (type: Array[String])
              "ignore": null, // Ignore errors and warnings (type: Array[String])
              "select": null, // Select errors and warnings (type: Array[String])
              "match": "(?!test_).*\\.py", // Check only files that exactly match the given regular expression; default is to match files that don't start with 'test_' but end with '.py'
              "matchDir": "[^\\.].*" // Search only dirs that exactly match the given regular expression; default is to match dirs which do not begin with a dot
            },
            "pyflakes": {
              "enabled": true // Enable or disable the plugin
            },
            "pylint": {
              "enabled": false, // Enable or disable the plugin
              "args": null // Arguments to pass to pylint (type: Array[String])
            },
            "rope_completion": {
              "enabled": true // Enable or disable the plugin
            },
            "yapf": {
              "enabled": true // Enable or disable the plugin
            }
          }
        }
      },
```

rchl:
We could add all I guess. I was sticking to some that would be worth having (in my mind).

AmjadHD:
@jwortmann just a side note in case anyone will need this in the future, i made a [script](https://github.com/AmjadHD/vsc_to_st) a while ago to automate converting from vscode to ST settings.

rwols:
@AmjadHD You might want to contribute that for https://github.com/sublimelsp/lsp_utils/issues/10

rchl:
But how would lsp_utils utilize that?
I would think we would rather do the conversion at package development time

rwols:
Maybe such a script can live in the LSP package. Expose it as an ST command like

```
LSP: Parse external package.json settings
```
Then open an input panel and let the user paste in a URL to a file, download the file, parse settings, dump output into a scratch view.

PR #1037, opened 2020-05-11
Use a WeakValueDictionary for storing the window managers

rwols:
This allows us to forget about the on_closed callback.
The self._workspace (ProjectFolders class) had a strong reference
to its containing WindowManager.

To check what the referrers are of a Python object, you can do:

    import gc
    import pprint
    pprint.pprint(gc.get_referrers(some_object))

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -21,6 +21,8 @@ from .workspace import enable_in_project
 from .workspace import get_workspace_folders
 from .workspace import ProjectFolders
 from .workspace import sorted_workspace_folders
+from weakref import WeakValueDictionary
+from weakref import ref
 import threading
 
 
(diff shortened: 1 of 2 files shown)


plugin/core/windows.py
```diff
@@ -310,12 +311,27 @@ def __init__(
...
-        self._workspace.on_changed = self._on_project_changed
-        self._workspace.on_switched = self._on_project_switched
+        weakself = ref(self)
+
+        # A weak reference is needed, otherwise self._workspace will have a strong reference to self, meaning a
+        # cyclic dependency.
+        def on_changed(folders: List[str]) -> None:
+            this = weakself()
+            if this is not None:
+                this._on_project_changed(folders)
```
rchl, line 324: Not sure if it would be of any use here but I've recently stumbled upon this code: https://github.com/SublimeText/PackageDev/blob/master/plugins/lib/weakmethod.py

rwols: Interesting. WeakMethod was introduced in py3.4. I was also thinking of backporting it to py3.3, but decided that it was okay to use the code above and wait for us to switch to py3.8.

PR #1039, opened 2020-05-12
document workspace/executeCommand

ayoub-benali:


rchl:
How about also adding a `debug` print for logging `executeCommandProvider. commands` so that the commands are at least somewhat discoverable?
Something like:

```py
execute_commands = session.get_capability('executeCommandProvider.commands')
if execute_commands:
  debug("{}: Server supported execute commands: {}".format(session.config.name, execute_commands))
```

PR #1040, opened 2020-05-12
Add more gutter icons for diagnostics

### LINKED ISSUE #1032: Additional gutter icons for diagnostics severities
I want to ask what you think about an additional gutter icon set for the different diagnostics severities. I adjusted a few lines of the code and added another option value for the "diagnostics_gutter_marker" setting (e.g. "symbols") for it. Here is how it looks like in the Monokai, Mariana and Brackets schemes (f.l.t.r.):

icons

The icons source are colored [SVG icons](https://github.com/microsoft/vscode-codicons) from VS Code. I removed the color and converted them to PNG for use in Sublime.

I didn't made a pull request yet because I wanted to hear some opinions first, and want to be sure to have no license violation. The original icons are released under the [Creative Commons Attribution 4.0 International Public License](https://creativecommons.org/licenses/by/4.0/legalcode), so I believe we could use them as long as we give attribution and include the license file.

What do you think about it?

jwortmann:
Add 3 more gutter icons for diagnostics: `error`, `warning`, `info` (in 3 sizes each), derived from https://github.com/microsoft/vscode-codicons.

Can be activated via a new option `"sign"` for the `"diagnostics_gutter_marker"` setting (please comment if you prefer another name - maybe "symbol"?). 

For preview images and more background info see https://github.com/sublimelsp/LSP/issues/1032#issuecomment-627017715

On high DPI displays the 32x32 pixel or 48x48 pixel versions are probably used. Note that the relative icon sizes strongly depend on the used font and font size setting! Therefore feedback with different display resolutions and your font settings would be welcome, to possibly adjust the icon size a bit if desired.

Closes #1032


rchl:
I would suggest making it default but unfortunately, I find it not super readable so maybe better to keep it as an option. We need SVG support...

rchl:
Or maybe some post-processing would help with readability? I'm not a designer but there is quite a bit of antialiasing which on such a small icon might do more bad than good.

jwortmann:
There should be no aliasing because I converted each size from the source SVG separetely. Perhaps you are experiencing the blurry rescaling effect I have mentioned before? You could try to open the PNG files with an image editing program, remove 1 pixel vertically and horizontally and see whether that helps. I would assume the higher pixel sizes are used on your display though.

rchl:
The right one is 15px.

Screenshot 2020-05-12 at 21 45 01

I find them both quite hard to read.
This is QHD screen so using 1x image.

LSP.sublime-settings
```diff
@@ -49,7 +49,7 @@
   },
 
   // Gutter marker for code diagnostics.
-  // Valid values are "bookmark", "circle", "cross", "dot" or ""
+  // Valid values are "bookmark", "circle", "cross", "dot", "sign" or ""
```
rchl, line 52: BTW. `dot` doesn't work.
It's not part one of the standard icons. Documentation says:
```
An optional unicode string specifying an icon to draw in the gutter next to each region. The icon will be tinted using the color associated with the scope. Standard icon names are "dot", "circle" and "bookmark". The icon may also be a full package-relative path, such as "Packages/Theme - Default/dot.png".
```
We should probably remove that option.

jwortmann: `dot` does work for me. You can even use just `"branch"` from the default theme in LSP, even though it is not documented.

jwortmann:
>I find them both quite hard to read.

Hm, I indeed can't spot much difference in your screenshot. Would you suggest to make them even bigger then?

In my opinion they still look relatively well even that small. What's your font and font size?

LSP.sublime-settings
rchl: Sorry, I meant "cross" doesn't work.

rchl:
> Hm, I indeed can't spot much difference in your screenshot. Would you suggest to make them even bigger then?

No, I think we can't do much with it. We should just leave as is.

LSP.sublime-settings
jwortmann: Can confirm and removed it. By the way, `label` and `warning` (from the default theme) have the correct icon size and work as well, even though undocumented.

icons/LICENSE
```diff
@@ -0,0 +1,395 @@
+﻿Attribution 4.0 International
```
rwols, line 1: I am not a lawyer. Can someone explain what implications this massive wall of text has?

jwortmann: I'm not a lawyer as well, but the conditions are explained in detail in section 3, or for a shorter summary see for example https://choosealicense.com/licenses/cc-by-4.0/ .

From that text I'm pretty sure that all conditions are met here:
* gave attribution to original author and included a hyperlink
* stated modifications
* there is no copyright notice provided in the original repository, so no further files need to be included

Actually I believe we don't have to include the whole license file, as it only says to retain
> iii. a notice that refers to this Public License;

but I wanted to be on the very safe side even though a link should be sufficient. If we wanted we could use another license for the modified PNG icons too, but I think this creative commons license is no bad choice.

Also we are not allowed to advertise the modified PNG icons under the name of Microsoft (no trademark use), and we can't blame Microsoft if the icons will crash or destroy our computer somehow (no liability and warranty).

jwortmann: Now I actually wonder whether we need a notice and license for the already existing lightbulb.png too, or where exactly did it come from in #489 ?

jwortmann: We should probably copy the very bottom section (Legal notice) of the Codicons' Readme too. I'm gonna add it to the attribution and modifications notice.

rwols: > From that text I'm pretty sure that all conditions are met here:

Yes, I think so too.

PR #1044, opened 2020-05-13
Improve whitespace and link parsing in documentation popup

### LINKED ISSUE #1043: Non-ideal rendering of links in text content in documentation popup
Link includes `>` and is not working:
Screenshot 2020-05-13 at 20 07 52

rchl:
Resolves #1043

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -67,11 +67,12 @@ class LspResolveDocsCommand(sublime_plugin.TextCommand):
         return content
 
     def show_popup(self, minihtml_content: str) -> None:
+        viewport_width = self.view.viewport_extent()[0]
         mdpopups.show_popup(
             self.view,
             minihtml_content,
             flags=sublime.COOPERATE_WITH_AUTO_COMPLETE,
-            max_width=800,
+            max_width=viewport_width,
             allow_code_wrap=True,
             on_navigate=self.on_navigate
         )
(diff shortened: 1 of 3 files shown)


plugin/completion.py
```diff
@@ -71,8 +71,7 @@ def show_popup(self, minihtml_content: str) -> None:
             self.view,
             minihtml_content,
             flags=sublime.COOPERATE_WITH_AUTO_COMPLETE,
-            max_width=480,
-            max_height=410,
+            max_width=800,
```
rchl, line 74: Any specific reason why width was restricted?

rwols: We could try:

- `view.em_width() * 100` (does this mean approximately hundred characters per line?)
- `view.layout_extent()[0]` (This seems to be viewport minus sidebar?)
- `view.viewport_extent()[0]` (This seems to be the entire window)

rchl: What do we want to achieve with that?
I believe popup is already restricted to viewport size by default.

rwols: If you use a 4K screen and your viewport width is 3840 pixels wide, then 800 pixels is tiny.

rchl: If I don't specify `max_width` then it breaks very prematurely:
Screenshot 2020-05-13 at 22 12 22

`em_width() * 100`:
Screenshot 2020-05-13 at 22 11 52

`view.viewport_extent()[0]`:
Screenshot 2020-05-13 at 22 11 07

rwols: I'll let you be the judge. I will say that 800 is not future-proof :)

rchl: I'd probably choose some ridiculously high value since ST will clamp to viewport size anyway but there might be gotchas with it.
Let's go for viewport width for now.

rwols: Yeah I tried `sys.maxsize` but that made the popup appear at the location of the AC for some reason.

PR #1046, opened 2020-05-13
Support MarkupContent in function signatures and hover content

### LINKED ISSUE #991: Support for MarkupContent in hover

rchl:
Those two, together with documentation for completion items covers all
places where MarkupContent is currently supported.

Resolves #991

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -195,9 +195,12 @@ def minihtml(view: sublime.View, content: Union[str, dict, list], prefer_plain_t
 
     :returns: Formatted string
     """
+    is_plain_text = True
+    result = ''
     if isinstance(content, str):
         # plain text string or MarkedString
-        return text2html(content) if prefer_plain_text else mdpopups.md2html(view, content)
+        is_plain_text = prefer_plain_text
+        result = content
     if isinstance(content, list):
         # MarkedString[]
         formatted = []
@@ -215,8 +218,8 @@ def minihtml(view: sublime.View, content: Union[str, dict, list], prefer_plain_t
             else:
                 formatted.append(value)
 
-        frontmatter_config = mdpopups.format_frontmatter({'allow_code_wrap': True})
-        return mdpopups.md2html(view, frontmatter_config + "\n".join(formatted))
+        is_plain_text = False
+        result = "\n".join(formatted)
     if isinstance(content, dict):
         # MarkupContent or MarkedString (dict)
         language = content.get("language")
@@ -224,11 +227,17 @@ def minihtml(view: sublime.View, content: Union[str, dict, list], prefer_plain_t
         value = content.get("value") or ""
         if kind:
             # MarkupContent
-            return mdpopups.md2html(view, value) if kind == "markdown" else text2html(value)
+            is_plain_text = kind != "markdown"
+            result = value
         if language:
             # MarkedString (dict)
-            return mdpopups.md2html(view, "```{}\n{}\n```\n".format(language, value))
-    return ''
+            is_plain_text = False
+            result = "```{}\n{}\n```\n".format(language, value)
+    if is_plain_text:
+        return text2html(result)
+    else:
+        frontmatter_config = mdpopups.format_frontmatter({'allow_code_wrap': True})
+        return mdpopups.md2html(view, frontmatter_config + result)
 
 
 REPLACEMENT_MAP = {
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -182,7 +182,7 @@ def minihtml(view: sublime.View, content: Union[str, dict, list], prefer_plain_t
 
     Content can be in one of those formats:
      - string: treated as plain text
-     - MarkedString: string | { language: string; value: string }
+     - MarkedString: string or { language: string; value: string }
      - MarkedString[]
      - MarkupContent: { kind: MarkupKind, value: string }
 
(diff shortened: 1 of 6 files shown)


plugin/core/views.py
```diff
@@ -176,20 +176,58 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
...
+    :param content
+    :param prefer_plain_text: Whether to treat *string* input as plain text (True) or markdown (False).
+
+    :returns: Formatted string
+    """
     if isinstance(content, str):
-        return text2html(content)
-    elif isinstance(content, dict):
-        value = content.get("value") or ""
+        return text2html(content) if prefer_plain_text else mdpopups.md2html(view, content)
```
predragnikolic, line 199: I believe you have two places where you return MarkedString `string`
line 199 and line 229

```
        if language:
            # MarkedString (dict)
            return mdpopups.md2html(view, "```{}\n{}\n```\n".format(language, value))
        # MarkedString (string)
        return mdpopups.md2html(view, value)   # this here can be deleted
```

plugin/core/views.py
```diff
@@ -176,20 +176,58 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
...
+
+            if language:
+                formatted.append("```{}\n{}\n```\n".format(language, value))
+            else:
+                formatted.append(value)
+
+        frontmatter_config = mdpopups.format_frontmatter({'allow_code_wrap': True})
+        return mdpopups.md2html(view, frontmatter_config + "\n".join(formatted))
+    if isinstance(content, dict):
+        language = content.get("language")
```
predragnikolic, line 219: I really like how you added comments like `# MarkupContent` or `# MarkupContent` after each if statement.
can you add them after all ifs, it really useful.(I don't have to look it up or lose 1min tring to figure out if it is a MarkedString or MarkupKind or a list of MarkedStrings)

predragnikolic:
Nice, 
I will use this branch tomorrow at work and I will let you know if I notice something :)
On the first look, everything seems ok

plugin/core/views.py
rchl: True, the second one shouldn't be needed.

plugin/core/views.py
```diff
@@ -176,20 +176,58 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
...
+            value = ""
+            language = None
+            if isinstance(item, str):
+                value = item
+            else:
+                value = item.get("value") or ""
+                language = item.get("language")
+
+            if language:
+                formatted.append("```{}\n{}\n```\n".format(language, value))
```
rchl, line 212: It probably makes sense to use frontmatter with `allow_code_wrap` for all cases where we use `md2html`.

plugin/core/views.py
```diff
@@ -176,20 +176,67 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
...
+            content = value
+        if language:
+            # MarkedString (dict)
+            is_plain_text = False
+            content = "```{}\n{}\n```\n".format(language, value)
+    if is_plain_text:
+        return text2html(content)
     else:
-        return ''
+        frontmatter_config = mdpopups.format_frontmatter({'allow_code_wrap': True})
```
rchl, line 238: We can revert if it proves problematic but we are now always setting `allow_code_wrap` for markup content.

plugin/core/views.py
```diff
@@ -176,20 +176,68 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
     })
 
 
-def minihtml(view: sublime.View, content: Union[str, dict]) -> str:
-    """ Content can be a string or MarkupContent. """
+def minihtml(view: sublime.View, content: Union[str, dict, list], prefer_plain_text: bool) -> str:
```
predragnikolic, line 179: What do you think that instead of `prefer_plain_text: bool` we have `marked_string: bool = False`? 

I know that I could have suggested `plain_text: bool = True` to be the default, 
but I think changing it to `marked_string` makes it more explicit.

There is only one call for now where marked_string would be set to True:
```
return minihtml(self.view, self._hover.get('contents') or "", marked_string=True)
```
the rest of the calls we will just be:
```diff
-   detail = minihtml(self.view, item.get('detail') or "", prefer_plain_text=True)
+  detail = minihtml(self.view, item.get('detail') or "")

rchl: Wfm

rwols: I agree the `prefer_plain_text` is a bit vague. `marked_string` communicates better what is happening.

rchl: To be fair I don't like either current or suggested solution too much. Latter also seems to suggest that you are passing MarkedString while it could be MarkupContent.

I'll try to do something with bitwase flags. I think that would be the most explicit and truthful solution.

rchl: Done now.

PR #1049, opened 2020-05-14
Dotted server settings

rwols:
Both styles work: you can put a dict in the "settings" of a server,
or you can put "paths" in the "settings" of a server.

- [x] deepcopy the existing DottedDict in configurations.py

### DIFF
--- a/plugin/core/collections.py
+++ b/plugin/core/collections.py
@@ -6,7 +6,7 @@ from .typing import Optional, Dict, Any
 
 class DottedDict:
 
-    __slots__ = ('_d')
+    __slots__ = ('_d',)
 
     def __init__(self, d: Optional[Dict[str, Any]] = None) -> None:
         """
--- a/plugin/core/collections.py
+++ b/plugin/core/collections.py
@@ -1,0 +1,115 @@
+"""
+Module with additional collections.
+"""
+from .typing import Optional, Dict, Any
+
+
+class DottedDict:
+
+    __slots__ = ('_d')
+
+    def __init__(self, d: Optional[Dict[str, Any]] = None) -> None:
+        """
+        Construct a DottedDict, optionally from an existing dictionary.
+
+        :param      d:    An existing dictionary.
+        """
+        self._d = {}  # type: Dict[str, Any]
+        if d is not None:
+            self.update(d)
+
+    def get(self, path: Optional[str] = None) -> Any:
+        """
+        Get a value from the dictionary.
+
+        :param      path:  The path, e.g. foo.bar.baz, or None.
+
+        :returns:   The value stored at the path, or None if it doesn't exist.
+                    Note that this cannot distinguish between None values and
+                    paths that don't exist. If the path is None, returns the
+                    entire dictionary.
+        """
+        if path is None:
+            return self._d
+        current = self._d  # type: Any
+        keys = path.split('.')
+        for key in keys:
+            if isinstance(current, dict):
+                current = current.get(key)
+            else:
+                return None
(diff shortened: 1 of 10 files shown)


plugin/core/collections.py
```diff
@@ -0,0 +1,115 @@
+"""
+Module with additional collections.
+"""
+from .typing import Optional, Dict, Any
+
+
+class DottedDict:
+
+    __slots__ = ('_d')
```
rchl, line 9: From https://stackoverflow.com/questions/472000/usage-of-slots:
> Requirements:
>  - To have attributes named in `__slots__` to actually be stored in slots instead of a `__dict__`, a class must inherit from object.
`

plugin/core/collections.py
```diff
@@ -0,0 +1,115 @@
...
+    def set(self, path: str, value: Any) -> None:
+        """
+        Set a value in the dictionary.
+
+        :param      path:   The path, e.g. foo.bar.baz
+        :param      value:  The value
+        """
+        current = self._d
+        keys = path.split('.')
+        for i in range(0, len(keys) - 1):
```
rchl, line 52: Isn't this the same while not looking like it's gonna make the brain explode:
```py
for i in range(len(keys)):
```
?

Or even shorter and cleaner:
```py
for i, key in enumerate(keys):
```

But then it doesn't look like you need index here so maybe just:
```py
for key in keys:
```
?

plugin/core/configurations.py
```diff
@@ -86,32 +86,22 @@ def _create_window_configs(self) -> List[ClientConfig]:
     def _apply_project_overrides(self, client_config: ClientConfig, project_clients: Dict[str, Any]) -> ClientConfig:
         overrides = project_clients.get(client_config.name)
         if overrides:
-            debug('window has override for {}'.format(client_config.name), overrides)
-            client_settings = _merge_dicts(client_config.settings, overrides.get("settings", {}))
-            client_env = _merge_dicts(client_config.env, overrides.get("env", {}))
+            debug('applying .sublime-project override for', client_config.name)
+            client_config.settings.update(overrides.get("settings", {}))
+            for key, value in overrides.get("env", {}).items():
+                client_config.env[key] = value
```
rchl, line 92: Why not using `update` also for `env`?

plugin/core/collections.py
rwols: I think that's old info. Every class implicitly inherits from `object` in python3. But there is a bug on that line... It should be `('_d',)`.

Without `__slots__`:
```
>>> from LSP.plugin.core.collections import DottedDict
>>> from sys import getsizeof
>>> getsizeof(DottedDict())
48
```

With `__slots__`:
```
>>> from LSP.plugin.core.collections import DottedDict
>>> from sys import getsizeof
>>> getsizeof(DottedDict())
40
```
What matters to me is the type-safety though.

plugin/core/configurations.py
rwols: Because `env` is a `dict`, whereas `settings` is a `DottedDict`.

plugin/core/collections.py
rwols: The last chunk of the path gets special treatment (`current[keys[-1]] = value`). For the earlier chunks: we want to turn the thing the chunk points to into a dict if it isn't a dict already:

```
            if not isinstance(next_current, dict):
                next_current = {}
                current[key] = next_current
```
so that we can traverse deeper into `self._d`.

rchl: Right. Sorry.

PR #1055, opened 2020-05-18
Normalize completion kinds and symbol kinds

rwols:
They live in views.py now as lists.
Let's also use smallcase latin letters for icons as that's what people are
converging towards in the Default packages.

KIND_ID_NAVIGATION is used for "file" and "reference".
KIND_ID_TYPE is used for "event".
KIND_ID_VARIABLE is used for "enum member".

@predragnikolic, would it be an idea to allow customization of the icons through a dict in LSP.sublime-settings?


rchl:
Kinda tricky to review but I'll run with it for a day.
Any particular things to be on a lookout for?

rwols:
It's purely style. We should not be too opinionated about the icons and just follow what the standard is. But I can see how users would want to customize this. That's why I propose to put a dict in the LSP.sublime-settings to allow users to customize the look.

rchl:
> That's why I propose to put a dict in the LSP.sublime-settings to allow users to customize the look.

Personally I would probably never be bothered to customize those.
I'd say we should wait with introducing settings until someone actually asks.
It's after all more work to maintain settings than not having them.

predragnikolic:
@rwols I also see that some user would like to change the icons, on the other side I also agree with @rchl about maintance. I let you be the judge. I am fine with both decisions.

PR #1056, opened 2020-05-18
Gather usages of encoded filename and put it in a common function

rwols:
The workspace/symbol request actually did this wrong.

Also put the code for the workspace/symbol in symbols.py as textDocument/symbol and workspace/symbol share a lot of code.


plugin/goto.py
```diff
@@ -30,71 +54,26 @@ def run(self, edit: sublime.Edit, event: Optional[dict] = None) -> None:
...
-                locations: List[Tuple[str, str, Tuple[int, int]]],
-                idx: int,
-                orig_view: sublime.View) -> None:
-            if idx >= 0:
-                open_location(window, locations[idx])
-            else:
-                if orig_view:
-                    window.focus_view(orig_view)
-
-        def highlight_entry(
```
rwols, line 72: These three closures actually did not capture any variables outside the function, so might as well make them plain functions.

plugin/symbols.py
```diff
@@ -156,3 +159,61 @@ def process_symbol_informations(self, items: List[Dict[str, Any]]) -> List[List[
...
+    def validate(self, txt: str) -> bool:
+        return txt != ""
+
+    def placeholder(self) -> str:
+        return "Symbol"
+
+
+class LspWorkspaceSymbolsCommand(LspTextCommand):
+    def __init__(self, view: sublime.View) -> None:
+        super().__init__(view)
```
rchl, line 175: Should skip overriding if you are not doing anything extra.
Or is it some style choice?

plugin/symbols.py
```diff
@@ -156,3 +159,61 @@ def process_symbol_informations(self, items: List[Dict[str, Any]]) -> List[List[
...
+    def input(self, _args: Any) -> sublime_plugin.TextInputHandler:
+        return SymbolQueryInput()
+
+    def run(self, edit: sublime.Edit, symbol_query_input: str = "") -> None:
+        if symbol_query_input:
+            request = Request.workspaceSymbol({"query": symbol_query_input})
+            client = self.client_with_capability('workspaceSymbolProvider')
+            if client:
+                self.view.set_status("lsp_workspace_symbols", "Searching for '{}'...".format(symbol_query_input))
+                client.send_request(request, lambda r: self._handle_response(symbol_query_input, r), self._handle_error)
```
rchl, line 219: Style nit: Could create a request just before sending it.

plugin/symbols.py
```diff
@@ -156,3 +159,61 @@ def process_symbol_informations(self, items: List[Dict[str, Any]]) -> List[List[
...
+        return name
+
+    def _open_file(self, symbols: List[Dict[str, Any]], index: int) -> None:
+        if index != -1:
+            symbol = symbols[index]
+            window = self.view.window()
+            if window:
+                window.open_file(location_to_encoded_filename(symbol['location']), sublime.ENCODED_POSITION)
+
+    def _handle_response(self, query: str, response: Optional[List[Dict[str, Any]]]) -> None:
```
rchl, line 190: Me being boring again: would love those to be ordered better (moved after `run`).

plugin/core/views.py
```diff
@@ -48,6 +49,21 @@ def region_to_range(view: sublime.View, region: sublime.Region) -> Range:
     )
 
 
+def encoded_filename(uri: str, position: Dict[str, Any]) -> str:
+    # WARNING: Cannot possibly do UTF-16 conversion :) Oh well.
+    return '{}:{}:{}'.format(uri_to_filename(uri), position['line'] + 1, position['character'] + 1)
```
rchl, line 54: Style nit: Since it's not reused and used only in one place maybe just inline it?

plugin/symbols.py
rwols: Done

PR #1058, opened 2020-05-20
Expand variables for workspace/executeCommand request arguments

jwortmann:
As it is now, the `lsp_execute` command is not really usefull if the request arguments require a file URI or other things dependent on current state. This PR expands (actually replaces) the following string arguments:

* `"${file_uri}"` for the file URI of the active view
* `"${offset}"` for the character offset of the (topmost) cursor position
* `"${selection_begin}"` for the character offset of the begin of the (topmost) selection
* `"${selection_end}"` for the character offset of the end of the (topmost) selection
* `"${selection}"` for the the content of the (topmost) selection
* `"${position}"` for Position object (from LSP specs) of the (topmost) cursor position
* `"${range}"` for Range object (from LSP specs) of the (topmost) selection 

In general the `arguments` parameter for the request is an array with items of any kind, but it should probably be sufficient to only check the top-level items.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -41,7 +41,7 @@ Example:
     "command": "lsp_execute",
     "args": { 
       "command_name": "thread-first",
-      "command_args": ["${file}", 0, 0]
+      "command_args": ["${file_uri}", 0, 0]
     }
   }
 ]
@@ -51,11 +51,13 @@ The following variables will be expanded, but only if they are top-level array i
 
 | Variable | Type | Description |
 | -------- | ---- | ----------- |
-| `"${file}"` | string | File URI of the active view |
+| `"${file_uri}"` | string | File URI of the active view |
+| `"${selection}"` | string | Content of the (rearmost) selection |
 | `"${offset}"` | int | Character offset of the (rearmost) cursor position |
 | `"${selection_begin}"` | int | Character offset of the begin of the (rearmost) selection |
 | `"${selection_end}"` | int | Character offset of the end of the (rearmost) selection |
-| `"${selection}"` | string | Content of the (rearmost) selection |
+| `"${position}"` | object | Mapping `{ 'line': int, 'character': int }` of the (rearmost) cursor position, see [Position](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#position) |
+| `"${range}` | object | Mapping with `'start'` and `'end'` positions of the (rearmost) selection, see [Range](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#range) |
 
 **Overriding keybindings**
 
--- a/docs/features.md
+++ b/docs/features.md
@@ -51,13 +51,13 @@ The following variables will be expanded, but only if they are top-level array i
 
 | Variable | Type | Description |
 | -------- | ---- | ----------- |
-| `"${file_uri}"` | string | File URI of the active view |
-| `"${selection}"` | string | Content of the (rearmost) selection |
-| `"${offset}"` | int | Character offset of the (rearmost) cursor position |
-| `"${selection_begin}"` | int | Character offset of the begin of the (rearmost) selection |
-| `"${selection_end}"` | int | Character offset of the end of the (rearmost) selection |
-| `"${position}"` | object | Mapping `{ 'line': int, 'character': int }` of the (rearmost) cursor position, see [Position](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#position) |
-| `"${range}` | object | Mapping with `'start'` and `'end'` positions of the (rearmost) selection, see [Range](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#range) |
+| `"$file_uri"` or `"${file_uri}"` | string | File URI of the active view |
+| `"$selection"` or `"${selection}"` | string | Content of the (rearmost) selection |
+| `"$offset"` or `"${offset}"` | int | Character offset of the (rearmost) cursor position |
+| `"$selection_begin"` or `"${selection_begin}"` | int | Character offset of the begin of the (rearmost) selection |
+| `"$selection_end"` or `"${selection_end}"` | int | Character offset of the end of the (rearmost) selection |
+| `"$position"` or `"${position}"` | object | Mapping `{ 'line': int, 'character': int }` of the (rearmost) cursor position, see [Position](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#position) |
+| `"$range"` or `"${range}` | object | Mapping with `'start'` and `'end'` positions of the (rearmost) selection, see [Range](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#range) |
 
 **Overriding keybindings**
 
--- a/docs/features.md
+++ b/docs/features.md
@@ -52,12 +52,12 @@ The following variables will be expanded, but only if they are top-level array i
 | Variable | Type | Description |
 | -------- | ---- | ----------- |
 | `"$file_uri"` or `"${file_uri}"` | string | File URI of the active view |
-| `"$selection"` or `"${selection}"` | string | Content of the (rearmost) selection |
-| `"$offset"` or `"${offset}"` | int | Character offset of the (rearmost) cursor position |
-| `"$selection_begin"` or `"${selection_begin}"` | int | Character offset of the begin of the (rearmost) selection |
-| `"$selection_end"` or `"${selection_end}"` | int | Character offset of the end of the (rearmost) selection |
-| `"$position"` or `"${position}"` | object | Mapping `{ 'line': int, 'character': int }` of the (rearmost) cursor position, see [Position](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#position) |
-| `"$range"` or `"${range}` | object | Mapping with `'start'` and `'end'` positions of the (rearmost) selection, see [Range](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#range) |
+| `"$selection"` or `"${selection}"` | string | Content of the (topmost) selection |
+| `"$offset"` or `"${offset}"` | int | Character offset of the (topmost) cursor position |
+| `"$selection_begin"` or `"${selection_begin}"` | int | Character offset of the begin of the (topmost) selection |
+| `"$selection_end"` or `"${selection_end}"` | int | Character offset of the end of the (topmost) selection |
+| `"$position"` or `"${position}"` | object | Mapping `{ 'line': int, 'character': int }` of the (topmost) cursor position, see [Position](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#position) |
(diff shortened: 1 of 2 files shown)


plugin/execute_command.py
```diff
@@ -12,13 +13,27 @@ def __init__(self, view: sublime.View) -> None:
...
         if client and command_name:
             window = self.view.window()
             if window:
                 window.status_message("Running command {}".format(command_name))
-            self._send_command(client, command_name, command_args)
+            self._send_command(client, command_name, self._expand_variables(command_args))
+
+    def _expand_variables(self, command_args: List[Any]) -> List[Any]:
+        for i, arg in enumerate(command_args):
+            if arg == "${file}":
```
rwols, line 26: It would be much nicer/easier to use `sublime.expand_variables` here. Something like this?

```python
variables = self.view.window().extract_variables()
region = self.view.sel()[-1]
variables['uri'] = uri_from_view(self.view)
variables['offset'] = region.b
variables['selection_begin'] = region.begin()
variables['selection_end'] = region.end()
variables['selection'] = self.view.substr(region)
return sublime.expand_variables(command_args, variables)
```

There are also already some concepts like Position and Range that should be included. `$position` and `$range`?

The `offset`, `selection_begin` and `selection_end` are UTF-32 offset-based, I don't think this will mesh well with the protocol, which generally uses a (row, col_utf16) pair to denote positions in a document. What server are you using?

jwortmann: Oh, I thought the mapping for `expand_variables()` must contain predefined strings. I will try your suggestion, which seems more elegant.

> The offset, selection_begin and selection_end are UTF-32 offset-based

I indeed saw that for the example `a𐐀b` in the specs, this implementation gives the wrong offset value. However, I guess it is impossible to solve with the current ST API. I quickly experimented with something like
```python
from .core.protocol import Point
from .core.views import point_to_offset

point_to_offset(Point(*self.view.rowcol(self.view.sel()[-1].b)), self.view)
```
but this seems useless since it only converts back and forth. Do you know a solution?

I use the Julia server, which expects a single int offset: https://github.com/julia-vscode/LanguageServer.jl/blob/cdb566d369389a43c1cdf88f3d7d1e69127dd73e/src/requests/actions.jl#L38

rwols: The question becomes whether Julia server then expects UTF-32 or UTF-16 offset. But communicating with UTF-16 offsets is madness, I suppose the only sane choice is UTF-32.

On the other hand, it *should* ask for a JSON dict `{"line": x, "character": y}` which is standard LSP.

> However, I guess it is impossible to solve with the current ST API. 

It is solved in the st4000-exploration branch with the new API.

jwortmann: I could not get `sublime.expand_variables()` to work like that, as it throws an error if the values for `variables` are not strings. Seems like this function is only for simple string replacement, and doesn't work with numbers, objects or evaluated functions.

I added `"${position}"` and `"${range}"` from the LSP specs.

> It is solved in the st4000-exploration branch with the new API.

I think it should work for most cases now, and ST3 users could at least profit from this implementation. Maybe it could be adjusted for the ST4 branch later then.

jwortmann: Also I noticed that we should not send `"arguments": null` if no arguments are provided.

rwols: Hmm, it's working here for me. Most likely SublimeHQ added this functionality for ST4.

```
>>> sublime.expand_variables(['$folder'], window.extract_variables())
['/home/raoul/.config/sublime-text-3/Packages/LSP']
>>> sublime.expand_variables({'foo': '$platform'}, window.extract_variables())
{'foo': 'Linux'}
```

jwortmann: Yes, but these two examples are both strings. You cannot add you own objects to `variables` and cannot evaluate functions for it.

jwortmann:
```
>>> sublime.expand_variables('$test', {'test': 42})
  File "<string>", line 1, in <module>
  File "C:\Program Files\Sublime Text 3\sublime.py", line 221, in expand_variables
    return sublime_api.expand_variables(val, variables)
TypeError: String required
```
And at least `view.substr()` didn't work for me when used as a value in `variables`.

rwols: Thanks for explaining the problem. You're correct in that we can't use sublime.expand_variables for this.

But this makes me question whether using this string-template notation is a good idea. The current impl does not recognize e.g. that `$offset` and `${offset}` are the same thing. Can we instead use something like a dict to denote the type/kind of argument? For instance:

```
[{"kind": "position"}, {"kind": "selection"}, {"kind": "constant", "value": 5}]
```
would define a command signature with three arguments, the first arg is the position, the second arg is the selection, and the third arg is the constant integer 5.

rwols: If you want to go far with this, you could also have a `{"kind": "input-text"}` that would show a TextInputHandler for the user to type in some text, and a `{"kind": "input-list", "value": ["option1, "option2"]}` that would show a ListInputHandler with the choices option1 and option2.

jwortmann: > The current impl does not recognize e.g. that $offset and ${offset} are the same thing.

No big problem, I could use `if arg in ['$offset', '${offset}']` then. The only thing that wouldn't work with this implementation is an argument like `'some text ${file_uri} more text'`, but this doesn't make much sense anyway, as most variables are replaced by an int or dict and I don't think there are many use cases for something like this example.

The third argument in
```
[{"kind": "position"}, {"kind": "selection"}, {"kind": "constant", "value": 5}]
```
doesn't make much sense, as it's the same as
```
[{"kind": "position"}, {"kind": "selection"}, 5]
```
but actually in my opinion this notation would only make it more complicated for users, and the `$offset` or `${offset}` notation is already a standard in many Sublime config files. If we want, we could additionally use `sublime.expand_variables()` for the other standard window variables like `$folder`, but I'm not sure if there are any use cases for these as well. If I should change the notation to `{"kind": "...."}`, I at least would like to see a third opinion who prefers that too.

jwortmann: I like the `input-text` and `-list` idea (preferably `"$input_text"`/`"${input_text}"` 😉 ) but there might be some uncertainties with how it should work:
* for multiple arguments like these, just show multiple quick panels?
* should the input be treated as a string - that would make it impossible to input numbers

Maybe we should better leave this for another pull request.

plugin/execute_command.py
```diff
@@ -12,14 +13,41 @@ def __init__(self, view: sublime.View) -> None:
...
+            elif arg in ["$position", "${position}"]:
+                position = self.view.rowcol(region.b)
+                command_args[i] = {"line": position[0], "character": position[1]}
+            elif arg in ["$range", "${range}"]:
+                start = self.view.rowcol(region.begin())
+                end = self.view.rowcol(region.end())
+                command_args[i] = {
+                    "start": {"line": start[0], "character": start[1]},
+                    "end": {"line": end[0], "character": end[1]}
+                }
```
rwols, line 48: I'm missing an `else` case for constant/unkown values (I need this for clangd :) )

plugin/execute_command.py
```diff
@@ -12,14 +13,41 @@ def __init__(self, view: sublime.View) -> None:
...
+                command_args[i] = self.view.substr(region)
+            elif arg in ["$offset", "${offset}"]:
+                command_args[i] = region.b
+            elif arg in ["$selection_begin", "${selection_begin}"]:
+                command_args[i] = region.begin()
+            elif arg in ["$selection_end", "${selection_end}"]:
+                command_args[i] = region.end()
+            elif arg in ["$position", "${position}"]:
+                position = self.view.rowcol(region.b)
+                command_args[i] = {"line": position[0], "character": position[1]}
```
rwols, line 41: Please use `offset_to_point(region.b).to_lsp()` from plugin/core/views.py, this function does UTF-16 correctly (for ST4).

plugin/execute_command.py
```diff
@@ -12,14 +13,41 @@ def __init__(self, view: sublime.View) -> None:
...
+            elif arg in ["$range", "${range}"]:
+                start = self.view.rowcol(region.begin())
+                end = self.view.rowcol(region.end())
+                command_args[i] = {
+                    "start": {"line": start[0], "character": start[1]},
+                    "end": {"line": end[0], "character": end[1]}
```
rwols, line 47: Please use `region_to_range(region).to_lsp()` from plugin/core/views.py, this function does UTF-16 correctly (for ST4).

docs/features.md
```diff
@@ -40,13 +40,24 @@ Example:
...
 ]
 ```
-Note: `command_args` is optional depending on the `workspace/executeCommand` that are supported by the LSP server. 
+Note: `command_args` is optional depending on the `workspace/executeCommand` that are supported by the LSP server.
+The following variables will be expanded, but only if they are top-level array items and not within nested arrays or objects:
+
+| Variable | Type | Description |
+| -------- | ---- | ----------- |
+| `"$file_uri"` or `"${file_uri}"` | string | File URI of the active view |
+| `"$selection"` or `"${selection}"` | string | Content of the (rearmost) selection |
```
rwols, line 55: Rearmost: usually we reason from the first selection instead of the last selection (e.g. completions take the first cursor position as the point to request completions for). Why do we have to take the last selection/cursor for commands?

jwortmann: From a usage perspective, I would argue that the rearmost is the better choice. When I use multiple cursors/selections, I usually add them from left to right and from top to bottom of a file. Then, the rearmost is the last added or "active" one, and when I suddenly see that I want to run the workspace/executeCommand for that selection, I don't have to reset all other selections first to do it.

However if we use the first selection in all other cases, I will change it here to the first as well for consistency. I think to always use the last cursor/selection would be the better way though.

plugin/execute_command.py
jwortmann: I don't really understand this, what would you put under `else`, `command_args[i] = command_args[i]`? This `_expand_variables` method only alters the list items with these specific variables in place. That said, I could omit the return statement / return None, but I don't think this function is performance relevant and didn't quite remember whether Python treats method arguments by copy or by reference.

docs/features.md
rwols: The AST is in most cases wrong when using the last cursor position for completions. For commands this probably doesn't matter as the text bufer is most likely in a good state (no errors). For consistency's sake I still think we should take the first region/cursor of the selection to avoid surprises for users.

plugin/execute_command.py
rwols: Oh I see :) It's a bit confusing that we're altering the list, but also returning the list. Perhaps return `None`? This communicates a bit better that we're altering the list.

plugin/execute_command.py
jwortmann: Thanks. By the way, the argument order in the functions `point_to_offset` and `range_to_region` could be made consistent with all the other functions in `views.py`.

docs/features.md
jwortmann: Yes I agree, changed it to the first.

plugin/execute_command.py
jwortmann: Ok.

PR #1059, opened 2020-05-20
Use LSPViewEventListener.has_supported_syntax consistently

rchl:
Instead of using the free-standing "is_supported_syntax" function.

(This doesn't fix the performance issue - just cleans up the code to make the fix easier.)

### DIFF
--- a/plugin/core/documents.py
+++ b/plugin/core/documents.py
@@ -1,8 +1,6 @@
 import sublime
 
-from .configurations import is_supported_syntax
 from .registry import LSPViewEventListener
-from .settings import client_configs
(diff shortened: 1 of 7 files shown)


rchl:
github CI not in the mood?

plugin/core/documents.py
```diff
@@ -1,6 +1,5 @@
 import sublime
 
-from .configurations import is_supported_syntax
 from .registry import LSPViewEventListener
 from .settings import client_configs
```
rwols, line 4: The CI tells me this import is unused.

PR #1062, opened 2020-05-21
Fix applying code action when content preceded by utf16 char (emoji)

rchl:
I've noticed that eslint goes crazy and breaks the code on CAOS when there are emojis. This fixes it.

### DIFF
--- a/tests/test_code_actions.py
+++ b/tests/test_code_actions.py
@@ -53,3 +53,13 @@ class CodeActionsTestCase(TextDocumentTestCase):
         ])
         run_code_action_or_command(self.view, self.config.name, code_actions)
         self.assertEquals(entire_content(self.view), initial_content)
+
+    # Keep this test last as it breaks pyls!
+    def test_applies_correctly_after_emoji(self) -> Generator:
+        self.insert_characters('🕵️hi')
+        yield from self.await_message("textDocument/didChange")
+        code_action = create_test_code_actions(self.view.change_count(), [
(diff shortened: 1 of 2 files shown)


rchl:
There is a couple more non-utf16 `rowcol`'s and one `text_point` left in the code. We should investigate if those need to be changed too (in a separate PR).

tests/test_code_actions.py
```diff
@@ -53,3 +53,13 @@ def test_does_not_apply_with_nonmatching_document_version(self) -> Generator:
         ])
         run_code_action_or_command(self.view, self.config.name, code_actions)
         self.assertEquals(entire_content(self.view), initial_content)
+
+    # Keep this test last as it breaks pyls!
+    def test_applies_correctly_after_emoji(self) -> Generator:
+        self.insert_characters('🕵️hi')
+        yield from self.await_message("textDocument/didChange")
+        code_action = create_test_code_actions(self.view.change_count(), [
+            ("bye", Range(Point(0, 3), Point(0, 5))),
```
rchl, line 62: It should be correct as it works for eslint but I don't quite understand this myself as ST shows that emoji character takes two characters and yet I have to start from 4th (1-based) for the server-returned code action...
(probably just me not understanding UTF though)

rwols: That's probably because this is "Detective | Man Modifier", if you paste in a "Detective" (without the "Man Modifier"), then it should be two UTF-16 points. The modifier probably adds more UTF-16 points.

PR #1064, opened 2020-05-21
Clean up regions and statuses of views on shutdown

rwols:


plugin/core/main.py
```diff
@@ -25,25 +25,26 @@ def startup() -> None:
...
+    if window:
+        windows.lookup(window).start_active_views()
 
 
 def shutdown() -> None:
     # Also needs to handle package being disabled or removed
     # https://github.com/sublimelsp/LSP/issues/375
     unload_settings()
-
+    # TODO: Move to __del__ methods
```
rchl, line 37: I wonder if it will be too late to call the ST API from __del__. Maybe we'll have to manually call our own "del" from here?

rwols: ViewEventListeners have the guarantee that their `__del__` is called when the view is closed, and their `__init__` is called when a view is opened. We basically have to design the entire architecture around these facts.

rchl: I was thinking about the case when package is disabled instead. In that case is `__del__` called at all?

And on closing view, do we have the guarantee that `view.settings()` or APIs like `view. erase_regions ` are not gonna fail from `__del__`?

PR #1065, opened 2020-05-21
Fix timeouts in tests

rchl:
Must reset cached object on removing/adding configs.


rchl:
BTW. `test_session.py` is completely broken as it tries to import things that don't exist (`set_dotted_value` and family).

rchl:
Hmm, maybe tests still timeout... I'd swear it worked...

rchl:
OK, fixed test_session.py issue that caused timeouts. This is something you've also fixed on `transports` branch (at least minimal version of it). Now all tests pass for me.

PR #1066, opened 2020-05-22
Hopefully fix AC disappearing

rwols:
Showing and updating popups from ST's main thread seems to fix the issue.

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -22,7 +22,8 @@ class LspResolveDocsCommand(sublime_plugin.TextCommand):
         # don't show the detail in the cooperate AC popup if it is already shown in the AC details filed.
         self.is_detail_shown = bool(detail)
         minihtml_content = self.get_content(documentation, detail)
-        self.show_popup(minihtml_content)
+        # NOTE: For some reason, ST does not like it when we show a popup from within this run method.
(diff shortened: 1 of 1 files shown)


plugin/completion.py
```diff
@@ -67,10 +67,9 @@ def handle_resolve_response(self, item: Optional[dict]) -> None:
         detail = self.format_documentation(item.get('detail') or "")
         documentation = self.format_documentation(item.get("documentation") or "")
         minihtml_content = self.get_content(documentation, detail)
-        if self.view.is_popup_visible():
-            self.update_popup(minihtml_content)
-        else:
-            self.show_popup(minihtml_content)
+        show = self.update_popup if self.view.is_popup_visible() else self.show_popup
+        # NOTE: Update/show popups from the main thread, or else the popup might make the AC widget disappear.
+        sublime.set_timeout(lambda: show(minihtml_content))
```
rwols, line 72: Hmm, this doesn't fix the issue for `pyls`'s json module documentation, but it does fix it solargraph (I'm guessing any server that requires a `completionItem/resolve`.

rwols:
1566ec5 also fixes the issue for already-resolved docs (e.g. pyls).

For some reason ST does not like it when we show a popup from within the `run` method of `LspResolveDocs`.

PR #1067, opened 2020-05-22
Optionally run a server command after inserting a completion

rwols:
I haven't tested this, but it seems straightforward. If anything comes up from a user we can improve upon it.

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -78,11 +78,14 @@ class LspResolveDocsCommand(sublime_plugin.TextCommand):
 
 class LspCompleteCommand(sublime_plugin.TextCommand):
 
-    def handle_additional_edits(self, item: Dict[str, Any]) -> None:
+    def epilogue(self, item: Dict[str, Any]) -> None:
         additional_edits = item.get('additionalTextEdits')
         if additional_edits:
             edits = [parse_text_edit(additional_edit) for additional_edit in additional_edits]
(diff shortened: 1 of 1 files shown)


plugin/completion.py
```diff
@@ -244,8 +247,8 @@ def format_completion(self, item: dict, index: int, can_resolve_completion_items
                 kind=kind,
                 details=st_details)
             completion.flags = sublime.COMPLETION_FLAG_KEEP_PREFIX
-        elif item.get("additionalTextEdits"):
-            # It's an insertText, but additionalEdits requires us to use a command completion.
+        elif item.get("additionalTextEdits") or item.get("command"):
+            # It's an insertText, but additionalEdits or a command requires us to use a command completion.
```
rchl, line 251: The logic flow is a bit weird here but I see that it should work in practice.

plugin/completion.py
```diff
@@ -78,11 +78,14 @@ def update_popup(self, minihtml_content: str) -> None:
 
 class LspCompleteCommand(sublime_plugin.TextCommand):
 
-    def handle_additional_edits(self, item: Dict[str, Any]) -> None:
+    def epilogue(self, item: Dict[str, Any]) -> None:
```
predragnikolic, line 81: Nitpic 🙂 maybe rename to on_completion_inserted or something like that? 

Epilogue reminds me a lot on  the word you offend use: vague

Issue #869, opened 2020-02-04
Auto close server session on last view close


vseryakov:
Hi,

This little change may be too big for a pull request. 

Working with different languages i do not want to have the dart analyst running and eating 300M+ of my memory, so it wil close the session if all dart views are closed. 

```
diff --git a/plugin/core/windows.py b/plugin/core/windows.py
index ca3f293..623e77f 100644
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -242,6 +242,8 @@ class WindowDocumentHandler(object):
             if session.client:
                 params = {"textDocument": {"uri": filename_to_uri(file_name)}}
                 session.client.send_notification(Notification.didClose(params))
+           if not len(self._document_states):
+              session.end()

     def handle_view_saved(self, view: ViewLike) -> None:
         file_name = view.file_name()
```

rwols:
This has been in the back of my mind for a while. It has both advantages and disadvantages. One major advantage is that we don't have to care about project switching anymore, since all old views from the previous project will be closed after switching a project. A disadvantage is that we might be starting+initializing a server quite a few times when the user opens and closes single views in a window.

vseryakov:
right, that was a proof of concept and it works very well. i think i will extend it with a config option so it will be flexible, will do a proper pull request this time, tnx

rwols:
I don't believe this is worth a configuration setting. It would add yet another point of maintenance.

vseryakov:
ok, another solution, only close after 30 secs, it will fire multiple timers on every close though but i do not see how to cancel sublime timer, i guess using Python timer

```
--- windows.py.orig	2020-02-04 18:04:08.000000000 -0500
+++ windows.py	2020-02-04 18:05:40.000000000 -0500
@@ -237,11 +237,23 @@
             self._document_states.remove(file_name)
         except ValueError:
             return
+        configs = []
         for session in self._get_applicable_sessions(view, 'openClose'):
             debug('closing', file_name, session.config.name)
             if session.client:
                 params = {"textDocument": {"uri": filename_to_uri(file_name)}}
                 session.client.send_notification(Notification.didClose(params))
+                configs.append(session.config.name)
+        if not len(self._document_states):
+           self._sublime.set_timeout(lambda: self._handle_view_closed(configs), 30000)
+
+    def _handle_view_closed(self, configs: 'Any') -> None:
+        if not len(self._document_states):
+           debug('closing sessions for', configs)
[...]

Issue #765, opened 2019-11-07
Provide a context key to check if LSP has a specific capability


randy3k:
<del>
Possible use case of the context key is to define custom keybind for go to definitions.
With the context key, it allows me to bind `super+alt+down` to both `goto_definition` and `lsp_symbol_definition`.

```js
{ 
   // this is the default key for `goto_definition`
  "keys": ["super+alt+down"], "command": "goto_definition" 
},
{ 
  "keys": ["super+alt+down"], "command": "lsp_symbol_definition" ,
  "context": [
			{ "key": "lsp.is.active", "operator": "equal", "operand": true }
		]
}
```
</del>

Update: @rwols pointed out that there is `lsp_active` to check, so this issue is now requesting a key for checking more specific capability.

randy3k:
I got my hand dirty and made this

```
import sublime_plugin
from LSP.plugin.core.registry import session_for_view


class LspCapability(sublime_plugin.EventListener):
    def on_query_context(self, view, key, operator, operand, match_all):
        if key == "lsp.client_with_capability":
            return session_for_view(view, operand)
```

and now I could do
```
    {
      "keys": ["super+alt+down"], "command": "lsp_symbol_definition" ,
      "context": [
                { "key": "lsp.client_with_capability", "operator": "equal", "operand": "definitionProvider" }
            ]
    }
```

Any interests in getting this merged?

predragnikolic:
Pretty cool :-)

jfcherng:
https://lsp.readthedocs.io/en/latest/features/

![image](https://user-images.githubusercontent.com/6594915/68409208-a6625700-01c1-11ea-973d-9ea0a2329c58.png)

Isn't this already the default behavior?

randy3k:
In a non lsp document, `lsp_symbol_definition` is [disabled](https://github.com/tomv564/LSP/blob/master/plugin/goto.py#L26), hence `f12` is simply malfunction.  It only fallbacks to `goto_definition` when lsp is at work.

jfcherng:
You are right... https://github.com/tomv564/LSP/blob/be60102/plugin/goto.py#L28-L58
No wonder sometimes I get my F12 not working :open_mouth:

rwols:
Related: https://github.com/tomv564/LSP/issues/651, https://github.com/tomv564/LSP/commit/88fc8f9786569aab2a09a62f873af7e46aaef025

I think this already works with `lsp_active`, no?

[...]

PR #1069, opened 2020-05-22
Introduce the SessionView class

rwols:
It's an object whose lifetime is controlled by a ViewEventListener. It's an object per view per session. So it's essentially view-state per session. Each Session object holds a WeakSet to all of its SessionViews. Each SessionView holds a strong reference to its Session. The WindowManager holds weak references to Sessions. The WindowManager also holds weak references to the ViewEventListeners. This means the SessionView is the only object that holds a strong reference to a Session. Which means that if all views applicable to a Session are removed, then so is the associated Session.

For now, it does document sync but it should do a lot more.


deathaxe:
You should be careful with depending on `ViewEventListener` lifetimes. see: https://github.com/sublimehq/sublime_text/issues/2333

rchl:
I'm seeing this issue on this branch:

hover

(The ST's own "references" popup showing up before LSP's).

rwols:
Yes, that is due to not disabling `"show_definitions"` in the `view.settings()`, but I'll get to that eventually.

rchl:
Seems like on starting ST, servers are started for each file that is open in every project.
My session usually contains a couple of projects and each has many files open and it looks like for all those files servers are initialized on starting ST which makes things busy but from responsive and fan spin point of view.
Is it intended?

rwols:
Fans are spinning up here too, for a short while. The CPU utilization is about 12% for the plugin_host-3.3 executable during that time (which means one full core is running 100% for me).

I added debug printing for the time it took to run _needed_config as I think that's probably the function that takes a lot of time to compute, and here are some results:

```
LSP: _needed_config duration:  508 ms
LSP: _needed_config duration:  630 ms
LSP: _needed_config duration:  480 ms
LSP: _needed_config duration:  483 ms
LSP: _needed_config duration:  547 ms
LSP: _needed_config duration:  546 ms
LSP: _needed_config duration:  442 ms
LSP: _needed_config duration:  615 ms
LSP: _needed_config duration:  452 ms
LSP: _needed_config duration:  437 ms
LSP: _needed_config duration:  62 ms
LSP: _needed_config duration:  2 ms
LSP: _needed_config duration:  3 ms
LSP: _needed_config duration:  3 ms
LSP: _needed_config duration:  2 ms
LSP: _needed_config duration:  3 ms
LSP: _needed_config duration:  2 ms
LSP: _needed_config duration:  2 ms
LSP: _needed_config duration:  3 ms
LSP: _needed_config duration:  2 ms
LSP: _needed_config duration:  2 ms
LSP: _needed_config duration:  2 ms
LSP: _needed_config duration:  2 ms
LSP: _needed_config duration:  1 ms
LSP: _needed_config duration:  1 ms
LSP: _needed_config duration:  1 ms
LSP: _needed_config duration:  1 ms
LSP: _needed_config duration:  1 ms
LSP: _needed_config duration:  1 ms
LSP: _needed_config duration:  1 ms
LSP: _needed_config duration:  1 ms
```
When ST is starting up the functions takes 500 times longer to run for some reason. After a while it's fast enough to be unnoticable. I'm not sure why this happens.

rwols:
> Is it intended?

Yes, but I'll also experiment with "lazy initialization", so, not starting active views until the user actually opened / focused the view. Or we can do it the old way of processing views only for the active window.

rchl:
I would guess that ST is loading syntaxes and the APIs that _needed_config uses are blocked on that. Or something similar.

tests/setup.py
```diff
@@ -255,10 +256,7 @@ def insert_characters(self, characters: str) -> int:
...
         yield from self.await_boilerplate_end()
-        super().tearDown()
-
-    def doCleanups(self) -> 'Generator':
-        # restore the user's configs
         close_test_view(self.view)
+        # restore the user's configs
         remove_config(self.config)
-        yield from super().doCleanups()
+        super().tearDown()
```
rchl, line 262: `doCleanups()` no longer used - can I know the specific reasons for this change?

I don't exactly know the differences between those two but I think `doCleanups` is called when something throws while the other doesn't so that could make a difference with `failfast` disabled.

plugin/diagnostics.py
```diff
@@ -366,7 +368,7 @@ def format_diagnostic(self, diagnostic: Diagnostic) -> str:
 
 class DiagnosticsPresenter(object):
 
-    def __init__(self, window: sublime.Window, documents_state: DocumentsState) -> None:
```
rchl, line 369: `DocumentsState` class also be removed.

plugin/core/workspace.py
```diff
@@ -25,6 +31,9 @@ def includes_path(self, file_path: str) -> bool:
         else:
             return True
 
+    def __contains__(self, view: sublime.View) -> bool:
+        return self.includes_path(view.file_name() or '')
```
rchl, line 35: `os.path.realpath()` with an empty string will return the working directory. Do we really want to follow that path or just return `False` early then?

plugin/core/views.py
```diff
@@ -385,7 +396,8 @@ def minihtml(view: sublime.View, content: Union[str, Dict[str, str], list], allo
     if is_plain_text:
         return text2html(result)
     else:
-        frontmatter_config = mdpopups.format_frontmatter({'allow_code_wrap': True})
+        frontmatter_config = mdpopups.format_frontmatter(
+            {'allow_code_wrap': True})
```
rchl, line 400: The line wasn't that long. Why extra line break?

plugin/core/views.py
```diff
@@ -241,12 +251,12 @@ def will_save_wait_until(view: sublime.View, reason: int) -> Request:
...
 
-def did_save(view: sublime.View, include_text: bool) -> Notification:
-    return Notification.didSave(did_save_text_document_params(view, include_text))
+def did_save(view: sublime.View, include_text: bool, file_name: Optional[str] = None) -> Notification:
+    return Notification.didSave(did_save_text_document_params(view, include_text, file_name))
 
 
-def did_close(view: sublime.View) -> Notification:
-    return Notification.didClose(did_close_text_document_params(view))
+def did_close(file_name: str) -> Notification:
```
rchl, line 258: The ST API calls it "file_name" but it's actually a file path.
Maybe we should be in line with ST API but the latter makes more sense to me.

plugin/core/views.py
```diff
@@ -164,9 +168,9 @@ def entire_content_range(view: sublime.View) -> Range:
     return region_to_range(view, entire_content_region(view))
 
 
-def text_document_item(view: sublime.View, language_id: str) -> Dict[str, Any]:
+def text_document_item(view: sublime.View, language_id: str, file_name: Optional[str] = None) -> Dict[str, Any]:
     return {
-        "uri": uri_from_view(view),
+        "uri": filename_to_uri(file_name) if file_name is not None else uri_from_view(view),
```
rchl, line 173: Why do we accept empty `file_name` for the string case? `uri_from_view` does not do that.

rchl:
(not finished checking the changes yet)

rwols:
I'll summarize the changes:

1) A session ends when there are no more views associated to it after three seconds. So this resolves https://github.com/sublimelsp/LSP/issues/869
2) The install phase for a plugin is smoother (it runs in the async thread now)
3) This handles file renames correctly, at least with respect to document sync (all other view event listeners fall flat on their face)
4) This (should) handle cloned views correctly, at least with respect to document sync
5) Session capabilities are registered per view in the view's settings. So it'll be easily to resolve https://github.com/sublimelsp/LSP/issues/765 once we change the keybindings. In fact we should deprecate the lsp_active boolean after this.
6) It's a stepping stone to workDoneProgress and partialResult tokens which I want to work on after this.
7) This'll allow showing the active request in the status bar, which I also want to work on next for a more reactive UX.

plugin/core/configurations.py
```diff
@@ -44,13 +44,18 @@ def match_view(self, view: sublime.View, include_disabled: bool = False) -> Gene
...
+            configs = self.match_document(view2scope(view))
+            if include_disabled:
+                yield from configs
+            else:
+                for config in configs:
+                    if config.enabled:
+                        yield config
+        except IndexError:
+            # We're in the worker thread, and the view is already closed. This means view.scope_name(0) returns an
+            # empty string.
```
rchl, line 57: General note: it's not always obvious whether a method runs on the main or separate thread. Maybe we could annotate methods with some custom decorator that would assert main/non-main thread? It would make things clearer for us (me) and do a runtime check in case it's used in the way it's not supposed to.

plugin/core/sessions.py
```diff
@@ -429,16 +448,22 @@ class Session(Client):
...
     def __init__(self, manager: Manager, logger: Logger, workspace_folders: List[WorkspaceFolder],
                  config: ClientConfig, plugin_class: Optional[Type[AbstractPlugin]]) -> None:
+        super().__init__(logger)
         self.config = config
         self.manager = weakref.ref(manager)
         self.window = manager.window()
         self.state = ClientStates.STARTING
         self.capabilities = DottedDict()
+        self.exiting = False
+        self.views_opened = 0
```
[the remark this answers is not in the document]
rchl: Possibly relevant https://stackoverflow.com/questions/52693294/is-it-safe-to-give-a-python-weakset-to-a-list-constructor

plugin/core/sessions.py
```diff
@@ -509,16 +547,15 @@ def should_notify_did_change_configuration(self) -> bool:
         return self.has_capability("didChangeConfigurationProvider")
 
     def handles_path(self, file_path: Optional[str]) -> bool:
+        if self._supports_workspace_folders():
+            return True
```
rchl, line 551: So this allows running sessions on files outside the workspace... Sneaky sneaky. ;)

Although there is an issue. I've just opened an external `js` file in an LSP project and `LSP-eslint` went bananas - initializing and closing itself in an infinite loop.

plugin/core/sessions.py
```diff
@@ -621,24 +658,40 @@ def m_textDocument_publishDiagnostics(self, params: Any) -> None:
...
+                debug("{}: registering capability:".format(self.config.name), capability_path)
+                self.capabilities.set(capability_path, registration.get("registerOptions", {}))
+                self.capabilities.set(registration_path, registration["id"])
+                toplevel_key = capability_path.split('.')[0]
+                if toplevel_key.endswith('Provider'):
+                    for sv in self.session_views_async():
+                        sv.register_capability(toplevel_key)
+            self.send_response(Response(request_id, None))
+
+        sublime.set_timeout_async(run)
```
rchl, line 676: Does running those async can potentially cause a race issue where a server would expect those to be registered already when it sends the next (one that relies on registration) message?

rwols: The server can only be sure we're done with our bookkeeping once we have replied to the request, which we do after we did all the bookkeeping. So everything checks out I think.

plugin/core/sessions.py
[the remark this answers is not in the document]
rwols: > Can't we just check the size of self._session_views directly in instead of introducing this count?

No:
```
t0: last view close, debounced check after 3 seconds
t0 + 1s: open another view
t0 + 2s: close that view
t0 + 3s: debounced check is done, all views have been closed so we end
```
This would mean the session ends 1 second after apparently closing the last view. Using a counter prevents this.

plugin/core/sessions.py
```diff
@@ -429,16 +448,22 @@ class Session(Client):
...
+        self.views_opened = 0
         self._workspace_folders = workspace_folders
+        self._session_views = WeakSet()  # type: WeakSet[SessionViewProtocol]
         self._progress = {}  # type: Dict[Any, Dict[str, str]]
         self._plugin_class = plugin_class
         self._plugin = None  # type: Optional[AbstractPlugin]
-        super().__init__(logger)
+
+    def __del__(self) -> None:
+        debug(self.config.binary_args, "ended")
```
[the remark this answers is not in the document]
rwols: I agree debug logging should be kept to a minimum. But in this case I'd like to keep it for a while. It also makes the debug log in transports.py symmetric where we log that a process is starting (here we log that the process definitely ended).

plugin/core/configurations.py
rwols: There are three types of threads for us to worry about: the main thread, the ST worker thread, and our own spawned threads. I've started adding an `_async` suffix to functions that should run on the ST worker thread, but I don't have any runtime checks. It is perhaps a possibility to add a decorator for this (in another PR).

plugin/core/sessions.py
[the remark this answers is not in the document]
rchl: I wonder why `SessionView` handles that then. Couldn't `Session` handle debouncing when called from `SessionView` through `unregister_session_view_async`?

plugin/core/sessions.py
rwols: That’s a bug that I will have to work on.

plugin/core/sessions.py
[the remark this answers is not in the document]
rwols: That’s probably a better idea, thanks.

plugin/core/sessions.py
rchl: So for this PR maybe don't enable starting sessions for files outside of the workspace?

plugin/core/windows.py
```diff
@@ -343,144 +90,177 @@ def __init__(
...
-            for session in sessions:
-                session.update_folders(workspace_folders)
-
-    def on_pre_close_project(self) -> None:
-        self.end_sessions()
+        for session in self._sessions:
+            session.update_folders(workspace_folders)
+
+    def register_listener(self, listener: AbstractViewListener) -> None:
+        sublime.set_timeout_async(lambda: self.register_listener_async(listener))
```
rchl, line 112: I'll complain about method ordering again - it would be so much easier to follow the flow if this would be just after `enable_config_async`. And similarly for many other new methods.

plugin/core/views.py
rwols: Turns out this change is not needed after all.

plugin/core/views.py
rwols: I cannot trust pyls to pick up the formatting settings in tox.ini apparently. I reverted these formatting changes.

tests/setup.py
rwols: Reverted.

plugin/core/sessions.py
rwols: Done.

plugin/core/windows.py
```diff
@@ -343,144 +90,177 @@ def __init__(
...
+            # debug("found new config for listener", listener)
+            self._new_listener = listener
+            self.start_async(config, listener.view)
+        else:
+            # debug("no new config found for listener", listener)
+            self._new_listener = None
+            return self._dequeue_listener_async()
+
+    def _publish_sessions_to_listener_async(self, listener: AbstractViewListener) -> None:
+        scope = view2scope(listener.view)
```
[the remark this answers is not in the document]
rwols: Fixed.

plugin/core/windows.py
```diff
@@ -343,144 +90,177 @@ def __init__(
...
+            for session in self._sessions:
+                if session.can_handle(listener.view):
+                    # debug("registering session", session.config.name, "to listener", listener)
+                    listener.on_session_initialized_async(session)
+        else:
+            for session in self._sessions:
+                if session.config.match_document(scope):
+                    # debug("registering session", session.config.name, "to listener", listener)
+                    listener.on_session_initialized_async(session)
+                    break
```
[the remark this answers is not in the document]
rwols: I've left outside-workspace sessions for another PR.

plugin/core/session_view.py
```diff
@@ -0,0 +1,189 @@
...
+        self.changes.extend(changes)
+
+
+class SessionView:
+
+    def __init__(self, listener: AbstractViewListener, session: Session) -> None:
+        self.view = listener.view
+        self.session = session
+        self.listener = weakref.ref(listener)
+        self._file_name = self.view.file_name() or ""
```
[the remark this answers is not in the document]
rwols: We do have to store the filename (and the buffer ID, and the view ID), because the `__del__` method of SessionView is usually called later than the `__del__` of ViewListener.
Moreover, we also need to store the file name in case of a rename of the file.

plugin/core/windows.py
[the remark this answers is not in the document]
rwols: I've replaced `__contains__` by `contains` https://github.com/sublimelsp/LSP/pull/1069/commits/cd3a6f19676dfcc9a89eabddffd56817015bc0bc

plugin/core/windows.py
rwols: I moved enabled_config_async and disable_config_async up https://github.com/sublimelsp/LSP/pull/1069/commits/efecf49a5222ec8e38a79907ba93e66db3a4fdac

plugin/diagnostics.py
rwols: done https://github.com/sublimelsp/LSP/pull/1069/commits/0717444988be660b0e1ec5b4bd1e72b297451be4

plugin/core/workspace.py
rwols: https://github.com/sublimelsp/LSP/pull/1069/commits/b20b02f4f2143aef978a3b7a99901291ec7cbf42

plugin/core/sessions.py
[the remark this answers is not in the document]
rwols: https://github.com/sublimelsp/LSP/pull/1069/commits/b20518d00d0544aaf7b28ef927fe36c1cb785921

plugin/core/views.py
rwols: Let's do that in another PR.

plugin/core/windows.py
```diff
@@ -343,144 +90,177 @@ def __init__(
...
+            # debug('Already starting on this window:', config.name)
             return
         try:
             workspace_folders = sorted_workspace_folders(self._workspace.folders, file_path)
             plugin_class = get_plugin(config.name)
             if plugin_class is not None:
                 if plugin_class.needs_update_or_installation():
-                    self._window.status_message('Installing {} ...'.format(config.name))
+                    config.set_view_status(initiating_view, "installing...")
                     plugin_class.install_or_update()
```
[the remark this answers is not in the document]
rwols: It would require more bookkeeping for maintaining the "starting session" (the `self._new_session`). The benefit of running this in the async thread is that less bookkeeping is needed. Because all pending listeners are now flowing through that thread. No locks needed. I don't see much benefit in spawning an extra thread. Perhaps we should also spawn an extra thread for starting a process? That also takes some time. And a thread for stopping a process. I'm not really in the business of "helping" other plugins in case they depend on async callbacks being invoked immediately.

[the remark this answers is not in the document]
rwols: Note that with `asyncio` this all becomes very easy.

plugin/core/session_view.py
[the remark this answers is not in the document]
rchl: I saw that another place uses it to handle renaming but not this class. So I think that that argument doesn't apply here.

If we need it then let's use it everywhere, also in `on_text_changed`.

plugin/core/windows.py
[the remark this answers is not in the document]
rchl: I'm just concerned that installation will block other plugins and all STs *_async events from triggering. It just doesn't seem like a "fair play" to do that :)

plugin/core/types.py
```diff
@@ -91,10 +103,21 @@ def __init__(self,
...
+        view.set_status(self.status_key, status)
+
+    def erase_view_status(self, view: sublime.View) -> None:
+        view.erase_status(self.status_key)
 
     def match_document(self, scope: str) -> bool:
         return any(language.match_document(scope) for language in self.languages)
 
+    def match_view(self, view: sublime.View) -> bool:
+        return self.match_document(view2scope(view))
```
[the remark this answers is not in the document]
rwols: https://github.com/sublimelsp/LSP/pull/1069/commits/19c14d843a51fcd51f755f821db6713fbd5f43b8

PR #1071, opened 2020-05-24
Support completion item tags

predragnikolic:
```
/**
 * Client supports the tag property on a completion item. Clients supporting
 * tags have to handle unknown tags gracefully. Clients especially need to
 * preserve unknown tags when sending a completion item back to the server in
 * a resolve call.
 *
 * @since 3.15.0
 */
```
The `deprecated` flag on a `CompletionItem` was deprecated in favor of `CompletionItem.tags`. This PR add support for tags.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -18,6 +18,7 @@ import os
 def get_initialize_params(workspace_folders: List[WorkspaceFolder], config: ClientConfig) -> dict:
     completion_kinds = list(range(1, len(COMPLETION_KINDS) + 1))
     symbol_kinds = list(range(1, len(SYMBOL_KINDS) + 1))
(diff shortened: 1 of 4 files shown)


plugin/core/sessions.py
```diff
@@ -35,7 +35,10 @@ def get_initialize_params(workspace_folders: List[WorkspaceFolder], config: Clie
                 "dynamicRegistration": True,
                 "completionItem": {
                     "snippetSupport": True,
-                    "deprecatedSupport": True
+                    "deprecatedSupport": True,
+                    "tagSupport": {
+                        "valueSet": [CompletionItemTag.Deprecated]
```
rwols, line 40: This should be more automatic. Try:

```python
[v for k, v in CompletionItemTag.__dict__.items() if not k.startswith('_')]
```

predragnikolic: Thanks :)

PR #1072, opened 2020-05-24
Add badges for license and docs to README

jwortmann:
Maybe add some nice badges with links to the license and docs?
And use rounded corners for the Discord badge, so it looks like the other badges too.


predragnikolic:
Thanks!

PR #1073, opened 2020-05-25
Trigger warning for ST4 users that are still using ST3-version of LSP

rchl:
Since users upgrading to ST4 won't get automatically updated (apparently) to ST4-compatible LSP, we should do something about it or it can become a major hassle.

This is a proof of concept how we could handle it.

It doesn't work as-is, because both ST3 and ST4 versions of LSP use the same `__version__`. We should IMO actually bump `__version__` of ST4 branch a bunch of notches so that we can differentiate them. Then we can implement this solution properly.

### DIFF
--- a/boot.py
+++ b/boot.py
@@ -1,12 +1,3 @@
-import sublime
-
-if int(sublime.version()) > 4000:
-    sublime.error_message(
-        """The currently installed version of LSP package is not compatible with Sublime Text 4. """
-        """Please remove and reinstall this package to receive a version compatible with ST4. """
-        """Remember to restart Sublime Text after."""
-    )
-
 # Please keep this list sorted (Edit -> Sort Lines)
 from .plugin.code_actions import LspCodeActionBulbListener
 from .plugin.code_actions import LspCodeActionsCommand
@@ -20,7 +11,7 @@ from .plugin.configuration import LspEnableLanguageServerGloballyCommand
 from .plugin.configuration import LspEnableLanguageServerInProjectCommand
 from .plugin.core.documents import DocumentSyncListener
 from .plugin.core.main import shutdown as plugin_unloaded
-from .plugin.core.main import startup as plugin_loaded
+from .plugin.core.main import startup
 from .plugin.core.panels import LspClearPanelCommand
 from .plugin.core.panels import LspUpdatePanelCommand
 from .plugin.core.panels import LspUpdateServerPanelCommand
@@ -52,3 +43,15 @@ from .plugin.symbols import LspDocumentSymbolsCommand
 from .plugin.symbols import LspSelectionAddCommand
 from .plugin.symbols import LspSelectionClearCommand
 from .plugin.symbols import LspWorkspaceSymbolsCommand
+import sublime
+
+
+def plugin_loaded():
+    startup()
+
+    if int(sublime.version()) > 4000:
+        sublime.error_message(
+            """The currently installed version of LSP package is not compatible with Sublime Text 4. """
+            """Please remove and reinstall this package to receive a version compatible with ST4. """
+            """Remember to restart Sublime Text after."""
+        )
--- a/boot.py
+++ b/boot.py
@@ -2,9 +2,9 @@ import sublime
 
 if int(sublime.version()) > 4000:
     sublime.error_message(
-        """The installed version of LSP package is not compatible with Sublime Text 4.\n"""
-        """Please remove and reinstall this package to receive version compatible with ST4.\n"""
-        """Remember to restart Sublime Text after.\n"""
+        """The currently installed version of LSP package is not compatible with Sublime Text 4. """
+        """Please remove and reinstall this package to receive a version compatible with ST4. """
+        """Remember to restart Sublime Text after."""
     )
 
 # Please keep this list sorted (Edit -> Sort Lines)
--- a/boot.py
+++ b/boot.py
@@ -10,8 +10,8 @@ from .plugin.configuration import LspDisableLanguageServerInProjectCommand
 from .plugin.configuration import LspEnableLanguageServerGloballyCommand
 from .plugin.configuration import LspEnableLanguageServerInProjectCommand
 from .plugin.core.documents import DocumentSyncListener
-from .plugin.core.main import shutdown as plugin_unloaded
-from .plugin.core.main import startup
+from .plugin.core.main import plugin_loaded
+from .plugin.core.main import plugin_unloaded
 from .plugin.core.panels import LspClearPanelCommand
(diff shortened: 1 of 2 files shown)


boot.py
```diff
@@ -1,3 +1,13 @@
+import sublime
+from .plugin import __version__
+
+if __version__ <= (0, 11, 0) and int(sublime.version()) > 4000:
+    sublime.error_message(
+        """Installed version of LSP package is not compatible with Sublime Text 4.\n"""
+        """Please remove and re-install this package to receive ST4-compatible version."""
```
predragnikolic, line 7: Just add the last step,and that is to reopen Sublime. 🙂

boot.py
```diff
@@ -1,3 +1,13 @@
+import sublime
+from .plugin import __version__
+
+if __version__ <= (0, 11, 0) and int(sublime.version()) > 4000:
```
rwols, line 4: Why do we need to check the `__version__` at all?
Just checking `sublime.version()` seems sufficient.

boot.py
```diff
@@ -1,3 +1,13 @@
...
+    sublime.error_message(
```
rwols, line 5: This entire check needs to move to plugin_loaded (also called startup in this plugin for some reason). Calling functions outside plugin_loaded is undefined.

rchl: From documentation:
```
At importing time, plugins may not call any API functions, with the exception of sublime.version(), sublime.platform(), sublime.architecture() and sublime.channel().
```

boot.py
rchl: We don't need to check as long as we won't merge this change to ST4.
I'm OK with that if we can ensure that.

boot.py
rwols: Hmm, I don’t see error_message among those functions, but I suppose that should always work too.

rwols: Perhaps it's safer to raise an ImportError instead, looking at this example: https://github.com/SublimeText/AFileIcon/blob/master/plugin.py

boot.py
predragnikolic: > Why do we need to check the __version__ at all?

To make a difference between
the LSP version for ST3 and
the LSP version for ST4. 

> Just checking sublime.version() seems sufficient.

Not really. 
In that case, If we just check the Sublime version, we will get the error popup each time we open ST no meather the LSP version.

The intent is to show the error only for user who are using ST4 and have LSP version for ST3. 
Users who are using ST4 and the LSP version for ST4 should not/won't see the error message.

Did I understood you correctly @rchl?

boot.py
rchl: And not initialize LSP?
ST3 still works, just not as well, so I wouldn't block it completely.

As for `sublime.error_message`, I have overlooked that one. I also think it should just work as it shouldn't rely on anything in the API but we can ask maybe.

rchl:
> Won't this now throw an error each time we start Sublime?

Yes, until user upgrades.
It's once per session so it's not that bad IMO.
Could be more annoying for us if we're gonna develop something.

boot.py
rchl: You understand correctly but I'm assuming we won't merge this change into ST4 branch and then it's fine.

predragnikolic:
And when he upgrade Sublime text,
User will still get that error because the installed version of ST will be grather than 4000..

if int(sublime.version()) > 4000:

boot.py
rwols: I got confused, I do want to get this into ST4 too. @predragnikolic is correct.
I guess it should say:

```python
if __version__ >= (0, 11, 0) and int(sublime.version()) < 4070:
```

predragnikolic:
Will this only be merged in the master branch? If so just innore me

rchl:
> Will this only be merged in the master branch? If so just innore me

I'm not repeating myself third time :)

boot.py
rchl: Then we should raise version of ST4 branch to be further away from ST3 branch.
Otherwise we can't handle that properly.
We could change ST4 branch version to 1.0.0 even. It doesn't really matter as long as they are separate and further apart.

predragnikolic: Both solutions are fine.
Either add tow different versions for LSP3 and LSP4 and do the LSP version check,

or.
Just check the sublime.version() (withouth the lsp version check)
merge it into master
merge master into st-4000
revert this commit on the st-4000

rwols: Are there any pitfalls in maintaining two versions? It's boring maintaining even one version number.

rwols: When we only do the check in master, and revert the commit in the st4 branch, then when we eventually merge the st4 branch into master, the version check is gone.

rwols: I'm okay with it becoming an issue again once we merge st4 into master, so let's roll with this and undo the commit when master is merged into st4.

boot.py
```diff
@@ -1,3 +1,12 @@
+import sublime
+
+if int(sublime.version()) > 4000:
```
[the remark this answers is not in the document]
rchl: Only if we want to be picky.
In practice people will be using latest version of ST4 or latest version of ST3.

boot.py
rchl: So what to do then?
Add version check for LSP version? (Then we need to set them apart first).
Or just keep as is?
If maintaining two versions is gonna be a chore then I guess we should first come up with some script that helps with releasing (maybe based on gitgutter's).

rchl: Re-read your last comment and as I understand we keep as is. :)

boot.py
```diff
@@ -1,3 +1,12 @@
...
+    sublime.error_message(
+        """The installed version of LSP package is not compatible with Sublime Text 4.\n"""
```
rwols, line 5: The installed -> The currently installed
... of LSP package ... -> ... of the LSP package ...

boot.py
```diff
@@ -1,3 +1,12 @@
...
+        """Please remove and reinstall this package to receive version compatible with ST4.\n"""
```
rwols, line 6: to receive version -> to receive a version

boot.py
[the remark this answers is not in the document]
rwols: Alright, sounds good.

predragnikolic:
https://discord.com/channels/280102180189634562/280157083356233728/714741294160674856

> Essentially all APIs that talk to the main process should not be called. version, etc. are implemented by the plugin host and don't talk to the main process. error_message and message_dialog are implemented by the main process, so shouldn't be called.

I hope that @rchl won't mind if I answer you instead of him :)

rchl:
> LGTM. Why did you not put this check in the `startup` function?

I'm not sure. I guess I'm used to seeing `plugin_loaded` and `plugin_unloaded` in the main plugin file.

PR #1075, opened 2020-05-29
Fix white space within links not being clickable

rchl:
ST is behaving weird and non-HTML-like when there are spaces within
an anchor. Those should still be part of the link and clickable but
in ST they are not. Work around by replacing spaces with non-breakable
ones.


rchl:
This fix will apply to ST4 with some minor conflicts due to minihtml being added there. But nothing serious so I've decided to put it on master.

rchl:
https://github.com/sublimehq/sublime_text/issues/3373

Turns out it's only when removing underline.

PR #1077, opened 2020-05-30
Remove client_from_session

rwols:
The extra indirection is no longer needed.

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -4,7 +4,7 @@ from .core.edit import parse_workspace_edit
 from .core.protocol import Diagnostic
 from .core.protocol import Request, Point
 from .core.registry import LspTextCommand
-from .core.registry import sessions_for_view, client_from_session
+from .core.registry import sessions_for_view
 from .core.settings import settings
 from .core.typing import Any, List, Dict, Callable, Optional, Union, Tuple, Mapping, TypedDict
 from .core.url import filename_to_uri
@@ -150,9 +150,8 @@ def is_command(command_or_code_action: CodeActionOrCommand) -> bool:
 
 def execute_server_command(view: sublime.View, config_name: str, command: Mapping[str, Any]) -> None:
     session = next((session for session in sessions_for_view(view) if session.config.name == config_name), None)
-    client = client_from_session(session)
-    if client:
-        client.send_request(
+    if session:
+        session.send_request(
             Request.executeCommand(command),
             handle_command_response)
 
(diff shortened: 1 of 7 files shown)


jwortmann:
Will this and the other recent or planned API changes will get backported to ST3? I have a plugin for the Julia server for personal use, that I might want to publish on Package Control some time in the future when it is ready, and it requires this client_from_session function. If the changes get backported, I can easily adjust the plugin to use the updated functions/API. But I don't want to support two different versions of the plugin and otherwise I would have to decide for ST3 or ST4.

rwols:
The changes that preceded this use the new `on_exit` callback from `sublime_plugin.EventListener`. It's used to fix an ancient problem where subprocesses are lingering after ST exits. So, that cannot be backported.

I'll try keeping current usages of LSP in other plugins in mind, but keep in mind there's no "official" stability promises. The old "API" also has serious problems:

1. Versioning. There's no versioning. (now fixed)
2. No real place or promise on what's a public API. We now place stuff in plugin/\_\_init\_\_.py and you may consider that "public", but expect to see some changes there still.
2. No plugin registration through functions like `register_plugin` but iterating over `__subclasses__()` which turned out to be a bad idea. (on my todo list)
3. Plugins cannot cooperate with the base LSP package for just-in-time installation of binaries (now fixed)

I'd suggest you go with the ST4 API; do `from LSP.plugin import AbstractPlugin`, and implement `name()`, `configuration()`, `needs_update_or_installation()` and `install_or_update()`.

There's no real public API to implement custom requests, but I think we should indeed make LspTextCommand public (and probably also add an LspWindowCommand).

rwols:
Also expect so see another large API change once we move to python3.8 and start using `async` *everywhere*.

jwortmann:
Ok, thanks for explanation. I noticed that I could still use sublime.version() to distinguish between ST3 and ST4, that would be a bit messy though but it should work for now.

> I'll try keeping current usages of LSP in other plugins in mind, but keep in mind there's no "official" stability promises. The old "API" also has serious problems:

Yes I understand, and I agree it's a good thing to move forward with those "API changes". Once ST4 is publicly released and the st4000 branch gets merged into master, we should create a st3 branch from master beforehand though, so we still have the possibility to keep some compatibily for ST3 users for breaking changes (or possible bug fixes).

> There's no real public API to implement custom requests, but I think we should indeed make LspTextCommand public (and probably also add an LspWindowCommand).

Oops, I actually didn't even notice that it was there... I will use LspTextCommand then instead of importing various functions. Sounds good to make it part of the "public API". And maybe also add functions `send_notification` and `send_request`, which could more easily used by plugins then.

PR #1078, opened 2020-05-30
Replace 'client_with_capability' and 'has_client_with_capability'

rwols:
Replace it with `self.session("capability.path")`.


predragnikolic:
This is a really neat change 🙂

PR #1079, opened 2020-05-30
Standardize on having a "capability" in your LspTextCommand

rwols:
It looks like we more or less have one LspTextCommand per server capability.

### DIFF
--- a/plugin/core/registry.py
+++ b/plugin/core/registry.py
@@ -93,26 +93,39 @@ def configurations_for_view(view: sublime.View) -> Generator[ClientConfig, None,
         yield from windows.lookup(window)._configs.match_view(view)
 
 
-def is_supported_view(view: sublime.View) -> bool:
-    # TODO: perhaps make this check for a client instead of a config
-    return any(configurations_for_view(view))
+def get_position(view: sublime.View, event: Optional[dict] = None) -> int:
+    if event:
+        return view.window_to_text((event["x"], event["y"]))
+    else:
+        return view.sel()[0].begin()
 
 
 class LspTextCommand(sublime_plugin.TextCommand):
-    def __init__(self, view: sublime.View) -> None:
-        super().__init__(view)
+
+    capability = ''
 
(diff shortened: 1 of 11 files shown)


plugin/core/registry.py
```diff
@@ -93,26 +93,39 @@ def configurations_for_view(view: sublime.View) -> Generator[ClientConfig, None,
...
+            return bool(self.session(self.capability, get_position(self.view, event)))
+        else:
+            # Any session will do.
+            return any(self.sessions())
+
+    def is_enabled(self, event: Optional[dict] = None) -> bool:
+        return self.is_visible(event)
+
+    def want_event(self) -> bool:
+        return True
```
rchl, line 119: There are two other places where this can be removed now.

plugin/core/registry.py
```diff
@@ -93,26 +93,39 @@ def configurations_for_view(view: sublime.View) -> Generator[ClientConfig, None,
...
-        return is_supported_view(self.view)
+        if self.capability:
+            # At least one active session with the given capability must exist.
+            return bool(self.session(self.capability, get_position(self.view, event)))
+        else:
+            # Any session will do.
+            return any(self.sessions())
+
+    def is_enabled(self, event: Optional[dict] = None) -> bool:
+        return self.is_visible(event)
```
rchl, line 116: I guess we could just skip `is_enabled` in the base class but it's a bit of a style choice, I suppose, so do whatever you fell doing. :)

plugin/core/registry.py
```diff
@@ -93,26 +93,39 @@ def configurations_for_view(view: sublime.View) -> Generator[ClientConfig, None,
         yield from windows.lookup(window)._configs.match_view(view)
 
 
-def is_supported_view(view: sublime.View) -> bool:
-    # TODO: perhaps make this check for a client instead of a config
-    return any(configurations_for_view(view))
+def get_position(view: sublime.View, event: Optional[dict] = None) -> int:
```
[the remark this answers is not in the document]
rchl: Or maybe not since event is optional..

PR #1082, opened 2020-06-01
Load "settings" from base file for plugin authors

rwols:
Then, users have to override settings with "settings".

I tested this with LSP-eslint. I modified Packages/LSP-eslint/LSP-eslint.sublime-settings as follows:

```js
{
	"languages": [
		{"languageId": "javascriptreact", "document_selector": "source.js.react | source.jsx"},		
		{"languageId": "javascript",      "document_selector": "source.js"},
		{"languageId": "vue",             "document_selector": "text.html.vue"}
	],
	"initializationOptions": {},
	"default_settings": {
		"validate": true,
		"packageManager": "npm",
		"autoFix": true,
		"autoFixOnSave": true,
		"options": {},
		"run": "onType",
		"nodePath": null,
		"quiet": false,
		"workspaceFolder": null,
		"codeAction.disableRuleComment.enable": true,
		"codeAction.disableRuleComment.location": "separateLine",
		"codeAction.showDocumentation.enable": true
	},
}
```

I then ran "Preferences: LSP-eslint settings" from the command palette and put the following in my Packages/User/LSP-eslint.sublime-settings:

```js
// Settings in here override those in "LSP-eslint/LSP-eslint.sublime-settings"
{
	// Overrides for "default_settings"
	"settings":
	{
		"codeAction.disableRuleComment.enable": false
	}
}
```

The workspace/didChangeConfiguration notification correctly sent the override:

```
::  -> lsp-eslint workspace/didChangeConfiguration: {'settings': {'codeAction': {'showDocumentation': {'enable': True}, 'disableRuleComment': {'location': 'separateLine', 'enable': False}}, 'quiet': False, 'workspaceFolder': None, 'run': 'onType', 'nodePath': None, 'validate': True, 'packageManager': 'npm', 'autoFixOnSave': True, 'autoFix': True}}
```

### DIFF
--- a/plugin/core/settings.py
+++ b/plugin/core/settings.py
@@ -113,6 +113,8 @@ class ClientConfigs(object):
             self._listener()
 
     def add_external_config(self, name: str, s: sublime.Settings) -> None:
+        settings = DottedDict(read_dict_setting(s, "default_settings", {}))  # defined by the plugin author
+        settings.update(read_dict_setting(s, "settings", {}))  # overrides from the user
         config = ClientConfig(
             name=name,
             binary_args=read_array_setting(s, "command", []),
@@ -121,7 +123,7 @@ class ClientConfigs(object):
             # Default to True, because an LSP plugin is enabled iff it is enabled as a Sublime package.
             enabled=bool(s.get("enabled", True)),
             init_options=s.get("initializationOptions"),  # type: ignore
-            settings=DottedDict(read_dict_setting(s, "settings", {})),
+            settings=settings,
             env=read_dict_setting(s, "env", {}),
             tcp_host=s.get("tcp_host", None),
             tcp_mode=s.get("tcp_mode", None),
--- a/plugin/core/settings.py
+++ b/plugin/core/settings.py
@@ -112,8 +112,9 @@ class ClientConfigs(object):
         if self._listener:
             self._listener()
 
-    def add_external_config(self, name: str, s: sublime.Settings) -> None:
-        settings = DottedDict(read_dict_setting(s, "default_settings", {}))  # defined by the plugin author
+    def add_external_config(self, name: str, s: sublime.Settings, file: str) -> None:
+        base = sublime.decode_value(sublime.load_resource(file))
+        settings = DottedDict(base.get("settings", {}))  # defined by the plugin author
         settings.update(read_dict_setting(s, "settings", {}))  # overrides from the user
         config = ClientConfig(
             name=name,
--- a/plugin/core/main.py
+++ b/plugin/core/main.py
@@ -19,7 +19,7 @@ from .settings import client_configs
 from .settings import settings, load_settings, unload_settings
 from .transports import kill_all_subprocesses
 from .types import ClientConfig
-from .typing import Optional, List, Type, Callable, Dict
+from .typing import Optional, List, Type, Callable, Dict, Tuple
 import weakref
 
 
@@ -61,7 +61,7 @@ def _forcefully_register_plugins() -> None:
                 return cls.handler.name  # type: ignore
 
             @classmethod
-            def configuration(cls) -> sublime.Settings:
+            def configuration(cls) -> Tuple[sublime.Settings, str]:
                 file_base_name = cls.name()
                 if file_base_name.startswith("lsp-"):
                     file_base_name = "LSP-" + file_base_name[len("lsp-"):]
@@ -78,7 +78,7 @@ def _forcefully_register_plugins() -> None:
                         "feature_selector": language.feature_selector
                     })
                 settings.set("languages", langs)
-                return settings
+                return settings, "Packages/{0}/{0}.sublime-settings".format(file_base_name)
 
             @classmethod
             def can_start(cls, window: sublime.Window, initiating_view: sublime.View,
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -291,14 +291,14 @@ class AbstractPlugin(metaclass=ABCMeta):
         raise NotImplementedError()
 
     @classmethod
-    def configuration(cls) -> sublime.Settings:
+    def configuration(cls) -> Tuple[sublime.Settings, str]:
         """
-        The Settings object that defines the "command", "languages", and optionally the "initializationOptions",
-        "default_settings", "env" and "tcp_port".
+        Return the Settings object that defines the "command", "languages", and optionally the "initializationOptions",
+        "default_settings", "env" and "tcp_port" as the first element in the tuple, and the path to the base settings
+        filename as the second element in the tuple.
 
-        The "settings" dict is also supported, but not recommended for use. Because when a user defines its own
-        settings, the entire dict is overwritten. Because sublime.Settings objects don't recurse overrides in JSON
-        objects. Define your default settings in the "default_settings" object.
+        The second element in the tuple is used to handle "settings" overrides from users properly. For example, if your
+        plugin is called LSP-foobar, you would return "Packages/LSP-foobar/LSP-foobar.sublime-settings".
 
         The "command", "initializationOptions" and "env" are subject to template string substitution. The following
         template strings are recognized:
@@ -329,7 +329,10 @@ class AbstractPlugin(metaclass=ABCMeta):
         When you're managing your own server binary, you would typically place it in sublime.cache_path(). So your
         "command" should look like this: "command": ["$cache_path/LSP-foobar/server_binary", "--stdio"]
         """
-        return sublime.load_settings('LSP-{}.sublime-settings'.format(cls.name()))
+        name = cls.name()
+        basename = "LSP-{}.sublime-settings".format(name)
+        filepath = "Packages/LSP-{}/{}".format(name, basename)
+        return sublime.load_settings(basename), filepath
 
     @classmethod
     def additional_variables(cls) -> Optional[Dict[str, str]]:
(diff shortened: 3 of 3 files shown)


rwols:
Project overrides is coming next...

rwols:
cc @jwortmann, does this look OK?

rchl:
Why is that necessary?
To not accidentally override all default settings when changing a single one?
But that is solved by "dotted" properties already anyway, right?

Also, if there is such treatment for "settings" then shouldn't there be the same for "languages" and all the other keys in default settings?

I feel that it complicates things for the user and not sure what's are the benefits.

jwortmann:
From a functionality point of view this should work fine (with the dotted settings), but I agree that it might be a bit confusing for users. This is basically the same as with the "clients" and "default_clients" in the LSP settings, but maybe there is a better approach of it for the "settings". I think to keep the same name for the default and user "settings" would be possible with some effort. We could still use the sublime.Settings object to get notifications about changes, but we would need to manually append all of those key-value pairs from the default "settings", which are not present in the user settings. Therefore we would need the path to the default settings file (package name and settings filename) and call `sublime.load_resource()` and `sublime.decode_value()` on it to find the additional values (one time at the start, and each time when the settings object changes).

> Why is that necessary?
> To not accidentally override all default settings when changing a single one?
> But that is solved by "dotted" properties already anyway, right?

I don't think so, because the dotted settings are still inside of a "settings" dict, and as soon as you have "settings" inside your User .sublime-settings file, it will overwrite the whole default "settings" dict. That's because `sublime.load_settings()` only overwrites values and doesn't merge them, if those values are dicts.

> Also, if there is such treatment for "settings" then shouldn't there be the same for "languages" and all the other keys in default settings?

The "languages" and other keys are probably rarely changed by users, so in my opinion it would be sufficient to only merge the "settings" dict.

rchl:
> I don't think so, because the dotted settings are still inside of a "settings" dict, and as soon as you have "settings" inside your User .sublime-settings file, it will overwrite the whole default "settings" dict. That's because `sublime.load_settings()` only overwrites values and doesn't merge them, if those values are dicts.

Yep, I see...

Then I guess we need to do that but I still think it might be confusing for people to have such special handling for settings. At the minimum, we would have to update the user settings template to match something like:
```js
// Settings in here override those in "LSP-eslint/LSP-eslint.sublime-settings"
{
	// Overrides for "default_settings"
	"settings":
	{
	}
}
```

rwols:
It also solves the ST3/ST4 maintenance problem. You can put  path-style settings in "default_settings", and keep the dict-style settings "settings" for ST3.

predragnikolic:
I really don't mind if this gets merged. 
But I really would not rush with it and I would maybe tell you to take some time to think about it once more. 
This PR only solves one problem, but not all related problems. 
And by one problem I mean, it fixes the merging of the `settings` key. 
What about the `initializationOptions` and `languages`? 

Is there a way that we can get away without adding the default_settings, and solve merging of all keys in the settings file? :)

rwols:
The initializationOptions and languages are keys that users will have to override in full because they're usually not very large dictionaries.

> Is there a way that we can get away without adding the default_settings, and solve merging of all keys in the settings file? :)

Well, we could manually load files with

```python
base = sublime.decode_value(sublime.load_resource("Packages/LSP-foobar/LSP-foobar.sublime-settings"))
overrides = sublime.decode_value(sublime.load_resource("Packages/User/LSP-foobar.sublime-settings"))
```
I don't like this approach because it requires more code, and it will load information from disk that's already in memory (in memory via the sublime.Settings object)

rchl:
Also, to take advantage of this change, and keep compatibility with ST3, we would have to have both `settings` and `default_settings` keys in *.sublime-settings files?

rwols:
Yes you would have both keys present, but only for the package maintainers. An end-user would only have the “settings” key (if that user overrides some setting).

rchl:
That's the part I don't like very much as it creates confusion. Sure, we can make it clearer by adding comments there but it would be better if we wouldn't have to...

jwortmann:
> Well, we could manually load files with
> 
> ```python
> base = sublime.decode_value(sublime.load_resource("Packages/LSP-foobar/LSP-foobar.sublime-settings"))
> overrides = sublime.decode_value(sublime.load_resource("Packages/User/LSP-foobar.sublime-settings"))
> ```

We only need the first line of that, because all information of the User settings are already inside the sublime.Settings object. And now where I look at it, we actually would need to do that only one single time when the server initializes, if we store the package's default settings into a variable, because that file never changes on disk. Something like this:
```python
base = sublime.decode_value(sublime.load_resource("Packages/LSP-foobar/LSP-foobar.sublime-settings"))
self._default_settings = base.get("settings", {})

user_settings = settings.get("settings", {})  # settings is the sublime.Settings object here
for key, value in self._default_settings.items():
  if key not in user_settings:
    user_settings[key] = value
```

When we get a notification that the User settings have changed, we onle need to do the bottom part again, but don't have to read anything from disk.
I think that would be worth it, because it's not too much code and wouldn't really affect performance.

rwols:
Alright, I'll go with the `sublime.decode_value` approach.

plugin/core/settings.py
```diff
@@ -112,7 +112,10 @@ def remove_for_testing(self, config: ClientConfig) -> None:
         if self._listener:
             self._listener()
 
-    def add_external_config(self, name: str, s: sublime.Settings) -> None:
+    def add_external_config(self, name: str, s: sublime.Settings, file: str) -> None:
+        base = sublime.decode_value(sublime.load_resource(file))
```
rwols, line 116: It should be saved somewhere, but because add_external_config and remove_external_config are actually useless (because of #899), I'll add that later.

randy3k:
I have a custom LanguageHandler

```
 class LspRLangPlugin(LanguageHandler):
        @property
        def name(self):
            return "rlang"
```

And got the following error. I am not sure if it is intended.

```
Failed to register plugin
  File "/Users/Randy/Dropbox/Sublime Text/Packages/LSP/plugin/core/sessions.py", line 406, in register_plugin
    client_configs.add_external_config(name, *plugin.configuration())
  File "/Users/Randy/Dropbox/Sublime Text/Packages/LSP/plugin/core/settings.py", line 119, in add_external_config
    base = sublime.decode_value(sublime.load_resource(file))
  File "/Applications/Sublime Text.app/Contents/MacOS/Lib/python33/sublime.py", line 240, in load_resource
    raise IOError('resource "%s" not found' % name)
OSError: resource "Packages/rlang/rlang.sublime-settings" not found
```

rwols:
Given that you're checking for ST build >4000 I'd suggest working with AbstractPlugin (use `from LSP.plugin import AbstractPlugin`, the "public API" is supposed to be `LSP.plugin` instead of the current wild-west situation. That class has a number of advantages over LanguageHandler: 1) it communicates with the base package about optionally installing dependencies 2) it's instantiated per *session* and not a singleton 3) it's more straightforward to write custom request handlers with `m_foobar` methods

I've written a [temporary transition class](https://github.com/sublimelsp/LSP/blob/2b5f2894215f84d482c5e9bd4f647ac18f985633/plugin/core/main.py#L42-L107) that wraps existing LanguageHandler classes into an AbstractPlugin class. It goes wrong in that place, because the assumption for the
```
@classmethod
def configuration(cls) -> Tuple[sublime.Settings, str]:
    ...
```
classmethod in that link assumes that the language handler plugin has the following layout:

1) If the name is Foo, then there exists a settings file called Packages/LSP-Foo/LSP-Foo.sublime-settings file
2) The settings file has the following entries: `"languages"`, `"initializationOptions"` (optional), `"command"`, `"settings"` (optional), `"env"` (optional)

The `"settings"` (server settings) have a long history of being a PITA. It's now established that:

1) It should be a dict
2) Keys are dotted paths: "foo.bar.baz"
3) Values are preferably not dicts

This mirrors what most vscode-based language servers present as their settings. For example, in their readme they might say:

> Settings:
> `"foo.bar.baz": 4` (analyze baz)
> `"foo.bar.qux": ["a", "b"]` (error levels a and b)

The `"settings"` dict should mirror this:

```
// Packages/LSP-Foo/LSP-Foo.sublime-settings
{
    "languages": [{"languageId": "r", "document_selector": "source.r|text.html.markdown.rmarkdown"}],
    "settings":
    {
        "foo.bar.baz": 4,
        "foo.bar.qux": ["a", "b"]
    }
}
```

Now if a user overrides `"settings"` like this:

```
// Packages/User/LSP-Foo.sublime-settings
{
    "settings":
    {
        "foo.bar.baz": 5,
    }
}
```
Things will just work; the `"foo.bar.qux"` setting is still sent in our `workspace/didChangeConfiguration` notification to servers and in `workspace/configuration` requests from servers.

rchl:
I'm not sure if how that plugin handles default settings is gonna be compatible with AbstractPlugin. Uses dynamically resolved settings rather than an actual sublime-settings file.

https://github.com/REditorSupport/sublime-ide-r/blob/master/ride/settings.py

rwols:
> Uses dynamically resolved settings rather than an actual sublime-settings file.

Not really. There's still an R-IDE.sublime-settings file. Load that and then fill in the `command`, `languages` in code and things will work.

There does seem to be a problem with reading `"settings"` from sublime.decode_value -- that's where we assume `"settings"` exist and is in the way I described above.

Well, we can argue until the cows come home but this is the way forward.

There will be another breaking change for fixing #899. Plugins will have to explicitly register themselves with `register_plugin` and unregister themselves with `unregister_plugin`. Those functions currently do nothing useful.

> BTW. I'm not a fan of the magical handling of LSP- prefix. I think I'd rather prefer to have everything explicit - that plugin can specify its name and the settings file name itself.

Then override `def configuration(cls)` and explicitly tell where your settings file lives.

PR #1087, opened 2020-06-02
Implement Code Actions On Save

rchl:
## TODO
 - [x] 1. On triggering CAOS (Code Actions On Save), only request code actions from servers that report `codeActionProvider.codeActionKinds` and only request those specific kinds. ESlint reacts very weirdly if we send `source.fixAll.stylelint` to it, for example.
 - [x] 2. Implement more sophisticated `lsp_code_actions_on_save` setting handling. There are various combinations of that setting that can be applied by the user. There is a couple of examples in [vscode-stylelint readme](https://github.com/stylelint/vscode-stylelint#editorcodeactionsonsave). For an example having `source.fixAll` set, that matches ESlint's more specific kind  `source.fixAll.eslint`, we send a request with  `source.fixAll.eslint` kind. When having both `source.fixAll: true` and `source.fixAll.eslint: false` in settings, we must not send the request.
 - [x] 3. Making a change and saving quickly fails to correct fixes. That's because, on a change, we clear diagnostics and send `didChange` notification, then find no diagnostics so we don't fire code action request. Only after that, we get diagnostics but it's too late.
 - [x] 4. Abandon sync handling in favor of overriding `save` (and `save_all`) command. Sync request blocks the main thread which is not an ideal experience. Especially if the fix takes a second or more.
 - [x] 5. Request and apply fixes from all servers on saving. That means we have to re-request code actions (which also means we have to wait for diagnostics) between each batch of fixes. And we need to request them until we exhaust all fixes or we time out...

## Questions/doubts
 - [x] Q: Should we only apply fixes with TextEdit. `Command`s need extra round-trip.
   A: We are applying all right now. I think it's just fine but haven't seen any servers using `commands` for this functionality.
 - [x] Q: When we get fixes from multiple servers then what do we do? It's possible that fixes from one server will invalidate fixes from another server. It's fine if the fixes were versioned with `VersionedTextDocumentIdentifier` but otherwise, they might apply to a document that has already changed.
  A: At current state, we're re-applying code actions as long as they make changes in the document, limited by time (2s).

## Related links
[Code Action Request Spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_codeAction)

### DIFF
--- a/Keymaps/Default (Windows).sublime-keymap
+++ b/Keymaps/Default (Windows).sublime-keymap
@@ -62,4 +62,7 @@
 
     // Symbol Hover
     // {"keys": ["UNBOUND"], "command": "lsp_hover", "context": [{"key": "setting.lsp_active"}]},
+
+    // Override native save to handle Code-Actions-On-Save
+    { "keys": ["ctrl+s"], "command": "lsp_save", "context": [{ "key": "setting.lsp_active"}] },
 ]
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -8,6 +8,22 @@
   // Valid values are "never", "always" and "saved"
   "auto_show_diagnostics_panel": "saved",
 
+  // A dictionary of code action identifiers that should be triggered on save.
+  //
+  // Code action identifiers are not officially standardized so refer to specific
+  // server's documentation on what is supported but `source.fixAll` is commonly
+  // used to apply fix-on-save code action.
+  //
+  // It's also supported to set this in project-specific settings in which case
+  // it will only apply in that specific project. Setting it here will apply globally,
+  // unless overridden by project-specific settings.
+  //
+  // Only "source.*" actions are supported.
+  "lsp_code_actions_on_save": {},
+
+  // The amount of time the code actions on save are allowed to run for.
+  "code_action_on_save_timeout_ms": 2000,
+
   // Open the diagnostics panel automatically
   // when diagnostics level is equal to or less than:
   // error: 1
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -8,19 +8,23 @@
   // Valid values are "never", "always" and "saved"
   "auto_show_diagnostics_panel": "saved",
 
+  // Run the server's formatProvider (if supported) on a document before saving.
+  // This option is also supported in syntax-specific settings and/or in project files.
+  "lsp_format_on_save": false,
+
   // A dictionary of code action identifiers that should be triggered on save.
   //
   // Code action identifiers are not officially standardized so refer to specific
   // server's documentation on what is supported but `source.fixAll` is commonly
-  // used to apply fix-on-save code action.
+  // used to apply fix-on-save code actions.
   //
-  // It's also supported to set this in project-specific settings in which case
-  // it will only apply in that specific project. Setting it here will apply globally,
-  // unless overridden by project-specific settings.
+  // This option is also supported in syntax-specific settings and/or in project files.
+  // Settings from all places will be merged and more specific (syntax and project) settings will
+  // override less specific (from LSP or Sublime settings).
   //
   // Only "source.*" actions are supported.
   "lsp_code_actions_on_save": {
-    // "source.fixAll": True
+    // "source.fixAll": true
   },
 
   // The amount of time the code actions on save are allowed to run for.
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -19,7 +19,9 @@
   // unless overridden by project-specific settings.
   //
   // Only "source.*" actions are supported.
-  "lsp_code_actions_on_save": {},
+  "lsp_code_actions_on_save": {
+    // "source.fixAll": True
+  },
 
   // The amount of time the code actions on save are allowed to run for.
   "code_action_on_save_timeout_ms": 2000,
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -9,7 +9,8 @@
   "auto_show_diagnostics_panel": "saved",
 
   // Run the server's formatProvider (if supported) on a document before saving.
-  // This option is also supported in syntax-specific settings and/or in project files.
+  // This option is also supported in syntax-specific settings and/or in the "settings" section
+  // of project files
   "lsp_format_on_save": false,
 
   // A dictionary of code action identifiers that should be triggered on save.
@@ -18,9 +19,9 @@
   // server's documentation on what is supported but `source.fixAll` is commonly
   // used to apply fix-on-save code actions.
   //
-  // This option is also supported in syntax-specific settings and/or in project files.
-  // Settings from all places will be merged and more specific (syntax and project) settings will
-  // override less specific (from LSP or Sublime settings).
+  // This option is also supported in syntax-specific settings and/or in the "settings" section of
+  // project files. Settings from all those places will be merged and more specific (syntax and
+  // project) settings will override less specific (from LSP or Sublime settings).
   //
   // Only "source.*" actions are supported.
   "lsp_code_actions_on_save": {
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -9,8 +9,8 @@
   "auto_show_diagnostics_panel": "saved",
 
   // Run the server's formatProvider (if supported) on a document before saving.
-  // This option is also supported in syntax-specific settings and/or in the "settings" section
-  // of project files
+  // This option is also supported in syntax-specific settings and/or in the
+  // "settings" section of project files.
   "lsp_format_on_save": false,
 
   // A dictionary of code action identifiers that should be triggered on save.
@@ -19,9 +19,10 @@
   // server's documentation on what is supported but `source.fixAll` is commonly
   // used to apply fix-on-save code actions.
   //
-  // This option is also supported in syntax-specific settings and/or in the "settings" section of
-  // project files. Settings from all those places will be merged and more specific (syntax and
-  // project) settings will override less specific (from LSP or Sublime settings).
+  // This option is also supported in syntax-specific settings and/or in the
+  // "settings" section of project files. Settings from all those places will be
+  // merged and more specific (syntax and project) settings will override less
+  // specific (from LSP or Sublime settings).
   //
   // Only "source.*" actions are supported.
   "lsp_code_actions_on_save": {
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -235,7 +235,7 @@ class LspSaveCommand(sublime_plugin.TextCommand):
 
     def _trigger_native_save(self) -> None:
         self._task = None
-        # Triggered from set timeout to preserve original semantics of on_pre_save handling
+        # Triggered from set_timeout to preserve original semantics of on_pre_save handling
         sublime.set_timeout(lambda: self.view.run_command('save', {"async": True}))
 
 
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -255,19 +255,12 @@ class CodeActionOnSaveTask:
         self._on_done = on_done
         self._completed = False
         self._canceled = False
-        # TODO: For limiting number of iterations the task can run. Maybe not needed since there is a timeout?
-        self._iteration = 0
 
     def run(self) -> None:
         sublime.set_timeout(self._on_timeout, settings.code_action_on_save_timeout_ms)
         self._request_code_actions()
 
     def _request_code_actions(self) -> None:
-        self._iteration += 1
-        if self._iteration > 3:
-            debug('Stopped applying Code-Actions-On-Save due to reaching iteration limit')
-            self._on_complete()
-            return
         self._manager.documents.purge_changes(self._view)
         request_code_actions_on_save(self._view, self._handle_response, self._on_save_actions)
 
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -1,60 +1,81 @@
-import sublime
-import sublime_plugin
 from .core.edit import parse_workspace_edit
 from .core.protocol import Diagnostic
 from .core.protocol import Request, Point
 from .core.registry import LspTextCommand
+from .core.registry import LSPViewEventListener
 from .core.registry import sessions_for_view
+from .core.registry import windows
 from .core.settings import settings
 from .core.typing import Any, List, Dict, Callable, Optional, Union, Tuple, Mapping, TypedDict
 from .core.url import filename_to_uri
-from .core.views import region_to_range
+from .core.views import entire_content_range, region_to_range
 from .diagnostics import filter_by_point, view_diagnostics
+import sublime
+import sublime_plugin
 
 CodeActionOrCommand = TypedDict('CodeActionOrCommand', {
     'title': str,
     'command': Union[dict, str],
-    'edit': dict
+    'edit': dict,
+    'kind': Optional[str]
 }, total=False)
 CodeActionsResponse = Optional[List[CodeActionOrCommand]]
 CodeActionsByConfigName = Dict[str, List[CodeActionOrCommand]]
 
 
-class CodeActionsAtLocation(object):
+class CodeActionsCollector:
+    """
+    Collects code action responses from multiple sessions. Calls back the "on_complete_handler" with
+    results when all responses are received.
+    """
 
-    def __init__(self, on_complete_handler: Callable[[CodeActionsByConfigName], None]) -> None:
-        self._commands_by_config = {}  # type: CodeActionsByConfigName
-        self._requested_configs = []  # type: List[str]
+    def __init__(self, on_complete_handler: Callable[[CodeActionsByConfigName], None]):
         self._on_complete_handler = on_complete_handler
+        self._commands_by_config = {}  # type: CodeActionsByConfigName
+        self._request_count = 0
+        self._response_count = 0
+        self._all_requested = False
 
-    def collect(self, config_name: str) -> Callable[[CodeActionsResponse], None]:
-        self._requested_configs.append(config_name)
-        return lambda actions: self.store(config_name, actions)
+    def __enter__(self) -> 'CodeActionsCollector':
+        return self
 
-    def store(self, config_name: str, actions: CodeActionsResponse) -> None:
+    def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
+        self._all_requested = True
+        self._notify_if_all_finished()
+
+    def create_collector(self, config_name: str) -> Callable[[CodeActionsResponse], None]:
+        self._request_count += 1
+        return lambda actions: self._collect_response(config_name, actions)
+
+    def _collect_response(self, config_name: str, actions: CodeActionsResponse) -> None:
+        self._response_count += 1
         self._commands_by_config[config_name] = actions or []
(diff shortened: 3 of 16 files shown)


rchl:
I think this is largely ready.

I will still work on adding more tests but need to extend test server to handle some cases that I'm thinking of.

Don't fully like the organization of the new code. Feel free to suggest any improvements, and be picky. :)

plugin/core/sessions.py
```diff
@@ -153,7 +153,7 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
                 "dynamicRegistration": True,
                 "codeActionLiteralSupport": {
                     "codeActionKind": {
-                        "valueSet": []
+                        "valueSet": ["quickfix", "refactor"]
```
rchl, line 156: We can add all types here, I think. This didn't really make any visible difference but spec says:

> The symbol kind values the client supports. When this property exists the client also guarantees that it will handle values outside its set gracefully and falls back to a default value when unknown. If this property is not present the client only supports the symbol kinds from `File` to `Array` as defined in the initial version of the protocol.

so I don't see why we wouldn't just put all kinds here.

plugin/code_actions.py
```diff
@@ -63,52 +84,215 @@ def get_location_key(self, view: sublime.View, point: int) -> str:
...
+            if key.startswith('source.'):
+                allowed_code_actions[key] = value
+        return allowed_code_actions
+
+    def _trigger_native_save(self) -> None:
+        self._task = None
+        self.view.run_command('save', {"async": True})
+
+
+class CodeActionOnSaveTask():
```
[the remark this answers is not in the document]
rwols: Hmm. This entire class looks like an `asyncio.Task`.

plugin/code_actions.py
```diff
@@ -63,52 +84,215 @@ def get_location_key(self, view: sublime.View, point: int) -> str:
...
+        def actions_filter(actions: CodeActionsResponse) -> List[CodeActionOrCommand]:
+            return [a for a in (actions or []) if a.get('kind') in kinds]
+        collector = actions_collector.create_collector(config_name)
+        return (
+            lambda actions: collector(actions_filter(actions)),
+            lambda error: collector([])
+        )
+
+    actions_collector = CodeActionsCollector(actions_handler)
+    with actions_collector:
```
rwols, line 132: It took me a while to understand why we need to have this `with` statement and why we need the `__enter__` and `__exit__` magic methods.

LSP.sublime-settings
```diff
@@ -8,6 +8,22 @@
...
+  // Code action identifiers are not officially standardized so refer to specific
+  // server's documentation on what is supported but `source.fixAll` is commonly
+  // used to apply fix-on-save code action.
+  //
+  // It's also supported to set this in project-specific settings in which case
+  // it will only apply in that specific project. Setting it here will apply globally,
+  // unless overridden by project-specific settings.
+  //
+  // Only "source.*" actions are supported.
+  "lsp_code_actions_on_save": {},
```
rwols, line 22: If you want this to be a view setting, then it should not be here in LSP.sublime-settings. Instead tell users to put it in Preferences.sublime-settings and/or in their "settings" of their .sublime-project and/or in their syntax-specific override settings. It's the same kind of setting as lsp_format_on_save.

plugin/code_actions.py
```diff
@@ -63,52 +84,215 @@ def get_location_key(self, view: sublime.View, point: int) -> str:
...
+    If user has enabled a generic action that matches more specific session action
+    (for example user's a.b matching session's a.b.c), then the more specific (a.b.c) must be
+    returned as servers must receive only actions that they advertise support for.
+    """
+    matching_kinds = []
+    for session_action in session_actions:
+        enabled = False
+        action_parts = session_action.split('.')
+        for i in range(len(action_parts)):
+            current_part = '.'.join(action_parts[0:i + 1])
```
rwols, line 187: This looks like an O(N^2) algorithm?

Can we instead iterate over the parts directly?
```
keys = session_action.split('.')
for key in keys:
   ...

File: plugin/code_actions.py [hunk 6]
```diff
@@ -63,52 +84,215 @@ def get_location_key(self, view: sublime.View, point: int) -> str:
...
+            user_value = user_actions.get(current_part, None)
+            if isinstance(user_value, bool):
+                enabled = user_value
```
rwols, line 190: `break` here?

Keymaps/Default (Windows).sublime-keymap
```diff
@@ -62,4 +62,7 @@
 
     // Symbol Hover
     // {"keys": ["UNBOUND"], "command": "lsp_hover", "context": [{"key": "setting.lsp_active"}]},
+
+    // Override native save to handle Code-Actions-On-Save
+    { "keys": ["ctrl+s"], "command": "lsp_save", "context": [{ "key": "setting.lsp_active"}] },
```
rwols, line 67: It should also have code actions present in the view settings.

```
{"key": "setting.lsp_code_actions_on_save"}
```

plugin/code_actions.py
```diff
@@ -1,60 +1,81 @@
...
 }, total=False)
 CodeActionsResponse = Optional[List[CodeActionOrCommand]]
 CodeActionsByConfigName = Dict[str, List[CodeActionOrCommand]]
 
 
-class CodeActionsAtLocation(object):
+class CodeActionsCollector:
+    """
+    Collects code action responses from multiple sessions. Calls back the "on_complete_handler" with
+    results when all responses are received.
```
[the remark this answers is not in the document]
rwols: This is just a comment, let's perhaps do that for another PR.

plugin/code_actions.py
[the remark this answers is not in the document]
rchl: Even though I barely know `asyncio`, I've intended to implement it in a way that will make the transition to asyncio smooth.
(based mostly on my knowledge of Promises which is very much like asyncio tasks, I think).

plugin/code_actions.py
rchl: I could add some comments in the implementation of `__enter__` or the collector class itself.

The previous implementation relied solely on synchronous execution to know how many responses are expected. With `with` it's now explicit that when the context is exited, we have triggered all requests by then.

Also, for the same reason, previous implementation did not work at all for sync implementation of code actions.

LSP.sublime-settings
rchl: Actually the idea is that you can set it in LSP.sublime-settings (then it will apply globally) or you can use it in project-specific settings.

The `LspSaveCommand._get_code_actions_on_save` has logic to take all of those into account.

plugin/code_actions.py
rchl: We need index to be able to create a "cut out" of current "parts" (like take "a.b" from "a.b.c").

In any case, I don't think we should be concerned about performance here. This runs on a very small input and only once per session per save command.

plugin/code_actions.py
rchl: We have to iterate through all parts as next, a more specific part, might override that value.

Keymaps/Default (Windows).sublime-keymap
rchl: Yeah, can probably do that. Whether it's it would really make a difference depends on whether we'll keep the default to be an empty dict and what does empty dict evaluates to (True or False) in this context (I suppose it should work like Python so `{}` should be "Falsey" but will have to check).

LSP.sublime-settings
rwols: But if it's put into Packages/User/Preferences.sublime-settings, then it's also global simply via checking `view.settings().get("lsp_code_actions_on_save")`.

rchl: But then it makes that setting less discoverable?
Or how would you suggest to handle discoverability then?
I think that only having a mention in documentation is not enough.

plugin/code_actions.py
```diff
@@ -63,52 +84,215 @@ def get_location_key(self, view: sublime.View, point: int) -> str:
...
+                    session.send_request(request, actions_collector.create_collector(config_name))
+    return actions_collector
+
+
+def request_code_actions_on_save(
+    view: sublime.View,
+    actions_handler: Callable[[CodeActionsByConfigName], None],
+    on_save_actions: Dict[str, bool]
+) -> CodeActionsCollector:
+    def filtering_collector(
```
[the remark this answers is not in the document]
rchl: Done.

plugin/code_actions.py
```diff
@@ -63,52 +84,215 @@ def get_location_key(self, view: sublime.View, point: int) -> str:
...
+            lambda actions: collector(actions_filter(actions)),
+            lambda error: collector([])
+        )
+
+    actions_collector = CodeActionsCollector(actions_handler)
+    with actions_collector:
+        file_name = view.file_name()
+        if file_name:
+            sessions = sessions_for_view(view, 'codeActionProvider')
+            for session in sessions:
```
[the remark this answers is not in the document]
rchl: Done.

plugin/code_actions.py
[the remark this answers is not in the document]
rchl: Done.

plugin/code_actions.py
rchl: Added explanation in the documentation block.

rwols:
When `lsp_format_on_save` is True (in the .sublime-project for example), then I have to press CTRL+S twice in order to save the buffer?

rchl:
> I tried this locally. When `lsp_format_on_save` is True (in the .sublime-project for example), then I have to press CTRL+S twice in order to save the buffer?

Nicely spotted. I don't have time to check but I assume what happens in that case is:
 - CAOS runs first, does its job and calls native save
 - `FormatOnSaveListener.on_pre_save` triggers and does the sync formatting.

Not sure why it would act as you've described. Is formatting reverting some changes maybe?

In any case, sync handling "sucks". Should we convert it to using async also (kinda out of scope of this task I guess).

rchl:
Which server supported `textDocumentSync.willSaveWaitUntil` BTW?

rwols:
> In any case, sync handling "sucks".

Yes :(

> Should we convert it to using async also (kinda out of scope of this task I guess)?

If you feel like it, do it in this PR. A feature should, after all, not break other existing features :)

rwols:
> Which server supported documentFormattingProvider BTW?

The server used was pyls.

rchl:
> I tried this locally. When `lsp_format_on_save` is True (in the .sublime-project for example), then I have to press CTRL+S twice in order to save the buffer?

Can't reproduce that in pyls...

rwols:
Below you can see me pressing enter, then ctrl+s. This formats the view and removes the extra whitespace, but the view is not saved. I press ctrl+s again and it saves the view.

Peek 2020-06-04 17-47

rchl:
> Below you can see me pressing enter, then ctrl+s. This formats the view and removes the extra whitespace, but the view is not saved. I press ctrl+s again and it saves the view.

You have to help me investigate it as I can't reproduce. Maybe server logs will tell something?
Or I can try on Linux...

rwols:
When I log ST commands with `sublime.log_commands(True)`, I see this sequence of commands when I press ctrl+s once:

```
command: lsp_save
command: save {"async": true}
command: lsp_apply_document_edit {"changes": ...
```
So, the view is actually saved, but right after that, the changes for the formatting request are applied.

rchl:
> So, the view is actually saved, but right after that, the changes for the formatting request are applied.

The `save {"async": true}` command means that the command was triggered, not that the view was saved. That command is supposed to trigger native save and thus `on_pre_save` first, that would handle formatting in a synchronous way.

rchl:
> When I log ST commands with `sublime.log_commands(True)`, I see this sequence of commands when I press ctrl+s once:
> 
> ```
> command: lsp_save
> command: save {"async": true}
> command: lsp_apply_document_edit {"changes": ...
> ```
> 
> So, the view is actually saved, but right after that, the changes for the formatting request are applied.

At the end of `on_pre_save` this is called to apply changes:
```
view.run_command('lsp_apply_document_edit', {'changes': edits})
```
It looks like the view is saved right when `on_pre_save` handler ends and before `LspApplyDocumentEditCommand` runs.

That behavior is different from when `save` command is triggered natively. So that's not good news, because that means programmatically triggering `save` command changes semantics a bit.

tests/test_code_actions.py
```diff
@@ -30,35 +33,147 @@ def edit_to_lsp(edit: Tuple[str, Range]) -> Dict:
...
+        yield from self.await_message('textDocument/codeAction')
+        self.assertEquals(entire_content(self.view), 'const x = 1;')
+        self.assertEquals(self.view.is_dirty(), False)
+
+    def test_no_fix_on_non_matching_kind(self) -> Generator:
+        yield from self._setup_document_with_missing_semicolon()
+        initial_content = 'const x = 1'
+        self.view.run_command('lsp_save')
+        self.assertEquals(entire_content(self.view), initial_content)
+        self.assertEquals(self.view.is_dirty(), True)
```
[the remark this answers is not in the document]
rchl: This test tests that "nothing happens" so it's in the category of tricky tests. It finishes before the view is actually saved. I'll add a short timeout.

plugin/code_actions.py
```diff
@@ -63,52 +84,215 @@ def get_location_key(self, view: sublime.View, point: int) -> str:
...
+
+    def __init__(self, view: sublime.View, on_save_actions: Dict[str, bool], on_done: Callable[[], None]) -> None:
+        self._view = view
+        self._manager = windows.lookup(view.window())
+        self._on_save_actions = on_save_actions
+        self._on_done = on_done
+        self._completed = False
+        self._canceled = False
+        # TODO: For limiting number of iterations the task can run. Maybe not needed since there is a timeout?
+        self._iteration = 0
```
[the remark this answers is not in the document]
rchl: I will remove it for now. I don't see a reason to optimize prematurely.

I've initially added it because `lsp-stylelint` will never shut up and will always return code action when asked. It's just that the code action will not include any text edits. Initially, I've handled that with this iteration limit but then I found a better solution (check if the view has changes after running code actions).

Keymaps/Default (Windows).sublime-keymap
rchl: - This check would not evaluate to True when `lsp_code_actions_on_save` is set in LSP settings
- Even when having it in global preferences, it would still evaluate to False regardless if `lsp_code_actions_on_save` Dict is empty or not. I guess those condition only handle booleans.

tests/test_code_actions.py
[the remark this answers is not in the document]
rchl: Done

rchl:
BTW. We should wait with this until #3323 is fixed (or worked around by us) because this feature can do some damage otherwise (reversible damage but annoying none the less).

tests/test_code_actions.py
```diff
@@ -100,7 +100,8 @@ def test_no_fix_on_non_matching_kind(self) -> Generator:
         initial_content = 'const x = 1'
         self.view.run_command('lsp_save')
         self.assertEquals(entire_content(self.view), initial_content)
-        self.assertEquals(self.view.is_dirty(), True)
+        yield 100
+        self.assertEquals(self.view.is_dirty(), False)
```
rwols, line 104: Why not do `yield lambda: not self.view.is_dirty()` ?

plugin/code_actions.py
```diff
@@ -236,7 +236,8 @@ def _get_code_actions_on_save(self, view_settings: Union[dict, sublime.Settings]
 
     def _trigger_native_save(self) -> None:
         self._task = None
-        self.view.run_command('save', {"async": True})
+        # Triggered from set timeout to preserve original semantics of on_pre_save handling
+        sublime.set_timeout(lambda: self.view.run_command('save', {"async": True}))
```
rwols, line 239: This seems to have fixed the problem, nice work.

tests/test_code_actions.py
rchl: Done

LSP.sublime-settings
predragnikolic, line 22: Instead of just having an empty object for `lsp_code_actions_on_save`,
we could give a hint of how that object looks:
```
  "lsp_code_actions_on_save": {
    // "source.fixAll": true
  },
```

Keymaps/Default (Windows).sublime-keymap
rwols: I learned that my above suggestion only works for boolean settings, not dictionaries.

LSP.sublime-settings
rwols: `lsp_format_on_save` already works the way I describe, why do things differently for `lsp_code_actions_on_save`?

rchl: I would suggest changing how `lsp_format_on_save` works then maybe.

I'd say it's better this way as I wouldn't even know the `lsp_format_on_save` existed if you wouldn't just tell me.

It's nice to be able to see all settings LSP supports when looking at its preferences.

And also, if it's not in Preferences then PackageDev marks pref as non-existent when user puts it in his global Preferences.

rwols: I'm okay with changing how `lsp_format_on_save` works, as long as these two settings are consistent.

LSP.sublime-settings
```diff
@@ -19,7 +19,9 @@
   // unless overridden by project-specific settings.
   //
   // Only "source.*" actions are supported.
-  "lsp_code_actions_on_save": {},
+  "lsp_code_actions_on_save": {
+    // "source.fixAll": True
```
predragnikolic, line 23: `True -> true`  
Because you already used the word minor, this would be a patch :)

rchl: Done

LSP.sublime-settings
rchl: Done

plugin/formatting.py
```diff
@@ -27,7 +28,8 @@ def on_pre_save(self) -> None:
             self._purge_changes_if_needed()
             self._will_save_wait_until(session)
 
-        if self.view.settings().get("lsp_format_on_save"):
+        format_on_save = self.view.settings().get("lsp_format_on_save", None)
+        if (format_on_save is not None and format_on_save) or settings.lsp_format_on_save:
```
rchl, line 32: Actually this is not correct logic. View setting should override global always, regardless if true or false.

LSP.sublime-settings
```diff
@@ -8,6 +8,28 @@
   // Valid values are "never", "always" and "saved"
   "auto_show_diagnostics_panel": "saved",
 
+  // Run the server's formatProvider (if supported) on a document before saving.
+  // This option is also supported in syntax-specific settings and/or in project files.
```
rwols, line 12: and/or in project files -> and/or in the `"settings"` section of project files

LSP.sublime-settings
```diff
@@ -8,6 +8,28 @@
...
+  "lsp_format_on_save": false,
+
+  // A dictionary of code action identifiers that should be triggered on save.
+  //
+  // Code action identifiers are not officially standardized so refer to specific
+  // server's documentation on what is supported but `source.fixAll` is commonly
+  // used to apply fix-on-save code actions.
+  //
+  // This option is also supported in syntax-specific settings and/or in project files.
```
rwols, line 21: and/or in project files -> and/or in the `"settings"` section of project files

PR #1092, opened 2020-06-04
Implement textDocument/selectionRange

rwols:
An LSP alternative to Sublime's own "Expand Selection".

Peek 2020-06-04 21-09

### DIFF
--- a/Keymaps/Default (Linux).sublime-keymap
+++ b/Keymaps/Default (Linux).sublime-keymap
@@ -64,5 +64,5 @@
     // {"keys": ["UNBOUND"], "command": "lsp_hover", "context": [{"key": "setting.lsp_active"}]},
 
     // Expand Selection (a replacement for ST's "Expand Selection")
-    // {"keys": ["ctrl+shift+a"], "command": "lsp_expand_selection", "context": [{"key": "setting.lsp_active"}]},
+    // {"keys": ["UNBOUND"], "command": "lsp_expand_selection", "context": [{"key": "setting.lsp_active"}]},
 ]
--- a/Keymaps/Default (Linux).sublime-keymap
+++ b/Keymaps/Default (Linux).sublime-keymap
@@ -62,4 +62,7 @@
 
     // Symbol Hover
     // {"keys": ["UNBOUND"], "command": "lsp_hover", "context": [{"key": "setting.lsp_active"}]},
+
+    // Expand Selection (a replacement for ST's "Expand Selection")
+    // {"keys": ["ctrl+shift+a"], "command": "lsp_expand_selection", "context": [{"key": "setting.lsp_active"}]}
 ]
--- a/Keymaps/Default (Linux).sublime-keymap
+++ b/Keymaps/Default (Linux).sublime-keymap
@@ -64,5 +64,5 @@
     // {"keys": ["UNBOUND"], "command": "lsp_hover", "context": [{"key": "setting.lsp_active"}]},
 
     // Expand Selection (a replacement for ST's "Expand Selection")
-    // {"keys": ["ctrl+shift+a"], "command": "lsp_expand_selection", "context": [{"key": "setting.lsp_active"}]}
+    // {"keys": ["ctrl+shift+a"], "command": "lsp_expand_selection", "context": [{"key": "setting.lsp_active"}]},
 ]
--- a/Keymaps/Default (Linux).sublime-keymap
+++ b/Keymaps/Default (Linux).sublime-keymap
@@ -64,5 +64,5 @@
     // {"keys": ["UNBOUND"], "command": "lsp_hover", "context": [{"key": "setting.lsp_active"}]},
 
     // Expand Selection (a replacement for ST's "Expand Selection")
-    // {"keys": ["UNBOUND"], "command": "lsp_expand_selection", "context": [{"key": "setting.lsp_active"}]},
+    // {"keys": ["ctrl+shift+a"], "command": "lsp_expand_selection", "context": [{"key": "setting.lsp_active"}]},
 ]
--- a/plugin/selection_range.py
+++ b/plugin/selection_range.py
@@ -1,0 +1,54 @@
+from .core.protocol import Range
+from .core.protocol import Request
+from .core.registry import get_position
+from .core.registry import LspTextCommand
+from .core.sessions import method_to_capability
+from .core.typing import Any, Dict, Optional, List
+from .core.views import range_to_region
+from .core.views import selection_range_params
+import sublime
+
+
+class LspExpandSelectionCommand(LspTextCommand):
+    method = 'textDocument/selectionRange'
+    capability = method_to_capability(method)[0]
+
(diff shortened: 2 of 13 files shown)


Keymaps/Default (Linux).sublime-keymap
```diff
@@ -62,4 +62,7 @@
 
     // Symbol Hover
     // {"keys": ["UNBOUND"], "command": "lsp_hover", "context": [{"key": "setting.lsp_active"}]},
+
+    // Expand Selection (a replacement for ST's "Expand Selection")
+    // {"keys": ["ctrl+shift+a"], "command": "lsp_expand_selection", "context": [{"key": "setting.lsp_active"}]}
```
rchl, line 67: Also other platforms please :)

rwols:
It seems vscode-css and vscode-json are also selectionRangeProviders:

Peek 2020-06-04 23-26

rwols:
For JSON it's not much different or better than ST's built-in expand selection to be honest.

plugin/selection_range.py
```diff
@@ -0,0 +1,55 @@
...
+        self._regions = []  # type: List[sublime.Region]
+        self._change_count = 0
+
+    def run(self, edit: sublime.Edit, event: Optional[dict] = None) -> None:
+        session = self.session(self.capability, get_position(self.view, event))
+        if session:
+            params = selection_range_params(self.view)
+            self._regions.extend(self.view.sel())
+            self._change_count = self.view.change_count()
+            session.send_request(Request(self.method, params), self.on_result, self.on_error)
```
rchl, line 27: We need to trigger original key binding if there is no session for position. Otherwise expanding will just be broken when there is a session but it doesn't support expanding.

plugin/selection_range.py
```diff
@@ -0,0 +1,55 @@
...
+
```
rchl, line 28: We need to trigger original key binding if there is no session for position. Otherwise expanding will just be broken when there is a session but it doesn't support expanding.

plugin/selection_range.py
```diff
@@ -0,0 +1,55 @@
...
+        if self._change_count != self.view.change_count():
+            return
+        if params:
+            regions = [self._smallest_containing(region, param) for region, param in zip(self._regions, params)]
+        else:
+            regions = self._regions
+            window = self.view.window()
+            if window:
+                window.status_message("Nothing to expand")
+        self.view.run_command("lsp_selection_set", {"regions": [(r.a, r.b) for r in regions]})
```
rchl, line 39: If there is nothing to expand then it looks like we will clear regions. Shouldn't we just leave current selection intact?

plugin/selection_range.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def on_result(self, params: Any) -> None:
+        if self._change_count != self.view.change_count():
+            return
+        if params:
+            regions = [self._smallest_containing(region, param) for region, param in zip(self._regions, params)]
+        else:
+            regions = self._regions
+            window = self.view.window()
+            if window:
```
rchl, line 37: What's the reason for setting status on the window and not view? This is a view command after all.

Also, is that message ever cleared?
I guess it might be cleared by "something else" eventually but should we rely on that?

I guess that's another reason to use view status as then we know our key and can clear it reliably.

tests/test_single_document.py
```diff
@@ -253,6 +253,32 @@ def test_implementation(self) -> 'Generator':
...
+        self.assertEqual(self.view.substr(self.view.sel()[0].a), "c")
+        response = [{
+            "parent": {
+                "parent": {
+                    "range": {"start": {"line": 0, "character": 0}, "end": {"line": 0, "character": 5}}
+                },
+                "range": {"start": {"line": 0, "character": 1}, "end": {"line": 0, "character": 4}}
+            },
+            "range": {"start": {"line": 0, "character": 2}, "end": {"line": 0, "character": 3}}
+        }]
```
rchl, line 270: Can you also test expanding only "on one side"? So, for example, "start" stays the same but "ends" moves forward.

Maybe also test expanding across the lines?

plugin/selection_range.py
rwols: Hmm, I tried to address this but I learned that the `is_enabled` method from LspTextCommand prevents the keybinding to trigger. But it also masks the original `expand_selection` command in this way (you cannot trigger it with ctrl+shift+a anymore, because another command (lsp_expand_selection) masked it with its is_enabled method returning False). So I've added the fallback behavior, but it doesn't do anything at this point. Implementing #1093 will make things work better anyway.

plugin/selection_range.py
rwols: The `window.status_message` method also clears the message after a short while.

tests/test_single_document.py
rwols: > Can you also test expanding only "on one side"? So, for example, "start" stays the same but "ends" moves forward.

Done.

> Maybe also test expanding across the lines?

Doesn't matter, I'm using sublime.Region which doesn't work with the notion of lines altogether.

plugin/selection_range.py
rwols: done.

plugin/selection_range.py
rwols: I have removed the ctrl+shift+a suggestion from the default keybindings for now.

plugin/selection_range.py
rchl: Should I use the same for CAOS then, to be consistent? I admit it shows in a better place then view status which tends to be crammed on the left and shift everything.

plugin/selection_range.py
```diff
@@ -25,31 +25,38 @@ def run(self, edit: sublime.Edit, event: Optional[dict] = None) -> None:
             self._regions.extend(self.view.sel())
             self._change_count = self.view.change_count()
             session.send_request(Request(self.method, params), self.on_result, self.on_error)
+        else:
+            self._run_builtin_expand_selection("No {} found".format(self.capability))
```
rchl, line 29: This message is quite internal to the protocol. I feel user shouldn't need to see it.

plugin/selection_range.py
rchl: Although I see that this would not be ideal for the context menu.

plugin/selection_range.py
rwols: Actually I think users have to get comfortable with fooProvider and barProvider and such as they'll end up in the contexts of keybindings eventually after #1069 (unless we go with #1093) (or we can do both).

plugin/selection_range.py
rwols: Yes, it will not be grayed out in the context menu when we always return True from is_enabled.

rchl: It might be OK in fact. Sure, it wouldn't necessarily use the LSP to expand the selection, but if user has made the effort to look for it in the submenu, maybe he should just be able to expand, regardless if it's done by LSP or ST.

PR #1097, opened 2020-06-06
Fix document synchronization issue when restoring from session

### LINKED ISSUE #1096: View synchronization issue when modifying file during LSP initialization
1. Install `lsp-eslint` (you might also need proper eslint configuration...)
2. Restart ST with such document open
```js
const a = 1;
const b = 2;
const c = 3;
```
3. While LSP is initializing, join all three lines with `cmd+j` (or equivalent on non-macs).
4. Wait for LSP to initialize fully
5. Undo changes with cmd+z (or equivalent).

The diagnostics will be messed up.
Log:
 - `textDocument/didOpen`:
```const a = 1; const b = 2; const c = 3;\n```
 - `textDocument/didChange`:
```
  "contentChanges": [
    {
      "rangeLength": 1,
      "range": {
        "start": {
          "line": 0,
          "character": 25
        },
        "end": {
          "line": 1,
          "character": 0
        }
      },
      "text": ""
    },
    {
      "rangeLength": 0,
      "range": {
        "start": {
          "line": 0,
          "character": 25
        },
        "end": {
          "line": 0,
          "character": 25
        }
      },
      "text": " "
    }
  ]
```

This is already wrong as replies changes that no longer apply for the view that was sent in `didOpen`.

(Issue found by @predragnikolic )

### PR DESCRIPTION
**Pull request by rchl**
When restoring from session, and making some changes in the document
quickly after plugin initialized but before LSP initialized the server,
we would end up with some accumulated pending buffer changes from
"on_text_changed" listener that we would send after sending "didOpen"
notification. That broke view synchronization as it applied old changes
to the current state of the document send in "didOpen".

Resolves #1096

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -214,6 +214,8 @@ class WindowDocumentHandler(object):
 
     def _notify_did_open(self, view: ViewLike, session: Session) -> None:
         language_id = self._view_language(view, session.config.name)
+        # There might be pending, no longer valid changes when restoring session.
+        self._pending_buffer_changes.pop(view.buffer_id(), None)
         # mypy: expected sublime.View, got ViewLike
         session.send_notification(did_open(view, language_id))  # type: ignore
 

### CODE REVIEW

File: plugin/core/windows.py [hunk 1]
```diff
@@ -214,6 +214,8 @@ def handle_did_open(self, view: ViewLike) -> None:
 
     def _notify_did_open(self, view: ViewLike, session: Session) -> None:
         language_id = self._view_language(view, session.config.name)
+        # There might be pending, no longer valid changes when restoring session.
+        self._pending_buffer_changes.pop(view.buffer_id(), None)
```
rwols, line 218: This doesn't work when there is more than one session attached to the view.
The first initialized session will discard the changes here, which is OK.
The second initialized session will also discard the changes, which is not OK for the previously initialized session.

However, I'm OK with merging this. I'll add the proper fix in #1069.

rchl: The loop that iterates session runs synchronously so it shouldn't be possible for any `on_text_changed` notifications to come during that iteration.

rchl: The code in `notify_did_change` looks suspicious though as it will call `handle_did_open` when didOpen wasn't notified yet. So even though we'll clear the buffer, it will still be sent...

rwols: Let's say the file content is `aaa`.
The user removes the last character, turning it into `aa`. Pending changes: "remove `a`".
Now session1 initialized, and it removes the pending changes. It sends `aa` as the initial document state. OK.
The user removes another character, turning the buffer into `a`. Pending changes: "remove `a`".
session2 initialized, and it removes the pending changes. Now, session1 will never get the pending change and will think the document is still `aa`.

rwols: One possible fix is to forget about delayed/batched text changes and send them immediately (but only for the case that the server understands incremental text sync, of course).

rchl: Right, I assume sessions can't be added dynamically like that.

rwols: I'm okay with merging this anyway, but just be aware that there's still some subtleties left to consider.

PR #1100, opened 2020-06-07
Use session_for_view and sessions_for_view less

rwols:
Instead use methods of the LSPViewEventListener.


plugin/core/windows.py
```diff
@@ -350,11 +350,13 @@ def window(self) -> sublime.Window:
 
     def sessions(self, view: sublime.View, capability: Optional[str] = None) -> Generator[Session, None, None]:
         file_name = view.file_name() or ''
+        scope = view2scope(view)
         for sessions in self._sessions.values():
             for session in sessions:
-                if capability is None or capability in session.capabilities:
-                    if session.state == ClientStates.READY and session.handles_path(file_name):
-                        yield session
+                if session.config.match_document(scope):
```
rchl, line 356: Style nit: Could filter the list in the `for` loop already? I don't like a lot of nesting. :)

plugin/color.py
```diff
@@ -35,10 +34,7 @@ def on_activated_async(self) -> None:
             self.initialize()
 
     def initialize(self, is_retry: bool = False) -> None:
-        if not any(configurations_for_view(self.view)):
-            self.initialized = True  # no server enabled, re-open file to activate feature.
-        sessions = list(sessions_for_view(self.view, 'colorProvider'))
-        if sessions:
+        if self.session('colorProvider'):
```
rchl, line 37: Style nit: I dislike such generic method names as it's really hard to search for uses of it in the code (especially when refactoring code). Could we maybe call those `get_session` and `get_sessions`?

plugin/core/types.py
```diff
@@ -197,3 +197,4 @@ def views(self) -> List[ViewLike]:
 
     def run_command(self, command_name: str, command_args: Dict[str, Any]) -> None:
         ...
+        ...
```
rchl, line 200: Intentional?

plugin/color.py
rchl, line 37: The logic has changed here a bit. Previously `self.initialized = True` was set if no sessions were found. Probably not a big deal. Someone would have to activate file twice really quickly for this to make a difference.

plugin/core/types.py
rwols: No :) must have been pyls going crazy with format-on-save again

plugin/core/windows.py
rwols: I've put all of the checks in `can_handle` of `Session`. Hope that makes it a bit better. I don't personally mind nested ifs.

plugin/core/windows.py
```diff
@@ -349,14 +349,10 @@ def window(self) -> sublime.Window:
...
     def sessions(self, view: sublime.View, capability: Optional[str] = None) -> Generator[Session, None, None]:
-        file_name = view.file_name() or ''
-        scope = view2scope(view)
         for sessions in self._sessions.values():
             for session in sessions:
-                if session.config.match_document(scope):
-                    if capability is None or capability in session.capabilities:
-                        if session.state == ClientStates.READY and session.handles_path(file_name):
-                            yield session
+                if session.can_handle(view, capability):
```
rchl: Could still be filtered in the `for` loop. :)
`for session in [session for session in sessions if session.can_handle(view, capability)]:`

plugin/color.py
rchl: I've approved but I don't like it.

rwols: I'll do it in another PR. This is currently more in line with the `self.session` from LspTextCommand.

PR #1102, opened 2020-06-07
Refactor "debouncing" into a function

rwols:
There's a bunch of places where we run a function at a later point
in time, but only when a certain stability condition holds.

### DIFF
--- a/plugin/color.py
+++ b/plugin/color.py
@@ -2,12 +2,13 @@ import sublime
 from .core.documents import is_transient_view
 from .core.protocol import Range
 from .core.protocol import Request
-from .core.registry import session_for_view, sessions_for_view, configurations_for_view
 from .core.registry import LSPViewEventListener
+from .core.registry import session_for_view, sessions_for_view, configurations_for_view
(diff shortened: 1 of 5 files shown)


plugin/color.py
```diff
@@ -2,12 +2,13 @@
...
 from .core.protocol import Range
 from .core.protocol import Request
-from .core.registry import session_for_view, sessions_for_view, configurations_for_view
 from .core.registry import LSPViewEventListener
+from .core.registry import session_for_view, sessions_for_view, configurations_for_view
 from .core.settings import settings
 from .core.typing import List, Dict, Optional
 from .core.url import filename_to_uri
 from .core.views import range_to_region
+from .core.windows import debounced
```
rchl, line 11: Not sure "windows.py" is the best place for it but don't have a better suggestion than a generic "utility".

rwols: I wanted to put it in registry.py, but then I got a circular import problem. And having another source file just for a function is a bit overkill.

PR #1109, opened 2020-06-09
Support selection ranges when requesting code actions

### LINKED ISSUE #1085: Request code actions at the cursor/selection when the selection is modified
The [rust-analyzer language server](https://rust-analyzer.github.io/manual.html#assists-code-actions) makes use of VSCode's behavior to request "diagnostic-less" code actions at the cursor whenever it changes. [A user reported this on discord](https://discord.com/channels/280102180189634562/645268178397560865/717091710739546182). We should mirror the behavior. This has already been observed and discussed [here](https://github.com/sublimelsp/LSP/pull/779#issuecomment-557704533).

Because code-actions returned from this type of request do not belong to any diagnostics, we may have to adjust the internal design of the code a little for this feature. For example, we should invalidate those type of code actions whenever the buffer is modified. Most likely we can re-use the logic present in highlights.py or color.py.

Also, the user cannot "see" these code actions unless they turned on the gutter bulb.

rchl:
- When requesting code actions using keyboard action, take whole selection
  range into account rather than just selection end point.
- Move code action requesting code into CodeActionsManager class to have
  one common endpoint. `get_matching_kinds` is left outside because
  it's imported by tests (but technically could also move it inside).
- (WIP) Trigger code actions on selection changes. This uses new ST4
  annotations feature to show the number of available code actions. It's
  better than bulb as bulb can't be clicked. The underline of the link
  looks a bit wrong in the annotation (seems like an ST bug). Also, it's
  not clear if we really want this feature. Maybe if we optimize it to
  not trigger for *all* selection changes (so not when typing).
- I guess we could support all selections for this feature instead of
  only the first one.

## Todo
- [x] Write tests

Resolves #1085


plugin/code_actions.py
```diff
@@ -323,14 +351,19 @@ def on_selection_modified_async(self) -> None:
...
+        diagnostics_by_config = filter_by_range(view_diagnostics(view), stored_range)
+        actions_manager.request_with_optional_diagnostics(view, diagnostics_by_config, self.handle_responses)
 
     def handle_responses(self, responses: CodeActionsByConfigName) -> None:
-        for _, items in responses.items():
-            self._actions.extend(items)
-        if len(self._actions) > 0:
-            self.show_bulb()
+        self._actions_by_config = responses
+        if any((action for action in self._actions_by_config.values())):
```
rwols, line 363: `if any(self._actions_by_config.values())` ?

plugin/code_actions.py
```diff
@@ -323,14 +351,19 @@ def on_selection_modified_async(self) -> None:
...
+            self.show_annotations()
+            if settings.show_code_actions_bulb:
+                self.show_bulb()
```
rwols, line 366: Shouldn't there be an `else` case here that removes bulbs and/or annotations in case the result params are empty?

plugin/code_actions.py
```diff
@@ -340,6 +373,17 @@ def show_bulb(self) -> None:
     def hide_bulb(self) -> None:
         self.view.erase_regions('lsp_bulb')
 
+    def show_annotations(self) -> None:
```
rwols, line 376: Things in the gutter are hoverable. Look at GitGutter for example. Perhaps we can use that to hover over this lightbulb and show code actions?

I was thinking to use annotations for diagnostics -- and remove the "active diagnostic" in the status bar to unclutter it. Not all servers use code actions, so reserving that feature just for code actions seems like a waste.

plugin/code_actions.py
```diff
@@ -382,7 +426,11 @@ class LspCodeActionsCommand(LspTextCommand):
     def run(self, edit: sublime.Edit, event: Optional[dict] = None) -> None:
         self.commands = []  # type: List[Tuple[str, str, CodeActionOrCommand]]
         self.commands_by_config = {}  # type: CodeActionsByConfigName
-        actions_manager.request(self.view, self.handle_responses, self.view.sel()[0].begin())
+        view = self.view
+        region = view.sel()[0]
```
rwols, line 430: Put this in a `try: ... except IndexError: ... ` block.

plugin/code_actions.py
```diff
@@ -340,6 +373,17 @@ def show_bulb(self) -> None:
     def hide_bulb(self) -> None:
         self.view.erase_regions('lsp_bulb')
 
+    def show_annotations(self) -> None:
+        flags = sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE
+        actions_count = sum([len(actions) for actions in self._actions_by_config.values()])
```
rwols, line 378: `actions_count = sum(map(len, self._actions_by_config.values()))` ?

plugin/code_actions.py
rchl: But those are removed when starting the request at `on_selection_modified_async`.

plugin/code_actions.py
rchl: > Perhaps we can use that to hover over this lightbulb and show code actions?

So how would the flow look like?
 - The bulb is hovered, which shows a minihtml popup with "3 Code actions"? And then clicking it reveals the native popup?
 - Or on hovering the bulb we show a minihtml popup with already populated lists of code actions?

> I was thinking to use annotations for diagnostics

I guess it's a separate discussion but I think that might not be very practical as there is limited space (at least for the annotations added by `add_regions`.

rwols: Ah, the horizontal spacing is limited? Then I suppose it's suitable only for telling the user there exist code actions.

But if we have the annotations show up, perhaps we can remove the bulb? It would otherwise be double information.

rwols:
I would rather see merge commits than rebased commits. It's hard to see what is changed. And github gets confused with rebased commits. For small PRs it's okay to do rebases but for larger stories please use merges.

plugin/code_actions.py
rchl: I would use the gutter icon for everything if it would be possible to make gutter icon clickable. There is a way to handle that using some hack (https://github.com/daveleroy/sublime_debugger/issues/68) but it's not cool (we don't really know the position or extend of the icon reliably).

So I'm up for using those annotations, removing the bulb and the `show_code_actions_bulb` setting.

rchl:
> I would rather see merge commits than rebased commits. It's hard to see what is changed. And github gets confused with rebased commits. For small PRs it's okay to do rebases but for larger stories please use merges.

Yes, I try to avoid both during PR. In this case, I've just updated the commit message so nothing really changed.

plugin/code_actions.py
```diff
@@ -323,23 +351,39 @@ def on_selection_modified_async(self) -> None:
...
+                              'Packages/LSP/icons/lightbulb.png', flags)
 
     def hide_bulb(self) -> None:
         self.view.erase_regions('lsp_bulb')
 
+    def show_annotations(self) -> None:
+        flags = sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE
+        actions_count = sum(map(len, self._actions_by_config.values()))
+        code_actions_link = make_link('subl:lsp_code_actions', '{} code action(s)'.format(actions_count))
+        self.view.add_regions('lsp_action_annottations', [self._stored_region], flags=flags,
```
rwols, line 380: lsp_action_annottations -> lsp_action_annotations

plugin/code_actions.py
```diff
@@ -323,23 +351,39 @@ def on_selection_modified_async(self) -> None:
...
+                              annotations=["<div class=\"actions\">{}</div>".format(code_actions_link)],
+                              annotation_color='yellow')
```
rwols, line 382: Can we use `var(--yellowish)` instead? Or `region.yellowish` ?

plugin/code_actions.py
```diff
@@ -323,23 +351,39 @@ def on_selection_modified_async(self) -> None:
...
+
+    def clear_annotations(self) -> None:
+        self.view.erase_regions('lsp_action_annottations')
```
rwols, line 385: lsp_action_annottations -> lsp_action_annotations

plugin/code_actions.py
rchl: None of those work.

rchl:
> I think the bulb is more suited for code actions than the annotations. The bulb is used in almost any editor/IDE to represent code actions.

The annotations are growing on me because I can see more information with it without really being distracted much.

I think that given how poor the gutter icon handling is in ST today, it doesn't make sense to use it.
When user sees it, he instinctively tries to click it and it does something very unexpected instead - changes selection. It's a bit of a disaster form UX perspective.
Other editors have better support for those.

rchl:
> Perhaps we can do both: show code actions and diagnostics as annotations. Because those diagnostic-less code actions appear on the same line as the caret “at worst”.

Not sure I get it.
Show diagnostics as annotations?
Often there is not much space for annotations - it depends on the length of the current line. So it would be impossible to really see the diagnostics properly in those cases. You would need to grab a mouse and hover the annotation.

rwols:
Yes: show diagnostics as annotations. Default/exec.py does it too, so they’re apparently supposed to be used for errors/warnings.

rchl:
I'm working on having at least one test for this but we don't really have a way to verify what was send in the request when using test server. So either we can skip the tests or I'll have to add something to test server.

rwols:
In a test, maybe you can use
```python
yield lambda: len(self.view.get_regions("lsp_action_annotations")) > 0
```
to "await" the moment when the code actions are visible? See tests/test_server_notifications.py for an example with diagnostics.

rchl:
But in this case I think I want to test something else. I want to assert that the range that was requested in `textDocument/codeAction` request matches document selection.

Checking for ranges won't help me as the ranges will be set to whatever I return in the response for that request. There is no validation for ranges at that point, we assume that server returned code actions filtered for given range.

rchl:
- There already is functionality to get the request's params, actually. Just need a small change to expose it.
- There is a bug with the code - when there are diagnostics in the range then the request has a range of the first diagnostics instead of whole selection.

rchl:
Rebased without any changes because I need new functions

rwols:
Again, I really prefer merges over rebases.

rchl:
But can github rebase onto one commit on merging? Because if not then merges have that minus that history looks like crap afterwise.

rchl:
I consider this ready.

PR #1111, opened 2020-06-11
Use the range formatter if there's no full range formatter

### LINKED ISSUE #1106: When there's only a documentRangeFormattingProvider, do "full" formatting via that capability
It can be the case that the server advertises only as documentRangeFormattingProvider, but not as documentFormattingProvider.

In that case, still allow the user to format the entire file simply by passing the entire range of the view to the range formatter.

see also https://github.com/microsoft/vscode/issues/99513

rwols:
close #1106

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -151,12 +151,16 @@ def text_document_identifier(view: sublime.View) -> Dict[str, Any]:
     return {"uri": uri_from_view(view)}
 
 
+def entire_region(view: sublime.View) -> sublime.Region:
+    return sublime.Region(0, view.size())
+
+
 def entire_content(view: sublime.View) -> str:
-    return view.substr(sublime.Region(0, view.size()))
+    return view.substr(entire_region(view))
 
 
 def entire_content_range(view: sublime.View) -> Range:
-    return region_to_range(view, sublime.Region(0, view.size()))
+    return region_to_range(view, entire_region(view))
 
 
 def text_document_item(view: sublime.View, language_id: str) -> Dict[str, Any]:
--- a/plugin/formatting.py
+++ b/plugin/formatting.py
@@ -5,7 +5,10 @@ from .core.registry import LSPViewEventListener
 from .core.sessions import Session
 from .core.settings import settings
 from .core.typing import Any, List, Optional
-from .core.views import will_save_wait_until, text_document_formatting, text_document_range_formatting
+from .core.views import entire_region
+from .core.views import text_document_formatting
+from .core.views import text_document_range_formatting
+from .core.views import will_save_wait_until
 
 
 def apply_response_to_view(response: Optional[List[dict]], view: sublime.View) -> None:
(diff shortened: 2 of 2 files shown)


plugin/formatting.py
```diff
@@ -59,13 +59,23 @@ class LspFormatDocumentCommand(LspTextCommand):
...
+        return super().is_enabled() or bool(self.session(LspFormatDocumentRangeCommand.capability))
+
     def run(self, edit: sublime.Edit, event: Optional[dict] = None) -> None:
         session = self.session(self.capability)
-        file_path = self.view.file_name()
-        if session and file_path:
-            session.send_request(
-                text_document_formatting(self.view),
-                lambda response: apply_response_to_view(response, self.view))
+        if session:
```
rchl, line 67: Why no `file_name()` check anymore?

plugin/formatting.py
```diff
@@ -59,13 +59,23 @@ class LspFormatDocumentCommand(LspTextCommand):
...
+            # Either use the documentFormattingProvider ...
+            session.send_request(text_document_formatting(self.view), self.on_result)
+        else:
+            session = self.session(LspFormatDocumentRangeCommand.capability)
+            if session:
+                # ... or use the documentRangeFormattingProvider and format the entire range.
+                req = text_document_range_formatting(self.view, sublime.Region(0, self.view.size()))
```
rchl, line 74: We could change `entire_content_range` to `entire_content_region` and use it here.
One place that uses `entire_content_range` could be changed to convert the region with `region_to_range`.

rchl:
Nice PR number

plugin/formatting.py
rwols: That's already taken care of in `self.session(...)`.

plugin/core/views.py
```diff
@@ -151,12 +151,16 @@ def text_document_identifier(view: sublime.View) -> Dict[str, Any]:
...
+def entire_region(view: sublime.View) -> sublime.Region:
+    return sublime.Region(0, view.size())
+
+
 def entire_content(view: sublime.View) -> str:
-    return view.substr(sublime.Region(0, view.size()))
+    return view.substr(entire_region(view))
 
 
 def entire_content_range(view: sublime.View) -> Range:
```
rchl, line 162: Would be more accurate with `_region` suffix.

plugin/core/views.py
```diff
@@ -151,12 +151,16 @@ def text_document_identifier(view: sublime.View) -> Dict[str, Any]:
     return {"uri": uri_from_view(view)}
 
 
+def entire_region(view: sublime.View) -> sublime.Region:
```
rchl, line 154: Kinda goes out-of-line with naming? `entire_content_region` would match equivalent function better?

rwols: Renamed in 62cf41e3b17a91316772fa416a061965a2bd19a9

PR #1113, opened 2020-06-11
Fix debouncing equality check on making selection

rchl:
On extending selection "sel().begin()" point does change as it denotes
beginning of selection, not end. So the check that compares current point
with stored point would always be True.

That resulted in trigger many many requests on making a selection as
all debounced request would go through.

### DIFF
--- a/plugin/color.py
+++ b/plugin/color.py
@@ -46,7 +46,7 @@ class LspColorListener(LSPViewEventListener):
             sel = self.view.sel()
             if len(sel) < 1:
                 return
-            current_point = sel[0].begin()
+            current_point = sel[0].b
             if self._stored_point != current_point:
                 self._stored_point = current_point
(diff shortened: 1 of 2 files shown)


rchl:
There is also another issue with debouncer - I would expect it to batch denounce requests and drop all but the latest (regardless of the lambda function check). That would require it to be a class that is instantiated and stored on the instance, I guess.

Why? Because we can still get handler triggered multiple times if we move the cursor left and right a couple of times quickly and end up at the initial location at the end. If you know what I mean.

rwols:
> There is also another issue with debouncer - I would expect it to batch denounce requests and drop all but the latest (regardless of the lambda function check). That would require it to be a class that is instantiated and stored on the instance, I guess.

Yes. There's also the problem that these debounced functions keep the object alive after it should have died (e.g. when the user closes the view, all pending functions for highlight / code actions / doc sync should be dropped).

PR #1117, opened 2020-06-12
Add release scripts from GitGutter

rchl:
I have tried the Build command locally and it worked. The new commit and tag were created.

The build command takes the latest version file from `/messages/*` and creates build based on that so the file should be created first.

Haven't tried the Release command obviously.

BTW. The script stores the new version number in `VERSION` file. We have our own version number in `sessions.py` so we should sync those. Maybe have a `version.py` file somewhere that the script would update?


rchl:
Messed up by including some local changes. Will try again.

PR #1118, opened 2020-06-12
Add release scripts from GitGutter

rchl:
I have tried the Build command locally and it worked. The new commit and tag were created.

The build command takes the latest version file from /messages/* and creates build based on that so the file should be created first.

Haven't tried the Release command obviously.

BTW. The script stores the new version number in VERSION file. We have our own version number in sessions.py so we should sync those. Maybe have a version.py file somewhere that the script would update?


release.py
```diff
@@ -0,0 +1,203 @@
...
+
+        def is_visible(self) -> bool:
+            settings = sublime.load_settings(SETTINGS)
+            return settings.has('github_token')
+
+        def run(self) -> None:
+            """Built a new release."""
+            build_release()
+
+    InternalBuildReleaseCommand.__name__ = '{}BuildReleaseCommand'.format(camel_case_prefix)
```
rwols, line 187: Why do it like this?

rchl: For reusability. I'd like to use this script in LSP packages also and having to change this code for each of them is cumbersome. It's easier if all configuration is at the top.

To be fair, it would be even better if the script would read an external configuration file instead so it doesn't have to be adjusted at all.

rwols: I see. I think a json configuration file would be better suited for this.

makefile
```diff
@@ -0,0 +1,12 @@
...
+
+release:
+	build_release
+	./release publish --token `cat $HOME/.github_access_token`
+
+build_docs:
+	mkdocs
+
+publish_docs:
+	mkdocs gh-deploy
```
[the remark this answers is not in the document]
rchl: Done all three.

release.py
rchl: Agreed, will do that.

Makefile
```diff
@@ -0,0 +1,12 @@
+build_release:
+	python3 ./release.py build
+
+release:
+	build_release
```
rwols, line 5: Should this not be in the prerequisites of the `release` target?

```make
release: build_release
    python3 ./release.py publish ...
```

All in all I think this makefile is pretty dumb. There's no mechanism to tell whether a target's prerequisites have already been satisfied. That's the whole point of Make. And Make is not installed on windows by default. So it can just as well be a bash script.

rchl: Agreed. Removed.

release.py
rchl: BTW. Alternatively, I could just remove those commands. The shell script should be enough really.

rwols: I agree the shell scripts should be enough :)

rchl:
Hmm, should move the script so it's not picked up by ST as a plugin...

PR #1120, opened 2020-06-14
Make log_server setting work again and remove log_payloads

rchl:
This setting was not used anywhere, instead log_debug controlled logging
in the panel.

Change default to False and make it control the log panel logging.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -1,6 +1,6 @@
 ## Self-help instructions
 
-Enable LSP logging: In LSP Settings enable the `log_debug` setting and `log_payloads` if needed.
+Enable LSP logging: In LSP Settings enable the `log_debug` setting.
 Enable server logging: set `log_server` and `log_stderr` to `true`
 Restart Sublime and open the console (``ctrl+` ``) to see the additional logging.
 If you believe the issue is with this package, please include the output from the Sublime console in your issue report!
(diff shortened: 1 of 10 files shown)


docs/troubleshooting.md
```diff
@@ -1,6 +1,6 @@
 ## Self-help instructions
 
-Enable LSP logging: In LSP Settings enable the `log_debug` setting and `log_payloads` if needed.
+Enable LSP logging: In LSP Settings enable the `log_debug` setting.
 Enable server logging: set `log_server` and `log_stderr` to `true`
 Restart Sublime and open the console (``ctrl+` ``) to see the additional logging.
```
rwols, line 5: This should probably say: Run LSP: Toggle Log Panel from the command palette. No restart is needed.

plugin/core/windows.py
```diff
@@ -666,7 +666,7 @@ def __init__(self, manager: WindowManager, server_name: str) -> None:
         self._manager = ref(manager)
         self._server_name = server_name
 
-    def log(self, message: str, params: Any, log_payload: bool) -> None:
+    def log(self, message: str, params: Any, log_payload: Optional[bool] = True) -> None:
```
rwols, line 669: The type of log_payloads should be `bool` and its default value should be `False`.

rchl: Why default `False`? Then we have to pass `True` in most cases where now we can just skip it.
Do you want thing more explicit?

rwols: Right, sorry, I think I need more coffee. I only have a problem with the Optional. That makes the type have 3 values whereas 2 possible values should be sufficient. The default should stay `True`.

PR #1121, opened 2020-06-14
Create a websocket logger for using with external log parsing tool

rchl:
- Change log_server setting from boolean to a list of logger types (boolean still supported for backward-compatibility)
- Use a simple, third party websocket server. The only one I found that didn't require async / coroutines support.

It can be tested together with https://lsp-log-parser.netlify.app/ (which hopefully is self-explanatory).

Note: There is no limit on the number of messages that the browser app can receive so it will just keep collecting all and memory usage can get out of hand probably, if left unattended. :)

Also, I think I've gotten ST to kinda freeze temporarily when typing a lot and sending a lot of messages quickly but haven't reproduced that lately. I need to stress-test it a bit more and verify that I'm using the server correctly.


rchl:
Updated linters on CI to latest versions as it was hard to fix issues when different ones were reported locally.

mypy.ini
```diff
@@ -8,3 +8,7 @@ mypy_path = stubs
 [mypy-tests]
 check_untyped_defs = True
 disallow_untyped_defs = False
+
+[mypy-third_party.*]
+ignore_errors = True
+ignore_missing_imports = True
```
rchl, line 14: This was supposed to help for linting issue when importing the module but it didn't so I had to add ` # type: ignore` there anyway.

rwols:
When I set server_log to ["panel", "remote"], and then open your link, and then press CTRL+D, I get this:

Screenshot from 2020-06-15 17-16-53

rchl:
That's normal state for when the server is not started or can't be connected to. I can try on Linux later in case it's something Linux specific.

rchl:
Btw log_server, not server_log but I guess it was just a typo.

rwols:
There's no indication that something is not starting from ST's console or the server log panel. So I'm kind of at a loss on how to debug this.

rchl:
A silly mistake during last-minute refactoring. Will push a fix in a minute.

A debug message is printed for it though:
```
LSP: Invalid logger type (remote) specified for log_server settings
```

Might be hard to see in the sea of other useless debug messages ;)

plugin/core/windows.py
```diff
@@ -739,3 +756,145 @@ def _format_request(self, direction: str, method: str, request_id: Any) -> str:
...
+                    raise ex
+        self._start_server()
+
+    def _start_server(self) -> None:
+        if RemoteLogger._ws_server:
+            def start_async() -> None:
+                if RemoteLogger._ws_server:
+                    RemoteLogger._ws_server.run_forever()
+            RemoteLogger._ws_server_thread = threading.Thread(target=start_async)
+            RemoteLogger._ws_server_thread.start()
```
rwols, line 791: I am allergic to threads. This seems to be starting a thread for every logger, in other words for every session. Is it not the intention that there is only one thread running for the entire application instead?

plugin/core/windows.py
```diff
@@ -739,3 +756,145 @@ def _format_request(self, direction: str, method: str, request_id: Any) -> str:
...
+
+class RemoteLogger(Logger):
+    PORT = 9981
+    DIRECTION_OUTGOING = 1
+    DIRECTION_INCOMING = 2
+    _ws_server = None  # type: Optional[WebsocketServer]
+    _ws_server_thread = None  # type: Optional[threading.Thread]
+
+    def __init__(self, manager: WindowManager, server_name: str) -> None:
+        self._manager = ref(manager)
```
rwols, line 769: Where is this member variable used?

plugin/core/windows.py
```diff
@@ -739,3 +756,145 @@ def _format_request(self, direction: str, method: str, request_id: Any) -> str:
...
+
+    def _start_server(self) -> None:
+        if RemoteLogger._ws_server:
+            def start_async() -> None:
+                if RemoteLogger._ws_server:
+                    RemoteLogger._ws_server.run_forever()
+            RemoteLogger._ws_server_thread = threading.Thread(target=start_async)
+            RemoteLogger._ws_server_thread.start()
+
+    def _stop_server(self) -> None:
```
rwols, line 793: Should this not be called in a `__del__` somewhere?

third_party/websocket_server/websocket_server.py
```diff
@@ -0,0 +1,370 @@
...
+        GUID = '258EAFA5-E914-47DA-95CA-C5AB0DC85B11'
+        hash = sha1(key.encode() + GUID.encode())
+        response_key = b64encode(hash.digest()).strip()
+        return response_key.decode('ASCII')
+
+    def finish(self):
+        self.server._client_left_(self)
+
+
+def encode_to_UTF8(data):
```
rwols, line 353: Note to self: could re-use `_encode` and `_decode` functions from plugin/core/transports.py here.

plugin/core/sessions.py
```diff
@@ -428,7 +428,7 @@ def get_plugin(name: str) -> Optional[Type[AbstractPlugin]]:
 
 class Session(Client):
 
-    def __init__(self, manager: Manager, logger: Logger, workspace_folders: List[WorkspaceFolder],
+    def __init__(self, manager: Manager, loggers: List[Logger], workspace_folders: List[WorkspaceFolder],
```
rwols, line 431: I'm not a fan of this list-of-loggers approach :/ We could just as well have a ListLogger class that delegates all of the methods to all of its subloggers.

(untested)
```python
class ListLogger(Logger):
    def __init__(self) -> None:
        self._loggers = []  # type: List[Logger]

    def append(self, logger: Logger) -> None:
        self._loggers.append(logger)

    def _foreach(self, method: str, *args: Any) -> None:
        for logger in self._loggers:
            getattr(logger, method)(*args)

    def outgoing_response(self, *args: Any) -> None:
        self._foreach("outgoing_response", *args)

    def outgoing_error_response(self, *args: Any) -> None:
        self._foreach("outgoing_error_response", *args)

    def outgoing_request(self, *args: Any) -> None:
        self._foreach("outgoing_response", *args)

    def outgoing_notification(self, *args: Any) -> None:
        self._foreach("outgoing_notification", *args)

    def incoming_response(self, *args: Any) -> None:
        self._foreach("incoming_response", *args)

    def incoming_request(self, *args: Any) -> None:
        self._foreach("incoming_request", *args)

    def incoming_notification(self, *args: Any) -> None:
        self._foreach("incoming_notification", *args)
```

Also untested _create_loggers modification:
```python
    def _create_loggers(self, config_name: str) -> Logger:
        logger_map = {
            "panel": PanelLogger,
            "remote": RemoteLogger,
        }
        loggers = []
        for logger_type in settings.log_server:
            if logger_type not in logger_map:
                debug("Invalid logger type ({}) specified for log_server settings".format(logger_type))
                continue
            loggers.append(logger_map[logger_type])
        if len(loggers) == 0:
            return ListLogger()  # logs nothing
        elif len(loggers) == 1:
            return loggers[0](self, config_name)
        else:
            logger = ListLogger()
            for logger_type in loggers:
                logger.append(logger_type(self, config_name))
            return logger
```

plugin/core/windows.py
rchl: `_ws_server` is a (static) class property so it's only initialized once.

plugin/core/windows.py
rchl: It probably isn't.

plugin/core/windows.py
rwols: But `_start_server` is always called in `__init__` of `RemoteLogger`? Should that not be behind an if-condition?

plugin/core/windows.py
rchl: Since it's initalialized once per however many sessions, it can't be stopped on any particular session being destroyed. It could only be stopped when all active sessions have been stopped but I'm not sure how to hook into that. Probably would require your session changes anyway?

plugin/core/sessions.py
rchl: To be clear, uou are not a fan of this implementation but you are OK with user-facing changes to `log_server` setting?

I'm OK with your suggestion although all I think of when seeing it is "more lines of code".

plugin/core/windows.py
rchl: The `if` condition could be there but `if RemoteLogger._ws_server:` check here makes sure it works like that. I can move the check there, I agree it makes sense.

plugin/core/sessions.py
rwols: Apologies for the confusion. I'm certainly OK with user-facing changes to log_server. The remote logging is pretty cool. I just don't agree with changing Session and Client for this. My proposed changes would keep Client as-is. It's perhaps more lines of code, but the functionality (doing stuff for a list-of-loggers) is contained in a single object.

plugin/core/windows.py
rchl: Removed.

(That also reminded me that I need to add handling for logging server's stderr.

plugin/core/windows.py
rchl: Done

rchl:
reverted the merge. Will re-do.

PR #1124, opened 2020-06-15
Allow registering notification/request handlers through AbstractPlugin

rchl:
So that the users of AbstractPlugin don't have to mess with internal
method2attr function themselves.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -6,11 +6,12 @@ from .protocol import TextDocumentSyncKindNone, TextDocumentSyncKindIncremental,
 from .protocol import WorkspaceFolder, Request, Notification, Response
 from .rpc import Client
 from .rpc import Logger
+from .rpc import method2attr
 from .settings import client_configs
 from .transports import Transport
 from .types import ClientConfig, ClientStates
 from .types import view2scope
-from .typing import Dict, Any, Optional, List, Tuple, Generator, Type
+from .typing import Callable, Dict, Any, Optional, List, Tuple, Generator, Type
 from .version import __version__
 from .views import COMPLETION_KINDS
 from .views import did_change_configuration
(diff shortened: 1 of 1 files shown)


rwols:
This is not the way to do it.

To quote the documentation string:
```
    Given a request/notification, replace the non-alphabetic characters with an underscore, and prepend it with "m_".
    This will be the name of your method.
    For instance, to implement the non-standard eslint/openDoc request, define the Python method

        def m_eslint_openDoc(self, params, request_id):
            session = self.weaksession()
            if session:
                webbrowser.open_tab(params['url'])
                session.send_response(Response(request_id, None))

    To handle the non-standard eslint/status notification, define the Python method

        def m_eslint_status(self, params):
            pass
```

Some examples from the Session class:

```python
    def m_workspace_applyEdit(self, params: Any, request_id: Any) -> None:
        """handles the workspace/applyEdit request"""
        edit = params.get('edit', {})
        self.window.run_command('lsp_apply_workspace_edit', {'changes': parse_workspace_edit(edit)})
        # TODO: We should ideally wait for all changes to have been applied. This is currently not "async".
        self.send_response(Response(request_id, {"applied": True}))

    def m_textDocument_publishDiagnostics(self, params: Any) -> None:
        """handles the textDocument/publishDiagnostics notification"""
        mgr = self.manager()
        if mgr:
            mgr.diagnostics.receive(self.config.name, params)  # type: ignore

    def m_client_registerCapability(self, params: Any, request_id: Any) -> None:
        """handles the client/registerCapability request"""
        registrations = params["registrations"]
        for registration in registrations:
            method = registration["method"]
            capability_path, registration_path = method_to_capability(method)
            debug("{}: registering capability:".format(self.config.name), capability_path)
            self.capabilities.set(capability_path, registration.get("registerOptions", {}))
            self.capabilities.set(registration_path, registration["id"])
            toplevel_key = capability_path.split('.')[0]
            if toplevel_key.endswith('Provider'):
                for sv in self.session_views():
                    sv.register_capability(toplevel_key)
        self.send_response(Response(request_id, None))
```

rchl:
I'm still in the transitional phase where I have to avoid changes to existing LSP packages so I can't use the official way of doing it.

But OK, I'll keep using the internal function for now then.

PR #1125, opened 2020-06-15
Expand variables in didChangeConfiguration notification

rwols:
I need this for my pyls helper package which I am working on.

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -279,8 +279,10 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
     })
 
 
-def did_change_configuration(d: DottedDict) -> Notification:
-    return Notification.didChangeConfiguration({"settings": d.get()})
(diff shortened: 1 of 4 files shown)


jwortmann:
This should be done in the `workspace/configuration` response too then, right?

rwols:
Correct you are.

plugin/core/views.py
```diff
@@ -279,8 +279,10 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
     })
 
 
-def did_change_configuration(d: DottedDict) -> Notification:
-    return Notification.didChangeConfiguration({"settings": d.get()})
+def did_change_configuration(d: DottedDict, variables: Dict[str, str]) -> Notification:
+    settings = d.get()
+    settings = sublime.expand_variables(settings, variables)
+    return Notification.didChangeConfiguration({"settings": settings})
```
rchl, line 285: Would be good to use the correct window here too.

PR #1129, opened 2020-06-15
Request code actions using range extended to include diagnostics

### LINKED ISSUE #1128: Different amount of code actions available for same cursor/pointer position
Depending on if we use the mouse pointer or caret, the number of available code actions can differ (annotations will show a different amount than hover popup).

One example where this can happen with LSP-typescript:
```js
const a = require('./types')
```

Screenshot 2020-06-15 at 22 03 29

Since `require('./types')` has a diagnostic, hovering within that region will include the region of the diagnostic in `textDocument/codeAction` request while putting a caret within that region will only include the range of the caret and server will not include some code actions then.

When making a code action request for the caret (selection) we should probably extend the range if there is a diagnostic intersecting the selection.

rchl:
Change filter_by_point and filter_by_range to also return an extended
range that includes filtered diagnostics.

This range is then used to request code actions with all filtered
diagnostics included within the range.

Previously the code only included range of the first diagnostics which
wasn't entirely correct and for "range" requests the range was not
extended at all, making those two types of calls inconsistent in number
of code actions returned.

Resolves #1128

### DIFF
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -6,7 +6,7 @@ from html import escape
 from .code_actions import actions_manager, run_code_action_or_command
 from .code_actions import CodeActionOrCommand
 from .core.popups import popups
-from .core.protocol import Request, DiagnosticSeverity, Diagnostic, DiagnosticRelatedInformation, Point
+from .core.protocol import Request, DiagnosticSeverity, Diagnostic, DiagnosticRelatedInformation
 from .core.registry import LspTextCommand
 from .core.registry import LSPViewEventListener
 from .core.registry import windows
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -13,6 +13,7 @@ from .core.registry import windows
 from .core.settings import settings
 from .core.typing import List, Optional, Any, Dict
 from .core.views import FORMAT_MARKED_STRING, FORMAT_MARKUP_CONTENT, minihtml
+from .core.views import offset_to_point
 from .core.views import make_link
 from .core.views import text_document_position_params
 from .diagnostics import filter_by_point, view_diagnostics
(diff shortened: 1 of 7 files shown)


tests/test_code_actions.py
```diff
@@ -257,6 +257,30 @@ def test_requests_with_no_diagnostics(self) -> Generator:
...
+        yield from self.await_message("textDocument/didChange")
+        yield from self.await_client_notification(
+            "textDocument/publishDiagnostics",
+            create_test_diagnostics([
+                ('diagnostic word', Range(Point(0, 2), Point(0, 12))),
+                ('all content', Range(Point(0, 0), Point(0, 12))),
+            ])
+        )
+        self.view.sel().clear()
+        self.view.sel().add(sublime.Region(0, 5))
```
rwols, line 273: You cannot modify the cursor without an edit object. This should use `self.view.run_command("lsp_selection_set", ...)`.

tests/test_code_actions.py
```diff
@@ -257,6 +257,30 @@ def test_requests_with_no_diagnostics(self) -> Generator:
...
+        self._trigger_on_selection_modified_async()
+        yield 100
```
rwols, line 275: Why do we need these `yield 100`?

plugin/hover.py
```diff
@@ -84,8 +85,8 @@ def run(self, edit: sublime.Edit, point: Optional[int] = None, event: Optional[d
         if self.is_likely_at_symbol(hover_point):
             self.request_symbol_hover(hover_point)
 
-        self._diagnostics_by_config = filter_by_point(view_diagnostics(self.view),
-                                                      Point(*self.view.rowcol(hover_point)))
+        request_point = offset_to_point(self.view, hover_point)
```
rwols, line 88: Nice one.

tests/test_code_actions.py
rchl: Cool stuff

rchl: That actually fixes the problem that `on_modified_async` was not triggered and I had to trigger it manually. :)

rchl: Done

tests/test_code_actions.py
rchl: `on_modified_async` is async and also it triggers code actions asynchronously (from `debounced`) so we need to delay things a bit to wait for `on_modified_async` especially as it seems to be a bit random when it's coming.

It might cause randomness in tests but seems pretty stable for me with 100ms.

rwols: But the `self.await_message('textDocument/codeAction')` already takes care of waiting?

rchl: But it will take previous response if there was one. And there is one triggered implicitly before on adding characters.

rwols: Hmm, I see.

PR #1132, opened 2020-06-16
Cache all code action requests (apart from on save)

rchl:
This allows us to more efficiently cache code action results. Hovering
some diagnostic will return a cached response on the whole range of diagnostic
rather just on specific, hovered point.

Remove "request_for_point()" in favor of using more generic endpoint.
It's also more efficient now as we don't have to get and filter view
diagnostics twice as they are passed from the caller which already did
the work.


plugin/code_actions.py
```diff
@@ -80,23 +79,7 @@ class CodeActionsManager:
...
-        current_location_key = "{}#{}:{}".format(view.buffer_id(), view.change_count(), request_point)
-        if current_location_key in self._point_cache:
-            actions_handler(self._point_cache[current_location_key].get_actions())
-        else:
-            self._point_cache.clear()
-            point = offset_to_point(view, request_point)
-            diagnostics_by_config, extended_range = filter_by_point(view_diagnostics(view), point)
-            results = self.request_with_diagnostics(view, extended_range, diagnostics_by_config, actions_handler)
-            self._point_cache[current_location_key] = results
+        self._response_cache = {}  # type: Dict[str, CodeActionsCollector]
```
rwols, line 82: Remind me again when items from this dict are popped? Or is it growing to infinity?

rchl: It's cleared on every request that doesn't match the cache: https://github.com/sublimelsp/LSP/blob/06a59d66e904d522b29209a8b42d4deb3aa296bb/plugin/code_actions.py#L139-L140

The fact that it's a dict is maybe a bit confusing here. Could be a tuple of cache_key, response.

rwols: I'd much prefer that. Dicts can grow indefinitely, tuples cannot :)

PR #1133, opened 2020-06-16
Fix diagnostics for lowercase drive letter in DocumentUri

jwortmann:
It might happen that the server sends diagnostics with a lowercase drive letter (on Windows OS) for the filename in the `uri` field (e.g. the Julia server does that). Sublimes `view.file_name()` function on the other side returns a uppercase drive letter. Therefore the filenames don't match and diagnostics will only be shown in the panel, but not as regions in the editor or on hover.

### DIFF
--- a/plugin/diagnostics.py
+++ b/plugin/diagnostics.py
@@ -243,7 +243,8 @@ class DiagnosticViewRegions(DiagnosticsUpdateWalk):
 
     def begin_file(self, file_name: str) -> None:
         # TODO: would be nice if walk could skip this updater
-        if os.path.samefile(file_name, self._view.file_name()):
+        file = self._view.file_name()
+        if file and os.path.samefile(file_name, file):
             self._relevant_file = True
 
     def diagnostic(self, diagnostic: Diagnostic) -> None:
--- a/plugin/diagnostics.py
+++ b/plugin/diagnostics.py
@@ -43,7 +43,9 @@ def view_diagnostics(view: sublime.View) -> Dict[str, List[Diagnostic]]:
         file_name = view.file_name()
         if file_name:
             window_diagnostics = windows.lookup(view.window()).diagnostics.get()
-            return window_diagnostics.get(file_name, {})
+            for file in window_diagnostics:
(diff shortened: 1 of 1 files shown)


jwortmann:
I'm not sure how to resolve the lint error from mypy. `samefile` is only called now when `self._view.file_name()` is not None, i.e. it must be a string.

rwols:
It is because that function might run on the async worker thread of sublime text, or a thread spawned by LSP. In that case, calling `.file_name()` and then a microsecond later `.file_name()` again might not guarantee that the second call also results in a valid filename. Thus, you'll have to do it like this:
```
file_name = view.file_name()
if file_name:
    ...
```

jwortmann:
Yes you're right, thanks!

plugin/diagnostics.py
```diff
@@ -43,7 +43,9 @@ def view_diagnostics(view: sublime.View) -> Dict[str, List[Diagnostic]]:
         file_name = view.file_name()
         if file_name:
             window_diagnostics = windows.lookup(view.window()).diagnostics.get()
-            return window_diagnostics.get(file_name, {})
+            for file in window_diagnostics:
```
rwols, line 46: I'm not sure if it's wise to have a for-loop here as this is called mainly by filter_by_range which also loops over items. We might need a Path class with a custom equality operator and custom hash operator to stash it into dicts as keys. I'm guessing this is an unnoticable performance loss?

jwortmann: I don't notice any performance loss caused by the for-loop and the function calls from `os.path`. But I agree it's not as clean as the `get` call or something similar. Another option could be to convert the drive letters in Windows paths to uppercase before storing the diagnostics. But that might be error prone because we would need to ensure to prevent doing that for Unix files and folders, where paths are case sensitive.

rchl:
This could be relevant: https://github.com/microsoft/language-server-protocol/pull/1019 and if so, then the proper solution would be to normalize to lower-case in LSP. But that would be a bigger task I guess.

PR #1134, opened 2020-06-17
Fix user keybindings template to use array rather than object

rchl:
Also remove reference to LSP as it's a global keybindings file, not
specific to LSP.

### DIFF
--- a/Commands/Default.sublime-commands
+++ b/Commands/Default.sublime-commands
@@ -12,7 +12,7 @@
         "args": {
             "base_file": "${packages}/LSP/Keymaps/Default (${platform}).sublime-keymap",
             "user_file": "${packages}/User/Default ($platform).sublime-keymap",
-            "default": "// Settings in here override those in \"LSP/Keymaps/Default (${platform}).sublime-keymap\",\n\n{\n\t$0\n}\n"}
+            "default": "[\n\t$0\n]\n"}
     },
     {
         "caption": "LSP: Setup Language Server",
--- a/Commands/Default.sublime-commands
+++ b/Commands/Default.sublime-commands
@@ -4,7 +4,7 @@
         "command": "edit_settings",
         "args": {
           "base_file": "${packages}/LSP/LSP.sublime-settings",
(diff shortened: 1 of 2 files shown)


Commands/Default.sublime-commands
```diff
@@ -12,7 +12,7 @@
         "args": {
             "base_file": "${packages}/LSP/Keymaps/Default (${platform}).sublime-keymap",
             "user_file": "${packages}/User/Default ($platform).sublime-keymap",
-            "default": "// Settings in here override those in \"LSP/Keymaps/Default (${platform}).sublime-keymap\",\n\n{\n\t$0\n}\n"}
+            "default": "[\n\t$0\n]\n"}
```
rwols, line 15: This seems to have an extraneous } at the end?

rchl: Not extraneous but in the wrong place.

Commands/Default.sublime-commands
```diff
@@ -4,15 +4,16 @@
         "command": "edit_settings",
         "args": {
           "base_file": "${packages}/LSP/LSP.sublime-settings",
-          "default": "// Settings in here override those in \"LSP/LSP.sublime-settings\",\n\n{\n\t$0\n}\n"}
+          "default": "// Settings in here override those in \"LSP/LSP.sublime-settings\"\n\n{\n\t$0\n}\n"}
```
rwols, line 7: Can you also remove the extra newline?

`\n\n{\n\t$0\n}\n"}` -> `\n{\n\t$0\n}\n"}`

Menus/Main.sublime-menu
```diff
@@ -16,15 +16,16 @@
                 "command": "edit_settings",
                 "args": {
                   "base_file": "${packages}/LSP/LSP.sublime-settings",
-                  "default": "// Settings in here override those in \"LSP/LSP.sublime-settings\",\n\n{\n\t$0\n}\n"
+                  "default": "// Settings in here override those in \"LSP/LSP.sublime-settings\"\n\n{\n\t$0\n}\n"
```
rwols, line 19: See above.

PR #1136, opened 2020-06-17
Add on_workspace_configuration to the AbstractPlugin API

rchl:


plugin/core/sessions.py
```diff
@@ -603,7 +612,10 @@ def m_workspace_configuration(self, params: Dict[str, Any], request_id: Any) ->
         items = []  # type: List[Any]
         requested_items = params.get("items") or []
         for requested_item in requested_items:
-            items.append(self.config.settings.get(requested_item.get('section') or None))
+            configuration = self.config.settings.get(requested_item.get('section') or None)
+            if self._plugin:
+                self._plugin.on_workspace_configuration(requested_item, configuration)
+            items.append(configuration)
```
rchl, line 618: The problem with this solution at the moment is that I have to access `_workspace_folders` from `on_workspace_configuration` handler, to use the "includes_uri" method. This is obviously a bit wrong so options are:
 1) make `_workspace_folders` public (remove the underscore)
 2) add some public method on session to do what `includes_uri` is doing
 3) pass the list of WorkspaceFolders to the handler

I think 3 is too specific at least.

rchl: BTW. Can we treat all public methods of `Session` as something that plugins should be able to access? I imagine that could break easily.

plugin/core/protocol.py
```diff
@@ -411,3 +411,6 @@ def to_lsp(self) -> Dict[str, str]:
 
     def uri(self) -> str:
         return filename_to_uri(self.path)
+
+    def includes_uri(self, uri: str) -> bool:
+        return uri.startswith(self.uri())
```
rwols, line 416: Does this always work correctly on Windows?

plugin/core/sessions.py
rwols: Having workspace_folders public is fine I think.

rchl: I'll do that then but then I don't think we'll be able to keep the Session API fully stable...

plugin/core/protocol.py
rchl: It looks like it will work. When converting from URI to file path the case is normalized (through the code used in `uri_to_filename`).
```
>>> import urllib
>>> urllib.request.url2pathname(urllib.parse.urlparse('file:///c:/Users/IEUser/LSP/plugin/core/url.py').path)
'C:\\Users\\IEUser\\LSP\\plugin\\core\\url.py'
>>> urllib.request.url2pathname(urllib.parse.urlparse('file:///C:/Users/IEUser/LSP/plugin/core/url.py').path)
'C:\\Users\\IEUser\\LSP\\plugin\\core\\url.py'

Comment (rchl) [reply plugin/core/protocol.py:416]: I should have probably tested `filename_to_uri` in this case so here it is:
```
>>> urllib.request.pathname2url('C:\\Users\\IEUser\\LSP\\plugin\\core\\url.py')
'///C:/Users/IEUser/LSP/plugin/core/url.py'
>>> urllib.request.pathname2url('c:\\Users\\IEUser\\LSP\\plugin\\core\\url.py')
'///C:/Users/IEUser/LSP/plugin/core/url.py'
```
also works as expected regardless of the drive case.

And if you are thinking about the rest of the path besides just the drive letter then I guess that won't be normalized.

rchl: But then I don't see how would it be possible for the case of the rest of the path to get out of sync with what we sent initially to the server. I understand that the server might normalize the drive letter's case but it shouldn't change anything else. What would be the point of that?

PR #1141, opened 2020-06-18
Don't show dialog for successful workspace/executeCommand

jwortmann:
It's annoying to have a dialog pop up each time `lsp_execute` is run.

### DIFF
--- a/plugin/execute_command.py
+++ b/plugin/execute_command.py
@@ -46,7 +46,9 @@ class LspExecuteCommand(LspTextCommand):
         if response:
             msg += "with response: {}".format(response)
 
-        sublime.message_dialog(msg)
+        window = self.view.window()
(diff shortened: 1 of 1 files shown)


rwols:
What do you think, @ayoub-benali ?

ayoub-benali:
> What do you think, @ayoub-benali ?

Fine by me :)

jwortmann:
It might depend on the server whether it's useful to show a dialog box on successful commands, but I don't think that would be worth another setting for example.
When I use it with the Julia server, `workspace/executeCommand` is mostly used for small things like to wrap a statement into an if-block. The responses then only contain `{'applied' True}` for that and it's more disrupting than useful to show a dialog box for it, so I guess the status bar is a better alternative here.

PR #1144, opened 2020-06-18
Move documents.py and session_view.py up one level

rwols:
I will need this for my one-view-event-listener mega plan.


rchl:
Is there any definition for what is core and what not? :)

rwols:
The idea was that files in core should not import `sublime` and/or `sublime_plugin`. In hindsight this is a very hard task.

Issue #1145, opened 2020-06-19
gopls: Format on save is disabled globally for the duration of the session after syntax error


uhthomas:
# System information
macOS Catalina 10.15.5 (19F101)
Sublime Text Dev Channel, Build 4074

# Install
Package Control

# Reproduction
Assuming the LSP configuration
```json
{
    "lsp_format_on_save": true,
    "lsp_code_actions_on_save": {
        "source.fixAll": true,
        "source.organizeImports": true
    },
    "show_symbol_action_links": true,
    "only_show_lsp_completions": true,
    "clients": {
        "gopls": {
            "enabled": true,
            "settings": {
                "gopls.usePlaceholders": true,
                "gopls.analyses": {
                  "fillreturns": true,
                  "nonewvars": true,
                  "undeclaredname": true,
                  "unusedparams": true
                },
                "gopls.completeUnimported": true,
                "gopls.staticcheck": true
            }
        }
    }
}
```

Then, given the snippet (note the superfluous indentation, and missing `.`)
```go
package main

import "fmt"

func main() {



                fmt Println("Hello, world!")
}
```

Save the snippet as `main.go`, open `main.go` in `subl` and save (to run the format on save).

Then replace
```go
fmt Println("Hello, world!")
```
with
```go
fmt.Println("Hello, world!")
```

From this point, saving the file does not do any formatting. However, it does still do imports, if
```go
import "fmt"
```
is removed.

# Logs
```
Traceback (most recent call last):
  File "/Applications/Sublime Text.app/Contents/MacOS/Lib/python33/sublime_plugin.py", line 795, in on_activated_async
    run_view_callbacks('on_activated_async', view_id)
  File "/Applications/Sublime Text.app/Contents/MacOS/Lib/python33/sublime_plugin.py", line 622, in run_view_callbacks
    callback(*args)
  File "/Applications/Sublime Text.app/Contents/MacOS/Lib/python33/sublime_plugin.py", line 167, in exception_handler
    return event_handler(*args)
  File "/Users/thomas/Library/Application Support/Sublime Text 3/Installed Packages/LSP.sublime-package/plugin/core/documents.py", line 52, in on_activated_async
    self.manager.activate_view(self.view)
  File "/Users/thomas/Library/Application Support/Sublime Text 3/Installed Packages/LSP.sublime-package/plugin/core/windows.py", line 419, in activate_view
    self._initialize_on_open(view)
[...]

PR #1147, opened 2020-06-19
Pass the window/logMessage notifications through the loggers

rchl:
Deferred notifications were never logged through the loggers. Those were
logged in the log panel through a different code path:
m_window_logMessage -> WindowManager.handle_log_message
but since those didn't pass through the loggers, the remote logger didn't
see those.

I've noticed this while looking at #1145.

### DIFF
--- a/plugin/core/rpc.py
+++ b/plugin/core/rpc.py
@@ -191,9 +191,12 @@ class Client(TransportCallbacks):
     def flush_deferred_notifications(self) -> None:
         for payload in self._deferred_notifications:
             try:
-                handler = self._get_handler(payload["method"])
+                method = payload["method"]
+                handler = self._get_handler(method)
+                result = payload["params"]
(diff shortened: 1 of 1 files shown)


plugin/core/rpc.py
```diff
@@ -191,9 +191,12 @@ def execute_request(
     def flush_deferred_notifications(self) -> None:
         for payload in self._deferred_notifications:
             try:
-                handler = self._get_handler(payload["method"])
+                method = payload["method"]
+                handler = self._get_handler(method)
+                result = payload["params"]
+                self._logger.incoming_notification(method, result, handler is None)
```
rwols, line 197: Surprising that this wasn't noticed earlier :)

rchl: There is also `Client.response_handler` where we, in some cases, construct our own error response and the original payload is not logged anywhere. That also makes debugging harder.

PR #1149, opened 2020-06-19
Make formatting tasks run asynchronously

rchl:
- There are now three separate tasks, one for code actions on save,
   one for formatting request, and one for WillSaveWait request.
 - Those tasks are registered onto LspSaveCommand and run sequentially on saving.
 - Tasks by default have the timeout set to 1s (matching former sync request) and code actions to 2s and controlled by a setting.

~~I still need to clean this up and figure out the timeout stuff (timeout separate for each task or for whole save action?).~~

### DIFF
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -152,9 +152,6 @@ class ErrorCode:
     RequestCancelled = -32800
     ContentModified = -32801
 
-    # Defined by us
-    Timeout = -40000
-
 
 class Error(Exception):
 
--- a/Keymaps/Default (Linux).sublime-keymap
+++ b/Keymaps/Default (Linux).sublime-keymap
@@ -18,7 +18,7 @@
             {
                 "key": "lsp.session_with_capability",
                 "operator": "equal",
-                "operand": "codeActionProvider"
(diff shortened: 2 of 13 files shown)


rchl:
When using `lsp_save` command the `on_pre_save` and thus potentially `will_save` notification does not run at the right time. Will have to fix it somehow so that it runs before save tasks and is skipped before native save (only when `lsp_save` was used).

rchl:
It should be ready now.
The way I handle `on_pre_save` might be a bit hacky but not that different from existing, similar listener hacks.

rwols:
So can we remove the SyncRequestStatus class now?

I do think we should keep the lock, in case we implicitly rely on different threads mutating the `self._response_handlers` dictionary.

rchl:
> So can we remove the SyncRequestStatus class now?

Should I do it here?

rwols:
> Should I do it here?

Please do! :+1:

plugin/core/protocol.py
```diff
@@ -152,9 +152,6 @@ class ErrorCode:
     RequestCancelled = -32800
     ContentModified = -32801
 
-    # Defined by us
-    Timeout = -40000
```
rwols, line 156: Nice catch.

PR #1151, opened 2020-06-20
Update flake8, mypy and yapf versions

rwols:


tox.ini
```diff
@@ -17,9 +17,9 @@ max-line-length = 120
 
 [testenv]
 deps =
-    flake8==3.7.9
-    mypy==0.750
-    yapf==0.29.0
+    flake8==3.8.3
+    mypy==0.781
+    yapf==0.30.0
```
rwols, line 22: I tried using yapf for the whole project but it's not good at mypy types. The black formatter does seem to understand mypy types.

PR #1160, opened 2020-06-24
Merge DocumentSyncListener and HoverHandler

rwols:
The disabled_capabilities check is now also not cached. So if you update disabled_capabilities in your settings, you will see the effect immediately.

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -60,8 +60,7 @@ class DocumentSyncListener(LSPViewEventListener, AbstractViewListener):
     def __del__(self) -> None:
         self._clear_async()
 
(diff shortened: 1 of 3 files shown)


plugin/documents.py
```diff
@@ -159,3 +160,10 @@ def on_query_context(self, key: str, operator: int, operand: Any, match_all: boo
 
     def __str__(self) -> str:
         return str(self.view.id())
+
+    def on_hover(self, point: int, hover_zone: int) -> None:
```
rchl, line 164: Would try to move it to a better position in the file :)

rwols: Is 36d1af3 better?

PR #1161, opened 2020-06-24
Merge ColorListener and DocumentSyncListener

rwols:


plugin/documents.py
```diff
@@ -102,6 +112,15 @@ def on_activated_async(self) -> None:
...
             self._register_async()
 
+    def on_modified_async(self) -> None:
+        sel = self.view.sel()
+        if len(sel) < 1:
+            return
+        current_point = sel[0].b
+        if self._stored_point != current_point:
+            self._stored_point = current_point
+            debounced(self._do_document_color, 800, lambda: self._stored_point == current_point, async_thread=True)
```
rchl, line 122: Can we avoid debouncing entirely when colorProvider is disabled (even if we have to check settings twice)?

rwols: Done in 0a0bd9b

plugin/documents.py
```diff
@@ -149,13 +171,27 @@ def on_hover(self, point: int, hover_zone: int) -> None:
...
         self.view.run_command("lsp_hover", {"point": point})
 
+    # --- textDocument/documentColor -----------------------------------------------------------------------------------
+
+    def _do_color_boxes_async(self) -> None:
+        session = self.session("colorProvider")
+        if session:
+            session.send_request(Request.documentColor(document_color_params(self.view)), self._on_color_boxes)
+
+    def _on_color_boxes(self, response: Any) -> None:
```
rchl, line 181: This also runs on async thread (not ST's own async thread though).

rwols: Yeah. I don't know what to do with these kinds of callbacks. I'm thinking of also running all response handlers on ST's async thread to obtain a more serialized state of the plugin. Because if everything runs on ST's async thread, we don't really need locks anymore.

PR #1163, opened 2020-06-24
Merge LspCodeActionsListener and DocumentSyncListener

rwols:


rchl:
Code actions test fail because it tries to import `LspCodeActionsListener`.

rwols:
We should start linting the test files too.

plugin/documents.py
```diff
@@ -63,6 +69,9 @@ def run() -> None:
 
 class DocumentSyncListener(LSPViewEventListener, AbstractViewListener):
 
+    ACTIONS_ANNOTATION_KEY = "lsp_action_annotations"
+    code_actions_debounce_time = 800
```
rchl, line 73: Probably would make sense to have those timeout defined here to have a good overview

PR #1164, opened 2020-06-25
Fix: respect tabs when applying TextEdits

rwols:
I couldn't find a way to make view.insert or view.replace to respect tabs. We have to temporarily change view setting.

### DIFF
--- a/plugin/edit.py
+++ b/plugin/edit.py
@@ -2,7 +2,20 @@ import sublime
 import sublime_plugin
 from .core.edit import sort_by_application_order, TextEdit
 from .core.logging import debug
-from .core.typing import List, Dict, Optional, Any
+from .core.typing import List, Dict, Optional, Any, Generator
+from contextlib import contextmanager
+
+
+@contextmanager
+def temporary_view_setting(view: sublime.View, key: str, value: Any) -> Generator[None, None, None]:
+    settings = view.settings()
+    previous_value = settings.get(key)
+    settings.set(key, value)
+    yield
+    if previous_value is None:
+        settings.erase(key)
+    else:
+        settings.set(key, previous_value)
 
 
 class LspApplyWorkspaceEditCommand(sublime_plugin.WindowCommand):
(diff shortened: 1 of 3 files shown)


plugin/edit.py
```diff
@@ -2,7 +2,20 @@
...
+@contextmanager
+def temporary_view_setting(view: sublime.View, key: str, value: Any) -> Generator[None, None, None]:
+    settings = view.settings()
+    previous_value = settings.get(key)
+    settings.set(key, value)
+    yield
+    if previous_value is None:
+        settings.erase(key)
+    else:
+        settings.set(key, previous_value)
```
rchl, line 18: In case `translate_tabs_to_spaces: true` is set in global preferences, the `settings.get()` will return `True`, not `None`. So then we'll also keep the `True` value set on the view while it wasn't really set initially.

My point: I think it makes sense to always erase. At least we won't leave view-local setting behind on all views that were saved.

rchl: And I suppose for this example it doesn't matter if `translate_tabs_to_spaces` is set to `True` initially.
It will be always either True or False unless somehow the user is missing default settings.

plugin/edit.py
```diff
@@ -40,30 +53,21 @@ class LspApplyDocumentEditCommand(sublime_plugin.TextCommand):
...
                 if version is not None and version != view_version:
                     debug('ignoring edit due to non-matching document version')
                     continue
                 region = sublime.Region(self.view.text_point(*start), self.view.text_point(*end))
-
-                if start[0] > last_row and newText[0] != '\n':
+                if start[0] > last_row and replacement[0] != '\n':
                     # Handle when a language server (eg gopls) inserts at a row beyond the document
                     # some editors create the line automatically, sublime needs to have the newline prepended.
                     debug('adding new line for edit at line {}, document ended at line {}'.format(start[0], last_row))
```
rchl, line 69: Should we remove this debug message now? The bug is fixed, not sure if there is any point in logging it now.

plugin/edit.py
```diff
@@ -40,30 +53,21 @@ class LspApplyDocumentEditCommand(sublime_plugin.TextCommand):
...
-
-    def apply_change(self, region: sublime.Region, newText: str, edit: Any) -> None:
-        if region.empty():
-            self.view.insert(edit, region.a, newText)
-        else:
-            if len(newText) > 0:
-                self.view.replace(edit, region, newText)
-            else:
-                self.view.erase(edit, region)
+                    self.view.replace(edit, region, replacement)
```
rchl, line 73: I wonder if using always "replace" can have some bearing on cursor placement...

rwols: Didn't want to find out. Reverted it.

plugin/edit.py
rwols: Remobved in 4638808 (huh, only numbers. What are the odds?)

plugin/edit.py
rwols: Oh I see what you mean. This might indeed be problematic.

rchl: I'm fine with this PR if you update this to the version that you've posted on Discord.

PR #1165, opened 2020-06-25
Fix completions not being requested from server in some cases

rchl:
When "on_query_completions" has been triggered before session has
been initialized (for example on start of ST), then the initialize()
has not seen the session yet and has disabled the completions
for the rest of the session for that file.

Fix by not storing "initialized" or "ready" state and instead handle
"initialize" (renamed to "apply_view_settings") on every request to
"on_query_completions".

To make that work nicely, also fixed issue with "auto_complete_triggers"
values being duplicated. Those values are stored in STs session so after
restarting ST, we would add duplicate entries to that list.
With "apply_view_settings" running on each completion request now that
would become even worse.


rchl:
> The real fix would be to move the settings manipulation to the `SessionView`

I feel those changes will just make it easier.

Issue #1029, opened 2020-05-07
Long log messages should be trimmed


genericptr:
Using the document/workspace symbol commands it's possible send payloads of many megabytes which is viewed in the LSP log panel will halt the program. These long messages should be identified and trimmed down to show only the first portion.

rwols:
Kind of fixed with #1167. However, that PR doesn't "cut off" the string, but rather replaces it with the string "<params with X characters>". But, if you need more serious logging I suggest you look into the remote logger from @rchl.

PR #1167, opened 2020-06-26
Log panel improvements

rwols:
- Don't print param payload if it exceeds 8k chars (can be modified).
- Do debounced logging to improve performance.
- Do not convert to JSON, send it to the plugin host, and then convert
  it back into a dict, only to then convert it to a JSON string again.
  Instead maintain a queue of log messages to be handled.
  This also unclutters the console when investigating commands with
  sublime.log_commands(True).

resolves https://github.com/sublimelsp/LSP/issues/1029

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -5,6 +5,7 @@ from .diagnostics import DiagnosticsStorage
 from .logging import debug
 from .logging import exception_log
 from .message_request_handler import MessageRequestHandler
+from .panels import update_server_panel
 from .protocol import Error
 from .rpc import Logger
 from .sessions import get_plugin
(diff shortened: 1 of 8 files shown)


plugin/core/windows.py
```diff
@@ -393,12 +382,14 @@ def stderr_message(self, message: str) -> None:
...
-    def log(self, message: str, params: Any, log_payload: bool = True) -> None:
+    def log(self, message: str, params: Any) -> None:
 
         def run_on_async_worker_thread() -> None:
             nonlocal message
-            if log_payload:
-                message = "{}: {}".format(message, params)
+            params_str = str(params)
+            if len(params_str) >= settings.log_max_size:
+                params_str = '<params with {} characters>'.format(len(params_str))
```
rwols, line 391: Hmm, this diff strips away all the information. The linked issue tells to "cut off" the params. That is however not possible given that we're highlighting it. It would screw up the syntax parser.

We could disable highlighting the params.

rchl:
Good stuff but I'm not sure I'm seeing how anything changed in this regard:

> Do not convert to JSON, send it to the plugin host, and then convert
it back into a dict, only to then convert it to a JSON string again.

Isn't the same is still happening, just not for every message?

rwols:
> Isn't the same is still happening, just not for every message?

When you pass arguments to `view.run_command`, or `window.run_command`, or `sublime.run_command`, then the arguments have to be convertible to JSON. Sublime then actually converts the args to JSON. It then looks up the appropriate TextCommand/WindowCommand/ApplicationCommand instance. Then it decodes the JSON to a Python object. And then it invokes the command in Python land.

This is also the reason why you can't pass arbitrary Python objects as arguments to an ST command.

rchl:
Yes, my point was that it still does it (run `lsp_update_server_panel`), just less often

PR #1170, opened 2020-06-28
Rework diagnostics

### LINKED ISSUE #852: other value for auto_show_diagnostics_panel: when_no_other_panel_visible
Currently, the only possible values for `auto_show_diagnostics_panel` are `never`, `always` and `saved`. It'd be nice if there was an other value, maybe `when_no_other_panel_visible` maybe (this name isn't the best, but I don't have a better idea).

The idea is to only show the diagnostic panel *if no other panel is already visible*. This is especially important when using build systems. I build my python code, and then as soon as I save, the build result is hidden by the diagnostics (it's even worse because I have `save_on_focus_lost` set to `true`).

The workflow I would like to have is something like this:

- diagnostic are shown as soon as I open the file
- I write code
- I build my file
- the build result prevents the diagnostics from showing up
- I press escape to hide the build results
- the diagnostic panel is now shown
- if I decide to re-show the build results, then the diagnostic panel isn't forced onto me, it just waits.

The behavior should be exactly the same for any other panel (eg the console).

It should be really easily to implement: just make sure `window.active_panel()` is `None` before showing the diagnostics.

In fact, this `when_no_other_panel_visible` could work with `always` and `saved`, so maybe have the allowed values could be something like:

- `never`
- `saved`
- `saved:when_no_other_panel_visible`
- `always`
- `always:when_no_other_panel_visible`

I'd be happy to contribute that feature if the maintainers think it's a good idea :slightly_smiling_face:

### LINKED ISSUE #858: Option to disable validation underline and hover
Similar to https://github.com/sublimelsp/LSP/issues/762 but not looking to disable completely, just the underline (and hover) in the document.  Since the gutter mark can be disabled, why not the underline?

As far as I can tell, the capabilities provided by linters can go well beyond what the language servers provide in terms of identifying (and fixing) code style type issues.  If not disabling entirely, disabling underline (and hover) would eliminate the "collision"/duplication issues.  Would be especially useful if configurable by language.

FWIW, VSCode provides options to disable validations for languages.  Examples for Javascript and CSS:

image

image

### LINKED ISSUE #1089: Increase the delay of checking
Hello,

It could be great to increase (or let the user decide) the delay of checking, because right now, having an kind of instant feedback when you're typing your code is quite disturbing for me/my eyes.

I would like to have a feedback when nothing happens (no type, like a debounce) after 2s for instance.

Thank you!

tl;dr: start checking after 2s using debounce.

### LINKED ISSUE #1130: Change diagnostics "underline" to distinguish between squiggly and straight
One of the things that bothers me is how `highlight` can distinguish between `squiggly` and `underline`, but for diagnostics, it's `underline` which is **actually** squiggly:
Screen Shot 2020-06-15 at 7 48 56 PM

```
// [LSP.sublime-settings]
// Highlighting style of code diagnostics.
// Valid values are "underline" or "box"
"diagnostics_highlight_style": "underline",
```

It'd be amazing if there was an ability to make diagnostics be solid underlines, not squiggly underlines. I've tried poking through the source code but python is a bit hard to navigate imo.

### LINKED ISSUE #1155: "View not found for ..." after closing file
After we send `textDocument/didClose`, servers like to send back `textDocument/publishDiagnostics` with empty diagnostics which causes a debug log to trigger - `view not found for ...` from `DiagnosticsPresenter.update`. I haven't investigated best way to fix it. The easy way could be to just remove the log but maybe we should rather unregister something on `didClose`.

rwols:
- Put the storage in the SessionBuffer class.
- Add delay setting.
- Add extra delay setting when the completion widget is visible
- Allow an empty string for the diagnostics style to suppress in-line
  diagnostics.
- Don't print a debug statement when a view is not found.
- Do not manipulate the diagnostics panel when another panel is visible.

Resolves #852
Resolves #858
Resolves #1089
Resolves #1155
Resolves #1130

This is still a WIP.

- [x] Diagnostics panel functionality
- [x] In-line diagnostics functionality
- [x] Hover popup shows diagnostics
- [x] Diagnostics cursor
- [x] Diagnostic in status bar
- [x] Figure out what to do with the project path prefix
- [x] Test clearing diagnostics with "LSP: Clear Diagnostics" command (do we even need this?)


rchl:
Maybe it's expected at this stage but just wanted to say that testing this with:
```
"diagnostics_delay": 2000,
"diagnostics_additional_delay_auto_complete": 1000,
```
and it seems quite erratic in that the diagnostics don't show for newly typed text that contains errors. I have to reopen the file for them to show up.

rchl:
> Test clearing diagnostics with "LSP: Clear Diagnostics" command (do we even need this?)

Kill it!

LSP.sublime-settings
```diff
@@ -56,9 +56,22 @@
...
+  // number). The delay will only kick into action when previously there were
+  // no diagnostics in the view. If there were previous diagnostics in the view,
+  // then the delay setting here is ignored and diagnostics are updated
+  // immediately.
+  "diagnostics_delay": 0.0,
+
+  // Add an additional delay when presenting diagnostics, and when the
+  // auto-complete widget is currently visible. See the setting
+  // "diagnostics_delay".
+  "diagnostics_additional_delay_auto_complete": 0.0,
```
rchl, line 69: Not clear to me if this value is treated as an extra on top of `diagnostics_delay` or should generally be bigger than the `diagnostics_delay`.

Also, the defaults here don't match the code in types.py.

rchl:
> Maybe it's expected at this stage but just wanted to say that testing this with:

Oh, it's seconds, not milliseconds...

rwols:
> Oh, it's seconds, not milliseconds...

Yes. Perhaps that's too confusing? The other, existing "duration" settings are in milliseconds.

rchl:
Then following existing convention would be better. And have `_ms` suffix for those.

rwols:
> Then following existing convention would be better. And have _ms suffix for those.

`diagnostics_delay_ms` is now in milliseconds, as well as `diagnostics_additional_delay_auto_complete_ms`. I've clarified in LSP.sublime-settings that the two are added together.

LSP.sublime-settings
```diff
@@ -56,9 +56,26 @@
   // hint: 4
   "show_diagnostics_severity_level": 4,
 
+  // Delay showing diagnostics by this many milliseconds.
+  // The delay will only kick into action when previously there were
+  // no diagnostics in the view. If there were previous diagnostics in the view,
+  // then the delay setting here is ignored and diagnostics are updated
+  // immediately.
+  "diagnostics_delay_ms": 0,
```
rchl, line 64: I think that this should be set to some decent value by default. Unless you feel that it's not finished or something (I haven't had a chance to try it out properly yet).

PR #1174, opened 2020-07-03
Add diagnostic panel severity level setting

willrowe:
This adds the `diagnostics_panel_include_severity_level` setting, which allows the diagnostics shown in the panel to be controlled independently from the ones shown in the code itself, which uses the `show_diagnostics_severity_level` setting.

This seems to be the least complex way to accomplish what I mentioned [on Discord](https://discord.com/channels/280102180189634562/645268178397560865/725710727712604171). While I may want to show hints and info level diagnostics in the code, which can be extremely helpful, I only want the diagnostic panel to be shown when I save a file and there are warning or error level diagnostics that would cause the code to not run. The `auto_show_diagnostics_panel_level` setting helps control when the panel is shown, but since the panel currently shows _all_ diagnostics that meet the `show_diagnostics_severity_level` setting from _all_ open files, it is very difficult to quickly grok what errors and warnings there are, since there can be a lot of info and hint level diagnostics.

The default for the new `diagnostics_panel_include_severity_level` setting matches the default value for the `show_diagnostics_severity_level` setting, but it may be a good idea to also have this set to whatever the user setting is when they update so that it does not unexpectedly start showing more diagnostics for users that have `show_diagnostics_severity_level` set to a value lower than `4`. I am not sure if that is possible or where to do it though.

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -41,6 +41,13 @@
   // hint: 4
   "auto_show_diagnostics_panel_level": 2,
 
+  // Only show diagnostics in the panel with a severity equal to or less than:
+  // error: 1
+  // warning: 2
+  // info: 3
+  // hint: 4
+  "diagnostics_panel_include_severity_level": 4,
+
   // Show errors and warnings count in the status bar
   "show_diagnostics_count_in_view_status": false,
 
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -41,13 +41,6 @@
   // hint: 4
   "auto_show_diagnostics_panel_level": 2,
 
-  // Only show diagnostics in the panel with a severity equal to or less than:
-  // error: 1
-  // warning: 2
-  // info: 3
-  // hint: 4
-  "diagnostics_panel_include_severity_level": 4,
-
   // Show errors and warnings count in the status bar
   "show_diagnostics_count_in_view_status": false,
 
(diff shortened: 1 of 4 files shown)


rwols:
Sorry for the delay on this. It looks good to me, but I'd like to get #1170 merged before this.

willrowe:
Sounds good.

Any thoughts on how to migrate the default value for this setting based on the current value set for `show_diagnostics_severity_level`?

rwols:
> Any thoughts on how to migrate

Well, it's just a simple global setting. I'll add it to the release notes and that should be sufficient.

willrowe:
Ok, works for me.

rwols:
Please merge with the updated base branch. The diagnostics logic is moved to session_buffer.py, so you probably need to put your logic there now.

willrowe:
It has been rebased.

LSP.sublime-settings
```diff
@@ -41,6 +41,13 @@
   // hint: 4
   "auto_show_diagnostics_panel_level": 2,
 
+  // Only show diagnostics in the panel with a severity equal to or less than:
+  // error: 1
+  // warning: 2
+  // info: 3
+  // hint: 4
+  "diagnostics_panel_include_severity_level": 4,
```
rwols, line 49: I think it's better if this setting were placed just under "show_diagnostics_severity_level".

rwols: Moreover, clarify in "show_diagnostics_severity_level" that it only applies to the squiggly lines (and the gutter icons).

plugin/session_buffer.py
```diff
@@ -181,7 +181,7 @@ def on_diagnostics_async(self, raw_diagnostics: List[Dict[str, Any]], version: O
                     total_errors += 1
                 elif diagnostic.severity == DiagnosticSeverity.Warning:
                     total_warnings += 1
-                if diagnostic.severity <= settings.show_diagnostics_severity_level:
+                if diagnostic.severity <= settings.diagnostics_panel_include_severity_level:
```
rwols, line 184: Ah yes. Life can be so simple sometimes :)

LSP.sublime-settings
willrowe: Done.

willrowe:
Thanks @rwols!

PR #1175, opened 2020-07-03
Import typing based on version rather than try/except

rchl:
We know exactly from which python version "typing" import is available
so no need to try and catch.

This also makes pyright type checker work well with the code base.


rwols:
Perhaps this can also go in the master branch?

rchl:
Sure.

PR #1180, opened 2020-07-08
Add a dev command to parse VSCode package.json files

rwols:
for setting up language server helpers.

### DIFF
--- a/Commands/Default.sublime-commands
+++ b/Commands/Default.sublime-commands
@@ -16,6 +16,10 @@
         }
     },
     {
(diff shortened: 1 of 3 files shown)


rchl:
Good stuff but it feels like it doesn't belong in this repo.
Normal users shouldn't need it so it could have been just a separate package that would have to be cloned manually.

rwols:
Yeah, I couldn't find a better place for the command. Having yet another package to maintain felt like too much of a burden. I don't think people will find it distracting.

PR #1181, opened 2020-07-08
Work around broken language servers part II: PowerShellEditorServices

rwols:
- The "tags" in a CompletionItem response are null (invalid per spec)
- The "signatures" in a textDocument/signatureHelp response is null (invalid per spec)

### DIFF
--- a/plugin/core/signature_help.py
+++ b/plugin/core/signature_help.py
@@ -163,20 +163,19 @@ class SignatureHelp(object):
 def create_signature_help(response: Optional[dict]) -> Optional[SignatureHelp]:
     if response is None:
         return None
-
-    signatures = list(parse_signature_information(signature) for signature in response.get("signatures", []))
-    active_signature = response.get("activeSignature", -1)
-    active_parameter = response.get("activeParameter", -1)
-
-    if signatures:
-        if not 0 <= active_signature < len(signatures):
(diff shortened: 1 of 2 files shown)


plugin/core/signature_help.py
```diff
@@ -163,20 +163,19 @@ def _build_overload_selector(self) -> str:
...
+        signatures = [parse_signature_information(signature) for signature in raw_signatures]
+        if signatures:
+            active_signature = response.get("activeSignature", -1)
+            active_parameter = response.get("activeParameter", -1)
+            if not 0 <= active_signature < len(signatures):
+                debug("activeSignature {} not a valid index for signatures length {}".format(
+                    active_signature, len(signatures)))
+                active_signature = 0
+            return SignatureHelp(signatures, active_signature, active_parameter)
+    return None
```
predragnikolic, line 178:
```suggestion
    signatures = response.get("signatures") or []
    signatures = [parse_signature_information(signature) for signature in signatures]
    if signatures:
        active_signature = response.get("activeSignature", -1)
        active_parameter = response.get("activeParameter", -1)
        if not 0 <= active_signature < len(signatures):
            debug("activeSignature {} not a valid index for signatures length {}".format(
                active_signature, len(signatures)))
            active_signature = 0
        return SignatureHelp(signatures, active_signature, active_parameter)
    return None
```

PR #1185, opened 2020-07-13
Update features docs

rwols:
- Update outdated keybindings explanation
- Update paragraph about "completion resolve docs" by adding a context to check if the AC widget is not visible
- Add missing settings (I'm not sure we even need this overview?)

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -61,21 +61,31 @@ The following variables will be expanded, but only if they are top-level array i
 
 ### Overriding keybindings
 
-Sublime's keybindings can be edited from the `Preferences: Key Bindings` command.
-The following example overrides `f12` to use LSP's go to definition when in javascript/typescript:
+LSP's keybindings can be edited from the `Preferences: LSP Keybindings` command from the command palette.
+There is a special context called `lsp.session_with_capability` that can check whether there is a language server active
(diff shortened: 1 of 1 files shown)


docs/features.md
```diff
@@ -95,13 +105,26 @@ Some completion items can have documentation associated with them.
...
 
-You can change the default keybinding by remaping the command bellow:
+You can change the default keybinding by remapping the command as below:
 
-```
-{ "keys": ["f12"], "command": "auto_complete_open_link", "context":
-	[
-		{ "key": "auto_complete_visible", "operator": "equal", "operand": true },
-	]
+```js
```
predragnikolic, line 110: The deleted code block meant to remap the `auto_complete_open_link`command to something else.
I imagined that people would just copy this template: 
```
{ "keys": ["f12"], "command": "auto_complete_open_link", "context":
	[
		{ "key": "auto_complete_visible", "operator": "equal", "operand": true },
	]
}
```
paste it in theirs keybinding preferences and remap this `auto_complete_open_link` to something other than `f12`.

Not sure why you deleted this and put a `lsp_symbol_definition` command instead of `auto_complete_open_link` :)

rwols: Whoops, that should indeed just be `auto_complete_open_link`

PR #1188, opened 2020-07-14
Use the escapeall markdown extension to handle \< and \> in markdown

rwols:
This does nothing at the moment, but should render `\<` and `\>` correctly once https://github.com/facelessuser/pymdown-extensions/commit/5fca01fa343b67c211ba5a1b51d4b0a40e939ccc is in https://github.com/facelessuser/sublime-pymdownx/tree/master/st3/pymdownx

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -406,8 +406,18 @@ def minihtml(view: sublime.View, content: Union[str, Dict[str, str], list], allo
     if is_plain_text:
         return text2html(result)
     else:
-        frontmatter_config = mdpopups.format_frontmatter({'allow_code_wrap': True})
-        return mdpopups.md2html(view, frontmatter_config + result)
+        frontmatter = {
+            "allow_code_wrap": True,
+            "markdown_extensions": [
+                {
+                    "pymdownx.escapeall": {
+                        "hardbreak": False,
+                        "nbsp": False
+                    }
+                }
+            ]
+        }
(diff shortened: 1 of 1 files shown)


plugin/core/views.py
```diff
@@ -406,8 +406,13 @@ def minihtml(view: sublime.View, content: Union[str, Dict[str, str], list], allo
     if is_plain_text:
         return text2html(result)
     else:
-        frontmatter_config = mdpopups.format_frontmatter({'allow_code_wrap': True})
-        return mdpopups.md2html(view, frontmatter_config + result)
+        frontmatter = {
+            "allow_code_wrap": False,
+            "markdown_extensions": [
+                {"pymdownx.escapeall": {}}
```
rchl, line 412: It looks like it won't do "nothing" right now as the value of `allow_code_wrap` has been changed from True to False.
Why is that? That will likely end up working worse.

Also it looks like `escapeall` will replace spaces with `&nbsp;` so that will also make blocks not have possibility to wrap on spaces.

rwols: Ah yes, allow_code_wrap must remain unchanged. I’m making a lot of mistakes lately 🤔

rwols: I think  e330a6b fixes all your points.

rchl: Hopefully. Have not tested but let's try it.

(Also I've realized that this will only process characters escaped with `\` so not sure what's the expected value for `hardbreak` and `nbsp` settings. I guess we should match commonmark if that's what servers assume typically.)

PR #1191, opened 2020-07-15
Do not override the built-in "Next Modification" keybinding

### LINKED ISSUE #620: ctrl+. hijacks bindings from other plugins
* OS and language server
Linux, gopls
* How you installed LSP
Cloned from git
* Minimal reproduction steps
N/A
* Log
N/A

Please consider changing the `lsp_code_actions` key binding from `ctrl+.` to something else. 

I'm using GoSublime and I assume there are people still using this plugin for features that are not in the gopls/golsp yet. Most GoSublime's key bindings are a combination of `ctrl+.` and another key. LSP's configuration hijacks it and the only way I've found to fix it is to copy all key binding to the user's key binding file, which is annoying.

`ctrl+., ctrl+a` would be a good choice.

rwols:
Fixes #620

### DIFF
--- a/Keymaps/Default (Linux).sublime-keymap
+++ b/Keymaps/Default (Linux).sublime-keymap
@@ -23,19 +23,19 @@
         ]
     },
     // Show Code Actions
-    {
-        "command": "lsp_code_actions",
-        "keys": [
-            "ctrl+."
-        ],
-        "context": [
-            {
-                "key": "lsp.session_with_capability",
-                "operator": "equal",
-                "operand": "codeActionProvider"
-            }
-        ]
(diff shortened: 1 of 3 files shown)


predragnikolic:
Just in the release notes put a special higlight on this 🙂
as people may think that code actions are broken.

rwols:
> Just in the release notes put a special higlight on this slightly_smiling_face
as people may think that code actions are broken.

Sounds good :)

rchl:
There was no conflict on Mac but I guess it might be better to be consistent...

BTW. I think PR had wrong title

PR #1193, opened 2020-07-15
"Prepare Rename" feature

rwols:
Since 3.12.

- [x] It reports an error when the selection cannot be renamed
- [x] When the session is a prepareProvider, do not check if the selection is on a word (the session can do that)
- [x] It's clumsy that `lsp_rename_basic` is visible in the command palette. How do we hide it, but still enable input handlers?
- [ ] placeholder logic is untested, my known language servers don't do this. Maybe gopls?


predragnikolic:
Raoul, can you tell me with what ls can I test prepareRename? :)

rwols:
A server needs to support is first of all. One such server is for example clangd. To test the feature, you can test whether the server rejects a rename of an invalid symbol, like a language constant or a fundamental type.

plugin/rename.py
```diff
@@ -37,30 +42,80 @@ def get_current_symbol_name(self) -> str:
...
+                    self.event = event
+                else:
+                    # trigger InputHandler manually
+                    raise TypeError("required positional argument")
+        else:
+            if new_name:
+                return self._do_rename(position, new_name)
+            else:
+                # trigger InputHandler manually
+                raise TypeError("required positional argument")
```
rwols, line 87: Kinda gross, but oh well.

PR #1194, opened 2020-07-18
Document symbol improvements

### LINKED ISSUE #1172: When a server returns SymbolInformation[] in textDocument/documentSymbol, show the containerName
Document symbols can and should support the containerName in the UI.

https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#textDocument_documentSymbol

`/**
	The name of the symbol containing this symbol. This information is for
	 user interface purposes (e.g. to render a qualifier in the user interface
	 if necessary). It can't be used to re-infer a hierarchy for the document
	 symbols.
	 */
	containerName?: string;`

For example something like this would be helpful to give context to the symbol without affecting the filtering behavior. 

Screen_Shot_2020-06-30_at_12 16 15_PM

rwols:
- Put the name (without spaces) in the first row
- Put the parent hierarchy in the second row
- Put the kind and detail in the third row

~~This only works when the server returns DocumentSymbol[] as response,
not when it returns SymbolInformation[].~~ Works for both DocumentSymbol[] and SymbolInformation[] return types.

close #1172

### DIFF
--- a/plugin/symbols.py
+++ b/plugin/symbols.py
@@ -17,6 +17,14 @@ def format_symbol_kind(kind: int) -> str:
     return str(kind)
 
 
+def _kind_and_detail(item: Dict[str, Any]) -> str:
+    kind = format_symbol_kind(item["kind"])
+    detail = item.get("detail")
+    if isinstance(detail, str) and detail:
+        return "{} | {}".format(kind, detail)
+    return kind
+
+
 def format_symbol_scope(kind: int) -> str:
     if 1 <= kind <= len(SYMBOL_KINDS):
         return SYMBOL_KINDS[kind - 1][1]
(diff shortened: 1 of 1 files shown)


plugin/symbols.py
```diff
@@ -119,28 +126,27 @@ def process_symbols(self, items: List[Dict[str, Any]]) -> List[List[str]]:
...
-        children = item.get('children') or []
-        for child in children:
-            self.process_document_symbol_recursive(quick_panel_items, child, depth + 1)
+        name = item['name']
+        with _additional_name(names, name):
+            kind_and_detail = format_symbol_kind(kind)
+            detail = item.get('detail')
+            if detail:
+                kind_and_detail += " - {}".format(detail)
+            quick_panel_items.append([name, " → ".join(names), kind_and_detail])
```
rchl, line 146: There is some redundancy for top-level symbols but no big deal.

Screenshot 2020-07-18 at 15 46 26

rwols:
This last commit compacts the items to two rows (instead of three rows).

clangd: (DocumentSymbol)
Screenshot from 2020-07-19 21-32-12

pyls: (SymbolInformation)
Screenshot from 2020-07-19 21-31-51

AmjadHD:
doesn't the quick panel support icons ?

AmjadHD:
@rwols doesn't the quick panel support icons?

rwols:
> @rwols doesn't the quick panel support icons?

Unfortunately, no.

AmjadHD:
well that's no right ! you should open a feature request.

rwols:
@AmjadHD great minds think alike https://github.com/sublimehq/sublime_text/issues/3489

PR #1196, opened 2020-07-18
Add "show_code_actions" setting

predragnikolic:
Some people like the annotation, some like the bulb. 
It just depends on personal preference. (although annotation have the advantage of being clickable)

This settings should make both worlds happy.


predragnikolic:
Currently I haven't added a settings to hide the code actions, like "show_code_actions": "none"
Instead I would suggest adding the `"codeAction"` value to `"disabled_capabilities"`, because that makes more sense. (... that can be a separate PR if you agree on having something like that)

plugin/documents.py
```diff
@@ -228,15 +228,25 @@ def _on_code_actions(self, responses: CodeActionsByConfigName) -> None:
...
+                                  [region],
+                                  flags=flags,
+                                  annotations=["<div class=\"actions\">{}</div>".format(code_actions_link)],
+                                  annotation_color='#2196F3')
+        elif global_settings.show_code_actions == 'bulb':
+            self.view.add_regions(self.CODE_ACTIONS_KEY,
+                                  [region],
+                                  'markup.changed',
+                                  'Packages/LSP/icons/lightbulb.png',
+                                  flags)
```
rwols, line 246: There are a few duplicated arguments for the `add_regions` function here. I would prefer if this was a single call to `add_regions`, like so:

```python
if global_settings.show_code_actions == 'annotation':
    ...
elif global_settings.show_code_actions == 'bulb':
    ...
else:
    # handle this case!
    pass
self.view.add_regions(self.CODE_ACTIONS_KEY, [region], ...)
```

predragnikolic: Done with cb2a51e

plugin/documents.py
```diff
@@ -228,22 +228,21 @@ def _on_code_actions(self, responses: CodeActionsByConfigName) -> None:
         if action_count == 0:
             return
         suffix = 's' if action_count > 1 else ''
```
rwols, line 230: This variable can move into the `else` block.

predragnikolic: a1f5272

rwols:
@rchl any thoughts on this?

rchl:
I'm fine with it.

I still think that having an icon that can't be clicked is a confusing and inferior solution but it can work for more experienced users that defined the keyboard action for showing code actions (I know that context menu is also an option but who's gonna reach for the mouse every time...).

The only problem is with less-experienced users that will enable it because they are more familiar with the bulb but won't understand the downsides. Maybe for those, we could at least expand the option comment a bit (but no clear message comes to mind now).

LSP.sublime-settings
```diff
@@ -8,6 +8,11 @@
   // Valid values are "never", "always" and "saved"
   "auto_show_diagnostics_panel": "saved",
 
+  // Show code actions:
+  // "annotation" - show an annotation on the right when code actions are available.
+  // "bulb" - show a bulb in the gutter when code actions are available.
+  "show_code_actions": "annotation",
```
rchl, line 14: Maybe group this option with `show_symbol_action_links`?

rwols:
> The only problem is with less-experienced users that will enable it because they are more familiar with the bulb but won't understand the downsides. Maybe for those, we could at least expand the option comment a bit (but no clear message comes to mind now.

This is a good point. Can this note be added in the comments in LSP.sublime-settings?

predragnikolic:
I have two suggestions:
```
  // Show code actions:
  // "annotation"(clickable) - show an annotation on the right when code actions are available.
  // "bulb"(not clickable) - show a bulb in the gutter when code actions are available.
```
or 
```
  // Show code actions:
  // "annotation" - show an annotation on the right when code actions are available.
  // "bulb" - show a bulb in the gutter when code actions are available.
  // Note: Annotation can be clicked. The bulb can not be clicked. 
```
?

rwols:
I would personally prefer the first one as it is a bit more clear :)

rchl:
> Note: Annotation can be clicked. The bulb can not be clicked.

I think it's not necessary to mention that annotations can be clicked. I think it should be assumed that they can if not mentioned. Just having a note for bulbs would be enough IMO.

Maybe:
````
Note: Due to API limitations, the "bulb" icon can not be clicked so the code actions must be triggered
using a keyboard shortcut or the context menu.
```

LSP.sublime-settings
```diff
@@ -107,6 +102,13 @@
   // Valid values are "dot", "circle", "bookmark", "sign" or ""
   "diagnostics_gutter_marker": "dot",
 
+  // Show code actions:
+  // "annotation" - show an annotation on the right when code actions are available.
+  // "bulb" - show a bulb in the gutter when code actions are available.
+  // Note: Due to API limitations, the "bulb" icon can not be clicked so the code actions can only triggered
```
rwols, line 108: Nitpick: can only triggered --> can only be triggered

rchl:
Not trying to throw a wrench in here but that setting could in theory be an array. ;)

jwortmann:
This is already merged, but let me add some suggestions here:

> but that setting could in theory be an array

I agree, that would allow maximum configuration ability for users, including the option to turn off displaying of code actions.

Also I would recommend to change the scope from `markup.changed` to one of the `region` scopes, if the lightbulb icon should be colored. For example the built-in Celeste scheme has no rule for `markup.changed`, and it's not among the recommended minimal scope coverage for color schemes (and furthermore it seems wrong semantically). And why using a hardcoded color `#2196F3` for annotations, instead of a color from the color scheme? I assume the annotation color supports minihtml variables, or doesn't it?

Here is an example for the relevant code (not tested):
```jsonc
  // Show code actions:
  // Should be zero or more of the following:
  // "annotation" - show an annotation on the right when code actions are available.
  // "bulb" - show a bulb in the gutter when code actions are available (not clickable).
  "show_code_actions": ["annotation"],
```

```python
        if 'bulb' in global_settings.show_code_actions:
            scope = 'region.yellowish'
        if 'annotation' in global_settings.show_code_actions:
            code_actions_link = make_link('subl:lsp_code_actions', '{} code action{}'.format(action_count, suffix))
            annotations = ["<div class=\"actions\">{}</div>".format(code_actions_link)]
            annotation_color = 'var(--cyanish)'
        if global_settings.show_code_actions:
            self.view.add_regions(self.CODE_ACTIONS_KEY, regions, scope, icon, flags, annotations, annotation_color)
```

predragnikolic:
I agree with these things:
- using `region.yellowish` instead of `markup.changed`
- using theme colors instead of hardcoded color for annotation(if it is supported)
- I do like the shorter settings description, but maybe even without this line.
```diff
  // Show code actions:
-  // Should be zero or more of the following:
  // "annotation" - show an annotation on the right when code actions are available.
  // "bulb" - show a bulb in the gutter when code actions are available (not clickable).
```

As for specifying things as an array does allow flexibility, I just don't have an opinion about it. Nor I am again it, nor for it... :)
Will some people ever enable both? (I don't know).
Note: Even if the put an `[]` array, they wont see the the code actions in the bulb or annotation,
but the code action request will still be send on_selection_modified(which can be considered wasteful). Here are my thoughts in the [comment]( https://github.com/sublimelsp/LSP/pull/1196#issuecomment-660509846) above. 

Feel free to open a PR :)

rchl:
> using theme colors instead of hardcoded color for annotation(if it is supported)

It's not supported.

rchl:
See https://github.com/sublimehq/sublime_text/issues/3397

jwortmann:
Okay, in that case we can still use `view.style_for_scope("region.cyanish")["foreground"]` to get the corresponding color from the color scheme.

PR #1197, opened 2020-07-19
Signature Help improvements

rwols:
- Cooperate with AC as ST4 likes to pop up the AC more eagerly
- Put it all in documents.py
- Make sure dynamic registration works with sigHelp. Basically
  do not cache anything. Caching is hard.

Peek 2020-07-19 21-35


rchl:
Is it on purpose that signature popup disappears as soon as the user starts typing the argument name (or selects different completion), or is it due to limitation of ST?

I think ideally I would want to see the signature help as long as I'm editing arguments. (Although I would have to verify that this is really what I want by testing.)

plugin/documents.py
```diff
@@ -216,6 +265,95 @@ def on_hover(self, point: int, hover_zone: int) -> None:
...
+                # all using a popup. Most of these systems assume they have exclusive access to a popup, while in
+                # reality there is only one popup per view.
+                self.view.hide_popup()
+
+    def _on_signature_help(self, response: Optional[Dict], point: int) -> None:
+        region = self._get_current_region_async()
+        if region is None:
+            return
+        if point != region.b:
+            return
```
rchl, line 311:
```js
callFn(arg1
```
Let's say I've typed `arg1` and I want to add another argument. I type `,[space]` and that makes signature popup not show up if I type those fast. I believe it's due to this point check that no longer holders.
I can reproduce it easily in typescript server in a bigger project than in pyls with LSP project.

Screen Recording 2020-07-19 at 23 13 27

rwols: I see. Perhaps we can redo the checks earlier to see if it’s still applicable to show signature help.

rwols: And by “checks earlier” I mean the checks that were done before sending the request.

plugin/documents.py
```diff
@@ -216,6 +265,95 @@ def on_hover(self, point: int, hover_zone: int) -> None:
...
+        provider = None  # type: Optional[Dict[str, Any]]
+        session = None  # type: Optional[Session]
+        for sv in self.session_views_async():
+            provider = sv.session.get_capability("signatureHelpProvider")
+            if provider:
+                session = sv.session
+                break
+        if not provider or not session:
+            return
+        triggers = provider.get("triggerCharacters") or []
```
rchl, line 287: I believe this block could be simplified to:
```py
        session = self.session("signatureHelpProvider")
        if not session:
            return
        triggers = session.get_capability("signatureHelpProvider.triggerCharacters") or []
```
Since we are handling only one session anyway.

plugin/documents.py
```diff
@@ -216,6 +265,95 @@ def on_hover(self, point: int, hover_zone: int) -> None:
...
+
+            def render_sighelp_on_main_thread() -> None:
+                if self.view.is_popup_visible():
+                    self._update_popup(content)
+                else:
+                    self._show_popup(content, point)
+
+            sublime.set_timeout(render_sighelp_on_main_thread)
+
+    def _show_popup(self, content: str, point: int) -> None:
```
rchl, line 324: Now that various functionality is within the same class it would probably be better to make method names more unique like `_show_signature_popup`, `_update_signature_popup`, `on_hide_signature_popup`...

AmjadHD:
how do you navigate overloads ?

rchl:
Apart from those mentioned issues, really good change and something that you didn't realize you were missing.

rchl:
Closing of the popup on navigating the list was implemented for the "completion more info" popup and it applies for this popup too but in this case, it doesn't really make sense to close the popup on changing selected completion.

plugin/documents.py
rwols: I think this is fixed in f34e7c3. The point check is unnecessary.

rchl:
LGTM

rchl:
Apart from:
 - not being able to navigate signatures when completions popup is open
 - would prefer the signature helper to stay open when navigating completions

rwols:
> not being able to navigate signatures when completions popup is open

Yes. I played around with this but I feel like "taking over" the up/down arrows that a user normally uses for browsing completions is too intrusive. On the other hand, a user can close the popup by pressing <kbd>esc</kbd>. I am unsure what is better...

> would prefer the signature helper to stay open when navigating completions

This would definitely be an improvement :)

PR #1203, opened 2020-07-21
Try fixing out-of-sync diagnostics

rwols:
I'm still not quite sure what the problem is because I cannot repro. @rchl @ayoub-benali could you try out this branch? Otherwise, the problem has to be around that part of the code somewhere.

### DIFF
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -169,7 +169,6 @@ class SessionBuffer:
         if version is None:
             version = change_count
         if version == change_count:
-            diagnostics_version = version
             for index, diagnostic in enumerate(map(Diagnostic.from_lsp, raw_diagnostics)):
                 diagnostics.append(diagnostic)
                 data = data_per_severity.get(diagnostic.severity)
@@ -185,14 +184,14 @@ class SessionBuffer:
                     data.panel_contribution.append(format_diagnostic_for_panel(diagnostic))
                 if diagnostic.severity <= settings.auto_show_diagnostics_panel_level:
                     should_show_diagnostics_panel = True
-            self._publish_diagnostics_to_session_views(
-                diagnostics_version,
-                diagnostics,
-                data_per_severity,
-                total_errors,
-                total_warnings,
-                should_show_diagnostics_panel
-            )
+        self._publish_diagnostics_to_session_views(
+            version,
+            diagnostics,
+            data_per_severity,
+            total_errors,
+            total_warnings,
+            should_show_diagnostics_panel
+        )
 
     def _publish_diagnostics_to_session_views(
         self,


rchl:
There is this assignment in one place in this file:
```py
self.diagnostics_version = diagnostics_version
```
Why is it assigned? I don't see the class property being referenced anywhere.

rchl:
It doesn't fix the issue. I have a document now that I can easily reproduce by undoing and re-doing a change so I'll investigate it.

rchl:
The issue triggers in this scenario:
1) Add an error in the code
2) Wait 500ms (but not too long) for the document to be synced and server to publish diagnostics
3) Undo the change

This results in server making two diagnostic notifications - first with an error and second empty.

Now the problem starts in `on_diagnostics_async` which is now triggered for both notifications and `view.change_count()` is the same for both (because it was triggered async, after both changes were applied).

And then in `_publish_diagnostics_to_session_views()` that results in the wrong logic where first notification triggers:
```py
            else:
                debounced(
                    present,
                    timeout_ms=int(1000.0 * delay_in_seconds),
                    condition=lambda: self.view.is_valid() and self.view.change_count() == diagnostics_version,
                    async_thread=True
                )
```

and then almost immediately the second notification triggers this block:
```py
        if not bool(diagnostics) and not self.diagnostics_are_visible:
            # Nothing was previously visible, and nothing will become visible. So do nothing.
            pass
```

So the first notification with an error is not yet processed (I'm testing with `diagnostics_delay_ms: 2000`) and second notification is ignored because the code thinks there are no diagnostics visible for the view (there aren't by they are pending).

I would probably fix that by creating a version of `debounced` that can be instantiated and attached as a property so that we can call `cancel` on currently denounced call. Something like:

```py
class DebounceHandler:
  def denounce(f, timeout_ms, condition, async_thread):
    ...

  def cancel():
    ...
```

Then we would cancel any pending denounce at the beginning of `_publish_diagnostics_to_session_views()`.

plugin/session_buffer.py
```diff
@@ -185,14 +184,14 @@ def on_diagnostics_async(self, raw_diagnostics: List[Dict[str, Any]], version: O
...
-                should_show_diagnostics_panel
-            )
+        self._publish_diagnostics_to_session_views(
+            version,
+            diagnostics,
+            data_per_severity,
+            total_errors,
+            total_warnings,
+            should_show_diagnostics_panel
+        )
```
rchl, line 194: Now that we know the actual problem we should revert this. Looks like this could result in another issue where we would show diagnostics but with all counts being 0.

rwols:
This debounced handler sounds like a good idea. It could be used in many more places.

PR #1204, opened 2020-07-22
Put a little more effort in the symbol scope names

rwols:
The scopes are taken from http://www.sublimetext.com/docs/3/scope_naming.html

It wasn't clear to me which scopes belong to these symbol kinds:

- Array
- Object
- Key
- Enum Member (what is this?)
- Event

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -47,7 +47,7 @@ SYMBOL_KINDS = [
     ("Boolean",        "constant.language"),
     ("Array",          "meta.sequence"),  # Unsure about this one [scope taken from JSON.sublime-syntax]
     ("Object",         "meta.mapping"),  # Unsure about this one [scope taken from JSON.sublime-syntax]
-    ("Key",            "string meta.mapping.key"),  # Unsure about this one [scope taken from JSON.sublime-syntax]
+    ("Key",            "meta.mapping.key string"),  # Unsure about this one [scope taken from JSON.sublime-syntax]
     ("Null",           "constant.language"),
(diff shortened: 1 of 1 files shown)


predragnikolic:
I just looked up in the [TypeScript Docs](https://www.typescriptlang.org/docs/handbook/enums.html)
```
enum Direction {
    Up,
    Down,
    Left,
    Right,
}
```
the enum is Direction,
enum members are Up, Down, Left, Eight

plugin/core/views.py
```diff
@@ -25,34 +25,35 @@
...
-    ("Null",           "comment"),
-    ("Enum Member",    "comment"),
-    ("Struct",         "comment"),
-    ("Event",          "comment"),
-    ("Operator",       "comment"),
-    ("Type Parameter", "comment"),
+    ("Boolean",        "constant.language"),
+    ("Array",          "meta.sequence"),  # Unsure about this one [scope taken from JSON.sublime-syntax]
+    ("Object",         "meta.mapping"),  # Unsure about this one [scope taken from JSON.sublime-syntax]
+    ("Key",            "string meta.mapping.key"),  # Unsure about this one [scope taken from JSON.sublime-syntax]
```
jwortmann, line 50: This should probably be `meta.mapping.key string`

plugin/core/views.py
```diff
@@ -25,34 +25,35 @@
...
+    ("Null",           "constant.language"),
+    ("Enum Member",    "storage.type.enum keyword.declaration.enum"),  # ???
```
jwortmann, line 52: `constant.other.enum` is used by Java and C# in the default syntaxes

PR #1205, opened 2020-07-23
Add support for side-by-side views for go-to-symbol commands

rchl:


plugin/goto.py
```diff
@@ -63,18 +67,18 @@ def handle_response(self, response: Any) -> None:
             else:
                 locations = process_response_list(response)
             if len(locations) == 1:
-                open_location(window, locations[0])
+                open_location(window, locations[0], side_by_side)
             elif len(locations) > 1:
                 window.show_quick_panel(
                     items=locations,
-                    on_select=lambda x: select_entry(window, locations, x, view),
+                    on_select=lambda x: select_entry(window, locations, x, view, side_by_side),
```
rchl, line 74: When quick panel is used the flow is maybe a bit weird. The files are previewed using full window view and on selecting a file it's opened in side-by-side view. I guess it's not that bad.

Keymaps/Default (Linux).sublime-keymap
```diff
@@ -152,6 +152,7 @@
     // Go To Definition
     // {
     //     "command": "lsp_symbol_definition",
+    //     // "args": { "side_by_side": true },
```
rchl, line 155: Could also uncomment `args` and set to `false` by default. I guess less friction that way?

rwols: Yes, just uncomment `args`. Also, I formatted all the json with vscode-json-language-server, perhaps we can keep the formatting consistent?

rwols:
I'm kind of missing the possibility to choose "side-by-side" via the hover popup (in the actions section). I know ST uses some kind of unicode symbol for that. Can we do the same?

rchl:
> Can we do the same?

I'll have a look

rchl:
Turned out to be quite a refactor but for the better. :)

plugin/hover.py
```diff
@@ -184,12 +191,6 @@ def on_hover_navigate(self, href: str, point: int) -> None:
             sel.add(sublime.Region(point, point))
 
             self.view.show_popup_menu(titles, lambda i: self.handle_code_action_select(config_name, i))
-        elif href.startswith('location'):
```
rchl, line 187: This was unused.

stubs/sublime.pyi
```diff
@@ -266,6 +268,10 @@ def list_syntaxes() -> Sequence[Dict[str, Any]]:
     ...
 
 
+def command_url(cmd: str, args: Optional[dict] = ...) -> str:
```
rwols, line 271: This wasn't added until 4075 I think?

Does this mean we are forced to make a 4075-x.y.z branch now?

rwols:
The actions links are a great improvement!

stubs/sublime.pyi
rchl: No. Let's not be crazy :)

People who are on ST4 dev builds are on the latest version. There is no reason to be on an older version and there is no reason to keep compatibility with older dev builds. :)

PR #1207, opened 2020-07-24
Fix diagnostics getting out of sync when having diagnostics_delay_ms set

### LINKED ISSUE #1198: Diagnostics get out of sync
During editing of JSON file the diagnostics become out of sync with the real state of the file. Happened before in other servers and files also.

```
	"diagnostics_delay_ms": 2000,
	"diagnostics_additional_delay_auto_complete_ms": 1000,
```
`st4000-exploration` (b1a1fa79a71d215dde2d6c396b559561feaeef5f).

```
::  -> LSP-json textDocument/didChange: {'textDocument': {'uri': 'file:///project/template/.lokalise.json', 'version': 6}, 'contentChanges': [{'text': '        "filter_langs": ["bg", "de", "en", "es", "fr", "it", "no", "pl", "pt", "pt_BR"],\n', 'range': {'end': {'character': 0, 'line': 7}, 'start': {'character': 0, 'line': 7}}, 'rangeLength': 0}]}
:: <-  LSP-json textDocument/publishDiagnostics: {'diagnostics': [{'message': 'Duplicate object key', 'source': 'json', 'code': 520, 'range': {'end': {'character': 22, 'line': 8}, 'start': {'character': 8, 'line': 8}}, 'severity': 2}, {'message': 'Duplicate object key', 'source': 'json', 'code': 520, 'range': {'end': {'character': 22, 'line': 7}, 'start': {'character': 8, 'line': 7}}, 'severity': 2}], 'uri': 'file:///project/template/.lokalise.json'}
::  -> LSP-json textDocument/didChange: {'textDocument': {'uri': 'file:///project/template/.lokalise.json', 'version': 7}, 'contentChanges': [{'text': '', 'range': {'end': {'character': 0, 'line': 9}, 'start': {'character': 0, 'line': 8}}, 'rangeLength': 86}]}
:: --> LSP-json textDocument/documentColor(10): {'textDocument': {'uri': 'file:///project/template/.lokalise.json'}}
:: <<< LSP-json 10: []
:: <-  LSP-json textDocument/publishDiagnostics: {'diagnostics': [], 'uri': 'file:///project/template/.lokalise.json'}
```

Screenshot 2020-07-20 at 12 53 10

As clear from the log, the last state from the server reported no diagnostics but ST is still showing the one that was reported before that.

rchl:
Resolves #1198

### DIFF
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -5,6 +5,7 @@ from .core.protocol import TextDocumentSyncKindNone
 from .core.sessions import SessionViewProtocol
 from .core.settings import settings
 from .core.types import debounced
+from .core.types import Debouncer
 from .core.typing import Any, Iterable, Optional, List, Dict
 from .core.views import DIAGNOSTIC_SEVERITY
 from .core.views import did_change
@@ -75,6 +76,7 @@ class SessionBuffer:
         self.total_errors = 0
         self.total_warnings = 0
         self.should_show_diagnostics_panel = False
+        self.diagnostics_debouncer = Debouncer()
 
     def __del__(self) -> None:
         mgr = self.session.manager()
@@ -217,6 +219,8 @@ class SessionBuffer:
                 should_show_diagnostics_panel
             )
 
+        self.diagnostics_debouncer.cancel_pending()
+
         if not bool(diagnostics) and not self.diagnostics_are_visible:
             # Nothing was previously visible, and nothing will become visible. So do nothing.
             pass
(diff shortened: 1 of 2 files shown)


rchl:
I still see some syncing issue. Will investigate

plugin/session_buffer.py
```diff
@@ -231,7 +235,7 @@ def present() -> None:
             if delay_in_seconds <= 0.0:
                 present()
```
rwols, line 236: Shouldn't we also cancel here?

rchl: We cancel earlier for all the branches so no? Or what do you mean? :)

rwols: You're right, I did not notice the above cancellation above.

rchl:
This fixes the original issue I was having but there is another one as mentioned in #1198.
I think it would be fine to merge this already anyway and work on another issue in separate PR.

PR #1210, opened 2020-07-25
Rename LspTextCommand.session -> LspTextCommand.best_session

rwols:
Done with pyls' now correctly working textDocument/rename, pretty nice.

All tests pass and tox doesn't complain, so I'm assuming everything went well.

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -59,7 +59,7 @@ class LspResolveDocsCommand(LspTextCommand):
         webbrowser.open(url)
 
     def do_resolve(self, item: dict) -> None:
-        session = self.session('completionProvider.resolveProvider')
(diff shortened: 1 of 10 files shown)


rchl:
Love the idea of not using generic "session" name for a method.

Can we also rename `LSPViewEventListener.session`?

rchl:
Maybe consider keeping the old method name as an alias for a few releases (with a deprecation message)?

rwols:
> Maybe consider keeping the old method name as an alias for few releases (with a deprecation message)?

For who? Nobody said you could use `LspTextCommand` :)

rchl:
> For who? Nobody said you could use `LspTextCommand` :)

But that doesn't mean that nobody did. :)
But maybe better to just visibly break those cases.

PR #1211, opened 2020-07-25
API: Get session by name

rwols:
Will be useful for e.g. LSP-metals and LSP-julia so that their commands are hidden in the command palette when there's no session with that name attached to the view.

### DIFF
--- a/plugin/core/registry.py
+++ b/plugin/core/registry.py
@@ -84,13 +84,27 @@ def get_position(view: sublime.View, event: Optional[dict] = None) -> int:
 
 
 class LspTextCommand(sublime_plugin.TextCommand):
+    """
+    Inherit from this class to define your requests that should be triggered via the command palette and/or a
+    keybinding.
+    """
 
+    # When this is defined in a derived class, the command is enabled if and only if there exists a session attached
+    # to the view that has the given capability. When both `capability` and `session_name` are defined, `capability`
+    # wins.
     capability = ''
+
+    # When this is defined in a derived class, the command is enabled if and only if there exists a session attached
+    # to the view that has the given name. When both `capability` and `session_name` are defined, `capability` wins.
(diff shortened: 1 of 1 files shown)


plugin/core/registry.py
```diff
@@ -86,11 +86,15 @@ def get_position(view: sublime.View, event: Optional[dict] = None) -> int:
 class LspTextCommand(sublime_plugin.TextCommand):
 
     capability = ''
+    session_name = ''
```
rchl, line 89: Since you want to make it a public API, can we add comments here explaining the relationship between those two (`capability` wins if set, `session_name` only checked if `capability` is not set)?

Or maybe it should rather work so that both are checked always (when set)? So that we can restrict the command to both at the same time?

rwols: > Or maybe it should rather work so that both are checked always (when set)? So that we can restrict the command to both at the same time?

Well, a helper package knows exactly which capabilities the server has, so this seems redundant.

rchl: I guess I don't have any counter-arguments for that :)

Just that it wouldn't hurt to make things more explicit and would also make the logic more straight-forward (all respected when set rather than one having higher priority when set).

plugin/core/registry.py
```diff
@@ -84,13 +84,26 @@ def get_position(view: sublime.View, event: Optional[dict] = None) -> int:
...
+    keybinding.
+    """
 
+    # When this is defined in a derived class, the command is enabled if and only if there exists a session attached
+    # to the view that has the given capability.
     capability = ''
 
+    # When this is defined in a derived class, the command is enabled if and only if there exists a session attached
+    # to the view that has the given name.
+    session_name = ''
```
rchl, line 98: It's still not explained that this is ignored when `capability` is set.

PR #1214, opened 2020-07-25
Replace diagnostic status bar with annotations

rwols:
A diagnostic annotation appears when the (first) region of the
selection intersects with a diagnostic.

This needs a bit more testing.

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -103,7 +103,6 @@ class ColorSchemeScopeRenderer:
 class DocumentSyncListener(LSPViewEventListener, AbstractViewListener):
 
     CODE_ACTIONS_KEY = "lsp_code_action"
-    ACTIVE_DIAGNOSTIC = "lsp_active_diagnostic"
     code_actions_debounce_time = 800
     color_boxes_debounce_time = 500
     highlights_debounce_time = 500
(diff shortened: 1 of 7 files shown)


rchl:
Immediately noticed that it fights with code action annotations. In the TS server that I was testing the code action typically (always?) exists for error diagnostic.

rwols:
That's true, but the code actions arrive later so get shown. Does that not happen for typescript-language-server?

rchl:
Error diagnostic appears for a split second and then gets replaced by code action.

rwols:
They're not replaced. You can click on the annotation and it will expand to a list of all the annotations on that line. I don't whether that's good UX though.

rchl:
Ah, I see now. Hovering the annotation is enough to see both.

Yeah, not a great UX but it would probably be acceptable if the diagnostic annotation would win over the code actions one. It's more relevant to see that then the code actions link (which you can still access by hovering annotation).

plugin/documents.py
```diff
@@ -365,7 +358,7 @@ def _on_code_actions(self, responses: CodeActionsByConfigName) -> None:
             suffix = 's' if action_count > 1 else ''
             code_actions_link = make_command_link('lsp_code_actions', '{} code action{}'.format(action_count, suffix))
             annotations = ["<div class=\"actions\">{}</div>".format(code_actions_link)]
-            annotation_color = '#2196F3'
+            annotation_color = self.view.style_for_scope("region.bluish")["foreground"]
```
rchl, line 361: Would probably make sense to cache it per-listener instance.

rwols:
I don't think we'll go this route.

rchl:
I still have some hope for it, I think. If only the diagnostics would also have priority over code actions...

PR #1223, opened 2020-07-28
Remove 'cquery' from the default configs

rwols:


rchl:
Not useful to have a default config for this one?

rwols:
The last commit in cquery was about 2 years ago. ccls is a fork of cquery and works well.

PR #1224, opened 2020-07-28
Remove intelephense from the default configs

rwols:
And point people to LSP-intelephense instead.

I also removed this "Felix Becker language server" because it's dead.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -584,22 +584,9 @@ Alternatively you can use the less maintained [lua-lsp](https://github.com/Alloy
 
 #### Intelephense
 
-```sh
-npm i intelephense -g
-```
-
-See [bmewburn/intelephense-docs](https://github.com/bmewburn/intelephense-docs)
-
-
-#### PHP Language server
-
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -584,22 +584,9 @@ Alternatively you can use the less maintained [lua-lsp](https://github.com/Alloy
...
-See: [felixfbecker/php-language-server](https://github.com/felixfbecker/php-language-server)
-
-Global installation:
-
-1. modify `~/.composer/composer.json` to set `"minimum-stability": "dev"` and `"prefer-stable": true`
-2. run `composer global require felixfbecker/language-server`
-3. run `composer run-script --working-dir=~/.composer/vendor/felixfbecker/language-server parse-stubs`
+1. Make sure you have `node` installed.
+2. Install the [LSP-intelephense](https://packagecontrol.io/packages/LSP-intelephense) helper package. It installs
+   and updates the server for you in $DATA/Cache.
```
rchl, line 589: Since there are quite a few helper packages now, maybe we could provide info about them all in more condensed and generic form like a table.

PR #1225, opened 2020-07-28
Remove all the ts/js stuff from the default configs

rwols:
And point people to LSP-typescript and LSP-eslint instead.


rchl:
A breaking change. Maybe we could highlight this in the changelog as breaking change?

rwols:
> Maybe we could highlight this in the changelog as breaking change?

Yes, agreed.

PR #1231, opened 2020-08-07
Fix compatibility with ST 4080+

rchl:
I've rushed it a bit so not sure if it's the best approach.
It can be scraped and redone if necessary.

(Have not created proper types for `Buffer` - just made it `Any`.)


rchl:
We could share the same `TextChangeListener` instance for multiple views with same buffer_id but I'm not sure if that optimization is really worth the trouble.

rwols:
Thanks for the quick fix :)

PR #1232, opened 2020-08-08
Pass copy of configuration to on_workspace_configuration

rchl:
Since the purpose of the "on_workspace_configuration" is for plugins to
modify the dict, don't send original dict as plugin can't possibly keep
track of what was modified and remove modifications on next API call.

### DIFF
--- a/plugin/core/collections.py
+++ b/plugin/core/collections.py
@@ -2,6 +2,7 @@
 Module with additional collections.
 """
 from .typing import Optional, Dict, Any
+from copy import deepcopy
 
 
 class DottedDict:
@@ -73,6 +74,19 @@ class DottedDict:
                 return
             current = next_current
         current.pop(keys[-1], None)
+
+    def copy(self, path: Optional[str] = None) -> Any:
+        """
+        Get a copy of the value from the dictionary or copy of whole dictionary.
+
+        :param      path:  The path, e.g. foo.bar.baz, or None.
+
+        :returns:   A copy of the value stored at the path, or None if it doesn't exist.
+                    Note that this cannot distinguish between None values and
+                    paths that don't exist. If the path is None, returns a copy of the
+                    entire dictionary.
+        """
+        return deepcopy(self.get(path))
 
     def __bool__(self) -> bool:
         """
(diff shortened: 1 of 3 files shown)


plugin/core/sessions.py
```diff
@@ -731,6 +732,7 @@ def m_workspace_configuration(self, params: Dict[str, Any], request_id: Any) ->
         for requested_item in requested_items:
             configuration = self.config.settings.get(requested_item.get('section') or None)
             if self._plugin:
+                configuration = deepcopy(configuration)
```
rwols, line 735: Perhaps it’s time to add a dedicated copy method to ClientConfig because we’re falling into this trap in a few more places I think. Please add a copy method.

rchl: `self.config.settings` is a DottedDict but `configuration` is a plain dict here. Are you saying we should copy the whole dotted dict and then get the `section` from that? That will be slightly less efficient.

rchl: On the other hand `section` is typically an empty string...

rchl:
Tests currently fail due to https://github.com/sublimehq/sublime_text/issues/3545
I've added try/except around the `detach()` call to be able to test.

plugin/core/sessions.py
rwols: Oh, I meant ClientConfig :D but it's good to have a copy method for DottedDict too.

rchl: You're asking to have a way to copy a ClientConfig, including deepcopy'ing all its properties? Why would that be useful?

rwols: Well, each window should have its own set of ClientConfigs because of project overrides.

plugin/core/collections.py
```diff
@@ -74,6 +75,19 @@ def remove(self, path: str) -> None:
...
+        Get a copy of the value from the dictionary or copy of whole dictionary.
+
+        :param      path:  The path, e.g. foo.bar.baz, or None.
+
+        :returns:   A copy of the value stored at the path, or None if it doesn't exist.
+                    Note that this cannot distinguish between None values and
+                    paths that don't exist. If the path is None, returns a copy of the
+                    entire dictionary.
+        """
+        return deepcopy(self.get(path))
```
rwols, line 89: Will this work OK if a value is involved? Say that the dict is `{"a": {"b" True}}` and we do `copy("a.b")` ?

rchl: Yes:
```
>>> import copy; copy.deepcopy(True)
True
```

PR #1233, opened 2020-08-09
[Improved Highlights] don't clear highlight on each cursor move

predragnikolic:
if the cursor is moving and it stays in the already highlighted regions,
the highlight won't be cleared.

Before:

output

After:

output

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -152,7 +152,6 @@ class DocumentSyncListener(LSPViewEventListener, AbstractViewListener):
         self._stored_region = sublime.Region(-1, -1)
         self._color_phantoms.update([])
         self.view.erase_status(AbstractViewListener.TOTAL_ERRORS_AND_WARNINGS_STATUS_KEY)
-        self._clear_highlight_regions()
         self._clear_session_views_async()
 
     # --- Implements AbstractViewListener ------------------------------------------------------------------------------
@@ -213,13 +212,16 @@ class DocumentSyncListener(LSPViewEventListener, AbstractViewListener):
             yield sv.session_buffer
 
     def on_text_changed_async(self, changes: Iterable[sublime.TextChange]) -> None:
-        self._clear_highlight_regions()
         different, current_region = self._update_stored_region_async()
         if self.view.is_primary():
             for sv in self.session_views_async():
                 sv.on_text_changed_async(changes)
         if not different:
             return
+        if "documentHighlight" not in global_settings.disabled_capabilities:
+            self._clear_highlight_regions()
+            self._when_selection_remains_stable_async(self._do_highlights, current_region,
+                                                      after_ms=self.highlights_debounce_time)
         if "colorProvider" not in global_settings.disabled_capabilities:
             self._when_selection_remains_stable_async(self._do_color_boxes_async, current_region,
                                                       after_ms=self.color_boxes_debounce_time)
@@ -249,7 +251,6 @@ class DocumentSyncListener(LSPViewEventListener, AbstractViewListener):
     def on_selection_modified_async(self) -> None:
         different, current_region = self._update_stored_region_async()
         if different:
-            self._clear_highlight_regions()
             self._clear_code_actions_annotation()
             if "documentHighlight" not in global_settings.disabled_capabilities:
                 self._when_selection_remains_stable_async(self._do_highlights, current_region,
@@ -419,11 +420,22 @@ class DocumentSyncListener(LSPViewEventListener, AbstractViewListener):
         for kind in global_settings.document_highlight_scopes.keys():
             self.view.erase_regions("lsp_highlight_{}".format(kind))
 
+    def is_in_higlighted_region(self, point: int) -> bool:
+        for kind in global_settings.document_highlight_scopes.keys():
+            regions = self.view.get_regions("lsp_highlight_{}".format(kind))
+            for r in regions:
+                if r.contains(point):
+                    return True
+        return False
+
     def _do_highlights(self) -> None:
-        self._clear_highlight_regions()
         if len(self.view.sel()) != 1:
             return
         point = self.view.sel()[0].begin()
+        if self.is_in_higlighted_region(point):
+            return
+        else:
+            self._clear_highlight_regions()
         session = self.session("documentHighlightProvider", point)
         if session:
             params = text_document_position_params(self.view, point)
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -153,6 +153,7 @@ class DocumentSyncListener(LSPViewEventListener, AbstractViewListener):
         self._color_phantoms.update([])
         self.view.erase_status(AbstractViewListener.TOTAL_ERRORS_AND_WARNINGS_STATUS_KEY)
         self._clear_session_views_async()
+        self._clear_highlight_regions()
 
     # --- Implements AbstractViewListener ------------------------------------------------------------------------------
 
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -252,10 +252,12 @@ class DocumentSyncListener(LSPViewEventListener, AbstractViewListener):
     def on_selection_modified_async(self) -> None:
         different, current_region = self._update_stored_region_async()
         if different:
-            self._clear_code_actions_annotation()
             if "documentHighlight" not in global_settings.disabled_capabilities:
-                self._when_selection_remains_stable_async(self._do_highlights, current_region,
-                                                          after_ms=self.highlights_debounce_time)
+                if not self.is_in_higlighted_region(current_region.b):
+                    self._clear_highlight_regions()
+                    self._when_selection_remains_stable_async(self._do_highlights, current_region,
+                                                              after_ms=self.highlights_debounce_time)
+            self._clear_code_actions_annotation()
             self._when_selection_remains_stable_async(self._do_code_actions, current_region,
                                                       after_ms=self.code_actions_debounce_time)
             self.update_diagnostic_in_status_bar_async()
@@ -433,9 +435,6 @@ class DocumentSyncListener(LSPViewEventListener, AbstractViewListener):
         if len(self.view.sel()) != 1:
             return
         point = self.view.sel()[0].begin()
-        if self.is_in_higlighted_region(point):
-            return
-        self._clear_highlight_regions()
         session = self.session("documentHighlightProvider", point)
         if session:
             params = text_document_position_params(self.view, point)
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -435,8 +435,7 @@ class DocumentSyncListener(LSPViewEventListener, AbstractViewListener):
         point = self.view.sel()[0].begin()
         if self.is_in_higlighted_region(point):
             return
-        else:
-            self._clear_highlight_regions()
+        self._clear_highlight_regions()
         session = self.session("documentHighlightProvider", point)
         if session:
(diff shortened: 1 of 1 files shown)


plugin/documents.py
```diff
@@ -152,7 +152,6 @@ def __del__(self) -> None:
         self._stored_region = sublime.Region(-1, -1)
         self._color_phantoms.update([])
         self.view.erase_status(AbstractViewListener.TOTAL_ERRORS_AND_WARNINGS_STATUS_KEY)
-        self._clear_highlight_regions()
```
predragnikolic, line 155: I removed this as I didn't see why there is a need to clear highlights in the `__del__` method. Should I put it back?

rwols: It's so that when the plugin is disabled we clean up our trash (not everything gets cleaned up still, but we're getting there :) ) Please keep this line in the `__del__`.

plugin/documents.py
```diff
@@ -213,13 +212,16 @@ def session_buffers_async(self) -> Generator[SessionBuffer, None, None]:
...
         different, current_region = self._update_stored_region_async()
         if self.view.is_primary():
             for sv in self.session_views_async():
                 sv.on_text_changed_async(changes)
         if not different:
             return
+        if "documentHighlight" not in global_settings.disabled_capabilities:
+            self._clear_highlight_regions()
+            self._when_selection_remains_stable_async(self._do_highlights, current_region,
+                                                      after_ms=self.highlights_debounce_time)
```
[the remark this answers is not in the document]
predragnikolic: Although ST fires both `on_text_changed_async` and `on_selection_modified_async` when we type,
There is this fn call to `_update_stored_region_async` in `on_selection_modified_async`,
that fn will for some reason return `different == False`, so we won't go into the if block and do the highlight.

```python
    def on_selection_modified_async(self) -> None:
        different, current_region = self._update_stored_region_async()
        if different:  #  different == False
            self._clear_code_actions_annotation()
            if "documentHighlight" not in global_settings.disabled_capabilities:
                self._when_selection_remains_stable_async(self._do_highlights, current_region,
                                                          after_ms=self.highlights_debounce_time)
```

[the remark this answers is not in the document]
predragnikolic: Here is the current behavior(st-4000 branch)

output
 
Note, when I stop typing,
the documentHighlight is not send, until I move the selection (press left, right)... I am not sure if this behaved like this in the past?

But with my PR, i send the documentHighlight on the `on_text_changed_async` so it will highlight the word when I stop typing

plugin/documents.py
predragnikolic: I added it back.
I wasn't aware that the `__del__` method was somehow related to the plugin_unload event. :)

rchl: Session shutdown which can be trigger with "restart servers" also

plugin/documents.py
```diff
@@ -213,13 +213,16 @@ def session_buffers_async(self) -> Generator[SessionBuffer, None, None]:
...
```
rchl, line 225: Duplicates existing code. Makes sense to create a helper function for it.

plugin/documents.py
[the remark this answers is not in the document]
rchl: Note that doing this from both `on_text_changed_async` and `on_selection_modified_async` will trigger request twice once "selection_remains_stable".

We could use `Debouncer` class to avoid that.

plugin/documents.py
```diff
@@ -419,11 +421,22 @@ def _clear_highlight_regions(self) -> None:
...
+            regions = self.view.get_regions("lsp_highlight_{}".format(kind))
+            for r in regions:
+                if r.contains(point):
+                    return True
+        return False
+
     def _do_highlights(self) -> None:
-        self._clear_highlight_regions()
         if len(self.view.sel()) != 1:
             return
```
rchl, line 434: Should we also clear highlight in case sel != 1 to match previous behavior?

plugin/documents.py
```diff
@@ -419,11 +421,22 @@ def _clear_highlight_regions(self) -> None:
...
         point = self.view.sel()[0].begin()
+        if self.is_in_higlighted_region(point):
+            return
+        else:
```
rchl, line 438: No need for `else` after return :)

plugin/documents.py
predragnikolic: I believe that check exist so the line `point = self.view.sel()[0].begin()` don't throw an Index Error when trying to access the first item but the `view.sel()` returned an empty list...

I think that the condition should just be:
```python
    if not len(self.view.sel()):
            return
   point = self.view.sel()[0].begin()
```

rchl: But it doesn't check that `len == 0`. I think it's supposed to also handle multiple cursors in which case we don't want to highlight matches.

plugin/documents.py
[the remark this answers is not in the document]
predragnikolic: > Note that doing this from both on_text_changed_async and on_selection_modified_async will trigger request twice once "selection_remains_stable".

It would make sense that it behaves like that, but that is not how logic currently behaves. 
Currently there wont be two request. 
Only one will be send and that is by on_text_changed_async. 
Because the on_selection_modified_async wont pass the `if different` [check](https://github.com/sublimelsp/LSP/pull/1233/files#diff-cbc07fb64262a8d87f66bb8f5fb32e69R254) .`different` will be `False`.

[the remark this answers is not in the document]
predragnikolic: I put two print statements on both `on_text_changed_async` and `on_selection_modified_async`
```python
   def on_text_changed_async(self, changes: Iterable[sublime.TextChange]) -> None:
        different, current_region = self._update_stored_region_async()
        print('on_text_changed_async fired') # here
        if self.view.is_primary():
                sv.on_text_changed_async(changes)
        if not different:
            return

        print('on_text_changed_async, debounce request')  # here
        if "documentHighlight" not in global_settings.disabled_capabilities:
            self._when_selection_remains_stable_async(self._do_highlights, current_region,
                                                      after_ms=self.highlights_debounce_time)

   def on_selection_modified_async(self) -> None:
        different, current_region = self._update_stored_region_async()
        print('on_selection_modified_async fired')  # here
        if different:
            print('on_selection_modified_async, debounce request')   # here
            self._clear_code_actions_annotation()
            if "documentHighlight" not in global_settings.disabled_capabilities:
                self._when_selection_remains_stable_async(self._do_highlights, current_region,
                                                          after_ms=self.highlights_debounce_time)
```

Logs where I just modify the selection:
```
# pressed left
on_selection_modified_async fired
on_selection_modified_async, debounce request
# pressed left
on_selection_modified_async fired
on_selection_modified_async, debounce request
# pressed left
on_selection_modified_async fired
on_selection_modified_async, debounce request
```
Logs where I type:
```
# typed a
on_text_changed_async fired
on_text_changed_async, debounce request
on_selection_modified_async fired
# typed b
on_text_changed_async fired
on_text_changed_async, debounce request
on_selection_modified_async fired
# typed a
on_text_changed_async fired
on_text_changed_async, debounce request
on_selection_modified_async fired
```
NOTE that on_selection_modified_async, debounce request is not fired when I type

rchl:
What I don't like about this change is that on moving selection out of the highlighted region, the regions are not immediately cleared.

I'm not sure if that's bad or I'm just not used to it and it's fine to change it like that.

plugin/documents.py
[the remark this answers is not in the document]
rchl: OK, good. I haven't checked it myself, just assumed that it would not work correctly.

predragnikolic:
> What I don't like about this change is that on moving selection out of the highlighted region.

Thanks for the feedback, I agree. I will change this behavior.

plugin/documents.py
predragnikolic: I would leave it as it is currently. 
But I will add a new function that will wrap this code if you really think it is needed. (in that case could you give me a name for that function, because I can't think of any good one, `schedule_highlight` was the best one I could think of) :)

rchl: Since one place checks `is_in_higlighted_region` and one not, I'm no longer sure if it's worth creating a helper for them.

plugin/documents.py
```diff
@@ -419,9 +423,16 @@ def _clear_highlight_regions(self) -> None:
...
+            regions = self.view.get_regions("lsp_highlight_{}".format(kind))
+            for r in regions:
+                if r.contains(point):
+                    return True
+        return False
+
     def _do_highlights(self) -> None:
-        self._clear_highlight_regions()
-        if len(self.view.sel()) != 1:
+        if not len(self.view.sel()):
```
rchl, line 435: I don't think we should allow multiple cursors unless we really support that by highlighting all different words under all cursors.

predragnikolic: Here is how VS Code handles it: 
image

We had the same issue with the completion request, because we can trigger AC, but the cursors can be on different places (symbols).
In that case the request is only send to the primary cursor(in our case that is the view.sel()[0].b point)

So that is the reason why I think this logic should stay.
If you still believe that we should handle this differently, I will revert the commit. :)

rwols:
This is pretty cool overall and a nice addition. I did notice that if you make a non-empty selection and go "outside" an existing region with the non-empty selection, there seems to be a logic mismatch somewhere between checking for region.begin() / region.end() / region.a / region.b.

plugin/documents.py
rchl: Completions are a bit different case. With multiple cursors the completion from the "primary" cursor is inserted at all cursor positions. So that is fine.

Here you have multiple cursors but only the primary one does trigger selection while other cursors don't have any function so I think it might be a bit confusing in the current state. I think it would be better to not select anything when there are multiple cursors. Like before

plugin/documents.py
```diff
@@ -152,8 +152,8 @@ def __del__(self) -> None:
         self._stored_region = sublime.Region(-1, -1)
         self._color_phantoms.update([])
         self.view.erase_status(AbstractViewListener.TOTAL_ERRORS_AND_WARNINGS_STATUS_KEY)
-        self._clear_highlight_regions()
         self._clear_session_views_async()
+        self._clear_highlight_regions()
```
rchl, line 156: The order of lines has changed here and probably there is no reason for it?
(Not a big deal but nonetheless an unnecessary change from git history perspective.)

plugin/documents.py
```diff
@@ -419,11 +423,18 @@ def _clear_highlight_regions(self) -> None:
         for kind in global_settings.document_highlight_scopes.keys():
             self.view.erase_regions("lsp_highlight_{}".format(kind))
 
+    def is_in_higlighted_region(self, point: int) -> bool:
```
rchl, line 426: For consistency can you prefix this method with an underscore?

plugin/documents.py
rchl: But feel free to leave as is. It's not that important.

plugin/documents.py
predragnikolic: 2afd270

plugin/documents.py
predragnikolic: fdb8e7f done :)

plugin/documents.py
predragnikolic: I would leave it just because of this:
When someone has multi cursors and edits the selections, I want to allow the highlight request. 

output

If I were to leave the highlight request to be send only where there is only one selection, in that case we would not see the highlights. Here is an example for `if len(self.view.sel()) != 1:`:
output

PR #1234, opened 2020-08-09
Remove unnecessary log message when websocket server is not created

rchl:
There will already be a relevant log message earlier in the console if
_ws_server server was failed to be created so there is no need to spam
console with more logs.


rchl:
This will trigger when having two instances of ST with LSP enabled running.

PR #1238, opened 2020-08-09
Don't close the transport and server on it writing empty line to stderr

### LINKED ISSUE #1237: Server is being shutdown on server sending empty stderr line
When server triggers stderr output that is an empty string (or becomes an empty string after `rstrip` then LSP closes the transports and thus the server.

Issue found when analyzing https://github.com/sublimelsp/LSP-angular/issues/1

rchl:
Resolves #1237


rwols:
Thanks for investigating this, @ihodev and @rchl

rwols:
As it turns out, this can make the stderr loop spin forever. I've reverted it for now so we can think of another solution :)

PR #1241, opened 2020-08-10
Fix server being shut down on it writing an empty line to stderr

### LINKED ISSUE #1237: Server is being shutdown on server sending empty stderr line

rchl:
Resolves #1237

### DIFF
--- a/plugin/core/transports.py
+++ b/plugin/core/transports.py
@@ -157,9 +157,9 @@ class JsonRpcTransport(Transport):
     def _stderr_loop(self) -> None:
         try:
             while self._stderr:
-                message = self._stderr.readline().decode('utf-8', 'replace').rstrip()
-                if not message:
+                if self._closed:
                     break
+                message = self._stderr.readline().decode('utf-8', 'replace').rstrip()
                 callback_object = self._callback_object()
                 if callback_object:
                     callback_object.on_stderr_message(message)


plugin/core/transports.py
```diff
@@ -157,9 +157,9 @@ def _write_loop(self) -> None:
     def _stderr_loop(self) -> None:
         try:
             while self._stderr:
-                message = self._stderr.readline().decode('utf-8', 'replace').rstrip()
-                if not message:
+                if self._closed:
```
rchl, line 160: Checking property set from another thread but I think there shouldn't be any race issues in this case.

rchl: Hmm, this doesn't work reliably. Seems like the thread is starting before that property has been created. Would maybe make sense to start those outside of `__init__`

rwols:
Looks good to me!

plugin/core/transports.py
rwols: Well, just move `self._closed = False` to the top of the `__init__` ?

rchl: Yes. can do that, but wanted to be "proper". Don't like doing things in __init__ that rely on class being fully initialized.

rwols: If the thread creation is the last thing that happens in `__init__` then it seems to me it is no different than a separate, and unnecessary, `start` method

rchl: In practice, most likely yes. But making intentions clear and potentially avoid future errors (since people tend to add new attributes at the end), it might make sense to make it explicit.

rwols: If the class were to have a `start` method, it would be theoretically possible to not call `start` for whatever reason. So it would be like a class having an `is_valid` method, I find those types of classes bad design (that includes the `Thread` class from the `threading` module :) )

Either the entire sequence of starting and maintaing threads should have succeeded, or you don't have an object.

PR #1245, opened 2020-08-10
Improve script for parsing VSCode's configuration

rchl:
- Sort keys by name
- Try to not mess up original description line-wrapping too much
- Pretty-print default dictionary values
- Don't convert default values to ascii


rchl:
Was testing this on https://github.com/vuejs/vetur/blob/master/package.json and the especially-problematic setting `vetur.completion.scaffoldSnippetSources`.

### Original formatting:
Screenshot 2020-08-10 at 22 27 58

### New formatting:
Screenshot 2020-08-10 at 22 28 38

PR #1246, opened 2020-08-10
TextChangeListener improvements

### LINKED ISSUE #1239: textDocument/didChange not sent in some cases when dealing with cloned views
1. Open a file and create a clone of it
2. Focus the cloned view
3. Restart ST so that language server only starts for the cloned view
4. Make an edit

**Expected**: The `textDocument/didChange` is sent (Note: make sure that server wants to receive it in the first place)

**Actual**: No didChange notification is triggered and state becomes out of sync

The root reason is that `DocumentSyncListener.on_session_initialized_async` is not called for the primary view (because it wasn't focused) and since the `DocumentSyncListener.on_text_changed_async()` only sends didChange notification for primary view, it won't.

Also, since `TextChangeListener` only attaches itself from `DocumentSyncListener.on_session_initialized_async`, the `DocumentSyncListener.on_text_changed_async()` won't even be called but even if it would, it wouldn't help due to the root issue.

rwols:
- It threw an exception when doing LSP: Restart Servers
- It threw an exception when modifying a .sublime-project
- Add is_applicable
- Let ST create and destroy the TextChangeListeners
- Do bookkeeping via weaksets and avoid double buffers for cloned views
- Fix a case where a clone view may be registered before the primary view (fixes #1239)

The tests make the plugin_host crash, unfortunately...


rchl:
> The tests make the plugin_host crash, unfortunately...

Also on closing global Preferences window.

plugin/documents.py
```diff
@@ -494,7 +506,15 @@ def _register_async(self) -> None:
         file_name = self.view.file_name()
         if file_name:
             self._file_name = file_name
-            self.manager.register_listener_async(self)
+            buf = self.view.buffer()
+            if buf is not None:
+                text_change_listener = TextChangeListener.ids_to_listeners.get(buf.buffer_id)
+                if text_change_listener is not None:
```
rchl, line 512: Why such verbose if-check (also for the buffer above)? Since we are dealing with class instances, an `if foo` should be enough (unless it implements some weird `__bool__ ` I guess but can we assume that it's not?).

plugin/documents.py
```diff
@@ -494,7 +506,15 @@ def _register_async(self) -> None:
...
+                    text_change_listener.view_listeners.add(self)
+                    self._text_change_listener = ref(text_change_listener)
+                    self.manager.register_listener_async(self)
+                    return
+            debug("not tracking bufferless view", self.view.id())
```
rchl, line 517: Will also trigger for "TextChangeListener-less" view, in theory.

plugin/documents.py
```diff
@@ -267,9 +280,8 @@ def on_selection_modified_async(self) -> None:
             self.update_diagnostic_in_status_bar_async()
 
     def on_post_save_async(self) -> None:
-        if self.view.is_primary():
-            for sv in self.session_views_async():
-                sv.on_post_save_async()
+        for sv in self.session_views_async():
+            sv.on_post_save_async()
```
rchl, line 284: The notifications triggered in SessionBuffer (like DidSave) will trigger twice for cloned views now.

plugin/documents.py
```diff
@@ -103,29 +112,38 @@ def _wrap_with_scope_style(self, content: str, scope: str, emphasize: bool = Fal
...
 
-    # Created to have symmetry with "attach". Weird naming in internal API...
-    def detach(self) -> None:
-        self.remove()
+    ids_to_listeners = WeakValueDictionary()  # type: WeakValueDictionary[int, TextChangeListener]
+
+    @classmethod
+    def is_applicable(cls, buffer: sublime.Buffer) -> bool:
+        views = buffer.views()
+        if views:
```
rchl, line 121: Is the return value of `views()` not always a list? We can't just directly iterate?

rchl:
Do you plan to create an ST issue for the crash?

rwols:
I don't know how to make a minimal example out of it. But, crashes should turn into crash reports which get sent to SublimeHQ. So if we just spam them by running tests all the time... ;)

plugin/documents.py
rwols: I decided that it was better to "force-register" clones, if there exist any, instead of this old approach from 2 days ago.

plugin/documents.py
```diff
@@ -103,29 +112,34 @@ def _wrap_with_scope_style(self, content: str, scope: str, emphasize: bool = Fal
...
             listener.on_reload_async()
 
     def on_revert_async(self) -> None:
-        listener = self.listener()
-        if listener:
+        for listener in list(self.view_listeners):
             listener.on_revert_async()
 
+    def __repr__(self) -> str:
+        return str(self.buffer.buffer_id)
```
rchl, line 141: If this is meant to help debugging then it would also worth adding that it's a TextChangeListener. Like
```
'TextChangeListener({}}'.format(self.buffer.buffer_id)
```

plugin/documents.py
```diff
@@ -492,9 +504,35 @@ def _when_selection_remains_stable_async(self, f: Callable[[], None], r: sublime
...
+            return
+        self_id = self.view.id()
+        for view in views:
+            view_id = view.id()
+            if view_id == self_id:
+                continue
+            listeners = list(sublime_plugin.view_event_listeners[view_id])
+            for listener in listeners:
+                if isinstance(listener, DocumentSyncListener):
+                    debug("registering clone listener", listener)
```
rchl, line 534: It's not necessarily a clone. The current one might be a clone.

rwols: Is the current wording OK?

rchl:
I think we should wait with this until plugin host crash is fixed because it triggers easily (for example on closing settings window).

rwols:
I've seen it happen once, but cannot reproduce it now anymore. How do you make the preferences window crash?

rchl:
I just open Preferences with a shortcut (cmd+,) and close them. I have LSP-json starting there which is maybe relevant.

rwols:
By "close them", do you mean super+w (ctrl+w on linux)?

rchl:
Yes, with a keyboard shortcut.

plugin/documents.py
```diff
@@ -492,9 +504,35 @@ def _when_selection_remains_stable_async(self, f: Callable[[], None], r: sublime
 
     def _register_async(self) -> None:
         file_name = self.view.file_name()
-        if file_name:
-            self._file_name = file_name
-            self.manager.register_listener_async(self)
+        if not file_name:
+            debug("listener", self, "has  no file name")
+            return
+        self._file_name = file_name
```
rchl, line 510: _file_name doesn't seem to be used anywhere (anymore?).

plugin/documents.py
```diff
@@ -492,9 +504,35 @@ def _when_selection_remains_stable_async(self, f: Callable[[], None], r: sublime
 
     def _register_async(self) -> None:
         file_name = self.view.file_name()
-        if file_name:
-            self._file_name = file_name
-            self.manager.register_listener_async(self)
+        if not file_name:
+            debug("listener", self, "has  no file name")
```
rchl, line 508: The `listener` word is redundant now that `__rep__` is more descriptive.
Similarly, in places below I guess.

plugin/documents.py
```diff
@@ -246,11 +258,11 @@ def on_reload_async(self) -> None:
     # --- Callbacks from Sublime Text ----------------------------------------------------------------------------------
 
     def on_load_async(self) -> None:
-        if self._is_regular_view():
+        if not self._text_change_listener and is_regular_view(self.view):
             self._register_async()
 
     def on_activated_async(self) -> None:
-        if self._is_regular_view() and not self.view.is_loading():
+        if not self._text_change_listener and not self.view.is_loading() and is_regular_view(self.view):
```
rchl, line 265: Did you mean to check that the real object is still alive here or just that weakref was created? Because you are doing the latter but if that's the intention, we could make it clearer and less error-prone with a boolean.

rwols:
I cannot reproduce the crashing on my MacBook Pro from 2017. Perhaps another package is interfering?

rchl:
It reproduces in a clean (LSP + LSP-json) ST too albeit a bit less often.

Screen Recording 2020-08-15 at 16 45 15

rwols:
Given that this is more idiomatic use of TextChangeListener, I'd argue to go ahead with this and live with the crashes for a while. They should end up as crash reports for @BenjaminSchaaf

PR #1249, opened 2020-08-13
add few comments on java jstls setting sample

zhongjis:
add few comments on java jdtls setting sample. Clarify the fields that need to be changed based on the user env. Try to make it beginner-friendly


rwols:
Many thanks for improving the documentation, @zhongjis

PR #1250, opened 2020-08-13
Ruby server: use stdio instead of socket

OlegSmelov:
The suggested configuration with TCP port is problematic:

* It doesn't pass port number to the command and assumes it will use 7658
* It doesn't allow multiple servers running at the same time
* Solargraph is the only server in sample configuration that uses sockets


rchl:
Have not verified that it works but makes sense to me.
Will also update ST4 branch.

PR #1251, opened 2020-08-14
Add LSP: Troubleshoot Server to Command Palette

rchl:
Opens a sheet with all information that should be necessary to diagnose
problems with a server. Important part of that is that it runs selected
server to catch and show its stderr output. Something that is a pain to
do manually.

### DIFF
--- a/Commands/Default.sublime-commands
+++ b/Commands/Default.sublime-commands
@@ -20,6 +20,10 @@
         "command": "lsp_parse_vscode_package_json"
     },
     {
+        "caption": "LSP: Troubleshoot Server",
+        "command": "lsp_troubleshoot_server"
+    },
+    {
         "caption": "LSP: Setup Language Server",
         "command": "lsp_setup_language_server",
         "args": {}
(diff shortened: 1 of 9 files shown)


rchl:
Not sure how to fix this:
```
plugin/tooling.py:214: error: Signature of "run" incompatible with supertype "ApplicationCommand"
```
so that both:
```
    def run(self, contents: str) -> None:
```
and
```
    def run(self) -> None:
```

Tried using **kwargs and *args but nothing seems to work.

rchl:
There is something going wrong with code highlighting and mdpopups. Need to investigate that.

rwols:
This is pretty cool. Perhaps we can add another command that dumps the server's capabilities (in another PR).

rchl:
> This is pretty cool. Perhaps we can add another command that dumps the server's capabilities (in another PR).

Hmm, maybe, but not sure.

I've tried to not go deeper with interacting with the server than just starting it. Having to exchange messages would be on slightly deeper level.

Also, that information is already possible to get by enabling normal logging. The point of starting the server in this change is to mainly catch the server's output on server failing to start, something that is completely lost in normal situations because we capture stderr but never show it to the user. That's why user's had to start servers manually from the terminal to debug those issues before.

PR #1252, opened 2020-08-15
Rework configurations

rwols:
- Use a DottedDict to merge "default_clients" and "clients"
- Use  a dict to force uniqueness of configs (previously a list)
- Use only one change listener for the LSP.sublime-settings settings object instead of two


rchl:
I have set `"auto_show_diagnostics_panel": false,` but the diagnostics panel popped-up after switching to this branch. So something is not reading/considering user prefs.

plugin/core/types.py
```diff
@@ -89,39 +89,124 @@ def cancel_pending(self) -> None:
             self._current_id = -1
 
 
+def read_bool_setting(settings_obj: sublime.Settings, key: str, default: bool) -> bool:
+    val = settings_obj.get(key)
+    if isinstance(val, bool):
+        return val
+    else:
+        return default
```
rchl, line 97: Just a style but maybe more compact:
```suggestion
    return val if isinstance(val, bool) else default
```

plugin/core/types.py
```diff
@@ -197,6 +282,67 @@ def __init__(self,
...
+
+    @classmethod
+    def from_dict(cls, name: str, d: Dict[str, Any]) -> "ClientConfig":
+        return ClientConfig(
+            name=name,
+            binary_args=d.get("command", []),
+            languages=_read_language_configs(d),
+            tcp_port=d.get("tcp_port", None),
+            enabled=d.get("enabled", False),
+            init_options=d.get("initializationOptions", dict()),
```
rchl, line 313: Would switching `init_options` to `DottedDict` be in scope of this change?
LSP-vue settings are provided through it and it's annoying that you can't use the dotted notation for it.

plugin/core/types.py
```diff
@@ -89,39 +89,124 @@ def cancel_pending(self) -> None:
...
+    show_references_in_quick_panel = None  # type: bool
+    disabled_capabilities = None  # type: List[str]
+    log_debug = None  # type: bool
+    log_server = None  # type: List[str]
+    log_stderr = None  # type: bool
+    log_max_size = None  # type: int
+    lsp_format_on_save = None  # type: bool
+    show_code_actions = None  # type: bool
+    lsp_code_actions_on_save = None  # type: Dict[str, bool]
+    code_action_on_save_timeout_ms = None  # type: int
```
rchl, line 159: Style: I would like those sorted by name (same for `r` calls) :)

plugin/core/types.py
```diff
@@ -89,39 +89,124 @@ def cancel_pending(self) -> None:
...
-        self.diagnostics_panel_include_severity_level = 4
-        self.show_diagnostics_count_in_view_status = False
-        self.show_diagnostics_in_view_status = True
-        self.show_diagnostics_severity_level = 2
-        self.only_show_lsp_completions = False
-        self.diagnostics_highlight_style = "underline"
-        self.document_highlight_style = "stippled"
-        self.document_highlight_scopes = {
+    # This is only for mypy
+    show_view_status = None  # type: bool
```
rchl, line 135: These types seem wrong as the value is `None` initially so the matching type would be `Optional[bool]`. `pyright` seems to agree on that but I guess mypy is fine as linter looks happy...

plugin/core/settings.py
```diff
@@ -1,176 +1,68 @@
 from .collections import DottedDict
 from .logging import debug
-from .types import Settings, ClientConfig, LanguageConfig, syntax2scope
-from .typing import Any, List, Optional, Dict, Callable, Union
-from copy import deepcopy
+from .types import ClientConfig
```
rchl, line 3: Is there are reason for `ClientConfig` being in `types.py`, separate from `ClientConfigs`?
It feels a bit wrong to have `ClientConfig` in `types.py`, I feel because types.py should in theory by for... types :)

plugin/core/collections.py
```diff
@@ -131,3 +138,6 @@ def _update_recursive(self, current: Dict[str, Any], prefix: str) -> None:
                 self._update_recursive(value, path)
             else:
                 self.set(path, value)
+
+    def __repr__(self) -> str:
+        return repr(self._d)
```
rchl, line 143: I feel like it would be useful to know that it's a DottedDict.

plugin/core/types.py
rchl: Thanks. :)

plugin/core/types.py
rchl: This function is removed but there are still two other that could use this style.

plugin/core/configurations.py
```diff
@@ -64,10 +63,18 @@ def is_supported(self, view: Any) -> bool:
...
+        for name in self._global_configs.keys():
+            config = self._global_configs[name]
+            overrides = project_settings.get(config.name)
+            if isinstance(overrides, dict):
+                debug('applying .sublime-project override for', config.name)
+                config = config.update(overrides)
+            self.all[name] = config
+        for name in self._temp_disabled_configs:
+            self.all[name].enabled = False
+        self._window.run_command("lsp_recheck_sessions")
```
rchl, line 77: This breaks some things.

This command triggers `restart_sessions_async` from a timeout.
But at that point the sessions might have already started initializing so that command, depending on timing, will interrupt that.

rwols: Well, I'm spamming CTRL+S in LSP.sublime-settings and it seems to keep working. I think I accounted for this in `WindowManager.register_listener_async`: it checks if a session is initializing via `self._new_session`. If so, we do nothing, otherwise we do things.

In `register_listener_async`:
```
        self._pending_listeners.appendleft(listener)
        if self._new_listener is None:
            self._dequeue_listener_async()
```
And then later, in `on_post_initialize`:
```
    def on_post_initialize(self, session: Session) -> None:
        sublime.set_timeout_async(self._dequeue_listener_async)
```
It's kind of a complex set up. This would be massively simplified by using `async` functions, or possibly promises.

plugin/core/types.py
rwols: That syntax is for python2, but it also works for Sublime's python3.3. I'm not sure who is right: mypy or pyright.

plugin/core/configurations.py
rchl: I can reproduce by switching projects using "quick switch project".
I'm switching from empty project to one that uses `gopls` with active file triggering the server and the server doesn't start after switching to gopls project.

rchl: Actually this seems to be due to the code that updates workspace folders being gone:

In `on_load_project_async` there was:
```py
        workspace_folders = self._workspace.update()
        for session in self._sessions:
            session.update_folders(workspace_folders)
```

plugin/core/types.py
rchl: Mypy documentation says:

> Note that attribute definitions in the class body that use a type comment are special: a None value is valid as the initializer, even though the declared type is not optional. This should be used sparingly, as this can result in None-related runtime errors that mypy can’t detect.

https://mypy.readthedocs.io/en/stable/class_basics.html#class-attribute-annotations

It seems like a mypy limitation since it basically says itself that this can result in runtime errors.

plugin/core/configurations.py
rwols: This line:
```
self._workspace.update()
```
is needed to update the folders. Otherwise we run into the "do or do not attach servers for files outside workspace" discussion.

rchl (approved):
SHIP IT

PR #1255, opened 2020-08-16
Improve code actions tests stability

rchl:
* Await "textDocument/didSave" to ensure that we only test document
   state after "lsp_save" command has done its job.
 * Trigger "save" command instead of "lsp_save" on tearing down tests
   as there is no point in triggering save actions in that case.

### DIFF
--- a/tests/setup.py
+++ b/tests/setup.py
@@ -152,10 +152,11 @@ class TextDocumentTestCase(DeferrableTestCase):
 
     def await_message(self, method: str, promise: Optional[YieldPromise] = None) -> 'Generator':
         """
-        Awaits until the server responds with data returned for a request with specified method.
-        If server has already received a request with specified method before, it will immediately
-        return the response for that previous request. If it hasn't received such request yet, it
-        will wait for it and then respond or we're time out.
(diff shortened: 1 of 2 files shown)


tests/setup.py
```diff
@@ -151,6 +151,17 @@ def ensure_document_listener_created(self) -> bool:
         return False
 
     def await_message(self, method: str, promise: Optional[YieldPromise] = None) -> 'Generator':
+        """
+        Awaits until the server responds with data returned for a request with specified method.
+        If server has already received a request with specified method before, it will immediately
+        return the response for that previous request. If it hasn't received such request yet, it
+        will wait for it and then respond or we're time out.
```
rchl, line 158: This is really tricky to describe properly. I did my best and it's probably still a mess.

rwols: It's not entirely correct either. It awaits the moment when the server has received any kind of payload with the given method. Not when it responds.

rchl: I think by saying "until the server responds" I meant when it responds to this test request, not the original request.

rchl: Improved it a bit.

PR #1256, opened 2020-08-16
Don't report empty line to stderr handler

rchl:
Break out of stderr loop on empty message.

We don't want to break on server returning a newline so check the
contents before doing rstrip() on the message.

This avoids printing some spurious empty messages to the stderr callback.


rchl:
There is a difference between empty that we want to display and empty that we don't.

The ones we want to display are the ones with just a newline. A completely empty string should still behave like before and end the stream.

PR #1257, opened 2020-08-19
Fixes for API changes in build 4082

rwols:


plugin/goto.py
```diff
@@ -59,9 +58,7 @@ def handle_response(self, response: Any, side_by_side: bool) -> None:
         if window is None:
             return
         if response:
-            # Save to jump back history.
-            get_jump_history_for_view(view).push_selection(view)
-            # TODO: DocumentLink support.
+            self.view.run_command("add_jump_record", {"selection": [(r.a, r.b) for r in self.view.sel()]})
```
rchl, line 61: Id null-check self `view.sel()` maybe?

Also, there is `view` in scope (short than `self.view`).

plugin/goto.py
```diff
@@ -58,7 +58,8 @@ def handle_response(self, response: Any, side_by_side: bool) -> None:
         if window is None:
             return
         if response:
-            self.view.run_command("add_jump_record", {"selection": [(r.a, r.b) for r in self.view.sel()]})
+            if any(view.sel()):
```
rchl, line 61: I don't think `any` is necessary. As long as list has some element, it's valid IMO.

rchl (approved):

plugin/goto.py
rchl: Wait, actually it can even cause bugs. We should still add jump record for empty selection (empty as in, starts and stops at the same point).

rchl (changes requested):
changed my mind...

plugin/goto.py
```diff
@@ -58,7 +58,7 @@ def handle_response(self, response: Any, side_by_side: bool) -> None:
         if window is None:
             return
         if response:
-            if any(view.sel()):
+            if len(view.sel()) > 0:
```
rchl, line 61: `if view.sel()` would be good enough. :)

rchl (approved):

plugin/goto.py
rwols: `view.sel()` returns a `sublime.Selection` object, which has a custom `__bool__` that doesn't do what you'd expect :/ I've used `len(view.sel()) > 0` instead.

plugin/goto.py
rwols: No, sublime.Selection has a custom `__bool__` method. It returns true if and only if the view is valid.

plugin/goto.py
rchl: yes, it returns False for empty selection so `any(view.sel())` would return False for empty selection (starts and ends in same point).

plugin/goto.py
rchl: Hmm... `bool(view.sel())` == True

rchl: The bool of Selection object is not relevant when using just `if view.sel()`. It would just check that the list is non-empty, ignoring the bool operator of the Selection object.

rwols: Really? I've learned something new. I thought Python would evaluate that as

```python
if view.sel().__bool__()
```
(for any object that has a custom `__bool__`)

rchl: Ah, sorry, my bad. I thought that `view.sel()` is a list of `Selection` objects while it's actually the `Selection` object. So you are right. :)

PR #1259, opened 2020-08-19
Fix unittests

rchl:


rwols (commented):
I think the `skipIf` in test_views.py can now also be removed.

tests/test_views.py
```diff
@@ -192,7 +190,7 @@ def test_minihtml_format_string(self) -> None:
 
     def test_minihtml_format_marked_string(self) -> None:
         content = "<div>text\n</div>"
-        expect = "<div>text</div>"
+        expect = "<div>text\n</div>"
```
rwols, line 193: Is this due to a change in mdpopups?

rwols (approved):
Yep, that's pretty much all the tests I broke. Nice work.

tests/test_views.py
rchl: Yes.
I thought it was supposed to convert newlines to spaces when `nl2br` is disabled but it seems like it keeps newline here for some reason. Maybe has something to do with our frontmatter config:
```py
        frontmatter = {
            "allow_code_wrap": True,
            "markdown_extensions": [
                {
                    "pymdownx.escapeall": {
                        "hardbreak": False,
                        "nbsp": False
                    }
                }
            ]
        }
```

But to be fair, I don't really care as whitespace doesn't change anything in block elements so there won't be any visual change.

rchl: All in all, a newline will render as space in block on inline element when rendered in HTML so as expected per the mdpopups fix.

PR #1260, opened 2020-08-19
Use sublime.QuickPanelItem for textDocument/documentSymbol

rwols:
A server returning DocumentSymbol[]:
Screenshot from 2020-08-19 21-33-18

A server returning SymbolInformation[]:
Screenshot from 2020-08-19 21-32-59

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -27,33 +27,33 @@ DIAGNOSTIC_SEVERITY = [
 
 # The scope names mainly come from http://www.sublimetext.com/docs/3/scope_naming.html
 SYMBOL_KINDS = [
-    # Display Name     ST Scope
-    ("File",           "string"),
-    ("Module",         "entity.name.namespace"),
-    ("Namespace",      "entity.name.namespace"),
-    ("Package",        "entity.name.namespace"),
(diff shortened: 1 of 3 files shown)


rchl:
DANeo shows it with a lot of padding but that's a theme issue:
Screenshot 2020-08-19 at 22 00 32

Nice and tidy change :)

rchl:
The annotation accepts HTML I guess so we could add some nice support for `deprecated` also. But that can be done later.

rchl (approved):

genericptr:
I've got to try the new panel now (very nice) and I'm still liking the idea to add the container name (for SymbolInformation) to the annotation field (instead of the kind name, which is redundant).

For example you can see the TOpenGLView.XXX methods which I've chosen to add the class name but it would be better from a usability standpoint (because I don't have extra text that could be filtered during typing) and visually more appealing (because it makes the list shorter and saves space) if the annotation field could be the containerName instead.

Could you consider making this an option at the very least?

Screen Shot 2020-08-31 at 9 50 43 PM

rwols:
I'm planning on implementing your idea but it's not a high priority for me. Anyone is welcome to make a pull request of course.

PR #1261, opened 2020-08-20
Report errors returned from initialize request

rchl:
- The "Manager.on_post_initialize" abstract method replaced with callback as Manager needs to have access to "initiating_view" to reset status also on error.
- Also, call the new callback on failure during "initialize". That was missing, causing the session to not be GC'ed due to missing clearing of self._new_session
- Remove showing of exception in the status bar as those are useless due to disappearing too quickly and having limited space.
- Include exception error in the "crash dialog", if any.
- Allow Troubleshooting disabled configs because when the server crashes it gets disabled and then it's tricky to test why it crashed. Also, print the root scope of the syntax.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -12,7 +12,7 @@ from .types import ClientConfig
 from .types import ClientStates
 from .types import debounced
 from .types import diff
-from .typing import Dict, Any, Optional, List, Tuple, Generator, Type, Protocol
+from .typing import Callable, Dict, Any, Optional, List, Tuple, Generator, Type, Protocol
 from .url import uri_to_filename
 from .version import __version__
 from .views import COMPLETION_KINDS
@@ -25,6 +25,9 @@ from weakref import WeakSet
 import os
 import sublime
 import weakref
+
+
+InitCallback = Callable[['Session', bool], None]
 
 
 class Manager(metaclass=ABCMeta):
@@ -86,14 +89,6 @@ class Manager(metaclass=ABCMeta):
     def on_post_exit_async(self, session: 'Session', exit_code: int, exception: Optional[Exception]) -> None:
         """
         The given Session has stopped with the given exit code.
-        """
-        pass
-
-    @abstractmethod
-    def on_post_initialize(self, session: 'Session') -> None:
-        """
-        The language server returned a response from the initialize request. The response is stored in
-        session.capabilities.
         """
         pass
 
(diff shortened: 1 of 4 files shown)


plugin/core/windows.py
```diff
@@ -279,6 +281,16 @@ def start_async(self, config: ClientConfig, initiating_view: sublime.View) -> No
             self._new_session = None
             sublime.set_timeout_async(self._dequeue_listener_async)
 
+    def _on_post_session_initialize(
+        self, initiating_view: sublime.View, session: Session, is_error: bool = False
+    ) -> None:
+        if is_error:
+            session.config.erase_view_status(initiating_view)
+            self._new_listener = None
+            self._new_session = None
```
rchl, line 290: All those changes, switching from `on_post_initialize` to a callback due to needing initiating_view to clear the view status.

Maybe we should just not set the status on initializing and not need to clear it then?

Init typically takes a short time, potentially it's not even noticeable.

plugin/tooling.py
```diff
@@ -98,7 +98,7 @@ class LspTroubleshootServerCommand(sublime_plugin.WindowCommand, TransportCallba
     def run(self) -> None:
         window = self.window
         active_view = window.active_view()
-        configs = [c for c in windows.lookup(window).get_config_manager().get_configs() if c.enabled]
+        configs = windows.lookup(window).get_config_manager().get_configs()
```
rchl, line 101: This will now show a lot of default-included servers that are disabled.

Ideally we would show only enabled ones and those that are temporarily disabled.

rchl: I'll revert that.
It's possible to work around the issue it's fixing by restarting ST without triggering servers.

rchl:
BTW. This fix was prompted by LSP-graphql which returns a helpful error message from `initialize`, saying that the project configuration file is missing.

plugin/core/windows.py
rwols: > Maybe we should just not set the status on initializing and not need to clear it then?

I think it's helpful for those times that something initializes very long (e.g. bash-language-server).

plugin/tooling.py
```diff
@@ -178,6 +178,10 @@ def line(s: str) -> None:
             for key in keys:
                 settings[key] = view_settings.get(key)
             line(self.json_dump(settings))
+            if isinstance(settings['syntax'], str):
+                syntax = sublime.syntax_from_path(settings['syntax'])
+                if syntax:
+                    line(' - root scope\n{}'.format(self.code_block(syntax.scope)))
```
rwols, line 184: root scope -> base scope

plugin/core/sessions.py
```diff
@@ -833,10 +836,17 @@ def on_transport_close(self, exit_code: int, exception: Optional[Exception]) ->
         self.state = ClientStates.STOPPING
         super().on_transport_close(exit_code, exception)
         self._response_handlers.clear()
+        if self._init_error:
+            init_code, init_exception = self._init_error
+            exit_code = init_code
+            exception = init_exception
```
rwols, line 842: Where are the variables `exit_code` and `exception` used?

rchl: Below, in the call to `mgr.on_post_exit_async`

rwols (approved):
LGTM

PR #1271, opened 2020-08-23
Make symbol highlight faster by reducing the debounce time

predragnikolic:
This makes the highlights show up a lot faster. (I haven't noticed any perf issue with this, but I like how much the highlight feels faster)


rchl:
The debounce time is a compromise between being too slow and too noisy. I'm not sure if 150ms is a better option. It will make it noisier when navigating the code. For example here I navigate between lines fairly fast and it still triggers:
highlight

Maybe we could get it down to 350ms for example?

rwols:
I am worried that this might indeed be too quick.

predragnikolic:
> 350ms

I bumped it to 300ms. Is it sufficient?

rchl:
I'm fine with that.

rchl (approved):

PR #1272, opened 2020-08-24
Fix merging of initializeOptions for LSP packages

rchl:


rwols (approved):
Thanks for spotting this. I guess we should make a new release for this.

rchl:
I think it's a good time, yes.

PR #1274, opened 2020-08-25
Extend sublime-project schema

rchl:
Also updated project-specific settings documentation.

### DIFF
--- a/sublime-package.json
+++ b/sublime-package.json
@@ -394,6 +394,32 @@
             }
           }
         }
+      },
+      {
+        "file_patterns": [
+          "/*.sublime-project"
+        ],
+        "schema": {
+          "properties": {
+            "settings": {
+              "properties": {
+                "LSP": {
+                  "type": "object",
+                  "markdownDescription": "The dictionary of your configured language servers or overrides for existing configurations. The keys of this dictionary are free-form. They give a humany-friendly name to the server configuration. They are shown in the bottom-left corner in the status bar once attached to a view (unless you have `\"show_view_status\"` set to `false`).",
+                  "additionalProperties": {
+                    "$ref": "sublime://settings/LSP#/definitions/ClientConfig"
+                  },
+                },
+                "lsp_format_on_save": {
+                  "$ref": "sublime://settings/LSP#/definitions/lsp_format_on_save",
+                },
+                "lsp_code_actions_on_save": {
+                  "$ref": "sublime://settings/LSP#/definitions/lsp_code_actions_on_save"
+                }
+              }
+            }
+          }
+        }
       }
     ]
   }
--- a/sublime-package.json
+++ b/sublime-package.json
@@ -410,12 +410,6 @@
                     "$ref": "sublime://settings/LSP#/definitions/ClientConfig"
                   },
                 },
-                "lsp_format_on_save": {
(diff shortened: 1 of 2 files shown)


sublime-package.json
```diff
@@ -394,6 +394,32 @@
...
+                  "markdownDescription": "The dictionary of your configured language servers or overrides for existing configurations. The keys of this dictionary are free-form. They give a humany-friendly name to the server configuration. They are shown in the bottom-left corner in the status bar once attached to a view (unless you have `\"show_view_status\"` set to `false`).",
+                  "additionalProperties": {
+                    "$ref": "sublime://settings/LSP#/definitions/ClientConfig"
+                  },
+                },
+                "lsp_format_on_save": {
+                  "$ref": "sublime://settings/LSP#/definitions/lsp_format_on_save",
+                },
+                "lsp_code_actions_on_save": {
+                  "$ref": "sublime://settings/LSP#/definitions/lsp_code_actions_on_save"
```
rwols, line 417: Does this mean everyone has to write the same pieces of json for both the `/Preferences.sublime-settings` pattern as well as the `*.sublime-settings` pattern? I would prefer if this was automated in LSP-json.

rchl (commented):
[deleted]

sublime-package.json
rchl: Do you mean `Preferences.sublime-settings` and `*.sublime-project`?

Note that in this case those are not the same. The LSP object is unique.

But yeah, ideally the global settings should be handled automatically somehow but I'm not sure how. Maybe we could automatically attach schemas to file patterns based on `$id` matching some pattern.

rwols: > Note that in this case those are not the same. The LSP object is unique.

True. So there would be three file patterns: `/LSP.sublime-settings`, `/Preferences.sublime-settings` and `*.sublime-project`. The schema from `/Preferences.sublime-settings` should be merged into `*.sublime-project`, together with anything from `*.sublime-project` itself (in this case, only this `LSP` object).

rwols: Maybe you can let the schema author define what should be merged into what? For example, an array of `"mergeWith": ["/Preferences.sublime-settings"]` for `*.sublime-project`. I'm not sure if it's technically possible to do such things because I'm not familiar with the code from LSP-json.

rchl: Are you thinking to have that `mergeWith` at the root level of the schema or possible within any object type?
If the former then there would still need to be some custom logic for specific files to know that properties from `Preferences.json` schema go into the `settings` object. And in both cases, the schemas would need to be manually altered in-memory to add `allOf` or such. It sounds like a bit of a pain.

rchl: I have some code to merge those things so it should be doable.

rwols: Yeah, LSP-json has to do this automatically because otherwise this will be a major roadblock for other plugin devs to adopt sublime-package.json. I also predict not many plugin devs actually fully understand the override behavior of ST config files. So if LSP-json does this automatically, devs learn something by accident (the best kind of learning).

rchl: This should address the issue: https://github.com/sublimelsp/LSP-json/pull/46

rwols (approved):
This is much better :)

PR #1284, opened 2020-08-30
Add wcmatch library (for st builds >=4070)

rwols:
As preparation for a follow-up PR for #1282, this adds the wcmatch library as another dependency.

### DIFF
--- a/dependencies.json
+++ b/dependencies.json
@@ -1,15 +1,6 @@
 {
     "*": {
-        "3124 - 4069": [
-            "pygments",
-            "python-markdown",
-            "mdpopups",
-            "python-jinja2",
-            "markupsafe",
-            "pymdownx",
-            "pyyaml"
-        ],
-        ">=4070": [
+        ">=3124": [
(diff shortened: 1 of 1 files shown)


dependencies.json
```diff
@@ -1,13 +1,23 @@
 {
     "*": {
-        ">=3124": [
+        "3124 - 4069": [
```
rwols, line 3: Hmm, I'm unsure if this is the right syntax.

rwols: Seems valid https://github.com/wbond/package_control/blob/67ef2e02fef941c01694a52c5dfc3a08f86fbe49/example-dependencies.json#L20-L21

rchl: I don't see the point of those version dependant checks anyway as this code base won't work with ST3 or even older ST4 builds.

rwols: Fair point, I've removed the checks.

rchl (approved):

PR #1286, opened 2020-08-31
Implement pattern matching for DocumentFilter

### LINKED ISSUE #1216: Implement documentSelector in dynamic capability registration
The OmniSharp server registers *all* of its capabilities dynamically. Two of these registrations are for example

```
:: <-- OmniSharp client/registerCapability(14): {'registrations': [{'id': '98c3fac0-c80e-46ae-86c1-4d6587a41ddd', 'registerOptions': {'documentSelector': [{'pattern': '**/*.cake'}]}, 'method': 'textDocument/documentSymbol'}]}
:: <-- OmniSharp client/registerCapability(15): {'registrations': [{'id': '0baf930f-d735-4017-a5b3-60415c72cec7', 'registerOptions': {'documentSelector': [{'pattern': '**/*.cs'}, {'pattern': '**/*.csx'}]}, 'method': 'textDocument/documentSymbol'}]}
```
Note how it uses different document selectors for the same capability (`documentSymbolProvider` in this case).

The first registration uses the document selector
```js
[{'pattern': '**/*.cake'}]
```
The second registration uses the document selector
```sj
[{'pattern': '**/*.cs'}, {'pattern': '**/*.csx'}]
```
Currently, we don't account for these selectors. Consequently the first registration is overwritten with the second registration.

rwols:
This finally makes OmniSharp work 100% correctly.

Fixes #1216

### DIFF
--- a/dependencies.json
+++ b/dependencies.json
@@ -1,12 +1,14 @@
 {
     "*": {
         ">=3124": [
-            "pygments",
-            "python-markdown",
-            "mdpopups",
-            "python-jinja2",
+            "backrefs",
+            "bracex",
             "markupsafe",
+            "mdpopups",
+            "pygments",
             "pymdownx",
+            "python-jinja2",
+            "python-markdown",
             "pyyaml",
             "wcmatch"
         ]
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -65,9 +65,83 @@ class TestDocumentSelector(unittest.TestCase):
         view = self._make_view("Packages/Python/Python.sublime-syntax", "hello.py")
         self.assertFalse(selector.matches(view))
 
-    def test_pattern(self) -> None:
+    def test_pattern_basics(self) -> None:
         selector = DocumentSelector([{"language": "html", "pattern": "**/*.component.html"}])
         view = self._make_view("Packages/HTML/HTML.sublime-syntax", "index.html")
         self.assertFalse(selector.matches(view))
         view = self._make_view("Packages/HTML/HTML.sublime-syntax", "components/foo.component.html")
         self.assertTrue(selector.matches(view))
+
+    def _make_html_view(self, file_name: str) -> sublime.View:
+        return self._make_view("Packages/HTML/HTML.sublime-syntax", file_name)
+
+    def test_pattern_asterisk(self) -> None:
+        """`*` to match one or more characters in a path segment"""
+        selector = DocumentSelector([{"language": "html", "pattern": "a*c.html"}])
+        # self.assertFalse(selector.matches(self._make_html_view("ac.html")))
+        self.assertTrue(selector.matches(self._make_html_view("abc.html")))
+        self.assertTrue(selector.matches(self._make_html_view("axyc.html")))
+
+    def test_pattern_optional(self) -> None:
+        """`?` to match on one character in a path segment"""
+        selector = DocumentSelector([{"language": "html", "pattern": "a?c.html"}])
+        self.assertTrue(selector.matches(self._make_html_view("axc.html")))
+        self.assertTrue(selector.matches(self._make_html_view("ayc.html")))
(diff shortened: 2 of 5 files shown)


dependencies.json
```diff
@@ -1,12 +1,14 @@
 {
     "*": {
         ">=3124": [
-            "pygments",
-            "python-markdown",
-            "mdpopups",
-            "python-jinja2",
+            "backrefs",
+            "bracex",
```
rwols, line 5: These two are needed by wcmatch: https://github.com/facelessuser/sublime-wcmatch#dependencies

rchl, line 5: Not clear why those are needed.

tests/test_types.py
```diff
@@ -34,3 +38,36 @@ def test_completely_new(self) -> None:
...
+        self.assertTrue(selector.matches(view))
+        view = self._make_view("Packages/Python/Python.sublime-syntax", "hello.py")
+        self.assertFalse(selector.matches(view))
+
+    def test_pattern(self) -> None:
+        selector = DocumentSelector([{"language": "html", "pattern": "**/*.component.html"}])
+        view = self._make_view("Packages/HTML/HTML.sublime-syntax", "index.html")
+        self.assertFalse(selector.matches(view))
+        view = self._make_view("Packages/HTML/HTML.sublime-syntax", "components/foo.component.html")
+        self.assertTrue(selector.matches(view))
```
rchl, line 73: Maybe test with curly braces also, if supported?

tests/test_types.py
```diff
@@ -34,3 +38,36 @@ def test_completely_new(self) -> None:
...
+        view.file_name = MagicMock(return_value=file_name)
+        return view
+
+    def test_language(self) -> None:
+        selector = DocumentSelector([{"language": "txt"}])
+        view = self._make_view("Packages/Text/Plain text.tmLanguage", "foobar.txt")
+        self.assertTrue(selector.matches(view))
+        view = self._make_view("Packages/Python/Python.sublime-syntax", "hello.py")
+        self.assertFalse(selector.matches(view))
+
```
rchl, line 67: Do we support all features from the spec?

```js
	/**
	 * A glob pattern, like `*.{ts,js}`.
	 *
	 * Glob patterns can have the following syntax:
	 * - `*` to match one or more characters in a path segment
	 * - `?` to match on one character in a path segment
	 * - `**` to match any number of path segments, including none
	 * - `{}` to group conditions (e.g. `**​/*.{ts,js}` matches all TypeScript and JavaScript files)
	 * - `[]` to declare a range of characters to match in a path segment (e.g., `example.[0-9]` to match on `example.0`, `example.1`, …)
	 * - `[!...]` to negate a range of characters to match in a path segment (e.g., `example.[!0-9]` to match on `example.a`, `example.b`, but not `example.0`)
	 */
```

rwols: It all works :) I had to enable the `BRACE` and `GLOBSTAR` flags.

tests/test_types.py
```diff
@@ -65,9 +65,83 @@ def test_language(self) -> None:
...
         view = self._make_view("Packages/HTML/HTML.sublime-syntax", "components/foo.component.html")
         self.assertTrue(selector.matches(view))
+
+    def _make_html_view(self, file_name: str) -> sublime.View:
+        return self._make_view("Packages/HTML/HTML.sublime-syntax", file_name)
+
+    def test_pattern_asterisk(self) -> None:
+        """`*` to match one or more characters in a path segment"""
+        selector = DocumentSelector([{"language": "html", "pattern": "a*c.html"}])
+        # self.assertFalse(selector.matches(self._make_html_view("ac.html")))
```
rwols, line 81: I'm unsure about the wording in the spec here. The wording says

> one or more characters

but in my experience, `*` also matches zero characters.

rchl (approved):

rwols:
Thanks for the excellent library, @facelessuser

PR #1288, opened 2020-09-01
Add id for color box phantoms [master]

jwortmann:
This allows styling of the color box phantoms from color schemes.
It also sets the default background color to the color scheme's background to prevent an ugly white 0.2em margin on top, in case the background color for `html` or `body` is not set in "phantom_css" (if used).

### DIFF
--- a/plugin/color.py
+++ b/plugin/color.py
@@ -98,12 +98,14 @@ class LspColorListener(sublime_plugin.ViewEventListener):
             alpha = color['alpha']
 
             content = """
-            <style>html {{padding: 0}}</style>
+            <style>html {{padding: 0; background-color: var(--background)}}</style>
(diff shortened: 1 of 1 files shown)


rwols:
For my curiosity's sake, do you have a screenshot of your custom color box?

rwols (approved):

jwortmann:
This is primarily to avoid the white border on top of the box, but still allowing to set a different background color for the generic `body` tag in phantoms. Here is an example how it currently looks on master if you add a customization for "phantom_css" in Mariana, but with no CSS rules:
```json
{
    "globals": {
        "phantom_css": ""
    }
}
```
mariana


And this PR allows for example rounded corners with `#lsp-color-box div{border-radius:7px;}` in "phantom_css":

brackets

PR #1290, opened 2020-09-01
Use common popups styling for the documentation popup

rchl:
This is a follow-up for 8b5454a38435ad33c3008bc4ca96584cd5b389f0 where
the common popups style was applied to the show_popup() call but was
forgotten in update_popup().

Also replaced deprecated "allow_code_wrap" with frontmatter option.

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -35,11 +35,11 @@ class LspResolveDocsCommand(LspTextCommand):
         return minihtml(self.view, content, allowed_formats=FORMAT_STRING | FORMAT_MARKUP_CONTENT)
 
     def get_content(self, documentation: str, detail: str) -> str:
-        content = ""
+        content = mdpopups.format_frontmatter({"allow_code_wrap": True})
         if detail and not self.is_detail_shown:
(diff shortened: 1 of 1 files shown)


plugin/completion.py
```diff
@@ -35,11 +35,11 @@ def format_documentation(self, content: str) -> str:
         return minihtml(self.view, content, allowed_formats=FORMAT_STRING | FORMAT_MARKUP_CONTENT)
 
     def get_content(self, documentation: str, detail: str) -> str:
-        content = ""
+        content = mdpopups.format_frontmatter({"allow_code_wrap": True})
```
rwols, line 38: Don't we need all of the other frontmatter configuration here, too? (escapeall and magiclink?)

rchl: I can add that.

rchl: Well, the code here uses `minihtml` function which adds frontmatter automatically. Though it's added multiple times (twice) and not at the root of the content which seems a bit weird...

rchl: No, that's fine as `minihtml` returns processed HTML already.
So I think we just don't need frontmatter here as the content will be already "wrappable".

PR #1292, opened 2020-09-02
Fix diagnostics not being displayed in the view for virtual files

rchl:
Diagnostics were not displays for default preferences, for example,
since those files don't physically exist on disk.

The purpose of the original is_same_file check was to match also files
that different in case of the drive letter but if they match exactly
then we don't have to call it.


rwols (approved):
Nice catch!

PR #1293, opened 2020-09-03
Fix related diagnostic links not working

rwols:
Also fix http:// and https:// links not working for some reason

Screenshot from 2020-09-03 20-52-09

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -520,7 +520,7 @@ def format_diagnostic_for_panel(diagnostic: Diagnostic) -> str:
     return formatted
 
 
-def format_diagnostic_related_info(info: DiagnosticRelatedInformation, base_dir: Optional[str] = None) -> str:
+def _format_diagnostic_related_info(info: DiagnosticRelatedInformation, base_dir: Optional[str] = None) -> str:
     file_path = info.location.file_path
     if base_dir and file_path.startswith(base_dir):
         file_path = os.path.relpath(file_path, base_dir)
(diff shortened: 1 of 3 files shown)


plugin/core/views.py
```diff
@@ -520,20 +520,20 @@ def format_diagnostic_for_panel(diagnostic: Diagnostic) -> str:
...
+def _format_diagnostic_related_info(info: DiagnosticRelatedInformation, base_dir: Optional[str] = None) -> str:
     file_path = info.location.file_path
     if base_dir and file_path.startswith(base_dir):
         file_path = os.path.relpath(file_path, base_dir)
     row = info.location.range.start.row + 1
     col = info.location.range.start.col + 1
     encoded_filename = "{}:{}:{}".format(info.location.file_path, row, col)
     file_path = "{}:{}:{}".format(file_path, row, col)
-    return '<a href="{}">{}</a>: {}'.format(encoded_filename, text2html(file_path), text2html(info.message))
+    return '<a href="file://{}">{}</a>: {}'.format(encoded_filename, text2html(file_path), text2html(info.message))
```
rchl, line 531: Should we maybe use `filename_to_uri` on the `file_path` first?
I'm not sure what your change will do for Windows (the slashes might be wrong)

rchl: Looking at the rest of the changes, it looks like this is actually pretty custom handling and not an URI.
So maybe to make that clearer, use a custom protocol like `location://`

plugin/hover.py
```diff
@@ -174,21 +175,27 @@ def _show_hover(self, point: int) -> None:
...
-        elif href.startswith('code-actions'):
+        elif href.startswith('code-actions:'):
             _, config_name = href.split(":")
             titles = [command["title"] for command in self._actions_by_config[config_name]]
             sel = self.view.sel()
             sel.clear()
             sel.add(sublime.Region(point, point))
-
             self.view.show_popup_menu(titles, lambda i: self.handle_code_action_select(config_name, i))
+        elif href.startswith("file://"):
```
rchl, line 190: Would prefer `location://` as it's not actually a valid `file://` URI.

plugin/core/views.py
rwols: Good suggestion, thanks.

plugin/hover.py
```diff
@@ -174,21 +175,27 @@ def _show_hover(self, point: int) -> None:
...
+            window = self.view.window()
+            if window:
+                window.open_file(href[len("file://"):], flags=sublime.ENCODED_POSITION)
         else:
-            webbrowser.open_new_tab(href)
+            if not (href.lower().startswith("http://") or href.lower().startswith("https://")):
+                href = "http://" + href
```
rchl, line 196: I don't like that too much. Assuming `http://` might actually not work. If it's due to removing protocol in frontmatter then I would probably prefer not removing protocol in the first place.

(But you can decide on that)

plugin/hover.py
rwols: c48dd48

rchl (approved):

plugin/hover.py
rwols: Based on this comment I played around with disabling magiclink and/or hide_protocol, but the problem remains: `webbrowser.open` for some reason doesn't open `github.com`, but it does open `http://github.com` and/or `https://github.com`.

rwols: hide_protocol turns `http://github.com` into `[github.com](http://github.com)` by the way, so it's not a problem.

rchl: Oh, then it's fine. We'll only potentially fail for something that didn't have the protocol in the first place.

rwols: Yeah so the py3.3 `webbrowser` module seems broken with respect to not doing anything when http:// is absent, but the py3.8 `webbrowser` module does work. One more reason to move to py3.8 :)

PR #1294, opened 2020-09-03
Add schema for using as a base in LSP plugins

rchl:


sublime-package.json
```diff
@@ -414,7 +428,29 @@
...
+          "$id": "sublime://settings/LSP-plugin-base",
+          "properties": {
+            "languages": {
+              "$ref": "sublime://settings/LSP#/definitions/ClientLanguages"
+            },
+            "initializationOptions": {
+              "$ref": "sublime://settings/LSP#/definitions/ClientInitializationOptions"
+            },
+            "settings": {
+              "ref": "sublime://settings/LSP#/definitions/ClientSettings"
```
rwols, line 443: `"ref"` -> `"$ref"` ?

rchl: Good catch

rwols (approved):
I’m worried we’re falling into the fragile base class trap. Consider that the LanguageConfig class may turn into the DocumentSelector class in the future.

rchl:
I don't know, we'd have to figure it out somehow. We can't just change it instantly one day anyway as it will break all plugins. We need to provide some backward compatibility. At least for a certain period of time.

PR #1296, opened 2020-09-03
Fix MarkedString display in documentation popup

rchl:
A plain markdown text (MarkedString) was rendered differently from
MarkupContent for example. For MarkupContent mdpopups would add
a paragraph while for MarkedString, we've only used text2html that
didn't add paragraph.

Checked also hover popups with MarkedString and it renders the same.


rchl:
Before:

Screenshot 2020-09-03 at 23 30 19

After:

Screenshot 2020-09-03 at 23 30 53

It looks like a bug that the content not wrapped in a paragraph is not displayed fully but still it makes sense to wrap it as it also makes padding more consistent with MarkupContent case.

rwols (approved):
lgtm

PR #1297, opened 2020-09-04
Showcase: Send diagnostics to SublimeLinter

kaste:
As (or rather not *as*) discussed on discord

This is the patch I made *before* starting to chat.  The main pain point here was that I could not figure out the name of the "provider" or server this response actually belongs to. 

When sending diags to SL, SL manages Goto prev/next, the squiggles, the status bar, the panel automatically.  SL has vast customization of the squiggles per linter name, severity, and rule name, and a very dynamic panel.  It also has a distraction free mode(s).

### DIFF
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -234,6 +234,57 @@ class SessionBuffer:
         return None
 
     def on_diagnostics_async(self, raw_diagnostics: List[Dict[str, Any]], version: Optional[int]) -> None:
+        try:
+            from LSP.plugin.core.views import format_severity
+            from SublimeLinter.lint import backend, style, util
+            from SublimeLinter import sublime_linter
+        except ImportError:
+            pass
+        else:
+            view = self.some_view()
+            if view is None:
+                return
+            file_name = util.get_filename(view)
+            errors = []
+            # In the following `self._source_name` is a hack, we need
+            # in SL speak a `linter_name` or provider name (type: str)
+            # from somewhere, even if `raw_diagnostics` is empty.
+            for diag in map(Diagnostic.from_lsp, raw_diagnostics):
+                self._source_name = diag.source
+                region = range_to_region(diag.range, view)
+                lint_error = {
+                    "filename": file_name,
+                    "linter": diag.source,  # sic! see above; `diag.source` is `Optional[str]`
+                    "line": diag.range.start.row,
+                    "start": diag.range.start.col,
+                    "end": diag.range.start.col + len(region),
+                    "region": region,
+                    "error_type": format_severity(diag.severity),
+                    # This is typically the rule name, like "semi" or "F802"
+                    "code": str(diag._lsp_diagnostic.get("code", "")),
+                    "msg": diag.message,
+
+                    # That's an optimization.  Very cheap because we work
+                    # on a "virtual" view over there, t.i. not querying the
+                    # real view.
+                    "offending_text": view.substr(region),
+                }
+                lint_error.update({
+                    "uid": backend.make_error_uid(lint_error),
+                    "priority": style.get_value("priority", lint_error, 0)
+                })
+
+                errors.append(lint_error)
+
+            try:
+                linter_name = self._source_name
+            except AttributeError:
+                # shit
+                ...
+            else:
+                sublime_linter.update_file_errors(file_name, linter_name, errors, reason=None)
+            return
+
         diagnostics = []  # type: List[Diagnostic]
         data_per_severity = {}  # type: Dict[int, DiagnosticSeverityData]
         total_errors = 0


rchl:
Definitively interesting and impressive how easy it is to hook up into SL's panel.

My personal thoughts:
 - The common panel would need to be separated into a dependency if we are to share implementation.
 - The panel should not be handling squiggles or hover popups as LSP needs to show its own with some more data than just errors.

For reference here is LSP's own implementation:
- hover
Screenshot 2020-09-04 at 22 01 19
 - panel
Screenshot 2020-09-04 at 22 01 48

And with this change:
 - hover
Screenshot 2020-09-04 at 22 02 26
- panel
Screenshot 2020-09-04 at 22 03 17

Showing that just for reference, not to point out any issues (which there are of course)

kaste:
I noticed that `code` which is for linters *the* common thing is not used or differently used here in LSP.  For linters these rule names `F703`, or `no-semi` etc.  are super important and stable because all the configuration usually involves them.  

I also noticed that for example pyright reports `Pyright (some-rule)` as `source` which is LOL and at least wouldn't be compatible with what we expect. For us the `source` or `provider` would be always `pyright` and the code in this case would be `some-rule`.  

In your screenshots @rchl, eslint's vue reports an object as `code` whose `value` is what we would use as `code` normally, and `target` refers to a doc link afaict which would map to a top level key `url` in SL.  Vetur reports just a number, just like the typescript server, which must be resolved somehow.  Likely they have a mapping in the source code (?).  Don't understand why they send numbers because then a client/user would always have to look it up in a dictionary.  It saves a few bytes. 

Now generally the (SL) views are decoupled from the backend.  For example you could literally just delete the squiggles or panels *file* and the rest will work.  The minimal dependency is to have an `events` module.  Then LSP and SL just send `events.broadcast("update_error_store", payload)`, and then "my" panel if installed or your or any other panel will subscribe to these events, and do whatever is necessary.  (E.g. `events.subscribe("update_error_store", fn)`) This would be just an event system. 

Another way *could* be to instead implement a problems "manager".  The API is basically the same in that you would `problems.update_error_store(payload)` and `problems.subscribe("error_store_changed")`.  But such a central manager would be not dumb but responsible for storing (and cleaning!) the error store.  It has thus some autonomy (and more state) that's why it probably needs some event listeners already.

rchl:
> I noticed that `code` which is for linters _the_ common thing is not used or differently used here in LSP. For linters these rule names `F703`, or `no-semi` etc. are super important and stable because all the configuration usually involves them.

The `code` field is optional in the Diagnostic object returned from LSP. So it might not be provided at all.
See interface at https://microsoft.github.io/language-server-protocol/specification#diagnostic

> I also noticed that for example pyright reports Pyright (some-rule) as source which is LOL and at least wouldn't be compatible with what we expect. For us the source or provider would be always pyright and the code in this case would be some-rule.

`source` is also optional.

> In your screenshots @rchl, eslint's vue reports an object as `code` whose `value` is what we would use as `code` normally, and `target` refers to a doc link afaict which would map to a top level key `url` in SL.

In the current spec version the `code` can be either a string or a number. In the upcoming spec version (3.16) an object with `value` and `target` properties will also be allowed). Eslint goes a bit ahead of itself and uses it already (because eslint LSP server is deemed "vscode-internal" as of now so it can do whatever it wants).

> Vetur reports just a number, just like the typescript server, which must be resolved somehow.

I think it's sending internal Typescript code and that might not have a map to map that to human-readable value. I mean, typescript doesn't expose that map in its API.

I'm not sure if I like the events or API approach better. I often think of global events as a dirty globals-like solution disguised as something clean. With the API solution like @rwols suggested on Discord, where we have a "provider" created by a package, we get a two-way communication with more explicit and transparent design.

rchl:
Extra note on the `code` and `source`. The diagnostics returned by the server would likely need to be pre-processed by the package (LSP or SL) before sending them to the panel. So we could make sure that `source` is set to something reasonable (like client name maybe or at least fallback to that). Though for the `code` we probably wouldn't have much to work with if it's not provided. So not having it would just have to be handled gracefully.

rwols:
I agree with everything that's been said so far by @rchl. Just to summarize:

1) We must have some sort of Provider class that registers itself in its init method and unregisters itself in its del method.
2) Providers are per-window and there can be more than one provider per window per plugin (in LSP's case, a Provider corresponds with a Session object)
3) Read the LSP spec carefully about what is possible.
4) IMO sending dicts as params is bad and will cause confusion in the future when things change. Please define "rigid" function signatures.
5) All in all we need to figure out how we can run event listeners from a package control dependency.

rwols:
And I have to admit it's pretty cool to see what one can do in a little over 50 lines of Python :)

rchl:
> 5\. All in all we need to figure out how we can run event listeners from a package control dependency.

BTW. It's not true that EventListener will be instanciated multiple time when imported from different places.

As long as the export is not subclass but just imported, it will be instantiated once even if imported from multiple places.

So if both LSP and SL do something like:

```py
from my_dependency import CustomViewEventListener
```

then it will import the same instance and ST will instantiate it only once per view.

kaste:
I can hardly find myself here.  Either you're not fully focused here or you already have a lot of details, usually problems, in your mind that may or definitely will arise. 

For example you never described what a `Provider` actually does. 
So, *you* know it should implement `__del__` but *I* actually don't even know what state it holds.  Maybe it is a LSP implementation detail only loosely related to the problem here so I don't have to know.  

At least I'm irritated.  I wouldn't know what the *next* step is, and what a desired, imagined, wanted end-system looks like. 

The next babysteps would be, from the perspective of the PR: How to get a meaningful, stable "linter_name" or plugin name (server/client name?)?  Do you already interpret/transform the "code" for the user, or is this too new and we must just omit here?  (Do the LSP plugins actually interpret/normalize the data received; that's basically _the_ main point of the SL plugin system: translate the data to the Sublime world.)  Since I obviously just `return` after sending the data to SL, to suppress the side-effects of drawing squiggles etc., do you know which parts of the system or features don't work after that? 

It also sounds a bit all-or-nothing which is a bit early. We don't know each other.  And I don't know the system you've written here. 

Which subsystems would you want to delegate or share or decouple actually? Let me describe the SL system:

In SL we have a 10 lines Event system, and an almost zero lines Store.  The BackendController receives for example a `modified` event from Sublime core, passes a request-to-lint to the backend, and evetually receives a response.  (The backend does not have access to the Store nor the Event system.)  If the BackendController receives a response it calls the same function as here in the patch[1] which writes to the store and broadcasts an event.  

The Views generally receive lots of events, coming from Sublime core, the commands (key bindings), and our events system.  They also access the Store obviously.  They don't have access to or query the backend.  Maybe you could say the backend as described here is a diagnistics provider, if you want.  (But really how many independent or interdependent objects were used and how many singletons to produce the response is not shared knowledge.  In other words I really don't care if it implements `__del__` or not.)   

There is another kind-of provider, providing quick actions.  Iirc when hovering over a squiggle we ask "hey, do you have any quick actions for this error?".  (I think iirc I made this because quick actions can come from two ends, either they come with the lint response, and thus can be part of the error/the Store, *or* by other means. So I ask someone who then combines and interprets these actions.) 

Anyhow, so in natural language I could say I have two things that *provide* information.  But these are very different things.  Just saying "Provider" doesn't click here. 

A key feature that drove to this design is that new features, like experiments with annotations or so, don't start in a feature branch in SL core but start as separate Sublime plugins.  As a user myself, it has importance that I can have some "views" or small subtle things as an "addon" running that don't reach wide consensus and thus don't get merged.  (It also avoids some options wars.)

[1] Almost, it calls the exact sibling function because linter may report for multiple files in one go so we have to filter/group in our case.

rwols:
> For example you never described what a Provider actually does.
So, you know it should implement __del__ but I actually don't even know what state it holds. Maybe it is a LSP implementation detail only loosely related to the problem here so I don't have to know.

My point here:

> We must have some sort of Provider class that **registers itself in its init method and unregisters itself in its del method**.

is that if we use a class for this that registers itself and unregisters itself, then things like removing/disabling a package, enabling a package, etc, should work automatically by cleaning up their lint/diagnostic/problem contributions in the panel. That is, the `__del__` of such a `Provider` class should remove all of its contributions to this panel. 

The same idea is used in a `sublime.PhantomSet`: it cleans up all of its phantoms in `__del__`. Likewise a problems provider must clean up all of its errors/warnings in the panel in its `__del__`.

Also the reason I'm hammering on this is that you could argue that it can be a cleanup function, but people will most definitely forget to call that.

> The next babysteps would be, from the perspective of the PR: How to get a meaningful, stable "linter_name" or plugin name (server/client name?)?

In the case of LSP it's `session.config.name`, where `session` is a `Session` :) it's probably suitable enough to get this linter name in the `__init__` of Provider?

```python

def Provider:
   def __init__(self, window: sublime.Window, name: str) -> None:
        self.name = name
        # ... register this provider to the panel with the given window ...

    def __del__(self) -> None:
        # ... remove all contributions from this provider from the panel ...
```

>  Do you already interpret/transform the "code" for the user, or is this too new and we must just omit here?

There's no clever translation happening because every language server has its own error codes and/or error strings. The `code` is also optional and may be omitted. To normalize it for the "SL panel", seeing as it's going to be a dict in the future spec, we should also normalize it to a dict

> Do the LSP plugins actually interpret/normalize the data received; that's basically the main point of the SL plugin system: translate the data to the Sublime world.

Not in the case of diagnostics.

>  Since I obviously just return after sending the data to SL, to suppress the side-effects of drawing squiggles etc., do you know which parts of the system or features don't work after that?

We have a few things which rely on linting data:

1) The squiggles
2) Hovering over a squiggle should present the diags in a popup
3) Every time the caret changes we do an intersection of the caret with the stored diags to see if there are any code actions associated to it
4) There's also an intersection being done to set the "active diagnostic" in the status bar
5) There's a global error and warning count in the status bar
6) There's also a "diagnostic walker" that's implemented as a phantomset
7) The diagnostics panel of course :)

So, if you want this problems panel to also be a storage for diagnostics, then it has to have a function that gives back a list of intersections given a `sublime.Region` or a point, and it has to be able to partition on the provider:

```python
region = self.view.sel()[0]
intersection = provider.intersect(view, region)  # only items from this particular provider, from the particular view
```

For code actions associated to diagnostics, we need to send back the original dict that got sent by the server, so we must be able to provide `userdata` when doing `provider.send(...)` and get it back when we're doing `provider.intersect`:

```python
diag = Diagnostic.from_lsp(raw_diagnostic)
region = range_to_region(diag.range, view)
provider.send(source=diag.source, code=diag.code, message=diag.message, region=region, userdata=raw_diagnostic)

# ... later ...

intersections = provider.intersect(view.sel()[0])
for intersection in intersections:
    raw_diagnostic = intersection.userdata  # we get back the original data
```

> Which subsystems would you want to delegate or share or decouple actually?

Not sure about this :) It's going to depend on what is optimal for performance. There's two ways to approach this (maybe more, let me know):

1) The "SL panel" acts as a view where plugins push their storage to and it'll show the data.
2) The "SL panel" acts as a storage itself and plugins do intersection queries on the storage.

Ideally I would like to see option (2). As I explained higher up in this comment, that would require the "SL panel" to store `userdata`, which can be anything, and also implement intersection queries.

> There is another kind-of provider, providing quick actions. Iirc when hovering over a squiggle we ask "hey, do you have any quick actions for this error?". (I think iirc I made this because quick actions can come from two ends, either they come with the lint response, and thus can be part of the error/the Store, or by other means. So I ask someone who then combines and interprets these actions.)

The correct abstraction here I think is again intersections.

> Anyhow, so in natural language I could say I have two things that provide information. But these are very different things. Just saying "Provider" doesn't click here.

I disagree that the "SL panel" or "problems panel" should also take care of fix-its, quick-actions, code-actions, or whatever actionable thing you want to call it :)

> A key feature that drove to this design is that new features, like experiments with annotations or so, don't start in a feature branch in SL core but start as separate Sublime plugins. As a user myself, it has importance that I can have some "views" or small subtle things as an "addon" running that don't reach wide consensus and thus don't get merged. (It also avoids some options wars.)

I'm okay with it being a more loose API.

One more thing: most of the code in this LSP plugin runs async and lockless. The shared problems library should also have a complete async interface, without any use of `threading.Lock`.

All in all I think we can achieve a lot with the following. I'm not saying this should be the one and only interface. It's a suggestion.

```python
class Problem:

    def __init__(
        self,
        view_or_file_name: Union[str, sublime.View],
        region: sublime.Region,
        severity: int,
        code: Optional[str],
        source: Optional[str],
        message: str,
        userdata: Optional[Any] = None
    ) -> None:
        self.view_or_file_name = view_or_file_name
        self.region = region
        self.severity = severity
        self.code = code
        self.source = source
        self.message = message
        self.userdata = userdata


class Provider:
    def __init__(self, window: sublime.Window, name: str) -> None:
        """Construct a provider in Sublime's worker thread."""
        self.name = name
        # ... register this Provider to the internal bookkeeping system ...

    def __del__(self) -> None:
        """Only safe to del a provider in Sublime's worker thread."""
        # ... Remove all problem contributions from the panel and storage for this provider ...

    def update_async(self, problems: Iterable[Problem]) -> None:
        """This method is meant to be called from Sublime's worker thread."""
        # ... Store the problems, update the panel, squiggles, etc ...

    def intersect_async(
        self,
        view_or_file_name: Union[str, sublime.View],
        region_or_point: Union[int, sublime.Region]
    ) -> List[Problem]:
        """This method is meant to be called from Sublime's worker thread."""
        # ... do intersection queries ...
```

I hope this clears things up!

PR #1298, opened 2020-09-06
Check that the view has a filename on-disk when starting the TextChangeListener

rwols:
I think this was (partially) fixed in 4084. The file_name() is now reliable when checking for it in is_applicable of the TextChangeListener.

There is still the issue of checking whether the view is transient being unreliable.


rchl:
BTW. Not sure if you caught that but ViewEventListener now has an `on_new_buffer`, `on_new_buffer_async`, `on_associate_buffer` and `on_associate_buffer_async` callbacks. Maybe something we should look into using?

rchl (approved):

PR #1300, opened 2020-09-06
Fix color box phantoms background in selections

jwortmann:
I have introduced a regression in https://github.com/sublimelsp/LSP/commit/32aaa5856e936118a9296035b18f35c2ddc63c65:
colorbox
The background color from the scheme is still shown on top when the color box phantom is inside a selection.

By using a complete transparent color (alpha value 0) it should work as expected now, both in selected text and when using "phantom_css" in the color scheme.

This should also be applied to the st4000-exploration branch.

### DIFF
--- a/plugin/color.py
+++ b/plugin/color.py
@@ -98,7 +98,7 @@ class LspColorListener(sublime_plugin.ViewEventListener):
             alpha = color['alpha']
 
             content = """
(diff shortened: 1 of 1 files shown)


jwortmann:
Actually the value `transparent` works as well in minihtml and is easier to understand.

rwols (approved):
lgtm

PR #1302, opened 2020-09-06
Fix: manually invoking signature help should not check the triggers

rwols:
Also, find relevant sessions via the stored session views / buffers

This is more direct than going via the manager.


rchl (approved):
Nice

PR #1304, opened 2020-09-07
Build 4085: account for "trim_trailing_white_space_on_save" being a string

rwols:


plugin/core/views.py
```diff
@@ -275,13 +275,22 @@ def did_close(file_name: str) -> Notification:
...
 def formatting_options(settings: sublime.Settings) -> Dict[str, Any]:
+    # Build 4085 allows "trim_trailing_white_space_on_save" to be a string so we have to account for that in a
+    # backwards-compatible way.
+    trim_trailing_white_space = settings.get("trim_trailing_white_space_on_save")
+    if isinstance(trim_trailing_white_space, bool):
+        pass
+    elif isinstance(trim_trailing_white_space, str):
+        trim_trailing_white_space = (trim_trailing_white_space != "none")
+    else:
+        trim_trailing_white_space = False
```
rchl, line 286: Can we just do a shorter:
```py
trim_trailing_white_space =trim_trailing_white_space not in (False, "none")

Comment (rchl) [inline plugin/core/views.py:286]: Can we just do a shorter:
```suggestion
trim_trailing_white_space = settings.get("trim_trailing_white_space_on_save") not in (False, "none")
```

PR #1305, opened 2020-09-09
Notify user about failed initialize

### LINKED ISSUE #1177: When a server returns an error response during the initialize request, show an error dialog to the user
Currently, we silently do the shutdown -> exit dance. There is no indication that something went wrong in a fatal way. The text in the dialog should show from which server the error came, and it should show the error message from the server.

See: https://github.com/sublimelsp/LSP-typescript/issues/15

rchl:
Resolves #1177

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -506,7 +506,7 @@ class Session(Client):
         self.exiting = False
         self._registrations = {}  # type: Dict[str, _RegistrationData]
         self._init_callback = None  # type: Optional[InitCallback]
-        self._exit_result = None  # type: Optional[Tuple[int, Optional[Exception]]]
+        self._initialize_error = None  # type: Optional[Tuple[int, Optional[Exception]]]
         self._views_opened = 0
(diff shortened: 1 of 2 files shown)


rchl:
The original intention of the change was to ignore errors on an explicit shutdown of the session but I think it ended up always ignoring the initialization error also.

This change makes sure that the initialization error is reported but doesn't handle ignoring errors on shutdown anymore.

rchl:
I have forced a crash in `shutdown` for eslint and this is currently not reported either.
```
:: --> LSP-eslint shutdown(3): None
:: <~~ LSP-eslint 3: {'code': -32603, 'message': 'Request shutdown failed with message: crash is not defined'}
```
So the only errors that we report are the ones that are violations of protocol I guess.

rchl:
Or more precisely, we show the crash dialog when on ask asking the process to close itself with `self._process.wait(1)` it answers with non-0 exit code.

rwols (approved):

PR #1308, opened 2020-09-10
Fix padding for "no documentation" text in completions info

rchl:
Always go through minihtml so that default wrapper element is added.


rwols (approved):

rwols:
Many thanks!

PR #1310, opened 2020-09-11
Fix completion info not shown when session doesn't support resolveProvider

### LINKED ISSUE #1309: [LSP-elm] Documentation popup is no logner visible when "More" link is clicked
* OS and language server - Ubunutu 20.04, LSP-elm, ST 4085 
* How you installed LSP - git, latest st-4000-exploration

When clicking the `More` link in the AC popup,
I expect to see a documentation popup.
But I see nothing. 

This commit introduced this behavior 19df9e19afeb0f32064a8b7e3a11ebaa4254f63c
If I checkout the commit before 19df9e19afeb0f32064a8b7e3a11ebaa4254f63c, everything works as expected.

rchl:
Resolves #1309

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -25,10 +25,12 @@ class LspResolveDocsCommand(LspTextCommand):
         if not detail or not documentation:
             # To make sure that the detail or documentation fields doesn't exist we need to resove the completion item.
             # If those fields appear after the item is resolved we show them in the popup.
-            self.do_resolve(item)
-        else:
-            minihtml_content = self.get_content(documentation, detail)
-            self.show_popup(minihtml_content)
+            session = self.best_session('completionProvider.resolveProvider')
+            if session:
+                session.send_request(Request.resolveCompletionItem(item), self.handle_resolve_response)
+                return
+        minihtml_content = self.get_content(documentation, detail)
+        self.show_popup(minihtml_content)
 
     def format_documentation(self, content: Union[str, Dict[str, str]]) -> str:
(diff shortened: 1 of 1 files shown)


plugin/completion.py
```diff
@@ -25,10 +25,12 @@ def run(self, edit: sublime.Edit, index: int, event: Optional[dict] = None) -> N
         if not detail or not documentation:
             # To make sure that the detail or documentation fields doesn't exist we need to resove the completion item.
             # If those fields appear after the item is resolved we show them in the popup.
-            self.do_resolve(item)
-        else:
-            minihtml_content = self.get_content(documentation, detail)
-            self.show_popup(minihtml_content)
+            session = self.best_session('completionProvider.resolveProvider')
```
rwols, line 28: This is a bit odd. We can only enter LspResolveDocs when either:

1) the session is a resolveProvider, or
2) there is actual documentation

From what I understand, elm-language-server is apparently *not* a resolveProvider? Then we must have checked if there was documentation present in the first in `format_completion` from views.py?

rwols: Ah, but if there's no detail field then we also enter this block. I think that's the mistake?

rchl: From the spec about the `completionItem/resolve` request:
> The request can only delay the computation of the detail and documentation properties.
So we have to try resolving if we don't have either of those, I think.

rchl: > From what I understand, elm-language-server is apparently _not_ a resolveProvider? Then we must have checked if there was documentation present in the first place in `format_completion` from views.py?

Yes, in this case:
 - the server is not a `resolveProvider`
 - there is a `documentation` property filled-in initially

rwols (approved):

PR #1312, opened 2020-09-11
Attach sessions to views outside the workspaces

### LINKED ISSUE #997: Figure out how to allow servers to start for files outside of project
With the emergence of `LSP-bash`, I have now realized we need to be able to start servers for files outside of the project directory. That's because I don't keep a project for my bash scripts in `~/` and, if I would, that would basically kill ST.

I haven't put any thought in it yet but there must be some way to handle that. Maybe decide per-server?

rwols:
Resolves #997

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -584,9 +584,11 @@ class Session(Client):
 
     # --- capability observers -----------------------------------------------------------------------------------------
 
-    def can_handle(self, view: sublime.View, capability: Optional[str] = None) -> bool:
+    def can_handle(self, view: sublime.View, capability: Optional[str], inside_workspace: bool) -> bool:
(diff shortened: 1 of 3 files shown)


rchl (approved):
Interesting that for files outside of the workspace you are reusing the workspace sessions. Or creating a new session with rootPath of the workspace if there is no session yet... It seems to work just fine for a few servers I've tried though.

I was thinking the whole time that for files outside of the workspace we would create a separate session with `rootPath` set to null. I think that might be more "correct" but at the same time might upset servers more as many might not like folder-less workspaces...

rwols:
It could be problematic when the user opens a file outside the workspaces from a goto-def from a session for rootPath X, but it attaches to a session for rootPath Y. In that case I say make your server handle workspaces instead of relying on rootPath.

PR #1317, opened 2020-09-12
More granular inhibit completion options

### LINKED ISSUE #1166: Feature request: allowing passthrough of sublime snippets in "only_show_lsp_completions" mode
* OS and language server
- Mac OS X Mojave
- Any language server (general LSP package feature request)

Hi there, this is a feature request as mentioned on Discord chat, being moved here. My set-up is Sublime Text 4, however I think this applies to all versions of ST and ST-LSP.

ST-LSP has the feature `only_show_lsp_completions`, which, when enabled, show only LSP completions. This drastically improves the ST autocomplete, which has never been worth much on its own.

However, there is one feature that lots of people use via autocomplete as the main interface — the sublime snippets. Enabling `only_show_lsp_completions` disables snippet completion.

The feature request is to make sublime snippets either by default passed through LSP in the case `only_show_lsp_completions` is enabled, or make this snippet passthrough a flag in the settings. My vote would be towards passing snippets through by default, since it is unlikely for a person to attempt disabling sublime snippets by enabling `only_show_lsp_completions` in LSP. 

Thanks for making this package — this is quite possibly the best ST package in existence in terms of usefulness.

rwols:
Replace "only_show_lsp_completions" with:

- "inhibit_explicit_completions"
- "inhibit_word_completions"

When inhibit_word_completions is enabled, this will also cause LSP
to disable ST's completion ordering and instead use the ordering of
the relevant language server.

Resolves #1166

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -112,7 +112,7 @@
   "show_symbol_action_links": false,
 
   // Disable Sublime Text's snippet completions.
-  "inhibit_explicit_completions": false,
+  "inhibit_snippet_completions": false,
 
   // Disable Sublime Text's word completions. When set to `true`, this also disables Sublime Text's internal completion
   // sorting algorithm and instead uses the sorting defined by the relevant language server.
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -111,8 +111,12 @@
   // Show symbol action links in hover popup if available
   "show_symbol_action_links": false,
 
-  // Disable Sublime Text's explicit and word completion.
-  "only_show_lsp_completions": false,
+  // Disable Sublime Text's snippet completions.
+  "inhibit_explicit_completions": false,
+
+  // Disable Sublime Text's word completions. When set to `true`, this also disables Sublime Text's internal completion
+  // sorting algorithm and instead uses the sorting defined by the relevant language server.
+  "inhibit_word_completions": true,
 
   // Show symbol references in Sublime's quick panel instead of the bottom panel.
   "show_references_in_quick_panel": false,
--- a/docs/features.md
+++ b/docs/features.md
@@ -161,8 +161,9 @@ Add these settings to LSP settings, your Sublime settings, Syntax-specific setti
 
 ### Package settings (LSP)
 
-* `only_show_lsp_completions` `false` *disable sublime word completion and snippets from autocomplete lists*
-* `show_code_actions` `annotation` *Show code actions: "annotation", "bulb" *
+* `inhibit_word_completions` `true` *disable sublime word completion*
+* `inhibit_explicit_completions` `true` *disable sublime snippet completion*
+* `show_code_actions` `annotation` *Show code actions: "annotation", "bulb"*
 * `code_action_on_save_timeout_ms` `2000` *the amount of time the code actions on save are allowed to run for*
 * `show_references_in_quick_panel` `false` *show symbol references in Sublime's quick panel instead of the bottom panel*
 * `show_view_status` `true` *show permanent language server status in the status bar*
@@ -177,7 +178,7 @@ Add these settings to LSP settings, your Sublime settings, Syntax-specific setti
 * `document_highlight_scopes`: *customize your sublime text scopes for document highlighting*
 * `diagnostics_gutter_marker` `"dot"` *gutter marker for code diagnostics: "dot", "circle", "bookmark", "sign" or ""*
 * `show_symbol_action_links` `false` *show links to symbol actions like go to, references and rename in the hover popup*
-* `disabled_capabilities`, `[]` *Turn off client capabilities (features): "hover", "completion", "documentHighlight", "colorProvider", "signatureHelp"
+* `disabled_capabilities`, `[]` *Turn off client capabilities (features): "hover", "completion", "documentHighlight", "colorProvider", "signatureHelp"*
 * `log_debug` `false` *show debug logging in the sublime console*
 * `log_server` `[]` *log communication from and to language servers*
 * `log_stderr` `false` *show language server stderr output in the console*
--- a/docs/features.md
+++ b/docs/features.md
@@ -162,7 +162,7 @@ Add these settings to LSP settings, your Sublime settings, Syntax-specific setti
 ### Package settings (LSP)
 
 * `inhibit_word_completions` `true` *disable sublime word completion*
-* `inhibit_explicit_completions` `true` *disable sublime snippet completion*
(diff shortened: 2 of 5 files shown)


rwols:
A "breaking" change in this diff is that word completions are now disabled by default. But I'm confident that most language servers are capable enough.

sublime-package.json
```diff
@@ -315,10 +315,15 @@
               "default": false,
               "markdownDescription": "Show symbol action links in hover popup if available."
             },
-            "only_show_lsp_completions": {
```
rchl, line 318: Would be better to "deprecate" it, mentioning alternatives.

LSP.sublime-settings
```diff
@@ -111,8 +111,12 @@
...
   "show_symbol_action_links": false,
 
-  // Disable Sublime Text's explicit and word completion.
-  "only_show_lsp_completions": false,
+  // Disable Sublime Text's snippet completions.
+  "inhibit_explicit_completions": false,
+
+  // Disable Sublime Text's word completions. When set to `true`, this also disables Sublime Text's internal completion
+  // sorting algorithm and instead uses the sorting defined by the relevant language server.
+  "inhibit_word_completions": true,
```
rchl, line 119: Have you thought of having one setting that takes an array of strings instead? I think it might be easier to manager for us and for users.

docs/features.md
```diff
@@ -161,8 +161,9 @@ Add these settings to LSP settings, your Sublime settings, Syntax-specific setti
 
 ### Package settings (LSP)
 
-* `only_show_lsp_completions` `false` *disable sublime word completion and snippets from autocomplete lists*
-* `show_code_actions` `annotation` *Show code actions: "annotation", "bulb" *
+* `inhibit_word_completions` `true` *disable sublime word completion*
+* `inhibit_explicit_completions` `true` *disable sublime snippet completion*
```
rchl, line 165: Default is `false`

plugin/documents.py
```diff
@@ -544,9 +544,11 @@ def _on_complete_result(self, response: Optional[Union[dict, List]], completion_
...
         response_items = []  # type: List[Dict]
         flags = 0
-        if userprefs().only_show_lsp_completions:
-            flags |= sublime.INHIBIT_WORD_COMPLETIONS
+        prefs = userprefs()
+        if prefs.inhibit_explicit_completions:
             flags |= sublime.INHIBIT_EXPLICIT_COMPLETIONS
+        if prefs.inhibit_word_completions:
+            flags |= sublime.INHIBIT_WORD_COMPLETIONS
             flags |= sublime.INHIBIT_REORDER
```
rchl, line 552: What's the motivation for not using INHIBIT_REORDER when inhibiting explicit completions?

sublime-package.json
```diff
@@ -315,10 +315,15 @@
...
+            "inhibit_explicit_completions": {
               "type": "boolean",
               "default": false,
-              "markdownDescription": "Disable Sublime Text's explicit and word completion."
+              "markdownDescription": "Disable Sublime Text's snippet completions."
+            },
+            "inhibit_word_completions": {
+              "type": "boolean",
+              "default": false,
```
rchl, line 325: The `default` here doesn't match reality but I'm thinking now that the default should probably be the opposite of what is the actual default since when the user is overriding an option, he normally wants to override it to be an opposite of default.

LSP.sublime-settings
```diff
@@ -111,8 +111,12 @@
   // Show symbol action links in hover popup if available
   "show_symbol_action_links": false,
 
-  // Disable Sublime Text's explicit and word completion.
-  "only_show_lsp_completions": false,
+  // Disable Sublime Text's snippet completions.
+  "inhibit_explicit_completions": false,
```
rchl, line 115: Could we give it a more understandable name than "explicit"? I know that ST calls it that but it's not really clear what it means and we could just give it a better and clearer name - namely `inhibit_snippet_completions`.

plugin/documents.py
```diff
@@ -526,7 +526,7 @@ def _apply_view_settings(self, session: Session) -> None:
         if any(trigger.get('server', None) == session.config.name for trigger in completion_triggers):
             return
         # This is to make ST match with labels that have a weird prefix like a space character.
-        settings.set('auto_complete_preserve_order', 'none')
+        # settings.set('auto_complete_preserve_order', 'none')
```
rchl, line 529: Hmm?

sublime-package.json
```diff
@@ -315,15 +315,20 @@
...
+            "inhibit_snippet_completions": {
               "type": "boolean",
               "default": false,
-              "markdownDescription": "Disable Sublime Text's word completions. When set to `true`, this also disables Sublime Text's internal completion sorting algorithm and instead uses the sorting defined by the relevant language server."
+              "markdownDescription": "Disable Sublime Text's (and your own) snippet completions. Note that your language server may still provide completion items of the \"snippet\" kind."
+            },
+            "inhibit_word_completions": {
+              "type": "boolean",
+              "default": true,
+              "markdownDescription": "Disable Sublime Text's word completions. \"word\" completion means Sublime Text's internal completer that takes words from the current buffer you're editing and presents them in the auto-complete widget. When set to `true`, this also disables Sublime Text's internal completion sorting algorithm and instead uses the sorting defined by the relevant language server."
```
rchl, line 331: > When set to `true`, this also disables Sublime Text's internal completion sorting algorithm

Now, this seems to happen irrespective of those options.

plugin/documents.py
rwols: Whoops :) I was playing around with this yesterday.

sublime-package.json
rwols: INHIBIT_REORDER doesn't seem to really do anything if word completions are enabled. ST seems to prefer word completions over plugin completions. So, I'd say just enable it when there's at least one item.

rchl (approved):

LSP.sublime-settings
rwols: Hmm, I don't think there will be other flags in the foreseeable future.

predragnikolic: Sorry for the late reply. 
I would also agree with @rchl that a list would me more suitable. 

And I would also like to point out that we should be careful how we name settings. 
It is easy to forget that not all ST are power users, so we should be make the settings really simple to understand.

```
// Exclude from the auto complete:
// "snippets", "words"
"exclude_completions": []
```
we can also have a similar setting for the quick_panel symbol list(ctrl+r) to

```
// exclude symbols from the symbol list:
// "property", "class", "method", ... the rest of SymbolKind but only lowercase
"exclude_symbols": ["property"]
```

PR #1321, opened 2020-09-13
Add support for DottedDict

rchl:
The public "read_client_config" API now takes an optional third argument
"base_settings_path" which must be passed by subclasses of LanguageHandler
to properly handle settings merging. It's an opt-in so without it
everything should work as before (no support for merging dotted settings).


rchl:
I've thrown it together quickly and I might have missed something.
Also, the tests will probably fail.

rwols (approved):
Looks good to me.

rwols:
Tried this locally and it seems to not break anything :) Man, what a world of difference is the completion system now.

PR #1322, opened 2020-09-14
Fix: restart a language server after it crashed by doing the registration dance

rwols:
Try to kill a language server from your process manager app.


plugin/core/windows.py
```diff
@@ -357,9 +357,8 @@ def on_post_exit_async(self, session: Session, exit_code: int, exception: Option
             msg += "\n\nDo you want to restart {0}?\n\nIf you choose Cancel, {0} will "\
                    "be disabled for this window until you restart Sublime Text.".format(config.name)
             if sublime.ok_cancel_dialog(msg, "Restart {}".format(config.name)):
-                view = self._window.active_view()
-                if view:
-                    self.start_async(config, view)
+                for listener in self._listeners:
+                    self.register_listener_async(listener)
```
rchl, line 361: I don't have the whole mapped out in my head but looks like a sound change.

rchl (approved):

PR #1325, opened 2020-09-15
Update code actions on diagnostics change

### LINKED ISSUE #1301: Request code actions when new diagnostics arrive on the current line
I'm starting to depend more on code actions and one thing that is somewhat annoying is that code action requests are triggered when the text changes or when the caret changes position. But not when diagnostics are updated. Frequently what happens is that I'm done typing, code actions are then requested. No code actions return. Then diagnostics come in and they carry code actions stored on the server. So we should re-request code actions in that case.

As a workaround I now have to press the left arrow button.

rchl:
Resolves #1301

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -72,7 +72,7 @@ class AbstractViewListener(metaclass=ABCMeta):
         raise NotImplementedError()
 
     @abstractmethod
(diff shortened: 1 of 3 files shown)


rwols (approved):
Awesome :)

PR #1326, opened 2020-09-15
Prevent accidental view input after triggering document symbols

### LINKED ISSUE #1323: LSP Document Symbols downside compared to native one
**Describe the bug**
I've been running with `cmd+r` overridden to use Document Symbols from LSP. The functionality is better than native but it has one downside compared to native behavior. Namely, with native, I can trigger the shortcut and start typing immediately. The quick panel appears immediately (synchronously or at least with almost no latency). When using LSP's document symbol, the quick panel only opens after we receive server response so, being used to just start typing my query right away, my query ends up being inserted into the active view instead.

Ideally, the quick panel would open right away, even before the response comes, but I guess the API might not let us do that.

Alternatively, we could potentially work around that somehow by capturing user input when the command is triggered and before the quick panel is shown.

rchl:
After triggering LSP's "Document Symbols", the user might start typing
the query immediately and if the server response is delayed the query
would end up changing the view. This is a stop-gap solution to avoid
such accidental input until https://github.com/sublimehq/sublime_text/issues/3338
is implemented.

Resolves #1323

### DIFF
--- a/Keymaps/Default (Linux).sublime-keymap
+++ b/Keymaps/Default (Linux).sublime-keymap
@@ -335,4 +335,16 @@
     //         }
     //     ]
     // },
+    // Internal key-binding
+    {
+        "keys": [
+            "<character>"
+        ],
+        "command": "noop",
+        "context": [
+            {
+                "key": "setting.lsp_suppress_input",
+            }
+        ]
+    },
 ]
--- a/plugin/symbols.py
+++ b/plugin/symbols.py
@@ -12,6 +12,9 @@ import sublime
 import sublime_plugin
 
 
+SUPPRESS_INPUT_SETTING_KEY = 'lsp_suppress_input'
+
+
 def unpack_lsp_kind(kind: int) -> Tuple[int, str, str, str]:
     if 1 <= kind <= len(SYMBOL_KINDS):
         return SYMBOL_KINDS[kind - 1]
@@ -76,7 +79,7 @@ class LspDocumentSymbolsCommand(LspTextCommand):
         self.is_first_selection = False
 
     def run(self, edit: sublime.Edit) -> None:
-        self.view.settings().set('lsp_supress_input', True)
+        self.view.settings().set(SUPPRESS_INPUT_SETTING_KEY, True)
         session = self.best_session(self.capability)
         if session:
             session.send_request(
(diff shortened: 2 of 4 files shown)


plugin/symbols.py
```diff
@@ -93,6 +96,9 @@ def handle_response(self, response: Any) -> None:
                 self.on_highlighted)
             self.view.run_command("lsp_selection_clear")
 
+    def handle_response_error(self, response: Any) -> None:
+        self.view.settings().erase('lsp_supress_input')
```
rwols, line 100: Also log the error to the status bar, like the default error handler does.

plugin/symbols.py
```diff
@@ -75,12 +75,15 @@ def __init__(self, view: sublime.View) -> None:
...
+        self.view.settings().set('lsp_supress_input', True)
         session = self.best_session(self.capability)
         if session:
             session.send_request(
-                Request.documentSymbols({"textDocument": text_document_identifier(self.view)}), self.handle_response)
+                Request.documentSymbols({"textDocument": text_document_identifier(self.view)}),
+                self.handle_response, self.handle_response_error)
 
     def handle_response(self, response: Any) -> None:
+        self.view.settings().erase('lsp_supress_input')
```
rwols, line 86: Hmm... `handle_response` runs in some thread somewhere. That means that theoretically what could happen is: the view setting is erased, then the user types a character, and only then the quickpanel is shown.

In order to guarantee atomicity here, we should run the handler on the UI thread with `sublime.set_timeout`.

plugin/symbols.py
```diff
@@ -75,12 +75,15 @@ def __init__(self, view: sublime.View) -> None:
         self.is_first_selection = False
 
     def run(self, edit: sublime.Edit) -> None:
+        self.view.settings().set('lsp_supress_input', True)
```
rwols, line 78: Typo: supress -> suppress

Keymaps/Default (Linux).sublime-keymap
```diff
@@ -335,4 +335,16 @@
     //         }
     //     ]
     // },
+    // Internal key-binding
+    {
+        "keys": [
+            "<character>"
+        ],
```
rwols, line 342: It feels dangerous, but let's try it anyway. Note that things like backspace and the enter key are not handled by `<character>`. But I think there's a low chance that anyone is pressing those keys.

rwols (changes requested):

rwols:
I'll merge this after a release because I'm still afraid that that `<character>` binding will break something.

rchl:
No problem. The scope the key-binding is enabled is very narrow but we can play it safe in case it works in an unexpected way.

PR #1335, opened 2020-09-18
Trigger autocomplete popup on "editor.action.triggerSuggest" command

### LINKED ISSUE #1139: Triggers completion command on the wrong session
When having both LSP-css and LSP-stylelint running for open file (*.css), the completions are requested from LSP-css but the `command` that the completion item triggers runs on `LSP-stylelint` triggering an error dialog.

An example completion item looks like this:
```
    {
      "command": {
        "command": "editor.action.triggerSuggest",
        "title": "Suggest"
      },
      "documentation": {
        "kind": "plaintext",
        "value": "Specifies the width of the content area, padding area or border area (depending on 'box-sizing') of certain boxes.\n\nSyntax: <viewport-length>{1,2}\n\nMDN Reference: https://developer.mozilla.org/docs/Web/CSS/width"
      },
      "insertTextFormat": 2,
      "kind": 10,
      "label": "width",
      "sortText": "d_0000",
      "tags": [],
      "textEdit": {
        "newText": "width: $0;",
        "range": {
          "end": {
            "character": 5,
            "line": 4
          },
          "start": {
            "character": 4,
            "line": 4
          }
        }
      }
    },
```

It has a `command` that runs after inserting completion.

rchl:
This is VSCode specific behavior that unfortunately CSS language server
(and possibly other) uses.

Also fixes arguments passed to "lsp_execute" as those were not correct.

Resolves #1139

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -89,8 +89,12 @@ class LspCompleteCommand(sublime_plugin.TextCommand):
             self.view.run_command("lsp_apply_document_edit", {'changes': edits})
         command = item.get("command")
         if command:
+            if command['command'] == "editor.action.triggerSuggest":
+                sublime.set_timeout(lambda: self.view.run_command("auto_complete"))
+                return
             debug('Running server command "{}" for view {}'.format(command, self.view.id()))
-            self.view.run_command("lsp_execute", {"command_name": command})
+            args = {"command_name": command["command"], "command_args": command.get("arguments")}
+            self.view.run_command("lsp_execute", args)
 
 
 class LspCompleteInsertTextCommand(LspCompleteCommand):
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -89,9 +89,6 @@ class LspCompleteCommand(sublime_plugin.TextCommand):
             self.view.run_command("lsp_apply_document_edit", {'changes': edits})
         command = item.get("command")
         if command:
-            if command['command'] == "editor.action.triggerSuggest":
-                sublime.set_timeout(lambda: self.view.run_command("auto_complete"))
-                return
             debug('Running server command "{}" for view {}'.format(command, self.view.id()))
             args = {"command_name": command["command"], "command_args": command.get("arguments")}
             self.view.run_command("lsp_execute", args)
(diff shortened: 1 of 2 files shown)


rchl:
This doesn't resolve #1139 per se, just avoids the issue triggering.

plugin/completion.py
```diff
@@ -89,8 +89,12 @@ def epilogue(self, item: Dict[str, Any]) -> None:
             self.view.run_command("lsp_apply_document_edit", {'changes': edits})
         command = item.get("command")
         if command:
+            if command['command'] == "editor.action.triggerSuggest":
```
rwols, line 92: I would rather see this check done in LspExecuteCommand itself. So that anyone who invokes that command doesn't have to bother with these checks.

plugin/completion.py
```diff
@@ -89,8 +89,12 @@ def epilogue(self, item: Dict[str, Any]) -> None:
             self.view.run_command("lsp_apply_document_edit", {'changes': edits})
         command = item.get("command")
         if command:
```
predragnikolic, line 91: Adding a comment saying that this was added to fix a vs code specific implementation would be nice for future reference.

Ideally this should be fixed in the language server, because it editor specific and nowhere in the spec does it say that editors should handle a editor.action.triggerSuggest command  🙂

rchl: > Ideally this should be fixed in the language server,

Don't think it's possible to do otherwise with LSP.

plugin/execute_command.py
```diff
@@ -15,6 +15,10 @@ def run(self,
             command_name: Optional[str] = None,
             command_args: Optional[List[Any]] = None,
             event: Optional[dict] = None) -> None:
+        # Handle VSCode-specific command for triggering suggestions popup.
+        if command_name == "editor.action.triggerSuggest":
+            sublime.set_timeout(lambda: self.view.run_command("auto_complete"))
+            return
```
rwols, line 21: Why does this need a set_timeout? I think it can be `return self.view.run_command("auto_complete")` no?

rchl: Just doesn't work without it. Possibly something to do with running one command within another command (and possibly specific to auto_complete).

rwols: Oh okay. The magical set_timeout solves most of the mysterious problems ;)

rwols (approved):

rwols (changes requested):
Actually, please add a comment about the set_timeout because it isn't obvious why it's needed.

rwols (approved):

PR #1336, opened 2020-09-18
Trigger commands on server that matches one that returned completions

### LINKED ISSUE #1139: Triggers completion command on the wrong session

rchl:
Pass config name through the whole chain of completions handling so
that we can trigger command on the same server that returned completion.

Resolves #1139

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -7,6 +7,7 @@ from .core.logging import debug
 from .core.edit import parse_text_edit
 from .core.protocol import Request, InsertTextFormat, Range
 from .core.registry import LspTextCommand
+from .core.types import SessionId
 from .core.typing import Any, List, Dict, Optional, Generator, Union
 from .core.views import FORMAT_STRING, FORMAT_MARKUP_CONTENT, minihtml
 from .core.views import range_to_region
@@ -82,7 +83,7 @@ class LspResolveDocsCommand(LspTextCommand):
 
 class LspCompleteCommand(sublime_plugin.TextCommand):
 
-    def epilogue(self, item: Dict[str, Any], session_name: Optional[str] = None) -> None:
+    def epilogue(self, item: Dict[str, Any], sid: SessionId = None) -> None:
         additional_edits = item.get('additionalTextEdits')
         if additional_edits:
             edits = [parse_text_edit(additional_edit) for additional_edit in additional_edits]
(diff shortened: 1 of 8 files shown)


plugin/completion.py
```diff
@@ -82,32 +82,36 @@ def update_popup(self, minihtml_content: str) -> None:
...
+                "command_args": command.get("arguments"),
+                "session_name": session_name
+            }
             self.view.run_command("lsp_execute", args)
 
 
 class LspCompleteInsertTextCommand(LspCompleteCommand):
 
-    def run(self, edit: sublime.Edit, **item: Any) -> None:
+    def run(self, edit: sublime.Edit, item: Any, session_name: Optional[str] = None) -> None:
```
rchl, line 103: Not sure why there were `kwargs` used here. That would "collect" all unrelated arguments into the same argument, right? Don't understand the use case...

plugin/core/views.py
```diff
@@ -635,7 +637,7 @@ def format_completion(item: dict, index: int, can_resolve_completion_items: bool
         completion = sublime.CompletionItem.command_completion(
             trigger=st_trigger,
             command="lsp_complete_text_edit",
-            args=item,
+            args={"session_name": session_name, "item": item},
```
rwols, line 640: I don't really like this solution because it duplicates a lot of the same data. If there are 200 completion items then every one of them contains the same session_name. Maybe we can store the name in a class-level variable instead, same as the completion items for the LspResolveDocs command?

rwols (changes requested):

plugin/completion.py
rwols: It was for performance because it doesn't allocate another dict. But then again, ST serializes everything to a string and stores those arguments somewhere in plugin_host, so it probably doesn't matter much.

plugin/core/views.py
rchl: I really prefer to do it correctly (this way) than doing a hack using a global.

Having a global can restrict us in the future if we decide to pull completions from multiple sessions.

Also, I think this is adding relatively small amount of data to already pretty big completion items. We could optimize it by having some unique IDs that would be smaller in size, if it's really an issue.

rchl: Pushed alternative solution that has smaller footprint but adds extra complexity. Let me know. :)

predragnikolic: Abort, abort, too much additional code  :)

An alternative solution would be to still use session_name instead of session_id:
image

But I still think your original solution with passing the config is better, because it wont cause any future issues,
on the other side I also understand @rwols because this info seems redundant, and we only send the completion request to one session. 

What about this solution here:
image

predragnikolic: But as said, I also don't mind passing the session name all the way down trough the chain.

rwols: I have to agree with @predragnikolic. The previous diff was better.

I've been thinking about this. If there's no `"command"` in the completion item we don't have to stash the session name in the args. I think that would keep 95% of completion items untouched.

rwols (approved):
Nice, thanks!

PR #1341, opened 2020-09-26
Act as a TCP server for langservers that want to connect to the client

rwols:
Needed for getting SonarLint to work. This language server assumes that the editor is playing the role of a TCP server.

Dug old code up from this commit: f1a53fc45466d008392a17f1c768c26723b14b27

It's a bit simplified with respect to configuration in the following sense:

- We don't start a listener socket on every ethernet device, only localhost.
- In the config, set `tcp_port` to `-1` to start a listener socket with an arbitrary OS-chosen port, or
- set `tcp_port` to `X` where `X < -1` to start a listener socket listening on port `-X`.

### DIFF
--- a/plugin/core/transports.py
+++ b/plugin/core/transports.py
@@ -197,71 +197,6 @@ def _fixup_startup_args(args: List[str]) -> Any:
     return startupinfo
 
 
-def _start_subprocess(
-    args: List[str],
-    stdin: int,
-    stdout: int,
-    stderr: int,
-    startupinfo: Any,
-    env: Dict[str, str],
-    cwd: Optional[str]
-) -> subprocess.Popen:
-    debug("starting {} in {}".format(args, cwd if cwd else os.getcwd()))
-    process = subprocess.Popen(
-        args=args,
-        stdin=stdin,
-        stdout=stdout,
-        stderr=stderr,
-        startupinfo=startupinfo,
-        env=env,
-        cwd=cwd)
-    _subprocesses.add(process)
-    return process
-
-
-def _start_tcp_listener(tcp_port: Optional[int]) -> socket.socket:
-    sock = socket.socket()
-    sock.bind(('localhost', tcp_port or 0))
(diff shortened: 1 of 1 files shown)


plugin/core/transports.py
```diff
@@ -176,27 +176,7 @@ def _stderr_loop(self) -> None:
...
-    env = os.environ.copy()
-    for var, value in config.env.items():
-        env[var] = sublime.expand_variables(value, variables)
-    if tcp_port is not None:
-        stdout = subprocess.DEVNULL
-        stdin = subprocess.DEVNULL
-    else:
-        stdout = subprocess.PIPE
-        stdin = subprocess.PIPE
+def _fixup_startup_args(args: List[str]) -> Any:
```
rchl, line 179: Can we sort the function in the order those are used (roughly at least)? Or are you intentionally putting private function on top?

rwols: Do the last three commits improve the ordering?

rchl: Yes. And diff is easier to read also.

plugin/core/transports.py
```diff
@@ -179,8 +179,18 @@ def _stderr_loop(self) -> None:
 def create_transport(config: ClientConfig, cwd: Optional[str], window: sublime.Window,
                      callback_object: TransportCallbacks, variables: Dict[str, str]) -> JsonRpcTransport:
     tcp_port = None  # type: Optional[int]
+    listener_socket = None  # type: Optional[socket.socket]
     if config.tcp_port is not None:
-        tcp_port = _find_free_port() if config.tcp_port == 0 else config.tcp_port
+        # < 0 means we're hosting a TCP server
+        if config.tcp_port < 0:
```
rchl, line 185: This condition (together with -1 check below) seems to suggest that other negative values apart from -1 are also supported. Is that the intention?

rchl: Never mind, I've read the comment now.

rchl (approved):

PR #1342, opened 2020-09-26
Add function to open a file asynchronously

rwols:
The function `open_file` in promises.py returns a Promise that shall be
resolved once the file is fully loaded. It's only safe to call `open_file`
from the main thread because the `on_load` event callback runs on the main
thread.

I want to use this for workspace edits as well as renames.

### DIFF
--- a/plugin/core/main.py
+++ b/plugin/core/main.py
@@ -7,6 +7,7 @@ from .css import unload as unload_css
 from .handlers import LanguageHandler
 from .logging import set_debug_logging, set_exception_logging
 from .panels import destroy_output_panels
+from .promise import opening_files
 from .protocol import Response
 from .protocol import WorkspaceFolder
 from .registry import windows
@@ -21,6 +22,7 @@ from .settings import userprefs
 from .transports import kill_all_subprocesses
 from .types import ClientConfig
 from .typing import Optional, List, Type, Callable, Dict, Tuple
+import os
 import weakref
 
 
@@ -153,3 +155,12 @@ class Listener(sublime_plugin.EventListener):
 
     def on_pre_close_window(self, w: sublime.Window) -> None:
         windows.discard(w)
+
+    def on_load(self, view: sublime.View) -> None:
+        file_name = view.file_name()
+        if not file_name:
+            return
+        for fn, tup in opening_files.items():
+            if fn == file_name or os.path.samefile(fn, file_name):
+                opening_files.pop(fn)
+                return tup[1](view)
--- a/plugin/core/main.py
+++ b/plugin/core/main.py
@@ -7,7 +7,6 @@ from .css import unload as unload_css
 from .handlers import LanguageHandler
 from .logging import set_debug_logging, set_exception_logging
 from .panels import destroy_output_panels
-from .promise import opening_files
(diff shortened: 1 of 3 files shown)


plugin/core/promise.py
```diff
@@ -184,3 +186,40 @@ def _is_resolved(self) -> bool:
     def _get_value(self) -> Any:
         with self.mutex:
             return self.value
+
+
+opening_files = {}  # type: Dict[str, Tuple[Promise, Callable[[sublime.View], None]]]
+
+
+def open_file(
```
rchl, line 194: I find the placement of this utility function rather weird. `promise.py` is a file containing an implementation of the Promise concept itself. This function doesn't have anything to do with the implementation of promise but is just a utility function that happens to use promise for execution (and would likely use asyncio instead if that would be available).

plugin/core/promise.py
```diff
@@ -184,3 +186,40 @@ def _is_resolved(self) -> bool:
...
+    window: sublime.Window,
+    file_path: str,
+    flags: int = 0,
+    group: int = -1
+) -> Promise:
```
rchl, line 199: Fits nicely on one line.
We need to save vertical space. :)

plugin/core/promise.py
```diff
@@ -184,3 +186,40 @@ def _is_resolved(self) -> bool:
...
+
+    # Is the view opening right now? Then return the associated unresolved promise
+    for fn, value in opening_files.items():
+        if fn == file_path or os.path.samefile(fn, file_path):
+            return value[0]
+
+    # Prepare new promise
+    def fullfill(resolve: ResolveFunc) -> None:
+        global opening_files
+        opening_files[file_path] = (None, resolve)  # type: ignore
```
rchl, line 217: So this reset the Promise to `None` when view is opened? What happens if that file is closed and then we get another open request? Won't the loop above just return None value then?

plugin/core/main.py
```diff
@@ -153,3 +155,12 @@ def on_new_window_async(self, w: sublime.Window) -> None:
...
         windows.discard(w)
+
+    def on_load(self, view: sublime.View) -> None:
+        file_name = view.file_name()
+        if not file_name:
+            return
+        for fn, tup in opening_files.items():
+            if fn == file_name or os.path.samefile(fn, file_name):
+                opening_files.pop(fn)
+                return tup[1](view)
```
rchl, line 166: Of course, it's not a problem in practice but I'd just `break` on a separate line instead to make intentions clearer.

plugin/core/promise.py
rwols: Where would you suggest we place this function?

plugin/core/promise.py
rwols: The fullfill func is ran immediately upon creating the promise. It puts the resolve func in the dict, to be called later by on_load

plugin/core/promise.py
rchl: I'm afraid of suggesting "utils" since it seems like you were avoiding that on purpose for some time now. ;)

Would `core/windows.py` work for now, since this is for opening file in a window so somewhat related.

PR #1344, opened 2020-09-26
Store inflight requests in SessionView

rwols:
We can now also show the active requests in the status bar.
It looks a bit frantic without a delay so I've added a delay of 100ms.

As an added bonus the "$test" requests also show up in the tests.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -267,6 +267,12 @@ class SessionViewProtocol(Protocol):
     def present_diagnostics_async(self, flags: int) -> None:
         ...
 
+    def on_request_started_async(self, request_id: int, request: Request) -> None:
+        ...
+
+    def on_request_finished_async(self, request_id: int) -> None:
+        ...
+
 
 class SessionBufferProtocol(Protocol):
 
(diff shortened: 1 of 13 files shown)


rchl:
> We can now also show the active requests in the status bar.
> It looks a bit frantic without a delay so I've added a delay of 100ms.

It's absolutely crazy and people should not be concerned with stuff like that. :) Please disable by default at least. It's basically very distracting as a small change in a document suddenly makes the status bar flash with multiple requests showing on and off.

plugin/core/sessions.py
```diff
@@ -1050,10 +1066,17 @@ def on_payload(self, payload: Dict[str, Any]) -> None:
...
 
     def response_handler(self, response_id: int, response: Dict[str, Any]) -> Tuple[Optional[Callable], Any, bool]:
-        handler, error_handler = self._response_handlers.pop(response_id, (None, None))
-        if not handler:
+        request, handler, error_handler = self._response_handlers.pop(response_id, (None, None, None))
+        if not request:
             error = {"code": ErrorCode.InvalidParams, "message": "unknown response ID {}".format(response_id)}
             return (print_to_status_bar, error, True)
+        if request.view:
+            sv = self.session_view_for_view_async(request.view)
```
rchl, line 1074: These should also be done from sublime's async thread (currently this is in stdio read thread).
(Just a note)

rchl (changes requested):

rwols:
Hmm okay, I'll remove the status bar printing. For workDoneProgress, I do think we should set the active request in the status bar when the server starts reporting progress for a particular active request.

rchl:
We shouldn't be too strict about it. IMO, there is really no point in showing progress for something that takes a short time. It's not useful and it's distracting. If something takes a long time then yes, it makes sense. But that means that the delay could be set to 1 or even 2s and I think that would be a good enough heuristic.

rchl (approved):

PR #1345, opened 2020-09-26
Allow plugin packages to define "command" themselves

rchl:


sublime-package.json
```diff
@@ -26,6 +26,17 @@
               "type": "string",
               "deprecationMessage": "Use \"languages\" instead."
             },
+            "ClientCommand": {
+              "type": "array",
+              "default": [
+                "path/to/language-server-binary",
+                "--some-option"
+              ],
```
rwols, line 34: I recently learned that vscode-json-language-server can also inserts snippets. Can we add some sweet snippet for the `"command"` completion? https://code.visualstudio.com/Docs/languages/json#_define-snippets-in-json-schemas

rchl: but `default` acts as a snippet also. It's automatically inserted when adding that key.

rwols (approved):

sublime-package.json
rwols: Yes but the `"path/to/language-server-binary`" and `"--some-option"` are not snippet fields. But maybe suitable for another pr.

PR #1346, opened 2020-09-26
Normalize path components in "command" option used to start the server

rchl:


plugin/core/transports.py
```diff
@@ -212,6 +212,7 @@ def create_transport(config: ClientConfig, cwd: Optional[str], window: sublime.W
         stdout = subprocess.PIPE
         stdin = subprocess.PIPE
     startupinfo = _fixup_startup_args(args)
+    args = list(map(os.path.normpath, args))
```
rwols, line 215: Hmm... What if one of the arguments is an URI like `file:///foo/bar.txt` ?

rwols: I think this os.path.normpath should only be done for the very first element of `args`, in which case we might as well do it in `_fixup_startup_args` ?

rchl: I guess it won't work correctly for `file://*`. But would it make sense to use `file://*` in the first place? This is a command that is passed to `subprocess.popen` so I doubt a URI would work?

> I think this os.path.normpath should only be done for the very first element of `args`,

The first argument can be `node` so that wouldn't be a good heuristic.

rwols: Coincidentally [this sonarlint config](https://github.com/sublimelsp/LSP-SonarLint/blob/93e2ccedde67647c478ac99adb862b49ea57e955/src/LSP-SonarLint.sublime-settings#L2-L12) I wrote earlier today wants file:// uris to additional jar files.

rchl: Hmm. OK, then I don't see a way to normalize path. Will just hope what you said is true and Windows will have no problems with that.

PR #1347, opened 2020-09-27
Use the http.client.parse_headers function to reduce code

rwols:


rwols:
The parse_headers function seems to have been removed from http.client in python3.8, so this is not future proof.

rwols:
I spoke too soon. It's right there in the py38 version of the plugin host:

```
>>> import sys
>>> sys.version_info
sys.version_info(major=3, minor=8, micro=6, releaselevel='final', serial=0)
>>> import http
>>> http.client
<module 'http.client' from '/opt/sublime_text/Lib/python3.8.zip/http/client.pyc'>
>>> http.client.parse_headers
<function parse_headers at 0x7f934f287700>
```

I'm not sure what's up with mypy then...

rchl (approved):
Been running for a day. Seems fine.

PR #1348, opened 2020-09-27
Run TransportCallbacks on ST's worker thread

rwols:
Previously, we ran the callbacks in the reader threads. We now run them
serialized through ST's worker thread. This saves us from some boring
boilerplate code to ensure callbacks are running serially.

This diff does not yet include the send_request and send_notification methods to be running in the ST worker thread.


rwols:
This breaks a bunch of tests so I'll have to look into that.

rwols:
Seeing some interesting log lines

```
:: --> pyls textDocument/codeAction(135): {'textDocument': {'uri': 'file:///home/raoul/.config/sublime-text-3/Packages/LSP/plugin/core/transports.py'}, 'context': {'diagnostics': []}, 'range': {'end': {'character': 16, 'line': 193}, 'start': {'character': 16, 'line': 193}}}
:: <<< pyls 135: []
:: <~~ pyls 135: {'message': 'unknown response ID 135', 'code': -32602}
```

rchl:
Yes, it's broken. With eslint I don't get publishDiagnostics notification but another one instead. Looks like reading from pipe is going wrong somewhere.

This is with this PR:
Screenshot 2020-09-28 at 19 52 09

This is how it's supposed to be (notice how the one above has `eslint/status` in place of diagnostics one:
Screenshot 2020-09-28 at 19 53 59

rwols:
The last commit fixes it I think. I'm still kind of confused about it. Apparently if you don't explicitly capture the `payload` variable, then the reference in `invoke` is the same as in the read loop, and once it gets re-assigned in the read loop it's also modified in the closure.

rchl:
> The last commit fixes it I think. I'm still kind of confused about it. Apparently if you don't explicitly capture the `payload` variable, then the reference in `invoke` is the same as in the read loop, and once it gets re-assigned in the read loop it's also modified in the closure.

Oh, yes, this is something that has made me confused before too. Different from JS closures and rather unexpected.

plugin/core/transports.py
```diff
@@ -90,14 +90,22 @@ def _read_loop(self) -> None:
...
+                def invoker(p: Dict[str, Any]) -> Callable[[], None]:
+
+                    def invoke() -> None:
+                        callback_object = self._callback_object()
+                        if callback_object:
+                            callback_object.on_payload(p)
+
+                    return invoke
+
+                sublime.set_timeout_async(invoker(payload))
```
rchl, line 108: Just a potentially easier to read suggestion:
```
from functools import partial
```

```suggestion
                def invoke(p: Dict[str, Any]) -> None:
                    callback_object = self._callback_object()
                    if callback_object:
                        callback_object.on_payload(p)

                sublime.set_timeout_async(partial(invoke, payload))
```
I think this should work, not 100% sure.

rchl (approved):

PR #1350, opened 2020-09-29
Set progress status on views instead of window

rwols:
Set the server-initiated progress status on all attached views, instead of a window. This has a couple of advantages:

1) The status remains sticky
2) A notification of kind 'end' can immediately clear the sticky status
3) There can be multiple statuses per session progressing at the same time
4) For views that don't apply to the session you'll see no status reports


plugin/core/sessions.py
```diff
@@ -894,34 +899,43 @@ def m_window_workDoneProgress_create(self, params: Any, request_id: Any) -> None
...
+            data['message'] = value.get('message')  # optional
+            progress_string = self._progress_string(data, value)
+            for sv in self.session_views_async():
+                sv.view.set_status(key, progress_string)
+        elif kind == 'report':
+            progress_string = self._progress_string(data, value)
+            for sv in self.session_views_async():
+                sv.view.set_status(key, progress_string)
+        elif kind == 'end':
+            for sv in self.session_views_async():
```
rchl, line 926: I guess we are not on the right thread so this could cause issues?
But with #1348 it should be fine?

rchl (approved):

plugin/core/sessions.py
rwols: > But with #1348 it should be fine?

Yes. I guess it's time to debug that now...

jwortmann:
>     2. A notification of kind 'end' can immediately clear the sticky status

The 'end' notification can still have a message, which won't be displayed in this implementation. Maybe keep the `window.status_message()` only for the 'end' notification? Not sure if it's really necessary to keep it though, and might be a bit awkward to mix it with the "sticky" status messages shown before.

```
        /**
	 * Optional, a final message indicating to for example indicate the outcome
	 * of the operation.
	 */
	message?: string;
```

PR #1352, opened 2020-09-30
Make the workspace/applyEdit handler truly async

rwols:
The plan is to refactor some code in code_actions.py to also make use of this
technique.

### DIFF
--- a/plugin/core/edit.py
+++ b/plugin/core/edit.py
@@ -59,13 +59,10 @@ def sort_by_application_order(changes: Iterable[TextEdit]) -> List[TextEdit]:
 
 def apply_workspace_edit(window: sublime.Window, changes: Dict[str, List[TextEdit]]) -> Promise:
     """Apply workspace edits. This function must be called from the main thread!"""
-    promises = []  # type: List[Promise]
-    for fn, edits in changes.items():
+    return Promise.all([open_file(window, fn).then(partial(_apply_edits, edits)) for fn, edits in changes.items()])
 
-        def apply_edits(edits: List[TextEdit], view: Optional[sublime.View]) -> None:
-            if view and view.is_valid():
-                # Text commands run blocking. After this call has returned the changes are applied.
-                view.run_command("lsp_apply_document_edit", {"changes": edits})
 
-        promises.append(open_file(window, fn).then(partial(apply_edits, edits)))
-    return Promise.all(promises)
+def _apply_edits(edits: List[TextEdit], view: Optional[sublime.View]) -> None:
+    if view and view.is_valid():
+        # Text commands run blocking. After this call has returned the changes are applied.
+        view.run_command("lsp_apply_document_edit", {"changes": edits})
--- a/plugin/core/edit.py
+++ b/plugin/core/edit.py
@@ -1,7 +1,12 @@
 from .logging import debug
+from .promise import open_file
+from .promise import Promise
 from .typing import List, Dict, Any, Iterable, Optional, Tuple
 from .url import uri_to_filename
 import operator
+import sublime
+from functools import partial
+
 
 # tuple of start, end, newText, version
 TextEdit = Tuple[Tuple[int, int], Tuple[int, int], str, Optional[int]]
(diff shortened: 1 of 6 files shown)


plugin/core/windows.py
```diff
@@ -510,35 +508,6 @@ def discard(self, window: sublime.Window) -> None:
...
-    def fullfill(resolve: ResolveFunc) -> None:
-        global opening_files
-        # Save the promise in the first element of the tuple -- except we cannot yet do that here
-        opening_files[file_path] = (None, resolve)  # type: ignore
-
-    promise = Promise(fullfill)
-    tup = opening_files[file_path]
-    # Save the promise in the first element of the tuple so that the for-loop above can return it
-    opening_files[file_path] = (promise, tup[1])
-    return promise
```
rwols, line 539: This had to go back to promise.py, because I got entangled in circular imports.

rwols: .edit imports .windows.open_file
.windows import .sessions.get_plugin
.sessions imports .edit.apply_workspace_edit
...

plugin/core/sessions.py
```diff
@@ -802,6 +803,18 @@ def run_command(self, command: Mapping[str, Any]) -> Promise:
...
+        """
+        Apply workspace edits, and return a promise that resolves on the async thread again after the edits have been
+        applied.
+        """
+        changes = parse_workspace_edit(edit)
+        return Promise.from_main_thread().then(
+            lambda _: apply_workspace_edit(self.window, changes).then(
+                lambda _: Promise.from_async_thread()
+            )
+        )
```
rchl, line 816: This doesn't look very pretty or readable. Maybe would make sense to use line-continuation marks here? 
```suggestion
        return Promise.from_main_thread() \
            .then(lambda _: apply_workspace_edit(self.window, changes)) \
            .then(lambda _: Promise.from_async_thread())
```

plugin/core/promise.py
```diff
@@ -67,6 +69,16 @@ def fullfill_func(resolve_fn: ResolveFunc) -> None:
 
         return cls(fullfill_func)
 
+    @classmethod
+    def from_main_thread(cls) -> 'Promise':
+        """Return a promise that resolves on the main thread."""
+        return Promise(lambda resolve: sublime.set_timeout(resolve))
+
+    @classmethod
+    def from_async_thread(cls) -> 'Promise':
```
rchl, line 78: Shouldn't those allow to do a pass-through of the value?

plugin/core/edit.py
```diff
@@ -50,3 +55,17 @@ def sort_by_application_order(changes: Iterable[TextEdit]) -> List[TextEdit]:
     # we use the index in the array as the key.
 
     return list(sorted(changes, key=operator.itemgetter(0)))
+
+
+def apply_workspace_edit(window: sublime.Window, changes: Dict[str, List[TextEdit]]) -> Promise:
```
rchl, line 60: Do you recall any server that I could use to test this with?

rchl: Never mind. Any code action.

plugin/core/promise.py
rchl: Like this:

```py
    @classmethod
    def from_main_thread(cls, value: Any = None) -> 'Promise':
        """Return a promise that resolves on the main thread."""
        return Promise(lambda resolve: sublime.set_timeout(lambda: resolve(value)))

    @classmethod
    def from_async_thread(cls, value: Any = None) -> 'Promise':
        """Return a promise that resolves on the worker thread."""
        return Promise(lambda resolve: sublime.set_timeout_async(lambda: resolve(value)))
```

plugin/core/edit.py
rwols: clangd uses workspace/applyEdit.

plugin/core/promise.py
```diff
@@ -67,6 +69,16 @@ def fullfill_func(resolve_fn: ResolveFunc) -> None:
 
         return cls(fullfill_func)
 
+    @classmethod
+    def from_main_thread(cls) -> 'Promise':
```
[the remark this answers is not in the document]
rwols: Renamed to `on_*_thread`

plugin/core/sessions.py
```diff
@@ -802,6 +803,16 @@ def run_command(self, command: Mapping[str, Any]) -> Promise:
...
 
+    def _apply_workspace_edit_async(self, edit: Any) -> Promise:
+        """
+        Apply workspace edits, and return a promise that resolves on the async thread again after the edits have been
+        applied.
+        """
+        changes = parse_workspace_edit(edit)
+        return Promise.on_main_thread() \
+            .then(lambda _: apply_workspace_edit(self.window, changes)) \
+            .then(Promise.on_async_thread)
```
rchl, line 814: You don't have to do anything about it for now but we should improve Promise types so that we can indicate what value type it returns.

In this case, it will pass the value from `apply_workspace_edit` which will be an array with `None` but in this case we don't care about the return value so it should be just `None`.

plugin/core/sessions.py
```diff
@@ -833,10 +844,8 @@ def m_workspace_configuration(self, params: Dict[str, Any], request_id: Any) ->
 
     def m_workspace_applyEdit(self, params: Any, request_id: Any) -> None:
         """handles the workspace/applyEdit request"""
-        edit = params.get('edit', {})
-        self.window.run_command('lsp_apply_workspace_edit', {'changes': parse_workspace_edit(edit)})
-        # TODO: We should ideally wait for all changes to have been applied. This is currently not "async".
-        self.send_response(Response(request_id, {"applied": True}))
+        self._apply_workspace_edit_async(params.get('edit', {})).then(
+            lambda _: self.send_response(Response(request_id, {"applied": True})))
```
rchl, line 848: This could use the other formatting style also.

PR #1355, opened 2020-09-30
Expose language_id on SessionBufferProtocol

rchl:
To fix https://github.com/sublimelsp/LSP-eslint/issues/36 I need to have
access to the "languageId" of the file that the server is requesting
configuration for.

~~This changes the "params" object so its schema will no longer match
the LSP spec but then it was nowhere stated that the params passed to
on_workspace_configuration() were matching the spec.~~

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -851,9 +851,6 @@ class Session(TransportCallbacks):
         for requested_item in requested_items:
             configuration = self.config.settings.copy(requested_item.get('section') or None)
             if self._plugin:
-                buffer_view = self.get_session_buffer_for_uri_async(requested_item['scopeUri'])
-                if buffer_view:
-                    requested_item['languageId'] = buffer_view.language_id
                 self._plugin.on_workspace_configuration(requested_item, configuration)
             items.append(configuration)
         self.send_response(Response(request_id, sublime.expand_variables(items, self._template_variables())))
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -449,6 +449,19 @@ class AbstractPlugin(metaclass=ABCMeta):
         """
         return None
 
+    def get_view_for_uri(self, uri: str) -> Optional[SessionBufferProtocol]:
+        """
+        Return the view for the requested URI.
+
+        :param    uri:  The URI to get view for
+
+        :returns: An optional view if found for URI.
+        """
+        session = self.weaksession()
+        if session:
+            return session.get_session_buffer_for_uri_async(uri)
+        return None
+
 
 _plugins = {}  # type: Dict[str, Type[AbstractPlugin]]
 
@@ -851,9 +864,6 @@ class Session(TransportCallbacks):
         for requested_item in requested_items:
             configuration = self.config.settings.copy(requested_item.get('section') or None)
             if self._plugin:
-                buffer_view = self.get_session_buffer_for_uri_async(requested_item['scopeUri'])
-                if buffer_view:
-                    requested_item['languageId'] = buffer_view.language_id
                 self._plugin.on_workspace_configuration(requested_item, configuration)
             items.append(configuration)
         self.send_response(Response(request_id, sublime.expand_variables(items, self._template_variables())))
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -449,19 +449,6 @@ class AbstractPlugin(metaclass=ABCMeta):
         """
         return None
 
-    def get_view_for_uri(self, uri: str) -> Optional[SessionBufferProtocol]:
-        """
-        Return the view for the requested URI.
-
-        :param    uri:  The URI to get view for
-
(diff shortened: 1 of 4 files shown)


plugin/core/sessions.py
```diff
@@ -850,6 +851,9 @@ def m_workspace_configuration(self, params: Dict[str, Any], request_id: Any) ->
         for requested_item in requested_items:
             configuration = self.config.settings.copy(requested_item.get('section') or None)
             if self._plugin:
+                buffer_view = self.get_session_buffer_for_uri_async(requested_item['scopeUri'])
+                if buffer_view:
+                    requested_item['languageId'] = buffer_view.language_id
```
rwols, line 856: Does this have the possibility to ever clash? What if there exists a server that asks for the `languageId` section?

I think clashes can be avoided by putting it in under an `_LSP` key and stashing all of the extra information in that dict... something like `requested_item['_LSP'] = {'languageId': buffer_view.language_id}`.

plugin/core/sessions.py
```diff
@@ -850,6 +851,9 @@ def m_workspace_configuration(self, params: Dict[str, Any], request_id: Any) ->
         for requested_item in requested_items:
             configuration = self.config.settings.copy(requested_item.get('section') or None)
             if self._plugin:
+                buffer_view = self.get_session_buffer_for_uri_async(requested_item['scopeUri'])
```
rwols, line 854: It's not guaranteed that `scopeUri` is in `requested_item`.

rwols, line 854: If there's a `scopeUri`, then I think a more natural thing to do is to search for all of the views that match the scopeUri and then pass those views as a list to `on_workspace_configuration`. Since LSP-eslint is the only one using `on_workspace_configuration` can we just break API here?

rwols: The languageId is also stored in the settings by the way:

```
>>> {k: v for k, v in view.settings().to_dict().items() if 'lsp' in k}
{'lsp_active': True, 'lsp_hover_provider_count': 2, 'lsp_language': {'clangd': 'cpp'}}
```

rwols: That `lsp_language` entry is a dict for now, but it should be a plain string. https://github.com/sublimelsp/LSP/blob/31376ed0f5072144e9989dca8c406b14a57ffffb/plugin/session_view.py#L67-L74

(should be resolved once we have DocumentSelector in the config and lookups go through language-ids.sublime-settings)

rwols (changes requested):
IMO a more natural thing to do is to return a list of views that match the scope URI. The plugin can then fetch the language ID through one of the view's settings.

plugin/core/sessions.py
rchl: > Does this have the possibility to ever clash? What if there exists a server that asks for the `languageId` section?

The `requested_item` has this structure:
```ts
interface ConfigurationItem {
	/**
	 * The scope to get the configuration section for.
	 */
	scopeUri?: DocumentUri;

	/**
	 * The configuration section asked for.
	 */
	section?: string;
}
```
So I'm not sure how it could clash, regardless of what server asks for. It could only clash if LSP spec would add a `languageId` key to that interface.

Regardless, I could add it under `_LSP` object I guess. Has kinda "internal" vibe to it and it would seems a bit wrong to access that from the plugin.

rwols: You're right it probably will not result in clashes with the protocol.

rwols: Regardless I think this is the wrong approach.

plugin/core/sessions.py
rchl: Let's not require plugins to fetch stuff from the view settings. That should stay internal to LSP so that we are free to change those in breaking ways. You said yourself that the languageId is gonna be in a dict structure eventually. The API should allow getting the language ID easily, not require digging through internal view settings.

As for passing the views, it might make sense from the conceptual point of view (since `scopeUri` theoretically be a URI of a directory also), but as far as getting the languageId it doesn't help much (because I don't want to use view settings).

Maybe we can pass a list of `SessionView` (or `SessionViewProtocol`) instead? Then it should be easy to expose languageId on it.

rwols: I’m okay with passing SessionViewProtocol list :)

rwols: Well, maybe in that case add another method to AbstractPlugin that gets all the SessionViewProtocols given an URI

rwols: It makes more sense to return an Optional[SessionBufferProtocol] because you can access the list of SessionViewProtocols through that.

rchl: Sounds good.

rchl: Just realized that I have access to `Session` in `AbstractPlugin` instance and it has a `get_session_buffer_for_uri_async` method already. So I don't really need to add anything.

rwols (approved):

PR #1357, opened 2020-10-03
Use the system font for sheet css

rwols:
I want to use this in LSP-SonarLint.


predragnikolic (approved):

rchl:
Is there a reason you have applied this font only to those specific elements and not in general (to `.lsp_sheet` class)?

Issue #1340, opened 2020-09-25
Diagnostic not highlighted in view when range is empty


rchl:
**Describe the bug**
A diagnostic that has an empty range is not visible in the view (only in the gutter and in diagnostics panel).

**To Reproduce**
Receive a diagnostic like:
```js
{
  "diagnostics": [
    {
      "code": {
        "target": "https://eslint.org/docs/rules/curly",
        "value": "curly"
      },
      "message": "Expected { after 'if' condition.",
      "range": {
        "end": {
          "character": 9,
          "line": 607
        },
        "start": {
          "character": 9,
          "line": 607
        }
      },
      "severity": 1,
      "source": "eslint"
    },
```

**Expected behavior**
See where on the line the problem is. Otherwise, it's difficult to know where code action can be triggered.

**Screenshots**
![Screenshot 2020-09-25 at 22 58 01](https://user-images.githubusercontent.com/153197/94315521-f6394080-ff82-11ea-9dc6-7adffb643465.png)

**Environment (please complete the following information):**
- OS: macOS
- Sublime Text version: 4087
- LSP version: 49ea7330301078035ad56b94218a9c6edb604498
- Language servers used: eslint

PR #1358, opened 2020-10-03
Fix some issues with textDocument/didSave and old-style text sync

rwols:
See https://github.com/microsoft/language-server-protocol/issues/288

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -376,10 +376,7 @@ def normalize_text_sync(textsync: Union[None, int, Dict[str, Any]]) -> Dict[str,
     result = {}  # type: Dict[str, Any]
     if isinstance(textsync, int):
         change = {"syncKind": textsync}  # type: Optional[Dict[str, Any]]
-        if textsync > TextDocumentSyncKindNone:
-            result["textDocumentSync"] = {"didOpen": {}, "didClose": {}, "change": change}
-        else:
-            result["textDocumentSync"] = {"change": change}
+        result["textDocumentSync"] = {"didOpen": {}, "save": {}, "didClose": {}, "change": change}
     elif isinstance(textsync, dict):
         new = {}
         change = textsync.get("change")
--- a/tests/test_session.py
+++ b/tests/test_session.py
@@ -185,7 +185,8 @@ class SessionTest(unittest.TestCase):
         self.assertEqual(session.text_sync_kind(), TextDocumentSyncKindIncremental)
         self.assertTrue(session.should_notify_did_change())
         self.assertFalse(session.should_notify_will_save())  # old-style text sync will never send willSave
-        self.assertEqual(session.should_notify_did_save(), (True, False))  # old-style text sync will always send didSave
+        # old-style text sync will always send didSave
+        self.assertEqual(session.should_notify_did_save(), (True, False))
 
         session.capabilities.assign({'textDocumentSync': TextDocumentSyncKindNone})
         self.assertTrue(session.should_notify_did_open())  # old-style text sync will always send didOpen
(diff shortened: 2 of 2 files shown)


predragnikolic:
Tested this, textDocument/didSave is now send.
And yes, this fixes the issue @KevinSjoberg reported on discord.

Elm language server now sends diagnostics when I save:
image

But if you take a closer look there is no squiggly line in the view.
That is because elm language server send back the diagnostics that has the same start and end range,
```
: <-  LSP-elm textDocument/publishDiagnostics: {'diagnostics': [{'message': 'PROBLEM IN EXPOSING - I want to parse exposed values, but I am getting stuck here:\n\n1| module Asset exposing Image, src, error, defaultAvatar, loading)\n                         #^#\nExposed values are always surrounded by parentheses. So try adding a #(# here?\n\n#Note#: Here are some valid examples of `exposing` for reference:\n\n    #import# Html #exposing# (..)\n    #import# Html #exposing# (Html, div, text)\n\nIf you are getting tripped up, you can just expose everything for now. It should\nget easier to make an explicit exposing list as you see more examples in the\nwild.', 'range': {'end': {'character': 22, 'line': 0}, 'start': {'character': 22, 'line': 0}}, 'source': 'Elm', 'severity': 1}], 'uri': 'file:///home/predrag/Downloads/elm-spa-example/src/Asset.elm'}
```

Take a closer look at the range:
`'range': {'end': {'character': 22, 'line': 0}, 'start': {'character': 22, 'line': 0}}`

Personally I don't know if we should draw a squiggly line that spawns one char in cases where the start range is equal to the end range. (but that can be done in a different PR, I just wanted to share what I discovered).

predragnikolic (approved):

plugin/core/types.py
```diff
@@ -376,10 +376,7 @@ def normalize_text_sync(textsync: Union[None, int, Dict[str, Any]]) -> Dict[str,
     result = {}  # type: Dict[str, Any]
     if isinstance(textsync, int):
         change = {"syncKind": textsync}  # type: Optional[Dict[str, Any]]
-        if textsync > TextDocumentSyncKindNone:
-            result["textDocumentSync"] = {"didOpen": {}, "didClose": {}, "change": change}
-        else:
-            result["textDocumentSync"] = {"change": change}
+        result["textDocumentSync"] = {"didOpen": {}, "save": {}, "didClose": {}, "change": change}
```
predragnikolic, line 379: Because I am still learning the LSP repo, 
I just did a quick search though the LSP spec to see more about `textDocumentSync`
and the closes thing I could find regarding `textDocumentSync` is this interface:
```
export interface TextDocumentSyncOptions {
	/**
	 * Open and close notifications are sent to the server. If omitted open close notification should not
	 * be sent.
	 */
	openClose?: boolean;
	/**
	 * Change notifications are sent to the server. See TextDocumentSyncKind.None, TextDocumentSyncKind.Full
	 * and TextDocumentSyncKind.Incremental. If omitted it defaults to TextDocumentSyncKind.None.
	 */
	change?: number;
	/**
	 * If present will save notifications are sent to the server. If omitted the notification should not be
	 * sent.
	 */
	willSave?: boolean;
	/**
	 * If present will save wait until requests are sent to the server. If omitted the request should not be
	 * sent.
	 */
	willSaveWaitUntil?: boolean;
	/**
	 * If present save notifications are sent to the server. If omitted the notification should not be
	 * sent.
	 */
	save?: boolean | SaveOptions;
}
```

but this object that you modified:
 
`{"didOpen": {}, "save": {}, "didClose": {}, "change": change}`

is not adhering to that interface.
So I am kind of confused why that object has keys: `didOpen`, `save`, `didClose`, `change`,
and not `openClose`, `willSave`, `willSaveWaitUntil`, `change`, `save`.

I should probalby take a closer look at this part of the codebase to figure it out, but at first sight I find it a bit strange.

rwols: `didOpen` and `didClose` each have their own key/dict because of dynamic registration. A server can register `textDocument/didOpen` and not register `textDocument/didClose`.

rchl:
> Personally I don't know if we should draw a squiggly line that spawns one char in cases where the start range is equal to the end range. (but that can be done in a different PR, I just wanted to share what I discovered).

https://github.com/sublimelsp/LSP/issues/1340

PR #1360, opened 2020-10-05
Expose more stuff in api

rwols:
See https://github.com/sublimelsp/LSP-elixir/pull/9#discussion_r499389741

### DIFF
--- a/plugin/__init__.py
+++ b/plugin/__init__.py
@@ -1,5 +1,4 @@
 from .core.css import css
-from .core.promise import Promise
 from .core.protocol import Notification
 from .core.protocol import Request
 from .core.protocol import Response
@@ -23,7 +22,6 @@ __all__ = [
     'filename_to_uri',
     'LanguageConfig',
     'Notification',
-    'Promise',
     'register_plugin',
     'Request',
     'Response',
--- a/plugin/__init__.py
+++ b/plugin/__init__.py
@@ -1,3 +1,5 @@
+from .core.css import css
+from .core.promise import Promise
 from .core.protocol import Notification
 from .core.protocol import Request
 from .core.protocol import Response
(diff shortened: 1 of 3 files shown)


plugin/__init__.py
```diff
@@ -8,19 +10,25 @@
...
 # This is the public API for LSP-* packages
 __all__ = [
     '__version__',
     'AbstractPlugin',
     'ClientConfig',
+    'css',
+    'filename_to_uri',
     'LanguageConfig',
     'Notification',
+    'Promise',
```
rchl, line 26: Should we really do that? For `on_pre_server_command` we could just expect an optional function to be returned and internally we would use a Promise.

rchl (approved):

plugin/core/sessions.py
```diff
@@ -791,8 +792,8 @@ def _template_variables(self) -> Dict[str, str]:
 
     def run_command(self, command: Mapping[str, Any]) -> Promise:
         if self._plugin:
-            promise = self._plugin.on_pre_server_command(command)
-            if promise:
+            promise, callback = Promise.packaged_task()
```
predragnikolic, line 795: I find the usage of `Promise.packaged_task` a bit confusing. 

If you want run a callback when the `on_pre_server_command` is done, why not write `self._plugin.on_pre_server_command(command).then(callback)`?

If we don't have the `Promise.packaged_task` I would find the code easier to understand.

rchl: > If you want run a callback when the `on_pre_server_command` is done, why not write `self._plugin.on_pre_server_command(command).then(callback)`?

`.on_pre_server_command(command)` doesn't return a Promise so you can call `.then` on it.

predragnikolic: Oh, I tough it was a promise based on the old code I saw:
```diff
-            promise = self._plugin.on_pre_server_command(command)
```

PR #1366, opened 2020-10-08
Allow hot-reloading of the plugin

rwols:
When you save boot.py, all modified py files should be picked up now.

There's still some quirks going wrong, but at least this kind of works.


boot.py
```diff
@@ -1,48 +1,249 @@
...
+        self.toposort = []
+        self._backend = backend
+
+    def find_module(self, fullname, path):
+        if fullname.startswith("LSP."):
+            _toposorted_modules.append(fullname)
+        return self._backend.find_module(fullname, path)
+
+    def invalidate_caches(self) -> None:
+        return self._backend.invalidate_caches()
```
rwols, line 25: We cannot use types here, because we cannot import LSP.plugin.core.typing

rchl:
> There's still some quirks going wrong, but at least this kind of works.

I'm not sure if "kinda working" is actually better than not working...

boot.py
```diff
@@ -1,48 +1,259 @@
...
+    from .plugin.tooling import LspDumpBufferCapabilities
+    from .plugin.tooling import LspDumpWindowConfigs
+    from .plugin.tooling import LspParseVscodePackageJson
+    from .plugin.tooling import LspTroubleshootServerCommand
+
+finally:
+    sys.meta_path.pop(0)
+
+
+def _get_final_subclasses(derived: List[Type], results: List[Type]) -> None:
```
rwols, line 100: This function is going to be needed after all. Because the LSP plugin can be disabled, and then enabled again. In that case, all LSP-* are still loaded so we have to iterate over the final subclasses anyway.

rwols:
We have to have proper support for disabling and enabling this plugin anyway, so this is a good start.

PR #1368, opened 2020-10-09
Create managers on plugin load

rwols:
Previously, we created the window managers in the on_init of the global
Listener object, but we might as well do this in the plugin_loaded.
We already remove them in plugin_unloaded.

### DIFF
--- a/boot.py
+++ b/boot.py
@@ -190,6 +190,8 @@ def plugin_loaded() -> None:
     load_css()
     _register_all_plugins()
     client_configs.update_configs()
+    for window in sublime.windows():
+        windows.lookup(window)
 
 
 def plugin_unloaded() -> None:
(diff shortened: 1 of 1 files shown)


boot.py
```diff
@@ -207,19 +209,6 @@ def plugin_unloaded() -> None:
 
 
 class Listener(sublime_plugin.EventListener):
-    def _register_windows(self) -> None:
-        for w in sublime.windows():
-            windows.lookup(w)
-
-    def __del__(self) -> None:
```
predragnikolic, line 214: when does an EventListener get destroyed?

rwols: It is destroyed when the plugin is loaded, not when it is unloaded. It was my assumption that these global listener objects are destroyed when the plugin unloads, but that is not the case. You can test this by creating a folder Packages/Foo with a file foo.py with the content

```python
import sublime_plugin


class Listener(sublime_plugin.EventListener):

    def __init__(self):
        print("__init__")

    def __del__(self):
        print("__del__")
```
If you disable the "Foo" package with package control, you won't see `__del__` printed to the console. Only when you enable the package again do you see `__del__` printed, followed immediately by `__init__`.

boot.py
```diff
@@ -190,6 +190,8 @@ def plugin_loaded() -> None:
     load_css()
     _register_all_plugins()
     client_configs.update_configs()
+    for window in sublime.windows():
+        windows.lookup(window)
```
rchl, line 194: Not a new issue but it's not great that a function called `lookup` has a side effect. Could have better name right `create_window_or_return`.

PR #1369, opened 2020-10-11
Record the change_count from the UI thread, and then dispatch to the worker thread

### LINKED ISSUE #1365: Document state can get out of sync when server supports incremental sync
**Describe the bug**
In some situations, the document state can get out of sync with the server. One case where I can reproduce is when activating a view that doesn't have server started yet and immediately changing something.

**To Reproduce**
1. Activate a view that doesn't have the server started yet.
2. Immediately type something that would cause an error

**Expected behavior**
The server having a correct state of the document.

- Sublime Text version: 4087
- LSP version: 1.0.17
- Language servers used: LSP-json

**Additional context**
I believe what happening is:
1. the file is opened and immediately some text is inserted:
   - the `change_count()` (revision) is now 1
   - server receives the text content of the document as seen in revision 1
2. The `on_text_changed_async` is triggered
   - the reported change is the insert done in point 1.
   - the revision is still 1
3. Server receives a didChange notification with insert done in point 1
   - now the server has an out-of-sync state since it has applied the change to the initial state which already included that change.

Here are logs from when a couple of `,` are inserted quickly after activating the view. The logs are for when `didOpen` and then `didChange` notifications are sent:
```
DID_OPEN 1
ON_TEXT_CHANGED 1 [(5, 5, ',')]
ON_TEXT_CHANGED 1 [(6, 6, ',')]
ON_TEXT_CHANGED 2 [(7, 7, ',')]
```

rwols:
Resolves #1365 

I cannot repro the bug with these changes.

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -137,7 +137,10 @@ class TextChangeListener(sublime_plugin.TextChangeListener):
         super().detach()
 
     def on_text_changed(self, changes: Iterable[sublime.TextChange]) -> None:
-        change_count = self.buffer.primary_view().change_count()
+        view = self.buffer.primary_view()
+        if not view:
+            return
+        change_count = view.change_count()
 
         def notify() -> None:
             for listener in list(self.view_listeners):
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -136,9 +136,14 @@ class TextChangeListener(sublime_plugin.TextChangeListener):
         self.ids_to_listeners.pop(self.buffer.buffer_id, None)
         super().detach()
 
-    def on_text_changed_async(self, changes: Iterable[sublime.TextChange]) -> None:
-        for listener in list(self.view_listeners):
-            listener.on_text_changed_async(changes)
+    def on_text_changed(self, changes: Iterable[sublime.TextChange]) -> None:
+        change_count = self.buffer.primary_view().change_count()
+
+        def notify() -> None:
+            for listener in list(self.view_listeners):
+                listener.on_text_changed_async(change_count, changes)
+
+        sublime.set_timeout_async(notify)
 
     def on_reload_async(self) -> None:
         for listener in list(self.view_listeners):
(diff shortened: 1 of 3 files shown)


predragnikolic:
I tested this with the elm language server and this seems to fix the issue I am experiencing.

plugin/documents.py
```diff
@@ -136,9 +136,17 @@ def detach(self) -> None:
...
+        view = self.buffer.primary_view()
+        if not view:
+            return
+        change_count = view.change_count()
+
+        def notify() -> None:
+            for listener in list(self.view_listeners):
+                listener.on_text_changed_async(change_count, changes)
+
+        sublime.set_timeout_async(notify)
```
predragnikolic, line 149: Will try to see if the timeout is really necessary.

predragnikolic: Yeah, it works even without the timeout:
```diff
         change_count = view.change_count()
 
-        def notify() -> None:
-            for listener in list(self.view_listeners):
-                listener.on_text_changed_async(change_count, changes)
-
-        sublime.set_timeout_async(notify)
+        for listener in list(self.view_listeners):
+            listener.on_text_changed_async(change_count, changes)
```

rwols: That would run a few `Session.send_request` calls in the main thread. The idea is to move all calls to the worker thread instead, and then remove that one lock in the Session class. All the response handlers and notification handlers already run in the worker thread.

plugin/documents.py
```diff
@@ -136,9 +136,17 @@ def detach(self) -> None:
         self.ids_to_listeners.pop(self.buffer.buffer_id, None)
         super().detach()
 
-    def on_text_changed_async(self, changes: Iterable[sublime.TextChange]) -> None:
-        for listener in list(self.view_listeners):
-            listener.on_text_changed_async(changes)
+    def on_text_changed(self, changes: Iterable[sublime.TextChange]) -> None:
+        view = self.buffer.primary_view()
+        if not view:
+            return
```
rchl, line 142: There can be a case when there is no primary view? And if so, is it really the right thing to not call the listeners then?

rwols: Yeah, I am kind of conflicted about this.

It is logically impossible to not have a primary view inside of `on_text_changed`.
But it is possible to not have a primary view when doing `buffer.primary_view()` after a 100 second timeout with `sublime.set_timeout`.
The mypy type system cannot describe this runtime behavior accurately. Or maybe we are not using a feature of the type system that could accurately describe this.

rwols: Essentially what we want to describe in the type system is something like this: if T is a sublime.View or sublime.Window, then any function which takes a T, becomes a function which takes an Optional[T] whenever it is passed to `sublime.set_timeout`.

rchl: I think mypy uses such system in relation to lambdas and variables that those are capturing in their scope. But not sure now in which cases exactly.

PR #1370, opened 2020-10-11
Register and unregister plugins properly

### LINKED ISSUE #899: plugin_loaded initializes LSP-* plugins before they have a chance to run plugin_loaded

rwols:
Resolves #899 

Helper packages are encouraged to call register_plugin and unregister_plugin
in their plugin_loaded and plugin_unloaded callbacks, respectively.

There is a bug in Sublime Text's `plugin_loaded` callback. When the package is
in the list of `"ignored_packages"` in
Packages/User/Preferences.sublime-settings, and then removed from that list,
the sublime.Settings object has missing keys/values. To circumvent this, we run
the actual registration one tick later. At that point, the settings object is
fully loaded. At least, it seems that way. For more context,
see https://github.com/sublimehq/sublime_text/issues/3379
and https://github.com/sublimehq/sublime_text/issues/2099

### DIFF
--- a/boot.py
+++ b/boot.py
@@ -113,7 +113,7 @@ def _register_all_plugins() -> None:
     plugin_classes = []  # type: List[Type[AbstractPlugin]]
     _get_final_subclasses(AbstractPlugin.__subclasses__(), plugin_classes)
     for plugin_class in plugin_classes:
-        register_plugin(plugin_class)
+        register_plugin(plugin_class, notify_listener=False)
     language_handler_classes = []  # type: List[Type[LanguageHandler]]
     _get_final_subclasses(LanguageHandler.__subclasses__(), language_handler_classes)
     for language_handler_class in language_handler_classes:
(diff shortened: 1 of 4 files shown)


predragnikolic:
I haven't tried this, but in practice this sounds like a ST restart will not be required when installing LSP-*? (if the LSP-* plugin after installation calls `register_plugin`) :)

rwols:
If you use `register_plugin` as well as `AbstractPlugin.install_or_update`, then yes, no restart needed after downloading from PC.

PR #1371, opened 2020-10-11
Show diagnostics with same begin and end range in the view

### LINKED ISSUE #1340: Diagnostic not highlighted in view when range is empty

predragnikolic:
closes #1340

### DIFF
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -127,6 +127,7 @@ class SessionView:
                 self.view.erase_regions(key)
             elif ((severity <= userprefs().show_diagnostics_severity_level) and
                     (data.icon or flags != (sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE))):
+
                 def handle_same_regions(region: sublime.Region) -> sublime.Region:
                     # this allows showing diagnostics with same begin and end in the view
                     if region.a == region.b:
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -132,6 +132,7 @@ class SessionView:
                     if region.begin() == region.end():
                         return sublime.Region(region.begin(), region.begin() + 1)
                     return region
+
                 underline_regions = list(map(handle_same_regions, data.regions))
                 self.view.add_regions(key, underline_regions, data.scope, data.icon, flags)
             else:
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -127,13 +127,13 @@ class SessionView:
                 self.view.erase_regions(key)
             elif ((severity <= userprefs().show_diagnostics_severity_level) and
                     (data.icon or flags != (sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE))):
-                def handle_same_regions(region: sublime.Region):
+                def handle_same_regions(region: sublime.Region) -> sublime.Region:
                     # this allows showing diagnostics with same begin and end in the view
                     if region.begin() == region.end():
                         return sublime.Region(region.begin(), region.begin() + 1)
                     return region
-                data.regions = list(map(handle_same_regions, data.regions))
-                self.view.add_regions(key, data.regions, data.scope, data.icon, flags)
+                underline_regions = list(map(handle_same_regions, data.regions))
+                self.view.add_regions(key, underline_regions, data.scope, data.icon, flags)
             else:
                 self.view.erase_regions(key)
         listener = self.listener()
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -127,6 +127,12 @@ class SessionView:
                 self.view.erase_regions(key)
             elif ((severity <= userprefs().show_diagnostics_severity_level) and
                     (data.icon or flags != (sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE))):
+                def handle_same_regions(region: sublime.Region):
+                    # this allows showing diagnostics with same begin and end in the view
+                    if region.begin() == region.end():
+                        return sublime.Region(region.begin(), region.begin() + 1)
+                    return region
+                data.regions = list(map(handle_same_regions, data.regions))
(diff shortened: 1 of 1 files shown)


plugin/session_view.py
```diff
@@ -127,6 +127,12 @@ def present_diagnostics_async(self, flags: int) -> None:
                 self.view.erase_regions(key)
             elif ((severity <= userprefs().show_diagnostics_severity_level) and
                     (data.icon or flags != (sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE))):
+                def handle_same_regions(region: sublime.Region):
```
predragnikolic, line 130: if you have a better name for this function let me know :)

plugin/session_view.py
```diff
@@ -127,6 +127,12 @@ def present_diagnostics_async(self, flags: int) -> None:
...
+                    # this allows showing diagnostics with same begin and end in the view
+                    if region.begin() == region.end():
+                        return sublime.Region(region.begin(), region.begin() + 1)
+                    return region
+                data.regions = list(map(handle_same_regions, data.regions))
```
rwols, line 135: I would rather see this check be done in `SessionBuffer.on_diagnostics_async`, because this "present" method is now also mutating the diagnostics data which can be confusing. See line 256 of session_buffer.py.

predragnikolic: Instead of mutating the data. I could just create a new variable:
```diff
-data.regions = list(map(handle_same_regions, data.regions))
+underline_regions = list(map(handle_same_regions, data.regions))

-self.view.add_regions(key, data.regions, data.scope, data.icon, flags)
+self.view.add_regions(key, underline_regions, data.scope, data.icon, flags)
```


I would leave it here, because I only wanted to fix the underlines. I don't want to modify the response from the server. 

What do you think about that,
or do you still think that it is better to put this check in `on_diagnostics_async`?

rwols: I agree, good point

plugin/session_view.py
```diff
@@ -127,7 +127,13 @@ def present_diagnostics_async(self, flags: int) -> None:
                 self.view.erase_regions(key)
             elif ((severity <= userprefs().show_diagnostics_severity_level) and
                     (data.icon or flags != (sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE))):
-                self.view.add_regions(key, data.regions, data.scope, data.icon, flags)
+                def handle_same_regions(region: sublime.Region) -> sublime.Region:
+                    # this allows showing diagnostics with same begin and end in the view
+                    if region.begin() == region.end():
```
rwols, line 132: It's sufficient to check `region.a == region.b` here.

plugin/session_view.py
```diff
@@ -127,7 +127,13 @@ def present_diagnostics_async(self, flags: int) -> None:
...
+                        return sublime.Region(region.begin(), region.begin() + 1)
```
rwols, line 133: `region.a`

plugin/session_view.py
```diff
@@ -127,7 +127,13 @@ def present_diagnostics_async(self, flags: int) -> None:
                 self.view.erase_regions(key)
             elif ((severity <= userprefs().show_diagnostics_severity_level) and
                     (data.icon or flags != (sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE))):
-                self.view.add_regions(key, data.regions, data.scope, data.icon, flags)
+                def handle_same_regions(region: sublime.Region) -> sublime.Region:
```
rwols, line 130: Please have an empty line before the start of this lambda for readability.

plugin/session_view.py
```diff
@@ -127,7 +127,13 @@ def present_diagnostics_async(self, flags: int) -> None:
...
+                    # this allows showing diagnostics with same begin and end in the view
+                    if region.begin() == region.end():
+                        return sublime.Region(region.begin(), region.begin() + 1)
+                    return region
```
rwols, line 134: Please have an empty line after this return statement for readability.

rwols (changes requested):
LGTM modulo nitpicks.

plugin/session_view.py
predragnikolic: I initially put those `a` and `b`,
but then I changed to `begin`, and `end` methods because it looked more readable. 

but I will bring it back to `a` and `b`

plugin/session_view.py
```diff
@@ -127,7 +127,15 @@ def present_diagnostics_async(self, flags: int) -> None:
                 self.view.erase_regions(key)
             elif ((severity <= userprefs().show_diagnostics_severity_level) and
                     (data.icon or flags != (sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE))):
-                self.view.add_regions(key, data.regions, data.scope, data.icon, flags)
+
+                def handle_same_regions(region: sublime.Region) -> sublime.Region:
```
rchl, line 131: There is a flag called `sublime.DRAW_EMPTY_AS_OVERWRITE` that could be used instead.

rchl: Or `sublime.DRAW_EMPTY`

PR #1372, opened 2020-10-11
Introduce the storage_path

rwols:
* New classmethod for AbstractPlugin: `storage_path()`.
* Extra template string for expansion: `${storage_path}` for use in .sublime-settings commands.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -408,7 +408,7 @@ class AbstractPlugin(metaclass=ABCMeta):
             @classmethod
             def basedir(cls) -> str:
                 # Do everything relative to this directory
-                return os.path.joim(cls.storage_path(), cls.name())
+                return os.path.join(cls.storage_path(), cls.name())
         ```
         """
         return get_storage_path()
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -26,6 +26,7 @@ from .version import __version__
 from .views import COMPLETION_KINDS
 from .views import did_change_configuration
 from .views import extract_variables
+from .views import get_storage_path
 from .views import SYMBOL_KINDS
 from .workspace import is_subpath_of
 from abc import ABCMeta
@@ -362,6 +363,7 @@ class AbstractPlugin(metaclass=ABCMeta):
 
         These are just the values from window.extract_variables(). Additionally,
 
+        $storage_path The path to the package storage (see AbstractPlugin.storage_path)
         $cache_path   sublime.cache_path()
         $temp_dir     tempfile.gettempdir()
         $home         os.path.expanduser('~')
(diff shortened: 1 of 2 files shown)


plugin/core/sessions.py
```diff
@@ -386,6 +388,31 @@ def additional_variables(cls) -> Optional[Dict[str, str]]:
...
+        class MyPlugin(AbstractPlugin):
+
+            @classmethod
+            def name(cls) -> str:
+                return "my-plugin"
+
+            @classmethod
+            def basedir(cls) -> str:
+                # Do everything relative to this directory
+                return os.path.joim(cls.storage_path(), cls.name())
```
rchl, line 411: it says `joim`. :)

plugin/core/sessions.py
```diff
@@ -386,6 +388,31 @@ def additional_variables(cls) -> Optional[Dict[str, str]]:
         """
         return None
 
+    @classmethod
+    def storage_path(cls) -> str:
+        """
+        The storage path. Use this as your base directory to install server files. Its path is '$DATA/Package Storage'.
```
rchl, line 394: We're referring to `$DATA` in a couple of places but this variable is something that LSP came up with and it doesn't really exist. I'm not sure how clear it is to people reading this code.

rchl:
> * Extra template string for expansion: `${storage_path}` for use in .sublime-settings commands.

For npm-based packages, it won't really work, if we want to include a dynamically generated node version in the path.

plugin/core/views.py
```diff
@@ -108,8 +108,17 @@ def get_line(window: Optional[sublime.Window], file_name: str, row: int) -> str:
         return linecache.getline(file_name, row + 1).strip()
 
 
+def get_storage_path() -> str:
+    """
+    The "Package Storage" is a way to store server data without influencing the behavior of Sublime Text's "catalog".
+    Its path is '$DATA/Package Storage'.
+    """
+    return os.path.abspath(os.path.join(sublime.cache_path(), "..", "Package Storage"))
```
rchl, line 116: As it was just mentioned on Discord, this won't be correct on Windows since cache is not part of "$DATA" directory there.

predragnikolic: maybe  `sublime.installed_packages_path()` is? 
I don't have a windows os.

jwortmann:
```
>>> sublime.installed_packages_path()
'C:\\Users\\<username>\\AppData\\Roaming\\Sublime Text 3\\Installed Packages'
>>> sublime.packages_path()
'C:\\Users\\<username>\\AppData\\Roaming\\Sublime Text 3\\Packages'
>>> sublime.cache_path()
'C:\\Users\\<username>\\AppData\\Local\\Sublime Text 3\\Cache'
```
for a standard installation (not portable) on Windows and ST build 3211.

rwols: > As it was just mentioned on Discord, this won't be correct on Windows since cache is not part of "$DATA" directory there.

This *is* the correct folder for windows. The roaming appdata folder is for config files, the local appdata is for machine-dependent files.

plugin/core/sessions.py
rwols: I'll clarify this in the docstrings

plugin/core/views.py
rchl: True. I guess that when making a comment I was thinking that the ST cache dir is a direct child of `Local`.

rchl (approved):

rwols:
It's time for a release I think :)

PR #1375, opened 2020-10-12
use DRAW_EMPTY_AS_OVERWRITE for empty regions

predragnikolic:
This is a follow up PR, based on feedback:
https://github.com/sublimelsp/LSP/pull/1371#discussion_r502964924

Current behavior (that was added with #1371):
image

DRAW_EMPTY_AS_OVERWRITE:
Screenshot from 2020-10-12 19-37-59

DRAW_EMPTY:
Screenshot from 2020-10-12 19-37-03

---
I think the current behavior is ok, but I think that DRAW_EMPTY_AS_OVERWRITE is also ok.
I wouldn't go with DRAW_EMPTY because it is a vertical line.

The only disadvantage about the DRAW_EMPTY_AS_OVERWRITE is that the line in not squiggly, or a box, or whatever the user set in the LSP.sublime-settings file.
Someone could open an issue telling us that they set `"diagnostics_highlight_style": "box",` 
but they still see a line:
- notice the first line - **box**, 
- notice the line bellow `import Url exposing (Url`  - **line**
image
 

Feel free to close this PR if you think that the current behavior is ok,
or feel free to merge this if you think that the behavior of DRAW_EMPTY_AS_OVERWRITE is ok.

### DIFF
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -127,15 +127,9 @@ class SessionView:
                 self.view.erase_regions(key)
             elif ((severity <= userprefs().show_diagnostics_severity_level) and
                     (data.icon or flags != (sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE))):
-
-                def handle_same_regions(region: sublime.Region) -> sublime.Region:
-                    # this allows showing diagnostics with same begin and end in the view
-                    if region.a == region.b:
-                        return sublime.Region(region.a, region.a + 1)
-                    return region
-
-                underline_regions = list(map(handle_same_regions, data.regions))
-                self.view.add_regions(key, underline_regions, data.scope, data.icon, flags)
+                # allow showing diagnostics with same begin and end range in the view
+                flags |= sublime.DRAW_EMPTY_AS_OVERWRITE
+                self.view.add_regions(key, data.regions, data.scope, data.icon, flags)
             else:
                 self.view.erase_regions(key)
         listener = self.listener()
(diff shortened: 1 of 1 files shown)


rwols:
But are these empty regions now hoverable with the mouse?

predragnikolic:
Yes, they are.

predragnikolic:
I don't know if people would be confused by the look of DRAW_EMPTY_AS_OVERWRITE:
https://github.com/sublimelsp/LSP/issues/858#issuecomment-577271147

To be honest I would expect to see a squiggle underline| box 
instead of a straight line.

rchl:
I might be leaning towards `DRAW_EMPTY` actually since the region is actually empty so the vertical line seems more appropriate. And also it eliminates the problem of underline being displayed differently from user preference.

rwols:
I would also prefer `DRAW_EMPTY` :)

predragnikolic:
To me it all comes down weather we want to display diagnostics consistently or not :)

not so consistent: (DRAW_EMPTY)
image


consistent: (Current behavior)
image

Except for that I have no other arguments against `DRAW_EMPTY`.

rwols (approved):

rchl:
It's not perfect either way since with DRAW_EMPTY the style is not consistent but with the current approach the range is not really accurate since it's expanded to the next character.

rchl:
Hmm, after working with it for a while, I find it a bit confusing that it looks almost exactly like the "or" operator:

Screenshot 2020-10-18 at 20 46 38

(The one at the end is an error)

predragnikolic:
Feel free to revert this PR 🙂

rwols:
This PR does the right thing, because if you start changing regions of length zero into regions of length one, then you cannot distinguish between actual regions of length one, and regions of length zero (to put it in mathematical terms, the mapping is no longer injective :) )

So the only thing we can debate is the presentation/rendering of these zero-length regions.

PR #1376, opened 2020-10-15
Replace version incompatibility dialog with a console message

rchl:
Latest version of Package Control is triggering update of packages
on updating to new ST version so there is no point in showing this
message now.


rchl:
Actually changed to print a message to the console instead. Just in case it does not get updated automatically for some reason and users report a problem.

rwols (approved):

PR #1379, opened 2020-10-20
This popup still keeps disappearing for some reason

rwols:
It must have something to do with the presence of an edit object, but I can only
guess. Anyway, doing a sublime.set_timeout solves problems.


rwols:
Disregard this PR, it works.

But if the AC is a large box, it can be the case that the popup makes the AC disappear still.

rwols:
And this diff doesn't fix that.

PR #1383, opened 2020-10-22
Update troubleshooting instructions related to PATH

rchl:


rchl:
I'm not sure if for `zsh` it should be `~/.zprofile` or still `~/.profile`. I could check that later.

predragnikolic:
I use zsh. You can check, but I believe it needs to be `~/. profile` the last time I checked, which was a year ago. 

I remember trying exporting the path from ~/.zprofile but it didn't work

rchl:
It might also depend on the plaform since on macOS ST queries shell manually while it doesn't do that on Linux.

On Mac it says:
```
environment variables loaded using: /bin/zsh -l
```
so I would assume that it actually reads `~/.zprofile`.

rwols:
We should really have three subsections for each of the three platforms:

```
## Windows %PATH%

Explain how to change the %PATH%

## macOS $PATH

### bash

What to do if you use bash

### zsh

What to do if you use zsh

### fish

What to do if you use fish

## Linux $PATH

### bash

What to do if you use bash

### zsh

What to do if you use zsh

### fish

What to do if you use fish
```

rwols:
## Windows %PATH%

Click on a bunch of GUI windows

## macOS $PATH

### bash

Edit ~/.profile

### zsh

Edit ~/.zprofile (???)

### fish

Edit ~/.profile

## Linux $PATH

### bash

Edit ~/.profile

### zsh

Edit ~/.profile (???)

### fish

Edt ~/.profile

rchl:
Just tested and on macOS it's definitively `~/.zprofile` when `zsh` is the default shell. It can even be tested without restarting the user session since ST reads those manually on start.

For Linux it's always `~/.profile`, I assume, since ST doesn't manually source those so it's only relevant what OS sources for GUI session.

Does anyone want to try fish on macOS? :)

rwols:
I'm on macOS with the fish shell and the startup script for fish, namely ~/.config/fish/config.fish, isn't read by ST. ~/.profile is read.

rwols:
I'm pretty sure ~/.config/fish/config.fish is also ignored by ST on Linux.

rchl:
Does ST print any console line about sourcing environment variables?

rwols:
Whoops, sorry for the confusion. The file ~/.config/fish/config.fish **is** read by ST on macOS.

```
environment variables loaded using: /usr/local/bin/fish -l
```

rwols:
But I'm pretty sure it is not read on Linux, hence why I got confused (I also work on Ubuntu).

rwols:
I've made edits to https://github.com/sublimelsp/LSP/pull/1383#issuecomment-714511119 based on what we know now.

rchl:
Pushed some updates. If you feel like changing something feel free to push directly.

Also, I'm not sure how the table will render on "readthedocs".

rwols (approved):

rchl:
Will also merge it to master so that official docs get updated.

PR #1390, opened 2020-10-25
Ensure more requests are ran from the async thread

rwols:
- code actions
- formatting
- rename
- expand selection

I had to change quite a bit in the tests to make this work.

### DIFF
--- a/plugin/formatting.py
+++ b/plugin/formatting.py
@@ -97,13 +97,24 @@ class LspFormatDocumentCommand(LspTextCommand):
         session = self.best_session(self.capability)
         if session:
             # Either use the documentFormattingProvider ...
-            session.send_request(text_document_formatting(self.view), self.on_result)
+            req = text_document_formatting(self.view)
+
+            def run_async() -> None:
+                assert session  # TODO: How to make mypy shut up about an Optional[Session]?
+                session.send_request(req, self.on_result)
+
+            sublime.set_timeout_async(run_async)
         else:
             session = self.best_session(LspFormatDocumentRangeCommand.capability)
             if session:
                 # ... or use the documentRangeFormattingProvider and format the entire range.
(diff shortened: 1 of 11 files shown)


plugin/formatting.py
```diff
@@ -97,13 +97,24 @@ def run(self, edit: sublime.Edit, event: Optional[dict] = None) -> None:
         session = self.best_session(self.capability)
         if session:
             # Either use the documentFormattingProvider ...
-            session.send_request(text_document_formatting(self.view), self.on_result)
+            req = text_document_formatting(self.view)
+
+            def run_async() -> None:
+                assert session  # TODO: How to make mypy shut up about an Optional[Session]?
```
rchl, line 103: You could assign `session` to another variable. I think it might be complaining because it's assuming that something else might change the `session` variable in another branch.

rchl:
Not specific to this branch but I'm getting a lot of those:
```
ERROR: test_from_no_position_backwards (test_diagnostics.DiagnosticsCursorTest)
  File "/Users/me/workspace/lsp/Sublime LSP.app/Contents/MacOS/Data/Packages/LSP/tests/test_diagnostics.py", line 97, in test_from_no_position_backwards
    walker.walk(test_diagnostics)
  File "/Users/me/workspace/lsp/Sublime LSP.app/Contents/MacOS/Data/Packages/LSP/plugin/core/diagnostics.py", line 225, in walk
    for _, diagnostics in source_diagnostics:
ValueError: too many values to unpack (expected 2)
```

rwols:
> Not specific to this branch but I'm getting a lot of those:

14cf95b1ea7a939136ab1e1f3d679ae5f7853ec1 broke it. But I'll get to fixing it eventually.

plugin/formatting.py
rwols: I have a plan to fix this in a neat way. These asserts are okay for now.

PR #1392, opened 2020-10-26
Remove the responses lock in Session

rwols:
Training wheels are off now :)

Resolves #1390


plugin/goto.py
```diff
@@ -40,22 +40,17 @@ def is_enabled(self, event: Optional[dict] = None) -> bool:
...
+            session.send_request(
+                Request(self.method, params, self.view),
+                lambda response: sublime.set_timeout(
+                    lambda: self.handle_response(
+                        response, side_by_side
                     )
                 )
-
-        sublime.set_timeout_async(run_async)
+            )
```
rchl, line 53: A little less space usage:
```py
            session.send_request(
                Request(self.method, params, self.view),
                lambda response: sublime.set_timeout(lambda: self.handle_response(response, side_by_side))
            )
```

rchl (approved):
It might be a stupid and potentially unnecessary idea but I suppose LSP could automatically detect the thread the `send_request` is called on and either send a request directly or through a `set_timeout_async`.

rchl (approved):

PR #1395, opened 2020-10-27
Re-read workspace folders on starting a server

### LINKED ISSUE #1374: The "path" argument must be one of type string, Buffer, or URL. Received type object
Is it a language server problem? No
Please include in your report:

* OS and language server? MacOS/Linux, LSP-typescript
* How you installed LSP (Package Control or from git?) git
* Minimal reproduction steps.
1. Open a typescript repo in sublime merge. 
2. Open command palette and choose `Open Repository in sublime text`.
3. Open a `.js`, `.ts` file in ST.

Expected behaviour:
Start the server in the file.

Actual behaviour:
I get an  error popup:
```
Request initialize failed with message: The "path" argument must be one of type string, Buffer, or URL. Received type object
```

Screenshot 2020-10-12 at 13 46 42

And I have  to quit ST and open it to make the error go away.

Additional info,
I suspect to it is due to not having a way to detect when we open ST like that. (similar that we don't have a callback to  detect when set_project_data was called, or maybe we do have a way, but I am not aware of it).

rchl:
This is a workaround for a case when user changes workspace folders
manually for which ST provides no notification. We'll just re-read set
folders every time the user starts any server.

Resolves #1374

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -122,11 +122,17 @@ class WindowManager(Manager):
         return self._configs
 
     def on_load_project_async(self) -> None:
-        self._workspace.update()
+        self.update_workspace_folders_async()
         self._configs.update()
 
     def on_post_save_project_async(self) -> None:
         self.on_load_project_async()
+
+    def update_workspace_folders_async(self) -> None:
+        if self._workspace.update():
+            workspace_folders = self._workspace.get_workspace_folders()
+            for session in self._sessions:
+                session.update_folders(workspace_folders)
 
     def enable_config_async(self, config_name: str) -> None:
         enable_in_project(self._window, config_name)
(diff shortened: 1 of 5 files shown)


plugin/core/windows.py
```diff
@@ -243,6 +243,9 @@ def start_async(self, config: ClientConfig, initiating_view: sublime.View) -> No
             # debug('Already starting on this window:', config.name)
             return
         try:
+            # Update folders in case those have been changed by the user. There is no currently
+            # no notification in ST that would notify about changed folder list.
+            self._workspace.update()
```
rchl, line 248: Should be done earlier actually, before the checks that check if the file is part of the workspace.

rchl:
I suppose the absolutely correct fix would be to notify the server if folders change. That might be more complicated. Is this OK or should we work on a proper fix?

rchl:
I've pushed something to notify servers of changed workspace folders but it's very rushed and needs to be re-checked properly.

rwols:
Seems to work okay on my end.

rwols (approved):

PR #1399, opened 2020-10-28
Call the action commands directly, instead of going through LspRunCommandFromPoint

rwols:
For now this seems problematic because it doesn't work for the first view that
you open when starting Sublime Text.

### DIFF
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -107,25 +107,20 @@ class LspHoverCommand(LspTextCommand):
             actions = []
             for link_kind in link_kinds:
                 if self.best_session('{}Provider'.format(link_kind.lsp_name)):
-                    command = 'lsp_run_command_from_point'
-                    args = {
-                        'command_name': link_kind.subl_cmd_name,
-                        'point': point,
-                    }
-                    link = make_command_link(command, link_kind.label, args)
+                    link = make_command_link(link_kind.subl_cmd_name, link_kind.label, {'point': point})
                     if link_kind.supports_side_by_side:
-                        args['command_args'] = {'side_by_side': True}
-                        link += ' ' + make_command_link(command, '◨', args, 'icon')
+                        args = {'point': point, 'side_by_side': True}
+                        link += ' ' + make_command_link(link_kind.subl_cmd_name, '◨', args, 'icon')
                     actions.append(link)
             if actions:
-                return "<p class='actions'>" + " | ".join(actions) + "</p>"
+                return '<p class="actions">' + " | ".join(actions) + "</p>"
         return ""
 
     def diagnostics_content(self) -> str:
         formatted = []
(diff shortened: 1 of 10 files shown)


plugin/hover.py
```diff
@@ -107,25 +107,20 @@ def symbol_actions_content(self, point: int) -> str:
             actions = []
             for link_kind in link_kinds:
                 if self.best_session('{}Provider'.format(link_kind.lsp_name)):
-                    command = 'lsp_run_command_from_point'
-                    args = {
-                        'command_name': link_kind.subl_cmd_name,
-                        'point': point,
-                    }
-                    link = make_command_link(command, link_kind.label, args)
+                    link = make_command_link(link_kind.subl_cmd_name, link_kind.label, {'point': point})
```
rwols, line 110: By setting `"mdpopups.debug": 3` in my Packages/User/Preferences.sublime-settings I can see the following html output (formatted a bit):

```html
<p class="actions">
    <a href='subl:lsp_symbol_definition {&quot;point&quot;: 939}'>
        Definition
    </a>
    <a href='subl:lsp_symbol_definition {&quot;side_by_side&quot;: true, &quot;point&quot;: 939}' class='icon'>
        ◨
    </a>
     | 
    <a href='subl:lsp_symbol_references {&quot;point&quot;: 939}'>References</a>
     |
    <a href='subl:lsp_symbol_rename {&quot;point&quot;: 939}'>Rename</a>
</p>
```

This doesn't work on the first view, but using `sublime.log_commands(True)` I can actually see it being called. I have no idea why this doesn't work.

rchl: It looks like the `subl:` command triggered from the popup runs on the active view rather than on the view in which the popup was triggered.

It also looks like the `point` argument is not passed to the command so with this refactored code, the navigation to the symbol in the current view is also broken (if the cursor is not on the symbol while clicking the `Definition` link)

rchl: > It also looks like the `point` argument is not passed to the command so with this refactored code, the navigation to the symbol in the current view is also broken (if the cursor is not on the symbol while clicking the `Definition` link)

Actually, this issue is due to the `is_enabled` check preventing the command from running:
```py
    def is_enabled(self, event: Optional[dict] = None) -> bool:
        return super().is_enabled(event) and is_at_word(self.view, event)
```

rchl: > Actually, this issue is due to the `is_enabled` check preventing the command from running:

I've pushed a fix for that.

I think there are still issues with:
 - ST calling `subl:` command on the wrong view if a popup is triggered for non-active view
 - Sometimes there are issues with things not working on the first view after opening ST as you've mentioned

rchl:
This doesn't fix the linked issue but it's an improvement over original behavior as it doesn't change selection when triggering the go to command. So this could be integrated as an improvement in it's own

PR #1401, opened 2020-10-31
Run the text command from the correct view for "subl:" links

### LINKED ISSUE #1313: Goto definition from popup not working properly when view is not focused
#1303 Is probably related.
When hovering over a symbol in a view that doesnt have cursor focus, goto definition from the popup window doesn't jump to the definition of the symbol that was hovered over. Instead, it seems to jump to the definition of the symbol the cursor is at in the focused view.

rwols:
When the hover popup appears on an unfocused view, hyperlinks which point to
a text command via the `subl:` protocol run in another, focused, view. This is
undesired behavior. To fix this, focus the view just in time.

The is_enabled workarounds do not work because ST never calls is_enabled with
a `point` argument (only with an `event` argument).

Fixes #1313


rchl:
The behavior now is that the first click on `Definition` in the popup focuses the view and only the second click on it navigates to the definition.

rchl:
> The is_enabled workarounds do not work because ST never calls is_enabled with
> a `point` argument (only with an `event` argument).

To be precise, the `point` is passed to `is_enabled` when we manually trigger the command with that argument (from `subl:` link for example).

plugin/goto.py
```diff
@@ -36,9 +35,6 @@ class LspGotoCommand(LspTextCommand):
 
     method = ''
 
-    def is_enabled(self, event: Optional[dict] = None, point: Optional[int] = None) -> bool:
-        return super().is_enabled(event, point) and is_at_word(self.view, event, point)
-
```
rchl, line 41: I'm not sure why have you removed the `is_at_word` check. Is it because you thought it was not effective or because you think we should be able to trigger "go to" on non-word selection also?

rwols: I've removed unrelated changes.

rchl:
> The behavior now is that the first click on `Definition` in the popup focuses the view and only the second click on it navigates to the definition.

It still behaves like that with the latest changes.

Layout with two columns. Focus on the other view. Try to trigger a "go to definition" on a non-focused view.

rwols:
> Layout with two columns. Focus on the other view. Try to trigger a "go to definition" on a non-focused view.

Can reproduce. I only tested this with semi-transient views (open a side-by-side, keep it unfocused, and then open the popup in the unfocused side-by-side view).

plugin/core/views.py
```diff
@@ -490,9 +491,22 @@ def make_link(href: str, text: str, class_name: Optional[str] = None) -> str:
...
-                      class_name: Optional[str] = None) -> str:
-    return make_link(sublime.command_url(command, command_args), text, class_name)
+def make_command_link(command: str, text: str, command_args: Optional[Dict[str, Any]] = None,
+                      class_name: Optional[str] = None, view: Optional[sublime.View] = None) -> str:
+    if view:
+        cmd = "lsp_run_text_command_helper"
+        args = {"view_id": view.id(), "command": command, "args": command_args}  # type: Optional[Dict[str, Any]]
+    else:
+        cmd = command
+        args = command_args
```
rchl, line 501: Shouldn't it be:
```suggestion
        args = {"args": command_args}
```
?

rchl:
This one doesn't seem to do anything for the non-active view.
Screenshot 2020-11-01 at 13 20 17

plugin/core/views.py
rchl: Probably not since it uses a different command in this case.

plugin/core/views.py
```diff
@@ -490,9 +491,22 @@ def make_link(href: str, text: str, class_name: Optional[str] = None) -> str:
...
+    return make_link(sublime.command_url(cmd, args), text, class_name)
+
+
+class LspRunTextCommandHelperCommand(sublime_plugin.WindowCommand):
+    def run(self, view_id: int, command: str, args: Optional[Dict[str, Any]] = None) -> None:
+        view = sublime.View(view_id)
```
rchl, line 507: Kinda clever, kinda scary. ;)

Looking at the implementation of `View`, it seems fine but I would myself probably go for finding the view in `window.views()`.

rchl (approved):

plugin/core/views.py
rwols: What is a sublime.View, other than an OOP wrapper around an integer :) ?

PR #1405, opened 2020-11-01
Fix issues with document state getting out of sync when changing and saving

rchl:
debounced call was incorrectly running on the sync thread causing potential
race issues when "pending_changes" was modified/checked from different
threads.

Also, made text document version reported in did_change notification
represents document version at the time of the change rather than at the
time of sending the notification. Without it, we sometimes have sent
consecutive notifications with different contents but the same text document
version. Which I imagine servers don't like.

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -181,7 +181,7 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         self._color_phantoms = sublime.PhantomSet(self.view, "lsp_color")
         self._sighelp = None  # type: Optional[SignatureHelp]
         self._sighelp_renderer = ColorSchemeScopeRenderer(self.view)
-        self._text_change_listener_found = False
+        self._registered = False
 
     def __del__(self) -> None:
         settings = self.view.settings()
@@ -217,7 +217,7 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         if removed_session:
             if not self._session_views:
                 self.view.settings().erase("lsp_active")
-                self._text_change_listener_found = False
+                self._registered = False
         else:
             # SessionView was likely not created for this config so remove status here.
             session.config.erase_view_status(self.view)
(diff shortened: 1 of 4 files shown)


rchl:
I'd like to try to make a test for this but I don't have time for it now so please hold off with integrating for now.

rchl:
9e56b3b73de22afc48a8d71df212890dd25cff09 can be reproduced with something like that:

```diff
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -652,6 +657,7 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
             debug("couldn't find a text change listener for", self)
             return
         self._text_change_listener_found = True
+        sublime.set_timeout(lambda: self.view.run_command('insert', {'characters': '123'}))
         self.manager.register_listener_async(self)
         views = buf.views()
         if not isinstance(views, list):
```
and re-opening the file

predragnikolic:
I am really happy to say that these changes fixes the error that I got when modifying the view while the server was initializing :)

rwols:
Is this done?

rchl:
Probably better to get this in than waiting for me to try to figure out how to test it as it would be non-trivial and I didn't manage to do it in 1h for the second issue at least.

I've just cleaned the history and would suggest merging those without squashing as those are two separate issues.

plugin/documents.py
```diff
@@ -645,7 +651,6 @@ def _register_async(self) -> None:
         if not text_change_listener:
             debug("couldn't find a text change listener for", self)
             return
-        text_change_listener.view_listeners.add(self)
         self._text_change_listener_found = True
```
rwols, line 654: This variable name, _text_change_listener_found, is now kind of confusing. Can we rename it to `_registered` or something?

rwols: All it really needs to do is to prevent registering the view to the window over and over again.

rchl: Changed to `_registered`.

rwols (approved):

PR #1406, opened 2020-11-01
Fixup Context menu item names

rwols:
- Edit Menu:
  * rename
  * format view
  * format selection

- Selection Menu:
  * expand selection

- Goto Menu:
  * goto symbol
  * goto definition
  * goto type definition
  * goto declaration
  * goto implementation
  * goto reference

  I did not add workspace symbols here because I couldn't get it to work with
  show_overlay.

- Tools Menu:
  Has an LSP submenuw with the following items:
  * restart servers
  * toggle log panel
  * toggle diagnostics panel
  * enable langserver globally
  * disable langserver globally
  * enable langserver in project
  * disable langserver in project

I also modified the Context menu because the word "Document" is a vscode
concept and we should call it "View" (or "File"?)

I also removed a non-existent command called "Setup Language Server" in the
command palette.


Context.sublime-menu
```diff
@@ -5,35 +5,35 @@
         [
```
predragnikolic, line 5: I noticed you added a `...` suffix to each command.
I noted that that that the ST devs add `...` in the context menu only to commands that open a new application, or switch focus to a different application.

See example:
- all commands that open Sublime Merge end with `...`
- open containing folder will open up the Files app
- open Terminal here, will open up the terminal :)
- I will stop with the examples

Context.sublime-menu
```diff
@@ -5,35 +5,35 @@
...
             {
                 "command": "lsp_symbol_references",
-                "caption": "Find Symbol References"
+                "caption": "Goto Reference…"
```
predragnikolic, line 8: Instead of `Goto Reference`, 
I would suggest `Find References`

predragnikolic: oh, now I see why you choose `Goto reference`, and the `...`.
But I would still consider `Find References` more suitable.
 
image

Context.sublime-menu
```diff
@@ -5,35 +5,35 @@
...
             },
             {
                 "command": "lsp_code_actions",
-                "caption": "Code Actions"
+                "caption": "Code Actions…"
             },
             {
                 "command": "lsp_format_document",
-                "caption": "Format Document"
+                "caption": "Format View"
```
predragnikolic, line 36: I would think that `Format Document` is more user friendly.

A `View` is a ST implementation detail. Users probably don't know what a `View` is, and that is the reason why I don't think `Format View` is suitable. Imagine if it was called `Format Buffer` ). 

`Format File` is a bit shorter,  
but again, there is nothing wrong with `Format Document` either :)

Default.sublime-commands
```diff
@@ -83,19 +70,17 @@
...
-        "args": {}
     },
     {
         "caption": "LSP: Previous Diagnostic",
         "command": "lsp_previous_diagnostic",
-        "args": {}
     },
     {
-        "caption": "LSP: Workspace Symbol",
+        "caption": "LSP: Goto Symbol In Project",
```
predragnikolic, line 79: Instead of `LSP: Goto Symbol In Project`
I would suggest:
- `Workspace Symbols` (like before, maybe familiar to other people coming from VS Code, Emac, Vim...)
- `Project Symbols` (new, but shorter).

rwols:
But all of the goto commands in the Goto menu have ellipses:

image

I searched the internet about this and found this: https://ux.stackexchange.com/questions/9544/what-is-the-significance-of-the-three-dots-on-menus-and-buttons-and-how-to

If I understand the discussion on that page correctly, an ellipsis signifies that more user input is required to finish the "action".

- For the four Goto variants, it's possible a quick panel may appear requiring more input of the user. So an ellipsis seems appropriate here.
- For code actions, more user input is required to select the code action.
- For find references, no more user interaction is required so I'll change this.

Main.sublime-menu
```diff
@@ -1,4 +1,126 @@
 [
+  {
```
predragnikolic, line 2: I would not mind if this file is added,
but I would not add it.

Here is why:
- We have the context menu which let a user choose to run a LSP command (for beginner people starting with LSP)
- We have the command palette which let us access any LSP command (for semi-beginner people starting with LSP)
- We have keybindings which let user choose what LSP command to run (for people who get a feel for how LSP works, and they assigned keybindings to be more productive)

Context.sublime-menu
rwols: I went with Format File. The word "Document" is not ST-native. You and I know what it means, but an end-user who is not familiar with VSCode or the LSP spec doesn't know what it means.

Default.sublime-commands
rwols: I would like to avoid the word "workspace" because it is not ST-native. Synonymous words for ST are "window" or "project".

I'll change it to "Project Symbols", but it should really be a substitute for ST's native "Goto Symbol In Project"

Main.sublime-menu
rwols: I'm convinced we're missing out on users that download this plugin, browse the menu, see nothing, and then uninstall this plugin. Not everyone is aware of a context menu.

Context.sublime-menu
rchl: I've noticed this change (`Format file`) while using LSP and it struck me as odd because it's not formatting the file physically but just the buffer/document/view.

PR #1407, opened 2020-11-01
Fix css docment highlight of media queries

predragnikolic:
When I worked on the `_is_in_higlighted_region`,
I assumed that it was safe to not send a document nightlight request while the cursor is in a highlighted region.
Well, my assumptions were wrong.

This PR will send a document nightlight request when the cursor stops moving, 
regardless if the cursor is in a highlighted region or not.

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -299,7 +299,7 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
                 if not self._is_in_higlighted_region(current_region.b):
                     self._clear_highlight_regions()
                 self._when_selection_remains_stable_async(self._do_highlights_async, current_region,
-                                                              after_ms=self.highlights_debounce_time)
+                                                          after_ms=self.highlights_debounce_time)
             self._clear_code_actions_annotation()
             self._when_selection_remains_stable_async(self._do_code_actions, current_region,
                                                       after_ms=self.code_actions_debounce_time)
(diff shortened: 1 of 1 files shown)


rwols:
Now that you're digging through that part of the code base, there is another issue that could use some tweaks. Namely, clangd returns regions which could be *outside* the requested position. As a consequence the clever algorithm to not request again when the caret is inside a region doesn't quite work.

For instance given the following C++ file:

```cpp
#include <iostream>

int add(const int& x, const int& y)
{
    return x + y;
} 

int main()
{
    std::cout << "sum: " << add(4, 5) << "\n";
}
```
when I put the caret on the word `const`, clangd highlights the `x` occurrences in the `add` function. Would you be willing to investigate this?

predragnikolic:
@rwols Yes :)

plugin/documents.py
```diff
@@ -501,6 +501,7 @@ def _do_highlights_async(self) -> None:
 
     def _on_highlights(self, response: Optional[List]) -> None:
         if not response:
+            self._clear_highlight_regions()
```
rchl, line 504: Could have one clear at the beginning of the function since the other branch also does that.

rchl (approved):

rwols (approved):
Consider looking into my request in another pr.

So to summarize: would be nice to not see disappearing regions whenever I move my caret.

PR #1408, opened 2020-11-01
New server configuration key: auto_complete_selector

### LINKED ISSUE #933: Keep using trigger characters, or start using selectors?

rwols:
The auto_complete_selector is used to fine-tune the behavior of the AC popping
up. This selector is added to an entry of the "auto_complete_triggers" of the
view, alongside the trigger characters advertised by the language server.

When both the selector and at least one trigger matches, the AC is shown.

Note that we have to notify all SessionViews of new capability registrations,
so that they can update their trigger chars.

Resolves #933

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -507,7 +507,7 @@ class ClientConfig:
                  binary_args: Optional[List[str]] = None,  # DEPRECATED
                  tcp_port: Optional[int] = None,
                  auto_complete_selector: Optional[str] = None,
-                 allow_completion_triggers_from_server: bool = True,
+                 ignore_server_triggers: bool = False,
                  enabled: bool = True,
                  init_options: DottedDict = DottedDict(),
                  settings: DottedDict = DottedDict(),
@@ -522,7 +522,7 @@ class ClientConfig:
         self.languages = languages
         self.tcp_port = tcp_port
         self.auto_complete_selector = auto_complete_selector
-        self.allow_completion_triggers_from_server = allow_completion_triggers_from_server
+        self.ignore_server_triggers = ignore_server_triggers
         self.enabled = enabled
         self.init_options = init_options
         self.settings = settings
@@ -543,7 +543,7 @@ class ClientConfig:
             languages=_read_language_configs(s),
             tcp_port=s.get("tcp_port"),
             auto_complete_selector=s.get("auto_complete_selector"),
-            allow_completion_triggers_from_server=bool(s.get("allow_completion_triggers_from_server", True)),
+            ignore_server_triggers=bool(s.get("ignore_server_triggers", False)),
             # Default to True, because an LSP plugin is enabled iff it is enabled as a Sublime package.
             enabled=bool(s.get("enabled", True)),
             init_options=init_options,
@@ -560,7 +560,7 @@ class ClientConfig:
             languages=_read_language_configs(d),
             tcp_port=d.get("tcp_port"),
             auto_complete_selector=d.get("auto_complete_selector"),
-            allow_completion_triggers_from_server=bool(d.get("allow_completion_triggers_from_server", True)),
+            ignore_server_triggers=bool(d.get("ignore_server_triggers", False)),
             enabled=d.get("enabled", False),
             init_options=DottedDict(d.get("initializationOptions")),
             settings=DottedDict(d.get("settings")),
@@ -578,8 +578,7 @@ class ClientConfig:
             languages=languages,
             tcp_port=override.get("tcp_port", self.tcp_port),
             auto_complete_selector=override.get("auto_complete_selector", self.auto_complete_selector),
-            allow_completion_triggers_from_server=bool(override.get("allow_completion_triggers_from_server",
-                                                                    self.allow_completion_triggers_from_server)),
+            ignore_server_triggers=bool(override.get("ignore_server_triggers", self.ignore_server_triggers)),
             enabled=override.get("enabled", self.enabled),
             init_options=DottedDict.from_base_and_override(self.init_options, override.get("initializationOptions")),
             settings=DottedDict.from_base_and_override(self.settings, override.get("settings")),
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -507,7 +507,7 @@ class ClientConfig:
                  binary_args: Optional[List[str]] = None,  # DEPRECATED
                  tcp_port: Optional[int] = None,
                  auto_complete_selector: Optional[str] = None,
-                 ignore_server_triggers: bool = False,
+                 ignore_server_trigger_chars: bool = False,
                  enabled: bool = True,
                  init_options: DottedDict = DottedDict(),
                  settings: DottedDict = DottedDict(),
@@ -522,7 +522,7 @@ class ClientConfig:
         self.languages = languages
         self.tcp_port = tcp_port
         self.auto_complete_selector = auto_complete_selector
-        self.ignore_server_triggers = ignore_server_triggers
+        self.ignore_server_trigger_chars = ignore_server_trigger_chars
         self.enabled = enabled
         self.init_options = init_options
         self.settings = settings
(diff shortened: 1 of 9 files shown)


rchl:
Trying it out on default preferences.
```
>>> view.settings().get('auto_complete_triggers')
[{'characters': '<', 'selector': 'text.html, text.xml'}, {'rhs_empty': True, 'selector': 'punctuation.accessor'}, {'characters': '":', 'selector': 'punctuation.separator.mapping.key-value | (string - punctuation.definition.string.end)', 'server': 'LSP-json'}]
```

Screen Recording 2020-11-02 at 21 09 55

Isn't it supposed to prevent completions popping up within a string?

rwols:
> Isn't it supposed to prevent completions popping up within a string?

That is an escaped character, which I didn't account for in the selector. In order to not trigger AC even for backslash-escaped double-quotes, we should use

```
punctuation.separator.mapping.key-value | (string - punctuation.definition.string.end - constant.character.escape)
```

which I will update in the PR for LSP-json.

rchl:
What about the colon (also pictured in my gif)?

rwols:
> What about the colon (also pictured in my gif)?

Oh, I didn't notice that one :) that cannot be fixed because the `:` doesn't receive a separate scope.

rchl:
This is not how I've imagined the fix would work. I thought that we can just plain prevent completion from triggering for those two characters (`:` and `"`) that LSP-json sets up completions for by for example completing those with `-string` scope.

rwols:
There are a couple options here.

1. Define a mapping from trigger chars -> selector and match the keys up with the server registrations
2. Like (1), but per language
3. Allow disabling trigger characters entirely with a boolean
4. Same as (3), and allow an auto_complete_selector per language like in (2)

I would personally go for only an extra boolean to disable trigger characters, as the other options get out of hand with respect to configuration options. What do you think?

The interaction between a server's trigger chars, an ST syntax, and ST's default auto_complete_selector seems to work out okay. For instance,

- pyright uses `.` and `[` as trigger characters. These now won't trigger in undesirable locations.
- clangd's selector is modified so that it won't trigger for the `:` of a label.
- intelephense can probably benefit from a custom selector too as I'd image `>` is a trigger character for that server
- I'd imagine solargraph would have `.` as trigger char, and that is scoped as `punctuation.accesor`, so using just that selector is sufficient

In almost all of these programming languages a trigger char would not be allowed to be part of a range normally used for identifiers. The exception here is the JSON syntax which may contain trigger chars in strings (where, also somewhat irregular, we do want to allow auto-complete).

rchl:
I'll play with it a bit more, maybe it's good enough, but option 1 would allow us to disable colon in scopes like strings so it does sound better. But maybe it's not worth the complexity.

rchl:
Could we make it so that `auto_complete_selector` applies to all trigger characters if it's a string value but can also be an object in which case it will set key=>scope mappings? I feel like we could be limited by just supporting setting scope for all characters and that also makes scopes quite hard to understand and reason about.

Looking at the `auto_complete_selector` configuration in a LSP package, it's not clear to which trigger characters that scope applies so that makes it hard to tweak and understand the implication of the change.

rwols:
I went with a boolean to disable the triggers from the server.

- Some syntaxes are really good
- Triggers from some servers are bad
- In most cases you don't even have to touch auto_complete_selector and allow_trigger_characters_from_server as most servers mostly just work fine
- These two settings have lowest maintenance cost

plugin/core/types.py
```diff
@@ -506,6 +506,8 @@ def __init__(self,
                  command: Optional[List[str]] = None,
                  binary_args: Optional[List[str]] = None,  # DEPRECATED
                  tcp_port: Optional[int] = None,
+                 auto_complete_selector: Optional[str] = None,
+                 allow_completion_triggers_from_server: bool = True,
```
rchl, line 510: Since we have 2 auto-complete-related options already, maybe we could group them under the same object?

Like:

```
"auto_complete": {
  "selector": "punctuation",
  "ignore_server_triggers": True
}
```

Just a suggestion so that things are a little less messy.

rchl: Also, I find the "default false" to be a better approach since it's more clear when key being missing == False.

plugin/session_view.py
```diff
@@ -75,6 +84,63 @@ def __del__(self) -> None:
...
+        settings: sublime.Settings,
+        trigger_chars: List[str],
+        registration_id: Optional[str] = None
+    ) -> None:
+        """This method actually modifies the auto_complete_triggers entries for the view."""
+        selector = self.session.config.auto_complete_selector
+        if not selector:
+            # If the user did not set up an auto_complete_selector for this server configuration, fallback to the
+            # "global" auto_complete_selector of the view.
+            selector = str(settings.get("auto_complete_selector"))
```
rchl, line 128: So setting this one is not equivalent to omitting it completely?

sublime-package.json
```diff
@@ -118,6 +118,16 @@
...
+                },
+                "allow_completion_triggers_from_server": {
+                  "type": "boolean",
+                  "default": true,
+                  "markdownDescription": "A language server may register a list of characters that, when typed, trigger the auto-complete. These days, there are Sublime syntaxes that are so good that it is possible to detect via a selector when the auto-complete should be triggered. This can result in a much better editor experience. When this is the case for your syntax, choose an appropriate selector for the `\"auto_complete_selector\"` and set this setting to `\"false\"`.\n\n**Hint.** You may look at `view.settings().get(\"auto_complete_triggers\")` in the Console to verify your settings."
+                },
```
rchl, line 130: Should have those as own definitions and also extend the `sublime://settings/LSP-plugin-base` schema at the bottom.

plugin/core/types.py
rwols: I agree with using `ignore_server_triggers`. But I don't like having another object as config item. I'm actually trying to flatten the config because I feel that users have trouble understanding the mental model. Which is understandable, just look at our sublime-package.json. I hope to be able to write

```
"disabled_capabilities": ["completionProvider.triggerCharacters"]
```

some time in the future (but not in this PR)

plugin/core/types.py
```diff
@@ -507,7 +507,7 @@ def __init__(self,
                  binary_args: Optional[List[str]] = None,  # DEPRECATED
                  tcp_port: Optional[int] = None,
                  auto_complete_selector: Optional[str] = None,
-                 allow_completion_triggers_from_server: bool = True,
+                 ignore_server_triggers: bool = False,
```
rchl, line 510: I think `ignore_server_triggers` is not clear what the "triggers" are. I think we need longer `ignore_server_trigger_chars`.

sublime-package.json
```diff
@@ -65,39 +75,6 @@
                 "$ref": "#/definitions/LanguageConfig"
               },
             },
-            "LanguageConfig": {
```
rchl, line 68: Why was this one removed? It's referenced from `ClientLanguages`

plugin/documents.py
```diff
@@ -522,7 +522,6 @@ def _on_query_completions_async(self, promise: sublime.CompletionList, location:
         if not session:
             resolve(promise, [])
             return
-        self._apply_view_settings(session)
```
rchl, line 525: Nice one. Was bothered sometimes that the first autocomplete request is a bit wonky.

plugin/session_view.py
rchl: I'm not certain we need this.

rchl (approved):

plugin/session_view.py
rwols: Yes it at least bounds the trigger chars to whatever the global ac selector is. Previously, trigger chars would trigger in comments as well.

PR #1411, opened 2020-11-03
Send back whole action object on choosing action in showMessageRequest

rchl:
MessageActionItem is specified as only containing the "title" property [1].
Yet, vscode-eslint sends an object with "title" and "id" properties and
expects to get that "id" back when user chooses an action.

Since it's not clear what is expected I'm for now changing behavior to
send back original object. This "spec issue" is discussed in more detail at [2].

[1] https://microsoft.github.io/language-server-protocol/specification#window_showMessageRequest
[2] https://github.com/microsoft/language-server-protocol/issues/230

### DIFF
--- a/plugin/core/message_request_handler.py
+++ b/plugin/core/message_request_handler.py
@@ -11,8 +11,8 @@ class MessageRequestHandler():
         self.request_id = request_id
         self.request_sent = False
         self.view = view
-        actions = params.get("actions", [])
-        self.titles = list(action.get("title") for action in actions)
+        self.actions = params.get("actions", [])
+        self.titles = list(action.get("title") for action in self.actions)
         self.message = params.get('message', '')
         self.message_type = params.get('type', 4)
         self.source = source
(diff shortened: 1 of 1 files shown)


rwols:
@ayoub-benali is this harmless for metals or does it expect a rigid json structure?

rwols (approved):
Let's try it out then.

ayoub-benali:
Sorry haven't had the time to test with Metals today. I will do it tomorrow and will let you know.

rwols:
@ayoub-benali it has turned into a client capability, so metals will have to adapt if it can't parse the json. Though, I predict it will have no problems with this. https://github.com/sublimelsp/LSP/pull/1417

ayoub-benali:
I just tested:
```
:: <-- LSP-metals window/showMessageRequest(3): {'message': 'sbt build needs to be re-imported', 'actions': [{'title': 'Import changes'}, {'title': 'Not now'}, {'title': "Don't show again"}], 'type': 3}
:: >>> LSP-metals 3: {'title': 'Import changes'}
```

You can see that the initial json sent to the client doesn't have the key `id` in the json, so when the whole object is sent back it is still compatible with Metals.

Bottom line everything is fine :)

rchl:
Good point :)

PR #1412, opened 2020-11-03
Make the right placeholder even for servers with prepare rename support

### LINKED ISSUE #1410: Wrong placeholder in the rename quick panel
**Describe the bug**

For LSP-intelephense, it seems that the placeholder in the quick panel is the one under the caret rather than the actual symbol which triggers that. The actual renaming operation is done right though.

Unfortunately `Rename` is a premium feature in intelephense... and I can't reproduce this with LSP-pyright... weird.

**To Reproduce**

```php
<?php

class Foo
{
    public static function getInstance()
    {
        return new static();
    }

    public function abcd()
    {
        echo 'haha';
    }
}

Foo::getInstance()->abcd();
```

1. Place caret on `haha`
2. Hover on `bar`
3. Click `Rename` on the popup
4. `haha` will be in the shown quick panel

**Expected behavior**

It should be `bar` in the shown panel.

**Screenshots**

Video recording: https://streamable.com/bw762h

- OS: Win7 x64
- Sublime Text version: 4090
- LSP version: the latest commit on `st4000-exploration` 653fd3c4e25c2ed5294f76b0b8b5db79f56f18c2 
- Language servers used: intelephense

**Additional context**

I see the server communication log but I didn't see any "line 11" in it but they are all correct "line 15".

LSP log:

```
:: --> LSP-intelephense textDocument/hover(87): {'position': {'character': 23, 'line': 15}, 'textDocument': {'uri': 'file:///E:/_Test/php/src/test.php'}}
:: <<< LSP-intelephense 87: {'range': {'start': {'character': 20, 'line': 15}, 'end': {'character': 24, 'line': 15}}, 'contents': {'kind': 'markdown', 'value': '__Foo::abcd__\n\n```php\n<?php\npublic function abcd() { }\n```\n\n_@return_ `void`'}}
:: --> LSP-intelephense textDocument/prepareRename(88): {'position': {'character': 23, 'line': 15}, 'textDocument': {'uri': 'file:///E:/_Test/php/src/test.php'}}
:: <<< LSP-intelephense 88: {'start': {'character': 20, 'line': 15}, 'end': {'character': 24, 'line': 15}}
```

rwols:
Fixes #1410

cc @jfcherng

### DIFF
--- a/plugin/rename.py
+++ b/plugin/rename.py
@@ -20,23 +20,13 @@ class RenameSymbolInputHandler(sublime_plugin.TextInputHandler):
         return "new_name"
 
     def placeholder(self) -> str:
-        if self._placeholder:
-            return self._placeholder
-        return self.get_current_symbol_name()
+        return self._placeholder
 
     def initial_text(self) -> str:
         return self.placeholder()
 
     def validate(self, name: str) -> bool:
         return len(name) > 0
-
-    def get_current_symbol_name(self) -> str:
-        pos = get_position(self.view)
-        current_name = self.view.substr(self.view.word(pos))
-        # Is this check necessary?
-        if not current_name:
-            current_name = ""
-        return current_name
 
 
 class LspSymbolRenameCommand(LspTextCommand):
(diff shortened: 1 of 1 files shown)


jfcherng:
I can confirm this patch working. Thanks :+1:

jfcherng (approved):

PR #1418, opened 2020-11-05
Set haskell-language-server as the default backend

moodmosaic:
`hie` and `ghcide` are no longer recommended for Haskell IDE users. See @ndmitchell's [post](https://neilmitchell.blogspot.com/2020/09/dont-use-ghcide-anymore-directly.html) on why [haskell-language-server](https://github.com/haskell/haskell-language-server) should be the default.

This PR makes Sublime LSP work out of the box if `haskell-language-server-wrapper` is in PATH. If we get that merged, we can then simplify the [equivalent section](https://github.com/haskell/haskell-language-server/blob/fc3cc65a527a687177df9973d82ed472f56bf309/README.md#using-haskell-language-server-with-sublime-text) in the haskell-language-server README.

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -299,15 +299,17 @@
         }
       ],
     },
-    "haskell-ide-engine": {
+    "haskell-language-server": {
       "command": [
-        "hie",
+        "haskell-language-server-wrapper",
         "--lsp"
       ],
-      "languages": [
-        {
-          "languageId": "haskell"
-        }
+      "languageId": "haskell",
+      "scopes": [
+        "source.haskell"
+      ],
+      "syntaxes": [
+        "Packages/Haskell/Haskell.sublime-syntax"
       ],
     },
     "ocaml": {


rwols:
Thanks. Can you target the master branch instead of the st4000-exploration branch?

predragnikolic:
Changing the default config is always a breaking change. 

Can you add the steps for setting up `haskell-language-server` and mark the previews example as deprecated? 
https://github.com/sublimelsp/LSP/blob/st4000-exploration/docs/index.md#haskell

The users should be notified(via the release notes).
Or maybe we can add this a new default client config and mark the previous as deprecated? That way we avoid making breaking changes.

moodmosaic:
Sure!

moodmosaic:
@rwols, done: https://github.com/sublimelsp/LSP/pull/1419
@predragnikolic, I am confused; on which branch? `master` or `st4000`?

---

Let's add further comments in https://github.com/sublimelsp/LSP/pull/1419 please.

PR #1419, opened 2020-11-05
Set haskell-language-server as the default backend

moodmosaic:
`hie` and `ghcide` are no longer recommended for Haskell IDE users. See @ndmitchell's [post](https://neilmitchell.blogspot.com/2020/09/dont-use-ghcide-anymore-directly.html) on why [haskell-language-server](https://github.com/haskell/haskell-language-server) should be the default.

This PR makes Sublime LSP work out of the box if `haskell-language-server-wrapper` is in PATH. If we get that merged, we can then simplify the [equivalent section](https://github.com/haskell/haskell-language-server/blob/fc3cc65a527a687177df9973d82ed472f56bf309/README.md#using-haskell-language-server-with-sublime-text) in the haskell-language-server README.

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -323,11 +323,11 @@
       "syntaxes": ["Packages/Babel/JavaScript (Babel).sublime-syntax", "Packages/JavaScript/JavaScript.sublime-syntax"],
       "languageId": "javascript"
     },
-    "haskell-ide-engine":
+    "haskell-language-server":
     {
       "command":
       [
-        "hie",
+        "haskell-language-server-wrapper",
         "--lsp"
       ],
       "languageId": "haskell",


moodmosaic:
/cc @rwols 

@predragnikolic please let me know how you'd like to have this change and I may be able to do it.

rwols:
Does this wrapper need any special attention for Windows?

moodmosaic:
No, not at all. It even works on Windows 8.1 (I just verified as I was curious) besides Windows 10 (20H1).

moodmosaic:
If we want to be backwards compatible, as I tend to agree with what @predragnikolic suggested, we can just *add* this one to the list, instead of replacing the existing one. It's just that it's no longer recommended to use the one you have right now in `master`.

rwols:
The overall plan is actually to remove more and more default configs. People can look up the docs and copy-paste their desired configuration. So, I would suggest adding a deprecation comment in LSP.sublime-settings and adding an entry in docs/index.md.

moodmosaic:
Sure, let's do it this way, then. I should be able to open a new PR over the next coming days.

moodmosaic:
>So, I would suggest adding a deprecation comment in LSP.sublime-settings

@rwols, thank you for your suggestion. How can I add such a comment? Are there any examples in that file I could copy from?

rwols:
It's rather free-form so do whatever you feel is necessary and suitable :)

moodmosaic:
Like that?

image

rwols:
sublime settings files can contain comment lines:

```
// deprecated: please use haskell-language-server, see lsp.readthedocs.io
"haskell-ide-engine": {
   ...
}
```

rwols:
put the config in this file https://github.com/sublimelsp/LSP/blob/st4000-exploration/docs/index.md

moodmosaic:
Thanks, will do. :+1:

moodmosaic:
Hope I got this correct this time: https://github.com/sublimelsp/LSP/pull/1422

moodmosaic:
...and another PR will follow with the settings for the `st4000-exploration` branch.

moodmosaic:
>...and another PR will follow with the settings for the `st4000-exploration` branch.

https://github.com/sublimelsp/LSP/pull/1423

PR #1420, opened 2020-11-05
Implement codeAction/resolve support

rwols:
This needs tests but the basic gist is here I think.


plugin/core/sessions.py
```diff
@@ -1104,6 +1125,11 @@ def send_request(
         """You can call this method from any thread. Callbacks will run in Sublime's worker thread."""
         sublime.set_timeout_async(functools.partial(self.send_request_async, request, on_result, on_error))
 
+    def request_promise_async(self, request: Request) -> Promise:
```
rchl, line 1128: The naming feels weird as it can be understood as "request a promise" while it's sending a request and returning a promise. Also, the `_async` suffix is confusing due to the pattern that we are using to mark methods that should be called from an async thread.

`send_request_promise` sounds stupid too, as well as `send_request_p`... ;)

Maybe `send_request_task` since it uses a concept of a task internally?

plugin/core/sessions.py
```diff
@@ -897,6 +904,20 @@ def run_code_action_async(self, code_action: Mapping[str, Any]) -> Promise:
         # At this point it cannot be a command anymore, it has to be a proper code action.
         # A code action can have an edit and/or command. Note that it can have *both*. In case both are present, we
         # must apply the edits before running the command.
+        if self.has_capability("codeActionProvider.resolveSupport"):
+            # TODO: Should we accept a SessionBuffer? What if this capability is registered with a documentSelector?
+            # We must first resolve the command and edit properties, because they can potentially be absent.
+            promise = self.request_promise_async(Request("codeAction/resolve", code_action))
```
rchl, line 910: I'm thinking that maybe this resolve call could be avoided if all properties of the item from the `codeActionProvider.resolveSupport` list are already resolved or missing. But I'm not sure it's worth the extra complexity. I guess we can wait and see how servers use it.

plugin/core/sessions.py
```diff
@@ -193,6 +193,13 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
                             "source.organizeImports"
                         ]
                     }
+                },
+                "dataSupport": True,
+                "resolveSupport": {
+                    "properties": [
+                        "edit",
+                        "command"
+                    ]
```
rchl, line 202: - What about `additionalTextEdits`?
- I don't think CompletionItems have `edit` property.

rwols: I don't understand the question. This defines what properties of a code action structure we can resolve. We can resolve all properties (namely the `edit` and the `command`).

plugin/core/sessions.py
rwols: > I guess we can wait and see how servers use it.

Agreed.

plugin/core/sessions.py
rwols: renamed to send_request_task

plugin/core/sessions.py
rchl: I was thinking of and looking at the documentation of `completion/resolve` the whole time ;)

rchl (approved):

PR #1424, opened 2020-11-06
Backport some stuff to LSP3

rchl:
I want to use those in lsp_utils

### DIFF
--- a/plugin/core/handlers.py
+++ b/plugin/core/handlers.py
@@ -1,6 +1,7 @@
 from .logging import debug
 from .types import ClientConfig
 from .typing import Dict, List, Callable, Optional, Type
+from .views import get_storage_path
 import abc
 
 
@@ -19,6 +20,14 @@ class LanguageHandler(metaclass=abc.ABCMeta):
     @classmethod
     def additional_variables(cls) -> Optional[Dict[str, str]]:
         return None
+
+    @classmethod
+    def storage_path(cls) -> str:
+        """
+        The storage path. Use this as your base directory to install server files. Its path is '$DATA/Package Storage'.
+        You should have an additional subdirectory preferrably the same name as your handler.
+        """
+        return get_storage_path()
 
     @classmethod
     def instantiate_all(cls) -> 'List[LanguageHandler]':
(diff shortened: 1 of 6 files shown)


plugin/core/handlers.py
```diff
@@ -20,6 +21,14 @@ def config(self) -> ClientConfig:
     def additional_variables(cls) -> Optional[Dict[str, str]]:
         return None
 
+    @classmethod
+    def storage_path(cls) -> str:
+        """
+        The storage path. Use this as your base directory to install server files. Its path is '$DATA/Package Storage'.
+        You should have an additional subdirectory preferrably the same name as your plugin.
```
rwols, line 28: s/plugin/handler/g

plugin/core/views.py
```diff
@@ -27,8 +27,21 @@ def get_line(window: Optional[sublime.Window], file_name: str, row: int) -> str:
         return linecache.getline(file_name, row + 1).strip()
 
 
+def get_storage_path() -> str:
+    """
+    The "Package Storage" is a way to store server data without influencing the behavior of Sublime Text's "catalog".
+    Its path is '$DATA/Package Storage', where $DATA means:
+
+    - on macOS: ~/Library/Application Support/Sublime Text
```
rwols, line 35: Sublime Text 3

plugin/core/views.py
```diff
@@ -27,8 +27,21 @@ def get_line(window: Optional[sublime.Window], file_name: str, row: int) -> str:
...
+    - on Windows: %AppData%/Sublime Text/Roaming
```
rwols, line 36: Sublime Text 3

plugin/core/views.py
```diff
@@ -27,8 +27,21 @@ def get_line(window: Optional[sublime.Window], file_name: str, row: int) -> str:
...
+    - on Linux: $XDG_CONFIG_DIR/sublime-text
```
rwols, line 37: sublime-text-3 :)

rwols (approved):

plugin/core/handlers.py
rchl: I think technically the most correct term should be "package" but it's really unimportant.

PR #1426, opened 2020-11-07
Fix issue with release notes not being picked up by PC on update

rchl:
The issue was that PC wouldn't show the release notes due to there being
a prefix. PC would report that it's not a valid SemVer version in the
console.

Fix by not using version prefix in release notes. Only use it for the
git tag and github release name.

### DIFF
--- a/scripts/release.py
+++ b/scripts/release.py
@@ -22,6 +22,8 @@ with open(os.path.join(PACKAGE_PATH, '.release.json'), 'r') as f:
 RELEASE_BRANCH = CONFIGURATION['push_branch']
 # The name of the GitHub repository in <owner>/<repo> format
 GITHUB_REPO = CONFIGURATION['publish_repo']
+# The prefix to use for release version number. For example with prefix "4000" the version will be "4000-x.y.z".
+RELEASE_VERSION_PREFIX = CONFIGURATION['publish_version_prefix'] or ''
 # The name of the settings file to get the release token from ("github_token" setting)
 SETTINGS = '{}.sublime-settings'.format(__package__)
 PYTHON_VERSION_PATH = CONFIGURATION.get('python_version_path', None)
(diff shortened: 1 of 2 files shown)


rchl:
The whole time the release notes you were creating were for nothing. Nobody saw them. ;)

rwols (approved):
> The whole time the release notes you were creating were for nothing. Nobody saw them. ;)

Nice.

jwortmann:
It wasn't completely "for nothing". At least I am someone who appreciates and always reads the release notes (here on the releases page) as a nice summary over the recent feature additions and bug fixes. Big thanks to both of you for all your good work on LSP in general! :clap:

rchl:
FYI: Published new ST3 release:
 - commit: 55fed5f26c8bc7910f55e4d294db4f90de11f810
 - release: https://github.com/sublimelsp/LSP/releases

All appears to be in order and we should be getting release notes now.

I'll do the same change for ST4 also.

rchl:
> I'll do the same change for ST4 also.

Never mind, you did that already. :)

scripts/release.py
```diff
@@ -22,6 +22,8 @@
 RELEASE_BRANCH = CONFIGURATION['push_branch']
 # The name of the GitHub repository in <owner>/<repo> format
 GITHUB_REPO = CONFIGURATION['publish_repo']
+# The prefix to use for release version number. For example with prefix "4000" the version will be "4000-x.y.z".
+RELEASE_VERSION_PREFIX = CONFIGURATION['publish_version_prefix'] or ''
```
rchl, line 26: Just realized that this `or ''` won't help as it will crash without the property. Should have `get()`'ed it.

rwols:
> Big thanks to both of you for all your good work on LSP in general!

Thanks. It helps that there's a community around this package :)

PR #1427, opened 2020-11-07
Add ability to set status for all views attached to the session

rwols:
Useful for LSP-* packages.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -715,6 +715,14 @@ class Session(TransportCallbacks):
                 return sv
         return None
 
+    def set_window_status_async(self, key: str, message: str) -> None:
+        for sv in self.session_views_async():
+            sv.view.set_status(key, message)
+
+    def erase_window_status_async(self, key: str) -> None:
+        for sv in self.session_views_async():
+            sv.view.erase_status(key)
+
     # --- session buffer management ------------------------------------------------------------------------------------
 
     def register_session_buffer_async(self, sb: SessionBufferProtocol) -> None:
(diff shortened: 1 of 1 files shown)


rchl (approved):

jwortmann:
If I want to use it in `on_post_start`, am I right assuming I'm supposed to do it like this?
```python
from LSP.plugin.core.registry import best_session, sessions_for_view

class MyPlugin(AbstractPlugin):
  @classmethod
  def on_post_start(cls, window, initiating_view, workspace_folders, resolved):
    session = best_session(sessions_for_view(initiating_view), 0)
    if session:
      session.set_window_status_async("my_status_key", "my_status_message")
```

Perhaps it would be more convenient to have a simple function available for this in AbstractPlugin using a `session_name` string like in LspTextCommand (or even better without `session_name`). Or maybe add `session` as a parameter to `on_post_start`?
By the way, the `session_name` in LspTextCommand should be the same as `cls.name()`, right?

rchl:
You have access to the `Session` of your plugin through `self.weaksession()`. Use that. :)

rchl:
Oh, sorry, can't do that from `@classmethod`...

I think this should be an instance method since the session is already started.

rwols:
@jwortmann you should instead use the `__init__`:

```python
class MyPlugin(AbstractPlugin):

    def __init__(self, weaksession):
        super().__init__(weaksession)
        session = weaksession()
        session.set_window_status_async("my_status_key", "my_status_message")
        # do NOT store this session in self
        # it is available via self.weaksession()
```

rwols:
An instance is created just before the `initialized` notification is sent.

rwols:
> By the way, the session_name in LspTextCommand should be the same as cls.name(), right?

Yes :)

jwortmann:
Thanks, I will use `__init__`!

PR #1428, opened 2020-11-07
Replace "languages" by "selector"

### LINKED ISSUE #1384: Replace `"languages"` with documentSelector
While it's an improvement to use ST selectors to attach langservers to views, this is still has some drawbacks:

- The server can dynamically register capabilities with its own cooked up documentSelector. A documentSelector doesn't know anything about ST selectors (or base scopes). So we are [forced to maintain a map anyway](https://github.com/sublimelsp/LSP/blob/st4000-exploration/language-ids.sublime-settings).
- Different ST selectors could be assigned to the same language ID. But nobody can guarantee this. This forces us to maintain a map of (client config -> language ID) in SessionView. This is really clunky. (See: https://github.com/sublimelsp/LSP-angular/blob/8a7ae24f573b74d7bb92b959636c67a1093c266c/LSP-angular.sublime-settings#L4-L14)

A language ID must be **unique** per view. Therefore we should change the ClientConfig to take a DocumentSelector instead of a LanguageConfig. I'm not sure what the backwards-compatibility will look like but figure it out.

Implemented the document_selector concept, but it is not sufficient.

rwols:
- Maintain a map of base scopes -> language IDs
- Rename the awkward feature_selector to priority_selector
- Solve the problem of non-unique language IDs in SessionView
- Update all default clients
- "languages", "document_selector", "syntaxes" are all backwards-compatible
- But not if you use LanguageConfig in your plugin somewhere

This is a minor version update because the public API changed.

Resolves #1384


rchl:
Will try to look at it later today.

rchl:
My not-very-clean ST shows one test fail:
```
FAIL: test_can_parse_client_settings_with_languages (test_configs.ConfigParsingTests)
  File "/Users/rchlodnicki/workspace/lsp/Sublime LSP.app/Contents/MacOS/Data/Packages/UnitTesting/unittesting/core/py33/case.py", line 26, in _executeTestPart
    deferred = function()
  File "/Users/rchlodnicki/workspace/lsp/Sublime LSP.app/Contents/MacOS/Data/Packages/LSP/tests/test_configs.py", line 28, in test_can_parse_client_settings_with_languages
    self.assertEqual(config.selector, "source.python")
AssertionError: '' != 'source.python'
+ source.python
```

rchl (approved):
This is a bit over my head right now but it seems to work so I approve. :)

PR #1430, opened 2020-11-08
Use all types from "typing" module on >=3.8.0

rchl:


rchl:
This actually fixes some type issues that Pyright has been reporting in `code_actions.py`.

rwols (approved):

PR #1431, opened 2020-11-09
Remove the "auto_complete_preserve_order" hack

rwols:
This was from the time that we didn't use the filterText as the trigger.
I don't think it's necessary to have this anymore. We don't need to
overwrite the user's personal preferences.


rchl (approved):
I think it's fine but time will tell.

The relevant discussion is at https://github.com/sublimelsp/LSP/pull/866#discussion_r379845105 but I haven't re-verified that the old issue doesn't exist anymore.

PR #1433, opened 2020-11-10
Put the menu items from the Context menu in the top Menu bar as well

rwols:
This helps discovery of capabilities for inexperienced users. Let the discussions begin :)


rchl (approved):
Good for discoverability 👍

PR #1434, opened 2020-11-10
Check for ValueError as well

pfultz2:
Json loads can throw ValueError if it has trouble parsing the json.


rwols (approved):
Thanks Paul!

PR #1438, opened 2020-11-10
Check for a listener for hover

### LINKED ISSUE #1303: LspHoverCommand should check whether the view is actually under the session's control
We try to find the most suitable session in LspHoverCommand, but that session may not actually have the view under document sync. For example, when hovering over a symbol with the mouse for a transient view via CTRL+P.

rwols:
When there's no ViewListener, the view is not under control of a language
server. So we should not request any textDocument/* requests.

textDocument/hover is the only one where this can happen due to mouse
interaction and transient views during CTRL+P file browsinng.

Fixes #1303

### DIFF
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -65,11 +65,7 @@ class LspHoverCommand(LspTextCommand):
         window = self.view.window()
         if not window:
             return
-        wm = windows.lookup(window)
-        listener = wm.listener_for_view(self.view)
-        if not listener:
-            return
-        self._base_dir = wm.get_project_path(self.view.file_name() or "")
+        self._base_dir = windows.lookup(window).get_project_path(self.view.file_name() or "")
         self._hover = None  # type: Optional[Any]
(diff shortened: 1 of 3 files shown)


plugin/hover.py
```diff
@@ -65,7 +65,11 @@ def run(self, edit: sublime.Edit, point: Optional[int] = None, event: Optional[d
         window = self.view.window()
         if not window:
             return
-        self._base_dir = windows.lookup(window).get_project_path(self.view.file_name() or "")
+        wm = windows.lookup(window)
+        listener = wm.listener_for_view(self.view)
+        if not listener:
+            return
```
rchl, line 71: Would checking if there exists a `hoverProvider` session be even better?

rchl: Maybe not since there is some code that fetches code actions for diagnostics. But then maybe just checking whether any session is running?

rwols: Your question inspired me to tackle the problem in a better way. See 23bd979

rchl (approved):

PR #1439, opened 2020-11-10
Implement window/showDocument server request

rwols:
Implements https://github.com/microsoft/language-server-protocol/commit/ec377055d48d1bc79bbee6ad89c23ec428059f8f

### DIFF
--- a/plugin/core/open.py
+++ b/plugin/core/open.py
@@ -1,0 +1,86 @@
+from .logging import exception_log
+from .promise import Promise
+from .promise import ResolveFunc
+from .protocol import Range
+from .typing import Any, Dict, Tuple, Callable, Optional
+from .url import uri_to_filename
+from .views import range_to_region
+import os
+import sublime
+import subprocess
+import webbrowser
+
+
+opening_files = {}  # type: Dict[str, Tuple[Promise, Callable[[Optional[sublime.View]], None]]]
+
+
+def open_file(window: sublime.Window, file_path: str, flags: int = 0, group: int = -1) -> Promise:
+    """Open a file asynchronously. It is only safe to call this function from the UI thread."""
+    view = window.open_file(file_path, flags, group)
+    if not view.is_loading():
+        # It's already loaded. Possibly already open in a tab.
+        return Promise.resolve(view)
+
+    # Is the view opening right now? Then return the associated unresolved promise
+    for fn, value in opening_files.items():
+        if fn == file_path or os.path.samefile(fn, file_path):
+            # Return the unresolved promise. A future on_load event will resolve the promise.
+            return value[0]
+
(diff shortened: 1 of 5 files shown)


rwols:
I haven't actually tested this, but it might be a good basis to start out with.

ayoub-benali:
Really cool, Metals have a custom request that does the same thing. Created an issue https://github.com/scalameta/metals-feature-requests/issues/164 to migrate it once the 3.16 spec is official.

plugin/core/sessions.py
```diff
@@ -236,12 +238,16 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
...
+            "showDocument": {
+                "support": True
+            },
             "showMessage": {
                 "messageActionItem": {
                     "additionalPropertiesSupport": True
                 }
             },
             "workDoneProgress": True
+
```
rchl, line 250: Oh no, no, no, no.

plugin/core/sessions.py
```diff
@@ -1032,6 +1038,21 @@ def m_client_unregisterCapability(self, params: Any, request_id: Any) -> None:
...
+        def success(b: bool) -> None:
+            self.send_response(Response(request_id, {"success": b}))
+
+        if params.get("external"):
+            success(open_externally(uri, bool(params.get("takeFocus"))))
+        else:
+            # TODO: ST API does not allow us to say "do not focus this new view"
+            filename = uri_to_filename(uri)
+            selection = params.get("selection")
+            open_file_and_center_async(self.window, filename, selection).then(lambda _: success(True))
```
rchl, line 1054: Should we take the `open_file_*` take a region right away? Feels weird to pass this poorly typed Dict.

plugin/core/open.py
```diff
@@ -0,0 +1,86 @@
...
+    A blocking function that invokes the OS's "open with default extension"
+    """
+    if uri.startswith("http:") or uri.startswith("https:"):
+        return webbrowser.open(uri, autoraise=take_focus)
+    file = uri_to_filename(uri)
+    try:
+        # TODO: handle take_focus
+        if sublime.platform() == "windows":
+            # os.startfile only exists on windows, but pyright does not understand sublime.platform().
+            # TODO: How to make pyright understand platform-specific code with sublime.platform()?
```
rchl, line 77: I don't know.

rchl (approved):

plugin/core/sessions.py
rwols: I pondered on this for a while and I think we should keep using a loose dict. The larger issue is that Range is such a weird class. We have sublime.Region which can already do everything that Range can do. Perhaps we can remove the Range class in the future?

PR #1442, opened 2020-11-12
Link to the client configuration also for C/C++

rchl:
As with all other languages, this one should also link to client
configuration page. That one also links to the guide.


rchl:
What I think is also missing from `https://lsp.readthedocs.io/en/latest/#cc` is how to enable the client/server. So something like "LSP: Enable Language Server Globally" or an example configuration. The guide itself doesn't cover that either.

rwols (approved):

PR #1443, opened 2020-11-12
Advertise support for regular expression syntax "ECMAScript"

rwols:
This is only used for completions at the moment. With respect to completions,
the regex flavor is actually Perl, but Perl is a strict *superset* of ECMAScript.

Because the spec only lists one "well-known" flavor, we restrict ourselves to
this subset to achieve maximum compatibility with language servers that may
check for this particular value.

I have intentionally left out the version because:

- I suspect language servers will not check this value rigorously, and
- I don't actually know what version of ECMAScript corresponds to ST's internal
  implementation.

See: https://github.com/microsoft/language-server-protocol/commit/d1bfc1014be2b530d85223f2b090c8a0c318eaad
See: https://www.sublimetext.com/docs/completions.html#snippets
See: https://www.boost.org/doc/libs/1_64_0/libs/regex/doc/html/boost_regex/syntax/perl_syntax.html

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -115,7 +115,7 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
             "regularExpressions": {
                 "engine": "ECMAScript"
             }
-        }
+        },
         "textDocument": {
             "synchronization": {
                 "dynamicRegistration": True,  # exceptional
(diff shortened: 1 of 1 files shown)


rchl:
> Perl is a strict _superset_ of ECMAScript.

Say what?
Do you mean the regexp implementations used by those languages?

rchl (approved):

rwols:
> Do you mean the regexp implementations used by those languages?

I think so yes :) I'm actually not entirely sure of my statements. Is every valid regular expressions in ecmascript also a valid regular expression in perl?

rwols:
If I can believe this comparison https://web.archive.org/web/20130830063653/http://www.regular-expressions.info/refflavors.html then yes.

Issue #535, opened 2019-03-20
Path mapping for remote tcp pyls


imposeren:
Usually I prefer to run my python projects in a docker environment, and there I usually have paths that differ from my local paths (sublime is running locally), e.g.: `/apps/project1/repo1` <-> `/home/USERNAME/work/project1/repo1`, and also all the dependencies are not installed locally, and are only available in docker.

I know I can:
1. start pyls in the docker manually
2. use port-mapping to expose server on localhost
3. use `tcp_host`, `tcp_port` in sublime-LSP settings, 
...
4. but paths are still not correct. 

Similar workflow can relatively easy be handled in [SublimeAnaconda](http://damnwidget.github.io/anaconda/):
```javascript
{  //project settings
    "folders": [
        {...}
    ],
    "settings": {
        "python_interpreter": "tcp://localhost:19362?pathmap=/home/USERNAME/work/project1/meta-repo/volumes/pipenv/,/apps/other-repo/.venv/&pathmap=/home/USERNAME/work/project1/,/apps/&extra=/apps&extra=/apps/tika/src&interpreter=/apps/other-repo/.venv/bin/python3.6m"
    }
}
```

Under the hood SublimeAnaconda server running in docker simply replaces matching parts of the filepaths when it receives request from sublime client one way, and then when sends anything with file path back to client, then it  replaces parts of path the other way.

There is no problem with "synchronization" because I'm using my local folders as volumes in docker

I'd like to implement something similar for LSP plugin, but I might need some advises:
1. Is it at all viable idea to try implementing this?
2. "Dumb" approach might be:
   * find all the calls to `plugins.core.url.filename_to_uri`, and add some extra argument to it (e.g. settings), 
   * in function `filename_to_uri` itself: do some transformations if some setting is set.
   * But I'm not sure what is the best way to get current settings (something like `plugin.core.configurations.get_scope_client_config(current_view, ...HOW_TOGET_CONFIGS, WHAT_IS_A_POINT)`)
   * I don't know where should "backward" transformation be done, and not sure that such place exists at all.
3. Is it better to do on the client (this repo), or maybe it's better to explore server-side changes? So that a `pyls` server makes all the work.

[...]

PR #1445, opened 2020-11-12
Introduce path mapping classes

rwols:
Path maps can be defined for each server configuration.
There is still the issue of actually applying the path maps everywhere. I will
defer this to another PR because it's huge.

I will also add the settings in sublime-package.json once everything is
actually working.

Related (but doesn't resolve): #535


rwols:
Note that this could also have been a workaround for the broken vue language server by translating `C:` to `c:` :)

plugin/core/types.py
```diff
@@ -621,6 +678,27 @@ def match_view(self, view: sublime.View) -> bool:
...
+                if mapped:
+                    break
+        return urllib.parse.urljoin('file:', urllib.request.pathname2url(path))
+
+    def map_server_uri_to_client_path(self, uri: str) -> str:
+        if os.name == 'nt':
+            # url2pathname does not understand %3A (VS Code's encoding forced on all servers :/)
+            path = urllib.request.url2pathname(urllib.parse.urlparse(uri).path).strip('\\')
+        else:
+            path = urllib.request.url2pathname(urllib.parse.urlparse(uri).path)
```
rchl, line 694: I wonder what cases it's trying to handle. Some tests would be nice to have to understand it better. But I guess Windows tests would have to run on Windows?

rwols: I took this directly from our good ol' `uri_to_filename` function in plugin/core/url.py

rchl: Why are we not using it directly then? :)

rwols: This is meant to replace `uri_to_filename` and `filename_to_uri` eventually. But for the time being I'll reuse the functions.

rchl (approved):

PR #1447, opened 2020-11-14
Add more plugin callbacks

### LINKED ISSUE #1373: Official way for deferring specifying the client "command" from plugin
As mentioned in https://github.com/sublimelsp/LSP/issues/899#issuecomment-706703808, it's possible to not specify the path for `command` from `AbstractPlugin.plugin_configuration()` and then manually update it later from `AbstractPlugin.can_start()`.

It works but it's neither clean nor obvious to users of the API that don't study the LSP code. It's not specified anywhere that the user is allowed to modify the `ClientConfig` like that (maybe the config was propagated to some other parts of the code already? Maybe `command` is not really considered public property?).

It basically looks like a hack to do it as LSP-Dart and LSP-flow are doing it and I think that we should have a clearer and more explicit API for that. I'm not sure how it should look like though.

rwols:
- on_pre_start: do just-in-time adjustments for the startup info here
- on_post_start: extra callback just in case :)
- ~~on_pre_window_settings~~ on_pre_workspace_configuration: do just-in-time adjustments for the server settings
  here

As a bonus there's much better separation between construction of the
jsonrpc transport and resolvement of settings.

Resolves #1373

### DIFF
--- a/plugin/core/collections.py
+++ b/plugin/core/collections.py
@@ -1,6 +1,7 @@
 """
 Module with additional collections.
 """
+import sublime
 from .typing import Optional, Dict, Any
 from copy import deepcopy
 
@@ -130,6 +131,16 @@ class DottedDict:
                 self._update_recursive(value, key)
             else:
                 self.set(key, value)
+
+    def resolve(self, variables: Dict[str, str]) -> "DottedDict":
+        """
+        Resolve a DottedDict that may potentially contain template variables like $folder
+
+        :param      variables:  The variables
+
+        :returns:   A copy of this DottedDict, but with the variables replaced
+        """
+        return DottedDict(sublime.expand_variables(self._d, variables))
 
     def _update_recursive(self, current: Dict[str, Any], prefix: str) -> None:
         if not current:
--- a/plugin/core/collections.py
+++ b/plugin/core/collections.py
@@ -1,9 +1,9 @@
 """
 Module with additional collections.
 """
-import sublime
 from .typing import Optional, Dict, Any
 from copy import deepcopy
+import sublime
 
 
 class DottedDict:
@@ -132,7 +132,7 @@ class DottedDict:
             else:
                 self.set(key, value)
 
-    def resolve(self, variables: Dict[str, str]) -> "DottedDict":
+    def create_resolved(self, variables: Dict[str, str]) -> "DottedDict":
         """
         Resolve a DottedDict that may potentially contain template variables like $folder
 
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -475,14 +475,14 @@ class AbstractPlugin(metaclass=ABCMeta):
                      workspace_folders: List[WorkspaceFolder], resolved: ResolvedStartupConfig) -> Optional[str]:
         """
         Callback invoked just before the language server subprocess is started. This is the place to do last-minute
-        adjustments to your "command" or "initializationOptions" in the passed-in "resolved" argument. You can also
-        choose to return a custom working directory, but consider that a language server shouldn't care about the
-        working directory.
+        adjustments to your "command" or "initializationOptions" in the passed-in "resolved" argument, or change the
+        order of the workspace folders. You can also choose to return a custom working directory, but consider that a
+        language server should not care about the working directory.
 
         :param      window:             The window
         :param      initiating_view:    The initiating view
-        :param      workspace_folders:  The workspace folders
-        :param      resolved:           The resolved configuration
+        :param      workspace_folders:  The workspace folders, you can modify these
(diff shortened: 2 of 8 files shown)


plugin/core/collections.py
```diff
@@ -1,6 +1,7 @@
 """
 Module with additional collections.
 """
+import sublime
 from .typing import Optional, Dict, Any
 from copy import deepcopy
```
rchl, line 6: not in order :)

plugin/core/collections.py
```diff
@@ -131,6 +132,16 @@ def update(self, d: Dict[str, Any]) -> None:
             else:
                 self.set(key, value)
 
+    def resolve(self, variables: Dict[str, str]) -> "DottedDict":
+        """
+        Resolve a DottedDict that may potentially contain template variables like $folder
```
rchl, line 137: Even though the `returns` documentation mentions it, it could be clearer from the name that it creates a copy instead of updating itself. Maybe `create_resolved()`?

It would potentially be even clearer if this would be a class method to which you would be a dict.

plugin/core/sessions.py
```diff
@@ -469,23 +470,62 @@ def can_start(cls, window: sublime.Window, initiating_view: sublime.View,
...
         """
-        Constructs a new instance.
+        Constructs a new instance. Your instance is constructed after a response to the initialize request.
 
         :param      weaksession:  A weak reference to the Session. You can grab a strong reference through
                                   self.weaksession(), but don't hold on to that reference.
         """
         self.weaksession = weaksession
 
+    def on_pre_window_settings(self, settings: DottedDict) -> None:
```
rchl, line 513: We have `on_workspace_configuration` which follows the LSP naming and you are adding one that completely doesn't match LSP naming. Can we be consistent with the existing method and call it `on_pre_workspace_configuration` or similar? For plugin authors, it might be clearer to follow the LSP naming as they should be familiar with that. This is not made for normal ST users.

plugin/core/sessions.py
```diff
@@ -469,23 +470,62 @@ def can_start(cls, window: sublime.Window, initiating_view: sublime.View,
...
+
     def on_workspace_configuration(self, params: Dict, configuration: Any) -> None:
         """
         Override to augment configuration returned for the workspace/configuration request.
 
         :param      params:         A ConfigurationItem for which configuration is requested.
         :param      configuration:  The resolved configuration for given params.
         """
-        pass
+        return None
```
rchl, line 528: It's not really expected to receive anything from this API so I think `pass` makes more sense semantically.

plugin/core/sessions.py
```diff
@@ -469,23 +470,62 @@ def can_start(cls, window: sublime.Window, initiating_view: sublime.View,
...
+        working directory.
+
+        :param      window:             The window
+        :param      initiating_view:    The initiating view
+        :param      workspace_folders:  The workspace folders
+        :param      resolved:           The resolved configuration
+
+        :returns:   A desired working directory, or None if you don't care
+        """
+        return None
```
rchl, line 489: Does it allow to alter `workspace_folders`? I feel that if we would have a use-case for changing the workspace directory, we would as well want to set workspace folders to something else.

(I'm thinking of a use-case where we have to read some configuration from disk in which it would be defined what are the roots of the projects. But that would probably be even more complicated to get right...)

plugin/core/collections.py
rwols: I went with create_resolved. I am too lazy to type out the name DottedDict if it were a classmethod :)

plugin/core/sessions.py
rwols: I added more wording to signal that it's okay to change the workspace folders.

rchl: It probably wouldn't work very well to change cwd as I imagine the server would still only be able to start once per window. But that's not important right now.

rchl (approved):

rchl:
The intention was that pyright would use those instead of the current code?

```py
    def on_settings_read(cls, settings: sublime.Settings) -> bool:
        super().on_settings_read(settings)

        if settings.get("dev_environment") == "sublime_text":
            server_settings = settings.get("settings", {})  # type: Dict[str, Any]

            # add package dependencies into "python.analysis.extraPaths"
            extraPaths = server_settings.get("python.analysis.extraPaths", [])  # type: List[str]
            extraPaths.extend(cls.find_package_dependency_dirs())
            server_settings["python.analysis.extraPaths"] = extraPaths

            settings.set("settings", server_settings)

        return False
```
?

Looks like `on_pre_start` is not suitable as it doesn't have access to `settings`. So is the intention that `on_pre_workspace_configuration` is used and settings are altered every time the configuration is sent to the server?

Issue #1356, opened 2020-10-01
Allow the user to invoke code actions manually 


ayoub-benali:
**Is your feature request related to a problem? Please describe.**
Some LSP servers like Metals expects `source.organizeImports` to be invoked manually, after the document is saved and compiled https://github.com/scalameta/metals/pull/2104#discussion_r496601113.
So adding `source.organizeImports` to `lsp_code_actions_on_save` setting wouldn't work and triggering the LSP code action in the import area yields "No action available".  

**Describe the solution you'd like**
LSP could provide a sublime command that takes the name of the code action as an argument so that the user can setup a key binding, or better `LSP-*` packages would provide these known code actions out of the box.

**Describe alternatives you've considered**
vscode seems to put the feature under a "source actions" menu: https://github.com/scalameta/metals/pull/1971#issuecomment-687085846 
![gif](https://user-images.githubusercontent.com/7843237/92234211-e5c90500-eeb1-11ea-8329-da98bd886094.gif) 

While Vim client seems to list all the `codeActionLiteralSupport` content: https://github.com/scalameta/metals/pull/1971#issuecomment-687139180
![gif](https://user-images.githubusercontent.com/13974112/92243382-b5896280-eec1-11ea-95f4-6545ef130ae9.gif)

rchl:
Here are literal code actions reported by some servers:
```
LSP: LSP-eslint: supported code action kinds: ['quickfix', 'source.fixAll.eslint']
LSP: LSP-pyright: supported code action kinds: ['quickfix', 'source.organizeImports']
```
In those cases it would only make sense to expose the `source.*` ones as `quickfix` wouldn't do anything. What are the code actions supported by metals?

As for supporting that feature we could:
 - expose "LSP: Source actions..." in the quick panel that would list those for active servers
 - add context menu item `LSP -> Source actions...`. That would probably be a bit confusing next to `code actions` though...

> or better `LSP-*` packages would provide these known code actions out of the box.

In what form? Just extend the command palette? In that case, every source action would have a separate entry.

ayoub-benali:
> What are the code actions supported by metals?
```
'codeActionKinds': ['quickfix', 'refactor', 'source.organizeImports']}
```

> expose "LSP: Source actions..." in the quick panel that would list those for active servers

I think this would be fine but can the user also set keyboard shortcut with this form ? I want to avoid going through the quick panel each time since I will trigger the same code action. 

> Maybe on all of them? Or allow specifying config_name also?

[...]

PR #1448, opened 2020-11-14
Add support for running source actions

rchl:
Expose a list of source actions that are provided by the servers.
Servers announce support for those through the
"codeActionProvider.codeActionKinds" capability.

### DIFF
--- a/Context.sublime-menu
+++ b/Context.sublime-menu
@@ -29,7 +29,12 @@
             },
             {
                 "command": "lsp_code_actions",
-                "caption": "Apply Code Action…"
+                "caption": "Code Action…"
+            },
+            {
+                "command": "lsp_code_actions",
+                "args": {"only_kinds": ["source"]},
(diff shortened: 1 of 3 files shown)


rchl:
Partially resolves #1356. That issue also asks for being able to run specific code actions from a shortcut/pallette command.

Context.sublime-menu
```diff
@@ -29,7 +29,12 @@
             },
             {
                 "command": "lsp_code_actions",
-                "caption": "Apply Code Action…"
+                "caption": "Code Action…"
```
rchl, line 32: Changed caption here as this command doesn't apply right away, it requires further input and then applies.

rwols: It should also be changed in Main.sublime-mennu

rwols: Never mind, it's not in the main menu

rwols (approved):

PR #1452, opened 2020-11-15
Use the Sheet to check if a view is transient

rwols:
Idea from @Ultra-Instinct-05

See: https://github.com/sublimehq/sublime_text/issues/3748

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -53,19 +53,10 @@ _kind2name = {
 }
 
 
-def is_transient_view(view: sublime.View) -> bool:
-    window = view.window()
-    if not window:
-        return True
-    if window.get_view_index(view)[1] == -1:
-        return True  # Quick panel transient views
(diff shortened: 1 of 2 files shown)


plugin/documents.py
```diff
@@ -53,19 +53,10 @@
...
-    if window.get_view_index(view)[1] == -1:
-        return True  # Quick panel transient views
-    return view == window.transient_view_in_group(window.active_group())
-
-
 def is_regular_view(v: sublime.View) -> bool:
     # Not from the quick panel (CTRL+P), must have a filename on-disk, and not a special view like a console,
     # output panel or find-in-files panels.
-    return not is_transient_view(v) and bool(v.file_name()) and v.element() is None
+    return not v.sheet().is_transient() and bool(v.file_name()) and v.element() is None
```
rchl, line 59: Not 100% confident that it will cover all cases but let's find out.

rchl (approved):

PR #1453, opened 2020-11-15
Expose ResolvedStartupConfig in plugin API

rchl:


plugin/__init__.py
```diff
@@ -1,4 +1,5 @@
 from .core.css import css
+from .core.collections import DottedDict
```
rwols, line 2: Not in the correct alphabetic order? :)

rwols (approved):

PR #1457, opened 2020-11-16
prev/next diagnostic selects affected region

mitranim:
Ideally this would be configurable via command arguments, for example:

```jsonc
"args": {"select": null}     // suppress
"args": {"select": "region"} // select entire region
"args": {"select": "start"}  // select start of region
```

But for now, for simplicity, the behavior is to select the entire region. Let me know if this needs any changes.

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -496,6 +496,8 @@ class WindowManager(Manager):
                 phantom_set.update([phantom])
                 self._window.focus_view(view)
                 view.show_at_center(phantom.region)
+                # Must be a tuple. Passing a `sublime.Region` causes a Sublime error.
+                view.run_command("lsp_selection_set", {"regions": [phantom.region.to_tuple()]})
                 self._diagnostic_phantom_set = phantom_set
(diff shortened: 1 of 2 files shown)


rwols:
Looks like we didn't add `to_tuple` yet to stubs/sublime.pyi.

mitranim:
Seems to be passing now.

predragnikolic:
Question?
Would it be a better experience if the cursor moved to the _start_ or _end_ of the phantom (instead of the whole selection)?

The whole region is not selected, so if the user start typing, the selected text wont be deleted.


image

mitranim:
@predragnikolic Selecting the entire region actually allows the user to choose on the fly. Pressing ← takes you to the start, pressing → takes you to the end.

The downside is reaching for the arrows... but not if your layout uses ⌘hjkl for arrows, which I highly recommend to everyone. 😉

rwols (approved):
Many thanks!

rwols:
I think this could be nice as optional arguments to the commands:

```
"args": {"select": null}     // suppress
"args": {"select": "region"} // select entire region
"args": {"select": "start"}  // select start of region
```

though, I would let the `select` be optional, in which case it would mean "suppress".

PR #1459, opened 2020-11-16
Allow specifying the value type for Promise

rchl:
- Make Promise a generic type so that the Promise (return) value can be
  typed. Many of our uses of Promise don't really expect any return
  value from the Promise so those are left alone without specifying
  the value type.
- Renamed some arguments to better match the (ES) Promise spec.


rchl:
Apart from checking on actual code, I've been testing with those lines (checking that there is no error and types on hover are correct):

```py
p = Promise(lambda resolve: resolve('str'))  # type: Promise[str]
p = Promise(lambda resolve: resolve())  # type: Promise[str]
p.then(lambda value: 'str')
p.then(handle_then)
r = p.then(lambda value: 1)  # Promise[Literal[1]]
pa2 = Promise.all([Promise.resolve(1), Promise.resolve(2)])
```

I can't guarantee that all types are 100% correct but it's a complex beast this one.

plugin/core/open.py
```diff
@@ -28,7 +30,7 @@ def open_file(window: sublime.Window, file_path: str, flags: int = 0, group: int
             return value[0]
 
     # Prepare a new promise to be resolved by a future on_load event (see the event listener in main.py)
-    def fullfill(resolve: ResolveFunc) -> None:
+    def fullfill(resolve: ResolveFunc[sublime.View]) -> None:
```
rwols, line 33: This should be `ResolveFunc[Optional[sublime.View]]`

plugin/core/sessions.py
```diff
@@ -966,7 +967,7 @@ def run_command(self, command: Mapping[str, Any]) -> Promise:
             )
         )
 
-    def run_code_action_async(self, code_action: Mapping[str, Any]) -> Promise:
+    def run_code_action_async(self, code_action: CodeAction) -> Promise[CodeAction]:
```
rwols, line 970: The input should  be `Union[Command, CodeAction]`, the output should be `Promise[None]` I think?

plugin/hover.py
```diff
@@ -67,7 +67,7 @@ def run(self, edit: sublime.Edit, point: Optional[int] = None, event: Optional[d
             return
         self._base_dir = windows.lookup(window).get_project_path(self.view.file_name() or "")
         self._hover = None  # type: Optional[Any]
-        self._actions_by_config = {}  # type: Dict[str, List[CodeActionOrCommand]]
+        self._actions_by_config = {}  # type: Dict[str, List[CodeAction]]
```
rwols, line 70: The type is `Dict[str, List[Union[Command, CodeAction]]]` I think.

plugin/core/open.py
rchl: It might be clearer to do that but `ResolveFunc` already ensures that `None` can also be passed:

```py
class ResolveFunc(Protocol[T]):
    def __call__(self, value: Union[T, 'Promise[T]'] = None) -> None:
        ...
```

That's also why I've used `Protocol` for it as it's not possible to represent an optional argument with `Callable`.

You can see here that the types are correct.
Screenshot 2020-11-16 at 20 52 42

But I could try anyway to make it Optional explicitly.

plugin/core/sessions.py
rchl: True about the return value. In the end the CodeAction is not returned.

The other one I have to check exactly

plugin/core/typing.py
```diff
@@ -93,7 +97,8 @@ class Tuple(Type):  # type: ignore
     class Union(Type):  # type: ignore
         pass
 
-    Protocol = object  # type: ignore
+    class Protocol(Type):  # type: ignore
+        pass
```
rchl, line 101: Changed this to make the `Protocol[T]` syntax not crash.
Those are just fake types so it doesn't matter what are those as long as they don't crash in Python <3.8.

rchl:
I wonder if I'll be able to convince mypy to like this code...

plugin/core/promise.py
```diff
@@ -126,26 +134,26 @@ def recheck_resolve_status(_: Any) -> None:
...
+        self.callbacks = []  # type: List[ResolveFunc[T]]
+        executor_func(lambda value=None: self._do_resolve(value))
 
     def __repr__(self) -> str:
         if self.resolved:
             return 'Promise({})'.format(self.value)
         return 'Promise(<pending>)'
 
-    def then(self, callback: ThenFunc) -> 'Promise':
+    def then(self, onfullfilled: FullfillFunc[T, TResult]) -> 'Promise[TResult]':
```
rwols, line 156: This doesn't seem right. And this place is probably the hardest to get right because it's the monadic join operation.

1. When you pass a `Callable[[T], TResult]` in here, then the return type of this method is `Promise[TResult]`.
2. When you pass a `Callable[[T], Promise[TResult]` in here, then the return type is **also** `Promise[TResult]`.

Point (2) is not taken care of here I think.

rwols: Oh never mind, it is taken care of :)

But not really in a nice way. Maybe we can make use of `@overload` here?

rwols: Something like

```python
@overload
def then(self, f: Callable[[T], TResult]) -> Promise[TResult]: ...
@overload
def then(self, f: Callable[[T], Promise[TResult]]) -> Promise[TResult]: ...

def then(self, f: Callable[[T], Union[TResult, Promise[TResult], None]]) -> Promise:
    # the actual implementation
```

By the way, awesome that you're trying to tackle this!

rchl: `@overload` is new in 3.8, as far as I can see. So won't work here.

But what is not nice about it?

plugin/core/sessions.py
rchl: > The input should be `Union[Command, CodeAction]`

Well, this is a nightmare to get working. Can't convince pyright that it's `Command` even with `isinstance` check or explicit `# type` comment.

Screenshot 2020-11-16 at 21 29 37

plugin/core/promise.py
rwols: Looks like I'm wrong, everything works fine here :)

rchl: I admit types get quite crazy when trying to wrap this around your head. :)

plugin/core/sessions.py
rchl: I have to ignore this. Doesn't seem like pyright can infer the type of the parameter based on `isinstance` check of one of its properties.

plugin/core/types.py
```diff
@@ -15,6 +15,21 @@
...
+})
+
+CodeAction = TypedDict('CodeAction', {
+    'title': str,
+    'kind': Optional[str],
+    'diagnostics': Optional[List[Any]],
+    'isPreferred': Optional[bool],
+    'edit': Optional[dict],
+    'command': Optional[Command],
+})
```
rwols, line 31: The best place for the typed dicts are in plugin/core/protocol.py I think :)

plugin/core/sessions.py
rchl: Used `cast` to fix it.

plugin/hover.py
rchl: Fixed.

plugin/core/types.py
rchl: Moved

plugin/code_actions.py
```diff
@@ -178,7 +174,7 @@ def filtering_collector(
     """
 
     def actions_filter(actions: CodeActionsResponse) -> List[CodeActionOrCommand]:
-        return [a for a in (actions or []) if a.get('kind') in kinds]
+        return [a for a in (actions or []) if a.get('kind') in kinds]  # type: ignore
```
rchl, line 177: The `get('kind')` triggers type error as `Command` can't have `kind` attribute.

Even checking with `'kind' in a` doesn't help.

plugin/core/promise.py
```diff
@@ -1,14 +1,24 @@
...
+from .typing import Callable, Generic, List, Optional, Protocol, Tuple, TypeVar, Union
 import functools
 import sublime
 import threading
 
-ResolveFunc = Callable[..., None]  # Optional argument not supported in Callable so using "..."
-FullfillFunc = Callable[[ResolveFunc], None]
-ThenFunc = Callable[[Any], Any]
+T = TypeVar('T')
+T_contra = TypeVar('T_contra', contravariant=True)
```
rchl, line 7: This one is crazy and I don't fully get it but mypy complains that the type used in `ResolveFunc` is expected to be contravariant. I'm not sure why. Googling reveals that it might be a bug in mypy but I haven't found exactly this case in people's reports.

rwols: Functions are always contravariant in their input arguments from what I understand.

plugin/core/promise.py
```diff
@@ -1,14 +1,24 @@
...
+TResult = TypeVar('TResult')
 
 
-class Promise:
+class ResolveFunc(Protocol[T_contra]):
+    def __call__(self, value: Union[T_contra, 'Promise[T_contra]'] = None) -> None:
+        ...
+
+
+FullfillFunc = Callable[[T], Union[TResult, 'Promise[TResult]', None]]
```
rwols, line 16: Does this work correctly? As in does mypy now understand `FullfillFunc[Optional[T], TResult]` ? I have a feeling that it doesn't understand the `[Optional[T], TResult]` part.

rchl: I mean, mypy complains here:
```py
f = lambda value: value + 'sss'  # type: FullfillFunc[Optional[str], str]
```
```
plugin/core/promise.py:21: error: Unsupported left operand type for + ("None")
plugin/core/promise.py:21: note: Left operand is of type "Optional[str]"
```
Pyright doesn't complain instead.

plugin/core/promise.py
rchl: But then why something like `Callable[[T], None` doesn't complain about requiring contravariant type?

rwols (approved):

PR #1460, opened 2020-11-16
Fix CI error due to use of ::add-path::

rchl:


rchl:
The `macOS` and `Linux` jobs don't run obviously but I hope the changes there are correct also.

rwols (approved):

PR #1461, opened 2020-11-16
Make renaming more careful

### LINKED ISSUE #873: Could you add following command: "Show all occurrences and rename symbol"
Hi!

I would like to see all occurrences of symbol before renaming.
It would be nice to have this command.

Thanks!

rwols:
When a server returns text edits for textDocument/rename that span more than
one file, we'll now show a dialog asking the user to confirm that he/she really
wants to apply the changes. The user can also select "Dry Run" to show the
locations of the text edits that will be applied in an output panel.

Resolves #873


plugin/rename.py
```diff
@@ -126,3 +141,45 @@ def on_prepare_result(self, response: Any, pos: int) -> None:
...
+        panel = ensure_rename_panel(window)
+        if not panel:
+            return
+        text = ''
+        for file, file_changes in changes.items():
+            text += '◌ {}:\n'.format(self._get_relative_path(file))
+            for edit in file_changes:
+                start = edit[0]
+                end = edit[1]
+                text += '\t{:>8}:{:<4} - {}:{}\n'.format(start[0] + 1, start[1] + 1, end[0] + 1, end[1] + 1)
```
rchl, line 168: Showing only the starting point might be enough as it's looking a bit busy/confusing.

rchl (approved):
Cool stuff.

plugin/rename.py
```diff
@@ -105,7 +109,18 @@ def on_rename_result(self, response: Any) -> None:
...
+                if file_count > 1:
+                    total_changes = sum(map(len, changes.values()))
+                    message = "Replace {} occurrences across {} files?".format(total_changes, file_count)
+                    choice = sublime.yes_no_cancel_dialog(message, "Replace", "Dry Run")
+                    if choice == sublime.DIALOG_YES:
+                        apply_workspace_edit(window, changes)
+                    elif choice == sublime.DIALOG_NO:
+                        self._render_rename_panel(changes, total_changes, file_count)
+                else:
+                    window.run_command('lsp_apply_workspace_edit', {'changes': changes})
```
rchl, line 123: I think this doesn't work when file_count is 1.

rchl (approved):

PR #1462, opened 2020-11-17
fix: renaming input has no placeholder if no "renameProvider.prepareProvider"

jfcherng:
Trigger the "lsp_symbol_rename" command via a keybinding with LSP-pyright. The popup input has no placeholder, which is supposed to be the original symbol name. 

This PR introduced a guessed symbol name for the placeholder which is better than nothing I think.


rwols (approved):

rwols:
Thanks!

PR #1465, opened 2020-11-17
Add fortran language id

predragnikolic:
Fortran syntax has some selectors that would not convert to the right language id.

See:
https://lsp.readthedocs.io/en/latest/#fortran

This PR should fix it.

### DIFF
--- a/language-ids.sublime-settings
+++ b/language-ids.sublime-settings
@@ -23,10 +23,10 @@
     "source.c++": "cpp",
     "source.cs": "csharp",
     "source.dosbatch": "bat",
+    "source.fixedform-fortran": "fortran", // https://packagecontrol.io/packages/Fortran
     "source.js": "javascript",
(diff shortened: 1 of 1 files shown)


language-ids.sublime-settings
```diff
@@ -26,6 +26,8 @@
     "source.js": "javascript",
     "source.js.react": "javascriptreact", // https://github.com/Thom1729/Sublime-JS-Custom
     "source.jsx": "javascriptreact",
+    "source.fixedform-fortran": "fortran", // https://packagecontrol.io/packages/Fortran
```
rwols, line 29: Please keep this list sorted.

rwols (changes requested):

language-ids.sublime-settings
predragnikolic: fixed. :)

Was the `languageId` deprecated?

rwols (approved):

PR #1466, opened 2020-11-18
Sync typing.py with ST4

rchl:


plugin/core/typing.py
```diff
@@ -1,11 +1,13 @@
-try:
+import sys
 
-    # ST builds >= 4000
+if sys.version_info >= (3, 8, 0):
 
-    from mypy_extensions import TypedDict
     from typing import Any
     from typing import Callable
+    from typing import cast
```
rwols, line 7: Good to think ahead :)

rwols (approved):

PR #1468, opened 2020-11-18
Change MkDocs theme to material 

predragnikolic:
changed:
- changed MkDocs theme to material

added:
- github actions for building documentation on push

breaking change:
- readthedocs [don't support other themes](https://github.com/readthedocs/readthedocs.org/issues/978#issuecomment-285212453), so https://lsp.readthedocs.io will eventually get deprecated.

The new docs will be available at https://sublimelsp.github.io/LSP/


stylesheets/extra.css
```diff
@@ -0,0 +1,3 @@
+:root > * {
```
rchl, line 1: Just `:root` should be enough no? The variable should be accessible to all children.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,15 @@
+name: ci
```
rchl, line 1: could call it docs to make it clear what it does

stylesheets/extra.css
predragnikolic: I saw an example they wrote to [override the highlight forground](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#custom-syntax-theme), so I just copied/pasted the css selector `:root > *`.

I tried now only `:root` but it doesn't work.

rchl: Could be that it's the way to override the variable defined on the `:root`.

docs/README
```diff
@@ -0,0 +1,27 @@
+# Welcome to MkDocs
```
rwols, line 1: Can we add a .md extension to this file?

Also, explain here that the user must install mkdocs-material, otherwise this all doesn't work.

mkdocs.yml
```diff
@@ -1,5 +1,35 @@
...
+site_name: LSP for Sublime Text
+repo_url: https://github.com/sublimelsp/LSP/
+edit_uri: edit/st4000-exploration/docs/
+
+theme:
+  name: material
+  palette:
+      scheme: slate
+      accent: orange
+      primary: orange
```
rwols, line 10: Indentation is a bit odd here. Keep indentation to 2 spaces for yaml files.

docs/README
rchl: In fact, we should probably just replace it without our own test if it's gonna show up on opening the `docs/` folder on github.

predragnikolic: > Can we add a .md extension to this file?

Yes. And yes I will add the step for installing mkdocs-material. Thanks.  :)

@rchl  I don't have enough context to get what you wanted to say. Can you clarify `it` and `test`? 

> In fact, we should probably just replace **it** without our own **test** if it's gonna show up on opening the docs/ folder on github.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,15 @@
+name: docs
+on:
+  push:
+    branches:
+      - st4000-exploration
```
jfcherng, line 5: I guess the docs only needs a rebuild when something in `docs/` is pushed?
https://github.blog/changelog/2019-09-30-github-actions-event-filtering-updates/

predragnikolic: So if I write:
```
on:
  push:
    paths:
    - "docs/**"
```
this github actions will be triggered when a file in the docs folder is changed/added/deleted?

jfcherng: Yes. And the workflow file itself. https://github.com/jfcherng-sublime/ST-ASS/blob/0db07aac67609f0731374601a1557b7f4723c2d5/.github/workflows/syntax.yml#L3-L15

predragnikolic: Thanks :)
done with 2f2f369

jfcherng: You forget to add `.github/workflows/docs.yml`!

predragnikolic: done 57b67eb :)

jwortmann:
You could also replace the docs badge in the repo's main README file then.
I think it needs to be
```
[![Documentation](https://github.com/sublimelsp/LSP/workflows/docs/badge.svg?branch=st4000-exploration)](https://sublimelsp.github.io/LSP/)
```

predragnikolic:
@jwortmann I am planning to do a couple of follow up PRs. 

Because of that I would wait a bit longer before changing the link of the docs in the main README. :)

docs/README
rchl: Meant to say "text", not "test" but never mind me. I no longer believe what I've said mysefl.

rwols (approved):

PR #1478, opened 2020-11-22
Add an optional ClientConfig argument to filename_to_uri and uri_to_filename

rwols:
This ensures backwards compatibility for the public API, because filename_to_uri
and uri_to_filename are indeed part of the API.

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -2,9 +2,11 @@ from .collections import DottedDict
 from .logging import debug, set_debug_logging
 from .protocol import TextDocumentSyncKindNone
 from .typing import Any, Optional, List, Dict, Generator, Callable, Iterable, Union, Set, Tuple, TypeVar
-from .url import filename_to_uri
-from .url import uri_to_filename
 from threading import RLock
+from urllib.parse import urljoin
+from urllib.parse import urlparse
+from urllib.request import pathname2url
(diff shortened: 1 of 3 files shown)


plugin/core/types.py
```diff
@@ -683,10 +685,14 @@ def map_client_path_to_server_uri(self, path: str) -> str:
                 path, mapped = path_map.map_from_local_to_remote(path)
                 if mapped:
                     break
-        return filename_to_uri(path)
+        return urljoin('file:', pathname2url(path))
```
rchl, line 688: I don't understand why we are duplicating the code here if we have to keep original functions anyway. There is a chance for implementations to get out of sync that way.

rwols: Because filename_to_uri and uri_to_filename would ideally be removed, but we cannot do that because they're part of the public api.

Once this is merged I'll go ahead and change all the functions in plugin/core/views.py to take a ClientConfig, and those functions will then call filename_to_uri and uri_to_filename with an actual ClientConfig. At the same time LSP-* packages have some time to upgrade.

rchl: I think those could already call filename_to_uri and uri_to_filename but I trust you will clean it up later. :)

rchl (approved):

PR #1482, opened 2020-11-22
Add AbstractPlugin.on_settings_changed API

rchl:
This replaces the old "on_workspace_did_change_configuration" with
something more generic that will be called whenever client settings
change (currently those cannot really change during the lifetime of the
Session but might in the future).

The Plugin instance receives a reference to the object that lives as
long as the session and is also **the** configuration that is sent to
the server in other cases like when responding to
"workspace/configuration".

Relates to https://github.com/sublimelsp/LSP-pyright/issues/30

### DIFF
--- a/plugin/core/collections.py
+++ b/plugin/core/collections.py
@@ -138,7 +138,7 @@ class DottedDict:
 
         :param      variables:  The variables
 
-        :returns:   A copy of underlying dictionary, but with the variables replaced
+        :returns:   A copy of the underlying dictionary, but with the variables replaced
         """
         return sublime.expand_variables(self._d, variables)
 
--- a/plugin/core/collections.py
+++ b/plugin/core/collections.py
@@ -132,15 +132,15 @@ class DottedDict:
             else:
                 self.set(key, value)
 
-    def create_resolved(self, variables: Dict[str, str]) -> "DottedDict":
+    def get_resolved(self, variables: Dict[str, str]) -> Dict[str, Any]:
         """
-        Resolve a DottedDict that may potentially contain template variables like $folder
+        Resolve a DottedDict that may potentially contain template variables like $folder.
 
         :param      variables:  The variables
 
-        :returns:   A copy of this DottedDict, but with the variables replaced
+        :returns:   A copy of underlying dictionary, but with the variables replaced
         """
-        return DottedDict(sublime.expand_variables(self._d, variables))
+        return sublime.expand_variables(self.copy(), variables)
 
     def _update_recursive(self, current: Dict[str, Any], prefix: str) -> None:
         if not current:
--- a/plugin/core/collections.py
+++ b/plugin/core/collections.py
@@ -140,7 +140,7 @@ class DottedDict:
 
         :returns:   A copy of underlying dictionary, but with the variables replaced
         """
-        return sublime.expand_variables(self.copy(), variables)
+        return sublime.expand_variables(self._d, variables)
 
     def _update_recursive(self, current: Dict[str, Any], prefix: str) -> None:
         if not current:
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -515,11 +515,12 @@ class AbstractPlugin(metaclass=ABCMeta):
         """
         self.weaksession = weaksession
 
-    def on_workspace_did_change_configuration(self, settings: DottedDict) -> None:
+    def on_settings_changed(self, settings: DottedDict) -> None:
         """
-        Override this method to alter the server settings for the workspace/didChangeConfiguration notification.
+        Override this method to alter the settings that are returned to the server for the
+        workspace/didChangeConfiguration notification and the workspace/configuration requests.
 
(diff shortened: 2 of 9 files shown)


rchl:
I see there is an issue with settings being re-used when a new session is started. That means we would potentially end up with plugin-changes stacking up on creating new sessions.

Either the settings object needs to be created from scratch for each new session or `on_settings_changed` needs to be called only when settings actually change.

I think it would have to be the former solution.

rchl:
Toying with the idea of creating a config copy but not sure it's safe yet.

rchl:
And also, we should discuss whether the new `ResolvedStartupConfig` is really necessary, or should we remove it before it spreads?

Do plugins really need the "resolved" config state or it would be enough to just pass a copy of `ClientConfig` to them? Are there any existing use-cases that need "resolved" state? Reusing `ClientConfig` would be simpler and better for ST3-compatibility.

rwols:
> And also, we should discuss whether the new ResolvedStartupConfig is really necessary, or should we remove it before it spreads?

Yeah, I'm okay with removing it. We make a copy in `start_async` now anyway.

plugin/core/windows.py
```diff
@@ -254,6 +254,7 @@ def _needed_config(self, view: sublime.View) -> Optional[ClientConfig]:
         return None
 
     def start_async(self, config: ClientConfig, initiating_view: sublime.View) -> None:
+        config = ClientConfig.from_config(config, {})
```
rwols, line 257: With this, I guess the line

```
resolved = config.resolve(variables)
```

can now become

```
config.resolve(variables)
```

where it mutates itself?

rchl: I'm thinking that even the copied config should always stay unresolved and we would only resolve (and by that I mean only expanding the settings) at the time of sending them to the server. So plugins would always receive "unresolved" values (as is the case for the `_plugin.on_workspace_configuration` call already, for example.

The part of `resolve` that resolves settings for the transport would be a separate function or something.

rwols:
Should `on_workspace_configuration` also get a `DottedDict`?

rchl:
> Should `on_workspace_configuration` also get a `DottedDict`?

I suppose it could...
Even though it would be a breaking change type-wise, I would think it shouldn't break anything in practice?

plugin/core/sessions.py
```diff
@@ -271,7 +271,7 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
         "rootPath": first_folder.path if first_folder else None,
         "workspaceFolders": [folder.to_lsp() for folder in workspace_folders] if workspace_folders else None,
         "capabilities": capabilities,
-        "initializationOptions": config.init_options.create_resolved(variables)
+        "initializationOptions": config.init_options.create_resolved(variables).get()
```
rwols, line 274: Now that we *always* resolve these DottedDicts after possible plugin intervention, maybe we can call the method `resolve` and it'll return a `Dict[str, Any]` ?

rchl:
> > Should `on_workspace_configuration` also get a `DottedDict`?
> 
> I suppose it could...
> Even though it would be a breaking change type-wise, I would think it shouldn't break anything in practice?

The more I think about it the more I doubt.

We could pass DottedDict but:
 - small or big, it's still a breaking change
 - plugin can create a DottedDict manually if it really wants to (for the purpose of simplifying modifications)

plugin/core/sessions.py
rchl: Would it resolve a copy of the Dict or the internal dict (`self._d`). Because current implementation actually resolves internal dict and I don't think that was intended.

rchl: Changed create_resolved => get_resolved because I think `resolve` is a poor name due to being hard to search for and also sounding like it's gonna change the internal state.

rchl: Oh, I think `sublime.expand_variables` will modify the passed-in object but it doesn't so I don't have to make a copy and the existing behavior was already correct.

plugin/core/collections.py
```diff
@@ -132,15 +132,15 @@ def update(self, d: Dict[str, Any]) -> None:
...
-        Resolve a DottedDict that may potentially contain template variables like $folder
+        Resolve a DottedDict that may potentially contain template variables like $folder.
 
         :param      variables:  The variables
 
-        :returns:   A copy of this DottedDict, but with the variables replaced
+        :returns:   A copy of underlying dictionary, but with the variables replaced
         """
-        return DottedDict(sublime.expand_variables(self._d, variables))
+        return sublime.expand_variables(self.copy(), variables)
```
rwols, line 143: There is no need to make a copy. `sublime.expand_variables` already does that

```
>>> a={"hello": "$world"}
>>> b=sublime.expand_variables(a, {"world": "foobar"})
>>> a==b
False
>>> a
{'hello': '$world'}
>>> b
{'hello': 'foobar'}
```

rchl: Yes, was just testing it too. Fixed.

rwols (approved):

PR #1504, opened 2020-12-02
Add Plugin API for overriding requests and notifications

rchl:
I've created this low-level API as it's difficult to otherwise create tailored and yet generic API that is not just to be used for a specific case.

LSP-json would utilize it something like this:

```py
RE_JSONC = re.compile(r'(?:\/package\.json)$')


class LspJSONPlugin(NpmClientHandler, StoreListener):

    def on_pre_send_notification_async(self, notification: Notification) -> None:
        if notification.method == 'textDocument/didOpen':
            text_document = notification.params['textDocument']
            if RE_JSONC.search(text_document['uri']):
                text_document['languageId'] = 'jsonc'
```

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -544,6 +544,25 @@ class AbstractPlugin(metaclass=ABCMeta):
         """
         return False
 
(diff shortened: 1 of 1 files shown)


rwols (approved):
I dig this.

PR #1505, opened 2020-12-02
Fix: server capability is codeAction.resolveProvider

rwols:
not codeAction.resolveSupport.


rchl (approved):
Oh my.

rchl:
Seems like a bug in the spec:
Screenshot 2020-12-02 at 22 31 16

rwols:
Yes. Around the actual textDocument/codeAction request, the server property is codeAction.resolveProvider.

PR #1507, opened 2020-12-02
Fix: document highlights cannot be sent earlier than didChange

rwols:
The text changes are buffered and sent after a 500ms timeout, provided that
nothing interesting happens like a trigger character.

The highlights are now sent after a 300ms timeout. This can cause the request
params to contain a document position that doesn't exist from the viewpoint of
the server, because those changes did not yet arrive (they will arrive another
200ms later).

This can be fixed by setting the timeout for highlights to 500ms.

Found in general rust-analyzer debugging.

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -144,7 +144,7 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
     ACTIVE_DIAGNOSTIC = "lsp_active_diagnostic"
     code_actions_debounce_time = 800
     color_boxes_debounce_time = 500
-    highlights_debounce_time = 300
+    highlights_debounce_time = 500
 
     @classmethod
     def applies_to_primary_view_only(cls) -> bool:
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -16,6 +16,7 @@ from .core.signature_help import create_signature_help
 from .core.signature_help import SignatureHelp
 from .core.types import basescope2languageid
 from .core.types import debounced
+from .core.types import FEATURES_TIMEOUT
 from .core.typing import Any, Callable, Optional, Dict, Generator, Iterable, List, Tuple, Union
 from .core.views import DIAGNOSTIC_SEVERITY
 from .core.views import document_color_params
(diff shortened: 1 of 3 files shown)


predragnikolic:
The only downside I see with a 500ms timeout is that the highlight feels slow(this is probably a subjective oppinion)

What if we reduced the timeout for buffered text changes to 300? Is that an option to consider?

That way technically the highlight will still feel snappier.
It would potentially lead to a chattier communication between the server and the client, but it also doesn't mean necessarly worse performance.

Not that I am agains increasing the highlight timeout, I would just like the plugin to feel fast.

plugin/documents.py
```diff
@@ -144,7 +144,7 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
     ACTIVE_DIAGNOSTIC = "lsp_active_diagnostic"
     code_actions_debounce_time = 800
     color_boxes_debounce_time = 500
-    highlights_debounce_time = 300
+    highlights_debounce_time = 500
```
rchl, line 147: I'm wondering if we should take the opportunity to clean it up and just use one timeout for all (and potentially make it customizable).

I do think that VSCode feels snappier in showing diagnostics and code actions, probably because it has a lower timeout so I would want to try with lower values also. But that can be considered a separate request. For now maybe just use the same timeout for both those two and also the text change timeout so that those stay in sync?

rchl:
Just something I've noticed now when looking at the code. We never check `should_notify_did_change`. We should either start using it or remove it. For example:

```diff
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -199,10 +199,8 @@ class SessionBuffer:
     on_reload_async = on_revert_async
 
     def purge_changes_async(self, view: sublime.View) -> None:
-        if self.pending_changes is not None:
+        if self.pending_changes is not None and self.should_notify_did_change():
             sync_kind = self.text_sync_kind()
-            if sync_kind == TextDocumentSyncKindNone:
-                return
             if sync_kind == TextDocumentSyncKindFull:
                 changes = None
                 version = view.change_count()
```

rwols:
I'll just set all the timeouts to 300ms.

rchl (approved):

PR #1511, opened 2020-12-04
Fix: do not clobber older edits with newer edits

rwols:
When parsing documentEdits there's no mention in the protocol that
the textDocument key is unique.


rwols:
I’m unsure what consequences this has for the code-actions-on-save functionality, but the tests pass.

PR #1512, opened 2020-12-06
Add support for $document_id in settings

### LINKED ISSUE #1499: Recursive variable expansion in lsp_execute command

The documentation states ...

```js
[
  // ...
  {
    "caption": "Thread First",
    "command": "lsp_execute",
    "args": {
      "command_name": "thread-first",
      "command_args": ["${file_uri}", 0, 0]
    }
  }
]
```

Note: `command_args` is optional depending on the `workspace/executeCommand` that are supported by the LSP server.
The following **variables will be expanded, but only if they are top-level array items** and not within nested arrays or objects:

The **LemMinX** language server provides a validation command which expects [textDocumentIdentifie](https://microsoft.github.io/language-server-protocol/specification#textDocumentIdentifier) as first parameter.

see: https://github.com/eclipse/lemminx/pull/938

The proper command definition in ST would look like

```
[
	{
		"caption": "XML: Validate File",
		"command": "lsp_execute",
		"args": {
			"command_name": "xml.validation.current.file",
			"command_args": [{"uri": "${file_uri}"}]
		}
	}
]
```

Unfortunatelly `${file_uri}` is not expanded as it is not in the top-level array.

The most flexible and straight forward solution would probably be to support recursive variable expansions in all nested arrays and objects.

An `$document_id` variable which is expanded to `{"uri": "file:///path/to/file.xml"}` would do the job as well. The command definition would look as follows then.

```
[
	{
		"caption": "XML: Validate File",
		"command": "lsp_execute",
		"args": {
			"command_name": "xml.validation.current.file",
			"command_args": ["$document_id"]
		}
	}
]
```

deathaxe:
Fixes #1499

### DIFF
--- a/docs/src/features.md
+++ b/docs/src/features.md
@@ -51,6 +51,7 @@ The following variables will be expanded, but only if they are top-level array i
 
 | Variable | Type | Description |
 | -------- | ---- | ----------- |
+| `"$document_id"` or `"${document_id}"` | object | Mapping `{ 'uri': string }` containing the file URI of the active view, see [Document Identifier](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentIdentifier) |
 | `"$file_uri"` or `"${file_uri}"` | string | File URI of the active view |
 | `"$selection"` or `"${selection}"` | string | Content of the (topmost) selection |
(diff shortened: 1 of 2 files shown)


rwols (approved):
Awesome

PR #1513, opened 2020-12-06
Rework textDocument/signatureHelp

### LINKED ISSUE #1001: The signature help "parentheses" detection needs more work
Screenshot from 2020-04-29 23-05-29

We've seen that for e.g. a dockerfile language server the parens detection doesn't work at all.
For gopls, parentheses are used after the function ends. The right decision might be to abandon this parsing. There's nothing in the spec that guarantees us parentheses of any kind.

rwols:
- Proper handling of markdown parameter docs.
- Don't assume that function signatures have an `(` and `)`.
- Put the *active parameter docs* at the top, font-size 1rem.
- Put the *signature docs* under that, font-size 0.9rem.
- Use 0.9rem font-size for the intro message.
- KISS.

close #1001

### DIFF
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -10,7 +10,7 @@ TextDocumentSyncKindFull = 1
 TextDocumentSyncKindIncremental = 2
 
 
-class DiagnosticSeverity(object):
+class DiagnosticSeverity:
     Error = 1
     Warning = 2
     Information = 3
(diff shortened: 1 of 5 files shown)


rchl:
When I've triggered signature help manually and gotten this response from the server:

```json
{
  "activeParameter": null,
  "activeSignature": null,
  "signatures": []
}
```

it opened a very tiny, empty popup.

rchl (approved):
It seems to work as well as before on few basic tests so I'll not dig too much into the code :)

Are you planning some further improvements to signature help in the near future? I think it works poorly in the sense that it disappears way too easily on writing the code or moving the cursor around. So if you are not planning work then I might look into it eventually.

rwols:
```
{
  "activeParameter": null,
  "activeSignature": null,
  "signatures": []
}
```

this is actually invalid. "activeParameter" and "activeSignature" are supposed to be an integer. The spec also states that if there are no signatures the server should return `null`. But we have to protect against these kinds of malformed responses anyway I suppose :)

rwols:
> Are you planning some further improvements to signature help in the near future?

No. Feel free to improve it. Though I feel that you'd be swimming against the ST API stream. There is no natural way to keep a popup open when typing.

predragnikolic:
> is no natural way to keep a popup open when typing.

Could `COOPERATE_WITH_AUTO_COMPLETE` help?

PR #1514, opened 2020-12-06
on_{reload,revert}_async is also invoked for the ViewEventListener

### LINKED ISSUE #1470: Plugin/client sends duplicate didChange requests
**Describe the bug**

The plugin (acting as LSP client) can sometimes send duplicate `didChange` request.

See https://github.com/hashicorp/terraform-ls/issues/310 for more, including snippet of the log (which contains whole request).

**Expected behavior**

Avoid sending duplicate request.

- OS: macOS 10.15.7
- Sublime Text version: 4093
- LSP version: 1.1.0
- Language servers used: terraform

**Additional context**

I am aware that the log may not provide sufficient details to debug this and you might also say that server should just ignore it - which would be all fair responses. I just thought I'd raise this as an issue, so you are aware, despite being unable to provide much details. So feel free to close this issue if you find it hard to fix or don't want to fix it, I will have no hard feelings about it! 😄

rwols:
This may or may not fix #1470. At least partially.

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -97,14 +97,6 @@ class TextChangeListener(sublime_plugin.TextChangeListener):
 
         sublime.set_timeout_async(notify)
 
-    def on_reload_async(self) -> None:
-        for listener in list(self.view_listeners):
-            listener.on_reload_async()
-
-    def on_revert_async(self) -> None:
-        for listener in list(self.view_listeners):
-            listener.on_revert_async()
-
     def __repr__(self) -> str:
         return "TextChangeListener({})".format(self.buffer.buffer_id)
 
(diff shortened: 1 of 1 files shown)


plugin/documents.py
```diff
@@ -97,14 +97,6 @@ def notify() -> None:
 
         sublime.set_timeout_async(notify)
 
-    def on_reload_async(self) -> None:
-        for listener in list(self.view_listeners):
-            listener.on_reload_async()
-
-    def on_revert_async(self) -> None:
-        for listener in list(self.view_listeners):
-            listener.on_revert_async()
```
rchl, line 106: The alternative would be to keep calling here and instead rename the methods in `DocumentSyncListener`. I think those two options would have slightly different behavior but I'm not sure what's the most correct way.

One thing that `TextChangeListener` does that `DocumentSyncListener` doesn't is it checks if the view is a `is_regular_view`.

rchl (approved):

Issue #1171, opened 2020-06-28
Blank area in popups


guest73:
* OS and language server
Ubuntu Mate, typescript-language-server and css-languageserver
* How you installed LSP (Package Control or from git?)
Package control
* Minimal reproduction steps
In .ts files: Hover over some function or class a blank square is shown where the function signature should be shown.
In .ts files: press Ctrl+Alt+Space - a blank square with some odd line is shown, no signature.
In .scss file: Hover over :last-child selector, a blank square is shown with a link to selector specification beneath.

![blank_area](https://user-images.githubusercontent.com/4907799/85957961-7b1fb880-b99a-11ea-8bf2-48c65cdeeb16.gif)

rchl:
Does anything show up in Sublime console?

I'd guess it has something to do with either color scheme or mdpopups (maybe local overrides?).

It's hard to help without asking 100 of questions so it would be best if you just [try a clean install](http://www.sublimetext.com/docs/3/revert.html) and see if you can reproduce.

predragnikolic:
What is your ST UI theme and color scheme?

guest73:
Well, tried with clean install. Unfortunately it did not help. Among installed plugins I had only 3
![Screenshot at 2020-06-29 17-15-13](https://user-images.githubusercontent.com/4907799/86016975-27f64600-ba2c-11ea-90ab-296bae25461f.png)
Everything was installed from package control. And, of course, I ran "Satisfy dependencies" to install mdpopups.

Here is a piece of LSP config I used

![Screenshot at 2020-06-29 17-18-21](https://user-images.githubusercontent.com/4907799/86017335-9935f900-ba2c-11ea-8327-b3de2b6964d7.png)


Also tried with different pre-installed ST themes and color schemas, did not help either, still having this blank area ;(

rchl:
I can see the issue in ST3 actually.
(We here are on the bleeding edge ST4 builds normally so tend to miss issues with ST3. :)

guest73:
Also, forgot to mention, nothing special in the console. No errors, or warnings.
And, I can see some blocks in some of the popups.

![blocks](https://user-images.githubusercontent.com/4907799/86018761-74db1c00-ba2e-11ea-8afb-3e72750b451f.gif)

So far, thanks for the response guys.

rchl:
Actually I can only reproduce similar thing when not having Typescript syntax installed so I've probably seen something else.

[...]

PR #1515, opened 2020-12-07
Use the system font for header elements in popups

rwols:


popups.css
```diff
@@ -13,6 +13,12 @@
 .lsp_popup li {
     font-family: system;
 }
+.lsp_popup h1,
+.lsp_popup h2,
+.lsp_popup h3,
+.lsp_popup h4 {
```
rchl, line 19: You could combine those rules to avoid duplication.

rchl (commented):
What kind of content uses header elements and should we really apply those to all headers, even ones that come from a server? Or should we even just apply that font style to the whole popup? At this point, I'm not sure what doesn't use it since the `p` element does which is probably used in a lot of places.

popups.css
rwols: I'm not very good at CSS. I think I combined the selectors in my latest commit?

rchl: Perfect.

rwols:
> What kind of content uses header elements and should we really apply those to all headers, even ones that come from a server?

\<h1\> is used a lot by clangd.

> Or should we even just apply that font style to the whole popup?

I would say *yes*, but I don't know how to do that in a neat way, because I suck at css.

rchl:
> I would say _yes_, but I don't know how to do that in a neat way, because I suck at css.

So we could just have a `font-family: system;` in the top `.lsp_popup` selector. That will be inherited by all children unless something overrides `font-family` (or `font`, if ST supports that) to something else. Also, ST's internal styles might override it if those specify `font-family` for certain elements (but I don't know if they do).

But we can just have `font-family` on the top selector and see if it doesn't apply to something and then make a specific override for that.

rwols:
Hmm... this also styles the diagnostics with the system font. I don't think that's an improvement.

rchl (approved):
It feels a bit random to me in which font-family style we style what. We target certain element types rather than certain content types.

But I guess it should be fine, we can think of a more generic solution later.

rwols:
I managed to simplify it I think :) or can something go wrong with this?

popups.css
```diff
@@ -1,52 +1,47 @@
...
 }
-.lsp_popup p {
+.diagnostics {
     margin-bottom: 0.5rem;
-    padding: 0 0.5rem;
-    font-family: system;
-}
-.lsp_popup li {
-    font-family: system;
+    font-family: monospace;
```
rwols, line 11: I'm not sure how to "reset" the font-family here. The value `monospace` doesn't actually do what it is supposed to do: https://github.com/sublimelsp/LSP/issues/1171#issuecomment-651800108

rchl: You are referring to a deleted line?
Not sure what do you exactly mean since the code doesn't look like that anymore.

rwols: I wrote `font-family: monospace;` here, but that doesn't actually work as you'd expect. It just so happens to work. But the behavior is unconfirmed on Windows.

The only special font-family is presumably `system`.

rchl (approved):

PR #1517, opened 2020-12-07
Always show the actions at the bottom of the hover popup

rwols:
I have a feeling some users miss out on these useful links. It doesn't hurt to
always show these links.

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -108,9 +108,6 @@
   // using a keyboard shortcut or the context menu.
   "show_code_actions": "annotation",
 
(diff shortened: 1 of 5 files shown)


jwortmann:
Why not just set the default setting to `true` then? There have been requests in the past to hide these links and I made a PR to introduce this setting some time ago. I'm one of the users who don't like to show these links for each small hover popup, and I remember that they didn't even work half of the time. I mostly use the hover popups to see descriptions and arguments of built-in functions, and I don't need a "Goto definition" or "Code actions" link in that case.

rchl (approved):

PR #1519, opened 2020-12-09
Display the diagnostic code

### LINKED ISSUE #1173: The user needs a way to know the name of the diagnostic being shown
Take a diagnostic like this for example:

```text
 ◌ src/main.c:
      159:30  	clang-tidy  	warning   	Narrowing conversion from 'double' to 'int'
```

Many linters/static-analyzers provide a way to suppress a specific issue on a specific line by having the user write a special comment, e.g. [in clang-tidy's case](https://clang.llvm.org/extra/clang-tidy/#suppressing-undesired-diagnostics):

`// NOLINT(???)`

But what should the LSP user write where `???` is? We're out of luck, because the diagnostic as shown doesn't tell us the name of the specific issue.

That name **is** already present inside the `textDocument/publishDiagnostics` message from the language server though, it's [`Diagnostic.code`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#diagnostic). In this example, that has the value `"bugprone-narrowing-conversions"`.

---

So if the diagnostic was instead shown including something like this:

`Narrowing conversion from 'double' to 'int' [bugprone-narrowing-conversions]`

...then the user would have the information they need to be able to write the comment:

`// NOLINT(bugprone-narrowing-conversions)`

---

I think some language servers duplicate the value of `Diagnostic.code` inside the `Diagnostic.message` string. That could be checked for to avoid showing the same information twice.

(These diagnostic names used by clangd+clang-tidy are quite wordy. The equivalent in the pyls+flake8 world are diagnostic names like `E731` and comments like `# noqa: E731`)

### LINKED ISSUE #1367: Implement codeDescriptionSupport
Version 3.16 adds a codeDescription to the Diagnostic structure. The codeDescription contains a url to "more information".

Advertise codeDescriptionSupport, and handle the codeDescription by allowing the user to click on the url.

N/A

**Additional context**
https://github.com/microsoft/language-server-protocol/commit/1de6e8e8d4f69ab7079f2626637de7c583c0822a

rwols:
Both in the hover popup, as well as in the diagnostic panel.
The hover popup additionally creates a clickable link if there exists a
codeDescription field with a URL.

I have also freshened up the diagnostics panel. It looks more like
SublimeLinter's panel now.

Resolves #1173
Resolves #1367

Impressions:

Schermafbeelding 2020-12-09 om 21 40 58

Schermafbeelding 2020-12-09 om 22 02 31

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -546,16 +546,16 @@ def format_diagnostic_for_panel(diagnostic: Diagnostic) -> str:
         code = make_link(diagnostic.code_description["href"], diagnostic.code)  # type: Optional[str]
     else:
         code = str(diagnostic.code) if diagnostic.code else None
-    formatted = ["[", diagnostic.source if diagnostic.source else "unknown-source"]
+    formatted = [diagnostic.source if diagnostic.source else "unknown-source"]
     if code:
         formatted.extend((":", code))
-    formatted.append("]")
-    return "{:>4}:{:<4}{:<8}{} {}".format(
+    # \u200B is the zero-width space
+    return "{:>4}:{:<4}{:<8}{} \u200B{}".format(
         diagnostic.range.start.row + 1,
         diagnostic.range.start.col + 1,
         format_severity(diagnostic.severity),
-        "".join(formatted),
-        (diagnostic.message.splitlines() or [""])[0]
+        (diagnostic.message.splitlines() or [""])[0],
+        "".join(formatted)
     )
 
 
@@ -575,23 +575,18 @@ def _with_scope_color(view: sublime.View, text: Any, scope: str) -> str:
 
 
 def format_diagnostic_for_html(view: sublime.View, diagnostic: Diagnostic, base_dir: Optional[str] = None) -> str:
-    formatted = ['<pre class="', DIAGNOSTIC_SEVERITY[diagnostic.severity - 1][1], '">']  # type: List[Any]
+    formatted = ['<pre class="', DIAGNOSTIC_SEVERITY[diagnostic.severity - 1][1], '">', text2html(diagnostic.message)]
     if diagnostic.code_description:
         code = make_link(diagnostic.code_description["href"], diagnostic.code)  # type: Optional[str]
     else:
-        code = _with_scope_color(view, diagnostic.code, "constant.numeric.code.lsp") if diagnostic.code else None
+        code = _with_scope_color(view, diagnostic.code, "comment.line.code.lsp") if diagnostic.code else None
     source = diagnostic.source if diagnostic.source else "unknown-source"
     formatted.extend((
-        _with_scope_color(view, "[", "punctuation.section.brackets.begin.lsp"),
-        _with_scope_color(view, source, "comment.line.lsp")
+        " ",
+        _with_scope_color(view, source, "comment.line.source.lsp")
     ))
     if code:
         formatted.extend((_with_scope_color(view, ":", "punctuation.separator.lsp"), code))
-    formatted.extend((
-        _with_scope_color(view, "]", "punctuation.section.brackets.end.lsp"),
-        " ",
-        text2html(diagnostic.message)
-    ))
     if diagnostic.related_info:
         formatted.append('<pre class="related_info">')
         formatted.extend(_format_diagnostic_related_info(info, base_dir) for info in diagnostic.related_info)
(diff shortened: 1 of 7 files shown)


rchl:
The knee-jerk reaction is that it's quite noisy and distracts from the error text which is more useful. Could we put it after the error text instead?

rchl:
In VSCode it is quite muted
Screenshot 2020-12-09 at 22 15 03

rwols:
Made it work in the .sublime-syntax with a zero-width space unicode character somehow.

Schermafbeelding 2020-12-09 om 22 58 26
Schermafbeelding 2020-12-09 om 22 58 51

rchl:
Looks very nice now IMO:
Screenshot 2020-12-09 at 23 00 03

But in the diagnostics panel the URL should probably be a plain URL:
Screenshot 2020-12-09 at 22 59 36

I think there is a special scope for URLs to make them stand out also.

rwols:
I decided not to display URLs in the panel, is that OK?

plugin/core/views.py
```diff
@@ -479,8 +479,9 @@ def text2html(content: str) -> str:
     return re.sub(REPLACEMENT_RE, _replace_match, content)
 
 
-def make_link(href: str, text: str, class_name: Optional[str] = None) -> str:
-    text = text.replace(' ', '&nbsp;')
+def make_link(href: str, text: Any, class_name: Optional[str] = None) -> str:
```
predragnikolic, line 482: I think it is safe to typehint `text` as `str` instead of `Any`,
or declare a type `DocumentUri = str`. What do you think?

predragnikolic:
I like that the diagnostic panel is now more compact. :+1: 

Here are two issues I noticed. (one is color-scheme specific, but still worth mentioning)

1. symbol_action_links are not rendered:

ST-4000 branch:

Screenshot from 2020-12-09 23-29-34

This branch:

Screenshot from 2020-12-09 23-28-57

2. Contrast might be a problem
The contrast in the popup between the red and gray might be a problem in some color schemes. 
This is a color scheme I made, so it is not a problem for me to tweak it, but it might potentially cause issues in other colors schemes.

rwols:
> This branch:

I could not repro this at that exact location, but will try a few more things tomorrow.

> Contrast might be a problem

I'm not good at CSS :) maybe someone can improve this?

jwortmann:
What you can try for this color would be instead of the foreground for `"comment"` (which could also be other colors than grey - e.g. green in some color schemes), use the global "foreground" color with a bit of transparency applied. Like this:
```diff
- code = _with_scope_color(view, diagnostic.code, "comment.line.code.lsp") if diagnostic.code else None
+ code = '<span style="color: color(var(--foreground) alpha(0.7));">{}</span>'.format(diagnostic.code) if diagnostic.code else None
```
Possibly adjust the alpha value until it looks good in most schemes.

plugin/core/views.py
rwols: The `diagnostic.code` is also passed into make_link, which might be an `int` (or `None`). But really, any stringifiable type works since it's passed into `.format`.

predragnikolic:
the contrast is great :+1: 
image

rchl:
I'm not sure I like that only the first line of the message is shown in the panel. I would think that the panel should show all information available.

Before:

Screenshot 2020-12-10 at 20 15 52


After:

Screenshot 2020-12-10 at 20 16 32

rchl:
And also I'm not sure why skipping the diagnostic link in the panel. We could add it in parenthesis or something?

The Diagnostics panel is all about diagnostics and it shows less information than the popup? :)

rwols:
> And also I'm not sure why skipping the diagnostic link in the panel. We could add it in parenthesis or something?

It's not clickable so I don't really see the point. Maybe possible with inline-phantoms though. Or we could turn the diagnostics panel into a HTML sheet. Can we defer this to another PR?

rchl:
> > And also I'm not sure why skipping the diagnostic link in the panel. We could add it in parenthesis or something?
> 
> It's not clickable so I don't really see the point. Maybe possible with inline-phantoms though. Or we could turn the diagnostics panel into a HTML sheet. Can we defer this to another PR?

Not clickable but with the right scope (or maybe even without) ST provides "open this link in the context menu

rchl (approved):

rwols:
> Not clickable but with the right scope (or maybe even without) ST provides "open this link" in the context menu

Really? I was not aware of that :) let me try a few things then

rwols:
On second thought, having a large url in the panel might be too cumbersome. I think an inline phantom might work best. But that would be a bit of work.

PR #1520, opened 2020-12-10
Clickable links in the diagnostics panel

rwols:
Untested, but I'll just put it on review while I try to set up eslint :)

### DIFF
--- a/Syntaxes/Diagnostics.sublime-syntax
+++ b/Syntaxes/Diagnostics.sublime-syntax
@@ -11,7 +11,7 @@ contexts:
     - include: line
 
   file:
-    - match: ^(.*)(:)$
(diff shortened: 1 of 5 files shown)


rwols:
image

rchl (approved):

rchl:
I've noticed something unrelated to this change that seems wrong.

I have three errors, 2 from typescript and 1 from eslint:

Screenshot 2020-12-11 at 21 55 09

Once I fix the eslint one, the panel closes automatically.
It could be that some code thinks that there are no more diagnostics messages left and closes it.

rwols:
> It could be that some code thinks that there are no more diagnostics messages left and closes it.

Yeah, I made a mistake there.

PR #1522, opened 2020-12-11
fix: diagnostics line regex

predragnikolic:
This PR will make sure the last line `1018:1   error   Invalid type "sublime.TextChange.str" mypy`
is also matched with the line regex. (so the f4 and shift+f4 will work for those lines).

```
plugin/documents.py:
   3:2   error   Cannot find module named '.completion' mypy
   4:2   error   Cannot find module named '.core.css' mypy
   ...
1018:1   error   Invalid type "sublime.TextChange.str" mypy
```

image


rwols (approved):
Nice work.

PR #1523, opened 2020-12-12
improve the popup ui

predragnikolic:
pyls/LSP-json:
before:
Screenshot 2020-12-12 at 00 21 54

after:
Screenshot 2020-12-12 at 00 19 55

LSP-typescript/LSP-pyright:
before:
Screenshot 2020-12-12 at 00 22 49

after:
Screenshot 2020-12-12 at 00 21 00

Feel free to test this with other LS :)

### DIFF
--- a/popups.css
+++ b/popups.css
@@ -38,6 +38,7 @@
     padding: 0.5rem;
 }
 .actions {
+    margin-bottom: 0.5rem;
     font-family: system;
     border-width: 0;
     background-color: color(var(--foreground) alpha(0.1));
--- a/popups.css
+++ b/popups.css
@@ -1,6 +1,9 @@
 .lsp_popup {
-    margin: 0.5rem;
+    margin: 0.5rem 0.5rem 0 0.5rem;
     font-family: system;
+}
+.lsp_popup p {
+    margin-bottom: 0.5rem;
 }
 .highlight {
     border-width: 0;


jfcherng:
I think the problem is in your theme as I don't experience this with Adaptive/Default theme?

predragnikolic:
To reproduce this, try setting `"show_symbol_action_links": false,`.

jfcherng (approved):

rchl:
But it sounds like your change breaks this case https://github.com/sublimelsp/LSP/pull/1523#issuecomment-743675945

The popup can be in different configurations, depending on settings or server response. And since ST doesn't support CSS selector like `:last-child` we need to carefully think of every possible case or have some generic solution.

I haven't looked more closely into this change, just talking in general.

predragnikolic:
I fixed the issue @jfcherng reported with the actions. But I agree that we need a more generic solution.

Please, don't merge this yet. I have one idea to try.(hopefully I will do it today)

predragnikolic:
I tried my other solution, but it doesn't work as I expected.

<details>
 <summary>If you want read what I tried</summary>

I wanted to apply only `margins-left` and `margins-right` (0.5rem) to the `.lsp_popup` 
and `margin-top`, `margin-bottom` to other children of the lsp_popup. 
Which would theoretically space the elements equally vertically, but it ended up looking like this:

Screenshot 2020-12-12 at 17 09 16

> NOTE: The margin at the top of the popup is bigger than it should be.


<details>
 <summary>here is the popup.css I tried for the other solution</summary>

```css
 .lsp_popup {
    margin: 0 0.5rem;
    font-family: system;
}
.lsp_popup p {
    margin: 0.5rem 0;
}
.highlight {
    margin: 0.5rem 0;
    border-width: 0;
    border-radius: 0;
}
.diagnostics {
    margin: 0.5rem 0;
    font-family: monospace;
}
.errors {
    border-width: 0;
    background-color: color(var(--redish) alpha(0.25));
    color: var(--foreground);
    padding: 0.5rem;
}
.warnings {
    border-width: 0;
    background-color: color(var(--yellowish) alpha(0.25));
    color: var(--foreground);
    padding: 0.5rem;
}
.info {
    border-width: 0;
    background-color: color(var(--bluish) alpha(0.25));
    color: var(--foreground);
    padding: 0.5rem;
}
.hints {
    border-width: 0;
    background-color: color(var(--bluish) alpha(0.25));
    color: var(--foreground);
    padding: 0.5rem;
}
.actions {
    margin: 0.5rem 0;
    font-family: system;
    border-width: 0;
    background-color: color(var(--foreground) alpha(0.1));
    color: var(--foreground);
    padding: 0.5rem;
}
.actions a.icon {
    text-decoration: none;
}
pre.related_info {
    border-top: 1px solid color(var(--background) alpha(0.25));
    margin: 0.5rem 0;
    padding-top: 0.7rem;
}
```

</details>
</details>

I am out of ideas :)

rchl:
It looks like the issue is coming from mdpopups which has both margin and padding to the `.highlight` class. Trying to compensate for that would be quite error-prone as it's not even exactly our code.

Screenshot 2020-12-12 at 20 40 24

A proper solution might be to override mdpopups styles.
I have a look a bit myself.

rchl:
Actually, there is also an ST bug here as it doesn't seem to support margin collapsing (or maybe not all aspects of it). If it did that margin of the popup container would get collapsed with the margin of mdpopups and the margin would not have double height.

predragnikolic:
I think I am gonna close this one, as it doesn't solve all the cases.

I sill believe that this can be fixed by overriding the mdpopup classes and without waiting ST to fix https://github.com/sublimehq/sublime_text/issues/3828. But this is not that PR :)

rchl:
I think it would be rather tricky to fix it properly with the current ST bug. You could override mdpopups bottom margin which would fix it when it's the last element in the popup but would also remove the spacing when there are adjacent mdpopup code blocks.

You'd need to target the last element in the popup with some special class but that would be either tricky to do because you would have to parse HTML or messy if dealing with pure string.

Issue #1429, opened 2020-11-07
Implement textDocument/callHierarchy


rwols:
**Is your feature request related to a problem? Please describe.**
I want to see an overview of the callers and callees of a callable.

**Describe the solution you'd like**
Implement textDocument/callHierarchy in the upcoming 3.16 spec.

**Describe alternatives you've considered**
textDocument/references

**Additional context**
clangd can do this so it would be nice to have this.

rwols:
Relevant requests and links:

textDocument/prepareCallHierarchy
https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-16.md#textDocument_prepareCallHierarchy

callHierarchy/incomingCalls
https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-16.md#callHierarchy_incomingCalls

callHierarchy/outgoingCalls
https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-16.md#callHierarchy_outgoingCalls

PR #1524, opened 2020-12-12
Make panels consistent

predragnikolic:
This is a draft that aims to make the diagnostics, references and rename panel look the same.

Diagnostics before:
Screenshot 2020-12-12 at 16 21 48

Diagnostics after:
Screenshot 2020-12-12 at 16 23 11

References before:
Screenshot 2020-12-12 at 16 21 58

References after:
Screenshot 2020-12-12 at 16 23 18

Rename before:
Screenshot 2020-12-12 at 16 22 05

Rename after:
Screenshot 2020-12-12 at 16 23 25

##### Questions

1. Should we have a \n after each as a separator between filenames?
```
plugin/symbols.py:
	   1:28   from .core.protocol import Request, Range
plugin/core/protocol.py:
	  98:7    class Request:
```
or
```
plugin/symbols.py:
	   1:28   from .core.protocol import Request, Range

plugin/core/protocol.py:
	  98:7    class Request:
```

2. I think I can reuse the same syntax file for references to both references and rename panels? Should I delete the `Rename.sublime-syntax` file? Please give me a suggestion on what to do here. :)

3. The line_regex now must contain a space in front. 
I put a `\t` in front of any line diagnostics, is that ok change?


rwols:
Looks like a good opportunity to refactor the "rename" and "references" panel to a common builder class. We're going to need such a class anyway, as we'll also be needing to present the "callees" and "callers" from the new set of call hierarchy requests of 3.16: https://github.com/sublimelsp/LSP/issues/1429

predragnikolic:
> a common builder class 

Something like this? 
```python
class Panel:
    def __init__(self, window: sublime.Window, name: str, unlisted=False):
        """
        :param      name:      The name of the output panel
        :param      unlisted:  Control if the output panel should be listed in the panel switcher
        """
        self.window = window
        self.panel = self.window.create_output_panel(name)
        self.name = "output.{}".format(name)

    def is_open(self):
        return self.window.active_panel() == self.name

    def open(self):
        self.window.run_command("show_panel", {"panel": self.name})

    def hide(self):
        self.window.run_command("hide_panel", {"panel": self.name})

    def update(self, content: str):
        self.panel.run_command("lsp_update_panel", {"characters": content})

    def destroy(self):
        self.window.destroy_output_panel(self.name)
```

Not sure what you have envisioned :)

plugin/core/views.py
```diff
@@ -564,7 +564,7 @@ def format_diagnostic_for_panel(diagnostic: Diagnostic) -> Tuple[str, Optional[i
             formatted.append(code)
     lines = diagnostic.message.splitlines() or [""]
     # \u200B is the zero-width space
-    result = "{:>4}:{:<4}{:<8}{} \u200B{}".format(
+    result = "\t{:>4}:{:<4}{:<8}{} \u200B{}\n".format(
```
rwols, line 567: The `\n` doesn't work for multi-line messages. You'd need to append the additional `\n` somewhere at the end of this function, after the additional indented lines of the message have been added.

Having said that, I prefer no empty lines :)

rwols: Actually, it's better if the WindowManager took care of that extra `\n`, if we're going to add that. But again I prefer no `\n` :)

plugin/core/windows.py
```diff
@@ -410,7 +410,7 @@ def handle_show_message(self, session: Session, params: Any) -> None:
         sublime.status_message("{}: {}".format(session.config.name, extract_message(params)))
 
     def update_diagnostics_panel_async(self) -> None:
-        to_render = []  # type: List[str]
+        to_render = ""  # type: str
```
rwols, line 413: What makes you think appending strings is faster than joining a list?

predragnikolic: I haven't tought about what is faster.
I saw the way the content for rename and reference was created, 
so I wanted to to use string concatenation to make it consistent for all three:

https://github.com/sublimelsp/LSP/blob/st4000-exploration/plugin/references.py#L123
https://github.com/sublimelsp/LSP/blob/st4000-exploration/plugin/rename.py#L163

That is the reason why I did that. :)

rwols:
> Not sure what you have envisioned :)

A builder class... to abstract away all of the boring indent logic and formatting logic. So something like this:


```python
class PanelBuilder:

    def preamble(self, text: str) -> None:
        """Set the preamble text"""
        ...

    def filename(self, filename: str) -> None:
        """Render a filename to the panel"""
        ...

    def diagnostic(self, diagnostic: Diagnostic) -> None:
        """Render a diagnostic to the panel"""
        ...

    def location(self, range: Dict[str, Any], identifier: Optional[str] = None) -> None:
        """Render a reference/rename location"""
        ...

    def __str__(self) -> str:
        """Flush the content as a string"""
        ...
```

does that make sense?

plugin/core/windows.py
rwols: `''.join` is the preferred way to build up chunks of a larger string: https://stackoverflow.com/a/1350289

predragnikolic:
@rwols I know I haven't implemented your suggestion yet. (for the PanelBuilder)
but I wanted to build a `Panel` class that abstracts away the interactions with the panel api.

My question is, should I revert the 7901a7c? 
If the abstraction is too much, or not necessary at the moment I can revert it with no problem :)

predragnikolic:
I reverted the changes 2a962e0 
Can PanelBuilder also be implemented in a different PR?
I just wanted the panels to be consistent :)

plugin/core/windows.py
predragnikolic: List or string?
Something tells me that it should be a list :)

rwols:
## 1

> Should we have a \n after each as a separator between filenames?

I don't see the need for it to be honest.

## 2

> Should I delete the Rename.sublime-syntax file?

Sounds good.

## 3

> The line_regex now must contain a space in front. I put a \t in front of any line diagnostics, is that ok change?

I don't understand why this is needed? Why doesn't SublimeLinter use a tab? http://www.sublimelinter.com/en/stable/

## 4

Please don't use string concatenation in windows.py

plugin/core/windows.py
predragnikolic: done with 02d6859

plugin/core/views.py
predragnikolic: I removed the `\n` and `\t` from those lines with 02d6859 
```diff
- \t{:>4}:{:<4}{:<8}{} \u200B{}\n\
+ {:>5}:{:<4}{:<8}{} \u200B{}\
```

predragnikolic:
4 - is done.
 3 - is done.
 2 - is done.
 1 - Although a `\n` is not necessary I would argue that a new line gives a visual separation of the results which help readability. Because of that I would keep the \n between filenames. :)

plugin/core/windows.py
```diff
@@ -439,6 +439,7 @@ def update_diagnostics_panel_async(self) -> None:
                 if offset is not None and code is not None and href is not None:
                     prephantoms.append((row, offset, code, href))
                 row += content.count("\n") + 1
+            to_render.append("")  # add spacing between filenames
```
rwols, line 442: Does this require another `row += 1` ?

rwols: You can check this by running with e.g. eslint and creating diagnostics in at least two files.

predragnikolic:
```diff
 to_render.append("")  # add spacing between filenames
+ row += 1
```
I don't see a difference.
with `row += 1`:

with
or without `row += 1`:
without

But I have noticed on the first line that `typescript:6133` is not highlighted. (I am not a ST syntax expert, not sure how to fix it)
```
     1:1   hint    'styles' is declared but its value is never read. ​typescript:6133
````

rwols: I'm talking about the clickable links, make sure to check those:

image

predragnikolic: Thanks for pointing this out.

> Does this require another row += 1 ?

It turns out that it does: 

Screenshot from 2020-12-18 21-21-23
the phantom placement is messed up.

It is fixed with 0bcd931

rchl: > But I have noticed on the first line that `typescript:6133` is not highlighted. (I am not a ST syntax expert, not sure how to fix it)

There is missing `pop: true` in here: https://github.com/sublimelsp/LSP/blob/0bcd93110d98f74879f274b2cb10fa50aec6a663/Syntaxes/Diagnostics.sublime-syntax#L54-L55

predragnikolic: Thanks @rchl,
I included the fix e729f45

rwols (approved):

PR #1525, opened 2020-12-12
Never close the diags panel automatically, and more

rwols:
Because the diags panel is never closed automatically, it makes little sense
to have it automatically open whenever there's an error/warning. So the
setting

    auto_show_diagnostics_panel

is removed. Instead

    auto_show_diagnostics_panel_level

is used to determine when to automatically open the panel *on save*. Put the
value `0` here to never open the panel.

I've also added a neutral "idle" message for when there are no diagnostics.

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -1,4 +1,3 @@
-import functools
 from ...third_party import WebsocketServer  # type: ignore
 from .configurations import ConfigManager
 from .configurations import WindowConfigManager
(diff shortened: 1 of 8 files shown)


plugin/core/windows.py
```diff
@@ -1,3 +1,4 @@
+import functools
```
rchl, line 1: wrong order

rchl (approved):

Issue #1530, opened 2020-12-22
streamingProcess: runInteractiveProcess: exec: does not exist (No such file or directory)


JeremyBois:
> Using a `stack` project I get the following error:
> 
> `streamingProcess: runInteractiveProcess: exec: does not exist (No such file or directory)'`

I check online and it seems [`nix` have the same issue with `stack`](https://github.com/commercialhaskell/stack/issues/2282). 
> ```For future reference (me): streamingProcess: runInteractiveProcess: exec: does not exist usually means it called an executable that failed to run since dynamic libraries weren't there, meaning you ran nix-collect-garbage.```

Related to commercialhaskell/stack#2282

_Originally posted by @JeremyBois in https://github.com/sublimelsp/LSP/issues/1529#issuecomment-749432945_

JeremyBois:
The solution is to update the **PATH** to include missing binaries or libraries.

One solution is to update the `~/.profile` with needed path:
```bash
### Home
export PATH=$PATH:~/.local/bin:~/bin

### ghcup-hs
export PATH=$PATH:~/.ghcup/bin

### Cabal
export PATH=$PATH:~/.cabal/bin
```

 Another solution is to update the `env` variable in the server configuration (replace `<Home_path>` by your Home path):
```json
"haskell-language-server":
{
    "command":
    [
        "haskell-language-server-wrapper",
        "--lsp"
    ],
    
    // ...

    "env":
    {
        "PATH": "/usr/local/bin:/usr/bin:<Home_path>/.local/bin:<Home_path>/bin:<Home_path>/.ghcup/bin:<Home_path>/.cabal/bin"
    }
}
```

rwols:
Awesome! @moodmosaic perhaps you can expand on this in the docs? Could be a blocker for more users.

JeremyBois:
> Awesome! @moodmosaic perhaps you can expand on this in the docs? Could be a blocker for more users.

Yeah sure I will.

Thanks again !

moodmosaic:
I'd be more than happy to do so, but first I must follow the thread. Perhaps the best person to do this is @JeremyBois. If not, I can look into it over the next few days.

JeremyBois:
Oh I misread @rwols comment ...

@moodmosaic, I will try to update the docs today if possible if it's ok. I just first want to finish to setup and test everything is working on my side in different project configurations.

If I have issues with the process I will ask for help.

moodmosaic:
[...]

Issue #1529, opened 2020-12-21
Language server haskell-language-server has crashed


JeremyBois:
**Describe the bug**
Immediately get the message asking to restart the server because it crashed.
Tested with `stack` and `direct` cradle.

I check using directly `haskell-language-server-wrapper` on the terminal without any issues for both `stack` and `direct` projects.
Also, both projects works on `vim` (using [CoC](https://github.com/neoclide/coc.nvim)) and `vscode` (using [haskell](https://github.com/haskell/vscode-haskell)).

**To Reproduce**
Start the `haskell-language-server-wrapper` LSP server. 

A minimal example can be downloaded [here](https://we.tl/t-2npcy5RAPF) and contains:
  - haskell source file - `Main.hs`
  - hie config - `hie.yaml`
  - `info`folder with 
    - LSP log trace
    - current configuration (same as the one provided below)
    - screenshot or the bug

**Expected behavior**
No crash. 

**Screenshots**
![LSP_crash](https://user-images.githubusercontent.com/38559279/102773953-58d78800-438a-11eb-9c9c-85da0cc5b47b.png)

**Environment (please complete the following information):**
- OS: Arch Linux (kernel 5.9.13-arch1-1)
- Sublime Text version: 3210
- LSP version: 0.14.3
- Language servers used: haskell-language-server

**Additional context**
I try with and without the `initializationOptions` part of the configuration below:
```json
{
	"clients":
	{
		"haskell-language-server":
		{
			"command":
			[
				"haskell-language-server-wrapper",
				"--lsp"
			],
			"enabled": true,
			"initializationOptions":
			{
				"languageServerHaskell":
				{
					"completionSnippetsOn": true,
					"hlintOn": true,
					"maxNumberOfProblems": 100
				}
			},
			"languageId": "haskell",
			"scopes":
			[
				"source.haskell"
			],
			"syntaxes":
			[
				"Packages/Haskell/Haskell.sublime-syntax",
				"Packages/SublimeHaskell/Syntaxes/Haskell-SublimeHaskell.sublime-syntax"
			]
		}
	},
	"log_debug": true,
	"log_payloads": true,
	"log_stderr": true
}
```

Thanks,

rchl:
You have the log panel open in the screenshot but you haven't included the last lines of the log which might give better idea of what is crashing.

JeremyBois:
The log can be found in the [minimal example](https://we.tl/t-2npcy5RAPF) `info` folder. 

```text
[...]

PR #1531, opened 2020-12-22
Documentation issue1529 issue1530

JeremyBois:
Add documentation to Troubleshooting section to explain how to solve #1529 and #1530.

@moodmosaic @rwols

### DIFF
--- a/docs/src/troubleshooting.md
+++ b/docs/src/troubleshooting.md
@@ -77,7 +77,7 @@ The complete procedure of updating the `PATH` used by Sublime Text depends on yo
 
 This error could occurs when using `stack` or `cabal` (`Haskell` language building tools) with the `haskell-language-server` if a library or binary is missing in the language server `PATH`.
 
-Depending on how you installed your `Haskell` tools, you will need to add one or more of the following to your Sublime Text `PATH`.
+Depending on how you installed your `Haskell` tools, you will need to add one or more of the following lines to your login script for bash-like shells on a unix-like operating system.
 ```bash
 ### Home specific and default Stack binaries location
 export PATH=$PATH:~/.local/bin:~/bin
@@ -95,7 +95,7 @@ The complete procedure of updating the `PATH` used by Sublime Text depends on yo
 ### 4. Popup error `Language server <your_server_language_name> has crashed`
 Assuming that the server is actually installed, and that you can start it from your shell, this issue is likely due to Sublime Text's internal environment not picking up the same `PATH` environment variable as you've configured in your shell.
 
-Server languages depend on other binaries that should also be in your `PATH` in addtion to the server binary itself.
+Language servers may depend on other binaries that should also be in your `PATH` in addtion to the server binary itself.
 
 For instance if you have installed the `haskell-language-server` using [ghcup-hs](https://gitlab.haskell.org/haskell/ghcup-hs) you should expose its specific installation folder `~/.ghcup/bin`. If the build process uses `stack` then it should also be in your `PATH`.
 
--- a/docs/src/troubleshooting.md
+++ b/docs/src/troubleshooting.md
@@ -91,6 +91,18 @@ export PATH=$PATH:~/.cabal/bin
 
 The complete procedure of updating the `PATH` used by Sublime Text depends on your platform and is explained [here](troubleshooting.md#updating-the-path-used-by-lsp-servers).
 
+
+### 4. Popup error `Language server <your_server_language_name> has crashed`
+Assuming that the server is actually installed, and that you can start it from your shell, this issue is likely due to Sublime Text's internal environment not picking up the same `PATH` environment variable as you've configured in your shell.
+
+Server languages depend on other binaries that should also be in your `PATH` in addtion to the server binary itself.
+
+For instance if you have installed the `haskell-language-server` using [ghcup-hs](https://gitlab.haskell.org/haskell/ghcup-hs) you should expose its specific installation folder `~/.ghcup/bin`. If the build process uses `stack` then it should also be in your `PATH`.
+
+The complete procedure of updating the `PATH` used by Sublime Text depends on your platform and is explained [here](troubleshooting.md#updating-the-path-used-by-lsp-servers).
+
+
+
 ## Known Issues
 
 ### Completions not shown after certain keywords
--- a/docs/src/troubleshooting.md
+++ b/docs/src/troubleshooting.md
@@ -5,25 +5,13 @@ Enable server logging: set `log_server` to ["panel"] and `log_stderr` to `true`
 Run "LSP: Toggle Log Panel" from the command palette. No restart is needed.
 If you believe the issue is with this package, please include the output from the Sublime console in your issue report!
 
-## Common problems
+## Updating the PATH used by LSP servers
 
-### 1. LSP doesn't start my language server
+You can confirm that your issue is due to `PATH` being different by starting Sublime Text from the command line so that it inherits your shell's environment.
 
-* Make sure you have a folder added in your Sublime workspace.
-* Make sure the document you are opening lives under that folder.
+The solution is to make ST read the same `PATH` that is read by your shell (or OS in general, in case of Windows).
 
-Your client configuration requires two settings to match the document your are editing:
-
-* Scope (eg. `source.php`): Verify this is correct by running "Show Scope Name" from the developer menu.
-* Syntax (eg. `Packages\PHP\PHP.sublime-syntax`): Verify by running `view.settings().get("syntax")` in the console.
-
-### 2. LSP cannot find my language server (`No such file or directory: 'xyz'`)
-
-Assuming that the server is actually installed, and that you can start it from your shell, this issue is likely due to Sublime Text's internal environment not picking up the same `PATH` environment variable as you've configured in your shell.
-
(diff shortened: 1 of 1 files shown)


docs/src/troubleshooting.md
```diff
@@ -42,15 +30,78 @@ Adjusting `PATH` can differ based on the operating system and the default shell
...
+export PATH=$PATH:~/.cabal/bin
+```
+
+The complete procedure of updating the `PATH` used by Sublime Text depends on your platform and is explained [here](troubleshooting.md#updating-the-path-used-by-lsp-servers).
+
+
+### 4. Popup error `Language server <your_server_language_name> has crashed`
+Assuming that the server is actually installed, and that you can start it from your shell, this issue is likely due to Sublime Text's internal environment not picking up the same `PATH` environment variable as you've configured in your shell.
+
+Server languages depend on other binaries that should also be in your `PATH` in addtion to the server binary itself.
```
rwols, line 98: Server languages --> Language servers

rwols, line 98: depend --> may depend

docs/src/troubleshooting.md
```diff
@@ -42,15 +30,78 @@ Adjusting `PATH` can differ based on the operating system and the default shell
...
+If, for example, you want to expose a `Node` binary to ST and you have it installed through a version manager like `nvm`, you need to insert its [initialization script](https://github.com/nvm-sh/nvm#install--update-script) in the location specified in [this table](troubleshooting.md#updating-the-path-used-by-lsp-servers)
+
+The complete procedure of updating the `PATH` used by Sublime Text depends on your platform and is explained [here](troubleshooting.md#updating-the-path-used-by-lsp-servers).
+
+### 3. Error in build `streamingProcess: runInteractiveProcess: exec: does not exist (No such file or directory)`
+
+This error could occurs when using `stack` or `cabal` (`Haskell` language building tools) with the `haskell-language-server` if a library or binary is missing in the language server `PATH`.
+
+Depending on how you installed your `Haskell` tools, you will need to add one or more of the following to your Sublime Text `PATH`.
```
rwols, line 80: you will need to add one or more of the following to your Sublime Text `PATH`.

-->

you will need to add one or more of the following lines to your login script for bash-like shells on a unix-like operating system.

rwols (approved):
Yeah, looks good to me. @rchl or @predragnikolic any thoughts?

predragnikolic:
Looking at this :)

predragnikolic:
I'll be a critique on this one, but bare with me :)

I don't thing that the [4 point - Popup error Language server <your_server_language_name> has crashed](https://github.com/sublimelsp/LSP/blob/4308f561621951d04c6ba7d5b86cc6f1eb6fcc82/docs/src/troubleshooting.md#4-popup-error-language-server-your_server_language_name-has-crashed) is necessary, because it was described in the [2 point](https://github.com/sublimelsp/LSP/blob/st4000-exploration/docs/src/troubleshooting.md#2-lsp-cannot-find-my-language-server-no-such-file-or-directory-xyz). Just moving it to a 4 point doesn't make a big difference. :) 

Lets compare [point 3 -  Error in build streamingProcess: runInteractiveProcess: exec: does not exist (No such file or directory)](https://github.com/sublimelsp/LSP/blob/4308f561621951d04c6ba7d5b86cc6f1eb6fcc82/docs/src/troubleshooting.md#3-error-in-build-streamingprocess-runinteractiveprocess-exec-does-not-exist-no-such-file-or-directory) 
with [point 2 - LSP cannot find my language server (No such file or directory: 'xyz')](https://github.com/sublimelsp/LSP/blob/4308f561621951d04c6ba7d5b86cc6f1eb6fcc82/docs/src/troubleshooting.md#2-lsp-cannot-find-my-language-server-no-such-file-or-directory-xyz).
They both try to solve the same issue: `does not exist (No such file or directory)`.

Point 3 is not a generic solution, it is tailored for the Haskell ecosystem, 
while the point 2 has is more generic. 

I am curious to see why the second step didn't help you to solve your problem in the first place? (Was it not clear, was it missing an example, what could we do to make the second step more)

Summary:
I think the [current troubleshooting guide](https://github.com/sublimelsp/LSP/blob/st4000-exploration/docs/src/troubleshooting.md#2-lsp-cannot-find-my-language-server-no-such-file-or-directory-xyz) is simpler, than the one in this PR. 

If you could just pin-point the problems you had and say that a particular paragraph/s wasn't clear in step 2. If that was written a little differently that would have helped me solve my issue.
so to make the current troubleshooting guide more simpler that would be nice :)

JeremyBois:
> I don't thing that the [4 point - Popup error Language server <your_server_language_name> has crashed](https://github.com/sublimelsp/LSP/blob/4308f561621951d04c6ba7d5b86cc6f1eb6fcc82/docs/src/troubleshooting.md#4-popup-error-language-server-your_server_language_name-has-crashed) is necessary, because it was described in the [2 point](https://github.com/sublimelsp/LSP/blob/st4000-exploration/docs/src/troubleshooting.md#2-lsp-cannot-find-my-language-server-no-such-file-or-directory-xyz). Just moving it to a 4 point doesn't make a big difference. :)

Actually I did not get the message `No such file or directory: "xyz"` but only a message indicating the crash of the plugin. So I think a proper section can help a future user. 

Also an alternative solution is added in the document. It consists of updating `env` variable directly in the **LSP** settings for a specific server. This approach allows to only expose a folder for a specific language server.

> Lets compare point 3 - Error in build streamingProcess: runInteractiveProcess: exec: does not exist (No such file or directory) with point 2 - LSP cannot find my language server (No such file or directory: 'xyz'). They both try to solve the same issue: does not exist (No such file or directory).

Again yes, the same solution for another issue. In this case the problem is a missing dependency with a **working** language server. For instance when the binary for the language server and one of its dependency is not in the same folder.

> Point 3 is not a generic solution, it is tailored for the Haskell ecosystem, while the point 2 has is more generic.

Yes I agree with this one. The error message is linked to Haskell in this case but the missing dependency is always specific to the server. Maybe I could remove this part and add a note on missing dependencies issue ?


> If you could just pin-point the problems you had and say that a particular paragraph/s wasn't clear in step 2. The goal is to make the current troubleshooting guide more simpler :)

As said above, I never get `No such file or directory: "xyz"`, so the current troubleshooting was not directly useful.


**Summary:**
  - Point 4 is different from point 2 even if the solution is the same
  - Point 3 is too specific and could be removed
  - A note about a missing language server dependencies is needed
  - Added alternative solution avoid polluting the global user space only for a specific server

predragnikolic:
I would just suggest removing the 3rd point from the troubleshooting guide.
The rest what you said in your summary is ok.

docs/src/troubleshooting.md
```diff
@@ -42,15 +30,60 @@ Adjusting `PATH` can differ based on the operating system and the default shell
...
+
+
+### 3. Popup error `Language server <your_server_language_name> has crashed`
+Assuming that the server is actually installed, and that you can start it from your shell, this issue is likely due to Sublime Text's internal environment not picking up the same `PATH` environment variable as you've configured in your shell.
+
+> **Note** : Language servers may have dependencies that should also be in your `PATH` in addition to the server binary itself.
+
+For instance if you have installed the `haskell-language-server` using [ghcup-hs](https://gitlab.haskell.org/haskell/ghcup-hs) you should expose its specific installation folder `~/.ghcup/bin`. If the build process uses `stack` then it should also be in your `PATH`.
+
+The complete procedure of updating the `PATH` used by Sublime Text depends on your platform and is explained [here](troubleshooting.md#updating-the-path-used-by-lsp-servers).
```
rchl, line 84: I don't think this should be here. The reasons for server crashing can be plenty and I don't think a wrong PATH is a frequent reason for those. I'd probably add it in the Haskell section in https://github.com/sublimelsp/LSP/blob/st4000-exploration/docs/src/language_servers.md instead.

docs/src/troubleshooting.md
```diff
@@ -42,15 +30,60 @@ Adjusting `PATH` can differ based on the operating system and the default shell
...
+
+Your client configuration requires two settings to match the document your are editing:
+
+* Scope (eg. `source.php`): Verify this is correct by running "Show Scope Name" from the developer menu.
+* Syntax (eg. `Packages\PHP\PHP.sublime-syntax`): Verify by running `view.settings().get("syntax")` in the console.
+
+### 2. LSP cannot find my language server (`No such file or directory: 'xyz'`)
+Assuming that the server is actually installed, and that you can start it from your shell, this issue is likely due to Sublime Text's internal environment not picking up the same `PATH` environment variable as you've configured in your shell.
+
+The exact changes to make can differ depending on what program you want to expose to Sublime Text. The simplest way is to extend the path like so (replacing `/usr/local/bin` with the path of your choice):
```
rchl, line 67: The `The exact changes to make` paragraph (including the code block) should be in the `Updating the PATH used by LSP servers` section. Just maybe add info that this is a Linux specific method.

rchl (changes requested):

rwols:
We can only ask so much from contributors. Maybe we can merge this and then you can make the adjustments, @rchl?

rwols (approved):

rchl:
> We can only ask so much from contributors. Maybe we can merge this and then you can make the adjustments, @rchl?

I can do that when I find time and courage.

PR #1533, opened 2020-12-23
Replace Diagnostic by a TypedDict

rwols:
In preparation for #1527, simplify a lot of the code.

Replacing Diagnostic by a TypedDict made me rethink the "next/prev diagnostic" functionality. I'm convinced it can just as well be done with the existing popup code using ST's built-in F4 (`next_result`) and shift+F4 (`prev_result`). This has the major advantage of having much less code to maintain. As well as not having to account for two different ways to render a diagnostic to minihtml.

It's quite a bit of changes. If you want to inspect it I'd suggest looking at the two commits separately as they should be stand-alone.


rchl:
Haven't looked at the changes but have a question. Are your changes compatible with the use-case where the user doesn't use the diagnostic panel but just wants to navigate the diagnostics shown in the view?

You said that you are relying on native next/previous_result but doesn't that require the results to be visible in the panel?

rwols:
> You said that you are relying on native next/previous_result but doesn't that require the results to be visible in the panel?

No. this works even when the panel is not visible :)

PR #1535, opened 2020-12-30
Implement client-initiated progress reporting

### LINKED ISSUE #1387: Implement client-initiated workDoneProgress
Some servers have the ability to report progress for long-running requests. For instance, when workspace/symbol takes a while it would be nice to see a progress indication in the status bar for this.

Implement workDoneProgress in a general manner that would work for any request.

rwols:
For a selected number of requests we'll report progress in the status bar.
They are precisely the requests that are explicitly ran by the user. So for
instance: references, rename, symbols.

The implicit requests like code actions, highlights, hover, completions and
signature help are not progress-enabled.

I tried this on pyright, which supposedly supports client-initiated progress,
but I can't seem to get it to work. It instead starts doing a server-initiated
progress scheme.

- [x] Advertise client-initiated progress depending on the particular request
- [x] Handle $/progress reports from servers that understand client-initiated progress
- [x] Have a fallback to just showing the method when the server doesn't support client-initiated progress
- [x] When the request is fast to complete, do not show any progress (using some form of debounce)

Resolves #1387


predragnikolic:
So when I open a `.js` file and wait for the server to initialize,
when I run rename for the first time I will see `textDocument/rename` for about a few second in the status bar. (Which is ok)

If I run rename for the second time I will see `textDocument/rename` for a really really short period of time(barely noticeable, but still noticeable) before it disappears. Because of that, other text in the status bar is shifted to right and and that quickly back to the left because the request status is visible for less than 30ms. 

The benefit of this PR is that it reports that a request is in progress, but do we need that?
Does the user really need to care that a request is in progress(because most of the time, the time between the request and response is not that big, but that depends on the LS and the size of the project)

rchl:
That issue could likely be solved by having status only be reported in the UI if the command takes more than x ms to run (for example 200ms). It can be useful to be able to see if the command is taking a long time to run but there is really no point in reporting the status of commands that finish quickly.

rwols:
>  visible for less than 30ms.

I feared someone would say this :)

We absolutely do need this. Working in very large code bases like [LLVM](https://github.com/llvm/llvm-project), language servers like clangd start becoming slow for some requests. In that case it's helpful to see that at least something is happening. In other words, it's good to see that the LSP client acknowledged the request of the user, and the request is in progress.

But as @rchl noted some sort of debounce should be implemented so that we don't show the active requests for the common case that the requests finish quickly.

PR #1536, opened 2020-12-30
Rename auto_show_diagnostics_panel_level to show_diagnostics_panel_on_save

### LINKED ISSUE #1528: Rename `auto_show_diagnostics_panel_level` to `show_diagnostics_panel_on_save`
`auto_show_diagnostics_panel_level` might not be a clear and simple setting name as it could be.

`show_diagnostics_panel_on_save` gives more context that `auto_show_diagnostics_panel_level`.

I considered even suggesting the removed `auto_show_diagnostics_panel` instead of `auto_show_diagnostics_panel_level`, but I still think that `show_diagnostics_panel_on_save` is more descriptive.

rwols:
Resolves #1528

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -34,7 +34,7 @@
   // warning: 2
   // info: 3
   // hint: 4
-  "auto_show_diagnostics_panel_level": 2,
+  "show_diagnostics_panel_on_save": 2,
 
   // Show errors and warnings count in the status bar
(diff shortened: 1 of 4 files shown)


rwols:
I think we can get away with just renaming this setting. Only ST4 users would be affected by this breaking change. Still backwards-compatible for ST3 users.

predragnikolic (approved):

PR #1539, opened 2021-01-05
Add A File Icon alias scopes to language-ids

deathaxe:
This PR adds all alias scopes of A File Icon, which point to source or text files.

Those aliases are created to attach unique icons to special named files, such as unittests or makefiles. All those alias files either use
text.plain or another scope of a syntax, shipped with Sublime Text.

As all those aliases add a sub-scope to the used base scope, they need to be added to this list here.

**Example:**

 All files ending with `.test.js` are scoped `source.js.unittest`.
 This commit makes sure to send the correct language id `javascript` to
 language-servers.

### DIFF
--- a/language-ids.sublime-settings
+++ b/language-ids.sublime-settings
@@ -21,12 +21,32 @@
 // request at github.com/sublimelsp/LSP if you believe an entry is missing.
 {
     "source.c++": "cpp",
+    "source.coffee.gulpfile": "coffeescript", // https://github.com/SublimeText/AFileIcon
     "source.cs": "csharp",
     "source.dosbatch": "bat",
     "source.fixedform-fortran": "fortran", // https://packagecontrol.io/packages/Fortran
+    "source.groovy.gradle": "groovy", // https://github.com/SublimeText/AFileIcon
+    "source.groovy.jenkins": "groovy", // https://github.com/SublimeText/AFileIcon
     "source.js": "javascript",
+    "source.js.eslint": "javascript", // https://github.com/SublimeText/AFileIcon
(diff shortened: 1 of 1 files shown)


rchl (approved):

rwols:
Many thanks!

rwols:
I did not realize you were working on it as well :) my hopeless attempt is here https://github.com/sublimelsp/LSP/commit/ab97ab7bafc2a3d9eaa83628b49caafeee6b92a7#diff-716f435a4afb9e9df93a69abab8e51b57cba8073ce7ef43c645b0d8ad43935cf

jwortmann:
Since all these scopes only append a subscope like e.g. `.unittest` to already existing base scopes `source.js` here, would it make sense to match the scopes in `language-ids.sublime-settings` against the base scope of the document and choose the best match? This way the best language-id is still chosen automatically if another syntax package uses such a scope extension for a document, without the need to update the scope-language-mapping (and new entries are only needed if a different language-id is desired for a specific scope extension).

Something like this (untested):
```python
def basescope2languageid(base_scope: str) -> str:
    # This the connection between Language IDs and ST selectors.
    base_scope_map = sublime.load_settings("language-ids.sublime-settings")
    best_match = None
    best_score = 0
    for selector, language_id in base_scope_map.items():
        score = sublime.score_selector(base_scope, selector)
        if score > best_score:
            best_score = score
            best_match = language_id
    if not best_match:
        best_match = base_scope.split(".")[-1]
    return result if isinstance(best_match, str) else ""
```

rwols:
It's an interesting idea, but it turns an amortized O(1) operation into an O(N) operation (N = number of exceptions). Although maybe unnoticeable in practice.

PR #1540, opened 2021-01-05
Make <hr> less noticeable in popups

predragnikolic:
Before:
the hr was a solid foreground color.
Screenshot from 2021-01-05 23-09-32

After:
the hr is an slightly opaque line.
Screenshot from 2021-01-05 23-09-58
Screenshot from 2021-01-05 23-09-48

### DIFF
--- a/popups.css
+++ b/popups.css
@@ -2,6 +2,9 @@
     margin: 0.5rem;
     font-family: system;
 }
+.lsp_popup hr {
+  border-color: color(var(--foreground) alpha(0.10));
+}
 .highlight {
     border-width: 0;
     border-radius: 0;


popups.css
```diff
@@ -1,4 +1,5 @@
 .lsp_popup {
+    --mdpopups-hr: color(var(--foreground) alpha(0.10));
```
rwols, line 2: While I don't know much of CSS, the `--mdpopups-hr` looks out of place. Why can't you style `hr` directly? I suppose that would influence the signature help as well?

predragnikolic: Changed it to style the hr directly.

rwols (approved):

PR #1541, opened 2021-01-10
Symbol Tag support

predragnikolic:
In this PR:

NEW:
- added deprecated tags to document symbols popup and  symbols in project popup
- symbols in project now looks consistent with the document symbol popup
- added new typed dicts

FIXES:
- This PR also fixes the invocation of the `LspWorkspaceSymbolsCommand` by removing the [fallback value](https://github.com/sublimelsp/LSP/compare/tags-support?expand=1#diff-514c816d75b08e41f44833deaf8e7281295700ce87fdfff07adbd5aabab403c3R240). - The problem is that I trigger the keybinding for `lsp_workspace_symbols` and I expect to see the popup, but no popup is shown. Some people don't experience this problem, but I do experience it on Linux and Mac. A discussion is held on Discord so I will update this PR if we conclude more about that.

Document symbol popup
document-symbol

Symbols in project popup - before
workspace-symbols-before

 Symbols in project popup - after
workspace-symbols-after

### DIFF
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -132,7 +132,7 @@ DocumentSymbol = TypedDict('DocumentSymbol', {
     'deprecated': Optional[bool],
     'range': Range_T,
     'selectionRange': Range_T,
-    'children': Optional[List[Any]]  # mypy doesn't support recurive types, Optional[List[DocumentSymbol]]
+    'children': Optional[List[Any]]  # mypy doesn't support recurive types like Optional[List[DocumentSymbol]]
 }, total=True)
 
 SymbolInformation = TypedDict('SymbolInformation', {
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -40,6 +40,15 @@ class SignatureHelpTriggerKind:
 
 DocumentUri = str
 
+Position = TypedDict('Position', {
+    'line': int,
+    'character': int
+})
+
+Range_T = TypedDict('Range_T', {
+    'start': Position,
+    'end': Position
+})
 
 CodeDescription = TypedDict('CodeDescription', {
     'href': str
@@ -70,7 +79,7 @@ CodeAction = TypedDict('CodeAction', {
 
 
 CodeLens = TypedDict('CodeLens', {
-    'range': Dict[str, Any],
+    'range': Range_T,
     'command': Optional[Command],
     'data': Any,
     # Custom property to bring along the name of the session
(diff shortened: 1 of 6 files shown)


plugin/symbols.py
```diff
@@ -135,27 +157,32 @@ def on_highlighted(self, index: int) -> None:
...
 
-    def process_symbols(self, items: List[Dict[str, Any]]) -> List[sublime.QuickPanelItem]:
+    def process_symbols(
+            self,
+            items: Union[List[DocumentSymbol], List[SymbolInformation]]
+    ) -> List[sublime.QuickPanelItem]:
         self.regions.clear()
         panel_items = []
         if 'selectionRange' in items[0]:
+            items = cast(List[DocumentSymbol], items)
```
predragnikolic, line 167: pyright cannot deduce the types here, so I cast them explicitly

stubs/sublime.pyi
```diff
@@ -1024,7 +1024,7 @@ class QuickPanelItem:
     def __init__(
         self,
         trigger: str,
-        details: str = "",
+        details: Union[str, List[str]] = "",
```
predragnikolic, line 1027: ST QuickPanelItem also supports a list of strings for details

plugin/symbols.py
```diff
@@ -170,30 +197,27 @@ def process_document_symbol_recursive(self, quick_panel_items: List[sublime.Quic
...
+            if SymbolTag.Deprecated in tags:
+                st_display_type = "⚠ {} - Deprecated".format(st_display_type)
             quick_panel_items.append(
                 sublime.QuickPanelItem(
                     trigger=name,
                     details=st_details,
                     annotation=st_display_type,
                     kind=(st_kind, st_icon, st_display_type)))
             children = item.get('children') or []
+            children = cast(List[DocumentSymbol], children)
```
predragnikolic, line 210: Ideally I shouldn't have used cast here,
but because mypy doesn't support recursive types, 
I couldn't typehint DocumentSymbol['children'] to be the type of `DocumentSymbol`.
https://github.com/sublimelsp/LSP/pull/1541/files#diff-049653877a86e4c220cd537d174655150e80b4ce0077d48742cc72b4654e2112R135

plugin/core/protocol.py
```diff
@@ -108,12 +121,31 @@ class SignatureHelpTriggerKind:
...
 
+DocumentSymbol = TypedDict('DocumentSymbol', {
+    'name': str,
+    'detail': Optional[str],
+    'kind': int,
+    'tags': Optional[List[int]],
+    'deprecated': Optional[bool],
+    'range': Range_T,
+    'selectionRange': Range_T,
+    'children': Optional[List[Any]]  # mypy doesn't support recurive types like Optional[List[DocumentSymbol]]
```
rchl, line 135: Are you just talking about it not being able to see definitions that are not yet defined? In that case you can use a string - `Optional[List['DocumentSymbol']]`

plugin/core/protocol.py
```diff
@@ -70,7 +83,7 @@ class SignatureHelpTriggerKind:
 
 
 CodeLens = TypedDict('CodeLens', {
-    'range': Dict[str, Any],
+    'range': Range_T,
```
rchl, line 86: Maybe `RangeLsp` instead? The `T` is a common naming convention when dealing with generics but it's not the case here.

plugin/core/protocol.py
rwols: No mypy doesn't support recursive types:

```
JsonDict = Dict[str, Union[None, int, float, JsonList, JsonDict]
```
This doesn't work because JsonDict references itself again. https://github.com/python/mypy/issues/731

plugin/symbols.py
rchl: I've removed this line and that didn't seem to break either mypy or pyright.
At the same time, this line doesn't seem to help pyright at least in understanding the type of `children` (still shows as `List[Unknown]`.

What seems to work for pyright is using the type comment (`# type: List[DocumentSymbol]`).

plugin/core/protocol.py
predragnikolic: done 5f1116f6373c899615e556fdaed08ffd03e3e2f8

plugin/symbols.py
predragnikolic: You are right, 

I swear I saw an error :D but apparently the error is no longer there 

I will add the `# type: List[DocumentSymbol]`

predragnikolic: 16ebb7e68cd895a92922dc92444236aeaf76c0bd

rwols (approved):
Very nice.

PR #1542, opened 2021-01-10
Be explicit where new line are added

predragnikolic:
I think this is easier to grasp and resolves this question: https://github.com/sublimelsp/LSP/commit/a9406f55e775a0be50c9d3d252787c8ba3be0a7e#r45265031

I checked multiline diagnostics, seems to work fine:
image

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -449,18 +449,18 @@ class WindowManager(Manager):
             file_path = listener.view.file_name() or ""
             base_dir = self.get_project_path(file_path)  # What about different base dirs for multiple folders?
             file_path = os.path.relpath(file_path, base_dir) if base_dir else file_path
-            to_render.append("{}:".format(file_path))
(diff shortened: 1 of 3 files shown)


rchl:
This is a very trivial change and I'm fine with either but I think the previous way was cleaner. If it wasn't clear enough then maybe the fix could be to just rename the variable from `to_render` to `to_render_lines` or `lines_to_render`?

predragnikolic:
1) I find the previous code cleaner, but maybe a bit harder to understand.
2) this PR adds a bit more noise with the `\n` everywhere, but is more explicit.

I like 1 more, but 2 is _maybe_(?) better for other people getting started with the codebase. 
I am ok with getting this PR closed, if you think that the previous code is ok.

predragnikolic:
I would not waste energy your time on this PR. The code is pretty much the same with or without this PR 🙂

PR #1545, opened 2021-01-17
Handle basic editor commands

rwols:
After some discussion on Discord it seems it's safer to handle vscode-specific commands.

Most notably, now that we don't check if the server is a commandProvider due to handling client-side code lens commands, rust-analyzer will sometimes complain about not understanding workspace/executeCommand. Because it puts the command `editor.action.triggerParameterHints` in one of its completion items.

See: https://github.com/microsoft/language-server-protocol/issues/1148

### DIFF
--- a/plugin/execute_command.py
+++ b/plugin/execute_command.py
@@ -14,6 +14,11 @@ class LspExecuteCommand(LspTextCommand):
             command_args: Optional[List[Any]] = None,
             session_name: Optional[str] = None,
             event: Optional[dict] = None) -> None:
+        # Handle VSCode-specific command for triggering suggestions popup.
+        if command_name == "editor.action.triggerSuggest":
+            # Triggered from set_timeout as suggestions popup doesn't trigger otherwise.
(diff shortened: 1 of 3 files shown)


plugin/execute_command.py
```diff
@@ -14,6 +15,18 @@ def run(self,
             command_args: Optional[List[Any]] = None,
             session_name: Optional[str] = None,
             event: Optional[dict] = None) -> None:
+        # Handle VSCode-specific command for triggering AC/sighelp
+        if command_name == "editor.action.triggerSuggest":
+            # Triggered from set_timeout as suggestions popup doesn't trigger otherwise.
+            return sublime.set_timeout(lambda: self.view.run_command("auto_complete"))
+        if command_name == "editor.action.triggerParameterHints":
+
```
predragnikolic, line 23: I thought that we had a `lsp_show_signature` command so we could have just 
`return sublime.set_timeout(lambda: self.view.run_command("lsp_show_signature", {"manual": False}))`
but I guess we don't have such a command :)

predragnikolic (approved):

PR #1550, opened 2021-01-20
Add support for old plist package

deathaxe:


rwols:
Thanks!

deathaxe:
May there be still an issue with replacing 2nd level scopes? I just realize LSP still not to start LemMinX for plists with `text.plist` top-level scope.

rwols:
I think you would also have to add it here: https://github.com/sublimelsp/LSP-lemminx/blob/a0d3654b366411e3e421e0c64b520ea5b7d194a1/LSP-lemminx.sublime-settings#L9

using `text.xml | text.plist`. In fact you can replace that entire `"languages"` array with `"selector": "text.xml | text.plist"`.

rwols:
Example: https://github.com/sublimelsp/LSP/blob/671bcad540eb2be91f899d2fbf23b80618da6b94/LSP.sublime-settings#L210

PR #1555, opened 2021-01-24
Extend disabled_capabilities with more options

### LINKED ISSUE #1554: Option to disable code actions
Currently code actions can be shown as bulbs or annotations, but there doesn't seem to be a way to completely disable them. Would be nice to get such an option.

I wrote a very simple patch that prevents them from rendering when `"show_code_actions"` is not set to a known value (currently it falls back on `'annotation'` instead). But ended up not submitting it because ideally we shouldn't even calculate the code actions when they should be disabled. Unfortunately, tracking down every code path that invokes code actions has proven too much. Requesting support. 🙂

predragnikolic:
closes #1554

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -121,7 +121,7 @@ class CodeActionsManager:
         on_save_actions: Optional[Dict[str, bool]] = None
     ) -> Optional[CodeActionsCollector]:
         if 'codeActionProvider' in userprefs().disabled_capabilities:
-            return
+            return None
 
         use_cache = on_save_actions is None
         if use_cache:
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -79,7 +79,7 @@ class CodeActionsManager:
         region: sublime.Region,
         session_buffer_diagnostics: Sequence[Tuple[SessionBufferProtocol, Sequence[Diagnostic]]],
         actions_handler: Callable[[CodeActionsByConfigName], None]
-    ) -> CodeActionsCollector:
+    ) -> Optional[CodeActionsCollector]:
         """
         Requests code actions *only* for provided diagnostics. If session has no diagnostics then
         it will be skipped.
@@ -93,7 +93,7 @@ class CodeActionsManager:
         session_buffer_diagnostics: Sequence[Tuple[SessionBufferProtocol, Sequence[Diagnostic]]],
         actions_handler: Callable[[CodeActionsByConfigName], None],
         only_kinds: Optional[Dict[str, bool]] = None
-    ) -> CodeActionsCollector:
+    ) -> Optional[CodeActionsCollector]:
         """
         Requests code actions with provided diagnostics and specified region. If there are
         no diagnostics for given session, the request will be made with empty diagnostics list.
@@ -105,7 +105,7 @@ class CodeActionsManager:
         view: sublime.View,
         actions_handler: Callable[[CodeActionsByConfigName], None],
         on_save_actions: Dict[str, bool]
-    ) -> CodeActionsCollector:
+    ) -> Optional[CodeActionsCollector]:
         """
         Requests code actions on save.
         """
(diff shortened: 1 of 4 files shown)


plugin/code_actions.py
```diff
@@ -79,7 +79,7 @@ def request_with_diagnostics_async(
         region: sublime.Region,
         session_buffer_diagnostics: Sequence[Tuple[SessionBufferProtocol, Sequence[Diagnostic]]],
         actions_handler: Callable[[CodeActionsByConfigName], None]
-    ) -> CodeActionsCollector:
+    ) -> Optional[CodeActionsCollector]:
```
rwols, line 82: Since we never use the return value anyway, might as well return None?

plugin/code_actions.py
```diff
@@ -93,7 +93,7 @@ def request_for_region_async(
         session_buffer_diagnostics: Sequence[Tuple[SessionBufferProtocol, Sequence[Diagnostic]]],
         actions_handler: Callable[[CodeActionsByConfigName], None],
         only_kinds: Optional[Dict[str, bool]] = None
-    ) -> CodeActionsCollector:
+    ) -> Optional[CodeActionsCollector]:
```
rwols, line 96: Since we never use the return value anyway, might as well return None?

plugin/code_actions.py
```diff
@@ -105,7 +105,7 @@ def request_on_save(
         view: sublime.View,
         actions_handler: Callable[[CodeActionsByConfigName], None],
         on_save_actions: Dict[str, bool]
-    ) -> CodeActionsCollector:
+    ) -> Optional[CodeActionsCollector]:
```
rwols, line 108: Since we never use the return value anyway, might as well return None?

rwols (commented):
A nice addition :) I only have nitpicks.

plugin/code_actions.py
predragnikolic: True

predragnikolic:
To be honest, I don't know how this affects code actions on save. (I didn't test that)

rchl:
> To be honest, I don't know how this affects code actions on save. (I didn't test that)
> 
> I will try to test that in the coming days.

Looking at the code, it seems like the save action would just get stuck and only trigger actual save after the 2s timeout.

rwols:
So it might need an extra check here?

https://github.com/sublimelsp/LSP/blob/6b5da033bb258ce720cb2092de41e38d0dbadf01/plugin/code_actions.py#L223-L225

rwols:
Wouldn't the previous code also get stuck if you had set `"lsp_code_actions_on_save"` in the settings but there was no codeActionProvider then?

rchl:
> So it might need an extra check here?

Potentially yes.
I think that ideally all those checks would be done earlier than done in this PR so that none of that request code is triggered when the code action provider is disabled globally. With this change also the callback (`actions_handler`) is not called so some code could potentially get stuck forever waiting for a reply (but as far as I checked, all of the current code actually should be fine).

> Wouldn't the previous code also get stuck if you had set `"lsp_code_actions_on_save"` in the settings but there was no codeActionProvider then?

Shouldn't as `CodeActionsCollector` is a context manager and should trigger the callback also if there are no sessions with codeActionProvider.

plugin/code_actions.py
```diff
@@ -119,7 +119,10 @@ def _request_async(
         only_with_diagnostics: bool,
         actions_handler: Callable[[CodeActionsByConfigName], None],
         on_save_actions: Optional[Dict[str, bool]] = None
-    ) -> CodeActionsCollector:
+    ) -> None:
+        if 'codeActionProvider' in userprefs().disabled_capabilities:
+            return None
```
rchl, line 124: If we keep this logic in here then at least `actions_handler` should be called with empty results instead of returning `None`. (Also can be just `return` instead of `return None`)

rchl: And BTW. The callback is called on async thread so we should be consistent here.

rwols: Good point about actions_handler. Shows another useful property of promises/futures as you wouldn't forget to return a resolved promise due to the type system.

plugin/code_actions.py
```diff
@@ -128,7 +131,6 @@ def _request_async(
                 cache_key, cache_collector = self._response_cache
                 if location_cache_key == cache_key:
                     sublime.set_timeout(lambda: actions_handler(cache_collector.get_actions()))
-                    return cache_collector
```
rchl, line 131: This return should not have been removed as it changes the logic.

predragnikolic: I will make a PR tonight that will hopefully solve each of the comments above  :)

PR #1557, opened 2021-01-24
Always look up the session name for lsp_execute

### LINKED ISSUE #1556: [regression] lsp_execute does nothing due to empty session
Since this [commit](https://github.com/sublimelsp/LSP/commit/7d05794fa3cc4ecd3931d09a90e801addc70d9fa) the `capability` variable got deleted which means that `self.best_session(self.capability)` is unable to find session.

The consequence is that [LSP-metals.sublime-commands](https://github.com/scalameta/metals-sublime/blob/master/LSP-metals.sublime-commands) aren't executed.

rwols:
The `self.capability` used to be an empty string, which meant that
the check for `capability is None` was always false when looking up
suitable sessions.

Resolves #1556

### DIFF
--- a/plugin/execute_command.py
+++ b/plugin/execute_command.py
@@ -27,7 +27,7 @@ class LspExecuteCommand(LspTextCommand):
                     listener.do_signature_help_async(manual=False)
 
             return sublime.set_timeout_async(run_async)
-        session = self.session_by_name(session_name) if session_name else self.best_session(self.capability)
+        session = self.session_by_name(session_name if session_name else self.session_name)
         if session and command_name:
             if command_args:
                 self._expand_variables(command_args)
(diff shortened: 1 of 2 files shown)


rwols:
On second thought this doesn't make sense... helper packages should use session_name otherwise you might end up running the keybinding for a second session that happens to be running for the view.

plugin/execute_command.py
```diff
@@ -27,7 +27,7 @@ def run_async() -> None:
                     listener.do_signature_help_async(manual=False)
 
             return sublime.set_timeout_async(run_async)
-        session = self.session_by_name(session_name) if session_name else self.best_session(self.capability)
+        session = self.session_by_name(session_name if session_name else self.session_name)
```
predragnikolic, line 30: Should we also have a check later if the session has a capability to execute the command?
```
session.has_capability('executeCommandProvider')
```

rwols: Let's do that when someone complains about it.

rchl, line 30: In other words `session_name or self.session_name`.

PR #1559, opened 2021-01-25
Fix issue with code action popup re-appearing after selecting item

rchl:
And call the callback when codeActionProvider is disabled.

This fixes more serious issues.
Whether the check for `codeActionProvider` should be implemented at a higher level is another matter.

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -121,7 +121,8 @@ class CodeActionsManager:
         on_save_actions: Optional[Dict[str, bool]] = None
     ) -> None:
         if 'codeActionProvider' in userprefs().disabled_capabilities:
-            return None
+            sublime.set_timeout_async(lambda: actions_handler({}))
+            return
 
         use_cache = on_save_actions is None
         if use_cache:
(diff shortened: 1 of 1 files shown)


rwols (approved):

plugin/code_actions.py
```diff
@@ -131,6 +132,7 @@ def _request_async(
                 cache_key, cache_collector = self._response_cache
                 if location_cache_key == cache_key:
                     sublime.set_timeout(lambda: actions_handler(cache_collector.get_actions()))
```
rwols, line 134: Wouldn't this also need to be `set_timeout_async`?

predragnikolic:
I was working on reverting some changes that I introduced lately.

I moved the if checks at a higher level,
and reverted all changes that I made to the _request_async method. 
Does this look good? https://github.com/sublimelsp/LSP/commit/10cbc775e49fb8281f39e8805bd0a2443ee0774c

rchl:
> Does this look good? [10cbc77](https://github.com/sublimelsp/LSP/commit/10cbc775e49fb8281f39e8805bd0a2443ee0774c)

 - you should rebase or otherwise your changes will conflict with #1559
 - no need to revert the return type change as the return value is not currently used
 - should `self._clear_code_actions_annotation()` always be called so that annotations are cleared in case the user disables codeactionProvider at runtime?
 - as for the placement of those checks, I don't have time now to think about that. Probably best to create PR.

PR #1567, opened 2021-02-03
Fix: '*InputHandler' object has no attribute 'want_event'

predragnikolic:
I don't know if we should wait for the next ST release that will fix this, 
but here is a fix for the `'*InputHandler' object has no attribute 'want_event'` error.


rwols (approved):
Let's roll with this.

Issue #1549, opened 2021-01-20
Option to change where type hinting is displayed in the suggestion window


dvcrn:
Similar to how vscode and other editors are inlining the type hints:

<img width="594" alt="Screen_Shot_2021-01-14_at_16 29 08" src="https://user-images.githubusercontent.com/688326/105137423-9b7db300-5b36-11eb-9b37-0f6f71312a2f.png">

While LSP does it this way:
<img width="380" alt="Screen_Shot_2021-01-14_at_16 28 26" src="https://user-images.githubusercontent.com/688326/105137474-b3553700-5b36-11eb-9cb2-5555ea55dec7.png">

Other plugins that integrate with sublimes autocomplete, for example sublime-go and snippet related extensions render it on the right as well, so when these results are combined with LSP results, it gets very confusing on where to look because LSP puts it at the bottom:
<img width="661" alt="Screen_Shot_2021-01-14_at_16 32 28" src="https://user-images.githubusercontent.com/688326/105137531-ca942480-5b36-11eb-907d-2ba0d306085f.png">

Because it's all a bit over the place, the results from LSP feel like they aren't actually relevant and just some random word completions, because on first glance they lack type information if you don't know where to look. 

It would be nice if the location where this is displayed is settable from withing sublime-LSP settings. 

I'm happy to help with adding support of this if it can get considered if you have any pointers which parts to touch :)

rchl:
We should even consider changing the default.

We'd just need to check various servers to see if this gives any unexpected results.

rwols:
All you would have to do is play around with the st_details and st_annotation variables here:

https://github.com/sublimelsp/LSP/blob/6c2521ba17c2ae57d3dc84233b2cfeed0a42f71e/plugin/core/views.py#L667-L679

dvcrn:
Thanks for the pointer. I put `lsp_detail` into the annotation and it's working perfectly for what I wanted:

<img width="801" alt="Screen Shot 2021-01-22 at 15 32 11" src="https://user-images.githubusercontent.com/688326/105455545-04943080-5cc7-11eb-818e-8897410c576a.png">

rwols:
So will you make a pull request?

dvcrn:
I can do that! I thought it was still under consideration if this project wants to change the default

rwols:
> I thought it was still under consideration if this project wants to change the default

Let's just see what you got in a pull request so we can all discuss it :)

dvcrn:
[...]

PR #1569, opened 2021-02-05
Move lsp detail into autocomplete annotation 

dvcrn:
Move `item.get("detail")` into st_annotation to effectively inline typehints 

Tested so far only with Golang 

References #1549

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -662,21 +662,19 @@ def format_completion(
 
     lsp_label = item["label"]
     lsp_filter_text = item.get("filterText")
-    lsp_detail = html.escape(item.get("detail") or "").replace('\n', ' ')
-
-    if lsp_filter_text and lsp_filter_text != lsp_label:
-        st_trigger = lsp_filter_text
-    else:
-        st_trigger = lsp_label
-
-    st_annotation = item.get("detail") or ""
+    st_annotation = (item.get("detail") or "").replace('\n', ' ')
 
     st_details = ""
     if can_resolve_completion_items or item.get("documentation"):
         st_details += make_command_link("lsp_resolve_docs", "More", {"index": index})
-        st_details += " | " if lsp_detail else ""
 
-    st_details += "<p>{}</p>".format(lsp_detail)
+    if lsp_filter_text and lsp_filter_text != lsp_label:
+        st_trigger = lsp_filter_text
+        if st_details:
+            st_details += " | "
+        st_details += "<p>{}</p>".format(html.escape(lsp_label))
+    else:
+        st_trigger = lsp_label
 
     # NOTE: Some servers return "textEdit": null. We have to check if it's truthy.
     if item.get("textEdit") or item.get("additionalTextEdits") or item.get("command"):
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -671,6 +671,8 @@ def format_completion(
         st_trigger = lsp_label
         st_annotation = ""
 
+    st_annotation = item.get("detail") or ""
+
     st_details = ""
     if can_resolve_completion_items or item.get("documentation"):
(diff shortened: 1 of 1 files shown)


rchl:
There is no need to have a duplicate text in the "detail" field (at the bottom of the popup). The logic should probably be to flip those values.

rwols:
For clangd, this results in:

**before**:
Schermafbeelding 2021-02-05 om 20 58 42

**after**:
Schermafbeelding 2021-02-05 om 21 00 36

rwols:
So the `label` is now "forgotten" with this PR. Perhaps the label can go into the st_details field?

dvcrn:
> For clangd, this results in:
> 
> **before**:
> Schermafbeelding 2021-02-05 om 20 58 42
> 
> **after**:
> Schermafbeelding 2021-02-05 om 21 00 36

Hrrm I wonder what a good solution for this is. Either come up with some logic to pick the right thing, or maybe have it settable per lsp?

rwols:
I think having the lsp_label in the st_details slot should be sufficient. Try this?

```diff
diff --git a/plugin/core/views.py b/plugin/core/views.py
index ac54b66..de70476 100644
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -662,21 +662,19 @@ def format_completion(
 
     lsp_label = item["label"]
     lsp_filter_text = item.get("filterText")
-    lsp_detail = html.escape(item.get("detail") or "").replace('\n', ' ')
-
-    if lsp_filter_text and lsp_filter_text != lsp_label:
-        st_trigger = lsp_filter_text
-    else:
-        st_trigger = lsp_label
-
-    st_annotation = item.get("detail") or ""
+    st_annotation = (item.get("detail") or "").replace('\n', ' ')
 
     st_details = ""
     if can_resolve_completion_items or item.get("documentation"):
         st_details += make_command_link("lsp_resolve_docs", "More", {"index": index})
-        st_details += " | " if lsp_detail else ""
 
-    st_details += "<p>{}</p>".format(lsp_detail)
+    if lsp_filter_text and lsp_filter_text != lsp_label:
+        st_trigger = lsp_filter_text
+        if st_details:
+            st_details += " | "
+        st_details += "<p>{}</p>".format(html.escape(lsp_label))
+    else:
+        st_trigger = lsp_label
 
     # NOTE: Some servers return "textEdit": null. We have to check if it's truthy.
     if item.get("textEdit") or item.get("additionalTextEdits") or item.get("command"):
```

dvcrn:
Sorry I didn't have time on hand yet to try this. I'll be getting back to this on the weekend

predragnikolic:
Keep in mind that `CompletionItem.detail` is optional.

> By default the request can only delay the computation of the detail and documentation properties.

Which means that there are probably servers that will return the `detail` only when a completion item is resolved.

dvcrn:
I applied the patch from rwols and wanted to test this a bit. I noticed that on tyescript/javascript the support is still lacking a good bit behind vscode, but that's probably unrelated to this because the information is neither in label nor detail 

ST3:
Screen Shot 2021-02-15 at 12 50 44
Screen Shot 2021-02-15 at 12 50 50



Vscode:
Screen Shot 2021-02-15 at 12 52 07

@rwols for clangd, how is the support with the latest patch that you provided?

rwols:
I totally missed that you wrote a reply, sorry for the delay.

> for clangd, how is the support with the latest patch that you provided?

It looks okay. I think we should move to putting the lsp_label in the st_detail slot anyway, because we can use minihtml in the st_detail slot. That means that we can add coloring for https://github.com/sublimelsp/LSP/issues/1570.

> Which means that there are probably servers that will return the detail only when a completion item is resolved.

True. But the lsp_details are then shown with the "More" button.

rwols (approved):
Let's try this out. It's a tiny diff so it's easy to blame/revert anyway.

PR #1571, opened 2021-02-10
Fix issue with diagnostics regions sometimes not being removed on undo

### LINKED ISSUE #1218: Diagnostics sync issue related to using undo
LSP 4070-1.0.6 (or thereabouts)

1. Create an issue in the code (I suggest to add a space where there shouldn't be one) to make diagnostic show up
2. If space was added, remove that space with a backspace
3. Very quickly press the shortcut for undo and immediately backspace to delete the space again

Expect: Diagnostic (added by undo) disappears
Actual: Diagnostic stays visible even though there are no diagnostic errors in the status bar or diagnostics panel.

What happens is that ST restores the error/warning diagnostic region on undoing and then, due to an unfortunate series of events and things being async, we fail to call `view.erase_regions`. And our model gets out of sync with the view.

This reproduces even with `diagnostics_delay_ms` set to 0 but it's a bit easier with higher value.

rchl:
The problem stems from the fact that ST restores regions on Undo so in
a situation when there is a diagnostic shown:

```
let a
     ^ missing semicolon
```

then semicolon is added:

```
let a;
```

and immediately (before the diagnostic is removed) the document is
changed again by for example adding something before semicolon:

```
let aa;
```

then LSP will remove the diagnostic (region) but pressing undo will
add it back (after the semicolon):

```
let a;
      ^ missing semi
```

At this point, our logic decides that there are no diagnostics visible and
we've just received empty diagnostics message so we don't need to do
anything while in fact, we do have to remove the region.

Fixes #1218

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -957,11 +957,6 @@ class Session(TransportCallbacks):
         if mgr:
             getattr(mgr, method)(*args)
 
-    def clear_diagnostics_async(self) -> None:
-        # XXX: Remove this functionality?
-        for sb in self.session_buffers_async():
-            sb.on_diagnostics_async([], None)
-
     def on_stderr_message(self, message: str) -> None:
         self.call_manager('handle_stderr_log', self, message)
         self._logger.stderr_message(message)
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -302,10 +302,7 @@ class SessionBuffer:
 
         self.diagnostics_debouncer.cancel_pending()
 
-        if not bool(diagnostics) and not self.diagnostics_are_visible:
-            # Nothing was previously visible, and nothing will become visible. So do nothing.
-            pass
-        elif self.diagnostics_are_visible:
+        if self.diagnostics_are_visible:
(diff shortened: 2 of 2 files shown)


rchl:
This fix is for the issue I was able to reproduce on demand, with out-of-sync regions. The actual diagnostics reported were correct. But I'm pretty sure there also exists a similar issue with actual diagnostics getting out of sync which I don't know how to reproduce but it could also get fixed by this fix.

Of course, this change adds some overhead so I was thinking to instead of removed code have a condition like "if there are diagnostic regions and received empty diagnostics" then "present()" which could be better for performance reason as I'm assuming that checking for regions is just a map lookup in ST. I feel like that solution would seem a bit "dirty" though.

predragnikolic:
I tried to reproduce the issue with the steps you provided, but I didn't succeed 

> and immediately (before the diagnostic is removed) the document is
changed again by for example adding something before semicolon:

I am not fast enough :)

rchl:
You can try utilizing code-actions-on-save to automatically add a semicolon and then immediately remove a character.

rwols (approved):
I'll take your word for it :)

PR #1574, opened 2021-02-11
Change popup style

### LINKED ISSUE #1566: Make header font-size same as text in popups
I do not find it very useful for the header text to be this big in popups. Just being bold makes them stand out just enough.

image

Reduce the font size of h1, h2, h3, h4, h5, h6 elements to be the same as the rest of the text.

predragnikolic:
closes #1566 

This PR makes all the font-size of header elements the same as the other text(1rem).
But header elements are bold, which makes them standout just enough from the rest of the text.
 
Header font size before:
headers

Header font size after:
the smallest

----

I also reduced the `hr` margin to be consistent with the popup margin spacing (0.5)

Before:
big

After
smaller

### DIFF
--- a/popups.css
+++ b/popups.css
@@ -6,6 +6,8 @@
     margin-top: 0.5rem;
 }
 .lsp_popup hr {
+    margin-top: 0.5rem;
+    margin-bottom: 0.5rem;
     border-color: color(var(--foreground) alpha(0.10));
 }
 .highlight {
--- a/popups.css
+++ b/popups.css
@@ -10,12 +10,14 @@
     margin-bottom: 0.5rem;
     border-color: color(var(--foreground) alpha(0.10));
 }
-.lsp_popup h1 { font-size: 1rem; }
-.lsp_popup h2 { font-size: 1rem; }
-.lsp_popup h3 { font-size: 1rem; }
-.lsp_popup h4 { font-size: 1rem; }
-.lsp_popup h5 { font-size: 1rem; }
-.lsp_popup h6 { font-size: 1rem; }
+.lsp_popup h1,
+.lsp_popup h2,
+.lsp_popup h3,
+.lsp_popup h4,
+.lsp_popup h5,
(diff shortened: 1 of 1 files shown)


rwols (approved):
Looks good to me :)

predragnikolic:
There was a comment from Rafal :)

> I would think that using the same size for all of them could make the content harder to read in case it's structured with multiple different heading levels. - https://github.com/sublimelsp/LSP/issues/1566#issuecomment-772839351

Having different sizes for different headers  makes sense for a web page, 
but I don't think that having different font-sizes for header would improve readability, because of the limited space in the popup.

rchl:
I'm fine with trying it but would it make sense to at least make them a little bit bigger than normal text? Maybe 1.1rem for all?

(I can't currently find any content I could test this with so maybe it's not that important either way ;))

rchl (approved):

popups.css
```diff
@@ -6,8 +6,16 @@
...
+    margin-top: 0.5rem;
+    margin-bottom: 0.5rem;
     border-color: color(var(--foreground) alpha(0.10));
 }
+.lsp_popup h1 { font-size: 1rem; }
+.lsp_popup h2 { font-size: 1rem; }
+.lsp_popup h3 { font-size: 1rem; }
+.lsp_popup h4 { font-size: 1rem; }
+.lsp_popup h5 { font-size: 1rem; }
+.lsp_popup h6 { font-size: 1rem; }
```
rchl, line 18: Avoid repeating yourself
```suggestion
.lsp_popup h1,
.lsp_popup h2,
.lsp_popup h3,
.lsp_popup h4,
.lsp_popup h5,
.lsp_popup h6 {
    font-size: 1rem;
}
```

predragnikolic:
There is really a small difference between 1.1rem and 1rem.
<details>

smallerheader

the smallest
</details>
I initially  chose 1rem because a little more text is visible :)

PR #1575, opened 2021-02-11
Handle moving views across windows

rwols:
Basically the old state is cleared and we pretend we are dealing with a fresh
new ViewEventListener again. After thinking about this, it feels like ST should
do this as well: delete the old ViewEventListener, and create a fresh new
ViewEventListener for the new window. Because the new window may have other
settings applied to all views due to project settings, the ViewEventListener's
is_applicable classmethod may not be applicable anymore for the new window. In
any case, for now we'll have to use on_pre_move and on_post_move.


boot.py
```diff
@@ -217,6 +217,20 @@ def on_new_window_async(self, w: sublime.Window) -> None:
...
+            return
+        sublime.set_timeout_async(listener.on_pre_move_window_async)
+
+    def on_post_move(self, view: sublime.View) -> None:
+        # Note: I'm afraid to use on_post_move_async because I don't know for sure whether that will run *after* the
+        # sublime.set_timeout_async in on_pre_move
+        listener = windows.listener_for_view(view)
+        if not listener:
+            return
+        sublime.set_timeout_async(listener.on_post_move_window_async)
```
rchl, line 232: Those are also called when dragging tabs within the same window (even the same group).

Not sure if optimization that checks if the window was changed is worth it...
Or maybe it should be reported as a bug that it fires even within the same window?

plugin/documents.py
```diff
@@ -152,6 +158,17 @@ def __del__(self) -> None:
...
+    def on_pre_move_window_async(self) -> None:
+        if self._registered and self._manager:
+            self._manager.unregister_listener_async(self)
+            self._registered = False
+            self._manager = None
+        self._cleanup()
+
+    def on_post_move_window_async(self) -> None:
+        self._setup()
+        self.on_activated_async()
```
rchl, line 170: It appears that if I drag a tab to a different position within the same window then LSP stops working for that view (no LSP messages trigger anymore).

rwols:
Some good remarks. I'll have to look at this another time.

boot.py
rwols: We're checking if the window IDs are the same and bail out if so.

plugin/documents.py
rwols: this should work now

rchl:
It behaves a bit weird after the file is moved to another window.

- The new window doesn't have the LSP log panel available.
- The error diagnostics that were visible before the view was moved no longer react to hover (no popup is shown)

rwols:
I cannot re-request a review, so I'll do it this way: ping @rchl :)

plugin/session_view.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, listener: AbstractViewListener, session: Session) -> None:
         self.progress = {}  # type: Dict[int, ViewProgressReporter]
         settings = self.view.settings()
         buffer_id = self.view.buffer_id()
-        key = (session.config.name, buffer_id)
+        key = (session.config.name, session.window.id(), buffer_id)
```
predragnikolic, line 39: what bad behavior can happen if we don't pass the window.id to the key?

I saw the commit description(https://github.com/sublimelsp/LSP/pull/1575/commits/349d76439acefd2d7585b75f8311d088337bd5cf), but I still don't understand why.

rwols: Because of [this variable](https://github.com/sublimelsp/LSP/blob/c0ae90fb3cbcc6e07ba1ce2b76ba8aedc600df73/plugin/hover.py#L78), the buffers are kept alive until the next time the `run` method is invoked.

Now, when we reset the view event listener due to a moved window, we create a new SessionView. That SessionView then looks up its SessionBuffer in the weak map. It finds that SessionBuffer in the weak map, because it's still alive due to the hover command data, and the session name and buffer ID are the same. So we have to key on the window ID as well.

PR #1580, opened 2021-02-13
Account for font size for popup width

### LINKED ISSUE #1578: mdpopup popups are rendered tiny
**Describe the bug**
I use `lsp_hover` function very often with my Rust development. It shows a popup window which has super small max_width and max_height properties. By default (an example "Dispatch" type) it looks like this: http://s.verknowsys.com/80a79f0d3b423fd7a7fdd335e15b9a4a.png

I altered User/mdpopups.css to decrease default font sizes, but I can't alter the width or height of the popup window.

Workaround:

edit `Packages/mdpopups/st3/mdpopups/__init__.py` file on line 610, and change "max_width=320, max_height=240," to some bigger numbers. I did "max_width=2160, max_height=2160,", but after ST4 restart I still have the width limited. Looks like this: http://s.verknowsys.com/4de808d955af370e9da630125f7c1295.png
It's better, but still width is not 2160 max.

**To Reproduce**
1. Latest ST4 build (b4098)
2. Use Rust Enhanced with rust-analyzer
3. Hover mouse on fern Dispatch type (or any well documented type?)
4. See small popup

**Expected behavior**
Popup should get as large as screen or at least as large as ST4 workspace/ canvas.

**Screenshots**
Yes, above.

- OS: macOS 11.2.1
- Sublime Text version: [build 4098]
- LSP version: 1.2.10 (latest)
- Language servers used: rust-analyzer (last non-nightly build available)

**Additional context**
None

rwols:
I can't actually find in the docs what "layout space" is, but from some
experimentation it seems to be the extent of the window minus the sidebar and
minus the minimap.

Resolves #1578

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -372,7 +372,7 @@ def show_lsp_popup(view: sublime.View, contents: str, location: int = -1, md: bo
         flags=flags,
         location=location,
         wrapper_class=wrapper_class,
-        max_width=view.viewport_extent()[0],
+        max_width=int(view.em_width() * 80.0),
         max_height=1000000,
         on_navigate=on_navigate)
 
@@ -395,6 +395,7 @@ def minihtml(view: sublime.View, content: Union[str, Dict[str, str], list], allo
     Formats provided input content into markup accepted by minihtml.
 
     Content can be in one of those formats:
+
      - string: treated as plain text
      - MarkedString: string or { language: string; value: string }
      - MarkedString[]
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -372,7 +372,7 @@ def show_lsp_popup(view: sublime.View, contents: str, location: int = -1, md: bo
         flags=flags,
         location=location,
         wrapper_class=wrapper_class,
-        max_width=int(view.em_width() * 80.0),
+        max_width=int(view.em_width() * 120.0),  # Around 120 characters per line
         max_height=1000000,
         on_navigate=on_navigate)
 
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -372,7 +372,7 @@ def show_lsp_popup(view: sublime.View, contents: str, location: int = -1, md: bo
         flags=flags,
         location=location,
         wrapper_class=wrapper_class,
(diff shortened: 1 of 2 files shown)


rchl:
> it seems to be the extent of the window minus the sidebar and
> minus the minimap.

Checking how it works, it looks like it depends on the longest line in the view so IMO it's not that great.

For example, if you have only `aaa` content in the view, the `layout_extent()[0]` will be around 21 pixels.

rchl:
Two things:
 - `viewport_extent` is limited to the view's width which could be unnecessarily limiting when having a layout with multiple columns
 - With a really big window width, it will look weird and be less practical to have a very wide and short popup because long lines are not that comfortable to read.

Maybe we can at least do the `min(viewport_extend_width, 800)` (or whatever arbitrary number we choose) logic?

And the original issue seems related to user settings or just scaling in 4k monitor and could be solved by just making sure that the current 800 min-width looks like actually 800 and not 400.

rchl:
I'm gonna run with it for a while.

plugin/core/views.py
```diff
@@ -373,8 +372,8 @@ def show_lsp_popup(view: sublime.View, contents: str, location: int = -1, md: bo
         flags=flags,
         location=location,
         wrapper_class=wrapper_class,
-        max_width=max_width,
-        max_height=max_height,
+        max_width=int(view.em_width() * 80.0),
```
rchl, line 375: `int(view.em_width() * 80.0)` is 575 for me so even less than previous default.

rwols: Yes, but this seems to be about 80 characters wide, which is a good width for reading text.

rchl: On the other hand, we use 120 characters for this codebase for example and the popup not only shows normal text but also the code. So maybe 120 would work?

rwols: @dmilith informed me on Discord that his font size is set to 29 and his em_width() is 14.5. So that would result in a width of 120chars * 14.5px/chars = 1740px, which is about 45% of the total width of a 4K resolution screen. Sounds good to me.

rchl (approved):

PR #1582, opened 2021-02-14
Send completion request to multiple sessions

predragnikolic:
resolves https://github.com/sublimelsp/LSP-stylelint/issues/11

Added:
- send completion request to all sessions that has the completion capability
- added LSP protocol types for `CompletionList` and `CompletionItem`

Fixed:
- fixed a failing test(that test was failing even before this PR)

I am in the process of testing this, but while I test, here is the code. :)


plugin/documents.py
```diff
@@ -606,46 +614,70 @@ def render_highlights_on_main_thread() -> None:
...
+        sublime.status_message('Completion error: ' + str(error.get('message')))
+
+    def _on_all_settled(
+        self,
+        responses: Optional[List[CompletionResponseWithSession]],
+        resolve_completion_list: ResolveCompletionsFn
+    ) -> None:
+        if not responses:
+            resolve_completion_list([], 0)
+            return
```
predragnikolic, line 655: The generic type T in the `then` call is wrapped as `Optional[T]`
>    def then(self, onfullfilled: FullfillFunc**[Optional[T]**, TResult]) -> 'Promise[TResult]': 
 https://github.com/sublimelsp/LSP/blob/st4000-exploration/plugin/core/promise.py#L160

Maybe it is not necessary the type to be optional, and could be just `T`, 
>    def then(self, onfullfilled: FullfillFunc[**T**, TResult]) -> 'Promise[TResult]': 

If the type wasn't optional I could delete this if check above. (I have a feeling that the if check is not necessary, it will always be an array of resolved promise values)

@rchl do you think that we can remove the `Optional[T]`  type and just leave `T`? (I would maybe not do it in this PR, this was just something that I have noticed)

plugin/core/protocol.py
```diff
@@ -169,6 +167,42 @@ class SignatureHelpTriggerKind:
...
+    'filterText': Optional[str],
+    'insertText': Optional[str],
+    'insertTextFormat': Optional[int],
+    'insertTextMode': Optional[int],
+    'textEdit': Optional[Union[TextEdit, InsertReplaceEdit]],
+    'additionalTextEdits': Optional[List[TextEdit]],
+    'commitCharacters': Optional[List[str]],
+    'command': Optional[Command],
+    'data': Any
+}, total=True)
```
predragnikolic, line 199: I misunderstand the meaning of total in typed dicts, i will change it to false.

plugin/documents.py
```diff
@@ -606,46 +614,70 @@ def render_highlights_on_main_thread() -> None:
...
-    def _on_complete_result(self, response: Optional[Union[dict, List]], resolve: ResolveCompletionsFn,
-                            can_resolve_completion_items: bool, session_name: str) -> None:
-        response_items = []  # type: List[Dict]
-        flags = 0
+
+        completion_promises = []  # type: List[Promise[CompletionResponseWithSession]]
+        for session in sessions:
+
+            def completion_request() -> Promise:
+                return Promise(lambda resolve: session.send_request_async(
```
rwols, line 628: Can't you use `session.send_request_task` here?

plugin/core/protocol.py
```diff
@@ -169,6 +167,42 @@ class SignatureHelpTriggerKind:
...
+    'insert': RangeLsp,
+    'newText': str
+}, total=True)
+
+CompletionItem = TypedDict('CompletionItem', {
+    'label': str,
+    'kind': Optional[int],
+    'tags': Optional[List[int]],
+    'detail': Optional[str],
+    'documentation': Optional[str],
```
rchl, line 186: str | MarkupContent

plugin/core/views.py
```diff
@@ -1,5 +1,5 @@
 from .css import css as lsp_css
-from .protocol import CompletionItemTag
+from .protocol import CompletionItem, CompletionItemTag
```
rchl, line 2: Separate lines? :)

plugin/documents.py
```diff
@@ -606,46 +614,70 @@ def render_highlights_on_main_thread() -> None:
...
+                    Request.complete(text_document_position_params(self.view, location), self.view),
+                    lambda res: self._on_complete_result(res, resolve, session),
+                    lambda res: self._on_complete_error(res, resolve, session)))
```
rchl, line 631: Passing the session is a bit unsafe since it could in theory die during request.
You could resolve `can_resolve_completion_items` right away here.
But then you would also need to pass the session name as it's needed in `format_completion` later.

I'm fine with being overridden here as it's probably a fairly low chance of this being a problem and we are probably doing that in other places already.

plugin/documents.py
```diff
@@ -606,46 +614,70 @@ def render_highlights_on_main_thread() -> None:
...
-        response_items = []  # type: List[Dict]
-        flags = 0
+
+        completion_promises = []  # type: List[Promise[CompletionResponseWithSession]]
+        for session in sessions:
+
+            def completion_request() -> Promise:
+                return Promise(lambda resolve: session.send_request_async(
+                    Request.complete(text_document_position_params(self.view, location), self.view),
+                    lambda res: self._on_complete_result(res, resolve, session),
```
rchl, line 630: Just a suggestion to remove `_on_complete_result` and just resolve here directly:

```suggestion
                    lambda res: resolve(res, session),
```

(although if you will follow my other comment then you'll have to pass more arguments).

plugin/documents.py
rchl: Changing types in the `Promise` can have some unexpected cascading effects but I can try. :)

plugin/core/protocol.py
predragnikolic: f8dc0ea

plugin/core/views.py
predragnikolic: c1d1948

plugin/documents.py
predragnikolic: I will try to pass only the session_name, and later retrieve the session by the session_name.
I would want to pass as little parameters as possible :)

plugin/documents.py
predragnikolic: I like this :)

plugin/documents.py
predragnikolic: I wasn't aware of the presence of `send_request_task` :)

But looking at the implementation of `send_request_task` I am not sure how I would pass in the session_name. I see that I can only pass in the `Request`

https://github.com/sublimelsp/LSP/pull/1582/files#diff-982c110867d9809c978a5597e4da66ce7538b327675a66b8f27214413b7c0ec1R630

Can I do something like (I haven't tried this, will try it tomorrow)
```python
        completion_promises = []  # type: List[Promise[CompletionResponseWithSession]]
        for session in sessions:

            def completion_request() -> Promise:
                return session.send_request_task(
                    Request.complete(text_document_position_params(self.view, location), self.view)
                ).then(lambda res: (res, session.config.name))

            completion_promises.append(completion_request())
```

predragnikolic:
I found one issue today.

LSP sends 2 completion request, 
and one server decides to not respond to a completion request
but rather it responds with a  custom request.

```
:: --> LSP-tailwindcss textDocument/completion(9): {'textDocument': {'uri': 'file:///Users/predrag/digilab-project-labena-base-frontend/src/App.vue'}, 'position': {'line': 9, 'character': 0}}
:: --> LSP-vue textDocument/completion(16): {'textDocument': {'uri': 'file:///Users/predrag/digilab-project-labena-base-frontend/src/App.vue'}, 'position': {'line': 9, 'character': 0}}
:: <?  LSP-tailwindcss tailwindcss/getConfiguration: {'languageId': 'vue', '_id': '24315d17375fde63b524046cdf5c96d9'}  // expected completion response, but got this
:: <<< LSP-vue 16: {'items': [...], 'isIncomplete': False}
```

I expected the tailwind css language server to return a response for the completions, but it returned `tailwindcss/getConfiguration`.

The end result is that even if the vue server returned completion items, we never show the the AC,
because the `_on_complete_result` and `_on_complete_error` are not called for the request made by tailwind session.

> note to reproduce this issue I had to trigger the AC in a scope outside of the attribute `class`
```
// App.vue
<template>
	<div id="app">
		<div class=""></div>
  	</div>
</template>
<script>
| // AC was triggered here

</script>
``` 

I will investigate more

rchl:
> I expected the tailwind css language server to return a response for the completions, but it returned `tailwindcss/getConfiguration`.

It's not that it responded to the completion request with that. It just triggered this separate request while handling the completion request. That should prevent completion result from coming after that. But I can see how failing to handle the custom request can "soft-crash" the server and not be able to return completions. I think it's a pure server (plugin helper) issue and not really related to this change.

Unless we would want to handle something like that with a timeout but I don't think we should.

predragnikolic:
> I think it's a pure server (plugin helper) issue and not really related to this change.
> Unless we would want to handle something like that with a timeout but I don't think we should.

Yes, I would avoid adding a timeout.

plugin/documents.py
predragnikolic: 64e3e29532f5bf8c3bab28457de52fd4ca023273

plugin/documents.py
predragnikolic: https://github.com/sublimelsp/LSP/pull/1582/commits/14c0e41f02c480352ffed4d54dc6c1d6ad07b8d3

plugin/core/registry.py
```diff
@@ -29,6 +29,17 @@ def best_session(view: sublime.View, sessions: Iterable[Session], point: Optiona
...
 
+def sessions(view: sublime.View, capability: Optional[str] = None) -> Generator[Session, None, None]:
+    yield from sessions_for_view(view, capability)
+
+
+def session_by_name(view: sublime.View, session_name: str) -> Optional[Session]:
+    for session in sessions(view):
+        if session.config.name == session_name:
+            return session
+    return None
```
rwols, line 40: I'm aiming to remove these free functions as they can be abused easily. Can you instead add them to LspTextCommand as methods?

plugin/core/protocol.py
```diff
@@ -169,6 +171,42 @@ class SignatureHelpTriggerKind:
...
+    'preselect': Optional[bool],
+    'sortText': Optional[str],
+    'filterText': Optional[str],
+    'insertText': Optional[str],
+    'insertTextFormat': Optional[int],
+    'insertTextMode': Optional[int],
+    'textEdit': Optional[Union[TextEdit, InsertReplaceEdit]],
+    'additionalTextEdits': Optional[List[TextEdit]],
+    'commitCharacters': Optional[List[str]],
+    'command': Optional[Command],
```
rwols, line 201: The types here are not `Optional` I think.

predragnikolic:
```

	/**
	 * A human-readable string that represents a doc-comment.
	 */
	documentation?: string | MarkupContent;
```

For the `documentation` key, the value type may be `string | MarkupContent`, 
but the `documentation?` indicates that that key might be missing in the object. (that is not the same as Optional in mypy)

I was not sure how to type hint that in mypy. 
There is `total=false`, but I see no clear way to type hint what keys might be missing. 
So for all those fields I added the `Optional` type wrapper to indicate that. (But that is not the correct interpenetration of the TypeScript `key?` type). 

Should I remove all those Optional wrapper types?

rchl: Why do you think so?

In a case like this:

```py
c = {'label': 'foo'}  # type: CompletionItem
kind = c.get('kind')
```

what would you expect the type of `kind` be? I'd say `int` or `None` because `kind` is optional in the spec.

plugin/core/registry.py
predragnikolic: > Can you instead add them to LspTextCommand as methods?

Unfortunately adding them to `LspTextCommand` wont help me, because I need to get the session  by name in the `DocumentSyncListener` class:
```
class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListener):
```
as you can see `DocumentSyncListener` doesn't extend from `LspTextCommand`.

I could add the get `session_by_name` as a method on `DocumentSyncListener`, but then I would end up with 2 definitions for `session_by_name` (one in `LspTextCommand`, and the other would be in `DocumentSyncListener` and both would do the same thing, return the session by name, so that seem like code duplication for me). 

> I'm aiming to remove these free functions as they can be abused easily

I am still not sure how those functions could be abused. What is your concern?

plugin/core/protocol.py
rwols: There's a difference between optional keys and optional values... although the difference is hard to detect :)

rchl: I would say that the difference is impossible to represent in Python types. To make those keys optional they would have to be missing completely from the Completion type which would then be impossible to know their type if they are present.

plugin/core/registry.py
rwols: Oh I did not realize it was needed in DocumentSyncListener :) In that case it's easy to find your session:

```python
for sb in self.session_buffers_async():
    if sb.session.config.name == session_name:
        # ...
```

plugin/core/protocol.py
rwols: > I would say that the difference is impossible to represent in Python types.

Not really true. You can detect the difference:

```python
# 1: optional keys
if "key" in obj:
    val = obj["key"]

# 2: optional values
val = obj["key"]
if val is not None:
    # ... handle val ...
```

Although that first (1) snippet does two dictionary lookups. It's faster to do one lookup:

```python
# 3: single lookup, but you cannot distinguish optional keys from optional values this way
val = obj.get("key")
if val is not None:
    # ... handle val ...
```

In any case, there are already usages of `total=False` exactly to tell mypy that some keys are missing. Although, with `total=False`, you cannot tell mypy which keys are mandatory. Either they are all mandatory or they are all optional. I thought I read somewhere that you can use the below trick to make a subset of the keys mandatory, and another subset of the keys optional:

```python
CompletionItem_Required(typing.TypedDict):
    label: str  # label is required

CompletionItem(CompletionItem_Required, total=False):
    kind: int,
    tags: List[int],
    detail: str   # kind, tags, detail, etc, are optional
```

Unfortunately, we cannot use this syntax as we're still on py33.

Moral of the story I think you can do one of two things:

1) Keep `total=False`, but then remove all the `Optional[...]`
2) Set `total=True`, and keep all the `Optional[...]`

rchl: I see what you mean now.

> Moral of the story I think you can do one of two things:

Both of those options are unfortunate because then the person writing the code will not know which properties need to be checked for existence and which are expected to be always there.

In other words, the types specified as they are now, help the person writing the code write it in a way that doesn't crash without adding unnecessary checks for properties that are always there.

predragnikolic: We can introduce a new type 
```
OptionalKey = Optional
```

And later use it like:
```
'documentation': OptionalKey[str]
```

That way we have at least one way to distinguish an optional key, from an optional value. So a person writing the code would know what to do.

What do you think of this approach?

plugin/core/registry.py
predragnikolic: Thanks 🙂

rwols:
IMO this pull request should be split into two pull requests:

1) introduce more protocol types
2) implement "promise.all" for completions

predragnikolic:
If you want me to split this PR, just confirm and I will split it :)

rwols:
Please do :)

predragnikolic:
Ok I split out the PR.

I wont close this PR, because this PR will be responsible for:
> 2. implement "promise.all" for completions

Another PR will be created after this and will be responsible for:
> 1. introduce more protocol types

plugin/core/registry.py
predragnikolic: 1f608ec

plugin/documents.py
predragnikolic: d19284c

plugin/documents.py
```diff
@@ -606,46 +615,65 @@ def render_highlights_on_main_thread() -> None:
...
+                continue
+
+            response_items = []  # type: List[CompletionItem]
+            if isinstance(response, dict):
+                response_items = response["items"] or []
+                if response.get("isIncomplete", False):
+                    flags |= sublime.DYNAMIC_COMPLETIONS
+            elif isinstance(response, list):
+                response_items = response
+            response_items = sorted(response_items, key=lambda item: item.get("sortText") or item["label"])
```
rwols, line 668: Shouldn't this be a global sort, sorting over all completions?

plugin/documents.py
```diff
@@ -606,46 +615,65 @@ def render_highlights_on_main_thread() -> None:
...
-        elif isinstance(response, list):
-            response_items = response
-        response_items = sorted(response_items, key=lambda item: item.get("sortText") or item["label"])
-        LspResolveDocsCommand.completions = response_items
-        items = [format_completion(response_item, index, can_resolve_completion_items, session_name)
-                 for index, response_item in enumerate(response_items)]
+
+        for response, session_name in responses:
+            if isinstance(response, Error):
+                sublime.status_message('Completion error: {}'.format(response))
```
rwols, line 654: If there are two errors, the first error is overwritten by the second error in the status bar.

plugin/documents.py
```diff
@@ -606,46 +615,65 @@ def render_highlights_on_main_thread() -> None:
...
+                if response.get("isIncomplete", False):
+                    flags |= sublime.DYNAMIC_COMPLETIONS
+            elif isinstance(response, list):
+                response_items = response
+            response_items = sorted(response_items, key=lambda item: item.get("sortText") or item["label"])
+            LspResolveDocsCommand.completions.extend(response_items)
+            can_resolve_completion_items = session.has_capability('completionProvider.resolveProvider')
+            items.extend(
+                [format_completion(response_item, len(items) + index, can_resolve_completion_items, session.config.name)
+                 for index, response_item in enumerate(response_items)])
```
rwols, line 673: You can omit the outer `[` and `]` here.

plugin/documents.py
```diff
@@ -606,46 +615,65 @@ def render_highlights_on_main_thread() -> None:
...
+        completion_promises = []  # type: List[Promise[ResolvedCompletions]]
+        for session in sessions:
+
+            def completion_request() -> Promise[ResolvedCompletions]:
+                return session.send_request_task(
+                    Request.complete(text_document_position_params(self.view, location), self.view)
+                ).then(lambda response: (response, session.config.name))
+
+            completion_promises.append(completion_request())
+
```
rwols, line 634: In general I think there's too many empty lines. Only before the start of an inner function (like completion_request) and the end of an inner function is an empty line needed.

rwols (changes requested):

plugin/documents.py
rwols: The sorting keys of the different language servers are likely to be unrelated to each other. So sorting globally would be a bit odd. I guess how it's written now is okay then.

plugin/documents.py
```diff
@@ -2,10 +2,14 @@
 from .code_actions import CodeActionsByConfigName
 from .completion import LspResolveDocsCommand
 from .core.logging import debug
+from .core.promise import Promise
 from .core.protocol import CodeLens
+from .core.protocol import CompletionItem
+from .core.protocol import CompletionList
 from .core.protocol import Command
```
rchl, line 9: Order off.

predragnikolic: 369c895

plugin/documents.py
predragnikolic: d2deb69

plugin/documents.py
predragnikolic: 04345e2

plugin/documents.py
predragnikolic: a6f7189

tests/test_completion.py
```diff
@@ -78,7 +78,7 @@ def verify(self, *, completion_items: List[Dict[str, Any]], insert_text: str, ex
     def test_none(self) -> 'Generator':
         self.set_response("textDocument/completion", None)
         self.view.run_command('auto_complete')
-        yield lambda: self.view.is_auto_complete_visible()
+        yield lambda: self.view.is_auto_complete_visible() is False
```
rwols, line 81: Very odd :) why is this test supposed to be inverted?

rwols (approved):
Alright let's try this out

rwols (changes requested):
Please resolve the merge conflicts

tests/test_completion.py
predragnikolic: that test was failing:
```
ERROR: test_none (test_completion.QueryCompletionsTests)
  File "/home/predrag/.config/sublime-text/Packages/UnitTesting/unittesting/core/py33/case.py", line 28, in _executeTestPart
    yield from deferred
  File "/home/predrag/.config/sublime-text/Packages/LSP/tests/test_completion.py", line 81, in test_none
    yield lambda: self.view.is_auto_complete_visible()
TimeoutError: Condition not fulfilled within 4.00 seconds
```

And than I thought about that, and it makes sense that the AC is not visible (is False) if we set an empty list: 
```
completion_list.set_completions([])
completion_list.set_completions(None)
```

In both cases I don't expect to see the AC.

plugin/documents.py
```diff
@@ -55,6 +55,12 @@
     DocumentHighlightKind.Write: "write"
```
predragnikolic: hmm... let me check if I need to remove this

predragnikolic: nope, looks fine

plugin/completion.py
```diff
@@ -25,8 +25,8 @@ def run(self, edit: sublime.Edit, index: int, event: Optional[dict] = None) -> N
         if not detail or not documentation:
             # To make sure that the detail or documentation fields doesn't exist we need to resove the completion item.
             # If those fields appear after the item is resolved we show them in the popup.
-            session = self.best_session('completionProvider.resolveProvider')
-            if session:
+            session = self.session_by_name(session_name)
+            if session and session.has_capability('completionProvider.resolveProvider'):
```
rwols, line 29: Hmm, this doesn't work with dynamic registration and a document selector. `session.has_capability` would return False while it could have been registered for a particular language id. The previous code:

```
self.best_session('completionProvider.resolveProvider')
```

took care of going through the listener.

You should probably modify `session_by_name` to take an extra optional parameter `capability`, look up the listener in `session_by_name`, then iterate over the session buffers filtering by the capability.

rwols (changes requested):

plugin/completion.py
rwols: I should probably add some tests for this document selector + dynamic registration thing to guard against this.

predragnikolic: I will try to wrap my head around this tomorrow 🙂

rwols: I'll try to write an architecture.md file to explain the high-level ideas if I can find the time

predragnikolic: > You should probably modify session_by_name to take an extra optional parameter capability, look up the listener in session_by_name, then iterate over the session buffers filtering by the capability.

I don't see a need to filter out session by capability and here is why :)

1. We have two sessions: A, B
2. Both sessions have the `completionProvider.resolveProvider` capability.

When we trigger AC 
We have:
- 2 completion items from session A 
- 1 from session B.

```
completinItem1  (session A)
completinItem2  (session A)
completinItem3  (session B)
```

To resolve the completion item, we need to send the resolve request to a the session that Completion Item bellogs to.
For example you cannot rersolve completinItem1 with session B (because it was provided by Session A)

so I added the session_name when requesting documentaion popup to know to what session to send the resove compeltion request:
```
st_details += make_command_link("lsp_resolve_docs", "More", {"index": index, "session_name": session_name})
																			 ^^^^^^^^^^^^^^
```

Given what I just said,
my question to you is why do I need to add the filtering by capability, because I already know to what exact session name to send the request to?

rwols: > Both sessions have the completionProvider.resolveProvider capability.

If both A and B use `client/registerCapability` for registering their `completionProvider` with a document selector, then

```
session.has_capability('completionProvider.resolveProvider')
```

will return false. `listener.has_capability_async` will return true.

plugin/documents.py
```diff
@@ -616,46 +625,65 @@ def render_highlights_on_main_thread() -> None:
...
+        completion_promises = []  # type: List[Promise[ResolvedCompletions]]
+        for session in sessions:
+
+            def completion_request() -> Promise[ResolvedCompletions]:
+                return session.send_request_task(
+                    Request.complete(text_document_position_params(self.view, location), self.view)
+                ).then(lambda response: (response, session.config.name))
+
+            completion_promises.append(completion_request())
+        LspResolveDocsCommand.completions = []
```
rchl, line 643: I'm thinking it would be better to clear the list from `_on_all_settled`. In theory, with slow-running requests, it could happen that some stale completions would sneak into the latest request results.

predragnikolic: > In theory, with slow-running requests, it could happen that some stale completions would sneak into the latest request results.

To add on this, here is how the code currently behaves.

If a language server has `isIncomplete` completions and we: 
>> type "a"
... request#1 pending, set LspResolveDocsCommand.completions = []
<< get completions from server, set LspResolveDocsCommand.completions = response1
>> type "b"
... request#2 pending, set LspResolveDocsCommand.completions = []
>> user triggers documentation popup for completion item from request#1, before the response for the request "b" was received.

In that case an index error will be thrown. 
In that case moving the clearing logic to `_on_all_settled` would solve the problem.

> I'm thinking it would be better to clear the list from _on_all_settled

But then we would clear the list and then immediately populate it: (why would we clear the list before assigning a new completion, when we can just assign the new completions)
https://github.com/sublimelsp/LSP/pull/1582/files#diff-982c110867d9809c978a5597e4da66ce7538b327675a66b8f27214413b7c0ec1R675

rchl: Good point. I thought that with incomplete completions the popup will close on typing the next character but it doesn't.

But on the other hand, with the current code, there could be another issue:

With `isIncomplete` completions, the user types "a" and then "b". Both requests are started but are slow to respond so when they both resolve, they both append items to the `LspResolveDocsCommand.completions` list, and triggering documentation could then fail in some cases.

plugin/completion.py
predragnikolic: I tried a few times to get this done, but I got stuck. (I am not sure of what to change although you explained it 2 times)

Can this be done with a different PR?

plugin/documents.py
predragnikolic: There a couple of approaches that came to my mind:

a) never clear the completions. (currently we only clear them when we trigger the AC)
LspResolveDocsCommand.completions = []

b) pass the completion item instead of the index: (feels like the a right approach, but performance might be affected)
```diff
- st_details += make_command_link("lsp_resolve_docs", "More", {"index": index, "session_name": session_name})
+ st_details += make_command_link("lsp_resolve_docs", "More", {"completion_item": completion_item, "session_name": session_name})

```

predragnikolic: I would choose `a)` because it is simpler and not that different from the current behavoior.

rchl: > a) never clear the completions.

How can this work if we never clear them?

predragnikolic: We assign the response to `LspResolveDocsCommand.completions` (keep in mind that I no longer use `LspResolveDocsCommand.completions.entend(response_items)` like I used to. Instead I just use `=` as this will prevent the list from growing to infinity)
```
LspResolveDocsCommand.completions[session_name] = response_items
```
https://github.com/sublimelsp/LSP/pull/1582/commits/01a98131d403ef89c79d442e5b9f36eb28e0c3b7#diff-982c110867d9809c978a5597e4da66ce7538b327675a66b8f27214413b7c0ec1R675

When we want to resolve the completion item, 
we will look up the completion item in the `LspResolveDocsCommand.completions`
```
    def run(self, edit: sublime.Edit, index: int, session_name: str, event: Optional[dict] = None) -> None:
        item = self.completions[session_name][index]
```

https://github.com/sublimelsp/LSP/pull/1582/commits/01a98131d403ef89c79d442e5b9f36eb28e0c3b7#diff-1e31f37d39e534dca2dca7fd81d29d7e5a039a4b55a5765fdb4292c7a87760aaR21

That is it. 

There is no need to clear the `LspResolveDocsCommand.completions = {}` (changed to dict, previously was an array)

We can just delete this line, to fix the index error that would be thrown.
https://github.com/sublimelsp/LSP/pull/1582/commits/01a98131d403ef89c79d442e5b9f36eb28e0c3b7#diff-982c110867d9809c978a5597e4da66ce7538b327675a66b8f27214413b7c0ec1R643

plugin/completion.py
rwols: I'll look at it maybe tomorrow

rwols: Servers like lua-language-server and OmniSharp register their completionProvider dynamically, so showing docs would break for those I guess.

plugin/documents.py
rchl: Another day, maybe I lost some context already, but clearing from `_on_query_completions_async` seems wrong because of this case:

(isIncomplete == true)
1. A completion request is completed and a completion popup is shown.
2. User types another letter - `LspResolveDocsCommand.completions` is cleared immediately
3. The previous request is slow to respond and the user tries to invoke documentation on some existing item - an error because `LspResolveDocsCommand.completions` is empty.

I think after the change to a dict it makes even more sense to clear it from the `_on_all_settled` because new completion request might not trigger for the same session names so at least from the memory/GC point of view it would make sense to clear that object at some point.

predragnikolic: I moved it to `_on_all_settled` to avoid errors :)

>  it would make sense to clear that object at some point

I guess the "ideal" place to clear the dict would be when the AC closes(either by the user selecting the completion item, or by pressing the esc key, or by pressing left right arrows). I know that there is a `is_auto_complete_visible()` api, but I don't yet have a simple solution for that approach(so I would avoid complicating the logic now).

But I believe that moving it to `_on_all_settled` improves things a bit.

plugin/completion.py
predragnikolic: Raoul is there any way to modify the codebase so `session.has_capability('completionProvider.resolveProvider')` is also affected when we register capabilities dynamically.  
```
lus_session = self.session_by_name("lua-language-server")
lus_session.has_capability('completionProvider.resolveProvider') # returns False
# in the mean time the server did registered that capability dynamically,
# so the next time we call:
lus_session.has_capability('completionProvider.resolveProvider') # it returns True
```

rwols: That doesn't work, because if an LS has two different document selectors, it might register completionProvider.resolveProvider for document selector A, but not for document selector B.

For example, SourceKit could register a colorProvider for swift files, but not for C/C++/Obj-C/Obj-C++ files. So an extra argument `view` is needed. But in any case, I think it's better to go through the listener for this anyway. See my latest commit.

rwols (approved):

rchl:
I've noticed that there is an empty popup shown for a brief moment now when resolving documentation.

rchl (approved):

PR #1583, opened 2021-02-14
Require a value when calling resolve function of a Promise

rchl:
Make the value passed to ResolveFunc non-optional. This allows to
skip some if-checks when handling resolved value in a fullfill
function.

This also aligns the behavior with a recent change in Typescript [1].

[1] https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#resolves-parameters-are-no-longer-optional-in-promises


rchl:
Tricky one that mypy type error.

Even though the ResolveFunc requires an argument now, `Promise.resolve()` can still be called without argument according to TS types. So that creates a bit of a problem that I'm not sure how to fix with Python...

rwols:
What's the plan here? Is this supposed to be merged before #1582 ?

predragnikolic:
If this gets merged I will remove one if statement in  #1582 :)
This PR will probably be merged before #1582

rchl:
I'd have to figure out how to fix the mypy issue first and so far I'm not sure.

rchl:
I've made the argument to Promise.[resolve|on_main_thread|on_async_thread] non-optional. You decide if that's an OK solution. I don't hate it, I think being explicit is not a bad thing.

rwols (approved):

PR #1586, opened 2021-02-16
updated haskell doc

AzogMugen:
Ghcide is no longer maintained and the repo is archived.
For a beginner in haskell like me, it was a real struggle to have a clean haskell environment to work with due to old and misleading documentation on all the web.
I don't want the next guy to try haskell to suffer (or at least less than I did).

Also, thanks a lot for this magical plugin for sublime text !
Really good job, keep going this way !

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -405,16 +405,19 @@ npm install -g flow-language-server
 
 ### Haskell
 
-1. Install [ghcide](https://github.com/digital-asset/ghcide).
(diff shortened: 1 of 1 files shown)


rwols (approved):
This was already updated in the ST4 docs, which aren't public yet: https://sublimelsp.github.io/LSP/language_servers/#haskell

but it doesn't hurt to have it in this spot as well.

AzogMugen:
Good to know and thanks ! Have a nice day !

PR #1589, opened 2021-02-20
Don't lookup the listener so many times

rwols:
A tiny improvement. Found during debugging of #1575. Most notably, the
`provider_exists` is called for a few different providers, and each time
we do the same lookup. This is now avoided by remembering with which
listener we are dealing with.


rchl (approved):
Unrelated to this change but while trying out this change, I've noticed that on going to next/previous diagnostics with F4/shift+F4, the popup is not always shown. I think that's because the popup is shown before the view is scrolled completely and ST closes the popup if it's not in the viewport.

PR #1591, opened 2021-02-23
Enable hardbreak in mdpopups to handle escaped newlines

rwols:
Discussion can be found here: https://github.com/facelessuser/sublime-markdown-popups/issues/119


rchl (approved):
Let's try!

PR #1592, opened 2021-02-23
Add a test for forbidden trigger characters

rwols:
Interestingly, this test succeeds. I have to dig further into this.


rchl:
Yes, this also seemed to work for me when I was testing manually.

Although completions returned by LSP-angular looked different from what @princemaple logged.

PR #1600, opened 2021-02-28
Remove ignore_server_trigger_chars setting

rwols:
An alternative is using disabled_capabilities.
I've also added more tests for dynamic registration for this.
Dynamic registration went through for disabled capabilities anyway.
But not anymore :)

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -713,16 +713,12 @@ class ClientConfig:
                     return True
         return False
 
-    def filter_disabled_capabilities(self, capability_path: str, options: Dict[str, Any]) -> bool:
-        if self.is_disabled_capability(capability_path):
-            return True
-        to_be_removed = []  # type: List[str]
-        for k in options.keys():
-            if self.is_disabled_capability("{}.{}".format(capability_path, k)):
-                to_be_removed.append(k)
-        for k in to_be_removed:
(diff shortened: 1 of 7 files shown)


plugin/core/types.py
```diff
@@ -719,6 +713,17 @@ def is_disabled_capability(self, capability_path: str) -> bool:
...
+        if self.is_disabled_capability(capability_path):
+            return True
+        to_be_removed = []  # type: List[str]
+        for k in options.keys():
+            if self.is_disabled_capability("{}.{}".format(capability_path, k)):
+                to_be_removed.append(k)
+        for k in to_be_removed:
+            options.pop(k)
+        return False
+
```
rchl, line 726: This is a rather confusing method:
 - It's not that obvious what that return value means
 - I'm not sure if the name really describes well what it does.
 - Mutating the passed-in `options` doesn't seem like a great practice

How about changing it like so:
 - return a new object with only enabled capabilities
 - don't check `is_disabled_capability` here, move it outside of this method
 - call it `filter_out_disabled_capabilities`? I feel like "filter" can mean either "return only disabled capabilities" or "return only enabled capabilities"

rwols: fixed in 08c6c80

rchl (approved):

PR #1606, opened 2021-03-06
Remove temp disabled config logic

rwols:
Putting a configuration into a temporarily disabled state has the major drawback that you can't enable it anymore, unless you close the window and re-open the window.

Instead, disable a failing config via the settings machinery. This way you can enable it again by clicking on Tools > LSP > Enable Language Server In Project.

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -339,10 +339,10 @@ class WindowManager(Manager):
             self._new_session = session
         except Exception as e:
             message = "".join((
-                "Failed tot start subprocess for {0}. Reason:\n\n",
+                "Failed to start subprocess for {0}. Reason:\n\n",
                 "{1}\n\n",
                 "{0} will be disabled for this project. Enable {0} again by running ",
-                "LSP: Enable Language Server In Project from the Command Palette."
+                "\"LSP: Enable Language Server In Project\" from the Command Palette."
             )).format(config.name, str(e))
             exception_log("Unable to start subprocess for {}".format(config.name), e)
             if isinstance(e, CalledProcessError):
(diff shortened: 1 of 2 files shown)


rchl (approved):
I'm a little skeptical about this because users might not understand how to turn those back on and that will lead to more "support" requests.

Maybe the error message should at least mention the specific command to use for enabling?

plugin/core/windows.py
```diff
@@ -338,10 +338,11 @@ def start_async(self, config: ClientConfig, initiating_view: sublime.View) -> No
                 lambda session, is_error: self._on_post_session_initialize(initiating_view, session, is_error))
             self._new_session = session
         except Exception as e:
-            message = "\n\n".join((
-                "Failed to start subprocess for {0}. Reason:",
-                "{1}",
-                "{0} will be disabled for this project. Enable {0} again using the Command Palette."
+            message = "".join((
+                "Failed tot start subprocess for {0}. Reason:\n\n",
```
rchl, line 342: typo "tot"

plugin/core/windows.py
```diff
@@ -338,10 +338,11 @@ def start_async(self, config: ClientConfig, initiating_view: sublime.View) -> No
...
+                "{1}\n\n",
+                "{0} will be disabled for this project. Enable {0} again by running ",
+                "LSP: Enable Language Server In Project from the Command Palette."
```
rchl, line 345: Quote the command name?

plugin/core/windows.py
```diff
@@ -425,8 +426,10 @@ def on_post_exit_async(self, session: Session, exit_code: int, exception: Option
             msg = "{} exited with status code {}".format(config.name, exit_code)
             if exception:
                 msg += " and message:\n\n---\n{}\n---".format(str(exception))
-            msg += "".join(("\n\nDo you want to restart {0}?\n\nIf you choose Cancel, {0} will be disabled for this ",
-                            "project. You can re-enable {0} manually using the Command Palette.")).format(config.name)
+            msg += "".join((
+                "\n\nDo you want to restart {0}?\n\nIf you choose Cancel, {0} will be disabled for this project. ",
+                "Enable {0} again by running LSP: Enable Language Server In Project from the Command Palette."
```
rchl, line 431: Quote the command name here too?

rchl (changes requested):

rchl (approved):

PR #1610, opened 2021-03-10
Fix completion docs not showing for servers that are not resolve providers

rwols:


rchl:
Yeah, should keep the return and just make sure that `self.handle_resolve_response_async` is called in all cases.

rchl (approved):

PR #1611, opened 2021-03-11
feat(code actions): display code actions within a quick panel 

BenoitPingris:
Display the code action in a quick panel instead of the context menu, this makes us able to use more keybindings to control which code action we want to use.
With the context menu you're only able to select an item with `up/down arrows` or the mouse, with a quick panel you can add custom keybindings and also use the search feature of the quick panel.

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -302,9 +302,14 @@ class LspCodeActionsCommand(LspTextCommand):
 
     def show_popup_menu(self) -> None:
         if len(self.commands) > 0:
-            self.view.window().show_quick_panel([command[1] for command in self.commands], 
-                    self.handle_select, 
-                    placeholder="Code action")
+            items = [command[1] for command in self.commands]
+            win = self.view.window()
+            if win:
+                win.show_quick_panel(items, 
+                        self.handle_select, 
+                        placeholder="Code action")
+            else:
+                self.view.show_popup_menu(items, self.handle_select)
         else:
             self.view.show_popup('No actions available', sublime.HIDE_ON_MOUSE_MOVE_AWAY)
 
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -306,8 +306,6 @@ class LspCodeActionsCommand(LspTextCommand):
             window = self.view.window()
             if window:
                 window.show_quick_panel(items, self.handle_select, placeholder="Code action")
-            else:
-                self.view.show_popup_menu(items, self.handle_select)
         else:
             self.view.show_popup('No actions available', sublime.HIDE_ON_MOUSE_MOVE_AWAY)
 
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -302,7 +302,9 @@ class LspCodeActionsCommand(LspTextCommand):
 
     def show_popup_menu(self) -> None:
         if len(self.commands) > 0:
-            self.view.show_popup_menu([command[1] for command in self.commands], self.handle_select)
+            self.view.window().show_quick_panel([command[1] for command in self.commands], 
+                    self.handle_select, 
+                    placeholder="Code action")
         else:
             self.view.show_popup('No actions available', sublime.HIDE_ON_MOUSE_MOVE_AWAY)
 
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -305,9 +305,7 @@ class LspCodeActionsCommand(LspTextCommand):
             items = [command[1] for command in self.commands]
             win = self.view.window()
             if win:
-                win.show_quick_panel(items, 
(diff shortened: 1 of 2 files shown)


rchl:
- Code actions can also be triggered from the hover popup and that is not handled in those changes so far.
- we don't like using abbreviated variable names (win => window)
- we should probably update naming while making this change. So change `show_popup_menu` to something more generic like `show_code_actions`.

BenoitPingris:
Are you referring to that hover popup? 

I don't see where the code actions should be displayed in that popup, or maybe I'm looking to the wrong one!

rchl:
When you have an error in the code and you hover the error.
Screenshot 2021-03-11 at 11 01 38

plugin/code_actions.py
```diff
@@ -302,7 +302,12 @@ def combine_commands(self) -> 'List[Tuple[str, str, CodeActionOrCommand]]':
 
     def show_popup_menu(self) -> None:
         if len(self.commands) > 0:
-            self.view.show_popup_menu([command[1] for command in self.commands], self.handle_select)
+            items = [command[1] for command in self.commands]
+            win = self.view.window()
+            if win:
+                win.show_quick_panel(items, self.handle_select, placeholder="Code action")
+            else:
```
predragnikolic, line 309: I think that there is no point in having this else branch. 
It can be removed.
Pretty certain that `if win:` will always be true.

predragnikolic:
I will try this PR today 🙂

plugin/code_actions.py
BenoitPingris: I needed to add the `if` otherwise mypy won't be happy as `window` is an `Optional[Window]`, so I also put the `else` in there in case `window` is indeed `None` that way we sure we'll have our precious code actions!

predragnikolic:
I tried this and I don't think that the quick panel provides a better user experience than the context menu:

Context menu:
context

Quick Panel:
output

rchl:
> note: how when the quick panel is shown it covered most of my code so I can loose context to where I initially invoked the code action

I'm not sure I agree with that because the focus goes back to the cursor and so there shouldn't be any confusion or disorientation IMO.

The fact that you can filter items is also a plus and usability improvement.

I've only used this new way for a minute or so, so far.

BenoitPingris:
Yea I understand the "fear" of losing some context. 

I must say in my case it's much more comfortable, even if I was working with a layout more or less like the one you gave in your example (small window size, sidebar opened etc) this wont annoy me as when I open code actions, I know where (in the code) I call it and for what purpose so it wont bother me.

I was also looking in a way to use the completion list (so we have keybindings, consistent styling and no code overflow) but my guess is that its only purpose is: completion? So we can't add a `on_done` callback when we select an item in it and use the code action's handler.

plugin/code_actions.py
predragnikolic: If a window doesn't exist, can the view exist?(because the view belongs to a window) 

I still think that there is no need to have both `window.show_quick_panel`(will work in 99% of the cases) and `view.show_popup_menu`(as fallback)
```diff
            if window:
                window.show_quick_panel(items, self.handle_select, placeholder="Code action")
-           else:
-                 self.view.show_popup_menu(items, self.handle_select)
```

The only situation where this will hit the else branch is when `window_id == 0` which is almost never. (I forgot the case where the window will be none. Maybe when we set a timeout and close the window in the mean time)
```python
# copied from sublime.py
class View:
    def window(self):
        window_id = sublime_api.view_window(self.view_id)
        if window_id == 0:
            return None
        else:
            return Window(window_id)
```

rwols (approved):
I don't have a strong opinion about this. All in all looks good to me.

rchl (approved):

predragnikolic:
I do like the ability to filter items. (but I still think that visually the context menu is more suited for this kind of stuff)
Maybe I just need more time to get used to it. :)

So lets merge this. We can act later based on the feedback from other people.

predragnikolic (approved):

predragnikolic:
BTW, thanks for the PR :)

princemaple:
auto save + auto format on save + LSP switching code action from inline context menu to command palette is not playing well

The problem with this is, I trigger auto import from LSP, it pops up the command palette, which causes loss of focus from the file, it auto saves and triggers auto format, and the auto import becomes invalid (either because of request id or code change, doesn't matter), choosing from the commands ends with up noop. I have to do it again, or save beforehand, which loses the point of auto save

Previously I'd write my code and go through all the unimported names and run auto import on them, then switch to browser or something, which triggers auto saves.
Now I have to go to a name, either save the file first then auto import, or be too quick and auto import directly and realize it's noop and have to do it again, and repeat for every name waiting to be imported.

PR #1613, opened 2021-03-11
Remove "log_stderr" in favor of always logging errors to the log panel

predragnikolic:
Remove `log_stderr` in favor of always logging errors to the log panel.

### DIFF
--- a/docs/src/guides/client_configuration.md
+++ b/docs/src/guides/client_configuration.md
@@ -12,7 +12,7 @@ Here is an example of the `LSP.sublime-settings` file with configurations for th
 ```js
 {
   // General settings
-  "log_debug": true,
+  "show_diagnostics_panel_on_save": 0,
 
   // Language server configurations
   "clients": {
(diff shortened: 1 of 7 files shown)


docs/src/guides/client_configuration.md
```diff
@@ -12,7 +12,7 @@ Here is an example of the `LSP.sublime-settings` file with configurations for th
 ```js
 {
   // General settings
-  "log_stderr": true,
+  "log_debug": true,
```
rchl, line 15: Should we suggest this one really?

predragnikolic: I could put `show_view_status` instead of it, or some other setting?

rwols: `"show_diagnostics_panel_on_save": 0` is IMO a good example

predragnikolic: b1fefb1

rwols (approved):

PR #1614, opened 2021-03-13
Make scopes in output panels consistent with Find results panel

jwortmann:
Some small tweaks for the scopes used in the diagnostics and references (rename) panels:

* `entity.name.filename` for the filenames (already used in Find in
  Files results panel and by SublimeLinter)
* `constant.numeric.line-number` for line numbers (already used in Find
  in Files results panel and by SublimeLinter)
* `constant.numeric.col-number` for column numbers (already used by
  SublimeLinter and seems to be a suitable scope name)

*Note*: this will change the color in Mariana for filenames in the diagnostics panel form green (string) to orange (entity.name), which may look a little bit worse than before, but I think it's more desirable to have a consistent style over different panels and with the built-in Find Results panel. And the Monokai scheme has a rule for `entity.name.filename`, so we should use that scope here as well. We could maybe keep the `string.unquoted` in front as a fallback, but since `entity.name` is within the minimal scope coverage for color schemes and all default schemes address it, I don't think that scope is of much use and therefore removed it.

### DIFF
--- a/Syntaxes/Diagnostics.sublime-syntax
+++ b/Syntaxes/Diagnostics.sublime-syntax
@@ -16,7 +16,7 @@ contexts:
     - match: ^(?!\s*\d+:\d+)(.*)(:)$
       captures:
         0: meta.diagnostic.preamble.lsp
-        1: string.unquoted.lsp
+        1: entity.name.filename.lsp
         2: punctuation.separator.lsp
 
   line:
@@ -68,9 +68,9 @@ contexts:
     - include: pop-at-end
     - match: (\d+)(:)(\d+)
       captures:
-        1: constant.numeric.integer.decimal.lsp
+        1: constant.numeric.line-number.lsp
         2: punctuation.separator.lsp
-        3: constant.numeric.integer.decimal.lsp
+        3: constant.numeric.col-number.lsp
       pop: true
 
   pop-at-end:
--- a/Syntaxes/Diagnostics.sublime-syntax
+++ b/Syntaxes/Diagnostics.sublime-syntax
@@ -68,9 +68,9 @@ contexts:
     - include: pop-at-end
     - match: (\d+)(:)(\d+)
       captures:
-        1: constant.numeric.line-number.lsp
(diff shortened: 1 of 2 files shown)


rwols:
A change in a syntax file that has an `entity.name` scope causes ST to re-index the world. I'm unsure if this is also the case when a packaged is upgraded via package control. But in any case that's the reason why I changed it to `string.unquoted` (it used to be `entity.name.filename` before).

jwortmann:
Afaik `entity.name` or `entity.name.filename` is not indexed by default. See https://github.com/sublimehq/Packages/issues/2719 for the default indexed scopes and some new suggestions.

This scope seems to always have been `string.unquoted` since you added it in https://github.com/sublimelsp/LSP/commit/744d2ee9b9c2e6766836b0c16dfa7913c83d686b in 2017. :thinking:

I don't think there will be any problem with `entity.name.filename`, otherwise wouldn't such an issue already been discovered in the built-in Find-in-Files feature or in SublimeLinter?

rwols:
Oh, yes, looks like the history in my brain is fuzzy. Luckily we still have the git history. But I remember now what my issue was. It's that References.sublime-syntax uses `entity.name.file.reference.lsp` like you pointed out.

But testing this now it seems that ST re-indexes the world no matter what a newly introduced .sublime-syntax contains. It'd be nice to communicate to ST "please don't reindex". Something like an `index: false` as a top-level key.

rwols (approved):

jwortmann:
> But testing this now it seems that ST re-indexes the world no matter what a newly introduced .sublime-syntax contains. It'd be nice to communicate to ST "please don't reindex". Something like an index: false as a top-level key.

Okay, I didn't know about that re-indexing before and assumed that a syntax change would only affect the files it actually applies to (here only the diagnostics panels). But the real behavior seems indeed to be exactly like you described. Might be a place where ST could be improved to be a bit smarter.

Anyway, I think it's useful now to have consistent colors across panels and it's easier to handle in color schemes.

PR #1616, opened 2021-03-19
Create CNAME

predragnikolic:
add CNAME to the docs_dir so that the file does not get removed the next time we deploy

https://www.mkdocs.org/user-guide/deploying-your-docs/#custom-domains


rwols:
That link says:

> you need to copy the file to your docs_dir

In mkdocs.yml I see that docs_dir is `docs/src`? https://github.com/sublimelsp/LSP/blob/st4000-exploration/docs/mkdocs.yml#L5

predragnikolic:
@rwols you are right.

I created that file via GitHub web and totally missed the directory.

rwols (approved):

PR #1617, opened 2021-03-20
Remove mutex from promise

rwols:
Completions are janky now. Removing the mutex from the promise solves this.


rchl:
Is it safe though?

And if it's safe is it only because of how it is used right now (only from a single thread)? And if so, would it be likely that we would introduce a case in the future where it's not safe?

Also, i haven't noticed completions being janky. Are you able to show the issue on video or something?

rchl:
Actually if it would only be used from a single thread then you should not be able to notice any jank since the threads would not be blocking. So it's it safe then?

predragnikolic:
I also haven't noticed completions being janky

rwols:
I can't seem to reproduce it now...

PR #1619, opened 2021-03-21
Put instructions at the top when langserver gets disabled

rwols:
Otherwise the instructions might be drowned out by the error message
on macOS especially due to the small error dialogs.

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -339,10 +339,10 @@ class WindowManager(Manager):
             self._new_session = session
         except Exception as e:
             message = "".join((
-                "Failed to start subprocess for {0}. Reason:\n\n",
-                "{1}\n\n",
-                "{0} will be disabled for this project. Enable {0} again by running ",
-                "\"LSP: Enable Language Server In Project\" from the Command Palette."
+                "Failed to start subprocess for {0}. {0} will be disabled for this window. ",
+                "Enable {0} again by running \"LSP: Enable Language Server In Project\" from the Command Palette.",
+                "\n\n",
+                "{1}"
             )).format(config.name, str(e))
             exception_log("Unable to start subprocess for {}".format(config.name), e)
             if isinstance(e, CalledProcessError):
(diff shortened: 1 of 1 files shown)


plugin/core/windows.py
```diff
@@ -339,10 +339,10 @@ def start_async(self, config: ClientConfig, initiating_view: sublime.View) -> No
             self._new_session = session
         except Exception as e:
             message = "".join((
-                "Failed to start subprocess for {0}. Reason:\n\n",
-                "{1}\n\n",
-                "{0} will be disabled for this project. Enable {0} again by running ",
-                "\"LSP: Enable Language Server In Project\" from the Command Palette."
+                "Failed to start subprocess for {0}. {0} will be disabled for this window. ",
+                "Enable {0} again by running \"LSP: Enable Language Server In Project\" from the Command Palette.",
```
rchl, line 343: - I think we can remove the "subprocess" word which is kinda a lower-level detail.
- The server name is repeated 3 times, once per-sentence.
Suggestion:
```suggestion
                "Failed to start \"{0}\" - disabling for this window. ",
                "Re-enable by running \"LSP: Enable Language Server In Project\" from the Command Palette.",
```

plugin/core/windows.py
```diff
@@ -339,10 +339,10 @@ def start_async(self, config: ClientConfig, initiating_view: sublime.View) -> No
...
+                "\n\n",
+                "{1}"
```
rchl, line 345: Let's try to make it clear that what follows is the error returned from the server. Maybe just a `Returned error: ` prefix would be enough.

plugin/core/windows.py
```diff
@@ -423,13 +423,13 @@ def on_post_exit_async(self, session: Session, exit_code: int, exception: Option
...
             config = session.config
-            msg = "{} exited with status code {}".format(config.name, exit_code)
-            if exception:
-                msg += " and message:\n\n---\n{}\n---".format(str(exception))
-            msg += "".join((
-                "\n\nDo you want to restart {0}?\n\nIf you choose Cancel, {0} will be disabled for this project. ",
+            msg = "".join((
+                "{0} exited with status code {1}. ",
+                "Do you want to restart {0}? If you choose Cancel, {0} will be disabled for this window. ",
                 "Enable {0} again by running \"LSP: Enable Language Server In Project\" from the Command Palette."
```
rchl, line 429: Again server name repeated 4 times.

```suggestion
                "{0} exited with status code {1}. ",
                "Do you want to restart it? If you choose Cancel, it will be disabled for this window. ",
                "Re-enable by running \"LSP: Enable Language Server In Project\" from the Command Palette."
```

plugin/core/windows.py
```diff
@@ -423,13 +423,13 @@ def on_post_exit_async(self, session: Session, exit_code: int, exception: Option
...
-            )).format(config.name)
+            )).format(config.name, exit_code)
+            if exception:
+                msg += "\n\n---\n{}\n---".format(str(exception))
```
rchl, line 432: Maybe:

```suggestion
                msg += "\n\n--- Error: ---\n{}".format(str(exception))
```

There is no need for the trailing dashes I think.

rchl (changes requested):

rchl (approved):

PR #1623, opened 2021-03-24
Fix null session completions

### LINKED ISSUE #1622: LSP sometimes confuses autocompletion even in unrelated syntaxes
**Describe the bug**

Every here and then I find me surprised by ST's mini_auto_completion to commit different content than it previews. See how it shows `true` and commits `terminator` in the following code snippet.

Animation

Why do I report it here? Well, in fact disabling the LSP package fixes the issue.

The strange thing is, the syntax in use not being supported by any language-server on earth. So there's none active.

The question is, how can or does LSP interact with ST's auto-completion in syntaxes which do not use LSP?

The only thing I saw on console was the `commit_completion` command being executed when hitting tab.

**Expected behavior**

The previewe'd completion should be committed.

This is what it looks like with LSP disabled:

Animation

**Screenshots**

- OS: Windows 10 x64 2004
- Sublime Text version: 4099
- LSP version: ST4000-exploration (commit 1bc9cff7f4385637cafdb0ae43ec843367cee19f)
- Language servers used: all disabled for debugging

deathaxe:
Fixes #1622

The `sessions()` method returns a generator object, which causes the query for `if not sessions` to always return false. As a result the whole promise machinery is running even if no session is available.

This commit therefore queries `completion_promises` instead, to shorten the route in case of no sessions being available.

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -639,9 +639,13 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
     # --- textDocument/complete ----------------------------------------------------------------------------------------
 
     def _on_query_completions_async(self, resolve_completion_list: ResolveCompletionsFn, location: int) -> None:
+        sessions = list(self.sessions('completionProvider'))
+        if not sessions:
+            resolve_completion_list([], 0)
+            return
         self.purge_changes_async()
         completion_promises = []  # type: List[Promise[ResolvedCompletions]]
-        for session in self.sessions('completionProvider'):
+        for session in sessions:
 
             def completion_request() -> Promise[ResolvedCompletions]:
                 return session.send_request_task(
@@ -649,10 +653,6 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
                 ).then(lambda response: (response, session.config.name))
 
             completion_promises.append(completion_request())
-
-        if not completion_promises:
-            resolve_completion_list([], 0)
-            return
 
         Promise.all(completion_promises).then(
             lambda responses: self._on_all_settled(responses, resolve_completion_list))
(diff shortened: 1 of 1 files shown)


plugin/documents.py
```diff
@@ -639,13 +639,9 @@ def render_highlights_on_main_thread() -> None:
     # --- textDocument/complete ----------------------------------------------------------------------------------------
 
     def _on_query_completions_async(self, resolve_completion_list: ResolveCompletionsFn, location: int) -> None:
-        sessions = self.sessions('completionProvider')
```
deathaxe, line 642: `sessions = tuple(self.sessions('completionProvider'))` may also work.

rchl: Why `tuple` and not `list`?

BTW. that sounds like a better solution because we wouldn't call `self.purge_changes_async()` unnecessarily when there are no sessions.
And you would only need this one-line change.

rwols: Yeah I think this can be a one-liner

deathaxe:
If there is no session, nothing needs to happen. Not even triggering some async stuff at all. Querying some lists is fast enough not to be called async at all. If no session is available the initial `on_query_completions` may exit without bothering `set_timeout_async` at all.

plugin/documents.py
deathaxe: `tuple` because there is no need to manipulate the set of sessions.

rwols:
> Querying some lists is fast enough not to be called async at all. If no session is available the initial on_query_completions may exit without bothering set_timeout_async at all.

The problem with that is that iterating over data structures is usually unsafe in this package, because they might mutate in the async thread. Almost everything has to be done in the async thread. It's not the speed of iteration that matters but it's the thread safety of mutable containers.

deathaxe:
I see.

rwols (approved):

Issue #1627, opened 2021-03-27
Remove some sections from the README


predragnikolic:
**Is your feature request related to a problem? Please describe.**
I believe that the README has some information that are not that important to the user. 

**Describe the solution you'd like**
1. [Customisation of the popups](https://github.com/sublimelsp/LSP#customisation-of-the-popups), is not the most important feature about LSP, so I would move it to the docs in the guides section. 

![image](https://user-images.githubusercontent.com/22029477/112727682-f3b0a700-8f23-11eb-9540-ee853ef6b689.png)

2. [LSP specification implementation status](https://github.com/sublimelsp/LSP#lsp-specification-implementation-status) is also not that important to a regular user. Would consider moving it either to:
* the wiki page [here](https://github.com/sublimelsp/LSP/wiki#protocol-compatibility) (there is already a section for it), 
* or the docs site (but I don't know where I would put it)

rwols:
> LSP specification implementation status is also not that important to a regular user.

There should be a section in the docs for language server maintainers explaining the peculiarities of this client. For instance, we use Python-Markdown, "document outline" is quite different from VSCode, ST doesn't really understand Language IDs, etc. The overall protocol implementation status can go in there as well.

predragnikolic:
Closed with https://github.com/sublimelsp/LSP/pull/1629

Thanks @jwortmann!

PR #1628, opened 2021-03-27
Move customization of popups to docs

predragnikolic:
References #1627

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,10 +30,6 @@ Open a document supported by the language server. LSP should report the language
 
 See more information in the [documentation](https://lsp.sublimetext.io) :open_book:.
 
-## Customisation of the popups
-
-LSP uses [mdpopups](https://github.com/facelessuser/sublime-markdown-popups) to display the popup. You can override its style by creating a `Packages/User/mdpopups.css` file. See the [mdpopups documentation](http://facelessuser.github.io/sublime-markdown-popups/) for more details.
-
 ## Getting help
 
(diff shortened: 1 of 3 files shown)


jwortmann:
There are currently some descriptions how to customize the regions for diagnostics, document highlights etc. under "Features > Configuring > Color configurations". Maybe it would be useful to have a general page "Configurations" and combine them all together there, with the popup custimization. And perhaps the "Package Settings (LSP)" section can be removed, because it's redundant and just a duplication of the LSP.sublime-settings file?

The hierarchy could then be:

Customization.md
* Sublime settings
* Hover popups
  LSP uses mdpopups to display....
* Color configurations (maybe rename to "Color scheme configurations"?)
  * Document Highlights
  * Diagnostics
  * Signature Help
  * Code Lens

Also, how about adding back a badge with a link to the docs like it is on the master branch? I find it useful so that you can easily click a link to the docs, without searching the whole Readme for it. For example something like this:
```markdown
[![Documentation](https://img.shields.io/badge/docs-ST4-blue)](https://lsp.sublimetext.io/)
```

predragnikolic:
+1 for adding the docs badge

As for the adding a new page "Configurations" and reorganizing the Docs is out of the scope of I wanted with this PR. I just want to free some space in the README. 

But feel free to open a PR if you think that you can improve the docs. :)

rwols (approved):

rwols:
Yeah the sections are a bit messy. I agree with @jwortmann.

PR #1629, opened 2021-03-31
Reorder some sections in the documentation

jwortmann:
* Combine mdpopups and color scheme customizations on a new page "Customizations"
* Remove the LSP settings from the "Features" page; they were just duplicated from the LSP.sublime-settings file without any additional information
* Move LSP spec implementation status from README to docs


rwols (approved):

rwols:
Hmm, some of the nested lists in https://lsp.sublimetext.io/lsp_implementation_status/ are not nested any longer.

jwortmann:
I'll see if I can fix it.

PR #1633, opened 2021-04-02
Add Find References to commands list

yeraydiazdiaz:
Hello, thank you for your work on LSP.

I noticed "Find references" was in the context menu but not in the commands list so I thought I would add it.

Seems to work on my Mac OS X using ST4 build 4099:

Screenshot 2021-04-02 at 12 54 08


rwols (approved):

rwols:
Many thanks!

PR #1637, opened 2021-04-07
Optimize pruning of messages over the limit in the log panel

rchl:
- Don't calculate rowcol after appending each line
 - Don't store and remove line regions separately. Do it in one erase instead.

### DIFF
--- a/plugin/core/panels.py
+++ b/plugin/core/panels.py
@@ -166,14 +166,10 @@ class LspUpdateServerPanelCommand(sublime_plugin.TextCommand):
             for prefix, message in to_process:
                 message = message.replace("\r\n", "\n")  # normalize Windows eol
                 self.view.insert(edit, self.view.size(), "{}: {}\n".format(prefix, message))
-                total_lines, _ = self.view.rowcol(self.view.size())
-                point = 0  # Starting from point 0 in the panel ...
-                regions = []  # type: List[sublime.Region]
+            last_region_end = 0  # Starting from point 0 in the panel ...
+            total_lines, _ = self.view.rowcol(self.view.size())
             for _ in range(0, max(0, total_lines - SERVER_PANEL_MAX_LINES)):
                 # ... collect all regions that span an entire line ...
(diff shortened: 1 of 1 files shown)


rchl:
Also realized that with this lazy update that I've introduced before, we'll append logs to the list without any limit. Only showing the log panel will clear it. So should do something about that...

plugin/core/panels.py
```diff
@@ -163,17 +163,18 @@ def run(self, edit: sublime.Edit, window_id: int) -> None:
         to_process = WindowPanelListener.server_log_map.get(window_id) or []
         WindowPanelListener.server_log_map[window_id] = []
         with mutable(self.view):
+            new_lines = []
             for prefix, message in to_process:
                 message = message.replace("\r\n", "\n")  # normalize Windows eol
-                self.view.insert(edit, self.view.size(), "{}: {}\n".format(prefix, message))
+                new_lines.append("{}: {}\n".format(prefix, message))
```
rchl, line 169: I've assumed that appending to list and concatenating later would be faster than appending to a string but that's not backed by any testing so I might be wrong.

rwols: The `view.insert` also does inter-process serialization. So it makes sense that building the string here in python is faster.

plugin/core/panels.py
```diff
@@ -148,6 +148,10 @@ def log_server_message(window: sublime.Window, prefix: str, message: str) -> Non
         return
     window_id = window.id()
     WindowPanelListener.server_log_map[window_id].append((prefix, message))
+    list_len = len(WindowPanelListener.server_log_map[window_id])
+    if list_len >= SERVER_PANEL_MAX_LINES:
+        # Trim leading items in the list, leaving only the max allowed count.
+        del WindowPanelListener.server_log_map[window_id][:list_len - SERVER_PANEL_MAX_LINES]
```
rwols, line 154: I did not know this syntax would remove a part of a list...

rwols (approved):

Issue #1456, opened 2020-11-15
[Windows] It's not possible to update running server


rchl:
On Windows files are locked when in use so on updating to new version, when trying to delete old `node_modules` directory an exception is thrown and server can be broken until restart.

I think we would need to be able to explicitly shut down server to fix that.

rchl:
Transferred this to LSP repo.

rchl:
Perhaps we should check whether the file/directory is locked (using some os API) and skip updating the server in that case. Then the server would only be updated on next start of the server (which could mean the next ST start) which should be OK.

jfcherng:
How about

1. `server_dir/` renames to `server_dir@bak/` (this should work on Windows)
2. Construct the new `server_dir/`
3. Restart the server
4. Delete `server_dir@bak/` maybe on the next ST starting

rchl:
Not having to deal with renaming and then cleaning up later would definitively be simpler. And if that means that the server would get updated one restart later then it shouldn't be such a big sacrifice.

Also, not sure you can actually rename directory that contains locked files. At least ChatGPT seems to think that you can't.

jfcherng:
> Also, not sure you can actually rename directory that contains locked files. At least ChatGPT seems to think that you can't.

I am not sure what makes a dir not rename-able on Windows. Sometimes it can be while sometimes it can't be. Both dirs have running executable in it. Both of them are not able to be deleted.

rchl:
Interestingly, when testing in Windows 11 Pro Arm64, I'm not able to delete or rename the `LSP-ruff/.venv` directory from File Explorer but lsp_utils code can do both without triggering errors.

That said, people are clearly [having problems with updates](https://github.com/sublimelsp/LSP-ruff/issues/154) so not sure what this depends on.

@vwheeler63

vwheeler63:
Is there a way, during an update, to have the server shut down?

rchl:
It is being shut down but the timing is tricky as mentioned earlier in this thread. It's because unloading server is asynchronous and update happens from a different thread.

Just curious, which version of Windows are you running?

vwheeler63:
> It is being shut down but the timing is tricky as mentioned earlier in this thread. It's because unloading server is asynchronous and update happens from a different thread.
> 
> Just curious, which version of Windows are you running?

Windows 10 x64

[...]

PR #1638, opened 2021-04-15
Add plugin APIs: "on_*_capability" and "on_session_end"

rchl:
I have a local implementation of `didChangeWatchedFiles` that I need those APIs for.

 - `on_register_capability_async` to be able to handle `workspace.didChangeWatchedFiles` capability that can specify `watchers` to set up.
 - `on_unregister_capability_async` to unregister them
 - `on_session_end_async` to unregister watchers when session ends.

I have implemented `didChangeWatchedFiles` in `lsp_utils` but not sure if I want to upstream it. It depends on `node` and I'm not sure I have finished it (have done it some time ago and just kept using it as it works ;)). With that in mind, I'm fine with the decision not to integrate this API but if it makes sense in general and doesn't hurt then why not.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -593,6 +593,17 @@ class AbstractPlugin(metaclass=ABCMeta):
         """
         pass
 
+    def on_register_capability_async(self, registration_id: str, capability_path: str, options: Dict[str, Any]) -> None:
+        pass
+
+    def on_unregister_capability_async(
+        self, registration_id: str, capability_path: str, options: Dict[str, Any]
+    ) -> None:
+        pass
+
+    def on_session_end_async(self) -> None:
+        pass
+
 
 _plugins = {}  # type: Dict[str, Tuple[Type[AbstractPlugin], SettingsRegistration]]
 
@@ -1136,6 +1147,10 @@ class Session(TransportCallbacks):
                     # Inform only after the response is sent, otherwise we might start doing requests for capabilities
                     # which are technically not yet done registering.
                     sublime.set_timeout_async(inform)
+            if self._plugin:
+                inform = functools.partial(
+                    self._plugin.on_register_capability_async, registration_id, capability_path, options)
+                sublime.set_timeout_async(inform)
         self.send_response(Response(request_id, None))
 
     def m_client_unregisterCapability(self, params: Any, request_id: Any) -> None:
@@ -1153,6 +1168,10 @@ class Session(TransportCallbacks):
                 if isinstance(discarded, dict):
                     for sv in self.session_views_async():
                         sv.on_capability_removed_async(registration_id, discarded)
+            if self._plugin:
+                inform = functools.partial(
+                    self._plugin.on_unregister_capability_async, registration_id, capability_path, registration_path)
+                sublime.set_timeout_async(inform)
         self.send_response(Response(request_id, None))
 
     def m_window_showDocument(self, params: Any, request_id: Any) -> None:
@@ -1232,7 +1251,9 @@ class Session(TransportCallbacks):
         if self.exiting:
             return
         self.exiting = True
-        self._plugin = None
+        if self._plugin:
+            self._plugin.on_session_end_async()
+            self._plugin = None
         for sv in self.session_views_async():
             sv.shutdown_async()
         self.capabilities.clear()
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -594,14 +594,38 @@ class AbstractPlugin(metaclass=ABCMeta):
         pass
 
     def on_register_capability_async(self, registration_id: str, capability_path: str, options: Dict[str, Any]) -> None:
+        """
+        Notifies about server dynamically registering a capability using the client/registerCapability request.
+        This API is triggered on async thread.
+
+        :param registration_id: The registration identifier
+        :param capability_path: The registration capability path
+        :param options: The registration options
+        """
         pass
 
     def on_unregister_capability_async(
         self, registration_id: str, capability_path: str, options: Dict[str, Any]
     ) -> None:
+        """
+        Notifies about server un-registering a capability using the client/unregisterCapability request.
+        This API is triggered on async thread.
+
+        :param registration_id: The registration identifier
+        :param capability_path: The registration capability path
+        :param options: The registration options
+        """
         pass
 
+
     def on_session_end_async(self) -> None:
+        """
+        Notifies about the session ending (also if the session has crashed). Provides an opportunity to clean up
+        any stored state or delete references to the session or plugin instance that would otherwise prevent the
+        instance from being garbage-collected. If the plugin hasn't crashed, a shutdown message will be send immediately
+        after this method returns.
+        This API is triggered on async thread.
+        """
         pass
 
 
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -617,7 +617,6 @@ class AbstractPlugin(metaclass=ABCMeta):
         """
         pass
 
-
     def on_session_end_async(self) -> None:
         """
         Notifies about the session ending (also if the session has crashed). Provides an opportunity to clean up
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1161,6 +1161,8 @@ class Session(TransportCallbacks):
             capability_path, registration_path = method_to_capability(unregistration["method"])
             debug("{}: unregistering capability:".format(self.config.name), capability_path)
             data = self._registrations.pop(registration_id, None)
+            if data and self._plugin:
+                self._plugin.on_unregister_capability_async(registration_id, capability_path, data.options)
             if data and not data.selector:
                 discarded = self.capabilities.unregister(registration_id, capability_path, registration_path)
                 # We must inform our SessionViews of the removed capabilities, in case it's for instance a hoverProvider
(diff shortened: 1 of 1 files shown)


plugin/core/sessions.py
```diff
@@ -1136,6 +1147,10 @@ def m_client_registerCapability(self, params: Any, request_id: Any) -> None:
                     # Inform only after the response is sent, otherwise we might start doing requests for capabilities
                     # which are technically not yet done registering.
                     sublime.set_timeout_async(inform)
+            if self._plugin:
+                inform = functools.partial(
+                    self._plugin.on_register_capability_async, registration_id, capability_path, options)
+                sublime.set_timeout_async(inform)
```
rchl, line 1153: Not sure now why I've decided to do it from `set_timeouta_async`. Maybe not to delay the response but it shouldn't really be needed, now that I think of it.

plugin/core/sessions.py
```diff
@@ -593,6 +593,17 @@ def on_pre_send_notification_async(self, notification: Notification) -> None:
...
+    def on_register_capability_async(self, registration_id: str, capability_path: str, options: Dict[str, Any]) -> None:
+        pass
+
+    def on_unregister_capability_async(
+        self, registration_id: str, capability_path: str, options: Dict[str, Any]
+    ) -> None:
+        pass
+
+    def on_session_end_async(self) -> None:
+        pass
```
rchl, line 605: Should add documentation if we decide to go forward with it.

predragnikolic:
Nice that you got it working.

> I have implemented didChangeWatchedFiles in lsp_utils but not sure if I want to upstream it.

IMO the LSP repo looks like a more suitable place for the didChangeWatchedFiles feature. Because the default LSP clients configuration and all LSP-* plugins will benefit with this feature.

Implementing it in lsp_utlis kind of gives me a hint that in only support node based LSP-* plugins. 

-----
BTW, the change in this PR are ok, 

> What do you thinking of creating a python class that would spawn a rust process that uses [rs-notify](https://docs.rs/crate/notify/4.0.16). The python class will listen on the STDOUT of the rust process which would return what files have been deleted, created, modified. That way node is not required. This is just a question, but I think it was worth sharing.

rchl:
> Implementing it in lsp_utlis kind of gives me a hint that in only support node based LSP-* plugins.

It's implemented in the most base class in lsp_utils so it supports all `lsp-utils`-based plugins (node or not node)

> What do you thinking of creating a python class that would spawn a rust process that uses rs-notify. The python class will listen on the STDOUT of the rust process which would return what files have been deleted, created, modified. That way node is not required. This is just a question, but I think it was worth sharing.

That's how it's implemented but using `node` and https://github.com/paulmillr/chokidar package.
I feel that node is more "portable" in this case since lsp_utils can install node on-demand for all any platform.
For rust, I'm not even sure how we would need to install the binary. Bundle it or require the user to install the crate himeselv? Soudns less practical...

rchl:
> IMO the LSP repo looks like a more suitable place for the didChangeWatchedFiles feature. Because the default LSP clients configuration and all LSP-* plugins will benefit with this feature.

Yes, but for the implementation to be in the LSP repo, I think we need to have some proper API from ST. Wouldn't want to have official implementation that depends on external binary/framework.

plugin/core/sessions.py
predragnikolic: Note, 
The `client/registerCapability` is sent from the server to the client, 
and the response for this request is void. (same for `client/unregisterCapability`)

rchl: OK, but I'm not sure I understand the point of this note. :)

predragnikolic:
If these changes will unblock LSP from having didChangeWatch implemented for all plugins that use lsp_utils, I am happy to have that. 
 
----
Just to reply to this: (but mostly ignore it, as it is not important for now)
> For rust, I'm not even sure how we would install the binary. Bundle it for all platforms or require the user to install the crate himself? Sounds less practical...

1. Write a rust program that will emit data over STDOUT.
2. Compile that file for each OS (MAC, Linux, Windows), result will be an executable file `./watcher`[.exe] . This means that the user does not need to install the crate himself, we already compiled it for them.  
3. From python have a pseudo code like this:
```
if (os.platform == 'Linux'):
    popen(['./wathcer'])
if (os.platform == 'Windows'):
    popen('[./watcher.exe]')
```
That way the user doesn't need to have Node. Performance is not affected, the rs-notify library has some debounce options, but we can implement the debouncing even on the python level, but doing it on the rust level would seem more logical.

plugin/core/sessions.py
predragnikolic: You wrote `"Maybe not to delay the response"`, 
So I took a look at the LSP spec and saw that the response is `void`. So I also started thinking that `set_timeout_async` is maybe not needed. 
... Hope it makes more sense now, if not just skip these messages :)

rchl:
OK, so you are suggesting bundling the executables. Not a big fan of that to be honest but maybe would be OK if the binaries are very small. Would also need to be able to compile for mac Arm for example, if there are no binaries publicly.

> That way the user doesn't need to have Node. Performance is (hopefully) not affected.

It would be easy to switch out node for rust binaries in my implementation, if there are good reasons for it.

rchl:
BTW. The performance cost would likely come from having to parse all the notifications and then trigger notifications for all those files. Not so much from running the watcher itself (as long as it's FSWatch-based). So I would expect pretty similar results whether it's node or rust.

predragnikolic:
> It would be easy to switch out node for rust binaries in my implementation, if there are good reasons for it.

Exactly :) 
You already have an implementation in node, and IMO it is better to have didChangeWatch files than not having it at all.

plugin/core/sessions.py
rchl: I see.
What I meant is that the blocking code in the plugin could delay sending of the void response. Probably either way is fine though.

plugin/core/sessions.py
```diff
@@ -1232,7 +1251,9 @@ def end_async(self) -> None:
         if self.exiting:
             return
         self.exiting = True
-        self._plugin = None
+        if self._plugin:
+            self._plugin.on_session_end_async()
+            self._plugin = None
```
rwols, line 1256: Wouldn't you be able to detect a session ending via the `__del__` of your `AbstractPlugin`?

plugin/core/sessions.py
rwols: This looks okay to me. We have to do it via `set_timeout_async` because otherwise the response to this `client/registerCapability` has not yet been sent back to the language server.

plugin/core/sessions.py
```diff
@@ -1153,6 +1168,10 @@ def m_client_unregisterCapability(self, params: Any, request_id: Any) -> None:
                 if isinstance(discarded, dict):
                     for sv in self.session_views_async():
                         sv.on_capability_removed_async(registration_id, discarded)
+            if self._plugin:
+                inform = functools.partial(
+                    self._plugin.on_unregister_capability_async, registration_id, capability_path, registration_path)
+                sublime.set_timeout_async(inform)
```
rwols, line 1174: This shouldn't be done via `set_timeout_async` because we need to be informed of an unregistration before the response is sent to the language server.

I see `client/registerCapability` as a kind of constructor/`__init__`, whereas its counterpart `client/unregisterCapability` is a destructor/`__del__`, so all operations need to be done in reverse. So in `client/registerCapability` we inform the `AbstractPlugin` after informing the SessionViews. Here we should do the callback before doing the callbacks for SessionViews.

plugin/core/sessions.py
rwols: Yeah. Although I'm not sure if `on_session_end_async` is really needed.

rwols (changes requested):

plugin/core/sessions.py
rchl: RIght. Probably that is what I was thinking too.

plugin/core/sessions.py
rchl: This is so that I can clear all references to the class so that the `__del__` is actually triggered.

That said, I have some bug in the current implementation that prevents it from being called, even with this API. Need to look into it.

As for this API vs. weak references, I kinda prefer being explicit because otherwise, it's easy to forget to add stuff that is not weak referenced and prevent destructor from being called, without realizing it.

But if you insist, I could get rid of it and change to weak references.

rchl: Found the issue with my code.

So then the question is: weak references or explicit API "on_session_end"?

I wouldn't like having to use weak references in this case. Using it internally (like within the LSP package) is fine but for an API it's not very explicit or clear. I'd need to make everything that stores reference to a class its method a weak pointer and then null-check every time it's used. If I have this API I know explicitly when things are available and when not (after `on_session_end_async`) and don't need extra null-checks everywhere. Also, an API like that is a good reminder to do things the right way to avoid leaks.

This would allow making a `session` pointer that would have an explicitly defined life rather than needing to use `weaksession` and having to null check it.

rchl: (In that case, I would also need to call it in case of a crash)

rwols: > Also, an API like that is a good reminder to do things the right way to avoid leaks.

I don't understand the context here, can you explain?

> So then the question is: weak references or explicit API "on_session_end"?

I don't know what's wisdom... I guess we can go ahead with `on_session_end`. Although I'm now wondering whether cyclic references are the cause of https://github.com/sublimelsp/LSP/issues/1456

rchl: > > Also, an API like that is a good reminder to do things the right way to avoid leaks.
> 
> I don't understand the context here, can you explain?

In my case, the plugin creates an instance of `FileWatcher` which is passed a function to the Plugin's method. This connection needs to be broken on session ending or the class will not be garbage-collected.

Or if I want to for instance create an instance of the ApiWrapper class (which takes plugin instance as an argument) and store it in `self.api`, for example, then I need to clear out that property on session ending or it will also block the plugin instance from being GC'ed.

All those can be solved by using weakref but I'm not a fan of solving it that way. Also, using weakref enforces its use down the call chain (for example in this case within the ApiWrapper class) which seems excessive. And what would I do if I would have to work with a third-party API that I can't change?

rwols: Alright, I’ll trust your judgement on it

rwols (approved):

PR #1641, opened 2021-04-17
docs: improve getting-started instructions

rchl:
- Remove mention of "LSP: Enable Language Server Globally" as setup
instructions enable them globally already and it doesn't apply to
LSP-* packages either.
- Don't mention the "PATH" as it doesn't apply to LSP-* packages.
Instead add a generic sentence linking to troubleshooting section.
- Remove links to LSP-* packages since it's redundant given that the
list of language servers already includes them and categorized by
programming language.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -1,7 +1,7 @@
 # Getting started
 
-1. Find a server for the programming language of your choice in the [list of language servers](language_servers.md) and follow its setup instructions.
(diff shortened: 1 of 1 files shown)


rchl:
Maybe should keep saying "language" instead of "programming language"... I guess not all languages supported by servers are programming languages

predragnikolic (approved):

rwols (approved):

PR #1647, opened 2021-04-22
Disable crashed servers in-memory instead of in project data

### LINKED ISSUE #1644: LSP dialog won't close when opening a single file and LSP failed to start
**Describe the bug**

When LSP fail to start a specific language server you're greeted with a dialog that informs you it will be disabled for the current window. As an example, this is the dialog I get when `gopls` fail to start.

```
Failed to start gopls - disabling for this window. Re-enable by running "LSP: Enable Language Server In Project" from the Command Palette.

--- Error: ---
[Errno 2] No such file or directory: 'gopls'
```

If you started Sublime Text using `subl <file>` and you get this dialog, it will go into loop. Pressing `OK` immediately show the same dialog again. However, if you started Sublime Text using `subl .` or already had an instance of Sublime Running (in a project), the behaves correctly.

**To Reproduce**
1. Quit Sublime Text
2. Open a file (that will enable LSP but make sure the language server will fail to start) using `subl <file>`
3. Wait until the dialog shows and press `OK`
4. The dialog should now be presented once again

**Expected behavior**

I would expect the dialog to only be presented once.

**Screenshots**

- OS: macOS  11.2.1
- Sublime Text version: 4102
- LSP version: 1.2.13
- Language servers used: gopls (probably related to any language server)

**Additional context**

Except from the Discord conversation:

Screenshot 2021-04-21 at 16 58 05

rchl:
Fixes #1644

### DIFF
--- a/plugin/core/configurations.py
+++ b/plugin/core/configurations.py
@@ -86,7 +86,8 @@ class WindowConfigManager(object):
         self._disabled_for_session.add(config_name)
 
     def _reenable_disabled_for_session(self, config_name: str) -> bool:
-        if config_name in self._disabled_for_session:
+        try:
             self._disabled_for_session.remove(config_name)
             return True
-        return False
+        except KeyError:
+            return False
--- a/plugin/core/configurations.py
+++ b/plugin/core/configurations.py
@@ -1,6 +1,6 @@
 from .logging import debug
 from .types import ClientConfig
-from .typing import Any, Generator, List, Dict
+from .typing import Any, Generator, List, Set, Dict
 from .workspace import enable_in_project, disable_in_project
 import sublime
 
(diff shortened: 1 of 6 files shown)


plugin/core/configurations.py
```diff
@@ -61,16 +62,31 @@ def update(self) -> None:
...
         self.update()
+
+    def _disable_for_session(self, config_name: str) -> None:
+        self._disabled_for_session.add(config_name)
+
+    def _reenable_disabled_for_session(self, config_name: str) -> bool:
+        if config_name in self._disabled_for_session:
+            self._disabled_for_session.remove(config_name)
+            return True
+        return False
```
rwols, line 92: This does two tree traversals ;) you can do it with one tree traversal by just doing the `.remove` call and catching the KeyError

rwols (changes requested):
tiny nitpick, looks good otherwise

plugin/core/configurations.py
rchl: Changed.

rwols (approved):

PR #1650, opened 2021-04-28
improve completion item typings

predragnikolic:
This PR tries to improve the completion item type.

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -99,7 +99,6 @@ class LspCompleteInsertTextCommand(LspCompleteCommand):
 class LspCompleteTextEditCommand(LspCompleteCommand):
 
     def run(self, edit: sublime.Edit, item: CompletionItem, session_name: Optional[str] = None) -> None:
-        # LspCompleteTextEditCommand command will only be run if "textEdit" exist on the completion item
         text_edit = item.get("textEdit")
         if not text_edit:
             return
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -3,9 +3,9 @@ import sublime_plugin
 import webbrowser
 from .core.logging import debug
 from .core.edit import parse_text_edit
-from .core.protocol import Request, InsertTextFormat, Range, CompletionItem, TextEdit
+from .core.protocol import Request, InsertTextFormat, Range, CompletionItem
 from .core.registry import LspTextCommand
-from .core.typing import Any, List, Dict, Optional, Generator, Union, cast
+from .core.typing import Any, List, Dict, Optional, Generator, Union
 from .core.views import FORMAT_STRING, FORMAT_MARKUP_CONTENT, minihtml
 from .core.views import range_to_region
 from .core.views import show_lsp_popup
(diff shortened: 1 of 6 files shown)


plugin/core/protocol.py
```diff
@@ -46,6 +46,11 @@ class SignatureHelpTriggerKind:
     ContentChange = 3
 
 
+class InsertTextMode:
+    AsIs = 1
```
predragnikolic, line 50: Should I use lowercase property names to follow the spec more closely? 

```ts
export namespace InsertTextMode {
	export const asIs: 1 = 1;
        export const adjustIndentation: 2 = 2;
```

Or should I leave them capitalized. (to be consistent with other class types like the property keys in: DiagnosticSeverity, DiagnosticTag, CompletionItemTag...)

plugin/completion.py
```diff
@@ -98,9 +98,10 @@ def run(self, edit: sublime.Edit, item: Any, session_name: Optional[str] = None)
 
 class LspCompleteTextEditCommand(LspCompleteCommand):
 
-    def run(self, edit: sublime.Edit, item: Any, session_name: Optional[str] = None) -> None:
-        text_edit = item["textEdit"]
-        new_text = text_edit['newText']
+    def run(self, edit: sublime.Edit, item: CompletionItem, session_name: Optional[str] = None) -> None:
+        # LspCompleteTextEditCommand command will only be run if "textEdit" exist on the completion item
+        text_edit = cast(TextEdit, item.get("textEdit"))
+        new_text = text_edit["newText"]
```
rchl, line 104: Would it not make sense to, instead of casting, add a null-check like:
```py
        if not text_edit:
            return
```
?
Even if currently all callers ensure that `textEdit` is set, it could save us in the feature if we do the check. And a cast always looks like a dirty hack.

plugin/core/protocol.py
```diff
@@ -171,7 +176,30 @@ class SignatureHelpTriggerKind:
...
+    'sortText': str,
+    'filterText': str,
+    'insertText': str,
+    'insertTextFormat': InsertTextFormat,
+    'insertTextMode': InsertTextMode,
+    'textEdit': TextEdit,
+    'additionalTextEdits': List[TextEdit],
+    'commitCharacters': str,
+    'command': Command,
+    'data': Any
```
rchl, line 201: Personally, I would favor alphabetical order but you don't have to follow my suggestion (unless @rwols thinks the same maybe).

plugin/core/protocol.py
```diff
@@ -171,7 +176,30 @@ class SignatureHelpTriggerKind:
...
+    'deprecated': bool,
+    'preselect': bool,
+    'sortText': str,
+    'filterText': str,
+    'insertText': str,
+    'insertTextFormat': InsertTextFormat,
+    'insertTextMode': InsertTextMode,
+    'textEdit': TextEdit,
+    'additionalTextEdits': List[TextEdit],
+    'commitCharacters': str,
```
rchl, line 199: This should be a `List[str]`

plugin/formatting.py
```diff
@@ -9,10 +9,11 @@
 from .core.views import text_document_range_formatting
 from .core.views import will_save_wait_until
 from .save_command import LspSaveCommand, SaveTask
+from .core.protocol import TextEdit
```
rchl, line 12: Not in the right order.

plugin/completion.py
predragnikolic: Yeah, I agree with you.

rchl (approved):

rwols (approved):

PR #1651, opened 2021-04-29
Resolve completions completion item before applying additional text edits

predragnikolic:
fixes https://github.com/sublimelsp/LSP-typescript/issues/46

Short description

This PR introduces resolving the completion item(if session supports resolving) before applying additional text edits and running the `epilogue`.

output

<details>
<summary>Longer description</summary>

# Change 1

**The current behavior:**

When a `LSP.CompletionItem` (*CI* for short) is passed to the `format_completion` function:

- if the *CI* contains a TextEdit,
it becomes a ST Command Completion Item.
When that *CI* is selected in the AC,
a `lsp_complete_text_edit` command is run passing in the `LSP.CompletionItem` and the `session_name` as arguments.
So we exactly know what `CI` was selected in the AC
and to what session to send the `CI` to resolve it(only if the session supports resolving completion items).

- else if the *CI* does not contains a TextEdit,
it becomes a ST Snippet Completion Item or a regular ST Completion Item.
When that *CI* is selected,
we don't know what **exact** completion item was selected or what session it belongs to, because ST.SnippetCompletionItem or ST.CompletionItem function a bit differently that Command Completion Items.

To conclude, IMO I see more benefits using Command Completion Items as they gives us the ability to always know what completion item was selected and what session the `CI` belongs.

**So in this PR:**
I change the `format_completion` to always return a command completion item. That way we always know what completion item was selected, what the session name is, and based on that we can insert TextEdits, or insert the label or newText, resolve completion items and apply additional text edits correctly.

# Change 2

**Current behavior:**
Additional text edits will not be applied if:
a) the unresolved `CI` don't have `additionalTextEdits`, but the resolved completion `CI` has `additionalTextEdits` . (which was the bug with the theia typescript lanugage server - https://github.com/sublimelsp/LSP-typescript/issues/46)
b) If we select ST Snippet Completion items or regular ST Completion items (no way to detect what exact `CI` was selected,
and currently there is no logic that applies additionalTextEdits for ST Snippet Completion items or regular ST Completion items.)

**Here is how this PR addresses those above mention issues.**

`format_completion_item` returns a Command Completion Item
When we select the Command Completion Item in the AC
a `lsp_select_completion_item` command is run.
We insert the text in the view(based on if it is a TextEdit, insertText, or label, Snippet or PlainText).
Than we run the `epilogue`,
where we will try to resolve the *CI* in the `epilogue` if the session supports resolving (this is a new change),
else if the session doesn't support resolving,
we will `applyAdditionalTextEdits` and everything else that was earlier already done in the `epilogue`.
</details>

-----

### TODO 

- [x] Fix failing test

There is currently one test that is failing and that is: 
```
FAIL: test_additional_edits (test_completion.QueryCompletionsTests)
  File "/home/predrag/.config/sublime-text/Packages/UnitTesting/unittesting/core/py33/case.py", line 28, in _executeTestPart
    yield from deferred
  File "/home/predrag/.config/sublime-text/Packages/LSP/tests/test_completion.py", line 377, in test_additional_edits
    expected_text='import asdf;\nasdf')
  File "/home/predrag/.config/sublime-text/Packages/LSP/tests/test_completion.py", line 76, in verify
    self.assertEqual(self.read_file(), expected_text)
AssertionError: 'asdf' != 'import asdf;\nasdf'
- asdf
+ import asdf;
asdf
```
The test is failing because of this [addition](https://github.com/sublimelsp/LSP/compare/try-to-resolve-completion-items-before-applying-additional-text-edits?expand=1#diff-1e31f37d39e534dca2dca7fd81d29d7e5a039a4b55a5765fdb4292c7a87760aaR105-R109):
If we comment it out the test will pass:
```py
    def epilogue(self, item: CompletionItem, session_name: str) -> None:
        # session = self.session_by_name(session_name, 'completionProvider.resolveProvider')
        # if session:
        #     request = Request.resolveCompletionItem(item, self.view)
        #     session.send_request_async(request, lambda response: self.on_resolved(response, session_name))
        # else:
        self.on_resolved(item, session_name)
```
My current conclusion of why the test is failing now is that now I added the check in the epilogue to see if session supports  `resolveProvider`. which when running tests will be `true`, but right now in the test I haven't added a mock response for the resolve completion item request, and thus the test failed.

Feel free to correct me, but I think that the test will pass if:
- we set the session in the test to not have the resolveCompletionItem capability
- or set the session to have the resolveCompletionItem capability and add a response for the resolveCompletion item request.


predragnikolic:
I think I put more effort to explain why I did things that the code itself :D 
If the text I wrote is too much, read the code.

plugin/core/views.py
```diff
@@ -718,34 +717,17 @@ def format_completion(
...
-            details=st_details)
-
+    args = {
+        "item": item,
+        "session_name": session_name
+    }  # type: Dict[str, Any]
+    completion = sublime.CompletionItem.command_completion(
+        trigger=st_trigger,
+        command="lsp_select_completion_item",
+        args=args,
```
rchl, line 727: We could save four lines with:

```suggestion
        args={"item": item, "session_name": session_name},
```

plugin/core/sessions.py
```diff
@@ -155,6 +155,9 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
                     "documentationFormat": ["markdown", "plaintext"],
                     "tagSupport": {
                         "valueSet": completion_tag_value_set
+                    },
+                    "resolveSupport": {
+                        "properties": ["detail", "documentation", "additionalTextEdits", "command"]
```
rchl, line 160: Maybe it doesn't make sense to specify `detail` and `documentation` since we resolve only after completing so we can't make use of those anyway.

plugin/completion.py
```diff
@@ -125,3 +99,26 @@ def translated_regions(self, edit_region: sublime.Region) -> Generator[sublime.R
             translation = region.b - primary_cursor_position
             translated_edit_region = sublime.Region(edit_region.a + translation, edit_region.b + translation)
             yield translated_edit_region
+
+    def epilogue(self, item: CompletionItem, session_name: str) -> None:
+        session = self.session_by_name(session_name, 'completionProvider.resolveProvider')
+        if session:
+            request = Request.resolveCompletionItem(item, self.view)
+            session.send_request_async(request, lambda response: self.on_resolved(response, session_name))
```
rchl, line 107: I think it would be good to jump back on the main thread before calling `on_resolved`.

rchl (changes requested):

tests/test_completion.py
```diff
@@ -3,6 +3,7 @@
 from LSP.plugin.core.views import format_completion
 from setup import TextDocumentTestCase
 import sublime
+from copy import deepcopy
```
rchl, line 6: order :)

predragnikolic: I always forget :)

plugin/core/sessions.py
predragnikolic: > Maybe it doesn't make sense to specify detail and documentation

For me also it doesn't make sense because only those two properties were allowed to be lazy loaded prior the 3.16 LSP spec release. So specifying the in the `resolveSupport` might seem redundant.

> we resolve only after completing so we can't make use of those anyway.

But I do not agree 100% with this one. We do resolve the completion item now in two cases.
- when we select the completion item(that is implemented with this PR)
- when we click the `More` link in the AC panel. (We resolve the completion item and grab the `detail` and `documentation ` of the resolved completion item, and that is the reason why I would also specify them in the `resolveSupport`). 

I feel like being explicit and leaving  `detail` and `documentation` in `resolveSupport` could not hurt us. 
What do you think?

boot.py
```diff
@@ -5,8 +5,7 @@
 
 # Please keep this list sorted (Edit -> Sort Lines)
 from .plugin.code_actions import LspCodeActionsCommand
-from .plugin.completion import LspCompleteInsertTextCommand
-from .plugin.completion import LspCompleteTextEditCommand
+from .plugin.completion import LspSelectCompletionItemCommand
```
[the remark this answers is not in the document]
predragnikolic: Damn it :D

plugin/core/sessions.py
rchl: Alright.

boot.py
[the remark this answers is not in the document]
rchl: Just an internal OCD linter.

plugin/completion.py
predragnikolic: Can you give me an answer of what will that change(moving to the main thread) improve(or fix)? (I am bad at threads)

rchl: `on_resolved` is expected to run on the main thread so it's more about the expectations than an actual issue.

Even though I don't remember the specific case, I remember that there were issues running LSP APIs or commands on async thread. Especially the ones that do some view edits.

predragnikolic: Thanks, done with 78fe8bb

predragnikolic:
I noticed one thing that I wasn't expecting. 

I moved the additional text edits from the completion item to the resolved completion item. And the test started to fail... I was not expecting that.

```diff
 def test_additional_edits_if_session_has_the_resolve_capability(self) -> 'Generator':
        completion_item = {
            'label': 'asdf'
-            'additionalTextEdits': [
-                {
-                    'range': {
-                        'start': {
-                            'line': 0,
-                            'character': 0
-                        },
-                        'end': {
-                            'line': 0,
-                            'character': 0
-                        }
-                    },
-                    'newText': 'import asdf;\n'
-                }
-            ]
        }
        self.set_response("completionItem/resolve", {
+            'label': 'asdf',
+            'additionalTextEdits': [
+                {
+                    'range': {
+                        'start': {
+                            'line': 0,
+                            'character': 0
+                        },
+                        'end': {
+                            'line': 0,
+                            'character': 0
+                        }
+                    },
+                    'newText': 'import asdf;\n'
+                }
+            ]
        })
        yield from self.verify(
            completion_items=[completion_item],
            insert_text='',
            expected_text='import asdf;\nasdf')
```

plugin/completion.py
```diff
@@ -125,3 +99,29 @@ def translated_regions(self, edit_region: sublime.Region) -> Generator[sublime.R
...
+        else:
+            self.on_resolved(item, session_name)
+
+    def on_resolved(self, item: CompletionItem, session_name: str) -> None:
+        def run_main() -> None:
```
rchl, line 112: This `run_main` doesn't really belong here. The expectation is that this function is called on the main thread but it shouldn't be the responsibility of that function to do that.

I would expect something likes this rather, in the epilogue:

```py
            def resolve_on_main_thread(item: CompletionItem, session_name: str) -> None:
                sublime.set_timeout(lambda: self.on_resolved(item, session_name))

            request = Request.resolveCompletionItem(item, self.view)
            session.send_request_async(request, lambda response: resolve_on_main_thread(response, session_name))
```

plugin/completion.py
```diff
@@ -102,26 +102,27 @@ def translated_regions(self, edit_region: sublime.Region) -> Generator[sublime.R
...
+        def resolve_on_main_thread(item: CompletionItem, session_name: str) -> None:
+            sublime.set_timeout(lambda: self.on_resolved(item, session_name))
+
         if session:
             request = Request.resolveCompletionItem(item, self.view)
-            session.send_request_async(request, lambda response: self.on_resolved(response, session_name))
+            session.send_request_async(request, lambda response: resolve_on_main_thread(response, session_name))
         else:
-            self.on_resolved(item, session_name)
+            resolve_on_main_thread(item, session_name)
```
rchl, line 113: Isn't this one already on the main thread and doesn't need to go through set_timeout?

predragnikolic: Correct

rchl (dismissed):

plugin/core/sessions.py
rwols: Here's what vscode does: https://github.com/microsoft/vscode-languageserver-node/blob/d58c00bbf8837b9fd0144924db5e7b1c543d839e/client/src/common/client.ts#L1598-L1600

predragnikolic: Should I remove the `command`?

rwols: Yeah, let's keep it at most as complicated as vscode :)

tests/test_completion.py
```diff
@@ -353,26 +355,30 @@ def test_filter_text_is_not_a_prefix_of_label(self) -> 'Generator':
...
-                            },
-                            'end': {
-                                'line': 0,
-                                'character': 0
-                            }
+    def test_additional_edits_if_session_has_the_resolve_capability(self) -> 'Generator':
+        completion_item = {
+            'label': 'asdf'
+        }
+        self.set_response("completionItem/resolve", {
```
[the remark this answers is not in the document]
predragnikolic: Thanks

plugin/core/sessions.py
predragnikolic: I guess that it wont hurt. 

The reason on why I also added the command is that we use  "additionalTextEdits", "command" after we resolve the completion item. So for our use case it would have been ok to announce that we support lazy loading for the `command` also 

https://github.com/sublimelsp/LSP/pull/1651/files#diff-1e31f37d39e534dca2dca7fd81d29d7e5a039a4b55a5765fdb4292c7a87760aaR116

https://github.com/sublimelsp/LSP/pull/1651/files#diff-1e31f37d39e534dca2dca7fd81d29d7e5a039a4b55a5765fdb4292c7a87760aaR120

But I will remove it.

predragnikolic: 843dcf6

plugin/completion.py
```diff
@@ -125,3 +99,30 @@ def translated_regions(self, edit_region: sublime.Region) -> Generator[sublime.R
...
+
+    def epilogue(self, item: CompletionItem, session_name: str) -> None:
+        session = self.session_by_name(session_name, 'completionProvider.resolveProvider')
+
+        def resolve_on_main_thread(item: CompletionItem, session_name: str) -> None:
+            sublime.set_timeout(lambda: self.on_resolved(item, session_name))
+
+        if session:
+            request = Request.resolveCompletionItem(item, self.view)
+            session.send_request_async(request, lambda response: resolve_on_main_thread(response, session_name))
```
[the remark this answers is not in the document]
predragnikolic: > We don't need to resolve if there's already additionalTextEdits present, no? 

Done with 4fc254e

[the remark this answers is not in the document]
rwols: It’s not that important for me :)

rwols (approved):

PR #1653, opened 2021-05-01
Change to pylsp vs pyls

s7726:
Palantir is no longer maintaining the project. It has been forked and taken up by a new group.

The final link to the complete options list will need to be updated when the new group determines their chosen documentation path. 
https://github.com/python-lsp/python-lsp-server/issues/24
The existing link should suffice for now since currently, the project is a drop-in replacement.


rwols (approved):

rwols:
Thanks!

Issue #1562, opened 2021-01-28
The code action is not displayed properly


husanjun:
**Describe the bug**
`code action` on the right is not displayed properly

**Screenshots**
![image](https://user-images.githubusercontent.com/49985266/106103227-909cd100-617b-11eb-80c3-807f4e79a234.png)
![image](https://user-images.githubusercontent.com/49985266/106103271-a4e0ce00-617b-11eb-8c6e-3f39aeecaab9.png)

**Environment (please complete the following information):**
- OS: win7
- Sublime Text version: 4095
- LSP version: 1.2.8
- Language servers used: intelephense

rchl:
This looks like a bug that belongs to ST bug tracker rather.
You should report it at https://github.com/sublimehq/sublime_text with extra information like the ST settings that you have overridden (font, line-height and stuff that might be related).

husanjun:
Hi @rchl ,
I'm not sure if it is `sublime` problem, but I submitted a PR that can solve this problem
#1663

husanjun:
It's a theme problem, Thanks @rchl @rwols @jwortmann

PR #1663, opened 2021-05-13
Fixed `code action` display issues

husanjun:
#1562


rwols:
Can you show a screenshot?

rwols:
I took the liberty of checking out your branch. It now looks like this

image

You should also fix up the rendering of the code lenses: https://github.com/sublimelsp/LSP/blob/82120a8b67124c22ea01d2723c6a63a3150f5936/plugin/documents.py#L528-L530

jwortmann:
While we're at it, I see the code action annotation uses a hardcoded color `#2196F3`, but the code lenses uses a green color from the color scheme and furthermore allows color schemes to override that color. Can we use the latter approach for all annotations?

Suggestion with slightly tweaked scope for the code lenses:
```diff
@@ -499,7 +499,7 @@
             suffix = 's' if action_count > 1 else ''
             code_actions_link = make_command_link('lsp_code_actions', '{} code action{}'.format(action_count, suffix))
             annotations = [annotations_template.format(text=code_actions_link)]
-            annotation_color = '#2196F3'
+            annotation_color = self.view.style_for_scope("region.bluish markup.accent.codeaction.lsp")["foreground"]
         self.view.add_regions(self.CODE_ACTIONS_KEY, regions, scope, icon, flags, annotations, annotation_color)
 
     def _clear_code_actions_annotation(self) -> None:
@@ -539,7 +539,7 @@
         else:
             annotation = "..."
         annotation = '<div class="codelens">{}</div>'.format(annotation)
-        accent = self.view.style_for_scope("region.greenish markup.codelens.accent")["foreground"]
+        accent = self.view.style_for_scope("region.greenish markup.accent.codelens.lsp")["foreground"]
         self.view.add_regions(self._code_lens_key(index), [region], "", "", 0, [annotation], accent)
 
     def _do_code_lenses_async(self) -> None:
```

And here is another suggestion for the lightbulb color:
```diff
@@ -480,7 +480,7 @@
         annotations = []
         annotation_color = ""
         if userprefs().show_code_actions == 'bulb':
-            scope = 'markup.changed'
+            scope = 'region.yellowish lightbulb.lsp'
             icon = 'Packages/LSP/icons/lightbulb.png'
         else:  # 'annotation'
             annotations_template = """
```

plugin/documents.py
```diff
@@ -483,9 +483,22 @@ def _on_code_actions(self, responses: CodeActionsByConfigName) -> None:
             scope = 'markup.changed'
             icon = 'Packages/LSP/icons/lightbulb.png'
         else:  # 'annotation'
+            annotations_template = """
+            <body id="lsp-line-annotation">
```
rchl, line 487: The other stylesheets (popup.css and an sheets.css) use `.lsp_popup` and `.lsp_sheet` respectively. Would like this to follow that convention which means it should be a class like `lsp_code_action`, for example.

plugin/documents.py
```diff
@@ -483,9 +483,22 @@ def _on_code_actions(self, responses: CodeActionsByConfigName) -> None:
...
+                <style>
+                    html, body {{
+                        font-family: system;
+                        margin: 0;
+                        padding: 0;
+                    }}
```
rchl, line 493: This creates this weird behavior on hovering where the bounding box gets bigger:
https://user-images.githubusercontent.com/153197/118176061-34f80a00-b431-11eb-8123-9e664782b84b.mov

It doesn't happen by default.

rchl (changes requested):

plugin/documents.py
jwortmann: This PR follows the best practices for minihtml content as described in https://www.sublimetext.com/docs/minihtml.html#best_practices.
For the simple annotations here it is probably not very relevant, but *in theory* this allows color schemes to tweak the CSS, for example to adjust the background color.

Afaik the `.lsp_popup` class is only used if the minihtml content is passed to mdpopups, which circumvents the intended way and forces its styles over color schemes anyway.

The color boxes at https://github.com/sublimelsp/LSP/blob/0a3ad53d56dc8e628ee051eb9a903ccf2e8b4193/plugin/core/views.py#L553-L561 are another example similar to this PR where LSP doesn't use mdpopups and follows the best practices to allow color schemes to tweak styles if desired.

rchl: OK, so let's at least change to `lsp-code-action` because "annotation" is just an implementation detail.

jwortmann: If you apply the fix for code lenses too, as suggested by rwols above, this HTML template can probably be used for both of them. Maybe then `lsp-annotation` would actually be okay?

husanjun:
Sorry, I can't test `code lens`, can someone help test it? I have submitted the modification

rwols:
One example of a language server that shows code lenses is OmniSharp:

image

It seems to render fine from my point of view.

rwols:
Before picture:

image

I don't quite understand what this solves ;) must be theme-related I guess.

rwols (approved):
LGTM. But I'll let it depend on @rchl.

jwortmann:
According to the screenshots in the linked issue there were some problems with the annotation line-height before. I could indeed imagine that this was related to the theme or color scheme, if one of it accidentally applied some additional padding or something similar. Would be interesting to know which theme and color scheme @husanjun uses. Perhaps this might really be a Sublime issue in case it is not intended that e.g. the `phantom_css` gets applied to annotations as well, or so?

Beyond that, I think this PR does the following improvements:
* the more condensed proportional font looks nicer (in my opinion) and saves some space
* the blue border fits to the color scheme now
* it can be styled from the user/color scheme if desired

Something I'm not sure about is whether `<div style="vertical-align: middle">` has any effect. I can't find that property in the [minihtml docs](https://www.sublimetext.com/docs/minihtml.html#css).

plugin/documents.py
```diff
@@ -139,6 +139,20 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
...
     code_lenses_debounce_time = FEATURES_TIMEOUT + 2000
 
+    ANNOTATIONS_TEMPLATE = """
+    <body id="lsp-annotation">
+        <style>
+        html, body {{
+            padding: 0;
+            margin: 0;
+            font-family: system;
+            height:{height}
```
rchl, line 149: I don't think this does anything. `view.line_height()` is in pixels and you don't have the unit here.
But if it was meant to fix the https://github.com/sublimelsp/LSP/pull/1663#discussion_r632047619 issue then it doesn't succeed in any case.

plugin/documents.py
```diff
@@ -139,6 +139,20 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
...
+        }}
+        </style>
+        <div style="vertical-align: middle">{text}</div>
```
rchl, line 152: As @jwortmann said, minihtml doesn't support that.

husanjun:
> Before picture:
> 
> image
> 
> I don't quite understand what this solves ;) must be theme-related I guess.

Thank you for your help. You're right. I'm sure it's the theme/color scheme issue, I use the default theme is fine

husanjun:
> According to the screenshots in the linked issue there were some problems with the annotation line-height before. I could indeed imagine that this was related to the theme or color scheme, if one of it accidentally applied some additional padding or something similar. Would be interesting to know which theme and color scheme @husanjun uses. Perhaps this might really be a Sublime issue in case it is not intended that e.g. the `phantom_css` gets applied to annotations as well, or so?
> 
> Beyond that, I think this PR does the following improvements:
> 
> * the more condensed proportional font looks nicer (in my opinion) and saves some space
> * the blue border fits to the color scheme now
> * it can be styled from the user/color scheme if desired
> 
> Something I'm not sure about is whether `<div style="vertical-align: middle">` has any effect. I can't find that property in the [minihtml docs](https://www.sublimetext.com/docs/minihtml.html#css).

Thank you for your tips

husanjun:
I will close the PR now, please let me know if it is necessary to change the font or color scheme, and I will submit the PR separately

PR #1665, opened 2021-05-16
Fix apply wrong region for code action

chendesheng:
The codefix for `remove unused opens` in fsharp language server hard coded `range.character` to 1000: https://github.com/fsharp/FsAutoComplete/blob/master/src/FsAutoComplete/CodeFixes/UnusedOpens.fs

This is actually legit according to the LSP spec, having `range.character` greater than the line length. (https://microsoft.github.io/language-server-protocol/specification#position)

--------------------

I'm not confident with python and the LSP codebase. And it seems this is like the core part of the codebase.  Feel free to reject and changes in a better way.

### DIFF
--- a/plugin/edit.py
+++ b/plugin/edit.py
@@ -33,7 +33,7 @@ class LspApplyDocumentEditCommand(sublime_plugin.TextCommand):
                 if version is not None and version != view_version:
                     debug('ignoring edit due to non-matching document version')
                     continue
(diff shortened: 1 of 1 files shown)


rwols:
Many thanks for this fix. I have created another commit that adds a unit test for this and adds you as a co-author: https://github.com/sublimelsp/LSP/pull/1666

PR #1669, opened 2021-05-20
Small tweaks for annotations and lightbulb colors

jwortmann:
This are some of the changes from the closed PR #1663 which make color scheme authors (and possibly users) happier. It affects only the colors, so there shouldn't be any unintended side affects like strange bounding box behavior on hover or so.

Short explanation regarding the scopes:
- The `markup.changed` which was previously used for the lightbulb icon is semantically incorrect and could be any color in theory. The scope `lightbulb` (can be used for potential color adjustments) on top of `region.yellowish` isn't probably used in any scheme, but that scope is comparable to the scopes `bookmarks` and `mark` which are used for the gutter icons by the internal "Goto > Bookmarks" and "Edit > Mark" features.
- If anyone used the former `markup.codelens.accent` scope for a color customization, they will need to update it now, but I think the way I've adjusted it here makes more sense. Maybe this is worth a small mention in the next update notes.

### DIFF
--- a/docs/src/customization.md
+++ b/docs/src/customization.md
@@ -74,8 +74,9 @@ Diagnostics will also optionally include the following scopes:
 | `entity.name.function.sighelp.lsp` | Function name in the signature help popup |
 | `variable.parameter.sighelp.lsp` | Function argument in the signature help popup |
 
-### Code Lens
+### Annotations
 
 | scope | description |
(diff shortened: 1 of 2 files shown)


rwols (approved):

rwols:
Thanks!

PR #1672, opened 2021-05-21
Update for ST4

rwols:
- Remove ST3 example configs.
- Redirect to install instructions of various new ST4 packages.

### DIFF
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -9,6 +9,8 @@ If there are no setup steps for a language server on this page, but a [language 
 !!! tip "We recommend installing [LSP-json](https://packagecontrol.io/packages/LSP-json)."
     [LSP-json](https://packagecontrol.io/packages/LSP-json) provides completions and diagnostics when editing JSON files that adhere to a JSON schema.
 
+!!! For legacy ST3 docs, see <a href="https://lsp.readthedocs.io">lsp.readthedocs.io</a>.
+
 
 ## Angular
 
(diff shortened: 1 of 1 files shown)


rchl (dismissed):

predragnikolic:
@rwols
> Because these plugins support only ST 4.
We should document them similar/consistent like the latex docs. For the ST4 the we will link to the plugin, but for ST3 we will leave the client configuration.
For LSP-Source​Kit we should specify that is Mac only (somewhere).

What do you think?

rwols:
Well, ST3 users can look at lsp.readthedocs.io still. In any case, there's little reason to still use ST3, and little reason for us to maintain double configs.

> For LSP-Source​Kit we should specify that is Mac only (somewhere).

That's noted in [its readme](https://github.com/sublimelsp/LSP-SourceKit).

rwols:
Moreover LSP-json will also point out deprecated keys :)

predragnikolic:
IMO removing the need to document ST 3 flows would ease the maintenance of the docs.
The only reasons why we document both flows for ST 4 and ST 3 is because in some conversation we had in some PR, we concluded that we want to keep both configuration.

Now that ST 4 is released, does this mean that we do not longer want to maintain double configs? @rwols @rchl

predragnikolic:
I am more than happy if the https://lsp.sublimetext.io/ documents only LSP for ST 4 features.
If necessary, we can put a paragraf on the readme for ST 3 to see documentation at  lsp.readthedocs.io

docs/src/language_servers.md
```diff
@@ -453,7 +197,6 @@ Follow installation instructions on [LSP-json](https://github.com/sublimelsp/LSP
 
 3. Open `Preferences > Package Settings > LSP > Settings` and add the `"julials"` client configuration to the `"clients"`:
 
-=== "Sublime Text 4"
     ```json
     {
         "clients": {
```
predragnikolic, line 202: Re-indent to the left each of these lines?

rwols: I was under the assumption that this indent was needed so that it is part of the ordered list. Is this not needed?

rwols:
I think everything is addressed here.

Issue #892, opened 2020-02-26
Diagnostic don't refresh in Vue files after types are changed


rchl:
- OSX
- LSP v0.9.5
- LSP-Vue v0.0.14

Vue project with TS configuration in`tsconfig.json` and custom types in `types/index.d.ts`.
Editing types in `types/index.d.ts` doesn't get reflected in *.vue file if I'm referring to those types there.
It works in plain *.js files (those pick up changes to types immediately).

My workaround is to trigger "LSP: Restart servers" command every time I change types.
No problem in VSCode.

rchl:
I'll create test project and investigate more when I have time. Just wanted to log this as it bothered me for a while now.

rwols:
My first instinct, without having used vue-lsp, is that we don't support workspace/didChangeWatchedFiles.

rchl:
Yes, you are most likely right.

I've set a project with instructions at https://github.com/rchl/test-sublime-lsp-type-bug

When checking VSCode, I can see it triggering
```
[Trace - 9:55:53 PM] Sending notification 'workspace/didChangeWatchedFiles'.
Params: {
    "changes": [
        {
            "uri": "file:///Users/me/minimal/types/index.d.ts",
            "type": 2
        }
    ]
}
```
right after I make a change in `types/index.d.ts`.

The fact that the pure JS or TS file sees the changes immediately is probably because for that file LSP-vue is actually not used. Just a pure typescript server is, instead.

rwols:
See: https://github.com/sublimelsp/LSP/issues/516, and https://github.com/sublimehq/sublime_text/issues/2669.

rchl:
Extra note:
While server can request watcher to be registered using `DidChangeWatchedFilesRegistrationOptions`, this is not how Vetur (used by LSP-vue) does it at least (yet?).

Instead, its VSCode extension part (since vetur is composed of server and vscode extension) registers watcher at https://github.com/vuejs/vetur/blob/8b82c92a232086167af2969b42a6cc0218cf5c2e/client/client.ts#L53 .

So which files are to be watched is server-specific and so that would have to be decided per-server (in this case in `LSP-vue` package).

ayoub-benali:
Does it make sense to have LSP provide a watcher API that language handlers can hook into ? 
Because it is a similar missing features for Metals integration.

rchl:
[...]

PR #1674, opened 2021-05-22
didChangeWatchedFiles using watchman

tomv564:
This is a proposal for an (interim?) solution for users who want file change notifications to their servers using facebook watchman before a potential API for this arrives in Sublime (so, issue #892)

I know Raoul commented that having a "hard" dependency (or a python package or a binary) would introduce a fair bit of maintenance pain - my thought would be to enable the support if the user opts-in and has watchman on their path for that reason.
It uses a named socket, so it is otherwise similar to TCP communication to language servers.

This was only partially tested using Metals which has a pretty nice set of globs it requests.
It would need quite a bit more testing:
* Windows
* A bunch of servers (javascript/typescript servers are probably important, any other likely heavy users?)
* Ensure watchers and watchman are correctly being removed when sessions, LSP or sublime are shut down.

### DIFF
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -382,6 +382,10 @@ class Notification:
         return Notification("workspace/didChangeWorkspaceFolders", params)
 
     @classmethod
+    def didChangeWatchedFiles(cls, params: dict) -> 'Notification':
+        return Notification("workspace/didChangeWatchedFiles", params)
+
+    @classmethod
     def exit(cls) -> 'Notification':
         return Notification("exit")
 
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -259,7 +259,7 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
                 "dynamicRegistration": True
             },
             "didChangeWatchedFiles": {
-                "dynamicRegistration": True
+                "dynamicRegistration": Watcher.is_available()
             },
             "executeCommand": {},
             "workspaceEdit": {
@@ -1190,6 +1190,8 @@ class Session(TransportCallbacks):
             capability_path, registration_path = method_to_capability(unregistration["method"])
             debug("{}: unregistering capability:".format(self.config.name), capability_path)
             data = self._registrations.pop(registration_id, None)
+            if capability_path == "didChangeWatchedFilesProvider":
+                Watcher.unregister(registration_id)
             if data and self._plugin:
                 self._plugin.on_unregister_capability_async(registration_id, capability_path, data.options)
             if data and not data.selector:
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -39,6 +39,7 @@ from .views import COMPLETION_KINDS
 from .views import extract_variables
 from .views import get_storage_path
 from .views import SYMBOL_KINDS
+from .watcher import Watcher
 from .workspace import is_subpath_of
 from abc import ABCMeta
 from abc import abstractmethod
@@ -155,9 +156,6 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
                     "documentationFormat": ["markdown", "plaintext"],
                     "tagSupport": {
                         "valueSet": completion_tag_value_set
-                    },
-                    "resolveSupport": {
-                        "properties": ["detail", "documentation", "additionalTextEdits"]
                     }
                 },
                 "completionItemKind": {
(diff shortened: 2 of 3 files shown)


rchl:
Haven't had a proper look at your changes yet but would just say that I have a "competing" solution that I've been using privately and is based on chokidar node package. It's not really finished though and there is likely no agreement whether we'd want that, even if finished.

Also, you seem to be reimplementing glob handling which we should already have support for through the [wcmatch dependency](https://facelessuser.github.io/wcmatch/glob/)

tomv564:
Hi @rchl! Saw your work also, but I may have misunderstood that you wanted to solve it through a few LSP-* language packages only?

Thanks for the hint about wcmatch - there is no wish to implement glob matching on the LSP side, but watchman did not understand the `{alternative1, alternativ2}` syntax. That might also make the watchman approach pointless, if servers are many glob features that watchman doesn't understand: https://facebook.github.io/watchman/docs/expr/match.html

rwols:
The glob patterns are described here: https://microsoft.github.io/language-server-protocol/specification.html#documentFilter

If watchman understands a subset we would need some kind of translation layer

rchl:
You are right. I forgot that it's not LSP that has to support it but the watcher itself...

Using chokidar at least guarantees that the LSP spec requirements are supported. For example `vetur` (my main usecase for this feature) uses a `{**/*.js,**/*.ts,**/*.json}` pattern.

It's true that currently my solution only works within the LSP-* packages that use `lsp_utils`. We could bring it into the core but the problem is that we don't really agree on whether we want it since it in theory gonna be a resource hog (although at least from my experience with LSP-vue and LSP-typescript, it isn't).

tomv564:
I agree the chokidar solution makes sense for the (majority?) of users who are already using node-based solutions as it's likely much more complete.

What would it take for your solution to become something like a LSP-watcher-chokidar package that works for all language servers?
Can I support in some way to make that happen?

Could we leave the implementation so that I could contribute an LSP-watcher-watchman package in the future?

rchl:
We should first agree on what everyone wants so that we don't work on something that won't be accepted later.

Do we want to extend LSP with some interface to allow packages like LSP-watcher-* to enable the watching functionality for LSP?

That would require supporting registering capability dynamically and also supporting the watcher configuration in the client configuration since some servers (LSP-vue, LSP-eslint) don't use dynamic capabilities for that and require client to define the watchers.

rwols:
> What would it take for your solution to become something like a LSP-watcher-chokidar package that works for all language servers?

I think that could go in lsp_utils, right?

> Could we leave LSP interface watcher so that I could contribute an LSP-watcher-watchman package in the future?

There are these two callbacks for plugins:

```
    def on_register_capability_async(self, registration_id: str, capability_path: str, options: Dict[str, Any]) -> None:
        """
        Notifies about server dynamically registering a capability using the client/registerCapability request.
        This API is triggered on async thread.

        :param registration_id: The registration identifier
        :param capability_path: The registration capability path
        :param options: The registration options
        """
        pass

    def on_unregister_capability_async(
        self, registration_id: str, capability_path: str, options: Dict[str, Any]
    ) -> None:
        """
        Notifies about server un-registering a capability using the client/unregisterCapability request.
        This API is triggered on async thread.

        :param registration_id: The registration identifier
        :param capability_path: The registration capability path
        :param options: The registration options
        """
        pass
```

rchl:
> > What would it take for your solution to become something like a LSP-watcher-chokidar package that works for all language servers?
> 
> I think that could go in lsp_utils, right?

It could but would we then make LSP depend on `lsp_utils`? Because how else would user enable the watching functionality? It's not possible to install a dependency on its own and installing some other server to enable that functionality would be "weird".

> There are these two callbacks for plugins:

I think we want to support that functionality on the ClientConfig level, even without having a plugin.

tomv564:
For simplicity's sake I suggest a `register_watcher` API that has nothing to do with ClientConfig or the existing plugin setup - the logic should not be server-specific I think?
 
> some servers (LSP-vue, LSP-eslint) don't use dynamic capabilities for that

This is not protocol based, right? Would it suffice to expose a `request_watch` API for these plugins to use?

rchl:
But to use the API one needs to create an LSP-* plugin right? In that case I already have solution that does that. But that means that clients defined within the `LSP.sublime-settings` can't use that functionality.

It might be OK of course but I wanted to have a complete solution that we could just plugin the future ST API into, once its ready.

> This is not protocol based, right? Would it suffice to expose a `request_watch` API for these plugins to use?

The notifications (`didChangeWatchedFiles`) are of course are dispatched through the protocol. The registrations of the watcher is complete outside of the server knowledge.

rchl:
> What would it take for your solution to become something like a LSP-watcher-chokidar package that works for all language servers?

Created something like that at https://github.com/sublimelsp/LSP/pull/1764

rwols:
superseded by #1764

rchl:
Note also released implementation through an additional package https://packagecontrol.io/packages/LSP-file-watcher-chokidar
See its readme for more info.

PR #1677, opened 2021-05-24
Improve `code action` op

husanjun:
If  only one option, display the `code action` method to reduce unnecessary operations

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -292,11 +292,10 @@ class LspCodeActionsCommand(LspTextCommand):
             actions_manager.request_for_region_async(
                 view, covering, session_buffer_diagnostics, self.handle_responses_async, dict_kinds)
 
-    def handle_responses_async(self, responses: CodeActionsByConfigName,
-                               has_commands_by_config: bool = False) -> None:
+    def handle_responses_async(self, responses: CodeActionsByConfigName, has_actions: bool = False) -> None:
         self.commands_by_config = responses
         self.commands = self.combine_commands()
-        if len(self.commands) == 1 and has_commands_by_config:
+        if len(self.commands) == 1 and has_actions:
             self.handle_select(0)
         else:
             self.show_code_actions()
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -274,7 +274,7 @@ class LspCodeActionsCommand(LspTextCommand):
     capability = 'codeActionProvider'
 
     def run(self, edit: sublime.Edit, event: Optional[dict] = None,
-            only_kinds: Optional[List[str]] = None, commands_by_config: CodeActionsByConfigName = None) -> None:
+            only_kinds: Optional[List[str]] = None, commands_by_config: Optional[CodeActionsByConfigName] = None) -> None:
         self.commands = []  # type: List[Tuple[str, str, CodeActionOrCommand]]
         self.commands_by_config = {}  # type: CodeActionsByConfigName
         if commands_by_config:
@@ -292,10 +292,10 @@ class LspCodeActionsCommand(LspTextCommand):
             actions_manager.request_for_region_async(
                 view, covering, session_buffer_diagnostics, self.handle_responses_async, dict_kinds)
 
-    def handle_responses_async(self, responses: CodeActionsByConfigName, has_actions: bool = False) -> None:
+    def handle_responses_async(self, responses: CodeActionsByConfigName, run_first: bool = False) -> None:
         self.commands_by_config = responses
         self.commands = self.combine_commands()
-        if len(self.commands) == 1 and has_actions:
+        if len(self.commands) == 1 and run_first:
             self.handle_select(0)
         else:
             self.show_code_actions()
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -283,7 +283,7 @@ class LspCodeActionsCommand(LspTextCommand):
         self.commands = []  # type: List[Tuple[str, str, CodeActionOrCommand]]
         self.commands_by_config = {}  # type: CodeActionsByConfigName
         if commands_by_config:
-            self.handle_responses_async(commands_by_config, True)
+            self.handle_responses_async(commands_by_config, run_first=True)
         else:
             view = self.view
             region = first_selection_region(view)
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -278,7 +278,7 @@ class LspCodeActionsCommand(LspTextCommand):
         self.commands = []  # type: List[Tuple[str, str, CodeActionOrCommand]]
         self.commands_by_config = {}  # type: CodeActionsByConfigName
         self.has_commands_by_config = False
-        if commands_by_config not None:
+        if commands_by_config:
             self.has_commands_by_config = True
             self.handle_responses_async(commands_by_config)
         else:
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -278,8 +278,7 @@ class LspCodeActionsCommand(LspTextCommand):
         self.commands = []  # type: List[Tuple[str, str, CodeActionOrCommand]]
         self.commands_by_config = {}  # type: CodeActionsByConfigName
         if commands_by_config:
-            self.commands_by_config = commands_by_config
-            self.handle_responses_async()
+            self.handle_responses_async(commands_by_config, True)
         else:
             view = self.view
             region = first_selection_region(view)
@@ -293,11 +292,11 @@ class LspCodeActionsCommand(LspTextCommand):
             actions_manager.request_for_region_async(
                 view, covering, session_buffer_diagnostics, self.handle_responses_async, dict_kinds)
 
-    def handle_responses_async(self, responses: CodeActionsByConfigName = None) -> None:
-        if responses:
-            self.commands_by_config = responses
+    def handle_responses_async(self, responses: CodeActionsByConfigName,
+                               has_commands_by_config: bool = False) -> None:
+        self.commands_by_config = responses
         self.commands = self.combine_commands()
-        if len(self.commands) == 1 and responses is None:
+        if len(self.commands) == 1 and has_commands_by_config:
             self.handle_select(0)
         else:
             self.show_code_actions()
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -273,8 +273,13 @@ class LspCodeActionsCommand(LspTextCommand):
 
     capability = 'codeActionProvider'
 
-    def run(self, edit: sublime.Edit, event: Optional[dict] = None,
-            only_kinds: Optional[List[str]] = None, commands_by_config: Optional[CodeActionsByConfigName] = None) -> None:
+    def run(
+        self,
+        edit: sublime.Edit,
+        event: Optional[dict] = None,
+        only_kinds: Optional[List[str]] = None,
+        commands_by_config: Optional[CodeActionsByConfigName] = None
+    ) -> None:
         self.commands = []  # type: List[Tuple[str, str, CodeActionOrCommand]]
         self.commands_by_config = {}  # type: CodeActionsByConfigName
         if commands_by_config:
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -277,10 +277,9 @@ class LspCodeActionsCommand(LspTextCommand):
             only_kinds: Optional[List[str]] = None, commands_by_config: CodeActionsByConfigName = None) -> None:
         self.commands = []  # type: List[Tuple[str, str, CodeActionOrCommand]]
         self.commands_by_config = {}  # type: CodeActionsByConfigName
-        self.has_commands_by_config = False
         if commands_by_config:
-            self.has_commands_by_config = True
-            self.handle_responses_async(commands_by_config)
+            self.commands_by_config = commands_by_config
+            self.handle_responses_async()
         else:
             view = self.view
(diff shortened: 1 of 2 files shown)


plugin/documents.py
```diff
@@ -483,8 +483,17 @@ def _on_code_actions(self, responses: CodeActionsByConfigName) -> None:
...
             icon = 'Packages/LSP/icons/lightbulb.png'
         else:  # 'annotation'
-            suffix = 's' if action_count > 1 else ''
-            code_actions_link = make_command_link('lsp_code_actions', '{} code action{}'.format(action_count, suffix))
+            if action_count > 1:
+                code_actions_link = make_command_link('lsp_code_actions', '{} code actions'.format(action_count))
+            else:
+                for name, commands in responses.items():
+                    command = commands[0].get('command')
+                code_actions_link = make_command_link('lsp_execute', commands[0].get('title'), {
```
rwols, line 491: This looks incomplete: the response type of a textDocument/codeAction request is

```
(Command | CodeAction)[] | null
```

you seem to be only accounting for type `CodeAction[]`, and then only when it has a `command` key. It could also have an `edit` key.

https://microsoft.github.io/language-server-protocol/specification#textDocument_codeAction

rwols: By the way, to run any kind of code action, use `session.run_code_action`.

husanjun: > By the way, to run any kind of code action, use `session.run_code_action_async`.

Thanks for tips, I have updated in 968ef53

plugin/documents.py
```diff
@@ -483,8 +483,17 @@ def _on_code_actions(self, responses: CodeActionsByConfigName) -> None:
...
+                title = '{} code actions'.format(action_count)
+            else:
+                command = {}
+                for commands in responses.values():
+                    command = commands[0]
+                title = command.get('title')
+
+            code_actions_link = make_command_link('lsp_code_actions', title, {
+                "commands_by_config": responses
+            })
```
rchl, line 496: There are some type issues here. Can try:
```suggestion
                title = ''
                for actions in responses.values():
                    title = actions[0].get('title')
            code_actions_link = make_command_link('lsp_code_actions', title, {"actions_by_config": responses})
```
I'd not call the variable "commands" because those are either `CodeAction`s or `Command`s so calling it "commands" is confusing.

husanjun: > There are some type issues here. Can try:
> 
> I'd not call the variable "commands" because those are either `CodeAction`s or `Command`s so calling it "commands" is confusing.

Thank you for your help, modified in f477b9c

plugin/documents.py
```diff
@@ -483,8 +483,16 @@ def _on_code_actions(self, responses: CodeActionsByConfigName) -> None:
...
             icon = 'Packages/LSP/icons/lightbulb.png'
         else:  # 'annotation'
-            suffix = 's' if action_count > 1 else ''
-            code_actions_link = make_command_link('lsp_code_actions', '{} code action{}'.format(action_count, suffix))
+            if action_count > 1:
+                title = '{} code actions'.format(action_count)
+            else:
+                title = ''
+                for actions in responses.values():
+                    title = actions[0].get('title')
```
rchl, line 491: Mypy doesn't like it. Maybe:

```suggestion
                    title = actions[0].get('title', 'Code action')
```

plugin/documents.py
```diff
@@ -483,8 +483,16 @@ def _on_code_actions(self, responses: CodeActionsByConfigName) -> None:
...
+
```
rchl, line 492: It's not in style of this code to have an empty lines in such places. :)

plugin/code_actions.py
```diff
@@ -273,25 +273,34 @@ class LspCodeActionsCommand(LspTextCommand):
...
-        listener = windows.listener_for_view(view)
-        if not listener:
-            return
-        session_buffer_diagnostics, covering = listener.diagnostics_intersecting_async(region)
-        dict_kinds = {kind: True for kind in only_kinds} if only_kinds else None
-        actions_manager.request_for_region_async(
-            view, covering, session_buffer_diagnostics, self.handle_responses_async, dict_kinds)
+        self.has_commands_by_config = False
+        if commands_by_config:
+            self.has_commands_by_config = True
```
rchl, line 282: I'd prefer to have this flag be an optional argument of `handle_responses_async` instead so that there is no state to keep track of and less code.

husanjun: > I'd prefer to have this flag be an optional argument of `handle_responses_async` instead so that there is no state to keep track of and less code.

Resolve in 6f96859

plugin/code_actions.py
```diff
@@ -277,10 +277,9 @@ def run(self, edit: sublime.Edit, event: Optional[dict] = None,
             only_kinds: Optional[List[str]] = None, commands_by_config: CodeActionsByConfigName = None) -> None:
         self.commands = []  # type: List[Tuple[str, str, CodeActionOrCommand]]
         self.commands_by_config = {}  # type: CodeActionsByConfigName
-        self.has_commands_by_config = False
         if commands_by_config:
-            self.has_commands_by_config = True
-            self.handle_responses_async(commands_by_config)
+            self.commands_by_config = commands_by_config
```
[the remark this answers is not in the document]
husanjun: > That's not what I meant. This is unnecessarily stateful. I meant to pass the `has_commands_by_config` flag to `handle_responses_async`.

Well, it seems more streamlined, To be honest I don't like to pass more parameters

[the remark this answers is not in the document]
rchl: But storing properties on the instance of TextCommand is a bad practice since it's initialized per-view rather than per-execution. So then there needs to be an extra code that resets the properties on run.

plugin/code_actions.py
```diff
@@ -293,11 +292,11 @@ def run(self, edit: sublime.Edit, event: Optional[dict] = None,
             actions_manager.request_for_region_async(
                 view, covering, session_buffer_diagnostics, self.handle_responses_async, dict_kinds)
 
-    def handle_responses_async(self, responses: CodeActionsByConfigName = None) -> None:
-        if responses:
-            self.commands_by_config = responses
+    def handle_responses_async(self, responses: CodeActionsByConfigName,
+                               has_commands_by_config: bool = False) -> None:
```
rchl, line 296: Single line is fine if it fits within 120 column mark. Also renamed to "*_actions_*".
```
    def handle_responses_async(self, responses: CodeActionsByConfigName, has_actions_by_config: bool = False) -> None:
```

husanjun: > actions

Thanks for your advice. Code specification is very important

rchl (approved):
I don't know which server and code I could test it with but the code looks fine to me.

plugin/code_actions.py
```diff
@@ -273,25 +273,32 @@ class LspCodeActionsCommand(LspTextCommand):
 
     capability = 'codeActionProvider'
 
-    def run(self, edit: sublime.Edit, event: Optional[dict] = None, only_kinds: Optional[List[str]] = None) -> None:
+    def run(self, edit: sublime.Edit, event: Optional[dict] = None,
+            only_kinds: Optional[List[str]] = None, commands_by_config: CodeActionsByConfigName = None) -> None:
```
rwols, line 277:
```suggestion
            only_kinds: Optional[List[str]] = None, commands_by_config: Optional[CodeActionsByConfigName] = None) -> None:
```

plugin/code_actions.py
```diff
@@ -273,25 +273,32 @@ class LspCodeActionsCommand(LspTextCommand):
...
+                return
+            listener = windows.listener_for_view(view)
+            if not listener:
+                return
+            session_buffer_diagnostics, covering = listener.diagnostics_intersecting_async(region)
+            dict_kinds = {kind: True for kind in only_kinds} if only_kinds else None
+            actions_manager.request_for_region_async(
+                view, covering, session_buffer_diagnostics, self.handle_responses_async, dict_kinds)
+
+    def handle_responses_async(self, responses: CodeActionsByConfigName, has_actions: bool = False) -> None:
```
rwols, line 295: Can you change the second argument name, `has_actions`, into `run_first`?

plugin/documents.py
```diff
@@ -483,8 +483,15 @@ def _on_code_actions(self, responses: CodeActionsByConfigName) -> None:
...
             icon = 'Packages/LSP/icons/lightbulb.png'
         else:  # 'annotation'
-            suffix = 's' if action_count > 1 else ''
-            code_actions_link = make_command_link('lsp_code_actions', '{} code action{}'.format(action_count, suffix))
+            if action_count > 1:
+                title = '{} code actions'.format(action_count)
+            else:
+                title = ''
+                for actions in responses.values():
+                    title = actions[0].get('title', 'code action')
```
rwols, line 491:
```suggestion
                title = next(iter(responses.values())['title']
```

rwols (changes requested):
I like it :) I only have a few nitpicks left.

plugin/documents.py
husanjun: Thank you for tips, Resolved in  6311861

plugin/documents.py
```diff
@@ -483,8 +483,13 @@ def _on_code_actions(self, responses: CodeActionsByConfigName) -> None:
             scope = 'region.yellowish lightbulb.lsp'
             icon = 'Packages/LSP/icons/lightbulb.png'
         else:  # 'annotation'
-            suffix = 's' if action_count > 1 else ''
-            code_actions_link = make_command_link('lsp_code_actions', '{} code action{}'.format(action_count, suffix))
+            if action_count > 1:
+                title = '{} code actions'.format(action_count)
+            else:
+                title = next(iter(responses.values()))['title']
```
rchl, line 489: This is unfortunately invalid code.

Code checker also says so. I recommend installing LSP-pyright which should notify you about most issues that the github checks does.

rwols: Seems to work fine in the console:

```
>>> foo={"a":"b"}
>>> next(iter(foo.values()))
'b'
```

husanjun: I just installed `LSP-pyright`  

image

Before that, it was `pyls`

rwols: Ah, I see my error now

```
>>> foo={"a":{"b":{"title": "hello there"}}}
>>> next(iter(foo.values()))["title"]
  File "__main__", line 1, in <module>
KeyError: 'title'
```

rwols: I think you can use `next(iter(responses.values())[0]['title']` then. There must be a way to get the title in one line :)

plugin/documents.py
```diff
@@ -483,8 +483,13 @@ def _on_code_actions(self, responses: CodeActionsByConfigName) -> None:
             scope = 'region.yellowish lightbulb.lsp'
             icon = 'Packages/LSP/icons/lightbulb.png'
         else:  # 'annotation'
-            suffix = 's' if action_count > 1 else ''
-            code_actions_link = make_command_link('lsp_code_actions', '{} code action{}'.format(action_count, suffix))
+            if action_count > 1:
+                title = '{} code actions'.format(action_count)
+            else:
+                title = next(iter(responses.values()))[0]['title']
```
rchl, line 489: Noticed a potential issue. With two or more servers running on a view, the dict can look like this:

```
{
   'LSP-eslint': [],
   'LSP-typescript': [{...}]
}
```

In which case this code can crash since the first key has no items.

husanjun: > Noticed a potential issue. With two or more servers running on a view, the dict can look like this:
> 
> ```
> {
>    'LSP-eslint': [],
>    'LSP-typescript': [{...}]
> }
> ```
> 
> In which case this code can crash since the first key has no items.

It only works when there is one `code action`
https://github.com/sublimelsp/LSP/blob/0bb5932bb9de33a68188ae790a871cf592088464/plugin/documents.py#L472-L475

rchl: There can still be just one code action in the case I've mentioned.

rwols: How about `next(itertools.chain.from_iterable(response.values()))["title"]`

husanjun: > How about `next(itertools.chain.from_iterable(response.values()))["title"]`

image

plugin/documents.py
husanjun: > I think you can use `next(iter(responses.values())[0]['title']` then. There must be a way to get the title in one line :)

Nice, it is very good

plugin/documents.py
husanjun: > How about `next(itertools.chain.from_iterable(response.values()))["title"]`

Nice, it is very good

rchl (approved):
just did some minor style changes. LGTM

rwols (approved):

rwols:
Thanks! The code actions are displayed in the hover popup too. Consider improving it there as well :)

rwols:
I'm wondering now if this is too intrusive:

image

jwortmann:
Maybe restrict the link text to 30 characters or so, add … if there are more than that, and use the full description for the new `title` attribute so that it shows as tooltip on mouse hover?

And/or get the proportional `system` font to work without side effects to save more space :wink:

husanjun:
> I'm wondering now if this is too intrusive:
> 
> image

I will make changes in another PR to limit the width, as @jwortmann  said

jfcherng:
I guess then the issue becomes there is no way to see the whole title :cry:

husanjun:
> I guess then the issue becomes there is no way to see the whole title when there is only a single code action 😂

It folds automatically and displays the entire title when the mouse hovers

PR #1678, opened 2021-05-24
Fix: do not ignore return types

rwols:
This fixes an edge case where registering the listener to the window manager could be "too late" because the window/view was already closed on the main thread.

```
  File "C:\Program Files\Sublime Text\Lib\python33\sublime_plugin.py", line 863, in on_load_async
    run_view_callbacks('on_load_async', view_id)
  File "C:\Program Files\Sublime Text\Lib\python33\sublime_plugin.py", line 714, in run_view_callbacks
    callback(*args)
  File "C:\Program Files\Sublime Text\Lib\python33\sublime_plugin.py", line 190, in exception_handler
    return event_handler(*args)
  File "C:\Users\mitranim\AppData\Roaming\Sublime Text\Installed Packages\LSP.sublime-package\plugin/documents.py", line 327, in on_load_async
    self._register_async()
  File "C:\Users\mitranim\AppData\Roaming\Sublime Text\Installed Packages\LSP.sublime-package\plugin/documents.py", line 785, in _register_async
    self.manager.register_listener_async(self)
AttributeError: 'NoneType' object has no attribute 'register_listener_async'
```

I'm assuming the solution is to not ignore type hints.

cc @mitranim

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -45,11 +45,6 @@ class AbstractViewListener(metaclass=ABCMeta):
 
     view = None  # type: sublime.View
 
-    @property
-    @abstractmethod
-    def manager(self) -> "WindowManager":
-        raise NotImplementedError()
-
     @abstractmethod
     def session(self, capability_path: str, point: Optional[int] = None) -> Optional[Session]:
         raise NotImplementedError()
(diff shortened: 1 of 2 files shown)


rchl:
LSP crashes with this. :)

Need to remove this also: https://github.com/sublimelsp/LSP/blob/4b052e7255895c71aad72169494b666427652d15/plugin/core/windows.py#L48-L52

rchl (approved):

PR #1679, opened 2021-05-24
Enable CI again

rwols:
Let's see if this works :) https://github.com/SublimeText/UnitTesting/issues/189


rchl (approved):
Obviously the linux failures should be looked at but the changes look good :)

rwols:
I'd like to enable it only if succeeds, with no flaky tests.

rchl:
Possibly I'll be jumping the gun but I believe the main issue is fixed.

Looks like it was related to `DocumentSyncListener` instance being created with the same View and Buffer ID while the existing one was not __del__'eted yet. Below you can see that `on_close` triggers and then there is `__init__` with the same IDs but without there being `__del__` for the existing one.

```
DocumentSyncListener.on_close view(View(61)), buffer(168)
ok
DocumentSyncListener.__init__ view(View(62)), buffer(171)
SessionView __del__ config(TEST) buffer_id(168)
TextDocumentTestCase.tearDownClass
DocumentSyncListener.__init__ view(View(61)), buffer(168)
```

That looks like our issue because after `on_close` is called, ST has the right to assume that the listener is garbage collected but since we didn't unregister it from WindowsManager, it was being kept alive and messing up some logic.

After that fix the `test_publish_diagnostics` test is consistently failing which I'm not sure why again. Doesn't look like it should be affected and the server communication looks correct. Will need to check that too...

I won't have time to continue today (maybe in the evening) but will eventually clean it up.

rchl:
It seems that the `"random" changes` commit made it more stable. Probably not 100% but potentially good enough.

rchl:
And just when I've said it, mac failed the `test_additional_edits_if_session_has_the_resolve_capability` test...

rchl:
I think it will be pretty stable now. The latest failure was due to missing import.

rwols (commented):
Bravo. You win. (I can't approve my own pull request)

PR #1685, opened 2021-05-26
Expose access to LspTextCommand from SaveTasks

### LINKED ISSUE #1684: Format JSON on save
Is there a way to get this LSP to format json files ons save?

It works for other LSPs but just not JSON,

Here are my configs:

Preferences
```
{
        ...
	"lsp_format_on_save": true
        ...
}
```

LSP
```
	"lsp_code_actions_on_save": {
		"source.organizeImports": true,
		"source.fixAll.eslint": true,
	}
```

All LSP-JSON settings are default

rchl:
I'm doing a little bit of cleanup and enhancement to allow SaveTasks to call `LspTextCommand.best_session` since I want to consolidate some code later and fix #1684.

I can do further changes as separate commits or separate PR. I didn't want to just throw all together as it would be hard to review.

This change should have no functional changes.
It does switch from global `sessions_for_view` function to `LspTextCommand.best_session` but I believe this should have no behavior changes? Or if it does then only for the better? I forgot what's the difference between those exactly but I believe the latter is preferred and works correctly in more cases?

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -254,7 +254,7 @@ class CodeActionOnSaveTask(SaveTask):
         tasks = []  # type: List[Promise]
         for config_name, code_actions in responses.items():
             if code_actions:
-                for session in sessions_for_view(self._task_runner.view, 'codeActionProvider'):
+                for session in self._task_runner.sessions('codeActionProvider'):
                     if session.config.name == config_name:
                         for code_action in code_actions:
                             tasks.append(session.run_code_action_async(code_action, progress=False))
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -244,22 +244,22 @@ class CodeActionOnSaveTask(SaveTask):
 
     def _request_code_actions_async(self) -> None:
         self._purge_changes_async()
-        on_save_actions = self._get_code_actions_on_save(self._view)
-        actions_manager.request_on_save(self._view, self._handle_response_async, on_save_actions)
+        on_save_actions = self._get_code_actions_on_save(self._task_runner.view)
+        actions_manager.request_on_save(self._task_runner.view, self._handle_response_async, on_save_actions)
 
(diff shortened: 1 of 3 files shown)


plugin/code_actions.py
```diff
@@ -244,22 +244,22 @@ def run_async(self) -> None:
...
     def _handle_response_async(self, responses: CodeActionsByConfigName) -> None:
         if self._cancelled:
             return
-        document_version = self._view.change_count()
+        document_version = self._task_runner.view.change_count()
         tasks = []  # type: List[Promise]
         for config_name, code_actions in responses.items():
             if code_actions:
-                for session in sessions_for_view(self._view, 'codeActionProvider'):
+                for session in sessions_for_view(self._task_runner.view, 'codeActionProvider'):
```
rwols, line 257: Can you also replace this one?

plugin/formatting.py
```diff
@@ -24,13 +24,13 @@ class WillSaveWaitTask(SaveTask):
...
-    def __init__(self, view: sublime.View, on_complete: Callable[[], None]) -> None:
-        super().__init__(view, on_complete)
+    def __init__(self, task_runner: LspTextCommand, on_complete: Callable[[], None]) -> None:
+        super().__init__(task_runner, on_complete)
         self._session_iterator = None  # type: Optional[Iterator[Session]]
 
     def run_async(self) -> None:
         super().run_async()
-        self._session_iterator = sessions_for_view(self._view, 'textDocumentSync.willSaveWaitUntil')
+        self._session_iterator = sessions_for_view(self._task_runner.view, 'textDocumentSync.willSaveWaitUntil')
```
rwols, line 33: this can also use `sessions(...)` I believe?

rchl: type-checking doens't complain :)

rchl: Also tested on actual server

rwols (approved):

PR #1686, opened 2021-05-27
Improve hover `code actions` UX

husanjun:
1. If only one, If the action link is clicked, it is ran immediately, circumventing the quick panel.
image

2. If only diagnostics, don't show navigation
image

3. Improved hover phantom problem
 

BTW, When hovering, the `code action` does not just occur during the diagnosis

### DIFF
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -103,11 +103,11 @@ class LspHoverCommand(LspTextCommand):
                 self.request_symbol_hover_async(listener, hover_point)
             self._diagnostics_by_config, covering = listener.diagnostics_touching_point_async(hover_point)
             if self._diagnostics_by_config:
-                if not only_diagnostics:
-                    actions_manager.request_with_diagnostics_async(
-                        self.view, covering, self._diagnostics_by_config,
-                        functools.partial(self.handle_code_actions, listener, hover_point))
                 self.show_hover(listener, hover_point, only_diagnostics)
+            if not only_diagnostics:
+                actions_manager.request_for_region_async(
+                    self.view, covering, self._diagnostics_by_config,
+                    functools.partial(self.handle_code_actions, listener, hover_point))
 
         sublime.set_timeout_async(run_async)
 
@@ -152,14 +152,21 @@ class LspHoverCommand(LspTextCommand):
                     format_diagnostic_for_html(self.view, sb.session.config, diagnostic, self._base_dir))
             for items in by_severity.values():
                 formatted.extend(items)
-            config_name = sb.session.config.name
-            if config_name in self._actions_by_config:
-                action_count = len(self._actions_by_config[config_name])
-                if action_count > 0:
-                    href = "{}:{}".format('code-actions', config_name)
-                    text = "choose code action ({} available)".format(action_count)
-                    formatted.append('<div class="actions">[{}] {}</div>'.format(config_name, make_link(href, text)))
             formatted.append("</div>")
+        return "".join(formatted)
+
+    def code_actions_content(self) -> str:
+        formatted = []
+        for config_name, actions in self._actions_by_config.items():
+            action_count = len(actions)
+            if action_count > 0:
+                href = "{}:{}".format('code-actions', config_name)
+                if action_count > 1:
+                    text = "choose code action ({} available)".format(action_count)
+                else:
+                    text = actions[0].get('title', 'code action')
+                formatted.append('<div class="actions">[{}] Code action: {}</div>'.format(
+                    config_name, make_link(href, text)))
         return "".join(formatted)
 
     def hover_content(self) -> str:
@@ -170,8 +177,9 @@ class LspHoverCommand(LspTextCommand):
         sublime.set_timeout(lambda: self._show_hover(listener, point, only_diagnostics))
 
     def _show_hover(self, listener: AbstractViewListener, point: int, only_diagnostics: bool) -> None:
-        contents = self.diagnostics_content() + self.hover_content()
-        if contents and not only_diagnostics:
+        hover_content = self.hover_content()
+        contents = self.diagnostics_content() + hover_content + self.code_actions_content()
+        if contents and not only_diagnostics and hover_content:
             contents += self.symbol_actions_content(listener, point)
 
         _test_contents.clear()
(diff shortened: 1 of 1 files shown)


plugin/hover.py
```diff
@@ -95,14 +95,13 @@ def run_async() -> None:
...
-            self._diagnostics_by_config, covering = listener.diagnostics_touching_point_async(hover_point)
-            if self._diagnostics_by_config:
-                if not only_diagnostics:
-                    actions_manager.request_with_diagnostics_async(
-                        self.view, covering, self._diagnostics_by_config,
-                        functools.partial(self.handle_code_actions, listener, hover_point))
+                actions_manager.request_for_region_async(
+                    self.view, covering, self._diagnostics_by_config,
+                    functools.partial(self.handle_code_actions, listener, hover_point))
+            else:
```
rchl, line 104: With the previous logic we would show the hover info as soon as possible and update the popup with diagnostics as soon as those arrive. With your change I believe that we will wait with showing the popup until diagnostics arrived?

husanjun: > With the previous logic we would show the hover info as soon as possible and update the popup with diagnostics as soon as those arrive. With your change I believe that we will wait with showing the popup until diagnostics arrived?

Diagnostic response is negligible, it does not require a request server

rchl: `actions_manager.request_for_region_async` does make a request (potentially even multiple) to the server.

husanjun: Sorry, I don t quite understand, it seems to me that this is a request for the `code action`. It s after hovering

rchl: In this highlighted code there are two types of requests made:
 1. one `textDocument/hover` request to get symbol information for hovered point
 2. one or more `textDocument/codeAction` requests to get code actions for hovered thing

And as said in the first comment, previously the first one would be triggered and shown without waiting for the 2 to finish.

plugin/hover.py
```diff
@@ -166,8 +171,8 @@ def show_hover(self, listener: AbstractViewListener, point: int, only_diagnostic
         sublime.set_timeout(lambda: self._show_hover(listener, point, only_diagnostics))
 
     def _show_hover(self, listener: AbstractViewListener, point: int, only_diagnostics: bool) -> None:
-        contents = self.diagnostics_content() + self.hover_content()
-        if contents and not only_diagnostics:
+        contents = self.diagnostics_content() + self.hover_content() + self.code_actions_content()
+        if contents and not only_diagnostics and self.hover_content():
```
rwols, line 175: Why do we call `hover_content()` twice?

plugin/hover.py
```diff
@@ -148,16 +147,22 @@ def diagnostics_content(self) -> str:
...
+        for config_name, actions in self._actions_by_config.items():
+            action_count = len(actions)
+            if action_count > 0:
+                href = "{}:{}".format('code-actions', config_name)
+                if action_count > 1:
+                    text = "choose code action ({} available)".format(action_count)
+                else:
+                    text = actions[0].get('title', 'code action')
+                formatted.append('<div class="actions">[{}] {}</div>'.format(config_name, make_link(href, text)))
+        return "".join(formatted)
```
rwols, line 164: Can this go back to where it was in `diagnostics_content`? (but with the `action > 1` check)

plugin/hover.py
rwols: I too don't understand this part.

plugin/hover.py
```diff
@@ -197,10 +202,14 @@ def _on_navigate(self, href: str, point: int) -> None:
...
-            if window:
-                window.show_quick_panel(titles, lambda i: self.handle_code_action_select(config_name, i),
-                                        placeholder="Code actions")
+            if len(titles) > 1:
+                window = self.view.window()
+                if window:
+                    window.show_quick_panel(titles, lambda i: self.handle_code_action_select(config_name, i),
+                                            placeholder="Code actions")
+            else:
+                self.handle_code_action_select(config_name, 0)
```
rwols, line 211: Looks good :)

rwols (changes requested):

plugin/hover.py
husanjun: The `code action` does not just occur during the diagnosis.Please see my screenshot

plugin/hover.py
husanjun: Resolved in 17d858f

plugin/hover.py
husanjun: Resolved in 17d858f

plugin/documents.py
```diff
@@ -490,8 +491,9 @@ def _on_code_actions(self, responses: CodeActionsByConfigName) -> None:
                 title = '{} code actions'.format(action_count)
             else:
                 title = next(itertools.chain.from_iterable(responses.values()))['title']
+                title = os.linesep.join([title[:30], title[30:]])
```
rchl, line 494: Why are you using `os.linesep`? In ST environment everything is pretty much normalized to `\n` and it should work with just that.

husanjun: Thank you for reminding me. I was not clear about it before, and I will revise it immediately

plugin/hover.py
```diff
@@ -99,11 +99,11 @@ def run_async() -> None:
...
             if self._diagnostics_by_config:
-                if not only_diagnostics:
-                    actions_manager.request_with_diagnostics_async(
-                        self.view, covering, self._diagnostics_by_config,
-                        functools.partial(self.handle_code_actions, listener, hover_point))
                 self.show_hover(listener, hover_point, only_diagnostics)
+            if not only_diagnostics:
+                actions_manager.request_for_region_async(
+                    self.view, covering, self._diagnostics_by_config,
+                    functools.partial(self.handle_code_actions, listener, hover_point))
```
rchl, line 106: Is this code shuffle intentional?

plugin/documents.py
husanjun: Resolved in 0eed685

plugin/hover.py
husanjun: > Is this code shuffle intentional?

Whether the diagnosis exists or not, the `code action` needs to be executed. They have to be independent, Any questions

rchl: Right, it does seem like quite a bit more expensive but correct approach.

rchl:
> 1. If only one, If the action link is clicked, it is ran immediately, circumventing the quick panel.
>    image

That code action title can look like just some random text there, depending on what it says, I feel.
Maybe we should prefix it with `Code action:`?

husanjun:
> > 1. If only one, If the action link is clicked, it is ran immediately, circumventing the quick panel.
> >    image
> 
> That code action title can look like just some random text there, depending on what it says, I feel.
> Maybe we should prefix it with `Code action:`?

Do you mean?
image

rchl:
Yes.
Lets discuss this first before doing any changes but something like that, yes. Ideally with a space after the colon :)

plugin/documents.py
```diff
@@ -490,8 +490,10 @@ def _on_code_actions(self, responses: CodeActionsByConfigName) -> None:
                 title = '{} code actions'.format(action_count)
             else:
                 title = next(itertools.chain.from_iterable(responses.values()))['title']
+                title = "\n".join([title[:30], title[30:]])
             code_actions_link = make_command_link('lsp_code_actions', title, {"commands_by_config": responses})
-            annotations = ["<div class=\"actions\">{}</div>".format(code_actions_link)]
+            annotations = [
+                "<div class=\"actions\" style=\"width:100%;font-family:system\">{}</div>".format(code_actions_link)]
```
rwols, line 496: Can you please make these changes in a separate pull request? This conflates two topics. Also, use `textwrap.wrap`, as the current code might split at non-word boundaries.

rwols (changes requested):

plugin/documents.py
husanjun: Reso

rwols:
Please merge with `main` and resolve the merge conflicts.

PR #1688, opened 2021-05-27
Allow no automatic code actions

### LINKED ISSUE #1687: There should be a way to have Code Actions enabled but not visually offering them

Some language servers have certain Code Actions that are offered very frequently. For example, on virtually every line, [ocaml-lsp](https://github.com/ocaml/ocaml-lsp) will offer a Code Action to add a type annotation. This gets annoying, because although it's technically a valid thing being offered, it's distracting for there _always_ to be a bulb or annotation visible in ST.

For it to be made possible to either:

**(A)** Disable the visual offering of all Code Actions. I guess by adding `false` as a possible value for this setting:

```jsonc
// Show code actions:
// "annotation" - show an annotation on the right when code actions are available.
// "bulb" - show a bulb in the gutter when code actions are available.
// Note: Due to API limitations, the "bulb" icon can not be clicked so the code actions can only be triggered
// using a keyboard shortcut or the context menu.
```

...then Code Actions would not be offered visually, but could still be summoned at the caret using a keybinding for the command `lsp_code_actions`.

or **(B)** Allowing the `disabled_capabilities` per-client setting to be used to mute certain Code Actions that the user doesn't want to be visually offered. Something like:

```jsonc
"disabled_capabilities": {
    "codeActionProvider": {
        "mute": ["type-annotate"],
    }
},
```

rwols:
Fixes #1687.

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -93,6 +93,7 @@
   // Show code actions:
   // "annotation" - show an annotation on the right when code actions are available.
   // "bulb" - show a bulb in the gutter when code actions are available.
+  // "" - don't show code actions.
   // Note: Due to API limitations, the "bulb" icon can not be clicked so the code actions can only be triggered
   // using a keyboard shortcut or the context menu.
   "show_code_actions": "annotation",
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -273,7 +273,8 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
 
     def on_diagnostics_updated_async(self) -> None:
         self._clear_code_actions_annotation()
-        self._do_code_actions()
+        if userprefs().show_code_actions:
+            self._do_code_actions()
         self._update_diagnostic_in_status_bar_async()
 
     def _update_diagnostic_in_status_bar_async(self) -> None:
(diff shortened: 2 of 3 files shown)


LSP.sublime-settings
```diff
@@ -93,6 +93,7 @@
   // Show code actions:
   // "annotation" - show an annotation on the right when code actions are available.
   // "bulb" - show a bulb in the gutter when code actions are available.
+  // "" - don't show code actions.
```
rchl, line 96: Would prefer `false` to be honest but I guess this simplifies types?

rchl (approved):

LSP.sublime-settings
rwols: yes, what would `true` mean then?

rchl: There doesn't need to be "true".
Or it can default to whatever is the default.

jwortmann: Maybe `"none"` as it is used in various ST preferences.

rwols: We already use `""` for diagnostics_highlight_style, document_highlight_style, diagnostics_gutter_marker.

jwortmann: Right, and "none" would probably only fit here if the setting was used to show only a subset of code actions or so. I guess empty string is fine.

PR #1689, opened 2021-05-27
Allow adjustable extents of the popups

### LINKED ISSUE #1635: Max width/length of popup window
The popup window that shows docstring is a bit too long for my own-taste. It would be nice to have a setting so it would do word-wrap after X chars.

### LINKED ISSUE #1657: Limit the height of the hover popup
When there is more hovering content, it is difficult for us to see important information, whether we can add a scroll bar to the introduction, and limit the height of the pop-up box

rwols:
Fixes #1635.
Fixes #1657.

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -366,10 +366,6 @@ def text_document_code_action_params(
 LSP_POPUP_SPACER_HTML = '<div class="lsp_popup--spacer"></div>'
 
 
-def _chars_to_pixels(view: sublime.View, characters: int) -> int:
-    return int(view.em_width() * float(characters))
-
-
 def show_lsp_popup(view: sublime.View, contents: str, location: int = -1, md: bool = False, flags: int = 0,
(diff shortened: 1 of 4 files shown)


plugin/core/views.py
```diff
@@ -379,8 +384,8 @@ def show_lsp_popup(view: sublime.View, contents: str, location: int = -1, md: bo
         flags=flags,
         location=location,
         wrapper_class=wrapper_class,
-        max_width=int(view.em_width() * 120.0),  # Around 120 characters per line
-        max_height=1000000,
+        max_width=_chars_to_pixels(view, userprefs().popup_max_characters_width),
+        max_height=_chars_to_pixels(view, userprefs().popup_max_characters_height),
```
rchl, line 388: It's an approximation so not a big deal but `view.line_height()` would be more appropriate for height.

rchl (dismissed):

PR #1701, opened 2021-05-29
Don't ascii-escape Unicode

rwols:
This makes the flow language server not crash with non-ascii characters.
However, inserting an emoji, and then removing that emoji brings flow into an
invalid state. It seems to be because when the emoji is removed, that counts
as two UTF-16 characters.

### DIFF
--- a/tests/test_protocol.py
+++ b/tests/test_protocol.py
@@ -2,6 +2,8 @@ from LSP.plugin.core.protocol import (
     Point, Range, Request, Notification
 )
 import unittest
+from LSP.plugin.core.transports import _encode, _decode
+
 
 LSP_START_POSITION = {'line': 10, 'character': 4}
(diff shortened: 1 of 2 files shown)


tests/test_protocol.py
```diff
@@ -2,6 +2,8 @@
     Point, Range, Request, Notification
 )
 import unittest
+from LSP.plugin.core.transports import _encode, _decode
```
rchl, line 5: that import section could use some cleanup :)

rchl (approved):

tests/test_protocol.py
rwols: I'm playing around with enabling `black` and `isort` for this stuff

BigGuy573 (commented):
Enhancement Updates👍

PR #1702, opened 2021-05-29
Add option to show diagnostics as annotations

rchl:
Add `show_diagnostics` option and support for `annotation` value to show diagnostics as annotations.

`show_diagnostics` is an array so user can decide whether to show just regions (highlights) or both.

Screenshot 2023-06-10 at 23 16 12


rchl:
- It can be noisy
- Will have priority over code action annotations so code actions will be hidden behind "hover" action. Or possibly it will even be random which ones are prioritized.
- Ideally the most severe diagnostics would be prioritized (showed without hovering) but it's the opposite way right now since annotations and regions are added by the same `add_regions` call, and regions needs to be added in the opposite order to annotations

Daggy1234:
Hey super excited for this feature. Just a few caveats:

1) It would be nice to make the colors less dark, perhaps more transparency
2) Some more padding on the surrounding boxes (just for readability)
3) Perhaps like errorlens color the text as well? (white doesn't go well on all colors)

These would be my choices however, feel free to interject!

rchl:
> * It would be nice to make the colors less dark, perhaps more transparency

I don't think it's possible to apply transparency to the background of those annotations.

> 2\. Some more padding on the surrounding boxes (just for readability)

Where exactly? Top and bottom padding is probably not a good idea because then it could get taller than line height.

> 3\. Perhaps like errorlens color the text as well? (white doesn't go well on all colors) . By which I mean a much lighter background accent color on which darker colored text that matches goes.

Can experiment with that.

rchl:
This is with the text color defined and no background. Compare to the first screenshot.
Screenshot 2021-05-29 at 23 56 26

Zomatree:
When hovering over an annotation it seems all other annotation's text color changed:
image

rchl:
> When hovering over an annotation it seems all other annotation's text color changed:

Yes. Looks like annotations don't play that well with custom styling.
It works better when just overriding text color. Just pushed that change.

Rapptz:
Thanks for taking the time to implement this! At first glance, I like it. Though I imagine this would probably require new settings for people who don't prefer this view. Error lens lets you configure the range of severity that you care to show. Obviously there aren't any settings since this is just a draft PR. 

Personally I couldn't get it to show all diagnostics, just the error ones. Not sure why that is.

Daggy1234:
> 
> 
> This is with the text color defined and no background. Compare with the first screenshot.
> 
> Screenshot 2021-05-29 at 23 56 26

Looks fantastic!

eproxus:
How would this look if you actually had a more "normal" line width where errors overlap the code? (I realize some people might have very wide editor panes, but I certainly don't).

rchl:
The errors would never overlap the code. Those are always shortened to fit the available space.

rchl:
Still experimenting.

Brought back the code from the original @rwols' branch that showed annotation only for the diagnostic under the cursor.

I think the behavior of this should be customisable. So at least have those options:
 - show annotations for all active diagnostics
 - show annotation only for the diagnostic under the cursor
 - ...?

The fact that we can't control what has priority in the annotation if there are multiple in the same region is a bit of a problem because what can happen is that on moving the cursor to a diagnostic, the diagnostic annotation will show up for a moment only to be replaced by code actions a moment later.

I feel like to solve that properly we would need either some support for prioritizing in the ST API or have a proper region/diagnostic manager class that would control the order and "drawing" of all those things from one central place. That would be major undertaking but I think that would also be needed if we would want to have some proper "problems panel".

rchl:
> I feel like to solve that properly we would need either some support for prioritizing in the ST API or have a proper region/diagnostic manager class that would control the order and "drawing" of all those things from one central place. That would be major undertaking but I think that would also be needed if we would want to have some proper "problems panel".

One more advantage - should allow implementing proper support for navigating to next/previous diagnostic instead of relying on built-in `next/previous_result` commands which don't always work as expected.

TheSecEng:
Have we played with showing dianostic as phantoms below the issue?

rchl:
> Have we played with showing diagnostics as phantoms below the issue?

I would consider this to be probably the worse option since it would make the content jump up and down constantly and create gaps in the code that would make it less readable.

TheSecEng:
> > Have we played with showing diagnostics as phantoms below the issue?
> 
> I would consider this to be probably the worst option since it would make the content jump up and down constantly and create gaps in the code that would make it less readable.

I might work on an implementation for POC. I personal would prefer it based on my hacked together implementation. I don't image code jumping a ton unless you have a ton of diagnostics.

TheSecEng:
Just for reference

image

rchl:
I've decided to abandon the option for showing all diagnostics inline and only support displaying inline the diagnostics for under the cursor. The other option could be added in the future if needed.

As far as the conflict between code action annotations and diagnostic annotations, I can't solve it currently so just warned about the incompatibility in the setting description.

plugin/session_buffer.py
```diff
@@ -49,9 +49,11 @@ def update(self, version: int, changes: Iterable[sublime.TextChange]) -> None:
 
 class DiagnosticSeverityData:
 
-    __slots__ = ('regions', 'regions_with_tag', 'annotations', 'scope', 'icon')
+    __slots__ = ('region_diagnostics', 'tag_diagnostics', 'regions', 'regions_with_tag', 'annotations', 'scope', 'icon')
 
     def __init__(self, severity: int) -> None:
+        self.region_diagnostics = []  # type: List[Diagnostic]
+        self.tag_diagnostics = []  # type: List[Diagnostic]
```
rchl, line 56: This change is a bit of a hack I guess since it stores whole diagnostics in a class whose name suggest that it should be only "diagnostic severity data"...

plugin/core/types.py
```diff
@@ -210,7 +210,7 @@ class Settings:
     show_diagnostics_count_in_view_status = None  # type: bool
     show_multiline_diagnostics_highlights = None  # type: bool
     show_diagnostics_in_view_status = None  # type: bool
-    show_diagnostics_inline = None  # type: Literal["all", "at-cursor", "off"]
+    show_diagnostics_inline = None  # type: Literal["all", "at-cursor", "none"]
```
predragnikolic, line 213: Remove "all" in the type

plugin/documents.py
```diff
@@ -308,7 +308,7 @@ def _update_diagnostic_in_status_bar_async(self) -> None:
         self.view.erase_status(self.ACTIVE_DIAGNOSTIC)
 
     def _update_inline_diagnostic_async(self) -> None:
-        region_key = "lsp_d-a"
+        region_key = "lsp_d-ann"
```
predragnikolic, line 311: Maybe be a more explicit about the region key... 
a -> ann -> annotations

I know that we have places in the code where we use one letter to describe the region key, but it takes time to understand what that means.
So i would suggest using a whole word over abbreviation.

rchl: I guess my thinking was that I might save a lot of bytes in memory but since I'm only showing active diagnostic, it's never gonna be an issue.

predragnikolic:
I used this for 2 days.
Here are my conclusions so far.

This works as expected,
when setting `"show_diagnostics_inline": "at-cursor"`,
the diagnostics is shown in the annotation:
output

In my daily use,
I tend to use next_result, prev_result commands:
```
    { "keys": ["alt+3"], "command": "next_result" },
    { "keys": ["alt+1"], "command": "prev_result" },
```

And when I do that in combination with `"show_diagnostics_inline": "at-cursor"`,
I get the following result:
output1


Note: the popup shows up each time I navigate to the next previous result,
so we end up showing diagnostic information in both the popup and annotation, 
which looks redundant to me.

Plus, I have the`"show_diagnostics_in_view_status": false`,
if I had that setting turned on I would see the same diagnostics on 3 places. 

----

But at the end I think that some people would find this useful.
ST also displays build errors in annotations.
And the code is nice.

So this looks good to me.

predragnikolic (approved):

rchl:
I could work to try to add support for showing all diagnostics in the file (as a phantom below maybe).

rchl:
Refactored by moving logic from `DocumentSyncListener` to `SessionView`. Mainly to be able to remove regions when SessionView is destroyed.

LSP.sublime-settings
```diff
@@ -47,6 +47,12 @@
   // under the cursor in status bar if available.
   "show_diagnostics_in_view_status": true,
 
+  // Show the diagnostics as inline annotations.
+  // When using the "at-cursor" value it's recommended to not use the "annotation"
+  // value for the "show_code_actions" option as then the code action annotations will
+  // show over the diagnostic annotations.
+  "show_diagnostics_inline": "none",
```
rchl, line 54: Not happy about introducing this option.

I'm thinking that maybe there should be an option like `diagnostics_style` that takes an array and the user can specify multiple values like `inline-after`, `inline-below` or similar.

Maybe there wouldn't be a need for an array if we keep the normal diagnostic highlights (`diagnostics_highlight_style`) out of it.

Though should also think of how to incorporate a setting for showing either all diagnostics at once or only ones intersecting the cursor... Maybe mash it all together into `diagnostics_style` but coming up with nice value names would be a challenge. Something like `inline-after-at-cursor`, `inline-after-all`, `inline-below-at-cursor`, `inline-below-all` sounds absolutely horrendous.

rchl:
> Note: the popup shows up each time I navigate to the next previous result,
> so we end up showing diagnostic information in both the popup and annotation,
> which looks redundant to me.

I find it that I quite often have to show the popup anyway as there is not enough space in the annotation to tell me what's wrong. I guess ideally the annotation would expand automatically instead but that's not possible.

FichteFoll:
Since SublimeLinter more or less recently got support for annotations, I've been using these to greate success and it looks as follows (second is when hovered):




I solved the noisiness by only showing the diagnostics' *code*, expanding to the full message text when hovering (including multiple diagnostics if any), which I believe to be a very elegant solution. The configuration needed for this can be found in [my dotfiles repo](https://github.com/FichteFoll/dotfiles/blob/1ebc2932ed5c590c52a85995e0dc70f309367c31/sublime-text/.config/sublime-text/Packages/User/SublimeLinter.sublime-settings#L5-L21).

As for diagnostics vs code lens, ideally LSP would be able to define a priority (either directly with the ST API or via a workaround where the last set of annotations gets appended to the list of annotations that ST renders) that a user can configure.

rchl:
> As for diagnostics vs code lens, ideally LSP would be able to define a priority (either directly with the ST API or via a workaround where the last set of annotations gets appended to the list of annotations that ST renders) that a user can configure.

ST API would be obviously preferred but not sure when or if we'd get that.

LSP can control the order but it would have to remove existing annotations and re-add them in case code action annotation comes when the diagnostic annotation is shown. That would likely create a bit of a flicker.

rchl:
I've integrated @TerminalFi port of https://github.com/Maan2003/lsp_lines.nvim as a proof of concept.

I like the instant visibility of diagnostics without having to open the hover popup or hover diagnostics.

I don't like quite a few about it though...
 - When typing, diagnostics can get created anywhere above the cursor. This shifts the view down and is both distracting, can hide the current line under the completions popup or can even hide the completions popup if the current line shift down far enough.
 - Documents with many multi-line diagnostics can be distracting to look at and navigate within due to how phantoms work.

Screenshot 2023-03-04 at 11 58 08

Screenshot 2023-03-04 at 11 59 55

rchl:
Here is a pretty bad case of the issue I'm talking about:

https://user-images.githubusercontent.com/153197/222897056-a8ba24e5-1626-4c84-ae68-9b0c98fea7ab.mov

TerminalFi:
Glad to see someone found value. Hopefully we can get native Sublime support. 

Text / font settings is one thing that needs test unless we can embed a font type

TerminalFi:
> Here is a pretty bad case of the issue I'm talking about:
> 
> 
> 
> https://user-images.githubusercontent.com/153197/222897056-a8ba24e5-1626-4c84-ae68-9b0c98fea7ab.mov
> 
> 
> 
> 

Yikes, almost need a longer debounce. I wonder how it works in VSCode and if they get the same experience. Also might be worth having it toggled and displayed on key press.

rchl:
The experience as is is pretty horrendous and distracting so I imagine it won't go in like that.

I don't think debounce will help. It will still do the same, just a little less often.

I'm not sure which VSCode extension does something like that.

rchl:
> Text / font settings is one thing that needs test unless we can embed a font type

I've refactored quite a lot compared to your version so maybe it doesn't have the same problems anymore. Do you have a setup that has issues with your version?

TerminalFi:
> > Text / font settings is one thing that needs test unless we can embed a font type
> 
> 
> 
> I've refactored quite a lot compared to your version so maybe it doesn't have the same problems anymore. Do you have a setup that has issues with your version?

Let me do some testing on my end. Maybe it's a none issue now :)

TerminalFi:
> > Text / font settings is one thing that needs test unless we can embed a font type
> 
> I've refactored quite a lot compared to your version so maybe it doesn't have the same problems anymore. Do you have a setup that has issues with your version?

Looks like your refactor fixed the font issues. Great work!

FichteFoll:
I wonder if the jumping could be mitigated a bit by adjusting the viewport by how much the result of `View.text_to_layout` changes for the current caret or whether that would be too jumpy/flicker. I just confirmed that this API does consider vertical space consumed by phantoms.

```py
>>> view.text_to_layout(view.sel()[0].a)
(0.0, 540.0)
>>> p = [sublime.Phantom(view.sel()[0], "test phantom", sublime.PhantomLayout.BELOW)]
>>> ps = sublime.PhantomSet(view, 'test')
>>> ps.update(p)
>>> view.text_to_layout(view.sel()[0].a)
(0.0, 564.0)
```

rchl:
I have pushed a hack for that but it results in visible flicker.

Probably makes sense to suggest an improvement in ST tracker.

Daggy1234:
I've been silently following this feature for a while now! I am not a fan of the vertical line structure, very distracting. Is it possible to go back to simple side annotations like SublimeLinter?

rchl:
Those are two alternative ways of showing diagnostics. One doesn't replace the other. At this stage of this draft PR there is no settings to toggle stuff as it's just a POC.

plugin/diagnostics.py
```diff
@@ -117,14 +117,17 @@ def __init__(self, view: sublime.View, highlight_line_background: bool = False)
...
+    def update_async(self, diagnostics: List[Tuple[Diagnostic, sublime.Region]]) -> None:
         sorted_diagnostics = self._sort_diagnostics(self._preprocess_diagnostic(diagnostics))
         line_stacks = self._generate_line_stacks(sorted_diagnostics)
         blocks = self._generate_diagnostic_blocks(line_stacks)
-        phantoms = []  # Type: List[sublime.Phantom]
+        phantoms = []  # type: List[sublime.Phantom]
         for block in blocks:
             content = self._generate_region_html(block)
             phantoms.append(sublime.Phantom(block['region'], content, sublime.LAYOUT_BELOW))
+        sublime.set_timeout(lambda: self._update_sync(phantoms))
```
predragnikolic, line 128: Yep, this looks like it should fix the flicker.

predragnikolic: The sublime.set_timeout for me is like the requestAnimationFrame in the web world 🙂

rchl:
I think I'm ready with this feature (diagnostic annotations).

It will conflict with code action annotations but documentation states that those should not be used together.

Otherwise its perhaps a bit flickery on editing the view but I think that's acceptable. And it's not enabled by default.

predragnikolic:
Q1: When navigating with the `lsp_next_diagnostics` and when show_diagnostics is set to annotation,
Would it be better to not show the hover popup that displays only the diagnostics in that case? (to avoid showing duplicate information in the hover popup and annotation)

https://github.com/sublimelsp/LSP/blob/84a4c2eca930c399adfc3a23e02d11dbffae8260/plugin/core/registry.py#L311

---

My initial answer to that question would be to hide the popup,
but than I changed my mind.

Annotation are not always visible, if the lines are wide...
See the video below:
[Screencast from 2023-06-11 14-27-41.webm](https://github.com/sublimelsp/LSP/assets/22029477/a12aba5d-a350-4c6a-9d47-8d493b75e165)

As you can see, the annotation is not visible,
and showing the popup helps see what the problem is immediately. 

I could use the mouse, 
and hover over the squiggly underline in the view (or the small annotation that is peaking from the right side)
to trigger the annotation to show the text. (but that behavior could sometime have a buggy behavior to show a annotation on the wrong line ... see the video below)
[Screencast from 2023-06-11 14-35-28.webm](https://github.com/sublimelsp/LSP/assets/22029477/06e43129-1763-4ad8-bdfc-b23e739dcaf7)

predragnikolic:
All in all, i haven't noticed any real issues, while testing this today.

I will test this PR in practice this week, 
and let you know how it goes. :)

LSP.sublime-settings
```diff
@@ -43,6 +43,14 @@
 
   // --- Diagnostics --------------------------------------------------------------------
 
+  // How to display diagnostics.
+  // Allowed options:
+  // - "highlight" - Show all diagnostic regions using highlights. Also see `diagnostics_highlight_style`.
+  // - "annotation" - Show diagnostics as annotations anchored at the right side of the corresponding line.
+  //                  When using this value it's recommended to not use the "annotation" value for the `show_code_actions`
+  //                  option as it's not possible to enforce which one is shown first then.
+  "show_diagnostics": ["highlight"],
```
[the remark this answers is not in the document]
rchl: Good point. I forgot to implement that, it looks like.

plugin/diagnostics.py
```diff
@@ -0,0 +1,49 @@
+from .core.protocol import Diagnostic
+from .core.protocol import DiagnosticSeverity
+from .core.typing import Dict, List, Tuple
+from .core.views import DIAGNOSTIC_KINDS
+from .core.views import diagnostic_severity
+from .core.views import format_diagnostics_for_annotation
+import sublime
+
+
+class DiagnosticsView():
```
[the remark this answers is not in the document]
rchl: I think my intention was to also move handling of diagnostic regions here. But not sure how much work would that be.

rchl:
> My initial answer to that question would be to hide the popup,
> but than I changed my mind.

Looks like you've answered yourself. :)
I agree that the popup should still show as annotation can only show limited number of characters so it might or might not show the whole info.

LSP.sublime-settings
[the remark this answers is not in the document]
rchl: Though I don't imagine someone wanting to disable highlights as then it wouldn't not be possible to tell where the issue is on the line.

Or maybe it should be possible to set `"show_diagnostics": []` but setting `annotation` should still force highlights to show...

plugin/documents.py
```diff
@@ -285,6 +287,12 @@ def on_diagnostics_updated_async(self, is_view_visible: bool) -> None:
         is_active_view = window and window.active_view() == self.view
         if is_active_view and self.view.change_count() == self._change_count_on_last_save:
             self._toggle_diagnostics_panel_if_needed_async()
+        self._diagnostics_view.clear_annotations()
+        if 'annotation' in userprefs().show_diagnostics:
+            all_diagnostics = []  # type: List[Tuple[Diagnostic, sublime.Region]]
+            for _, diagnostics in self._diagnostics_async(allow_stale=True):
+                all_diagnostics.extend(diagnostics)
+            self._diagnostics_view.update_diagnostic_annotations_async(all_diagnostics)
```
rchl, line 295: Just realized that this should be tied to `SessionView` and and `DocumentSyncListener`. Otherwise it won't work in split views.

rchl: No, never mind, the current code is correct. It's just there is issue with ST that the LSP doesn't immediately initialize on splitting the view. Something needs to change in the file for it to trigger.
Sounds like one of the issues we had but not sure if it's a known issue, it regressed or it's a new one.

LSP.sublime-settings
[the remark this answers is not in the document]
jwortmann: I think it should be a boolean setting `"show_diagnostics_annotations": false`.

I'd imagine to use an array for the value is confusing to users, because it's not obvious from the description that you can have both at the same time (`["highlight", "annotation"]`). Also I see no real point having the "highlight" option here as well, because we already have the `"diagnostics_highlight_style": ""` option for to disable the underlines in the view. Now this setting would introduce the same again under a different setting name, and it's unclear what should happen if the settings use conflicting values.

jwortmann:
I think it was mentioned before, but wouldn't it be better to only show the annotation for the diagnostic under the cursor? Or maybe for all diagnostics on the line of the cursor (i.e. even if the cursor is not directly on a diagnostic, but somewhere on the same line).

So this would be an alternative to showing the active diagnostic in the status bar.

Ideally it should be configurable, so perhaps a setting with string value would be better, like
```jsonc
  // Allowed options:
  // - "all"
  // - "active_line"
  // - "under_cursor"  (maybe)
  // - ""
  "show_diagnostics_annotations": "",
```

LSP.sublime-settings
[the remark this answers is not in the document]
rchl: I guess I wanted to do a little "reset" and create some more consistency in the settings.

Your `show_diagnostics_annotations` sounds OK but we don't have `show_diagnostics_highlights` so it would be a little inconsistent.

But I'll go with your suggestion because I don't mind it and I don't want this PR to take extra year to finish :)

plugin/documents.py
```diff
@@ -285,6 +287,12 @@ def on_diagnostics_updated_async(self, is_view_visible: bool) -> None:
         is_active_view = window and window.active_view() == self.view
         if is_active_view and self.view.change_count() == self._change_count_on_last_save:
             self._toggle_diagnostics_panel_if_needed_async()
+        self._diagnostics_view.clear_annotations()
+        if userprefs().show_diagnostics_annotations:
+            all_diagnostics = []  # type: List[Tuple[Diagnostic, sublime.Region]]
+            for _, diagnostics in self._diagnostics_async(allow_stale=True):
+                all_diagnostics.extend(diagnostics)
+            self._diagnostics_view.update_diagnostic_annotations_async(all_diagnostics)
```
rchl, line 295: I guess it might be more efficient and potentially less flickery if we would do that from SessionView, per-server. Like we do highlight regions currently.

predragnikolic (approved):

predrag-codetribe:
Usability feedback,
after using it for three days.

**The good.**
It is ok if there are not a lot of errors in the view.

**The awkward.**

If each line has errors, 
the annotations can be really distracting:
Screenshot 2023-06-15 at 15 26 00

Yesterday, I opened a file... I could not focus on the code, because the red annotations on each line, spanning long red text from the right of the screen... 
My whole screen was filled with text, (code coming from left[which is ok], and annotations coming from the right,  basically there was no blank space in the view :D )

**My subjective conclusion.**
I would not use this feature, 
although I find it cool.

Keep in mind, this feedback comes from a Person who likes to see less stuff on the screen. (example. I like to hide the line numbers) 

----

I approve it.
Although If the LSP enabled a way for plugins to hook into it,
I would delegate this feature to the a plugin, than inside LSP. (but that is a different topic, and requires a lot of thinking)

rchl:
I suppose that proper code in the projects we are working on should not have files that have errors on each line. If they do, we would probably fix them. So it would be more of an issue for some random third party or minimized files that happen to show diagnostics? In that case normal diagnostic regions would also be distracting (maybe not as much) and maybe the solution to that is just being able to disable diagnostics temporarily or per file?

rchl:
Switched to `show_diagnostics_annotations_severity_level` and made it so that regions are created per-session which mainly makes a difference when running multiple servers in a single file.

predragnikolic (approved):

FichteFoll:
Just a quick question here, now that it is going to be shipped in the next update, is it will it also be possible to configure the annotations in a way that they only show an error code for example? I configured that for SublimeLinter and it makes the annotations much less noisy. Example below:



when hovered:


Of course the full message also still shows in a popup when hovering over the annotated region.

rchl:
I suppose you could crate a feature request for it but my initial thought is that this wouldn't work well as a global LSP setting as with many servers the error code won't be present or won't mean much to most. So this would probably need to be a per-server configuration.

FichteFoll:
> my initial thought is that this wouldn't work well as a global LSP setting as with many servers the error code won't be present
Yeah, this was also my suspicion. Thanks for clarifying. I'll think of a proper feature request once I've had the opportunity to test it.

PR #1705, opened 2021-05-30
shellcheck works only for bash

rwols:
This avoids starting it for fish https://github.com/Phidica/sublime-fish/blob/6697fdd7becc3e5991484d4365b036b9fbb1d99d/fish.sublime-syntax#L8


rchl (approved):
It also doesn't support zsh but that one doesn't have its own scope.

I know that there will be false positives if we let it run in zsh or fish but I guess a lot of issues can still be valid? Just a thought, I don't really work with those languages apart from some basic scrips.

PR #1706, opened 2021-05-30
Apply all status messages when a SessionView is opened

jwortmann:
Apply all permanent status messages that were set via `Session.set_window_status_async` when a file is opened.

I'm not sure why, but this also fixes an issue I saw, that `set_window_status_async` has no effect when called within `AbstractPlugin.__init__`.

Note that the `set_window_status_async` method ignores the "show_view_status" setting, but it seems much more useful to me this way, because plugin authors could add their own setting for this (or query the "show_view_status" setting from the LSP preferences) and then check that before calling `set_window_status_async`.

### DIFF
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -49,8 +49,6 @@ class SessionView:
         self.session_buffer = session_buffer
         session.register_session_view_async(self)
         session.config.set_view_status(self.view, "")
-        for status_key, message in session._status_messages.items():
-            self.view.set_status(status_key, message)
         if self.session.has_capability(self.HOVER_PROVIDER_KEY):
             self._increment_hover_count()
         self._clear_auto_complete_triggers(settings)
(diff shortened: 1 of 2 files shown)


rchl (approved):

plugin/session_view.py
```diff
@@ -49,6 +49,8 @@ def __init__(self, listener: AbstractViewListener, session: Session) -> None:
         self.session_buffer = session_buffer
         session.register_session_view_async(self)
         session.config.set_view_status(self.view, "")
+        for status_key, message in session._status_messages.items():
+            self.view.set_status(status_key, message)
```
rwols, line 53: Wouldn't it work better if this was done in `Session.register_session_view_async`?

rwols: I think this is a bit asymmetric, you're also accessing "private" variables in `session` here.

rwols (approved):

PR #1707, opened 2021-05-31
Improve code action annotation UX

husanjun:


plugin/documents.py
```diff
@@ -489,9 +490,13 @@ def _on_code_actions(self, responses: CodeActionsByConfigName) -> None:
...
             icon = 'Packages/LSP/icons/lightbulb.png'
         else:  # 'annotation'
-            suffix = 's' if action_count > 1 else ''
-            code_actions_link = make_command_link('lsp_code_actions', '{} code action{}'.format(action_count, suffix))
-            annotations = ["<div class=\"actions\">{}</div>".format(code_actions_link)]
+            if action_count > 1:
+                title = '{} code actions'.format(action_count)
+            else:
+                title = next(iter(responses.values()))[0]['title']
+                title = "<br/>".join(textwrap.wrap(title, width=30))
```
rchl, line 497: `<br>` was never a self-closing element in the spec. `<br>` is enough. :)

husanjun: Yes,  but in my test, it works and `\n` doesn't work

rwols (approved):
I like it :)

image

plugin/documents.py
```diff
@@ -489,9 +490,13 @@ def _on_code_actions(self, responses: CodeActionsByConfigName) -> None:
             scope = 'region.yellowish lightbulb.lsp'
             icon = 'Packages/LSP/icons/lightbulb.png'
         else:  # 'annotation'
-            suffix = 's' if action_count > 1 else ''
-            code_actions_link = make_command_link('lsp_code_actions', '{} code action{}'.format(action_count, suffix))
-            annotations = ["<div class=\"actions\">{}</div>".format(code_actions_link)]
+            if action_count > 1:
+                title = '{} code actions'.format(action_count)
+            else:
+                title = next(iter(responses.values()))[0]['title']
```
rchl, line 496: Something weird happened here. This is some old code that is wrong. This was supposed to be:
```py
title = next(itertools.chain.from_iterable(responses.values()))['title']
```

PR #1710, opened 2021-05-31
Refine diagnostics presentation

### LINKED ISSUE #1547: Hide inline diagnostics if they span multiple lines
LemMinX tends to mark huge portions of a document (see https://github.com/eclipse/lemminx/issues/963). It marks everything from the opening to the expected position of a closing tag in case it is missing.

An author therefore sees only red when writing XML which is not valid most of the time.

I'd like to only see underlined issues which don't exceed one single line. Larger diagnostics regions should be represented by a gutter icon and the panel diagnostics only in order to reduce the noise within the editor control.

A possibility to quickly enable/disable dynamic linting while writing. It's useful in general though.

**Additional context**

These kinds of markups don't help much.

grafik

rwols:
If the diagnostic spans one line:

- Draw errors and warnings as squigglies
- Draw infos and hints as stippled

If the diagnostic spans more than one line: always draw as a box.

The diagnostics_highlight_style setting is removed, as these provide good
defaults.

Resolves #1547

Infos and hints as stippled:
Schermafbeelding 2021-05-31 om 23 08 23

Warnings and errors as squigglies:
Schermafbeelding 2021-05-31 om 23 09 02

If more than one line: always a box
Schermafbeelding 2021-05-31 om 23 09 50

cc @deathaxe, @kornelski

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -75,11 +75,6 @@
   // See also: "diagnostics_delay_ms".
   "diagnostics_additional_delay_auto_complete_ms": 0,
 
-  // Highlighting style of code diagnostics.
-  // Valid values are "fill", "box", "underline", "stippled", "squiggly" or "".
-  // When set to the empty string (""), no diagnostics are shown in-line.
-  "diagnostics_highlight_style": "squiggly",
-
   // Highlighting style of "highlights": accentuating nearby text entities that
   // are related to the one under your cursor.
   // Valid values are "fill", "box", "underline", "stippled", "squiggly" or "".
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -99,6 +99,11 @@
   // Show code actions in hover popup if available
   "show_code_actions_in_hover": true,
 
+  // Show diagnostics (error/warning squigglies) in the view. When disabled,
+  // gutter markers are still drawn, unless "diagnostics_gutter_marker" is
+  // set to "".
+  "show_diagnostics_highlights": true,
+
   // Show symbol action links in hover popup if available
   "show_symbol_action_links": true,
 
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -151,7 +151,6 @@ class Settings:
     diagnostics_additional_delay_auto_complete_ms = None  # type: int
     diagnostics_delay_ms = None  # type: int
     diagnostics_gutter_marker = None  # type: str
-    diagnostics_highlight_style = None  # type: str
     diagnostics_panel_include_severity_level = None  # type: int
     disabled_capabilities = None  # type: List[str]
     document_highlight_style = None  # type: str
@@ -187,7 +186,6 @@ class Settings:
         r("diagnostics_additional_delay_auto_complete_ms", 0)
         r("diagnostics_delay_ms", 0)
         r("diagnostics_gutter_marker", "dot")
-        r("diagnostics_highlight_style", "underline")
         r("diagnostics_panel_include_severity_level", 4)
         r("disabled_capabilities", [])
         r("document_highlight_style", "stippled")
@@ -238,9 +236,6 @@ class Settings:
 
     def document_highlight_style_to_add_regions_flags(self) -> int:
         return _settings_style_to_add_regions_flag(self.document_highlight_style)
-
-    def diagnostics_highlight_style_to_add_regions_flag(self) -> int:
-        return _settings_style_to_add_regions_flag(self.diagnostics_highlight_style)
 
 
 class ClientStates:
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -167,6 +167,7 @@ class Settings:
     show_code_actions = None  # type: bool
     show_code_actions_in_hover = None  # type: bool
     show_diagnostics_count_in_view_status = None  # type: bool
+    show_diagnostics_highlights = None  # type: bool
     show_diagnostics_in_view_status = None  # type: bool
     show_diagnostics_panel_on_save = None  # type: int
     show_diagnostics_severity_level = None  # type: int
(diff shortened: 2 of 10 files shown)


predragnikolic:
Let see where this goes. 👏
 
BTW, just noticed one problem, I get no underline if I set `"diagnostics_gutter_marker": "",` :)

rchl:
Will have a proper look tomorrow but after quick check noticed that warnings are drawn over errors whereas previously it was the other way around.

The most severe issue should be surfaced, I'd think.

Before:
Screenshot 2021-06-01 at 00 23 23

After:
Screenshot 2021-06-01 at 00 25 21

(Also noticed how gutter icon severity that is drawn is the opposite of the region severity. But that's not new and potentially unavoidable)

rchl:
Should that logic also apply to symbol highlight?

Screenshot 2021-06-01 at 13 40 33

deathaxe:
grafik

rwols:
> after quick check noticed that warnings are drawn over errors whereas previously it was the other way around.

see 9cff9d3

rwols:
> BTW, just noticed one problem, I get no underline if I set "diagnostics_gutter_marker": "", :)

This behavior also occurs on `main`, so is worth a separate pull request.

LSP.sublime-settings
```diff
@@ -78,11 +78,6 @@
   // See also: "diagnostics_delay_ms".
   "diagnostics_additional_delay_auto_complete_ms": 0,
 
-  // Highlighting style of code diagnostics.
-  // Valid values are "fill", "box", "underline", "stippled", "squiggly" or "".
-  // When set to the empty string (""), no diagnostics are shown in-line.
-  "diagnostics_highlight_style": "squiggly",
```
rwols, line 84: Would anyone be using `""`? I'm gambling no.

rwols:
The "fill" trick for unnecessary and deprecated tags is also still working

image

Anybody against merging this?

plugin/session_buffer.py
```diff
@@ -259,12 +259,13 @@ def on_diagnostics_async(self, raw_diagnostics: List[Diagnostic], version: Optio
             diagnostics_version = version
             diagnostics = []  # type: List[Tuple[Diagnostic, sublime.Region]]
             for diagnostic in raw_diagnostics:
+                region = range_to_region(Range.from_lsp(diagnostic["range"]), view)
```
rchl, line 262: Pyright doesn't like it:
```
Could not access item in TypedDict ​Pyright:
                   "range" is not a required key in "Diagnostic", so access may result in runtime exception
```

rchl: But it's just moved code so...

plugin/session_view.py
```diff
@@ -173,42 +174,44 @@ def shutdown_async(self) -> None:
...
             listener.on_diagnostics_updated_async()
 
+    def _draw_diagnostics(self, severity: int, flags: int, multiline: bool) -> None:
+        key = self.diagnostics_key(severity, multiline)
+        key_tags = {tag: '{}_tags_{}'.format(key, tag) for tag in DIAGNOSTIC_TAG_VALUES}
+        for key_tag in key_tags.values():
+            self.view.erase_regions(key_tag)
+        data = self.session_buffer.data_per_severity.get((severity, multiline))
+        if data is None:
+            self.view.erase_regions(key)
```
rchl, line 201: I guess it would be fine to remove this if now and move the data check to the next one.

LSP.sublime-settings
deathaxe: Yes, me. It's sometimes better to have gutter icons only. It doesn't distract that much as all the inline highlighting which updates too late or is normal while statements are still incomplete.

rchl: `show_diagnostics_highlights: true/false` then

plugin/session_buffer.py
rwols: Yeah. That’s because of total=False in protocol.py

LSP.sublime-settings
rwols: Can't you use `show_diagnostics_severity_level` for that (set it to 0)?

rwols: Oh, that also hides the gutter icon...

plugin/core/types.py
```diff
@@ -236,6 +219,14 @@ def r(name: str, default: Union[bool, int, str, list, dict]) -> None:
             r("inhibit_snippet_completions", False)
             r("inhibit_word_completions", True)
 
+        # Backwards-compatible with "diagnostics_highlight_style"
+        diagnostics_highlight_style = s.get("diagnostics_highlight_style")
+        if isinstance(diagnostics_highlight_style, str):
+            if diagnostics_highlight_style:
+                self.show_diagnostics_highlights = True
```
rchl, line 226: If the old setting is non empty then we should not force enable IMO.

plugin/core/types.py
```diff
@@ -236,6 +219,14 @@ def r(name: str, default: Union[bool, int, str, list, dict]) -> None:
             r("inhibit_snippet_completions", False)
             r("inhibit_word_completions", True)
 
+        # Backwards-compatible with "diagnostics_highlight_style"
+        diagnostics_highlight_style = s.get("diagnostics_highlight_style")
+        if isinstance(diagnostics_highlight_style, str) and not s.has("show_diagnostics_highlights"):
+            if diagnostics_highlight_style:
+                self.show_diagnostics_highlights = True
+            else:
+                self.show_diagnostics_highlights = False
```
rchl, line 228: Does this code sees only user overrides or the final state of the settings, including default setting? Because in the latter case the `has` check will not work as expected.

What I meant is that we could do:

```suggestion
        if isinstance(diagnostics_highlight_style, str):
            if not diagnostics_highlight_style:
                self.show_diagnostics_highlights = False
```

Since the fact that `show_diagnostics_highlights` is set to non-empty value shouldn't force the diagnostics to be enabled, even if `show_diagnostics_highlights` is `false`. The user overrode it to indicate that a different style is to be used but if he also adds `show_diagnostics_highlights: false` then that should win.

rchl: That said, your added `s.has("show_diagnostics_highlights")` I guess makes sense also but does it work as expected though?

rwols: You're probably right about `has` always being true

rchl (approved):

Issue #289, opened 2018-03-04
OSError: Bad path: D::\src\addobjectwindow.h


evandrocoan:
After setting up https://github.com/tomv564/LSP/issues/208 (cquery), I got this problem when using the go to definition feature:
```
2018-03-04 15:18:31,884: 6260: ERROR: can not send request; node process not running
Traceback (most recent call last):
  File "F:\SublimeText\sublime_plugin.py", line 389, in run_callback
    expr()
  File "F:\SublimeText\sublime_plugin.py", line 456, in <lambda>
    run_callback('on_close', callback, lambda: callback.on_close(v))
  File "F:\SublimeText\Data\Packages\LSP\plugin\core\documents.py", line 205, in on_close
    Events.publish("view.on_close", view)
  File "F:\SublimeText\Data\Packages\LSP\plugin\core\events.py", line 28, in publish
    listener(*args)
  File "F:\SublimeText\Data\Packages\LSP\plugin\core\documents.py", line 158, in notify_did_close
    params = {"textDocument": {"uri": filename_to_uri(file_name)}}
  File "F:\SublimeText\Data\Packages\LSP\plugin\core\url.py", line 8, in filename_to_uri
    return urljoin('file:', pathname2url(path))
  File "./python3.3/nturl2path.py", line 58, in pathname2url
OSError: Bad path: D::\src\addobjectwindow.h
```

From the error, we can see it is adding an additional colon after the `D:` letter.

```
22:00:00:000.000000 0.00e+00 - LSP.plugin.core.rpc.No Function:0 - LSP server: 2018-03-04 16:54:22.683 (  86.447s) [scan_includes]          utils.cc:319      0| Reading D:/tmp/cquery/D@@User@Dropbox@SoftwareVersioning@GraphicsComputing@src@mainwindow.cpp
16:54:22:687.258959 1.53e-03 - LSP.plugin.core.rpc.notification_handler:384 - <--  $cquery/progress
[...]

PR #1711, opened 2021-05-31
Open links with absolute filepaths in a new view

jwortmann:
The Julia language server uses Markdown links in hover popups with absolute paths to other files (but without `file://` schema) and a :row suffix. For example `C:\\\\path\\\\to\\\\file.jl:123`, which corresponds to `C:\path\to\file.jl:123` without the JSON and Markdown escaping for the backslashes (on Windows).

This commit checks if the path describes an existing file and allows to open such links in a new view, instead of in a webbrowser.

I didn't find an exact description in the LSP specs how links to files should be specified (other than that Markdown should follow GFM), so I guess it's up to the client what kind of link formats are supported. I guess this implementations seems safe to not accidentally open actual weblinks in the editor.

### DIFF
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -14,6 +14,7 @@ from .core.views import diagnostic_severity
 from .core.views import first_selection_region
 from .core.views import format_diagnostic_for_html
 from .core.views import FORMAT_MARKED_STRING, FORMAT_MARKUP_CONTENT, minihtml
+from .core.views import is_file
 from .core.views import is_location_href
 from .core.views import make_command_link
 from .core.views import make_link
@@ -24,6 +25,7 @@ from .core.views import update_lsp_popup
 from .core.windows import AbstractViewListener
 from urllib.parse import urlparse
 import functools
+import os
 import sublime
 import webbrowser
 
@@ -212,6 +214,10 @@ class LspHoverCommand(LspTextCommand):
                 position = {"line": row, "character": col_utf16}  # type: Position
                 r = {"start": position, "end": position}  # type: RangeLsp
                 sublime.set_timeout_async(functools.partial(session.open_uri_async, uri, r))
+        elif os.path.isabs(href) and is_file(href):
+            window = self.view.window()
+            if window:
+                window.open_file(href, flags=sublime.ENCODED_POSITION)
         else:
             # NOTE: Remove this check when on py3.8.
             if not (href.lower().startswith("http://") or href.lower().startswith("https://")):
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -704,6 +704,15 @@ def is_location_href(href: str) -> bool:
     return href.startswith("location:")
 
 
+def is_file(href: str) -> bool:
+    """
+    Check whether this href is a path to an existing file, possibly with :row or :row:col suffix.
+    """
+    drive, tail = os.path.splitdrive(href)
+    path = tail.split(':')[0]
+    return os.path.isfile(drive + path)
+
+
 def _format_diagnostic_related_info(
     config: ClientConfig,
     info: DiagnosticRelatedInformation,


plugin/hover.py
```diff
@@ -212,6 +214,10 @@ def _on_navigate(self, href: str, point: int) -> None:
                 position = {"line": row, "character": col_utf16}  # type: Position
                 r = {"start": position, "end": position}  # type: RangeLsp
                 sublime.set_timeout_async(functools.partial(session.open_uri_async, uri, r))
+        elif os.path.isabs(href) and is_file(href):
```
rwols, line 217: This won't work when the Julia LS is running in a Docker or other remote location. You don't expect such scenarios?

rwols: Shouldn't this be fixed in the Julia LS? There's already a handler for `file:` schemes.

jwortmann: > This won't work when the Julia LS is running in a Docker or other remote location. You don't expect such scenarios?

Good to know, I did not consider this. I have very little experience with Docker, so I cannot tell if this is a realistic scenario or not.

I've looked at the language server code and saw that they even have special casing only for VS Code to use URIs with `file:` scheme (https://github.com/julia-vscode/LanguageServer.jl/blob/8d8408edb3b4bd8d1266901b9aac3f540b04fdfe/src/requests/hover.jl#L114-L116). So I've adjusted the code in the server to do the same for Sublime Text and wanted to prepare a PR there, but it still doesn't work because apparently we can't handle percent-encoding for `:` in paths (after the drive letter) - the server reports for example URIs like `file:///C%3A/Users/jwortmann/AppData/Local/Programs/Julia-1.6.1/share/julia/base/iobuffer.jl#473` in that case, but Sublime cannot open the filepath starting with `C%3A`. Iirc we had seen this problem earlier already, but I don't remember whether the percent-encoding is strictly neccessary or not. According to https://en.wikipedia.org/wiki/Percent-encoding#Reserved_characters the colon is a reserved character. But regardless of that, the percent-encoding is probably allowed, so I guess we should handle that on the client side?

jwortmann: According to https://en.wikipedia.org/wiki/Uniform_Resource_Identifier the percent-encoding is required for the `:` in the path:
> Of the ASCII character set, the characters : / ? # [ ] @ are reserved for use as delimiters of the generic URI components and must be percent-encoded

We can just use `urllib.parse.unquote` to decode the URI. But ST still cannot open the file (at least for me on Windows), probably because the path begins with a `/`. But that URI with 3 `/` seems still correct and only 2 slashes would actually be invalid: [https://en.wikipedia.org/wiki/File_URI_scheme#How_many_slashes?](https://en.wikipedia.org/wiki/File_URI_scheme#How_many_slashes?)

So I guess we also need to strip the leading `/` from the path first.
How/why do `file:` URIs work for other servers with the current implementation?

rwols: It doesn't matter what specs say when Microsoft is involved. What this code does is the authority: https://github.com/microsoft/vscode-uri

jwortmann: Well, I'm mostly interested in getting the language server with ST to work, and apparently they are following the specs, but we don't support the specs here. Since the server targets VS Code in the first place, that would imply that VS Code does also support the specs, right?

rwols: I haven't read through the RFCs but you're probably correct. The current file: scheme handling in hover.py works well for Linux and macOS, but probably doesn't work well for Windows. There is still doubt about what to do in plugin/core/url.py. Not many people dare to change it because the major contributors to this package are not using Windows.

I would have expected [urlparse](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) to handle the complexity for us, but apparently it doesn't. Your proposal of using urllib.parse.unquote sounds appealing. See also this issue for some history https://github.com/sublimelsp/LSP/issues/289 on why we have this in url.py in the first place.

rwols: BTW, after applying `unquote` I've been toying with the solution of applying `re.sub` with regex `r'^/([a-zA-Z]:)'`:

```
>>> re.sub(r'^/([a-zA-Z]:)', r'\1', '/C:/foo/bar')
'C:/foo/bar'
```

jwortmann: Yeah the documentation for `urlparse` says
> The delimiters as shown above are not part of the result, except for a leading slash in the path component, which is retained if present.

which is expected on Linux/macOS, but results in an additional `/` preceding the drive letter on Windows and ST doesn't like that.

My solution would be something like this:
```python
parsed = urlparse(unquote(href))
filepath = parsed.path
if sublime.platform() == "windows" and filepath[0] == "/":
    filepath = filepath[1:]
fn = "{}:{}".format(filepath, parsed.fragment) if parsed.fragment else parsed.path
```

jwortmann:
I've made another PR to fix the file: URI handling with percent-encoded characters and possible slash preceding the drive letter on Windows.

This can be closed now, assuming the Julia server will change link destinations to use a file: URI for ST, instead of the absolute_path:row format. I will prepare a PR there.

PR #1712, opened 2021-06-01
Use box-style for multiline doc highlights

rwols:
Related: #1710

I've removed the "squiggly", "dotted" and "box" options for single-line highlights. "fill" and "underline" are possible. When set to "fill", "fill" is used for multi-line as well.

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -190,7 +190,7 @@ class Settings:
         r("diagnostics_highlight_style", "underline")
         r("diagnostics_panel_include_severity_level", 4)
         r("disabled_capabilities", [])
-        r("document_highlight_style", "stippled")
+        r("document_highlight_style", "underline")
         r("log_debug", False)
         r("log_max_size", 8 * 1024)
         r("lsp_code_actions_on_save", {})
@@ -236,8 +236,11 @@ class Settings:
 
         set_debug_logging(self.log_debug)
 
-    def document_highlight_style_to_add_regions_flags(self) -> int:
-        return _settings_style_to_add_regions_flag(self.document_highlight_style)
+    def document_highlight_style_to_add_regions_flags(self) -> Tuple[int, int]:
+        if self.document_highlight_style == "fill":
+            return sublime.DRAW_NO_OUTLINE, sublime.DRAW_NO_OUTLINE
+        else:
+            return sublime.DRAW_NO_FILL, sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE | sublime.DRAW_SOLID_UNDERLINE
 
     def diagnostics_highlight_style_to_add_regions_flag(self) -> int:
         return _settings_style_to_add_regions_flag(self.diagnostics_highlight_style)
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -236,7 +236,7 @@ class Settings:
 
         set_debug_logging(self.log_debug)
 
-    def document_highlight_style_to_add_regions_flags(self) -> Tuple[int, int]:
+    def document_highlight_style_region_flags(self) -> Tuple[int, int]:
         if self.document_highlight_style == "fill":
             return sublime.DRAW_NO_OUTLINE, sublime.DRAW_NO_OUTLINE
         else:
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -43,11 +43,11 @@ from functools import partial
 from weakref import WeakSet
 from weakref import WeakValueDictionary
 import functools
+import itertools
 import sublime
 import sublime_plugin
 import textwrap
 import webbrowser
-import itertools
 
 
 SUBLIME_WORD_MASK = 515
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -343,8 +343,9 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         if different:
             if not self._is_in_higlighted_region(current_region.b):
                 self._clear_highlight_regions()
-            self._when_selection_remains_stable_async(self._do_highlights_async, current_region,
-                                                      after_ms=self.highlights_debounce_time)
+            if userprefs().document_highlight_style:
+                self._when_selection_remains_stable_async(self._do_highlights_async, current_region,
+                                                          after_ms=self.highlights_debounce_time)
             self._clear_code_actions_annotation()
             if userprefs().show_code_actions:
                 self._when_selection_remains_stable_async(self._do_code_actions, current_region,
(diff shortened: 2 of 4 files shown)


rchl:
The comment above the `document_highlight_style` should be updated also.

rchl:
The `fill` style looks bad but it was like that before also:

Screenshot 2021-06-02 at 09 06 29

rchl:
In `on_selection_modified_async` there should also be a user pref check added.

plugin/documents.py
```diff
@@ -597,13 +598,21 @@ def _unresolved_code_lenses(
...
+            self.view.erase_regions(self._highlights_key(kind, False))
+            self.view.erase_regions(self._highlights_key(kind, True))
 
     def _is_in_higlighted_region(self, point: int) -> bool:
         for kind in range(1, 4):
-            regions = self.view.get_regions("lsp_highlight_{}".format(_kind2name[kind]))
+            regions = self.view.get_regions(self._highlights_key(kind, False))
+            for r in regions:
+                if r.contains(point):
+                    return True
```
rchl, line 614: Sorry, not an issue but could be shorter?
```suggestion
            if any((r.contains(point) for r in regions)):
                return True
```
especially since it's repeated twice

plugin/core/types.py
```diff
@@ -236,8 +236,11 @@ def r(name: str, default: Union[bool, int, str, list, dict]) -> None:
 
         set_debug_logging(self.log_debug)
 
-    def document_highlight_style_to_add_regions_flags(self) -> int:
-        return _settings_style_to_add_regions_flag(self.document_highlight_style)
+    def document_highlight_style_to_add_regions_flags(self) -> Tuple[int, int]:
```
rchl, line 239: Nit. I would just name it `document_highlight_style_region_flags` since from the perspective of the caller, the fact that it maps the user setting to to regions shouldn't be relevant.

plugin/documents.py
```diff
@@ -621,31 +630,25 @@ def _do_highlights_async(self) -> None:
...
-                    scope = _kind2scope[kind]
-                    self.view.add_regions(
-                        "lsp_highlight_{}".format(_kind2name[kind]),
-                        regions,
-                        scope=scope,
-                        flags=flags)
+                    flags = flags_multi if multiline else flags_single
+                    self.view.add_regions(key, regions, scope=_kind2scope[kind], flags=flags)
+                else:
+                    self.view.erase_regions(key)
```
rchl, line 651: There is a `self._clear_highlight_regions()` call at the beginning. Is this really needed?

rchl:
> In `on_selection_modified_async` there should also be a user pref check added.

Still not addressed. Otherwise looks very good.

rchl (approved):
Perfect

agnivade:
I am very surprised with this change. I would be totally fine with changing the default to `underline`, but removing the other settings altogether seems like an extreme step, and is affecting my setup choices.

I do not like `underline` at all, and want to use `stippled`. What choice do I have as an user apart from directly hot-wiring the code?

rwols:
Thanks for actually looking at possible customization points. The vast majority of users don't actually do that.

We traded settings-maintenance-burden complexity with overall better default behavior here. This project is maintained by volunteers in their free time so if you really need `stippled` customizability, feel free to make a pull request **and** maintain the behavior in the long run.

What you can also do is use `fill` and set up your color scheme overrides (by running _UI: Customize Color Scheme_ from the command palette) so that it'll behave somewhat like VSCode.

So for instance for my Mariana color scheme I have this override file:

```jsonc
{
    "rules": [
        {
            "scope": "markup.error",
            "foreground": "hsl(357, 79%, 65%)"
        },
        {
            "scope": "markup.warning",
            "foreground": "hsl(32, 93%, 66%)"
        },
        {
            "scope": "markup.info",
            "foreground": "hsl(0, 0%, 100%)"
        },
        {
            "scope": "markup.info.hint",
            "foreground": "hsl(0, 0%, 97%)"
        },
        {
            "scope": "markup.highlight.text",
            "background": "color(var(white2) alpha(0.2))"
        },
        {
            "scope": "markup.highlight.read",
            "background": "color(var(green) alpha(0.2))"
        },
        {
            "scope": "markup.highlight.write",
            "background": "color(var(red2) alpha(0.2))"
        },
        {
            "scope": "markup.unnecessary.lsp",
            "background": "color(var(blue3) alpha(0.9))"
        }
    ],
}
```


resulting in:

image

agnivade:
Thanks, yep totally with you on the burdens of open source. Doesn't look like I can do a stippled line via a color scheme override, so I'll have to maintain my own git patch.

Thanks for all your work!

jwortmann:
You could for example set the underline color to 50% opacity via "foreground" in the color scheme rule, then it doesn't look much different from the dotted style:

stippled

underline

But I must say I actually considered to suggest keeping the "stippled" style option when I saw this PR, but then I changed my mind because I don't need it personally and the default has been "underline" even before iirc. I guess the motivation was that "stippled" is now used for the diagnostics with severity `info` and `hint`, so currently it's ensured that the document highlights are not only distinguishable by color, but also by the drawing style.

However, the "settings maintenance burden" is only two lines of code (plus re-adding that option in the settings description and .sublime-package schema). Maybe it's reasonable to keep that style option?
```diff
diff --git a/plugin/core/types.py b/plugin/core/types.py
index 49fdd3b..512bdcb 100644
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -235,6 +235,8 @@ class Settings:
     def document_highlight_style_region_flags(self) -> Tuple[int, int]:
         if self.document_highlight_style == "fill":
             return sublime.DRAW_NO_OUTLINE, sublime.DRAW_NO_OUTLINE
+        elif self.document_highlight_style == "stippled":
+            return sublime.DRAW_NO_FILL, sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE | sublime.DRAW_STIPPLED_UNDERLINE
         else:
             return sublime.DRAW_NO_FILL, sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE | sublime.DRAW_SOLID_UNDERLINE
```

agnivade:
>  I guess the motivation was that "stippled" is now used for the diagnostics with severity info and hint, so currently it's ensured that the document highlights are not only distinguishable by color, but also by the drawing style.

Yep I get that. But 99% of the time, I place my cursor under a symbol to get a feeling of other symbols around it. That overshadows my workflow far more than the times I need diagnostics for info and hint. (And if needed, I'd want to be able to customize it).

> However, the "settings maintenance burden" is only two lines of code (plus re-adding that option in the settings description and .sublime-package schema). Maybe it's reasonable to keep that style option?

I'd really like that. However, I think I might fall into the 1% here, so I'd let you guys take the call. Thanks for pasting the snippet btw!

thewchan:
I'd like to add myself to the 1% here, it was kinda confusing and disappointing to see those appearance customization options disappeared.

PR #1716, opened 2021-06-02
In an early exit continue handling listeners

rwols:
This isn't all that pretty but this function should be async-ified anyway.


rchl (approved):

predragnikolic:
Currently I do not have time for a full report, But here is a short one.

If I open a `index.js` file supported by LSP-typesript and LSP-tailwindcss and don't have `tailwind.config.js` file.
The LSP-typescript server will initialize,
but the LSP-tailwindcss will show the error `cannot start LSP-tailwindcss: No tailwind.config.js ...`.
And the LSP-tailwind will try to install again and I will see in the status bar `LSP-tailwindcss: Installing ...`, 
and try to start again, but will again show the message `cannot start LSP-tailwindcss: No tailwind.config.js ...`. 
And it will repeat that forever.

Will look into this tonight.

rwols:
Whoops! Yeah this async server spawning without asyncio is certainly one of the most confusing and dangerous parts of the code base now.

rwols:
The config should probably be disabled in the project to make it stop.

rwols:
Right, that’s what I meant.

predragnikolic:
Is there a way to disable a session for the current workspace?

rwols:
What do you mean by workspace? A window or a folder?

predragnikolic:
In ST I can drag multiple folders in the sidebar.

Lets say that I drag 2 folders.
In one workspace folder (folderA) LSP-tailwindcss can start, but in the other it cannot (folderB).

If I a file supported by LSP-tailwind in `folderA`
The LSP-tailwindcss will start the server.

But if I open a file in `folderB`, becase the `folderB` doesn't meet the requirements for the LSP-tailwindcss to start, 
if we disable the current session, will that also disable/stop the LSP-tailwindcss server in the `folderA` (I think that it stops the server process, or I am wrong)?

The question is can we disable the session per workspace folder? (so that disabling it in one workspace folder wont stop/disable the session in another workspace folder)

rwols:
No, only disabling per window(/project) or for the whole application is implemented.

plugin/core/windows.py
```diff
@@ -315,6 +315,9 @@ def start_async(self, config: ClientConfig, initiating_view: sublime.View) -> No
                 if cannot_start_reason:
                     config.erase_view_status(initiating_view)
                     message = "cannot start {}: {}".format(config.name, cannot_start_reason)
+                    # Continue with handling pending listeners
```
predragnikolic, line 318:
```suggestion
                    self._configs.disable_config(config.name, only_for_session=True)
                    # Continue with handling pending listeners
```

predragnikolic (approved):
This works, thanks

PR #1717, opened 2021-06-03
Hover code action show remains consistent

husanjun:
if the `"show_code_actions": ""` then code action also shouldn't show in the hover

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -101,6 +101,9 @@
   // using a keyboard shortcut or the context menu.
   "show_code_actions": "annotation",
 
+  // Show code actions in hover popup if available
+  "show_code_actions_in_hover": true,
+
   // Show symbol action links in hover popup if available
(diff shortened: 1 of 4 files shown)


rchl:
I am not sure about this.
I can easily imagine a use case where someone doesn't want to be bothered about the visual noise but would still want to have access to those when triggering explicit action like hover.

husanjun:
> I am not sure about this.
> I can easily imagine a use case where someone doesn't want to be bothered about the visual noise but would still want to have access to those when triggering explicit action like hover.

if so, should we add a setting for the user to choose, because there must be users who don't want to show it, such as me

rwols:
> if so, should we add a setting for the user to choose, because there must be users who don't want to show it, such as me

Agreed. How about a boolean setting `"show_code_actions_in_hover"`?

rchl (approved):

rwols (approved):

PR #1722, opened 2021-06-04
Handle percent-encoding and leading slash in file URIs

jwortmann:
... when used in Markdown link destinations in hover popups.

Resulting from the discussion in #1711.

I've only tested on Windows.


rwols (approved):
Maybe you can take a look at plugin/core/url.py as well? The logic there is incorrect for UNC paths like `\\192.168.0.100\foo\bar` from what I remember.

PR #1724, opened 2021-06-04
Fix the `LSP Development` command

Daggy1234:
With this fix the tooling command for generating a schema is fixed. As discussed in the discord server!. Just adds a few is-instance checks!

### DIFF
--- a/plugin/tooling.py
+++ b/plugin/tooling.py
@@ -167,47 +167,48 @@ class LspParseVscodePackageJson(sublime_plugin.ApplicationCommand):
         self.view.assign_syntax("Packages/JSON/JSON.sublime-syntax")
         self.writeline("{")
         for k, v in sorted(properties.items()):
-            if isinstance(v, dict):
-                description = v.get("description")
-                if not isinstance(description, str):
-                    description = v.get("markdownDescription")
-                if isinstance(description, str):
-                    for line in description.splitlines():
-                        for wrapped_line in textwrap.wrap(line, width=100 - 8 - 3):
-                            self.writeline4('// {}'.format(wrapped_line))
+            if not isinstance(v, dict):
+                continue
+            description = v.get("description")
+            if not isinstance(description, str):
+                description = v.get("markdownDescription")
+            if isinstance(description, str):
+                for line in description.splitlines():
+                    for wrapped_line in textwrap.wrap(line, width=100 - 8 - 3):
+                        self.writeline4('// {}'.format(wrapped_line))
+            else:
+                self.writeline4('// unknown setting')
+            enum = v.get("enum")
+            has_default = "default" in v
+            default = v.get("default")
+            if isinstance(enum, list):
+                self.writeline4('// possible values: {}'.format(", ".join(enum)))
+            if has_default:
+                value = default
+            else:
+                typ = v.get("type")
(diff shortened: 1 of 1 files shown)


rwols (changes requested):
can you instead use this?

```diff
diff --git a/plugin/tooling.py b/plugin/tooling.py
index 9bcdc6f..21febe0 100644
--- a/plugin/tooling.py
+++ b/plugin/tooling.py
@@ -1,4 +1,5 @@
 from .core.css import css
+from .core.logging import debug
 from .core.registry import windows
 from .core.sessions import get_plugin
 from .core.transports import create_transport
@@ -41,7 +42,12 @@ def _preprocess_properties(translations: Optional[Dict[str, str]], properties: D
 
     - Replace description translation placeholders by their English translation
     - Discard the "scope" key
+    - Removes key/values whose value is not a dict
     """
+    non_dict_keys = [k for k, v in properties.items() if not isinstance(v, dict)]
+    for k in non_dict_keys:
+        debug("discarding key:", k)
+        properties.pop(k)
     for v in properties.values():
         v.pop("scope", None)
         descr = v.get("description")
```

rwols:
The exception is caused by these two lines: https://github.com/rust-analyzer/rust-analyzer/blob/d647568db161db585f4f2ee3af9160e445338ac1/editors/code/package.json#L392

https://github.com/rust-analyzer/rust-analyzer/blob/d647568db161db585f4f2ee3af9160e445338ac1/editors/code/package.json#L844

plugin/tooling.py
```diff
@@ -161,46 +167,47 @@ def run(self, base_package_name: str) -> None:
...
-                elif typ == "boolean":
-                    value = False
-                elif typ == "array":
-                    value = []
-                elif typ == "object":
-                    value = {}
-                elif typ == "number":
-                    value = 0
+            if isinstance(v, dict):
+                description = v.get("description")
```
rwols, line 171: Can you invert the condition here?

```
if not isinstance(v, dict):
    continue
```

rwols (changes requested):

plugin/tooling.py
```diff
@@ -161,46 +167,47 @@ def run(self, base_package_name: str) -> None:
...
-                    value = ""
-                elif typ == "boolean":
-                    value = False
-                elif typ == "array":
-                    value = []
-                elif typ == "object":
-                    value = {}
-                elif typ == "number":
-                    value = 0
+            if isinstance(v, dict):
```
rchl, line 170: Can you instead do?
```suggestion
            if not isinstance(v, dict):
               continue
```
Because the current way increases indentation of the whole block which is not ideal.

rchl (approved):

plugin/tooling.py
```diff
@@ -161,6 +167,8 @@ def run(self, base_package_name: str) -> None:
         self.view.assign_syntax("Packages/JSON/JSON.sublime-syntax")
         self.writeline("{")
         for k, v in sorted(properties.items()):
+            if not isinstance(v, dict):
```
rwols, line 170: I don’t think this check is actually necessary now. But better safe than sorry I suppose :)

rwols (approved):

PR #1733, opened 2021-06-09
Support restart specific server

husanjun:
The from https://github.com/sublimelsp/LSP/pull/1023

image

1. If there was only one server it would be the same as before
2. If there are more than one server, the server options are listed, and there are options to restart all of them
3. Compatible with command line running and support running single server parameters


rchl:
Did the fix not work as expected or what's the reason for closing?

I think it's a welcome change but a bit tricky to do the last time I've looked at it so I wanted to find proper time to review it.

plugin/core/registry.py
```diff
@@ -101,11 +101,36 @@ def sessions(self, capability: Optional[str] = None) -> Generator[Session, None,
         yield from sessions_for_view(self.view, capability)
 
 
-class LspRestartClientCommand(sublime_plugin.TextCommand):
-    def run(self, edit: Any) -> None:
+class LspRestartServerCommand(LspTextCommand):
+    def run(self, edit: Any, config_name: str = None, show_quick_panel: bool = False) -> None:
```
rwols, line 105: Where is the parameter `config_name` supposed to be used?

plugin/core/registry.py
```diff
@@ -101,11 +101,36 @@ def sessions(self, capability: Optional[str] = None) -> Generator[Session, None,
...
+                self.restart_server(0)
+            else:
+                window.show_quick_panel(self._config_names, self.restart_server)
+
+    def restart_server(self, index: int) -> None:
+        if index > -1:
+
+            def run_async() -> None:
+                config_name = self._config_names[index]
+                if not config_name or not self.session_by_name(config_name):
```
rchl, line 127: It would be nice to be more explicit here and instead of restarting all sessions when config_name doesn't match any session, check for "All Servers" specifically. There could be a class constant for "All Server" to make it DRY.

plugin/core/registry.py
```diff
@@ -101,11 +101,36 @@ def sessions(self, capability: Optional[str] = None) -> Generator[Session, None,
...
+                    self._wm.restart_sessions_async()
+                else:
+                    self._wm.end_config_sessions_async(config_name)
+                    self._wm.register_listener_async(self._wm.listener_for_view(self.view))  # type: ignore
```
rchl, line 131: Please don't ignore types. It's easy to handle the type issue with an extra "if".

plugin/core/registry.py
```diff
@@ -101,11 +101,36 @@ def sessions(self, capability: Optional[str] = None) -> Generator[Session, None,
...
+            self.restart_server(0)
+        else:
+            self._config_names.insert(0, 'All Servers')
+            if not show_quick_panel:
+                self.restart_server(0)
+            else:
+                window.show_quick_panel(self._config_names, self.restart_server)
+
+    def restart_server(self, index: int) -> None:
+        if index > -1:
```
rchl, line 123: Style nit but avoid extra indentation (an early return) would be preferred

plugin/core/registry.py
rchl: Probably would be useful to simplify https://github.com/sublimelsp/LSP-intelephense/pull/63

rwols: Ah, I understand now.

Default.sublime-commands
```diff
@@ -52,8 +53,11 @@
         "command": "lsp_format_document_range",
     },
     {
-        "caption": "LSP: Restart Servers",
-        "command": "lsp_restart_client",
+        "caption": "LSP: Restart Server",
+        "command": "lsp_restart_server",
+        "args": {
+            "show_quick_panel": true
+        }
```
rwols, line 60: If the `config_name` parameter is `None`, wouldn't that already imply we have to show a quick panel? I think you can remove the `show_quick_panel` parameter.

rwols (changes requested):

plugin/core/registry.py
```diff
@@ -95,11 +95,39 @@ def sessions(self, capability_path: Optional[str] = None) -> Generator[Session,
...
+            self._config_names.insert(0, self.ALL_SERVERS)
+            window.show_quick_panel(self._config_names, self.restart_server)
+
+    def restart_server(self, index: int) -> None:
+        if index < 0:
+            return
+
+        def run_async() -> None:
+            config_name = self._config_names[index]
+            if not config_name or config_name != self.ALL_SERVERS:
```
rchl, line 122:
```suggestion
            if not config_name or config_name == self.ALL_SERVERS:
```

husanjun: Thanks!

rchl: I'm thinking that the "All servers" option is a bit inconsistent because when there are active servers in the view, it will be available and restart **all** servers, even the ones that are not active on the page.

On the other hand, if there is only one server or none, this option won't even show up.

Should this option:
 - only restart all **active** servers in the file
 - or be always available, as long as there are any active servers somewhere?
 - not be provided at all?

The last option would be the most consistent probably. The other would require some more non-trivial changes.

husanjun: Well, it's not common to restart all services, I agree

rchl (approved):

rwols:
Thanks, @husanjun

PR #1736, opened 2021-06-09
Mention deno in docs

rwols:


docs/src/language_servers.md
```diff
@@ -509,7 +509,8 @@ Follow installation instructions on [LSP-tagml](https://github.com/HuygensING/LS
 
 ## TypeScript / JavaScript
 
-Follow installation instructions on [LSP-typescript](https://github.com/sublimelsp/LSP-typescript).
+- Node.js: Follow installation instructions on [LSP-typescript](https://github.com/sublimelsp/LSP-typescript).
+- deno.land: Follow installation instructions on [LSP-Deno](https://github.com/sublimelsp/LSP-Deno).
```
rchl, line 513: I believe it's called `Deno`. The `deno.land` is just a domain of the website.

rchl, line 513: Saying "Node.js" is maybe too narrow because the server can be used for plain JS without the Node server also. Could reformat it similar to:

```suggestion
Follow installation instructions on [LSP-typescript](https://github.com/sublimelsp/LSP-typescript).

For development with the Deno framework follow installation instructions on [LSP-Deno](https://github.com/sublimelsp/LSP-Deno).
```

PR #1742, opened 2021-06-11
Remove sessions_for_view

### LINKED ISSUE #1741: Don't send codeAction request for file previews
This client sends the `textDocument/codeAction` request even for file previews of selected items in the quick panel, which appears in case of multiple location results for "Goto Definition...", but without sending a `textDocument/didOpen` notification first. This may lead to a server crash, because apparently VSCode behaves different and servers might expect the VSCode behavior, for more context see https://github.com/julia-vscode/LanguageServer.jl/issues/951#issuecomment-858783505. Here is a video with the payload (and where I've fixed the server code to not crash as a workaround):

https://user-images.githubusercontent.com/6579999/121701934-b5159c00-cad1-11eb-86ec-1c2c6a972956.mp4

I don't know if this is regulated somewhere in the specs, or if the server should handle these requests even before `didOpen`. But regardless of that, I think it is neither necessary nor useful to send the code action request for the file previews. Instead, we should wait until the user confirms the selection with <kbd>Enter</kbd> and only send the code action request after `didOpen`.

- OS: Windows 10
- Sublime Text version: 4107
- LSP version: main branch
- Language servers used: Julia language server (LanguageServer.jl)

rwols:
Fixes #1741

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -5,7 +5,6 @@ from .core.protocol import Diagnostic
 from .core.protocol import Error
 from .core.protocol import Request
 from .core.registry import LspTextCommand
-from .core.registry import sessions_for_view
 from .core.registry import windows
 from .core.sessions import SessionBufferProtocol
 from .core.settings import userprefs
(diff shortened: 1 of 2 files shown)


rchl (approved):
Good to finally get rid of it.

PR #1744, opened 2021-06-11
Use "_async" suffix in listener APIs

rchl:
Just a tiny refactor to include the _async suffix and add
a "sessions_async" API to listener for parity.


rchl:
Renaming made me aware that we are using async `session_async` API within `on_query_context`. But I guess there is no way around that.

rwols (approved):

PR #1747, opened 2021-06-12
Fixup window/showDocument request handler

rwols:
The selection is optional.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -8,7 +8,7 @@ from .open import open_externally
 from .progress import WindowProgressReporter
 from .promise import PackagedTask
 from .promise import Promise
-from .protocol import CodeAction, Location, LocationLink
+from .protocol import CodeAction, Location, LocationLink, Position
 from .protocol import Command
 from .protocol import CompletionItemTag
 from .protocol import Diagnostic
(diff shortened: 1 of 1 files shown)


plugin/core/sessions.py
```diff
@@ -1089,11 +1089,18 @@ def run_code_action_async(self, code_action: Union[Command, CodeAction], progres
...
+        r: Optional[RangeLsp],
+        flags: int = 0,
+        group: int = -1
+    ) -> Promise[bool]:
         if uri.startswith("file:"):
             # TODO: open_file_and_center_async seems broken for views that have *just* been opened via on_load
             path = self.config.map_server_uri_to_client_path(uri)
-            self.window.open_file(to_encoded_filename(path, r["start"]), flags | sublime.ENCODED_POSITION, group)
+            pos = r["start"] if r else {"line": 0, "character": 0}  # type: Position
+            self.window.open_file(to_encoded_filename(path, pos), flags | sublime.ENCODED_POSITION, group)
```
rchl, line 1103: I guess another option would be to not position at all when not provided.
Could make a different (for better or worse, not sure) when the file is already open.

rchl (approved):

plugin/core/sessions.py
rwols: Yes, good idea for improvement

PR #1748, opened 2021-06-13
Add back stippled option for document highlights

jwortmann:
Because it is requested in #1712 by multiple users and there is already the setting for it, so I think it doesn't hurt to keep this option.

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -81,7 +81,7 @@
 
   // Highlighting style of "highlights": accentuating nearby text entities that
   // are related to the one under your cursor.
-  // Valid values are "fill", "underline", or "".
+  // Valid values are "fill", "underline", "stippled", or "".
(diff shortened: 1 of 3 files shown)


thewchan:
Would other options like box it filled also be available as well?

agnivade:
Thanks @jwortmann !

jwortmann:
> Would other options like box it filled also be available as well?

You mean use the box style instead of background color for multiline regions, if the setting is set to "fill"?
The box style was introduced because it doesn't look well if the underline style is applied to big parts of a file (which only certain servers do). But I think that is no problem for the "fill" style:

fill

So another option for this is currently not supported, and I don't think it is necessary. Do you think it would be useful?

thewchan:
This might be a weird me thing but I mostly stay away for the underline squiggles I possible because I basically reserve that or spell check...

rwols (approved):

PR #1750, opened 2021-06-15
Add setting to render code lenses as a phantom

Rapptz:
This also adds a setting to disable the rendering entirely.

In order to satisfy those who like their code lenses as an annotation I kept the setting as-is rather than replacing them. I did change the default setting to use a phantom since that representation is probably more common for those coming from VSCode.

I've tested this code and it works, but I'm unsure if there are any state related bugs. One thing I noticed is that the code lenses only show up when the file is reopened. This felt undesirable to me, however I did not know where exactly to fix it. In a normal plugin lifecycle I would hook into `on_load_async`, `on_activated_async`, and (potentially) `on_post_save_async`. There was the `_register_async` helper function which would have felt like an appropriate place to send the code lenses request, but I was unsure if the state was properly set there or if this was even desirable. Personally, reopening the file to get code lenses is a bad experience.

Code lenses with the `phantom` setting are rendered similarly to VSCode, as seen below:

| Sublime Text | VSCode |
| :---: | :---: |
| image | image |

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -97,6 +97,11 @@
   // using a keyboard shortcut or the context menu.
   "show_code_actions": "annotation",
 
+  // Show code lens:
+  // "annotation" - show an annotation on the right when code actions are available
+  // "phantom" - show a phantom on the top when code actions are available
+  "show_code_lens": "annotation",
+
   // Show code actions in hover popup if available
   "show_code_actions_in_hover": true,
 
--- a/plugin/code_lens.py
+++ b/plugin/code_lens.py
@@ -67,6 +67,7 @@ class CodeLensView:
     def __init__(self, view: sublime.View) -> None:
         self.view = view
         self._init = False
+        self._phantom = sublime.PhantomSet(view, self.CODE_LENS_KEY)
         self._code_lenses = {}  # type: Dict[Tuple[int, int], List[CodeLensData]]
 
     def clear(self) -> None:
@@ -86,6 +87,7 @@ class CodeLensView:
         return '{0}.{1}'.format(self.CODE_LENS_KEY, index)
 
     def clear_view(self) -> None:
+        self._phantom.update([])
         self.clear_annotations()
 
     def handle_response(self, session_name: str, response: List[CodeLens]) -> None:
@@ -136,10 +138,36 @@ class CodeLensView:
             if lens.is_resolved():
                 yield lens
 
-    def render(self) -> None:
-        accent = self.view.style_for_scope("region.greenish markup.accent.codelens.lsp")["foreground"]
-        for index, lens in enumerate(self._flat_iteration()):
-            self.view.add_regions(self._region_key(index), [lens.region], "", "", 0, [lens.small_html], accent)
+    def _get_phantom_region(self, region: sublime.Region) -> sublime.Region:
+        """Takes the current region and puts the phantom on the line above, properly indented.
+
+        Code-lenses are typically shown above the code in VSCode.
+        """
+        line = self.view.substr(self.view.line(region))
+        offset = 0
+        for ch in line:
+            if ch.isspace():
+                offset += 1
+            else:
+                break
+
+        row, _ = self.view.rowcol_utf16(region.a)
+        point = self.view.text_point_utf16(max(row - 1, 0), offset, clamp_column=True)
+        return sublime.Region(point)
+
+    def render(self, mode: str) -> None:
+        if mode == 'annotation':
+            accent = self.view.style_for_scope("region.greenish markup.accent.codelens.lsp")["foreground"]
+            for index, lens in enumerate(self._flat_iteration()):
+                self.view.add_regions(self._region_key(index), [lens.region], "", "", 0, [lens.small_html], accent)
+        else:  # 'phantom'
+            phantoms = []
+            for key, group in self._code_lenses.items():
+                region = sublime.Region(*key)
+                phantom_region = self._get_phantom_region(region)
+                html = '\n<small style="font-family: system">|</small>\n'.join(lens.small_html for lens in group)
+                phantoms.append(sublime.Phantom(phantom_region, html, sublime.LAYOUT_BELOW))
+            self._phantom.update(phantoms)
 
     def get_resolved_code_lenses_for_region(self, region: sublime.Region) -> Generator[CodeLens, None, None]:
         region = self.view.line(region)
--- a/plugin/code_lens.py
+++ b/plugin/code_lens.py
@@ -134,16 +134,15 @@ class CodeLensView:
                 yield lens
 
     def _get_phantom_region(self, region: sublime.Region) -> sublime.Region:
-        line = self.view.substr(self.view.line(region))
+        line = self.view.line(region)
+        code = self.view.substr(line)
         offset = 0
-        for ch in line:
+        for ch in code:
             if ch.isspace():
                 offset += 1
             else:
                 break
-        row, _ = self.view.rowcol(region.a)
-        point = self.view.text_point(max(row - 1, 0), offset, clamp_column=True)
-        return sublime.Region(point)
+        return sublime.Region(line.a + offset, line.b)
 
     def render(self, mode: str) -> None:
         if mode == 'phantom':
--- a/plugin/code_lens.py
+++ b/plugin/code_lens.py
@@ -128,21 +128,12 @@ class CodeLensView:
         for group in self._code_lenses.values():
             yield from group
 
-    def unresolved_visible_code_lens(self, visible: sublime.Region) -> Iterable[CodeLensData]:
+    def unresolved_visible_code_lenses(self, visible: sublime.Region) -> Iterable[CodeLensData]:
         for lens in self._flat_iteration():
             if not lens.is_resolved() and visible.intersects(lens.region):
                 yield lens
 
-    def resolved_code_lens(self) -> Iterable[CodeLensData]:
-        for lens in self._flat_iteration():
-            if lens.is_resolved():
-                yield lens
-
     def _get_phantom_region(self, region: sublime.Region) -> sublime.Region:
-        """Takes the current region and puts the phantom on the line above, properly indented.
-
-        Code-lenses are typically shown above the code in VSCode.
-        """
         line = self.view.substr(self.view.line(region))
         offset = 0
         for ch in line:
@@ -150,17 +141,12 @@ class CodeLensView:
                 offset += 1
             else:
                 break
-
-        row, _ = self.view.rowcol_utf16(region.a)
-        point = self.view.text_point_utf16(max(row - 1, 0), offset, clamp_column=True)
+        row, _ = self.view.rowcol(region.a)
+        point = self.view.text_point(max(row - 1, 0), offset, clamp_column=True)
         return sublime.Region(point)
 
     def render(self, mode: str) -> None:
-        if mode == 'annotation':
-            accent = self.view.style_for_scope("region.greenish markup.accent.codelens.lsp")["foreground"]
-            for index, lens in enumerate(self._flat_iteration()):
-                self.view.add_regions(self._region_key(index), [lens.region], "", "", 0, [lens.small_html], accent)
-        else:  # 'phantom'
+        if mode == 'phantom':
             phantoms = []
             for key, group in self._code_lenses.items():
                 region = sublime.Region(*key)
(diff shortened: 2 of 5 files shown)


LSP.sublime-settings
```diff
@@ -97,6 +97,12 @@
   // using a keyboard shortcut or the context menu.
   "show_code_actions": "annotation",
 
+  // Show code lens:
+  // "annotation" - show an annotation on the right when code actions are available
+  // "phantom" - show a phantom on the top when code actions are available
+  // "" - don't show code lens
```
rwols, line 103: Contrary to code actions, code lenses are enabled only in one place. So they can be disabled already with

```js
"disabled_capabilities": {
    "codeLensProvider": true
}
```

The different settings for code actions are needed because they appear in the hover popup as well as in the view, as well as can be requested manually with a keybinding.

plugin/core/views.py
```diff
@@ -808,3 +808,46 @@ def format_completion(
     if item.get("textEdit"):
         completion.flags = sublime.COMPLETION_FLAG_KEEP_PREFIX
     return completion
+
+
+class CodeLensPhantom:
+    def __init__(self, view: sublime.View, region: sublime.Region):
```
rwols, line 814: Maybe use `__slots__`? It's a small class.

plugin/documents.py
```diff
@@ -545,13 +550,30 @@ def _render_code_lens(self, name: str, index: int, region: sublime.Region, comma
...
+            self.view.add_regions(self._code_lens_key(index), [region], "", "", 0, [annotation], accent)
+
+    def _show_code_lens_phantom(self) -> None:
+        phantoms = [phantom.to_phantom() for phantom in self._code_lens_phantom_regions.values() if phantom.is_visible()]
+        self._code_lens_phantom.update(phantoms)
 
     def _do_code_lenses_async(self) -> None:
         session = self.session_async("codeLensProvider")
-        if session and session.uses_plugin():
+        if session and session.uses_plugin() and userprefs().show_code_lens:
```
rwols, line 576: This additional check can go away, it's already taken care of by `"disabled_capabilities"`

plugin/documents.py
```diff
@@ -545,13 +550,30 @@ def _render_code_lens(self, name: str, index: int, region: sublime.Region, comma
...
-        self.view.add_regions(self._code_lens_key(index), [region], "", "", 0, [annotation], accent)
+
+        show_code_lens = userprefs().show_code_lens
+        if show_code_lens == "phantom":
+            key = region.to_tuple()
+            try:
+                phantom = self._code_lens_phantom_regions[key]
+            except KeyError:
+                self._code_lens_phantom_regions[key] = phantom = CodeLensPhantom(self.view, region)
+
```
rwols, line 561: Personally not a huge fan of empty lines. Only use empty lines when it's a function/method/closure please?

rwols (changes requested):

predragnikolic:
IMO, the phantoms looks nice! Really Nice! :+1: 

I did a quick test run with the LSP-elm server
and noticed a few issues, which lead me to believe that there is still some polishing to do in the UX.

Like for example:
1. if I insert new lines, the code lens will on the same line (than inserted back to where it should be)
2. if I delete a line before a code lens, the code lens will be deleted (than inserted back in) - I think that this is ST behavior and that we cannot prevent the deletion of phantoms.  
output

there is a problem where one code lens appears multiple times.

notice the `exposed | 4 references | 4 references`

Screenshot from 2021-06-15 17-22-45
notice the `exposed | 7 references | 7 references | 7 references | 7 references`

LSP.sublime-settings
Rapptz: Yeah that works too. I wrote the setting in a similar fashion since some people might prefer a quick toggle and I was already there. I'll remove this setting though if it's duplicating another.

Rapptz:
I've fixed the review comments.

> if I insert new lines, the code lens will on the same line (than inserted back to where it should be)

Ideally, the code lens would be re-calculated when there are modifications. I'd personally use `on_modified_async` for this with the 300ms delay to prevent spamming. I'm not sure if this would be the approach I should take though. There are other nits I want to fix like the one I mentioned in the starting comment above.

> if I delete a line before a code lens, the code lens will be deleted (than inserted back in) - I think that this is ST behavior and that we cannot prevent the deletion of phantoms.

I think this one's in the same vain as the one above, which is to re-calculate code lenses on modification. I'm not entirely sure if it fixes it but I'm willing to try.

> there is a problem where one code lens appears multiple times.

I've seen this one before myself! I don't know what causes it other than maybe the server sending duplicates (unlikely?) or some state not being cleared properly. I think I need LSP server logs to figure it out but I'd love to get it fixed as well.

plugin/documents.py
```diff
@@ -587,6 +611,14 @@ def _resolve_visible_code_lenses_async(self) -> None:
...
                 session.send_request_async(Request("codeLens/resolve", code_lens, self.view), callback)
 
+        # To prevent excessive jumping around, it would be desirable to only call
+        # phantom_set.update when most of the data is already resolved.
+        # Since this function is called whenever code lenses are actually rendered
+        # this function is a good place to put the actual rendering after some delay
+        # to wait for the code lenses to resolve.
+        if userprefs().show_code_lens == "phantom":
+            sublime.set_timeout_async(self._show_code_lens_phantom, FEATURES_TIMEOUT)
+
```
rwols, line 621: How about using `Promise.all` in plugin/core/promise.py to wait for all lenses in the viewport to resolve?

rwols: Would also prefer to have the same resolvement logic for both annotations and phantoms

rwols: So basically use `session.send_request_task` and collect them all in a list and use `Promise.all`. See textDocument/completion

plugin/core/views.py
```diff
@@ -808,3 +808,52 @@ def format_completion(
...
+
+        row, col = self.view.rowcol_utf16(self.region.a)
+        point = self.view.text_point_utf16(max(row - 1, 0), offset, clamp_column=True)
+        return sublime.Region(point)
+
+    def to_phantom(self) -> sublime.Phantom:
+        if len(self.lens) == 1:
+            html = self.lens[0]
+        else:
+            html = '\n<small style="font-family: system">|</small>\n'.join(lens for lens in self.lens)
```
rchl, line 853: Shouldn't this work?
```suggestion
            html = '\n<small style="font-family: system">|</small>\n'.join(self.lens)
```

plugin/core/views.py
```diff
@@ -808,3 +808,52 @@ def format_completion(
...
+
+        html = '<div class="codelens">{}</div>'.format(html)
```
rchl, line 855: Is the `class` used for anything?
I don't think it's possible to target it with some external overrides either? And if it is, it should be more specific to this package.

plugin/documents.py
```diff
@@ -545,9 +550,24 @@ def _render_code_lens(self, name: str, index: int, region: sublime.Region, comma
...
+        if show_code_lens == "phantom":
+            key = region.to_tuple()
+            try:
+                phantom = self._code_lens_phantom_regions[key]
+            except KeyError:
+                self._code_lens_phantom_regions[key] = phantom = CodeLensPhantom(self.view, region)
+            if annotation != "...":
+                phantom.add_annotation(annotation)
+        elif show_code_lens == "annotation":
+            annotation = '<div class="codelens">{}</div>'.format(annotation)
```
rchl, line 564: Is `class` used for anything? And if not, there is likely no need to wrap it in a `div` also.

Rapptz: This class was not mine, it was there from before I sent the PR. I do not know why it is there but I can remove it if it's unnecessary.

plugin/core/views.py
Rapptz: Yep. The original code wasn't a List[str] so I missed this during a refactoring.

plugin/documents.py
Rapptz: That's a good suggestion. I'll incorporate it in my next set of patches.

Rapptz:
I've force pushed the changesets to rebase from the previous refactoring and it's ready for review again.

plugin/code_lens.py
```diff
@@ -131,10 +133,36 @@ def resolved_code_lens(self) -> Iterable[CodeLensData]:
...
+        line = self.view.substr(self.view.line(region))
+        offset = 0
+        for ch in line:
+            if ch.isspace():
+                offset += 1
+            else:
+                break
+
+        row, _ = self.view.rowcol_utf16(region.a)
+        point = self.view.text_point_utf16(max(row - 1, 0), offset, clamp_column=True)
```
rwols, line 150: use `rowcol` and `text_point`, we're already in UTF32 space.

rwols (changes requested):

plugin/code_lens.py
```diff
@@ -82,6 +83,7 @@ def _region_key(self, index: int) -> str:
         return '{0}.{1}'.format(self.CODE_LENS_KEY, index)
 
     def clear_view(self) -> None:
+        self._phantom.update([])
         self.clear_annotations()
```
rchl, line 87: Maybe rename the method now since it can't be either annotation or phantom? Something more generic that doesn't mention the UI concept. Maybe just `clear()` since the class name `CodeLensView` already makes it clear that it's something UI related.

plugin/code_lens.py
```diff
@@ -131,10 +133,36 @@ def resolved_code_lens(self) -> Iterable[CodeLensData]:
...
+                break
+
+        row, _ = self.view.rowcol_utf16(region.a)
+        point = self.view.text_point_utf16(max(row - 1, 0), offset, clamp_column=True)
+        return sublime.Region(point)
+
+    def render(self, mode: str) -> None:
+        if mode == 'annotation':
+            accent = self.view.style_for_scope("region.greenish markup.accent.codelens.lsp")["foreground"]
+            for index, lens in enumerate(self._flat_iteration()):
```
rchl, line 156: Just a nit but that `_flat_iteration` method seems kinda redundant. It's an extra 3-lines method with an extra, redundant iteration loop when you can just use `self._code_lenses.values()` in place of it.

plugin/code_lens.py
```diff
@@ -131,10 +133,36 @@ def resolved_code_lens(self) -> Iterable[CodeLensData]:
...
+                self.view.add_regions(self._region_key(index), [lens.region], "", "", 0, [lens.small_html], accent)
+        else:  # 'phantom'
```
rchl, line 158: Since the `annotation` is the default option, it would be better to have it in the `else` branch so that we 
default to that when the setting is set to an unknown value.

plugin/code_lens.py
```diff
@@ -131,10 +133,36 @@ def resolved_code_lens(self) -> Iterable[CodeLensData]:
...
 
-    def render(self) -> None:
-        accent = self.view.style_for_scope("region.greenish markup.accent.codelens.lsp")["foreground"]
-        for index, lens in enumerate(self._flat_iteration()):
-            self.view.add_regions(self._region_key(index), [lens.region], "", "", 0, [lens.small_html], accent)
+    def _get_phantom_region(self, region: sublime.Region) -> sublime.Region:
+        """Takes the current region and puts the phantom on the line above, properly indented.
+
+        Code-lenses are typically shown above the code in VSCode.
+        """
```
rchl, line 140: Nits:
- this doesn't strictly put the phantom anywhere, just creates a region
- Not sure if the note about VSCode is necessary in that context.

rchl (commented):
There is unused `resolved_code_lens` method in `CodeLensView`.

plugin/code_lens.py
rchl: No, sorry, this is actually only about annotations.

plugin/code_lens.py
Rapptz: It's a helper method because I repeatedly call it. If flattening it in-line is better is considered better then that's fine. I'm not sure how it's redundant though. It's taking a nested structure and flattening it into a single iteration instead of two.

Rapptz: Note that this is equivalent to: 
```py
for group in self._code_lenses.values():
   for lens in group:
       ...
```
The helper method is using `yield from` not `yield`.

rchl: Right. I'm wrong here. Misunderstood the logic

rwols (changes requested):
The major problem I have with this is that phantoms are not attached to their actual region they represent. This causes all kinds of problems when modifying the buffer, for instance:

https://user-images.githubusercontent.com/2431823/123509879-63941200-d678-11eb-8fbe-d0ec4fd64078.mov

Use `sublime.LAYOUT_BELOW` and attach to the region it's meant to be attached to. We will have to wait for https://github.com/sublimehq/sublime_text/issues/4469, if ever implemented.

Rapptz:
This does not look pleasing.

image

I'm fine with changing it to the bottom but only if the region follows the same indentation rules as the code, i.e. the region is expanded to the line and indented, sort of like LAYOUT_BLOCK but not really.

TheSecEng:
This is starting to look really clean. Hopefully we get a `sublime.LAYOUT_ABOVE` option for phantoms here in the near future and then this can become user configurable

rwols (approved):
Phantoms:
Schermafbeelding 2021-07-03 om 17 13 52


Annotations:
Schermafbeelding 2021-07-03 om 17 15 20


Seems to work OK now.

Issue #1593, opened 2021-02-24
Inconsistent behavior when drawing overlapping styled regions


jwortmann:
This describes a bug discovered in #1585, when drawing multiple styled regions on top of each other, which is caused by a race condition. Sublime Text derives the order how overlapping regions are drawn (via the `view.add_regions` API method), from the time when the corresponding region key is created *the first time* in a particular view.

The following minimum example will illustrate this (for actual LSP examples see the screenshots/gif in #1585):

1. Create a new view with some text in it, and paste the following lines into the console:
```python
>>> view.add_regions("filled", [sublime.Region(20, 30)], scope="region.bluish", flags=sublime.DRAW_NO_OUTLINE)
>>> view.add_regions("squiggly", [sublime.Region(10, 40)], scope="region.redish", flags=sublime.DRAW_NO_FILL|sublime.DRAW_NO_OUTLINE|sublime.DRAW_SQUIGGLY_UNDERLINE)
```

![region2](https://user-images.githubusercontent.com/6579999/109009640-1181b680-76af-11eb-8dc2-71c048332163.png)

Here everything looks fine and the squiggly underline can be seen on top of the blue background.

2. Now create another new view with some text in it, and use the following commands instead:
```python
>>> view.add_regions("squiggly", [sublime.Region(0, 0)], scope="region.redish", flags=sublime.DRAW_NO_FILL|sublime.DRAW_NO_OUTLINE|sublime.DRAW_SQUIGGLY_UNDERLINE)
>>> view.erase_regions("squiggly")
>>> view.add_regions("filled", [sublime.Region(20, 30)], scope="region.bluish", flags=sublime.DRAW_NO_OUTLINE)
>>> view.add_regions("squiggly", [sublime.Region(10, 40)], scope="region.redish", flags=sublime.DRAW_NO_FILL|sublime.DRAW_NO_OUTLINE|sublime.DRAW_SQUIGGLY_UNDERLINE)
```

![region1](https://user-images.githubusercontent.com/6579999/109009699-22cac300-76af-11eb-8829-a4a14d578757.png)

Even thoug "squiggly" is drawn last here, it is now hidden underneath the "filled" region. This happens because the "squiggly" key has been created earlier in this view. It also shows that `view.erase_regions` will not reset the drawing order of region keys.

[...]

PR #1751, opened 2021-06-15
Semantic highlighting

litezzzout:
adding semantic highlighting using "add regions" not an ideal solution but works for time being. 

instruction for editing colorscheme, to work with this is in -- LSP/plugin/semantic.py

I'm not very familiar with LSP package or python, code probably needs little tune up.
I've tested with "clangd" it works, other server I tried was pyrigtht doesn't work with that, MS is keeping semantic highlighting for themselves in Pylance.
Edit: attached screenshot
semantic_highlighting

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -95,6 +95,42 @@
   // "" - don't show code actions.
   // Note: Due to API limitations, the "bulb" icon can not be clicked so the code actions can only be triggered
   // using a keyboard shortcut or the context menu.
+  
+
+  // Enable semantic highlighting:
+  
+  // Add the following to your color scheme
+  //  "rules":
+  //  [
+  //     {
+  //         "scope": "meta.semantic-token.lsp",
+  //         "background": "#01010101"
+  //     }
+  //  ]
+
+  // Note: most color schemes don't utilise many available SublimeText scopes,
+  // this will result in subpar semantic highlighting. Follow the example given
+  // below to customize your color_scheme.
+
+  // syntax for scopes: meta.semantic.tokenType.modifierType
+  // modifierTypes are optional.
+
+  // example rules:
+  // {
+  //     "scope": "meta.semantic.variable",
+  //     "foreground": "var(blue)",
+  // },
+  // {
+  //     "scope": "meta.semantic.method",
+  //     "foreground": "var(yellow)",
+  // },
+  // {
+  //     "scope": "meta.semantic.method.static",
+  //     "foreground": "var(red)",
+  // },
+  "enable_semantic_tokens": false,
+
+
   "show_code_actions": "annotation",
 
   // Show code lens:
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -272,6 +272,56 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
         },
         "codeLens": {
             "dynamicRegistration": True
+        },
+        "semanticTokens": {
+            "dynamicRegistration": True,
+            "tokenTypes": [
+                "namespace",
+                "type",
+                "class",
+                "enum",
+                "interface",
+                "struct",
+                "typeParameter",
+                "parameter",
+                "variable",
+                "property",
+                "enumMember",
+                "event",
+                "function",
+                "method",
+                "macro",
+                "keyword",
+                "modifier",
+                "comment",
+                "string",
+                "number",
+                "regexp",
+                "operator"
+            ],
+            "tokenModifiers": [
+                "declaration",
+                "definition",
+                "readonly",
+                "static",
+                "deprecated",
+                "abstract",
+                "async",
+                "modification",
+                "documentation",
+                "defaultLibrary"
+            ],
+            "formats": [
+                "relative"
+            ],
+            "requests": {
+                "range": False,
+                "full": {
+                    "delta": False
+                }
+            },
+            "multilineTokenSupport": True,
+            "overlappingTokenSupport": True
         }
     }
     workspace_capabilites = {
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -139,7 +139,6 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
     color_boxes_debounce_time = FEATURES_TIMEOUT
     highlights_debounce_time = FEATURES_TIMEOUT
     code_lenses_debounce_time = FEATURES_TIMEOUT
-    code_lenses_debounce_time = FEATURES_TIMEOUT + 2000
     semantic_tokens_debounce_time = FEATURES_TIMEOUT
 
     _uri = None  # type: str
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -695,19 +695,12 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         return scope2regions
 
     def _update_semantic_token_regions(self, scope2regions: Dict[str, List[sublime.Region]]) -> None:
-        keys = []
         for scope, regions in scope2regions.items():
             if not scope:
                 continue
             # in case of multiple modifiers, getting a _single modifier_ unique key
             # instead of a long _combined key_, to comply with init_region_keys()
             key = scope.split(', ')[0]
-            if key in keys:
-                for i in range(1, len(scope.split(', ')) - 1):
-                    if scope.split(', ')[i] not in keys:
-                        key = scope.split(', ')[i]
-                        break
-            keys.append(key)
             self.view.add_regions(key, regions, 'meta.semantic-token.lsp ' + scope, flags=sublime.DRAW_NO_OUTLINE)
 
     def _clear_semantic_token_regions(self) -> None:
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -38,6 +38,7 @@ from .core.views import text_document_position_params
 from .core.views import update_lsp_popup
 from .core.windows import AbstractViewListener
 from .core.windows import WindowManager
+from .semantic import get_semantic_scope_from_modifier
 from .session_buffer import SessionBuffer
 from .session_view import SessionView
 from functools import partial
@@ -49,7 +50,6 @@ import sublime_plugin
 import textwrap
 import weakref
 import webbrowser
-from .semantic import get_semantic_scope_from_modifier
 
 SUBLIME_WORD_MASK = 515
 
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -192,6 +192,7 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         self._color_phantoms.update([])
         self.view.erase_status(AbstractViewListener.TOTAL_ERRORS_AND_WARNINGS_STATUS_KEY)
         self._clear_highlight_regions()
+        self._clear_semantic_token_regions()
         self._clear_session_views_async()
 
     def _reset(self) -> None:
@@ -661,47 +662,57 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
                 Request.semanticTokens(params, self.view),
                 lambda response: self._on_semantic_tokens_async(response, semantic_tokens_legend['legend']))
 
-    def _on_semantic_tokens_async(self, response: dict, semantic_tokens_legend: dict) -> None:
-        regions = cast(dict, {})
+    def _on_semantic_tokens_async(self, response: dict, legend: dict) -> None:
+        scope2regions = self._process_semantic_tokens_response(response, legend)
+        self._update_semantic_token_regions(scope2regions)
+
+    def _process_semantic_tokens_response(
+        self,
+        response: Dict[str, List[int]],
+        legend: dict
+    ) -> Dict[str, List[sublime.Region]]:
+        scope2regions = {}  # type: Dict[str, List[sublime.Region]]
         data = response['data']
-        prev_row = None
-        prev_col = None
-        for x in range(0, len(data), 5):
-            encoded_token = data[x:x+5]
+        row = 0
+        col = 0
+        for i in range(0, len(data), 5):
+            token = data[i:i+5]
 
-            if prev_row is not None:
-                if encoded_token[0] == 0:
-                    encoded_token[1] += prev_col
-                    encoded_token[0] = prev_row
-                else:
-                    encoded_token[0] += prev_row
+            if token[0] == 0:
+                token[1] += col
+                token[0] = row
+            else:
+                token[0] += row
 
-            point1 = self.view.text_point(encoded_token[0], encoded_token[1])
-            point2 = self.view.text_point(encoded_token[0], encoded_token[1]+encoded_token[2])
-            my_region = sublime.Region(point1, point2)
+            scope = get_semantic_scope_from_modifier(token, legend)
+            region = sublime.Region(
+                self.view.text_point_utf16(token[0], token[1]),
+                self.view.text_point_utf16(token[0], token[1] + token[2]))
+            scope2regions.setdefault(scope, []).append(region)
 
-            scope = get_semantic_scope_from_modifier(encoded_token, semantic_tokens_legend)
+            row = token[0]
+            col = token[1]
+        return scope2regions
 
-            try:
-                regions[scope].append(my_region)
-            except KeyError:
-                regions[scope] = []
-                regions[scope].append(my_region)
+    def _update_semantic_token_regions(self, scope2regions: Dict[str, List[sublime.Region]]) -> None:
+        keys = []
+        for scope, regions in scope2regions.items():
+            if not scope:
+                continue
+            # in case of multiple modifiers, getting a _single modifier_ unique key
+            # instead of a long _combined key_, to comply with init_region_keys()
+            key = scope.split(', ')[0]
+            if key in keys:
+                for i in range(1, len(scope.split(', ')) - 1):
+                    if scope.split(', ')[i] not in keys:
+                        key = scope.split(', ')[i]
+                        break
+            keys.append(key)
+            self.view.add_regions(key, regions, 'meta.semantic-token.lsp ' + scope, flags=sublime.DRAW_NO_OUTLINE)
 
-            prev_row = encoded_token[0]
-            prev_col = encoded_token[1]
-
-        region_keys = []
-        for key, value in regions.items():
-            if key:
-                reg_key = key.split(', ')[0]  # in case of multiple modifiers, getting a _single modifier_ unique key
-                if reg_key in region_keys:    # instead of a long _combined key_, to comply with init_region_keys()
-                    for i in range(1, len(key.split(', ')) - 1):
-                        if key.split(', ')[i] not in region_keys:
-                            reg_key = key.split(', ')[i]
-                            break
-                region_keys.append(reg_key)
-                self.view.add_regions(reg_key, value, 'meta.semantic-token.lsp ' + key, flags=sublime.DRAW_NO_OUTLINE)
+    def _clear_semantic_token_regions(self) -> None:
+        # TODO
+        pass
 
     # --- textDocument/complete ----------------------------------------------------------------------------------------
 
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -21,7 +21,7 @@ from .core.types import basescope2languageid
 from .core.types import debounced
 from .core.types import FEATURES_TIMEOUT
 from .core.types import SettingsRegistration
-from .core.typing import Any, Callable, Optional, Dict, Generator, Iterable, List, Tuple, Union
+from .core.typing import Any, Callable, Optional, Dict, Generator, Iterable, List, Tuple, Union, cast
 from .core.url import parse_uri
 from .core.url import view_to_uri
 from .core.views import DIAGNOSTIC_SEVERITY
@@ -33,6 +33,7 @@ from .core.views import lsp_color_to_phantom
 from .core.views import make_command_link
 from .core.views import range_to_region
 from .core.views import show_lsp_popup
+from .core.views import text_document_identifier
 from .core.views import text_document_position_params
 from .core.views import update_lsp_popup
 from .core.windows import AbstractViewListener
@@ -48,7 +49,7 @@ import sublime_plugin
 import textwrap
 import weakref
 import webbrowser
-
+from .semantic import get_semantic_scope_from_modifier
 
 SUBLIME_WORD_MASK = 515
 
@@ -138,6 +139,8 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
     color_boxes_debounce_time = FEATURES_TIMEOUT
     highlights_debounce_time = FEATURES_TIMEOUT
     code_lenses_debounce_time = FEATURES_TIMEOUT
+    code_lenses_debounce_time = FEATURES_TIMEOUT + 2000
+    semantic_tokens_debounce_time = FEATURES_TIMEOUT
 
     _uri = None  # type: str
 
@@ -227,6 +230,7 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         if added:
             self._do_color_boxes_async()
             self._do_code_lenses_async()
+            self._do_semantic_tokens_async()
 
     def on_session_shutdown_async(self, session: Session) -> None:
         removed_session = self._session_views.pop(session.config.name, None)
@@ -340,6 +344,9 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         self.do_signature_help_async(manual=False)
         self._when_selection_remains_stable_async(self._do_code_lenses_async, current_region,
                                                   after_ms=self.code_lenses_debounce_time)
+
+        self._when_selection_remains_stable_async(self._do_semantic_tokens_async, current_region,
+                                                  after_ms=self.semantic_tokens_debounce_time)
 
     def get_uri(self) -> str:
         return self._uri
@@ -634,6 +641,68 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
                 self.view.add_regions(key, regions, scope=_kind2scope[kind], flags=flags)
 
         sublime.set_timeout(render_highlights_on_main_thread)
+
+    # --- textDocument/semanticTokens -------------------------------------------------------------------------------
+
+    def _do_semantic_tokens_async(self) -> None:
+        if ("background" not in self.view.style_for_scope("meta.semantic-token.lsp") or
+                not userprefs().enable_semantic_tokens):
+            return
+        session = self.session_async("semanticTokensProvider")
+        if session:
+            semantic_tokens_legend = {}
+            try:
+                semantic_tokens_legend = cast(dict, session.get_capability('semanticTokensProvider'))
+            except ValueError:
+                # print('semantic highlighting capabilies not provided by the server!!!')
+                return
+
+            params = {"textDocument": text_document_identifier(self.view)}
+            session.send_request_async(
+                Request.semanticTokens(params, self.view),
+                lambda response: self._on_semantic_tokens_async(response, semantic_tokens_legend['legend']))
+
+    def _on_semantic_tokens_async(self, response: dict, semantic_tokens_legend: dict) -> None:
+        regions = cast(dict, {})
+        data = response['data']
+        prev_row = None
+        prev_col = None
+        for x in range(0, len(data), 5):
+            encoded_token = data[x:x+5]
+
+            if prev_row is not None:
+                if encoded_token[0] == 0:
+                    encoded_token[1] += prev_col
+                    encoded_token[0] = prev_row
+                else:
+                    encoded_token[0] += prev_row
+
+            point1 = self.view.text_point(encoded_token[0], encoded_token[1])
+            point2 = self.view.text_point(encoded_token[0], encoded_token[1]+encoded_token[2])
+            my_region = sublime.Region(point1, point2)
+
+            scope = get_semantic_scope_from_modifier(encoded_token, semantic_tokens_legend)
+
+            try:
+                regions[scope].append(my_region)
+            except KeyError:
+                regions[scope] = []
+                regions[scope].append(my_region)
+
+            prev_row = encoded_token[0]
+            prev_col = encoded_token[1]
+
+        region_keys = []
+        for key, value in regions.items():
+            if key:
+                reg_key = key.split(', ')[0]  # in case of multiple modifiers, getting a _single modifier_ unique key
+                if reg_key in region_keys:    # instead of a long _combined key_, to comply with init_region_keys()
+                    for i in range(1, len(key.split(', ')) - 1):
+                        if key.split(', ')[i] not in region_keys:
+                            reg_key = key.split(', ')[i]
+                            break
+                region_keys.append(reg_key)
+                self.view.add_regions(reg_key, value, 'meta.semantic-token.lsp ' + key, flags=sublime.DRAW_NO_OUTLINE)
 
     # --- textDocument/complete ----------------------------------------------------------------------------------------
 
--- a/plugin/semantic.py
+++ b/plugin/semantic.py
@@ -1,0 +1,104 @@
+NON_MODIFIER_SCOPES = {
+                'variable': 'variable.other.lsp',
+                'parameter': 'variable.parameter.lsp',
+                'function': 'variable.function.lsp',
+                'method': 'variable.function.lsp',
+                'property': 'variable.other.member.lsp',
+                'class': 'support.type.lsp',
+                'enum': 'variable.enum.lsp',
+                'enumMember': 'constant.other.enum.lsp',
+                'type': 'storage.type.lsp',
+                'macro': 'variable.other.constant.lsp',
+                'namespace': 'variable.other.namespace.lsp',
+                'typeParameter': 'variable.parameter.generic.lsp',
+                'comment': 'comment.block.documentation.lsp',
+                'dependent': '',
+                'concept': '',
+                'module': '',
+                'magicFunction': '',
+                'selfParameter': '',
+
+            }
+
+DECLARATION_SCOPES = {
+            'namespace': 'entity.name.namespace.lsp',
+            'type': 'entity.name.type.lsp',
+            'class': 'entity.name.class.lsp',
+            'enum': 'entity.name.enum.lsp',
+            'interface': 'entity.name.interface.lsp',
+            'struct': 'entity.name.struct.lsp',
+            'function': 'entity.name.function.lsp',
+            'method': 'entity.name.function.lsp',
+            'macro': 'entity.name.macro.lsp'
+            }
+
+STAIC_SCOPES = NON_MODIFIER_SCOPES.copy()
+DEPRECATED_SCOPES = NON_MODIFIER_SCOPES.copy()
+ABSTRACT_SCOPES = NON_MODIFIER_SCOPES.copy()
+ASYNC_SCOPES = NON_MODIFIER_SCOPES.copy()
+MODIFICATION_SCOPES = NON_MODIFIER_SCOPES.copy()
+DEFAULT_LIB_SCOPES = NON_MODIFIER_SCOPES.copy()
+
+[NON_MODIFIER_SCOPES.update({k: v + ' meta.semantic.' + k + '.lsp'}) for k, v in NON_MODIFIER_SCOPES.items()]
+
+[DECLARATION_SCOPES.update({k: v + ' meta.semantic.' + k + '.declaration.lsp'}) for k, v in DECLARATION_SCOPES.items()]
+[STAIC_SCOPES.update({k: v + ' meta.semantic.' + k + '.satic.lsp'}) for k, v in STAIC_SCOPES.items()]
+[DEPRECATED_SCOPES.update({k: v + ' meta.semantic.' + k + '.deprecated.lsp'}) for k, v in DEPRECATED_SCOPES.items()]
+[ABSTRACT_SCOPES.update({k: v + ' meta.semantic.' + k + '.abstract.lsp'}) for k, v in ABSTRACT_SCOPES.items()]
+[ASYNC_SCOPES.update({k: v + ' meta.semantic.' + k + '.async.lsp'}) for k, v in ASYNC_SCOPES.items()]
+[MODIFICATION_SCOPES.update({k: v + ' meta.semantic.' + k + '.modification.lsp'})
+ for k, v in MODIFICATION_SCOPES.items()]
+[DEFAULT_LIB_SCOPES.update({k: v + ' meta.semantic.' + k + '.defaultLibrary.lsp'})
+ for k, v in DEFAULT_LIB_SCOPES.items()]
+
+SEMANTIC_SCOPES = {
+            '': NON_MODIFIER_SCOPES,  # if no modifiers are provided
+            "declaration": DECLARATION_SCOPES,
+            "definition": DECLARATION_SCOPES,
+            "readonly": {'variable': 'constant.other.lsp, semantic.variable.readonly.lsp'},
+            "static": STAIC_SCOPES,  # these are temporary, should be filled with real scopes
+            "deprecated": DEPRECATED_SCOPES,
+            "abstract": ABSTRACT_SCOPES,
+            "async": ASYNC_SCOPES,
+            "modification": MODIFICATION_SCOPES,
+            "documentation": {'comment': 'comment.block.documentation'},
+            "defaultLibrary": DEFAULT_LIB_SCOPES
+        }
+
+
+def get_semantic_scope_from_modifier(encoded_token: list, semantic_tokens_legends: dict) -> str:
+    token_int = encoded_token[3]
+    modifier_int = encoded_token[4]
+
+    token_type = semantic_tokens_legends['tokenTypes'][token_int]
+
+    if not modifier_int:
+        try:
+            return(SEMANTIC_SCOPES[''][token_type])
+        except KeyError:
+            pass  # print('scope not defined for tokenType: ', token_type)
+
+    modifier_binary = [int(x) for x in reversed(bin(modifier_int)[2:])]
+    modifiers = [semantic_tokens_legends['tokenModifiers'][i] for i in range(len(modifier_binary)) if i]
+
+    # print('tokenType: ' + token_type + '-- modifiers: ',modifiers)
+    scopes = []
+    for modifier in modifiers:
+        scope = None
+        try:
+            scope = SEMANTIC_SCOPES[modifier][token_type]
+        except KeyError:
+            pass  # print('scope not defined for modifier/tokenType: ', modifier+'/'+token_type)
+
+        if scope and scope not in scopes:
+            scopes.append(scope)
+
+    if not scopes:
+        try:
+            return(SEMANTIC_SCOPES[''][token_type])
+        except KeyError:
+            pass  # print('scope not defined for tokenType: ', token_type)
+    else:
+        return ', '.join(scopes)
+
+    return ''
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -305,6 +305,18 @@ class Request:
             "params": self.params
         }
 
+    @classmethod
+    def semanticTokens(cls, params: Mapping[str, Any], view: sublime.View) -> 'Request':
+        return Request("textDocument/semanticTokens/full", params, view)
+
+    @classmethod
+    def semanticTokensDelta(cls, params: Mapping[str, Any], view: sublime.View) -> 'Request':
+        return Request("textDocument/semanticTokens/full/delta", params, view)
+
+    @classmethod
+    def semanticTokensRange(cls, params: Mapping[str, Any], view: sublime.View) -> 'Request':
+        return Request("textDocument/semanticTokens/range", params, view)
+
 
 class ErrorCode:
     # Defined by JSON RPC
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -158,6 +158,7 @@ class Settings:
     diagnostics_panel_include_severity_level = None  # type: int
     disabled_capabilities = None  # type: List[str]
     document_highlight_style = None  # type: str
+    enable_semantic_tokens = None  # type: bool
     inhibit_snippet_completions = None  # type: bool
     inhibit_word_completions = None  # type: bool
     log_debug = None  # type: bool
@@ -196,6 +197,7 @@ class Settings:
         r("diagnostics_panel_include_severity_level", 4)
         r("disabled_capabilities", [])
         r("document_highlight_style", "underline")
+        r("enable_semantic_tokens", False)
         r("log_debug", False)
         r("log_max_size", 8 * 1024)
         r("lsp_code_actions_on_save", {})
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -18,6 +18,7 @@ from weakref import ref
 from weakref import WeakValueDictionary
 import functools
 import sublime
+from .semantic import SEMANTIC_SCOPES
 
 DIAGNOSTIC_TAG_VALUES = [v for (k, v) in DiagnosticTag.__dict__.items() if not k.startswith('_')]
 
@@ -37,6 +38,7 @@ class SessionView:
     _session_buffers = WeakValueDictionary()  # type: WeakValueDictionary[Tuple[int, int], SessionBuffer]
 
     def __init__(self, listener: AbstractViewListener, session: Session, uri: DocumentUri) -> None:
+        self.init_region_keys(listener.view, session.config.name, userprefs().document_highlight_style)
         self.view = listener.view
         self.session = session
         self.active_requests = {}  # type: Dict[int, Request]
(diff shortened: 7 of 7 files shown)


jwortmann:
I'm a bit pessimistic about using this `add_regions` workaround, because
* it relies on the undocumented (and possibly unintended) behavior that it is possible to change the foreground color with `add_regions` under certain circumstances
* it requires adjustment(s) for each color scheme
* the ST developers already stated that plugins should not modify the foreground colors from syntax highlighting (due to latency etc.) - maybe it's acceptable if only the existing color from the color scheme are used

But (after some discussion) we already make use of this for the `unnecessary` and `deprecated` tags for diagnostics. So *if* support for semantic highlighting should really be implemented here even without a dedicated ST API that can be used, and if this PR should be considered to merge, than I would at least suggest the following improvements:

1. It should be checked first, if the corresponding color scheme rule(s) are present, and only in that case send the request and draw regions - similar to how it is done with the diagnostic tags at the moment

2. Instead of those arbitrary new scope names like `semantic_param` etc., we should reuse established scopes and map them to the LSP semantic token types, e.g. `variable.parameter`. Then we can prepend the scope with something like `meta.semantic-token`, i.e. the full scope to use for `add_regions` would be `meta.semantic-token.lsp variable.parameter.lsp`. Then the color scheme needs only a single rule like the following, and it should work for all of the regions:
    ```json
    {
        "scope": "meta.semantic-token.lsp",
        "background": "#00000001"
    },
    ```

rwols:
Also: would this hide the diagnostics? Or are the diagnostics still visible?

litezzzout:
> Also: would this hide the diagnostics? Or are the diagnostics still visible?
> 
> EDIT: I see @jwortmann already noted this :)

I haven't rigorously tested it but squiggly error/warning works but outline doesn't as you can see in screen shot.
btw what's the default scope for a variable name, seems like ST doesn't even try to color variable names, at least in C/Cpp. can't find it by ctrl+shift+alt+p.

rwols:
Scopes are explained here https://www.sublimetext.com/docs/scope_naming.html

there should be a mapping from LSP semantic tokens + LSP token modifiers -> ST scopes

rwols:
Look at the symbol table in plugin/core/views.py for an example of another mapping (from LSP symbol names to ST scope names)

litezzzout:
> lugin/core/views.p

> The first transformation step is to encode tokenType and tokenModifiers as integers using the legend. As said, token types are looked up by index, so a tokenType value of 1 means tokenTypes[1]. Multiple token modifiers can be set by using bit flags, so a tokenModifier value of 3 is first viewed as binary 0b00000011, which means [tokenModifiers[0], tokenModifiers[1]] because bits 0 and 1 are set. Using this legend, the tokens now are: -- MS spec 

is this what you referring to? will map to SYMBOL_KINDS

jwortmann:
The appropriate scope for variable names as used in a majority of syntaxes would be `variable.other`.

The problem for cooperating with the diagnostics (and document highlights) is, that the behavior in ST, which of the regions to draw on top, is dependent on which region *key* is created first in a view. This can depend on the order of certain user operations, or the time when diagnostics are reported by a server. So even if it might work in your tests, there can be other circumstances where it doesn't work correctly. For a more detailed explanation see #1593. So with the help of something like a mentioned "region manager" I think we would need to create (empty) regions in the correct order and in each new view first, even if no diagnostics/documentHighlights/semantic tokens are reported yet. And only after that "initialization", draw the actual regions when applicable.

litezzzout:
> The appropriate scope for variable names as used in a majority of syntaxes would be `variable.other`.

default theme that comes with st4 doesn't seem to have this value set to anything, all I could find resembling to this was

`        {
            "name": "JavaScript Dollar",
            "scope": "variable.other.dollar.only.js, variable.other.object.dollar.only.js, variable.type.dollar.only.js, support.class.dollar.only.js",
            "foreground": "var(blue5)"
        },`
setting scope to 'variable.other' didn't change color for me

jwortmann:
> default theme that comes with st4 doesn't seem to have this value set to anything

Yes only the "Sixteen" color scheme from the default schemes target this scope (via `variable` selector). So it is expected that the variables use the default foreground color in most schemes.

litezzzout:
> > default theme that comes with st4 doesn't seem to have this value set to anything
> 
> Yes only the "Sixteen" color scheme from the default schemes target this scope (via `variable` selector). So it is expected that the variables use the default foreground color in most schemes.

thanks for blinding me btw.

litezzzout:
sublime_marina_semantic

default theme loves orange, uses for everything, variable, parameter, class/struct name different shade of orange for operator symbols, if we rely on default scopes and color themes to do the right thing, end user probably won't receive any benefits from semantic highlighting.

another approach could be like what vscode does. prompt user to customize their color scheme 
like this 
    "editor.semanticTokenColorCustomizations": {
        "rules": {
            "variable": "#94dbfd",
            "property": "#DADADA",
            "parameter": "#C586C0",
            "class": "#4eaab0",
            "macro": "#beb7ff",
            "function": "#dbdbaa",
            "type": "#4ec9b0",
            "enum": "#4eaacf",
            "enumMember": "#DADADA",
            "operator": "#C586C0"
        },
        "[Default Dark+]": {}
    },

jwortmann:
To customize the color scheme to adjust colors seperately for each token is still possible in ST. For example by using the selector `meta.semantic-token variable.parameter` in this case.

I can see why all this orange from your screenshot happens in Mariana. In the function body, now all variables which are function parameters get the "parameter" token, i.e. it is expected that they all are highlighted with the same color as within the parentheses for the argument list. But I would think it must be the same in VS Code, just that most VS Code themes perhaps use a more subtle color for function parameters instead of this prominent orange. Sublime color schemes have always worked this way to highlight parameters in the function declaration head (`variable.parameter`) with a special color. But I agree that it doesn't look nice here. Maybe we need a setting to ignore certain token kinds? Like `"ignored_semantic_tokens": ["parameter"]`, and then the user can fine-tune which of the tokens not to highlight, so that it looks good for their color scheme? (But actually, this would give no further means than what is already possible with a color scheme customization, it would just be easier to discover/apply for end-users.)

But I would expect class/struct names to use a different color than orange. Which exact scope name did you use for those?

rwols:
I'm guessing or hoping that a server would use the `declaration` or `definition` [SemanticTokenModifier](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#semanticTokenModifiers) when a class/record/struct is declared/defined, and not have that token modifier when it is _used_. With that I think you'd be able to align the scopes better to what the Scope Naming Guide demands.

For instance a **call** to a **class constructor** should probably use `variable.function.constructor`. But a **declaration** of a **class** should use `entity.name.class`

rwols:
What I'm also noticing from the screenshot is that **function declarations** and **function calls** are colored the same -- that should not happen in ST land. Same issue I guess.

litezzzout:
Capture

It doesn't color variable orange btw, I probably added some scope for testing and forgot to remove it.
here's are the scope I'm using

`
        self.view.add_regions('1',     my_macros,  'meta.semantic-token.lsp variable.other.constant.lsp',flags= sublime.DRAW_NO_OUTLINE)
        self.view.add_regions('2',       my_vars,  'meta.semantic-token.lsp variable.other.lsp',         flags= sublime.DRAW_NO_OUTLINE)
        self.view.add_regions('4',     my_params,  'meta.semantic-token.lsp variable.parameter.lsp',     flags= sublime.DRAW_NO_OUTLINE)
        self.view.add_regions('5',    my_structs,  'meta.semantic-token.lsp entity.name.class.lsp',      flags= sublime.DRAW_NO_OUTLINE)
        self.view.add_regions('6',      my_types,  'meta.semantic-token.lsp storage.type.lsp',           flags= sublime.DRAW_NO_OUTLINE)
        self.view.add_regions('7',      my_funcs,  'meta.semantic-token.lsp entity.name.function.lsp',   flags= sublime.DRAW_NO_OUTLINE)
        self.view.add_regions('8',      my_enums,  'meta.semantic-token.lsp variable.enum.lsp',          flags= sublime.DRAW_NO_OUTLINE)
        self.view.add_regions('9',     my_fields,  'meta.semantic-token.lsp variable.other.member.lsp',  flags= sublime.DRAW_NO_OUTLINE)
        self.view.add_regions('0', my_enum_fields, 'meta.semantic-token.lsp entity.name.enum.lsp',       flags= sublime.DRAW_NO_OUTLINE)`

I 've hard coded them, I couldn't get what rwols suggested  to work, probably I'm doing something wrong.

token_types = session.get_capability('semanticTokensProvider')
this return the following.
`{'range': False, 'full': {'delta': True}, 'legend': {'tokenTypes': ['variable', 'variable', 'parameter', 'function', 'method', 'function', 'property', 'variable', 'class', 'enum', 'enumMember', 'type', 'dependent', 'dependent', 'namespace', 'typeParameter', 'concept', 'type', 'macro', 'comment'], 'tokenModifiers': []}}`

and this prints
`if token_types[encoded_token[3]] ==  'function':
    print('Scope: ',get_symbol_scope_from_lsp_kind(encoded_token[3]+encoded_token[4]))`

it prints: 'entity.name.namespace' , I would expect it to print something like 'entity.name.function'

jwortmann:
> I'm guessing or hoping that a server would use the `declaration` or `definition` SemanticTokenModifier ...

Interesting, yes I think we should take the modifiers into account to improve the highlighting. As far as I understand the specs, each token type can have any of the modifiers and even multiple modifiers at the same time. This would make the mapping more complex, but we probably only need to consider a few of those possible combinations. And apparently servers can also define their own types and modifiers.

> For instance a call to a class constructor should probably use variable.function.constructor. But a declaration of a class should use entity.name.class

Hm, depends on which token type would be used for the class constructor, I would probably guess it is `function`, but could also be `class`, not sure. Seems like there is no modifier for "calls", and using `variable.function` by default in case it is the `class` type may result in unexpected highlighting, especially if servers don't use any modifiers at all. Without having tried the semantic highlighting with any server, I would probably assume something like the following tokens and modifiers in this Java example:
```java
public class SomeClass {
//           ^^^^^^^^^ TokenType: "class", TokenModifier: "declaration"/"definition"
    AnotherClass foo;
//  ^^^^^^^^^^^^ TokenType: "class"
    public SomeClass() { // this is the constructor
//         ^^^^^^^^^ TokenType: "function"/"method" or is this maybe also "class"?
        foo = new AnotherClass();
//                ^^^^^^^^^^^^ TokenType: "class" or "function"/"method" (?)
    }
}
```

So even for this small example I find it pretty hard to tell which tokens to expect and how to distinguish them in a good way for mapping to ST scopes. I guess this needs a bit of testing what kind of tokens/modifiers are actually used by servers. And even the ST scopes have inconsistencies between syntaxes, for example types/classes sometimes use `storage.type` and sometimes `support.type`/`support.class`.

But from the token and modifier lists in the LSP specs, I could imagine a mapping to ST scopes like this (probably needs more work):

| TokenType | TokenModifiers | ST scope |
| --- | --- | --- |
| namespace | declaration, definition | entity.name.namespace |
| namespace | any but above | variable.other.namespace (?) |
| type | declaration, definition | entity.name.type |
| type | any but above | storage.type or support.type (?) |
| class | declaration, definition | entity.name.class |
| class | any but above | storage.type.class or support.class (?) |
| enum | declaration, definition | entity.name.enum |
| enum | any but above | ? |
| interface | declaration, definition | entity.name.interface |
| interface | any but above | entity.other.inherited-class (?) |
| struct | declaration, definition | entity.name.struct |
| struct | any but above | storage.type.struct or support.struct (?) |
| typeParameter | any | variable.parameter.generic (?) |
| parameter | any | variable.parameter or even variable.other (see problem from screenshot above) |
| variable | readonly | constant.other or variable.other.constant (?) |
| variable | any but above | variable.other |
| property | any | ? (support.type.property-name if this means properties like in CSS) |
| enumMember | any | constant.other.enum |
| event | any | ? what is this supposed to mean ? |
| function | declaration, definition | entity.name.function |
| function | any but above | variable.function |
| method | declaration, defintion | entity.name.function |
| method | any but above | variable.function |
| macro | declaration, defintion | entity.name.macro |
| macro | any but above | variable.macro |
| keyword | any | keyword |
| modifier | any | storage.modifier |
| comment | documentation | comment.block.documentation |
| comment | any but above | comment |
| string | any | string |
| number | any | constant.numeric |
| regexp | any | string.regexp |
| operator | any | keyword.operator |

Lots of question marks....

***

@litezzzout This looks like you didn't decode the integer values to token kinds correctly. According to the specs, servers define their own [SemanticTokensLegend](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#semanticTokensLegend), so are you sure you are using this server's token list and correctly map the integer values to its token kinds? P.S. please don't blind me in the night with huge screenshots from "Sixteen" 😄

litezzzout:
I didn't use modifiers value for anything since I like to have same color for both delc and use, and forgot to add it before making a pull request. I also didn't do a lot of due diligence to comply with other color scheme since I explicitly defined scopes and color values for tokens I wanted.

litezzzout:
> According to the specs, servers define their own [SemanticTokensLegend](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#semanticTokensLegend)

`{'range': False, 'full': {'delta': True}, 'legend': {'tokenTypes': ['variable', 'variable', 'parameter', 'function', 'method', 'function', 'property', 'variable', 'class', 'enum', 'enumMember', 'type', 'dependent', 'dependent', 'namespace', 'typeParameter', 'concept', 'type', 'macro', 'comment'], 'tokenModifiers': []}}`

isn't this what they're referring to ?
we get token type with this  `session.get_capability('semanticTokensProvider')['legend']['tokenTypes']`
or am I wrong?

jwortmann:
> isn't this what they're referring to ?

Yes, for example here a value `4` would correspond to `method` (they seem to use 0-based indexing), but for another server it most likely is a different token type. This server seems to not use any modifiers at all.

litezzzout:
> or I don't really understand what you're doing with this).

rwols suggested this 

> Scopes are explained here https://www.sublimetext.com/docs/scope_naming.html
>there should be a mapping from LSP semantic tokens + LSP token modifiers -> ST scopes
> Look at the symbol table in plugin/core/views.py for an example of another mapping (from LSP symbol names to ST scope names)


I thought I could just directly map semantic tokens to LSP kinds, I probably misunderstood his point.

jfcherng:
I can confirm that pyright has no sematic highlighting while pylance does.

After https://github.com/litezzzout/LSP/pull/1/files, Pylance:
image

litezzzout:
> I can confirm that pyright has no sematic highlighting while pylance does.
> 
> After https://github.com/litezzzout/LSP/pull/1/files, Pylance:
> image

how to install pylance?

jfcherng:
> > I can confirm that pyright has no sematic highlighting while pylance does.
> > After [litezzzout/LSP#1 (files)](https://github.com/litezzzout/LSP/pull/1/files), Pylance:
> > image
> 
> how to install pylance?

I do that for myself. Not legal though, so not public.

litezzzout:
> I do that myself. Not legal though, so not public.

Oh ok. I though I was missing out on something, everyone knew about.

litezzzout:
@jfcherng btw did you get colors to work correctly?

jfcherng:
> @jfcherng btw did you get colors to work correctly?

I guess yes. I just forget to add `background1` in variables.
<details>

image

</details>

plugin/semantic.py
```diff
@@ -0,0 +1,214 @@
...
+        '''
+
+import sublime_plugin
+import time
+
+class SemanticListener(sublime_plugin.EventListener, sublime_plugin.TextChangeListener):  	
+    def on_post_save_async(self, view):
+        view.run_command('lsp_semantic')
+
+    def on_activated_async(self, view): # this needs work, doesn't always highlight on the first try when activating a view
```
jfcherng, line 207: maybe `on_load_async()` ?

jfcherng:
Noticed a little problem. If diagnostic red underline may be replaced with the sematic highlight (due to `add_region()`).
image

litezzzout:
> Noticed a small problem. A diagnostic error (red underline) may be replaced with the sematic highlight (due to `add_region()`).
Capture

because of this I was under the impression that, squiggly error/warning works, but just now realized it's only drawing on ''non add_region" characters

litezzzout:
> Noticed a small problem. A diagnostic error (red underline) may be replaced with the sematic highlight (due to `add_region()`).
diagnostics_works

this should work now

litezzzout:
> * It should be checked first, if the corresponding color scheme rule(s) are present, and only in that case send the request and draw regions - similar to how it is done with the diagnostic tags at the moment
> * Instead of those arbitrary new scope names like `semantic_param` etc., we should reuse established scopes and map them to the LSP semantic token types, e.g. `variable.parameter`. Then we can prepend the scope with something like `meta.semantic-token`, i.e. the full scope to use for `add_regions` would be `meta.semantic-token.lsp variable.parameter.lsp`. Then the color scheme needs only a single rule like the following, and it should work for all of the regions:
>   ```json
>   {
>       "scope": "meta.semantic-token.lsp",
>       "background": "#00000001"
>   },
>   ```
> * Also it should be tested carefully to ensure that it works together with the diagnostics and document highlights, and I believe currently it would be pretty buggy, because the regions are drawn in a non-deterministic order by LSP at the moment (I made another issue about this before). I think to solve this we would need some kind of central region manager as rhcl suggested in another PR (or issue) recently.

I have tried to improve the code all these areas, 
1. was trivial added a check.
2. Maped symanticTokens to ST scopes, used the table you provided as a guide, it could possibly be filled with many more scopes, someone who's expert in ST scopes should  look into it.
3. For now fixed diagnostic/highlighting drawing with a dirty hack. needs to be tested more to see what I have broke.

litezzzout:
added custom scopes that directcly correspond to tokenTypes/modifierTypes:
syntax:- `scope: semantic.tokenType.modifierType`

examples:
`semantic.method.static`
`semantic.variable`
`semantic.function.declaration`

this will result in simple straight-forward and more robust ColorScheme customization ability.

plugin/semantic.py
```diff
@@ -0,0 +1,232 @@
...
+        Color Scheme :
+        Error: Scope not defined in ColorSchme
+        
+        Add the following to your color scheme
+        
+        "rules":
+        [
+            {
+                "scope": "meta.semantic-token.lsp",
+                "background": "#00000001"
```
jfcherng, line 138: I think `#01010101` will cause less problem. This doesn't work for me as my color scheme background is perfect black (i.e., `#000000`).

litezzzout: I don't even know what it does, I'm guessing it's setting the lower bit,
jwortmann suggested using it, so I copy/pasta it.

jfcherng: To make `add_region()` able to change to foreground color, the background color for `add_region()` has to be different with the background color in the user's color scheme. To make the difference subtle so human eyes can't even distinguish it, people tend to just change the background color a little bit. But changing the alpha of the color doesn't work, at least on my side.

jwortmann: Yes, you are absolutely right @jfcherng, if the background is already completely black, then `#00000001` won't work. This was more of an example, and we should better describe it in the docs, similar to the unnecessary/deprecated diagnostic tags https://lsp.sublimetext.io/customization/#diagnostics.

plugin/core/sessions.py
```diff
@@ -257,6 +257,56 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
...
+                    "async",
+                    "modification",
+                    "documentation",
+                    "defaultLibrary"
+                ],
+                "formats": [
+                    "relative"
+                ],
+                "requests": {
+                    "range": True,
```
jwortmann, line 303: This doesn't reflect what happens in the code.

plugin/core/sessions.py
```diff
@@ -257,6 +257,56 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
...
+                    "full": {
+                        "delta": True
```
jwortmann, line 305: Same here.

plugin/semantic.py
litezzzout: > To make `add_region()` able to change the foreground color, the background color for `add_region()` has to be different with the `--background` color in the user's color scheme ([ST core bug](https://github.com/sublimehq/sublime_text/issues/817)?). To make the difference subtle so human eyes can't even distinguish it, plugin devs tend to just change the `--background` color a little bit. But changing the alpha of the color doesn't work, at least on my side.
> 
> That said, if you use
> 
> ```js
> {
>     "scope": "meta.semantic-token.lsp",
>     "background": "#00000001",
> }
> ```
> 
> then your trick won't work for people who use `#000000` as their `--background` color.
> 
> Similarly, if you use
> 
> ```js
> {
>     "scope": "meta.semantic-token.lsp",
>     "background": "#01010101",
> }
> ```
> 
> then your trick won't work for people who use `#010101` as their `--background` color. But there should be less people use `#010101` as their `--background` color.

I knew this much but, I wasn't entirely sure adding "background": "#00000001",  is masking the existing  background color or what?

plugin/core/sessions.py
```diff
@@ -257,6 +257,56 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
...
+                    }
+                },
+                "multilineTokenSupport": False,
```
jwortmann, line 308: Is there a reason why we wouln't support multiline tokens? I think it should already work automatically with add_regions for multiline tokens, no?

plugin/semantic.py
```diff
@@ -0,0 +1,232 @@
...
+from .core.registry import windows
+from .core.sessions import SessionBufferProtocol
+from .core.typing import List, Optional, Any, Dict, Tuple, Sequence, cast
+from .core.views import text_document_identifier
+
+import sublime
+import sublime_plugin
+import time
+
+SUBLIME_WORD_MASK = 515
```
jwortmann, line 14: What is this magic number and why isn't it used in the code?

plugin/semantic.py
jfcherng: Anyway, it should be a color which is close to user's `--background` but not exactly the same, for best UX.

plugin/semantic.py
```diff
@@ -0,0 +1,232 @@
...
+
+NON_MODIFIER_SCOPES = {
+            'variable'     :'variable.other.lsp',
+            'parameter'    :'variable.parameter.lsp',
+            'function'     :'variable.function.lsp',
+            'method'       :'variable.function.lsp',
+            'property'     :'variable.other.member.lsp',
+            'class'        :'entity.name.class.lsp',
```
jwortmann, line 22: I think this may lead to problems, because `entity.name` scopes should only used for definitions. See my Java example in a comment above. But I'm not sure what would be the best scope name here. Do you maybe have a screenshot where this scope is marked in e.g. magenta, so we can see where exactly this "class" token type is used by a server?

plugin/semantic.py
```diff
@@ -0,0 +1,232 @@
...
+            'enum'         :'variable.enum.lsp',
+            'enumMember'   :'entity.name.enum.lsp',
```
jwortmann, line 24: This is `constant.other.enum` in most syntaxes.

plugin/semantic.py
```diff
@@ -0,0 +1,232 @@
...
+            'type'         :'storage.type.lsp',
+            'macro'        :'variable.other.constant.lsp',
```
jwortmann, line 26: The (3rd-party) syntax for the Julia language (where macros are commonly used) uses `variable.macro`, which I think would fit better here. Not sure in which other languages something like macros are used and how they are scoped elsewhere.

plugin/semantic.py
litezzzout: by masking I meant bit-masking

jfcherng: > by masking I meant bit-masking

no. that trailing 01 is alpha.

plugin/semantic.py
```diff
@@ -0,0 +1,232 @@
...
+            'namespace'    :'variable.other.namespace.lsp',
+            'typeParameter':'variable.parameter.generic.lsp',
+            'comment'      :'comment.block.documentation.lsp',
+            'dependent'    :'',
+            'concept'      :'',
+            'module'       :'',
+            'magicFunction':'',
+            'selfParameter':'',
```
jwortmann, line 34: I think here in this LSP package we should only provide scopes for the standardized token types from the specs. And then we should provide a way for LSP-* packages to add mappings with such custom types only used by a certain server via `AbstractPlugin.configuration`. Those should be merged together then, also allowing LSP-* packages to override scopes for the predefined types, if needed.

plugin/semantic.py
```diff
@@ -0,0 +1,232 @@
...
+[MODIFICATION_SCOPES.update({k:v + ', semantic.'+ k + '.modification.lsp'})   for k,v in MODIFICATION_SCOPES.items()]
+[DEFAULT_LIB_SCOPES.update( {k:v + ', semantic.'+ k + '.defaultLibrary.lsp'}) for k,v in DEFAULT_LIB_SCOPES.items()]
+
+SEMANTIC_SCOPES = {
+            ''              :NON_MODIFIER_SCOPES,# if no modifiers are provided
+
+            "declaration"   :DECLARATION_SCOPES,
+            "definition"    :DECLARATION_SCOPES,
+
+            "readonly"      :{'variable': 'constant.other.lsp, variable.other.constant.lsp, semantic.variable.readonly.lsp'},
```
jwortmann, line 73: This should only use one of the scopes.

plugin/core/sessions.py
litezzzout: @jfcherng added this to support pylance, clangd worked without it.

plugin/core/sessions.py
jfcherng: These are simply copied from what VSCode sends to Pylance. I have no idea what it should be accurately in ST.

jwortmann:
> 3. For now fixed diagnostic/highlighting drawing with a dirty hack. needs to be tested more to see what I have broke.

Could you explain how it would work? As already pointed out, we probably need to initialize *all* regions keys for each new view in the desired order, and `View.erase_regions` will do nothing in this regard. Also the order should depend on user settings for document highlights, i.e. if it is set to "fill", then it should be drawn below diagnostics, and otherwise on top of diagnostics. I think the regions for semantic tokens can always be drawn at the bottom (i.e. its region key initialized first).

> added custom scopes that directcly correspond to tokenTypes/modifierTypes:
syntax:- scope: semantic.tokenType.modifierType

This would be 22 × 10 = 220 combinations, not counting additional custom types and modifiers from servers. Also this logic cannot work at all if multiple modifiers are used for a type. Customization of the scopes used in the mapping in combination with the `meta.semantic-token` should already be sufficient for color schemes. If you want you could add a `meta.tokenType` on top, to ensure that the mapping is injective (ignoring modifiers).

***

In general, I guess most of the code could go into plugin/documents.py, maybe with something like the following as a starting point?:

```diff
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -262,6 +262,18 @@ class Request:
     def documentHighlight(cls, params: Mapping[str, Any], view: sublime.View) -> 'Request':
         return Request("textDocument/documentHighlight", params, view)
 
+    @classmethod
+    def semanticTokens(cls, params: Mapping[str, Any], view: sublime.View) -> 'Request':
+        return Request("textDocument/semanticTokens/full", params, view)
+
+    @classmethod
+    def semanticTokensDelta(cls, params: Mapping[str, Any], view: sublime.View) -> 'Request':
+        return Request("textDocument/semanticTokens/full/delta", params, view)
+
+    @classmethod
+    def semanticTokensRange(cls, params: Mapping[str, Any], view: sublime.View) -> 'Request':
+        return Request("textDocument/semanticTokens/range", params, view)
+
     @classmethod
     def resolveCompletionItem(cls, params: CompletionItem, view: sublime.View) -> 'Request':
         return Request("completionItem/resolve", params, view)

--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -140,6 +140,7 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
     color_boxes_debounce_time = FEATURES_TIMEOUT
     highlights_debounce_time = FEATURES_TIMEOUT
     code_lenses_debounce_time = FEATURES_TIMEOUT + 2000
+    semantic_tokens_debounce_time = FEATURES_TIMEOUT
 
     @classmethod
     def applies_to_primary_view_only(cls) -> bool:
@@ -210,6 +211,7 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         if added:
             self._do_color_boxes_async()
             self._do_code_lenses_async()
+            self._do_semantic_tokens_async()
 
     def on_session_shutdown_async(self, session: Session) -> None:
         removed_session = self._session_views.pop(session.config.name, None)
@@ -318,6 +320,8 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         self.do_signature_help_async(manual=False)
         self._when_selection_remains_stable_async(self._do_code_lenses_async, current_region,
                                                   after_ms=self.code_lenses_debounce_time)
+        self._when_selection_remains_stable_async(self._do_semantic_tokens_async, current_region,
+                                                  after_ms=self.semantic_tokens_debounce_time)
 
     def get_uri(self) -> str:
         return self._uri
@@ -719,6 +723,21 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
             sublime.status_message('Completion error: {}'.format(error_messages))
         resolve_completion_list(items, flags)
 
+    # --- textDocument/semanticTokens -------------------------------------------------------------------------------
+
+    def _do_semantic_tokens_async(self) -> None:
+        session = self.session_async("semanticTokensProvider")
+        if session and "background" in self.view.style_for_scope("meta.semantic-token"):
+            params = {"textDocument": text_document_identifier(self.view)}
+            session.send_request_async(
+                Request.semanticTokens(params, self.view), self._on_semantic_tokens_async)
+
+    def _on_semantic_tokens_async(self, response: Optional[Dict]) -> None:
+        session = self.session_async("semanticTokensProvider")
+        token_legend = ...
+        # then encode tokens and draw regions etc.
+
     # --- Public utility methods ---------------------------------------------------------------------------------------
 
     def session_async(self, capability: str, point: Optional[int] = None) -> Optional[Session]:
```

plugin/semantic.py
litezzzout: @jwortmann 
you mean this one:
`"readonly" :{'variable': 'constant.other.lsp, variable.other.constant.lsp, semantic.variable.readonly.lsp'},`?

plugin/core/sessions.py
jfcherng: Yes. Here we only implement the `full` one.

plugin/core/sessions.py
jfcherng: Yes, `textDocument/semanticTokens/full/delta` is not implemented.

plugin/core/sessions.py
```diff
@@ -257,6 +257,56 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
...
+                "overlappingTokenSupport": False
```
jfcherng, line 309: Wondering what does this mean 🤔 

```js
	/**
	 * Whether the client supports tokens that can overlap each other.
	 */
	overlappingTokenSupport?: boolean;
```

plugin/core/sessions.py
litezzzout: @jfcherng added this to support his pylance server, clangd worked but adding this.

> Could you explain how it would work? As already pointed out, we probably need to initialize _all_ regions keys for each new view in the desired order, and `View.erase_regions` will do nothing in this regard. Also the order should depend on user settings for document highlights, i.e. if it is set to "fill", then it should be drawn below diagnostics, and otherwise on top of diagnostics. I think the regions for semantic tokens can always be drawn at the bottom (i.e. its region key initialized first).

key used for diagnostic/doc-highlight drawing was a static string. but if you used a new key for every draw call and store that key and used it in clear call, diagnostic/highlighting should draw on top on any 'add_region'.

It works so far I have tested I didn't face any issue drawing diagnostics/highlighting. It didn't worked before at, only drew on 'non add_region' scopes or if server stopped drawing on typo variable name.
diag_highlighting_underline
diag
doc_highlight_fill

plugin/core/sessions.py
jwortmann: Maybe this could be used for things like tags, parameters etc. within doc comments (for JavaDoc/JSDoc highlighting). So the whole doc comment block would get the "comment" token type with "documentation" modifier, and the tags within that block get another token added to that. Not sure if we handle this correctly, I guess one region would just override the other if they are overlapping. This is probably something where the initializing order for the region keys will be important again. But I guess we could set this parameter to `True`.

jfcherng:
> > Noticed a small problem. A diagnostic error (red underline) may be replaced with the sematic highlight (due to `add_region()`).
> 
> diagnostics_works
> 
> this should work now

Just found another problem. Those red underlines won't goes always after fixing that error.

jwortmann:
> key used for diagnostic/doc-highlight drawing was a static string. but if you used a new key for every draw call and store that key and used it in clear call, diagnostic/highlighting should draw on top on any 'add_region'.

Ah okay, yeah this is kind of an ugly workaround :D 
But it doesn't achieve what we want to do, right? We want to draw diagnostics *always* on to of semantic tokens. If you just add a timestamp to a region key, it still depends on the order when some user operations happen or the exact time when diagnostics are reported by a server, and it is pure luck whether it will work or not. I have tried to explain this problem several times now, but my comments seem to be ignored xD
Here is the link again: #1593
So I would say we need something like this (pseudocode!):
```python
r = [sublime.Region(0,0)]
view.add_regions('lsp_semantic_token_namespace', r)
view.add_regions('lsp_semantic_token_class', r)
view.add_regions('lsp_semantic_token_type', r)
... # and all other token types here
if user_settings.document_highlight_style == "fill":
   view.add_regions('lsp_document_highlights', r)   # I haven't checked the correct key name here and in the lines below
view.add_regions('lsp_diagnostics_hint', r)
view.add_regions('lsp_diagnostics_info', r)
view.add_regions('lsp_diagnostics_warning', r)
view.add_regions('lsp_diagnostics_error', r)
# also need to initialize the diagnostic tags unnecessary/deprecated somewhere here
if user_settings.document_highlight_style != "fill":
    view.add_regions('lsp_document_highlights', r)
```

This must be put somewhere when a view/SessionView initializes. So that it will be executed before any actual regions are drawn.

litezzzout:
> This would be 22 × 10 = 220 combinations, not counting additional custom types and modifiers from servers. 

I was thinking more the 10 or 15 combinations. just like I have in vscode.
`    "editor.semanticTokenColorCustomizations": {
        "rules": {
            "variable": "#94dbfd",
            "property": "#DADADA",
            "parameter": "#C586C0",
            "class": "#4eaab0",
            "macro": "#beb7ff",
            "function": "#dbdbaa",
            "type": "#4ec9b0",
            "enum": "#4eaacf",
            "enumMember": "#DADADA",
            "operator": "#C586C0"
        },
        "[Default Dark+]": {}
    },`
modifierTypes are optional, only set it you want it.

>Also this logic cannot work at all if multiple modifiers are used for a type. Customization of the scopes used in the mapping in combination with the `meta.semantic-token` should already be sufficient for color schemes. If you want you could add a `meta.tokenType` on top, to ensure that the mapping is injective (ignoring modifiers).

I'm entirely not sure about this, but if one scope has multiple modifiers, and if user only wants to target one
 or same color for multiple ones,
 like this 
`        {
            "name": "markup bold/italic",
            "scope": "markup.italic markup.bold | markup.bold markup.italic",
            "font_style": "bold italic"
        },`
wouldn't that work?

jwortmann:
> I was thinking more the 10 or 15 combinations. just like I have in vscode.

Hm okay, maybe you are right and this could work. But then I think at least we can use the `meta.semantic-token` scope for this, but now added on top, for example:
* `variable.function.lsp meta.semantic-token.function.lsp` for type=function, no modifier
* `entity.name.function.lsp meta.semantic-token.function.declaration.lsp` for type=function, modifier=declaration

I think this should still work with the background color rule for only `meta.semantic-token`, but provides an easy way to target each of the token types (possibly with modifiers) in the color scheme. Could you try this? (Btw, there must *not* be a comma between the scope parts, only a regular whitespace. Commas can only be used in scope *selectors* in the color scheme.)

litezzzout:
> ```python
> r = [sublime.Region(0,0)]
> view.add_regions('lsp_semantic_token_namespace', r)
> ```

works

litezzzout:
@jwortmann 
what to use for these two
`            'class'        :'',
`
 `           'enumMember'   :'',
`
i tried https://www.sublimetext.com/docs/scope_naming.html but didn't find anything that fits the bill

jwortmann:
"enumMember" should be `constant.other.enum`.

For "class", there is the following suggestion in the guidelines:

> While also used for base frameworks, many syntaxes apply these to scopes unrecognized classes and types, effectively scoping all user constructs.
>
> * support.type
> * support.class

But in some of the default syntaxes this is handled differently, so there seems to be no clear convention. Maybe you can test it and see how it looks with the server you use.

I see there is also a "defaultLibrary" modifier for tokens. I think we could use this e.g. for functions/methods, i.e. the scope with defaultLibrary could be `support.function.lsp meta.semantic-token.function.defaultLibrary.lsp`, instead of `variable.function`.

litezzzout:
> * support.type
> * support.class

both of these did the same thing to python class, (it was orange before)
only other server I have is clangd and I only use structs there
support

im not familiar with this color scheme so can't tell if it correct or not

litezzzout:
@jwortmann doc highlighting is still hit or miss, not sure if I'm doing something wrong.

`def init_semantic_draw_keys(view: sublime.View, server: str, doc_highlight_style: str) -> None:`
   r = [sublime.Region(0, 0)]
    for modifiers in SEMANTIC_SCOPES.values():
       for scope in modifiers.values():
            view.add_regions(scope, r)

    if doc_highlight_style == 'fill':
        view.add_regions('lsp_highlight_texts',  r)
        view.add_regions('lsp_highlight_reads',  r)
        view.add_regions('lsp_highlight_writes', r)

        view.add_regions('lsp_highlight_textm',  r)
        view.add_regions('lsp_highlight_readm',  r)
        view.add_regions('lsp_highlight_writem', r)
    for i in range(1, 5):
        view.add_regions('lsp{}d{}{}'.format(server, 's', i), r)
        view.add_regions('lsp{}d{}{}'.format(server, 'm', i), r)

    if doc_highlight_style != 'fill':
        view.add_regions('lsp_highlight_texts',  r)
        view.add_regions('lsp_highlight_reads',  r)
        view.add_regions('lsp_highlight_writes', r)


I'm calling this in `session_veiw.py`

`class SessionView:
...`
`    def __init__(self, listener: AbstractViewListener, session: Session) -> None`:
`        init_semantic_draw_keys(listener.view, session.config.name, userprefs().document_highlight_style)`
`        self.view = listener.view`

rwols (changes requested):
I have been closely following this PR and it certainly looks cool and tempting. I do think we need to play with the scope mapping more to get a better mapping.

Could you also add a boolean setting to enable semantic tokens? It should be disabled by default.

plugin/documents.py
```diff
@@ -319,6 +321,11 @@ def on_text_changed_async(self, change_count: int, changes: Iterable[sublime.Tex
         self._when_selection_remains_stable_async(self._do_code_lenses_async, current_region,
                                                   after_ms=self.code_lenses_debounce_time)
 
+        for change in changes:
+            if len(change.str) == 1 and (change.str == ' ' or change.str == ';' or change.str == '.'):
+                self._when_selection_remains_stable_async(self._do_semantic_tokens_async, current_region,
+                                                          after_ms=self.semantic_tokens_debounce_time)
```
rchl, line 327: What's the reasoning for this logic here? Why do we need to trigger semantic tokens on those changes and not the any any other and is this logic universal to all languages?

plugin/documents.py
```diff
@@ -657,6 +664,62 @@ def render_highlights_on_main_thread() -> None:
...
         sublime.set_timeout(render_highlights_on_main_thread)
 
+    # --- textDocument/semanticTokens -------------------------------------------------------------------------------
+
+    def _do_semantic_tokens_async(self) -> None:
+        if not userprefs().enable_semantic_tokens:
+            return
+        if "background" not in self.view.style_for_scope("meta.semantic-token.lsp"):
+            color_scheme_error()
+            return
```
rchl, line 674: I think that we shouldn't be triggering this error/warning every time the tokens are requested.
Maybe we should only have this information in the documentation and the documentation should be linked from the description of the setting that enables this feature.

plugin/semantic.py
```diff
@@ -0,0 +1,173 @@
...
+        view.add_regions('lsp_highlight_readm',  r)
+        view.add_regions('lsp_highlight_writem', r)
+    for i in range(1, 5):
+        view.add_regions('lsp{}d{}{}'.format(server, 's', i), r)
+        view.add_regions('lsp{}d{}{}'.format(server, 'm', i), r)
+
+    if doc_highlight_style != 'fill':
+        view.add_regions('lsp_highlight_texts',  r)
+        view.add_regions('lsp_highlight_reads',  r)
+        view.add_regions('lsp_highlight_writes', r)
```
rchl, line 173: - This is quite bad from maintenance point of view.
 - It's not really specific to semantic tokens
 - It can easily get out of sync with the actual implementation of those regions

I think this should be both in a different, more generic place (maybe `views.py`) and be done in a way that doesn't risk getting out of date easily. Should maybe be done using some sort of pattern where various places that create regions provide the regions they operate on and the "priority" those should be drawn.

This in fact could be done as a separate PR since it's not necessarily specific to this task.

rchl (commented):
Apart the individual comments, there are also various stylisting, ordering and type issues that should be fixed first but it would probably be easier for me or @rwols to do that before integration than reporting dozens of nitpicky issues.

plugin/semantic.py
rchl: Also there probably would be issues with that if the user changes settings at runtime.

plugin/documents.py
litezzzout: basically the idea is to only trigger semantic tokens after typing a whole word, otherwise sometimes it freaks out and draws incorrectly, especially while holding `backspace` or `return` keys

plugin/core/views.py
```diff
@@ -28,6 +28,7 @@
 import sublime
 import sublime_plugin
 import tempfile
+from ..semantic import SEMANTIC_SCOPES
```
rchl, line 31: It's not really "allowed" to import non-core stuff from 'core'.
So it looks like it wasn't not a good idea to have `init_region_keys` here. Or `SEMANTIC_SCOPES` should maybe be a core thing.

litezzzout: also initializing keys in this order doesn't really gives us desired behavior every time, seems like this only works 50% of the time,  `lsp_highlight_{}` keys somehow manages to get initialized before this ever gets called.
would you happen to know what causes that?

plugin/core/views.py
```diff
@@ -808,3 +809,24 @@ def format_completion(
...
+    r = [sublime.Region(0, 0)]
+    for modifiers in SEMANTIC_SCOPES.values():
+        for scope in modifiers.values():
+            view.add_regions(scope, r)
+
+    if doc_highlight_style == 'fill':
+        for name in doc_highlight_kinds:
+            view.add_regions('lsp_highlight_{}{}'.format(name, 's'), r)
+            view.add_regions('lsp_highlight_{}{}'.format(name, 'm'), r)
+    for i in range(1, 5):
```
jwortmann, line 825: I think these should be initialized in reversed order, so that errors are drawn over warnings and infos, if they span the same region.

There are also currently missing the region keys for diagnostics with tags from
https://github.com/sublimelsp/LSP/blob/a31b6b45301d612e69bc9a492fb75251b98bd393/plugin/session_view.py#L212-L213
and those should be initialized before the "regular" diagnostics.

***

I don't know why the documentHighlights would only work 50% of the time, maybe you can debug it by adding a print statement here and another one in https://github.com/sublimelsp/LSP/blob/a31b6b45301d612e69bc9a492fb75251b98bd393/plugin/documents.py#L607
to see if the documentHighlights request is somehow sent before `init_region_keys` was called?

rwols:
Please keep up-to-date with the main branch by merging it into your branch.

plugin/core/views.py
litezzzout: I've tried to debug it with print statements before, ` def render_highlights_on_main_thread() -> None: `, isn't getting called before `init_region_keys` but still sometimes doc_highlight refuses to draw.

litezzzout: sometimes `def _clear_highlight_regions(self) -> None:` is getting called before `init_region_keys`,  in those cases , doc_highlighting  doesn't work. I tried to prevent 'def _clear_highlight_regions(self) -> None:' from getting called before `init_region_keys` but it didn't help.

LSP.sublime-settings
```diff
@@ -150,30 +186,56 @@
...
   // "clients":
   // {
   //   // Each new server configuration must have the following structure.
   //   "CLIENT_NAME": // some human-friendly name
   //   {
-  //     # Must-have settings (for new configurations):
+  //     # ----------- The following three settings are **required** -----------
+  //
+  //     // Enable or disable this client configuration.
+  //     "enabled": true,
```
rwols, line 202: What went wrong here? this is already in `main`

litezzzout: I might have merged incorrectly, in my machine it's showing 7 files changed, here it's 27

litezzzout: I forgot to do `fetch` i guess

rwols:
Please merge with `main` :)

jwortmann:
May I ask what is the status of this PR? There seem to be a few issues still unaddressed, for example the default token types and scope format.
If it's okay, maybe I could give it another try based on this PR. But I don't really know which servers support semantic highlighting, so I'm not sure how to test it.

rwols:
clangd has semantic highlighting. I think gopls has it as well if you enable a feature flag.

rwols:
I'm certainly interested in this feature as well. But overall I see these drawbacks:

- `add_regions` is a god function. It's used for a lot of features.
- It requires manual changes to the user's color scheme
- Does it work correctly and is it "out of my way" in all cases?
- What about the render performance? Is `add_regions` ready for potentially rendering regions spanning the whole buffer?

I want to keep the work done in this pull request nearby and not close it. At the same time I am hesitant to merge it. Once a feature is in, people will complain if that feature is removed again.

So my strategy here is to sit back and see how the situation will evolve.

jwortmann:
Btw, I had started a local re-implementation of this, with various changes and additions. I haven't had the time to finish it yet, so it's not in a state that works right now. But some things that I noticed and which are missing here are:

* should also implement the `workspace/semanticTokens/refresh` server request. I've tested with the Lua server (only) and that server will only send semantic tokens if this request is supported.
* tokens should be stored in the SessionView/SessionBuffer like diagnostics. Then we should send the full request only at the first time, and after that use the delta request to get only the changes, so that we can update only the affected regions.
* I think regions with tokens that don't exist anymore are not cleared here in this PR, so the region keys for the previous request result must be stored in the buffer/view too somehow
* I added an API function for LSP-* plugins to provide a tokenType -> scope mapping for custom token types
* added a command lsp_show_scope_name as an extension to the built-in show_scope_name to also show the token type and modifiers at the cursor position
* to map token types with modifiers to scope names gives headaches and I haven't figured out a good way yet for the modifiers (which would work together with the mentioned API function). But I prefer a scheme with something like this for the scope name: `<default_scope>.lsp meta.semantic-token.<token_type>.lsp` with `<token_type>` being the token type name, but all lowercased . Not sure how to best add modifiers into this scheme (especially if there are multiple) or whether it is really necessary.
* entry in the docs for semantic highlighting + how to customize the color scheme for token types. And sublime-package.json should be updated. I'd prefer a setting named `"semantic_highlighting": false`

rwols:
OK, I'll suspend my activities then :)

One thing I'm noticing is that showing the active line highlight causes the highlight to be behind the regions. But perhaps that's a core issue.

image

<hr>

> to map token types with modifiers to scope names gives headaches and I haven't figured out a good way yet for the modifiers (which would work together with the mentioned API function). But I prefer a scheme with something like this for the scope name: <default_scope>.lsp meta.semantic-token.<token_type>.lsp with <token_type> being the token type name, but all lowercased . Not sure how to best add modifiers into this scheme (especially if there are multiple) or whether it is really necessary.

Yes, the modifiers give me headaches as well. One problem is that the `add_regions` keys would explode combinatorically if we really want to account for all modifiers. As each combination of modifiers would require its own separate `add_regions` key.

For the record we have these modifiers:

- declaration
- definition
- readonly
- static
- deprecated
- abstract
- async
- modification
- documentation
- defaultLibrary

That's 10 modifier values. If we really wanted to account for all possible modifiers, we would need 2^10 = 1024 different region keys for a single token type (either the modifier is present or not, order doesn't matter). Multiplied by the number of token types (22), we would have in total 22528 different region keys for semantic highlighting.

We could restrict to at most 2 possible modifiers. Choosing at most 2 modifiers out of 10 has $0 \choose 10 + 1 \choose 10 + 2 \choose 10 = 56$ possible region keys for one token type. In total 1232. That's still a lot.

We could restrict to at most 1 possible modifier. Choosing at most 1 modifier out of 10 has $0 \choose 10 + 1 \choose 10 = 11$ possible region keys for one token type. In total 242. Still a lot!

I'm assuming here that maintaining lots of `add_regions` keys is somehow expensive. But maybe that's not the case.

In any case, this modifier issue turns out to be quite complex. Maybe leave it out for now as you suggest.

rwols:
Superseded by #1839. Thanks for your work @litezzzout.

PR #1752, opened 2021-06-16
Special-case env.PATH extending to prepend the original PATH

rchl:


rwols:
Could you add a unit test for this? So that an alarm bell goes off if this is ever refactored for some reason?

rwols (approved):

PR #1755, opened 2021-06-19
Refactor code lenses

Rapptz:
This changes the following:

* Code lens state is stored in its own class
* Code lenses are moved to SessionView instead of DocumentSyncListener
* The code lens debounce time is lowered to 300ms from 2300ms

I'm aware this is a big change. The development was discussed in the Discord server in preparation of allowing to render code lenses as phantoms. As a result the PR for that one is blocked until this one is finalised since it depends on the refactoring done here.

I tried my best to separate the code that dealt with the phantom representation but some of the data changes might still linger and seem bizarre.

### DIFF
--- a/plugin/code_lens.py
+++ b/plugin/code_lens.py
@@ -1,0 +1,185 @@
+from .core.protocol import CodeLens, Range
+from .core.typing import List, Tuple, Dict, Iterable, Generator
+from .core.registry import LspTextCommand
+from .core.registry import windows
+from .core.views import make_command_link
+from .core.views import range_to_region
+from html import escape as html_escape
+import itertools
+import sublime
+
+
+class CodeLensData:
+    __slots__ = (
+        'data',
+        'region',
+        'session_name',
+        'annotation',
+    )
+
+    def __init__(self, data: CodeLens, view: sublime.View, session_name: str) -> None:
+        self.data = data
+        self.region = range_to_region(Range.from_lsp(data['range']), view)
+        self.session_name = session_name
+        self.annotation = '...'
+        self.resolve_annotation()
+
+    def __repr__(self) -> str:
+        return 'CodeLensData(resolved={0}, region={1!r})'.format(self.is_resolved(), self.region)
+
+    def is_resolved(self) -> bool:
+        """A code lens is considered resolved if the inner data contains the 'command' key."""
+        return 'command' in self.data
+
+    def to_lsp(self) -> CodeLens:
+        copy = self.data.copy()
+        copy['session_name'] = self.session_name
+        return copy
+
+    @property
+    def small_html(self) -> str:
+        return '<small style="font-family: system">{}</small>'.format(self.annotation)
+
+    def resolve_annotation(self) -> None:
+        command = self.data.get('command')
+        if command is not None:
+            command_name = command.get('command')
+            if command_name:
+                self.annotation = make_command_link('lsp_execute', command['title'], {
+                    'session_name': self.session_name,
+                    'command_name': command_name,
+                    'command_args': command.get('arguments', []),
+                })
+            else:
+                self.annotation = html_escape(command['title'])
+        else:
+            self.annotation = '...'
+
+    def resolve(self, view: sublime.View, code_lens: CodeLens) -> None:
+        self.data = code_lens
+        self.region = range_to_region(Range.from_lsp(code_lens['range']), view)
+        self.resolve_annotation()
+
+
+class CodeLensView:
+    CODE_LENS_KEY = 'lsp_code_lens'
+
+    def __init__(self, view: sublime.View) -> None:
+        self.view = view
+        self._code_lenses = {}  # type: Dict[Tuple[int, int], List[CodeLensData]]
+
+    def clear(self) -> None:
+        self._code_lenses.clear()
+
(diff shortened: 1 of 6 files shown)


plugin/documents.py
```diff
@@ -169,7 +162,6 @@ def _cleanup(self) -> None:
         triggers = [trigger for trigger in triggers if 'server' not in trigger]
         settings.set("auto_complete_triggers", triggers)
         self._stored_region = sublime.Region(-1, -1)
-        self._color_phantoms.update([])
```
rwols, line 172: where did this line go?

Rapptz: I messed this up during a rebase when I was splitting up the PR, sorry. I've reverted it.

plugin/core/sessions.py
```diff
@@ -340,6 +340,15 @@ def on_request_finished_async(self, request_id: int) -> None:
...
 
+    def start_code_lenses_async(self) -> None:
+        ...
+
+    def resolve_visible_code_lenses_async(self) -> None:
+        ...
+
+    def get_resolved_code_lenses_for_region(self, region: sublime.Region) -> Generator[CodeLens, None, None]:
+        ...
+
```
rwols, line 351: I don't think these methods are required to be declared here? You should be able to remove them and mypy won't complain.

rwols: BTW, you can run `tox` locally to check mypy and flake8.

Rapptz: If they're not given a type error is given when you eventually access it down the chain since e.g. `AbstractViewListener` makes use of it and failing to meet that interface leads to a type error.

rwols: I see. get_resolved_code_lenses_for_region is used in the text command. start_code_lenses_async and resolve_visible_code_lenses_async can go away.

plugin/documents.py
```diff
@@ -529,80 +515,19 @@ def _on_color_boxes(self, response: Any) -> None:
...
-            region = range_to_region(Range.from_lsp(c["range"]), self.view)
-            self._code_lenses.append((c, region))
-            if "command" in c:
-                # We consider a code lens that has a command to be already resolved.
-                self._on_resolved_code_lens_async(name, index, region, c)
-            else:
-                self._render_code_lens(name, index, region, None)
-        self._code_lenses = list((c, range_to_region(Range.from_lsp(c["range"]), self.view)) for c in response)
-        self._resolve_visible_code_lenses_async()
+                    sv.start_code_lenses_async()
```
rwols, line 523: Because the data is now stored in SessionView, can't we now request code lenses for _all_ sessions attached to the view?

plugin/documents.py
```diff
@@ -529,80 +515,19 @@ def _on_color_boxes(self, response: Any) -> None:
...
-        visible: sublime.Region
-    ) -> Generator[Tuple[int, CodeLens, sublime.Region], None, None]:
-        for index, tup in enumerate(self._code_lenses):
-            code_lens, region = tup
-            if not code_lens.get("command") and visible.intersects(region):
-                yield index, code_lens, region
+        if session and session.uses_plugin():
+            for sv in self.session_views_async():
+                if sv.session == session:
+                    sv.resolve_visible_code_lenses_async()
```
rwols, line 530: resolve for _all_ sessions attached to the view because the data is stored per session now?

plugin/code_lens.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def is_empty(self) -> bool:
+        return not self._code_lenses
+
+    def clear_annotations(self) -> None:
+        for index, _ in enumerate(self._flat_iteration()):
+            self.view.erase_regions(self._region_key(index))
+
+    def _region_key(self, index: int) -> str:
+        return '{0}.{1}'.format(self.CODE_LENS_KEY, index)
```
rwols, line 82: We're still keying on only the index, not the session. Oh well, maybe for a next step.

rwols (approved):
Alright, let's roll with this. Couple things to do still:

- it probably doesn't work with cloned views (it didn't work with cloned views before either I think), code lens data needs to move to SessionBuffer for that. SessionView is for presentation of the data
- there's some flickering of the annotations when they're updated. Probably because they're updated from the worker thread.

plugin/session_view.py
```diff
@@ -267,5 +273,41 @@ def _start_progress_reporter_async(self, request_id: int, title: str) -> ViewPro
...
+        for code_lens in self._code_lenses.unresolved_visible_code_lens(self.view.visible_region()):
+            callback = functools.partial(code_lens.resolve, self.view)
+            promise = self.session.send_request_task(
+                Request("codeLens/resolve", code_lens.data, self.view)
+            ).then(callback)
+            promises.append(promise)
+        Promise.all(promises).then(lambda _: self._code_lenses.render())
+
+    def get_resolved_code_lenses_for_region(self, region: sublime.Region) -> Generator[CodeLens, None, None]:
+        yield from self._code_lenses.get_resolved_code_lenses_for_region(region)
```
rchl, line 310: This code is equivalent to the old one so not a new issue but I don't really see the point of returning a generator when the caller is always gonna iterate through all the results anyway. I haven't measured it but it must be quite a bit slower then just returning a plain or even a tuple.

PR #1758, opened 2021-06-20
Allow attaching to URI schemes other than the 'file' scheme

### LINKED ISSUE #1739: Attach to non-files
I would like to see language servers attached to views without filenames.

Attach to any view whose base scope matches the selector.

**Additional context**
Motivated by new views created ad-hoc with CTRL+N.
And motivated by language servers that understand a custom scheme like `deno:` or `json-schema:`.
One problem is that we should probably first listen for setting changes in the view to keep an eye on the current syntax.

rwols:
This allows servers to attach to schemes like:

- `buffer`: in-memory buffers
- `res`: ST resource files in .sublime-packages
- other foreign schemes, like `deno`, `json-schema`, `jdt`, etc

The way this works is that we have a new client configuration setting
called `"schemes"` which is a list of allowable URI schemes.

If `"schemes"` is not specified, then it's by default `["file"]`.

Resolves #1739

I've tried this with LSP-json, LSP-pyright and these language servers
are working well with non-file URIs. clangd on the other hand refuses
to work with in-memory buffers.


plugin/core/sessions.py
```diff
@@ -354,8 +359,12 @@ class SessionBufferProtocol(Protocol):
 
     session = None  # type: Session
     session_views = None  # type: WeakSet[SessionViewProtocol]
-    file_name = None  # type: str
-    language_id = None  # type: str
```
rwols, line 358: I may have broken some helper package with this, need to preserve the behavior just in case.

plugin/core/sessions.py
```diff
@@ -1120,6 +1138,8 @@ def open_uri_async(
                 result = Promise.packaged_task()  # type: PackagedTask[bool]
 
                 def open_scratch_buffer(title: str, content: str, syntax: str) -> None:
+                    global _pending_uris_to_set
+                    _pending_uris_to_set[title] = uri
```
rchl, line 1142: Could we use something more unique than title. Maybe view id?

plugin/core/sessions.py
```diff
@@ -658,6 +667,12 @@ def on_session_end_async(self) -> None:
 
 
 _plugins = {}  # type: Dict[str, Tuple[Type[AbstractPlugin], SettingsRegistration]]
+_pending_uris_to_set = {}  # type: Dict[str, DocumentUri]
+
+
+def get_possible_uri_to_set(title: str) -> Optional[str]:
+    global _pending_uris_to_set
+    return _pending_uris_to_set.get(title)
```
rchl, line 675: Shouldn't we remove the `title` from the dict once we find it?

plugin/documents.py
```diff
@@ -765,5 +787,12 @@ def clear_async() -> None:
 
         sublime.set_timeout_async(clear_async)
 
+    def _on_settings_object_changed(self) -> None:
+        new_syntax = self.view.settings().get("syntax")
+        if new_syntax != self._current_syntax:
+            self._current_syntax = new_syntax
+            self._cleanup()
+            self._setup()
+
```
rchl, line 796: Don't we have to unregister the listener and then register it back and trigger "on_activated" to trigger didOpen?

rchl:
Don't have time to investigate now but noticed that `languageId` is empty for LSP-eslint (at least in a Vue file).

plugin/documents.py
rwols: No need: container of SessionViews is cleared, _is_registered is set to False, and then on_activated_async is called by ST.

rwols:
@rchl in what circumstances is languageid empty? do you have an example?

plugin/documents.py
rchl: I haven't checked but I've assumed that `on_activated_async` won't necessarily be called. Not unless the user changes the focus of the files manually.

rchl:
We don't seem to be handling file renames and are sending the old file name to the server even after rename.

rchl:
Also found some bad issue where LSP gets confused with what file (URI) was closed and reports one for a different file.

I'm not entirely sure how this happens but have sort of solid repro so can look close when I have more time.

rwols:
> We don't seem to be handling file renames and are sending the old file name to the server even after rename.

Works fine for me in pyright:

```
::  -> LSP-pyright textDocument/didClose: {'textDocument': {'uri': 'buffer://sublime/156'}}
::  -> LSP-pyright textDocument/didOpen: {'textDocument': {'version': 15, 'text': 'import functools\n\n', 'uri': 'file:///Users/raoulwols/Desktop/test2.py', 'languageId': 'python'}}
```

pyright does have problems with publishing diagnostics though. It publishes diagnostics for `file:///156` instead of `buffer://sublime/156`. I think that's a bug in pyright.

rwols:
Saving existing files on disk to a different filename also works:

```
::  -> LSP-pyright textDocument/didClose: {'textDocument': {'uri': 'file:///Users/raoulwols/Desktop/test2.py'}}
::  -> LSP-pyright textDocument/didOpen: {'textDocument': {'languageId': 'python', 'version': 1, 'uri': 'file:///Users/raoulwols/Desktop/test3.py', 'text': 'import functools\n'}}
```

rchl:
I don't mean saving files. I mean open a normal file and use `rename` command on it.

rwols:
Seems to be missing callbacks in the ST API.

rwols:
made an issue for it https://github.com/sublimehq/sublime_text/issues/4540

rchl:
Instead of waiting for a fix in ST, maybe we should consider only falling back to a saved URI when the file has no filename so that we in most cases always access native view.file_name(). Otherwise it's easy for regressions and corner cases to bite is (besides the already mentioned one).

rwols:
>  maybe we should consider only falling back to a saved URI when the file has no filename

Sorry, could you expand more on what you mean here?

rchl:
Basically what I'm thinking is to only set `lsp_uri` in the view if the view has no file name. Then the `get_uri` method would get the `lsp_uri` from view, if set, otherwise calculate the URI from `view.file_name()`. I'm not sure how much of a performance impact would it have but it would ensure that we always ask the real source of truth for the file name instead of relying on our cached value and having to worry about invalidating it in all cases.

rwols:
Trying to figure out why the CI fails

plugin/documents.py
rwols: fixed in 5294c95

rwols:
Something in 3a495bc makes the CI fail spectacularly :D

rchl:
What's your plan regarding this issue: https://github.com/sublimelsp/LSP/pull/1758#issuecomment-868991028 ?

Integrate with that flaw, wait or have my suggested workaround?

While previously we didn't detect the renames either (I guess those would ideally cause didClose and didOpen), now renames cause various errors with various servers (I guess because we notify about file that is no longer on disk).

rwols:
I would go ahead with this in spite of not handling renames as it doesn’t work on main as well.

plugin/session_buffer.py
```diff
@@ -56,24 +57,20 @@ class SessionBuffer:
...
-            raise ValueError("missing filename")
         self.opened = False
         # Every SessionBuffer has its own personal capabilities due to "dynamic registration".
         self.capabilities = Capabilities()
         self.session = session_view.session
         self.session_views = WeakSet()  # type: WeakSet[SessionViewProtocol]
         self.session_views.add(session_view)
-        self.file_name = file_name
-        self.language_id = language_id
+        self.uri = uri
```
rchl, line 73: It's not really clear why this class has both `self.uri` and `self.get_uri()` and what's the difference between those. Maybe renaming this would help? Like `last_known_uri` maybe? It would also help with accidentally using the wrong wrong.

plugin/core/url.py
```diff
@@ -1,19 +1,63 @@
...
+    """
     parsed = urlparse(uri)
     assert parsed.scheme == "file"
     if os.name == 'nt':
         # url2pathname does not understand %3A (VS Code's encoding forced on all servers :/)
         return url2pathname(parsed.path).strip('\\')
     else:
         return url2pathname(parsed.path)
+
+
```
rchl, line 49: Can we have an alternative helper function for getting filename from URI? I'd need to use something in eslint.

Having to play with urllib doesn't sound like fun and also I wonder if such helper needs to apply some path transformations to stay consistent.

I imagine such helper could return a tuple with scheme and file path (if `file` scheme at least).

rchl:
> I would go ahead with this in spite of not handling renames as it doesn’t work on main as well.

But I could argue that the situation is worse now. Previously at least the server would more or less happily work after rename. Now there are errors like:

```
LSP-eslint: ENOENT: no such file or directory, realpath '/xxx/app/pages/notifications.vue'
```

Though, won't hold it against this change as previous behavior wasn't perfect either.

rwols:
Let's merge this then and consider the rename thing a separate issue.

plugin/core/url.py
rwols: Let's also do that in another PR?

PR #1759, opened 2021-06-21
Prevent infinite loop when server returns empty code lenses

rwols:
Happens with LSP-PowerShellEditorServices for instance.

### DIFF
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -294,9 +294,7 @@ class SessionView:
 
     def resolve_visible_code_lenses_async(self) -> None:
         if self._code_lenses.is_empty():
-            self.start_code_lenses_async()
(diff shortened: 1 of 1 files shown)


Rapptz (commented):
I think the fix for this one might have to be a little different. Probably by using a different sentinel to see if code lenses have been initialised at all. Right now I used an empty list but it seems that sentinel was wrong.

The idea behind this change was to populate code lenses without having to reopen the file.

deathaxe:
Just noticed that issue with LSP-pylsp, too. Causing plugin host to run with high CPU usage.

This PR at least causes it to calm down again.

rwols:
I’ll let you think of a better solution @Rapptz

PR #1762, opened 2021-06-26
Index SessionBuffer weak value dictionary by id(session)

rwols:
Instead of the name. The config name caused weird behavior when
using the hover functionality. The text command of hover keeps the
session buffers alive. This ensures new session buffers are created
when doing a LSP: Restart Server.

### DIFF
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -33,7 +33,7 @@ class SessionView:
     COMPLETION_PROVIDER_KEY = "completionProvider"
     TRIGGER_CHARACTERS_KEY = "completionProvider.triggerCharacters"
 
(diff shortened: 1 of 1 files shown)


plugin/session_view.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, listener: AbstractViewListener, session: Session) -> None:
         self._code_lenses = CodeLensView(self.view)
         settings = self.view.settings()
         buffer_id = self.view.buffer_id()
-        key = (session.config.name, session.window.id(), buffer_id)
+        key = (id(session), session.window.id(), buffer_id)
```
rchl, line 47: Since `session` id is so unique, we probably don't need the `window.id()` in the tuple anymore?

rchl (approved):

rchl (approved):

PR #1764, opened 2021-06-26
Add API for registering an external file watcher implementation

rchl:
Allows to create external package that implements the file watcher API.

See example implementation at https://github.com/sublimelsp/LSP-file-watcher-chokidar/pull/1

### DIFF
--- a/plugin/core/file_watcher.py
+++ b/plugin/core/file_watcher.py
@@ -1,0 +1,76 @@
+from .typing import List, Optional, Protocol, Tuple, Type
+from abc import ABCMeta
+from abc import abstractmethod
+
+
+DEFAULT_IGNORES = ['**/.git/**', '**/node_modules/**', '**/.hg/**']
+
+WatchKind = int
+FilePath = str
+FileEvent = Tuple[WatchKind, FilePath]
+
+
+class WatchKindValue:
+    CREATE = 1
+    CHANGE = 2
+    DELETE = 4
+
+
+WATCHER_TYPE_TO_KIND = {
+    'create': WatchKindValue.CREATE,
+    'change': WatchKindValue.CHANGE,
+    'delete': WatchKindValue.DELETE,
+}
+
+
+class FileWatcherProtocol(Protocol):
+    def on_file_event(self, events: List[FileEvent]) -> None:
+        ...
+
+
+class FileWatcher(metaclass=ABCMeta):
+    """
+    A public interface of a file watcher implementation.
+    """
+
+    @classmethod
+    @abstractmethod
+    def create(
+        cls,
+        root_path: str,
+        glob: str,
+        kind: WatchKind,
+        ignores: List[str],
+        handler: FileWatcherProtocol
+    ) -> 'FileWatcher':
+        """
+        Creates a new instance of the file watcher.
+
+        :param glob: The glob pattern to enable watching for.
+        :param kind: The kind of events that should be watched.
+        :param ignores: The list of glob patterns that should excluded from file watching.
+
+        :returns: A new instance of file watcher.
+        """
+        pass
+
+    @abstractmethod
+    def destroy(self) -> None:
+        """
+        Called before the file watcher is disabled.
+        """
+        pass
+
+
+watcher_implementation = None  # type: Optional[Type[FileWatcher]]
+
+
+def register_file_watcher_implementation(file_watcher: Type[FileWatcher]) -> None:
+    global watcher_implementation
+    if watcher_implementation:
+        print('LSP: Watcher implementation already registered. Overwriting.')
+    watcher_implementation = file_watcher
+
+
+def get_file_watcher_implementation() -> Optional[Type[FileWatcher]]:
+    return watcher_implementation
--- a/plugin/core/file_watcher.py
+++ b/plugin/core/file_watcher.py
@@ -32,13 +32,22 @@ def file_watcher_kind_to_lsp_file_change_type(kind: FileWatcherKind) -> FileChan
 
 
 class FileWatcherProtocol(Protocol):
-    def on_file_event(self, events: List[FileWatcherEvent]) -> None:
+    def on_file_event_async(self, events: List[FileWatcherEvent]) -> None:
+        """
+        Called on file watcher events.
+        This API must be triggered on async thread.
+
+        :param events: The list of events to notify about.
+        """
         ...
 
 
 class FileWatcher(metaclass=ABCMeta):
     """
     A public interface of a file watcher implementation.
+
+    The interface implements the file watcher and notifies the `handler` (through the `on_file_event_async` method)
+    on file event changes.
     """
 
     @classmethod
--- a/plugin/core/file_watcher.py
+++ b/plugin/core/file_watcher.py
@@ -1,30 +1,38 @@
-from .typing import List, Optional, Protocol, Tuple, Type
+from .protocol import FileChangeType, FileChangeTypeCreated, FileChangeTypeChanged, FileChangeTypeDeleted
+from .protocol import WatchKind, WatchKindCreate, WatchKindChange, WatchKindDelete
+from .typing import List, Literal, Optional, Protocol, Tuple, Type, Union
 from abc import ABCMeta
 from abc import abstractmethod
 
-
 DEFAULT_IGNORES = ['**/.git/**', '**/node_modules/**', '**/.hg/**']
+DEFAULT_KIND = WatchKindCreate | WatchKindChange | WatchKindDelete
 
-WatchKind = int
+FileWatcherKind = Union[Literal['create'], Literal['change'], Literal['delete']]
 FilePath = str
-FileEvent = Tuple[WatchKind, FilePath]
+FileWatcherEvent = Tuple[FileWatcherKind, FilePath]
 
 
-class WatchKindValue:
-    CREATE = 1
-    CHANGE = 2
-    DELETE = 4
+def lsp_watch_kind_to_file_watcher_kind(kind: WatchKind) -> List[FileWatcherKind]:
+    kinds = []  # type: List[FileWatcherKind]
+    if kind & WatchKindCreate:
+        kinds.append('create')
+    if kind & WatchKindChange:
+        kinds.append('change')
+    if kind & WatchKindDelete:
+        kinds.append('delete')
+    return kinds
 
 
-WATCHER_TYPE_TO_KIND = {
-    'create': WatchKindValue.CREATE,
-    'change': WatchKindValue.CHANGE,
-    'delete': WatchKindValue.DELETE,
-}
+def file_watcher_kind_to_lsp_file_change_type(kind: FileWatcherKind) -> FileChangeType:
+    return {
+        'create': FileChangeTypeCreated,
+        'change': FileChangeTypeChanged,
+        'delete': FileChangeTypeDeleted,
+    }[kind]
 
 
 class FileWatcherProtocol(Protocol):
-    def on_file_event(self, events: List[FileEvent]) -> None:
+    def on_file_event(self, events: List[FileWatcherEvent]) -> None:
         ...
 
 
@@ -39,7 +47,7 @@ class FileWatcher(metaclass=ABCMeta):
         cls,
         root_path: str,
         glob: str,
-        kind: WatchKind,
+        kind: List[FileWatcherKind],
         ignores: List[str],
         handler: FileWatcherProtocol
     ) -> 'FileWatcher':
--- a/plugin/core/file_watcher.py
+++ b/plugin/core/file_watcher.py
@@ -55,16 +55,16 @@ class FileWatcher(metaclass=ABCMeta):
     def create(
         cls,
         root_path: str,
-        glob: str,
-        kind: List[FileWatcherKind],
+        pattern: str,
+        events: List[FileWatcherKind],
         ignores: List[str],
         handler: FileWatcherProtocol
     ) -> 'FileWatcher':
         """
         Creates a new instance of the file watcher.
 
-        :param glob: The glob pattern to enable watching for.
-        :param kind: The kind of events that should be watched.
+        :param pattern: The glob pattern to enable watching for.
+        :param events: The type of events that should be watched.
         :param ignores: The list of glob patterns that should excluded from file watching.
 
         :returns: A new instance of file watcher.
--- a/plugin/core/file_watcher.py
+++ b/plugin/core/file_watcher.py
@@ -7,23 +7,23 @@ from abc import abstractmethod
 DEFAULT_IGNORES = ['**/.git/**', '**/node_modules/**', '**/.hg/**']
 DEFAULT_KIND = WatchKindCreate | WatchKindChange | WatchKindDelete
 
-FileWatcherKind = Union[Literal['create'], Literal['change'], Literal['delete']]
+FileWatcherEventType = Union[Literal['create'], Literal['change'], Literal['delete']]
 FilePath = str
-FileWatcherEvent = Tuple[FileWatcherKind, FilePath]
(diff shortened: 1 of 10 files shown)


plugin/core/sessions.py
```diff
@@ -1023,6 +1018,17 @@ def _handle_initialize_success(self, result: Any) -> None:
         code_action_kinds = self.get_capability('codeActionProvider.codeActionKinds')
         if code_action_kinds:
             debug('{}: supported code action kinds: {}'.format(self.config.name, code_action_kinds))
+        if self._watcher_impl:
+            watcher_config = self.config.file_watcher
+            if watcher_config and 'glob' in watcher_config:
+                kind = 0
+                for kind_type in watcher_config.get('kind', []):
+                    kind |= WATCHER_TYPE_TO_KIND.get(kind_type, 0)
+                if kind:
```
rchl, line 1027: `kind` should default to all types if not specified.

plugin/core/file_watcher.py
```diff
@@ -0,0 +1,76 @@
+from .typing import List, Optional, Protocol, Tuple, Type
+from abc import ABCMeta
+from abc import abstractmethod
+
+
+DEFAULT_IGNORES = ['**/.git/**', '**/node_modules/**', '**/.hg/**']
```
rwols, line 6: Would be better if this were binary_file_patterns and/or index_exclude_patterns. But the glob syntax/features of ST is unfortunately incomplete compared to VSCode.

plugin/core/file_watcher.py
```diff
@@ -0,0 +1,76 @@
...
+
+WatchKind = int
+FilePath = str
+FileEvent = Tuple[WatchKind, FilePath]
+
+
+class WatchKindValue:
+    CREATE = 1
+    CHANGE = 2
```
predragnikolic, line 15: Does CHANGE mean: 
1) File Modified
2) or File Renamed

predragnikolic: Shuld have seen the spec first:
```
export namespace WatchKind {
	/**
	 * Interested in create events.
	 */
	export const Create = 1;

	/**
	 * Interested in change events
	 */
	export const Change = 2;

	/**
	 * Interested in delete events
	 */
	export const Delete = 4;
}
```

rchl:
On the high-level this seems fine?

I'm open to nitpicking too. :)

rwols:
I’m quite busy with IRL responsibilities, will review maybe Thursday or Friday. But one thing I do miss is some sort of integration test with the test/server.py file.

plugin/core/file_watcher.py
```diff
@@ -0,0 +1,76 @@
...
+        pass
+
+
+watcher_implementation = None  # type: Optional[Type[FileWatcher]]
+
+
+def register_file_watcher_implementation(file_watcher: Type[FileWatcher]) -> None:
+    global watcher_implementation
+    if watcher_implementation:
+        print('LSP: Watcher implementation already registered. Overwriting.')
```
predragnikolic, line 71: Question:
Prefer `debug` instead of `print`?
Yes or no

rchl: I thought that in this case it makes sense to print it regardless if the debug setting is enabled since normal user should know that something is off if he/she manages to install two different watchers.

plugin/core/file_watcher.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        Called before the file watcher is disabled.
+        """
+        pass
+
+
+watcher_implementation = None  # type: Optional[Type[FileWatcher]]
+
+
+def register_file_watcher_implementation(file_watcher: Type[FileWatcher]) -> None:
```
predragnikolic, line 68: Lets say we have two plugins `LSP-angular` and `LSP-file-watcher-chokidar`.

With the way that ST loads plugins, 
`LSP-angular` will be loaded first, than `LSP-file-watcher-chokidar`. 

Do you maybe see some issue for plugins that would be loaded before the file watcher plugin is registered (with `register_file_watcher_implementation(FileWatcherChokidar)`)?

rchl: I don't foresee issues with that. `register_file_watcher_implementation` is called at the import stage while packages and LSP only initializes sessions after `plugin_loaded` which is guaranteed to happen later.

plugin/core/file_watcher.py
```diff
@@ -0,0 +1,76 @@
...
+
+WatchKind = int
+FilePath = str
+FileEvent = Tuple[WatchKind, FilePath]
+
+
+class WatchKindValue:
+    CREATE = 1
+    CHANGE = 2
+    DELETE = 4
```
predragnikolic, line 16: Why didn't you use `3` (by the spec) instead of `4` for the DELETE?

```
export namespace FileChangeType {
	export const Created = 1;
	export const Changed = 2;
	export const Deleted = 3;
}
```

sublime-package.json
```diff
@@ -26,6 +26,41 @@
               "type": "string",
               "deprecationMessage": "Use the \"selector\" key instead."
             },
+            "FileWatcher": {
```
predragnikolic, line 29: should we expose the FileWatcher in the settings and let the user (miss)configure it?

plugin/core/file_watcher.py
rchl: Well, spotted.
I've also noticed this while refactoring this (can't decide how to write that code since there is the LSP-side and API-side that should likely use different types...)

sublime-package.json
rchl: You mean global (LSP) settings? I don't think that would make sense. It's something specific to a language server.

plugin/core/file_watcher.py
predragnikolic: I saw that you mainly used 4, because it is convenient to use with `|=`.

IMO I think that LSP should follow the LSP spec, thus I would suggest `3`.
Plus, I would try to avoid any confusion latter for DELETE somewhere being 3 and somewhere being 4. 
But you decide. :)

predragnikolic:
> On the high-level this seems fine?

Yes, the api looks nice (not to many methods exposed and the one that are exposed are reasonably easy to understand) 🙂

plugin/core/file_watcher.py
rchl: I just didn't notice that the kind used when registering the watcher was a bit field while on notifying was just a number.

sublime-package.json
predragnikolic: I read the spec a bit more:

> The watched files notification is sent from the client to the server when the client detects changes to files and folders watched by the language client (note although the name suggest that only file events are sent it is about file system events which include folders as well). 

Who defines the glob patterns?
The server or the client.

I thought it was the server, now I think it is the client who defines `glob`s and `kind`s.

plugin/core/sessions.py
```diff
@@ -1253,10 +1269,16 @@ def m_client_registerCapability(self, params: Any, request_id: Any) -> None:
                     # Inform only after the response is sent, otherwise we might start doing requests for capabilities
                     # which are technically not yet done registering.
                     sublime.set_timeout_async(inform)
-            if self._plugin:
-                inform = functools.partial(
-                    self._plugin.on_register_capability_async, registration_id, capability_path, options)
-                sublime.set_timeout_async(inform)
+            if self._watcher_impl and capability_path == "didChangeWatchedFilesProvider":
```
predragnikolic, line 1272: `didChangeWatchedFilesProvider` is this string correct?

I think that it should be `didChangeWatchedFiles`

rchl: Our `method_to_capability` transforms the original `workspace/didChangeWatchedFiles` capability into `didChangeWatchedFilesProvider`. It's not a capability that is defined by the spec but that shouldn't really be a problem.

One benefit how this works is that the user can disable this capability using the `disabled_capabilities` option as long as he/she knows to use that specific string.

plugin/core/file_watcher.py
rchl: BTW. This is fixed.

plugin/core/sessions.py
```diff
@@ -979,6 +974,18 @@ def should_notify_did_save(self) -> Tuple[bool, bool]:
...
+
+    def on_file_event_async(self, events: List[FileWatcherEvent]) -> None:
+        changes = []  # type: List[FileEvent]
+        for event in events:
+            kind, filepath = event
+            changes.append({
+                'uri': filename_to_uri(filepath),
+                'type': file_watcher_kind_to_lsp_file_change_type(kind),
+            })
+        self.send_notification(Notification.didChangeWatchedFiles({'changes': changes}))
```
predragnikolic, line 987: If we have a file watcher installed,
then the LSP client will say that it support dynamicRegistration.
```   
if get_file_watcher_implementation():
    text_document_capabilities["didChangeWatchedFiles"] = {"dynamicRegistration": True}
```

> It is recommended that servers register for these file system events using the registration mechanism. _In former implementations clients pushed file events without the server actively asking for it._

Will the LSP client send `didChangeWatchedFiles` notifications to severs who are not interested in these type of notifications? 

My answer to that would be only if the server dynamically registered to that capability. If the server didn't registered than I would guess that dynamic file watchers would [not get instancated](https://github.com/sublimelsp/LSP/pull/1764/files#diff-8fe432afb96c719bc734bcb2a1045d852979d09ea395956f1b979861432c2b05R1293-R1294).
If file watchers are not instantiated than `on_file_event_async` would never be called, and the server will never be notified with the didChangeWatchedFiles notification.

... Sorry for these kind of questions, I am just trying to grasp this part of the spec :)

sublime-package.json
rchl: It's the server that wants to be notified about specific file events. For example vetur wants to know about changes in *.js files as it doesn't run in JS files but changes in JS files might affect the *.vue files. Or eslint server wants to know about changes in .eslintrc.* configuration file(s).

A generic client doesn't know what the server is interested in watching.

With dynamic registrations being available, the server triggers those to register the watchers.

Before dynamic registrations there was no way for server to register a watcher since there is no server capability for it or anything like that so before it worked because the client (vscode extension) had hardcoded patterns for the server that it was running.

plugin/core/sessions.py
predragnikolic: Thanks :+1:

plugin/core/sessions.py
rchl: It's like you said.

For servers that don't dynamically register the watcher but want to be notified about some file changes (like vetur or eslint), we need to define the configuration in their ClientConfig.

rchl:
I'll say it's ready for review although would be good to also handle ST's exclude patterns. Maybe it wouldn't be too hard to convert those to globs.

rchl:
Note:
 - changes in `sessions.py` are much more readable with "hide whitespace changes" enabled. Had to extract each capability group into a variable to avoid type issue.
 - remove the `on_register_capability_async` and `on_unregister_capability_async` APIs which to my knowledge were only used by me in my previous file watcher implementation.

tests/test_file_watcher.py
```diff
@@ -0,0 +1,181 @@
...
+        registration_params = {
+            'registrations': [
+                {
+                    'id': '111',
+                    'method': 'workspace/didChangeWatchedFiles',
+                    'registerOptions': {
+                        'watchers': [
+                            {
+                                'globPattern': '*.py',
+                                'kind': WatchKindCreate | WatchKindChange,
```
predragnikolic, line 157: Maybe use 
```
FileChangeTypeCreated = 1
FileChangeTypeChanged = 2
```
instead of 
```
WatchKindCreate = 1
WatchKindChange = 2
```
?

predragnikolic: For the concrete test, it would make no difference what ever you choose.

But if someone would add `WatchKindDelete` than it would cause a problem. 
But because this is a test, I guess it is not that important.

rchl: I will add `WatchKindDelete` to the test also.

But I can't use `FileChangeType` in place of `WatchKind` as those have different purpose. The latter is used in a bit field when registering the watcher dynamically. The former is a single value used to notify about file event through the `workspace/didChangeWatchedFiles` notification.

sublime-package.json
rwols: But now that dynamic capabilities exist for a while, is it still needed to have static glob patterns?

Do you have an example of a server that relies on the client knowing the correct glob patterns?

rchl: vetur and eslint, as mentioned, still rely on the watcher being registered by the client (the extension in case of vscode).

plugin/core/sessions.py
```diff
@@ -1267,8 +1289,12 @@ def m_client_unregisterCapability(self, params: Any, request_id: Any) -> None:
             capability_path, registration_path = method_to_capability(unregistration["method"])
             debug("{}: unregistering capability:".format(self.config.name), capability_path)
             data = self._registrations.pop(registration_id, None)
-            if data and self._plugin:
-                self._plugin.on_unregister_capability_async(registration_id, capability_path, data.options)
+            if self._watcher_impl and capability_path == "workspace.didChangeWatchedFiles":
+                file_watchers = self._dynamic_file_watchers.get(registration_id)
```
rwols, line 1293:
```suggestion
                file_watchers = self._dynamic_file_watchers.pop(registration_id, None)
```

plugin/core/sessions.py
```diff
@@ -1267,8 +1289,12 @@ def m_client_unregisterCapability(self, params: Any, request_id: Any) -> None:
...
+                if file_watchers:
+                    for file_watcher in file_watchers:
+                        file_watcher.destroy()
+                    del self._dynamic_file_watchers[registration_id]
```
rwols, line 1297:
```suggestion
```

sublime-package.json
```diff
@@ -26,6 +26,41 @@
...
+              "type": "object",
+              "examples": [
+                {
+                  "glob": "{**/*.js,**/*.ts,**/*.json}",
+                  "kind": ["create", "change", "delete"],
```
rchl, line 34: Maybe this should be called `type`...
Spec uses `kind` when registering but `type` when triggering events.

sublime-package.json
```diff
@@ -26,6 +26,41 @@
...
+                  "ignores": ["**/.git/**", "**/node_modules/**", "**/.hg/**"]
```
rchl, line 35: Might be good to revise the key naming here.
- `kind` - maybe it should be called `type`. Spec uses `kind` when registering but `type` when triggering events.
- `glob` and `ignores` - maybe `patternGlob` and `ignoresGlob` for consistency.

plugin/core/file_watcher.py
rwols: Wouldn't you need `**/.svn/**` here as well?

sublime-package.json
rwols: I think `glob` and `ignores` is fine. Best to avoid snake_case vs camelCase by using a single noun ;) I have no real opinion about kind vs type

rwols (approved):
Looks like you've put a lot of work into this.

sublime-package.json
rchl: If `ignores` is a glob and I'm not mentioning that in the key name then, for consistency, I think I'll rename `glob` to `pattern`. That will be more consistent.

plugin/core/file_watcher.py
rchl: Someone is using SVN still? :)

I can add of course but was planning too look into fetching ignores from ST and converting to glob but that might turn out to be some work.

rwols: > Someone is using SVN still? :)

Unfortunately, yes, these companies exist ;) But I think it's okay to forget about it and focus on fetching the ignore list from ST instead.

rwols:
Let's go ahead with this for now and work out the kinks.

PR #1766, opened 2021-06-27
Send a code lens request if one hasn't been sent yet

Rapptz:
This is an extended fix related to #1759. The original intent of the code was to allow code lenses to be filled without resorting to
reopening the file. However, the naive implementation caused an infinite loop so it requires a separate sentinel value to differentiate between an actual empty request vs an uninitialised request.


rwols (approved):
I'm getting confused by all this logic. It might be time for some integration tests with respect to code lenses. Look at tests/test_single_document.py

PR #1768, opened 2021-06-29
Ignore plugin classes that don't return name()

rchl:
Otherwise LSP will attempt to register intermediary classes like
NpmClientHandler and throw an error.

### DIFF
--- a/boot.py
+++ b/boot.py
@@ -80,6 +80,11 @@ def _register_all_plugins() -> None:
     plugin_classes = []  # type: List[Type[AbstractPlugin]]
     _get_final_subclasses(AbstractPlugin.__subclasses__(), plugin_classes)
     for plugin_class in plugin_classes:
+        try:
+            if not plugin_class.name():
+                continue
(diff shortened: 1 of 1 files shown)


predragnikolic (approved):

boot.py
```diff
@@ -80,6 +80,11 @@ def _register_all_plugins() -> None:
     plugin_classes = []  # type: List[Type[AbstractPlugin]]
     _get_final_subclasses(AbstractPlugin.__subclasses__(), plugin_classes)
     for plugin_class in plugin_classes:
+        try:
+            if not plugin_class.name():
+                continue
+        except NotImplementedError:
+            continue
```
rwols, line 87: Can't you check in _get_final_subclasses whether `d.__abstractmethods__` is not falsy?

rchl: I don't think this would work.

NpmClientHandler doesn't have any abstract methods, same as a subclass of it like LspPyright.

PR #1769, opened 2021-06-29
Update sublime-package.json for 'schemes'

rwols:
Also updated LSP.sublime-settings


sublime-package.json
```diff
@@ -80,6 +80,20 @@
...
+              "items": {
+                "type": "string"
+              },
+              "defaultSnippets": [
+                {
+                  "body": [
+                    "${1:file}"
+                  ]
+                }
+              ]
```
rchl, line 95: Could have `uniqueItems` also.

rchl (approved):

PR #1771, opened 2021-06-30
Account for FS case-insensitivity

### LINKED ISSUE #1770: On Windows, the drive letter in server responsed file URIs are lowercase.
**Describe the bug**

I tried both intelephense and pyright, they both returned lowercased drive letter thus I suspect it's a standard. (or maybe VSCode's LSP lib does it)

https://user-images.githubusercontent.com/6594915/123961095-96286c80-d9e2-11eb-8ada-0da9af754a55.mp4

In "Goto Definition...", this causes ST to open a file whose drive letter is in lowercase. And that may cause various mysterious problem sometimes... Or maybe, this should be fixed in ST core.

**To Reproduce**
1. Install LSP-intelephense with a Windows build ST
2. Open a PHP project
3. Make sure the definition file is not opened in a tab already
4. Do "Goto Definition"
5. The newly opened tab should have a lower drive letter

**Expected behavior**

The drive letter should be uppercase.

- OS: Win10 21H1 x64
- Sublime Text version: 4109
- LSP version: 4070-1.6.1
- Language servers used: intelephense, pyright

**Additional context**

This is a Windows-only issue as it's case-insensitive.

rwols:
~~Fixes #1770~~

Fixes diagnostics on Windows.

### DIFF
--- a/plugin/core/url.py
+++ b/plugin/core/url.py
@@ -1,6 +1,5 @@
 from .typing import Any, Tuple
 from urllib.parse import quote
-from urllib.parse import unquote
 from urllib.parse import urljoin
 from urllib.parse import urlparse
 from urllib.request import pathname2url
--- a/plugin/core/url.py
+++ b/plugin/core/url.py
@@ -47,19 +47,19 @@ def uri_to_filename(uri: str) -> str:
         return url2pathname(parsed.path)
 
 
-def parse_uri(uri: str) -> Tuple[bool, str]:
+def parse_uri(uri: str) -> Tuple[str, str]:
     """
-    Parses an URI into a tuple where the first element is True if the URI is a
-    file URI, otherwise False. The second element is the local filesystem path
-    if the URI is a file URI, otherwise the second element is the original URI.
+    Parses an URI into a tuple where the first element is the URI scheme. The
+    second element is the local filesystem path if the URI is a file URI,
+    otherwise the second element is the original URI.
     """
     parsed = urlparse(uri)
     if parsed.scheme == "file":
         if os.name == 'nt':
             # TODO: this is wrong for UNC paths
-            return True, url2pathname(parsed.path).strip('\\')
-        return True, url2pathname(parsed.path)
-    return False, uri
+            return parsed.scheme, url2pathname(parsed.path).strip('\\')
+        return parsed.scheme, url2pathname(parsed.path)
+    return parsed.scheme, uri
 
 
 def _to_resource_uri(path: str, prefix: str) -> str:
--- a/plugin/core/url.py
+++ b/plugin/core/url.py
@@ -1,4 +1,4 @@
-from .typing import Any
+from .typing import Any, Tuple
 from urllib.parse import quote
 from urllib.parse import unquote
 from urllib.parse import urljoin
@@ -48,12 +48,19 @@ def uri_to_filename(uri: str) -> str:
         return url2pathname(parsed.path)
 
 
-def path_url_to_fs_path(path: str) -> str:
+def parse_uri(uri: str) -> Tuple[bool, str]:
     """
-    Convert the "path" part of an URI to a path on the local filesystem.
-    The URI must have the "file" scheme.
+    Parses an URI into a tuple where the first element is True if the URI is a
+    file URI, otherwise False. The second element is the local filesystem path
+    if the URI is a file URI, otherwise the second element is the original URI.
     """
-    return re.sub(r'^/([a-zA-Z]:/.+)', r'\1', unquote(path))
+    parsed = urlparse(uri)
+    if parsed.scheme == "file":
+        if os.name == 'nt':
+            # TODO: this is wrong for UNC paths
+            return True, url2pathname(parsed.path).strip('\\')
+        return True, url2pathname(parsed.path)
+    return False, uri
 
 
 def _to_resource_uri(path: str, prefix: str) -> str:
--- a/plugin/core/url.py
+++ b/plugin/core/url.py
@@ -1,5 +1,6 @@
 from .typing import Any
 from urllib.parse import quote
+from urllib.parse import unquote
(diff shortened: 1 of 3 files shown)


rchl:
```
Error handling notification
  File "C:\Users\IEUser\AppData\Roaming\Sublime Text\Packages\LSP\plugin\core\sessions.py", line 1510, in on_payload
    handler(result)
  File "C:\Users\IEUser\AppData\Roaming\Sublime Text\Packages\LSP\plugin\core\sessions.py", line 1246, in m_textDocument_publishDiagnostics
    sb = self.get_session_buffer_for_uri_async(uri)
  File "C:\Users\IEUser\AppData\Roaming\Sublime Text\Packages\LSP\plugin\core\sessions.py", line 951, in get_session_buffer_for_uri_async
    session_buffer = next(filter(predicate, self.session_buffers_async()), None)
  File "C:\Users\IEUser\AppData\Roaming\Sublime Text\Packages\LSP\plugin\core\sessions.py", line 942, in compare_by_samefile
    return os.path.samefile(needle, urllib.parse.unquote(p.path))
  File "./python3.3/ntpath.py", line 603, in samefile
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: '/c:/Users/IEUser/AppData/Roaming/Sublime Text/Packages/lsp_utils/st3/lsp_utils/activity_indicator.py'
```

plugin/core/url.py
```diff
@@ -47,6 +48,14 @@ def uri_to_filename(uri: str) -> str:
         return url2pathname(parsed.path)
 
 
+def path_url_to_fs_path(path: str) -> str:
```
rchl, line 51: How about instead having a function that returns a tuple (scheme, filepath) where filepath is None for non-file schemas?

There is a lot of repetition right now, having to parse manually using some low-level functions.

jwortmann:
How is this PR related to the linked issue? The only place where `get_session_buffer_for_uri_async` is called, seems to be for diagnostics in https://github.com/sublimelsp/LSP/blob/a31b6b45301d612e69bc9a492fb75251b98bd393/plugin/core/sessions.py#L1220 and I already fixed the diagnostics with lowercase drive letter a year ago in #1133. (Maybe the code got refactored away, but it still works for me even with lowercase drive letter from the server)

The issue is about how the drive letter is displayed in the title bar for newly opened file URIs that *have not yet* an associated view. So if I understand it correctly, it is only a cosmetic issue.

Due to a bug (https://github.com/sublimehq/sublime_text/issues/4429) ST already treats all paths as case insensitive on Windows - even if they are not. To always display the drive letter as uppercase, I guess it would be okay to normalize it somewhere in https://github.com/sublimelsp/LSP/blob/a31b6b45301d612e69bc9a492fb75251b98bd393/plugin/core/sessions.py#L1107 (or in a related function), though.

rwols:
@jwortmann, are you sure you're up-to-date on the main branch? I recently introduced handling non-file URIs in https://github.com/sublimelsp/LSP/commit/7673446a46c161f3fa28ce240138949eeb622736, but forgot to account for `os.path.samefile` in this method.

plugin/core/url.py
rwols: I think this is a good idea, went with 3194319

plugin/core/url.py
```diff
@@ -47,6 +47,21 @@ def uri_to_filename(uri: str) -> str:
...
+    file URI, otherwise False. The second element is the local filesystem path
+    if the URI is a file URI, otherwise the second element is the original URI.
+    """
+    parsed = urlparse(uri)
+    if parsed.scheme == "file":
+        if os.name == 'nt':
+            # TODO: this is wrong for UNC paths
+            return True, url2pathname(parsed.path).strip('\\')
+        return True, url2pathname(parsed.path)
+    return False, uri
```
rchl, line 62: Well, this doesn't return the schema. Many places (in LSP at least) would benefit from a helper function like that if it would return the schema. I'm not sure if external code (helper packages) would benefit from knowing the schema. If not then maybe it's OK like that...

Can we also export this function in public plugin API?

jwortmann:
I have just checked out this branch at eb68fd3 and I can still reproduce the issue #1770, just as I would expect because this PR doesn't touch the related code at all.

> I recently introduced handling non-file URIs in 7673446, but forgot to account for os.path.samefile in this method.

All servers that I use only return file URIs for "Goto Definition", so I can't test it for non-file URIs. So is this PR not at all related to #1770? (because the video in that issue also shows a file URI) I'm confused now because you wrote `Fixes #1770` in the opening post.

rchl:
@jfcherng reported the issue that is fixed here on Discord but also reported #1770 so that's where the confusion is coming from I guess. :)

The issue fixed here is related to diagnostics not working on Windows.

rwols:
Yes, after a careful reading #1770 is unrelated.

plugin/core/url.py
rwols: > Can we also export this function in public plugin API?

Yes, I think this should be the public facing function for handling URIs.

rwols:
This fixes diagnostics for windows users in case the path differs in case. I have updated the opening post accordingly.

plugin/core/url.py
rwols: I'm still not totally happy with this function. If it's not a file URI, then you might need the URI fragment instead (for example). But then you would have to use the low-level urllib.parse functions again.

jwortmann:
> Yes, after a careful reading #1770 is unrelated.

Okay then :D

Yes, I was a few commits behind main but after updating to the latest commit, I can see diagnostics not working anymore on Windows (and the functionality restored on this branch).

rchl (approved):

plugin/core/url.py
rchl: It probably would be OK to provide two separate functions. Anything is better than using urlparse directly :)

PR #1772, opened 2021-07-01
Normalize drive letter in Windows filepaths to uppercase

### LINKED ISSUE #1770: On Windows, the drive letter in server responsed file URIs are lowercase.

jwortmann:
This fixes a bug with wrong labels for quick panel items after "Find References" and "Goto Definition..." if the server returns filepaths with lowercase drive letter, and also for Sublime's VCS status in the status bar when such a file link is opened.

Closes #1770.

### DIFF
--- a/plugin/core/url.py
+++ b/plugin/core/url.py
@@ -21,7 +21,6 @@ def filename_to_uri(file_name: str) -> str:
     if file_name.startswith(prefix) and not os.path.exists(file_name):
         return _to_resource_uri(file_name, prefix)
     path = pathname2url(file_name)
-    re.sub(r"^([A-Z]):/", _lowercase_driveletter, path)
     return urljoin("file:", path)
 
 
@@ -77,10 +76,3 @@ def _uppercase_driveletter(match: Any) -> str:
     For compatibility with Sublime's VCS status in the status bar.
     """
     return "{}:".format(match.group(1).upper())
-
-
-def _lowercase_driveletter(match: Any) -> str:
-    """
-    For compatibility with certain other language clients.
-    """
-    return "{}:/".format(match.group(1).lower())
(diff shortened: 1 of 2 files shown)


rwols:
How about removing the _lowercase_driveletter instead? I thought it would solve the duplicate goto-def issue but it looks like servers will have to account for that themselves. Which Intelephense does.

rchl:
> How about removing the _lowercase_driveletter instead? I thought it would solve the duplicate goto-def issue but it looks like servers will have to account for that themselves. Which Intelephense does.

Possibly it would even be safest to uppercase in both cases since the `filename_to_uri` might be used with file paths that come from somewhere else than ST (either server or user configuration).

rwols:
Can you add some tests for this in tests/test_url.py so we don't regress in the future?

jwortmann:
> Possibly it would even be safest to uppercase in both cases since the filename_to_uri might be used with file paths that come from somewhere else than ST (either server or user configuration).

Seems like the `pathname2url` already does the uppercasing for us if the filename has a lowercase drive letter. I removed the obsolete function.

> Can you add some tests for this in tests/test_url.py so we don't regress in the future?

Yes, I guess these tests need to be adjusted now anyway to pass.

jwortmann:
I adjusted the outdated test, but I think the tests should already cover the critical cases. Anything more to add there?

rwols (approved):

PR #1777, opened 2021-07-03
Allow styling of code lens phantoms

jwortmann:
I installed LSP-jdtls and dug out an old Java project to try out the new code lens phantoms, and saw that I could not target them specifically to adjust their style. Fixed it now.

Example with `"phantom_css": "html{background-color:transparent;color:var(--foreground)}a{color:var(--bluish)}#lsp-code-lens a{color:color(var(--foreground) alpha(0.5));text-decoration: none}",` in Mariana:

codelens

***

By the way, there is still a bug (and error in the console) with file URIs to filepaths conversion when I click on the phantom (or annotation). The normal references request e.g. via the context menu works. I will look into that later.

### DIFF
--- a/plugin/code_lens.py
+++ b/plugin/code_lens.py
@@ -150,7 +150,8 @@ class CodeLensView:
             for key, group in self._code_lenses.items():
                 region = sublime.Region(*key)
                 phantom_region = self._get_phantom_region(region)
-                html = '\n<small style="font-family: system">|</small>\n'.join(lens.small_html for lens in group)
+                html = '<body id="lsp-code-lens">{}</body>'.format(
+                    '\n<small style="font-family: system">|</small>\n'.join(lens.small_html for lens in group))
                 phantoms.append(sublime.Phantom(phantom_region, html, sublime.LAYOUT_BELOW))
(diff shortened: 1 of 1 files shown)


rwols (approved):

plugin/code_lens.py
```diff
@@ -150,7 +150,8 @@ def render(self, mode: str) -> None:
             for key, group in self._code_lenses.items():
                 region = sublime.Region(*key)
                 phantom_region = self._get_phantom_region(region)
-                html = '\n<small style="font-family: system">|</small>\n'.join(lens.small_html for lens in group)
+                html = '<body id="lsp-code-lens">{}</body>'.format(
```
predragnikolic, line 153: Why `id` instead of the `class` attribute?

jwortmann: Because it is the recommended way to do it: https://www.sublimetext.com/docs/minihtml.html#best_practices

And maybe it helps to avoid accidentally using it here in this package to style the phantoms (if we want to do that, we could wrap the inner html into a div tag with another class and use that class to add styles).

PR #1778, opened 2021-07-04
Fix location_to_encoded_filename function

rwols:


plugin/core/views.py
```diff
@@ -194,7 +202,11 @@ def location_to_encoded_filename(location: Union[Location, LocationLink]) -> str
     """
     DEPRECATED
     """
-    return to_encoded_filename(*get_uri_and_position_from_location(location))
+    uri, position = get_uri_and_position_from_location(location)
+    scheme, parsed = parse_uri(uri)
+    if scheme == "file":
+        return to_encoded_filename(parsed, position)
```
rchl, line 208: It looks like `to_encoded_filename` received an URI before and now it will receive file path.

rchl (changes requested):

plugin/core/views.py
rwols: Yes, that was a bug :)

rwols: I would actually like to remove the function, but it's still in use by LSP-jdtls and LSP-metals I believe.

rchl (approved):

PR #1779, opened 2021-07-04
Open existing views with custom URIs

rwols:
This doesn't actually work correctly for session buffers that get a different
session attached to them. For example, when you open a remote JSON schema with
the yaml-language-server, then the vscode-json-language-server is attached to
that virtual view. In that case this code fails.

We should instead have something like a per-window iteration over all session
buffers for all session. But this works okay for now for e.g. deno.

https://user-images.githubusercontent.com/2431823/124381676-0e817d00-dcc4-11eb-8271-1136b39835ba.mov

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -944,8 +944,6 @@ class Session(TransportCallbacks):
         else:
 
             def compare_by_string(sb: Optional[SessionBufferProtocol]) -> bool:
-                if not sb:
-                    return False
                 return sb.get_uri() == parsed if sb else False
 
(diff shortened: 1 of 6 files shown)


plugin/core/sessions.py
```diff
@@ -939,6 +944,8 @@ def compare_by_samefile(sb: Optional[SessionBufferProtocol]) -> bool:
         else:
 
             def compare_by_string(sb: Optional[SessionBufferProtocol]) -> bool:
+                if not sb:
+                    return False
```
rchl, line 948: seem redundant

PR #1780, opened 2021-07-04
Fix key errors for log messages when running tests

rwols:


rwols:
I realize now why this keeps breaking for me. UnitTesting reloads the package but in doing so breaks our invariants. In particular, we assume `WindowPanelListener.on_new_window` is always balanced with `WindowPanelListener.on_pre_close_window`. But that is not the case if you forcefully start reloading python modules.

The fix is to set

```
    "reload_package_on_testing": false,
```

in unittesting.json

rchl (approved):

rchl:
Yeah, I didn't really see the point of having that enabled myself and didn't set it in lsp_utils tests for example.

rchl:
Also we probably could get rid of `adjust_unittesting_config_for_ci.py` because using same values locally and on CI seems just fine.

rwols:
The idea is that you don't have to restart ST, but reloading python modules while there are various data structures busy in the worker thread just doesn't work that well. @randy3k

rchl:
Reloading something as complex as LSP doesn't seem viable to me.

randy3k:
> The idea is that you don't have to restart ST, but reloading python modules (in the main thread) while there are various data structures busy in the worker thread just doesn't work that well. @randy3k 

Ya, that's one part of it. Another part is that reloading is necessary to achieve 100% coverage because otherwise lines, that are executed only when the module is imported but not when the tests are running, are never covered.

PR #1782, opened 2021-07-05
Even more fixes for matching up buffers with URIs

rwols:
If the paths compare equal as strings, then we consider this a match, *even* if
the candidate file does not actually exist on-disk. This catches a case where
you delete a file on the file system but you have a view of it open in
Sublime Text. There is no callback in the API for that. And even so,
view.file_name() will still return the (now non-existing) file name.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -928,8 +928,6 @@ class Session(TransportCallbacks):
     def get_session_buffer_for_uri_async(self, uri: DocumentUri) -> Optional[SessionBufferProtocol]:
         scheme, path = parse_uri(uri)
         if scheme == "file":
-            if not os.path.exists(path):
-                return None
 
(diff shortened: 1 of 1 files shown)


plugin/core/sessions.py
```diff
@@ -938,7 +938,14 @@ def compare_by_samefile(sb: Optional[SessionBufferProtocol]) -> bool:
                 if not isinstance(candidate, str):
                     return False
                 candidate_scheme, candidate_path = parse_uri(candidate)
-                return candidate_scheme == "file" and os.path.samefile(parsed, candidate_path)
+                if candidate_scheme != "file":
+                    return False
+                if parsed == candidate_path:
```
rchl, line 943: Would rename `parsed` to `path`.

rchl (approved):

PR #1792, opened 2021-07-14
LSP-Serenata doesn't support ST4

vdupom:
As stated in https://github.com/Cloudstek/LSP-serenata/issues/1#issuecomment-851055572
Until LSP-serenata supports ST4 I think the only client for PHP is LSP-intelephense

### DIFF
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -311,15 +311,7 @@ Follow installation instructions on [LSP-promql](https://github.com/prometheus-c
 
 ## PHP
 
-There are multiple options:
-
-### Intelephense
-
 Follow installation instructions on [LSP-intelephense](https://github.com/sublimelsp/LSP-intelephense).
-
-### Serenata
-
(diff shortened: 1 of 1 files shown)


rwols:
Thanks. @mdeboer how about moving LSP-serenata to the sublimelsp organization so we can better track what's happening to the package? Once LSP upgrades to python 3.8 we would also need to upgrade all helper packages in one go.

mdeboer:
I would be happy to, I really don't have time to look into it unfortunately.

mdeboer:
Okay I can't transfer the repo as I don't have permission to create public repositories for sublimelsp. If you give me (temporary) permissions, I will transfer the repo :)

rwols:
@mdeboer invite sent.

mdeboer:
@rwols I will make some time today to get this sorted 👍🏻

mdeboer:
It has been moved 👍🏻

PR #1796, opened 2021-07-17
Remove more usage of uri_to_filename

rwols:
Doesn't work flawlessly yet as `center_selection` doesn't actually center the selection but positions the viewport such that the selection is at the top of the viewport, not in the middle of the viewport.


rwols:
Broken since 2014 :) https://github.com/sublimehq/sublime_text/issues/538

rwols:
It works now with build 4122, but there is still an animation playing, which is annoying.

htmue:
While there, what about going even further with something like this:
```python
class URI:
    __slots__ = ('_scheme', '_uri', '_path')

    @classmethod
    def from_document_uri(cls, document_uri: DocumentUri) -> 'URI':
        return cls(*parse_uri(document_uri))

    @classmethod
    def from_file_path(cls, file_path: Path) -> 'URI':
        return cls.from_document_uri(filename_to_uri(str(file_path)))

    def __init__(self, scheme: str, path: str) -> None:
        self._scheme = scheme
        self._uri = unparse_uri((scheme, path))
        self._path = Path(path).resolve() if self.is_file_uri else None

    def __hash__(self) -> int:
        return hash(self._uri)

    @property
    def document_uri(self) -> DocumentUri:
        """Normalized `DocumentUri`"""
        return self._uri

    @property
    def file_path(self) -> Optional[Path]:
        """Normalized `Path` if `scheme` is `file`"""
        return self._path

    @property
    def is_file_uri(self) -> bool:
        return self._scheme == "file"

    # ... plus workspace/project folder related methods
```
Consistently used for URIs and file paths instead of `str`, this would add better type checks, ensure normalization and put an end to thinking about `parse_uri`/`uri_to_filename`/`filename_to_uri` calls at all.

rwols:
Yeah, we should probably have an uri class like that.

PR #1798, opened 2021-07-20
Make hover popup also follow the show_diagnostics_severity_level setting

### LINKED ISSUE #1797: Add ability to hide diagnostics on hover
There is a quite big diagnostic window appears on hover. 
Screenshot from 2021-07-20 05-54-15

But there is no setting to disable/hide it. Could you kindly add it? Something like "show_hover_diagnostics_severity_level". Or can you point me to line of code where it called, so i can override it?

rchl:
This change also affects what is shown in the status bar with "show_diagnostics_in_view_status" enabled.

I don't think it would make sense to have yet another setting specifically for the hover popup and status bar.

I've tried to make test for it but we don't have ability to override or mock the global LSP settings per-test so I couldn't find a way to test it.

Fixes #1797

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -8,6 +8,7 @@ from .message_request_handler import MessageRequestHandler
 from .panels import log_server_message
 from .promise import Promise
 from .protocol import Diagnostic
+from .protocol import DiagnosticSeverity
 from .protocol import Error
 from .protocol import Location
 from .sessions import get_plugin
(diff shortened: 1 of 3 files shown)


rwols (commented):
Doesn’t this change the code action requesting behavior?

rchl:
> Doesn’t this change the code action requesting behavior?

Yes, I suppose it does.
Can add argument to that function then.

rwols (approved):

PR #1799, opened 2021-07-20
Improve keyboard shortcuts documentation

rchl:
Moved the keyboard and mouse configuration documentation to the `Customization` section and refactored.

### DIFF
--- a/docs/src/customization.md
+++ b/docs/src/customization.md
@@ -44,7 +44,7 @@ But generally you should not need to restrict your key bindings to specific scop
 
 ## Mouse map configuration
 
-If you want to bind some action to a mouse, open `Preferences / Browser Packages` from the main menu and, depending on your platform, create a sublime-mousemap file in the following location within the Packages folder:
+If you want to bind some action to a mouse, open `Preferences / Browser Packages` from the main menu and create a sublime-mousemap file in the following location within the Packages folder:
 
 | Platform | Path |
 | -------- | ---- |
--- a/docs/src/customization.md
+++ b/docs/src/customization.md
@@ -13,6 +13,59 @@ html {
 to `Packages/User/mdpopups.css`.
 See the [mdpopups documentation](http://facelessuser.github.io/sublime-markdown-popups/) for more details.
 
+## Keyboard shortcuts (key bindings)
+
+LSP's key bindings can be edited from the `Preferences: LSP Key Bindings` command from the Command Palette. Many of the default key bindings (visible in the left view) are disabled to avoid conflicts with default or user key bindings. To enable those, copy them to your user key bindings on the right, un-comment and pick the key shortcut of your choice.
+
+If you want to customize the provided key bindings (or create new ones) and make them only be active when there is a language server with a specific [LSP capability](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#initialize) (refer to the `ServerCapabilities` structure in that link) running, you can make use of the `lsp.session_with_capability` context. For example, the following example overrides `ctrl+r` to use LSP's symbol provider but only when the current view has a language server with the `documentSymbolProvider` capability and we're in a javascript or a typescript file:
+
+```js
+{
+    "command": "lsp_document_symbols",
+    "keys": [
+        "ctrl+r"
+    ],
+    "context": [
+        {
+            "key": "lsp.session_with_capability",
+            "operator": "equal",
+            "operand": "documentSymbolProvider"
+        },
+        {
+            "key": "selector",
+            "operator": "equal",
+            "operand": "source.ts, source.js"
+        }
+    ]
+},
+```
+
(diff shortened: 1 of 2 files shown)


docs/src/features.md
```diff
@@ -66,43 +66,6 @@ You can include special variables in the `command_args` array that will be autom
...
-},
-```
-
-More useful keybindings (OS-X), edit Package Settings -> LSP -> Key Bindings
-```js
-  { "keys": ["f2"], "command": "lsp_symbol_rename" },
-  { "keys": ["f12"], "command": "lsp_symbol_definition" },
-  { "keys": ["super+option+r"], "command": "lsp_document_symbols" },
-  { "keys": ["super+option+h"], "command": "lsp_hover"}
-```
```
rchl, line 104: I've removed that part completely since those examples are bad as they don't check the context. And there are better versions of those in the default key bindings.

docs/src/features.md
```diff
@@ -66,43 +66,6 @@ You can include special variables in the `command_args` array that will be autom
 | `"$position"` or `"${position}"` | object | JSON object `{ 'line': int, 'character': int }` of the (topmost) cursor position, see [Position](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#position) |
 | `"$range"` or `"${range}"` | object | JSON object with `'start'` and `'end'` positions of the (topmost) selection, see [Range](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#range) |
 
-### Overriding keybindings
-
-LSP's keybindings can be edited from the `Preferences: LSP Keybindings` command from the command palette.
-There is a special context called `lsp.session_with_capability` that can check whether there is a language server active
-with the given [LSP capability](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#initialize).
-Refer to the `ServerCapabilities` structure in that link.
-The following example overrides `ctrl+r` to use LSP's symbol provider when we're in a javascript or typescript view:
```
rchl, line 75: I'll paste here the new text since it's not clear from the diff how this has changed.
Basically made the first paragraph talk about default (disabled) key bindings and how to enable them.
The second paragraph is more advanced part about customizing key bindings (which I imagine most won't need as the provided default ones should be enough).

--- 

LSP's key bindings can be edited from the `Preferences: LSP Key Bindings` command from the Command Palette. Many of the default key bindings (visible in the left view) are disabled to avoid conflicts with default or user key bindings. To enable those, copy them to your user key bindings on the right, un-comment and pick the key shortcut of your choice.

If you want to customize the provided key bindings (or create new ones) and make them only be active when there is a language server with a specific [LSP capability](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#initialize) (refer to the `ServerCapabilities` structure in that link) running, you can make use of the `lsp.session_with_capability` context. For example, the following example overrides `ctrl+r` to use LSP's symbol provider but only when the current view has a language server with the `documentSymbolProvider` capability and we're in a javascript or a typescript file:

docs/src/features.md
```diff
@@ -117,7 +80,7 @@ You can change the default keybinding by remapping the command as below:
 ```js
 {
     "command": "auto_complete_open_link",
-    "keys": ["f12"],
+    "keys": ["f13"],
```
rchl, line 83: Since this example is about overriding the default F12 shortcut, I've changed it to something else than F12.

docs/src/features.md
```diff
@@ -127,31 +90,3 @@ You can change the default keybinding by remapping the command as below:
...
-        },
-        {
-            "key": "auto_complete_visible",
-            "operator": "equal",
-            "operand": false
-        }
-    ]
-},
-```
-There is an example of this in LSP's default keybindings.
```
rchl, line 152: This part was in a very weird place, talking about F12 potentially conflicting with default binding after the example that shown how to override default F12 key binding without conflicting with "goto definition". I've deemed that part completely unnecessary and it was removed.

docs/src/customization.md
```diff
@@ -13,6 +13,59 @@ html {
...
+[
+    {
+        "button": "button1",
+        "count": 1,
+        "modifiers": ["ctrl"],
+        "press_command": "drag_select",
+        "command": "lsp_symbol_definition",
+    }
+]
+```
```
rchl, line 67: Instead of linking to the stack overflow page, I've inlined the specified example.

docs/src/features.md
jwortmann: Maybe this example isn't really necessary at all, because it is just a default keybinding from ST and not specific to LSP? And you have already described how to define/override a keybinding in customization.md.
Btw, I think F-keys larger than F12 are macOS exclusive.

rchl: Yeah, I agree, removed.

rwols (approved):

PR #1805, opened 2021-07-22
Include the list of key bindings (shortcuts) in the documentation

rchl:
Create a table of supported key bindings / commands in the documentation.

I've also added some missing commands to the sublime-commands file (and removed some that were not working anymore). I'm a bit uncertain if we want them or not there. It looks like maybe we didn't have commands that rely on cursor position there on purpose but then there were exceptions like `Rename` or `Find References` so maybe those were just overlooked.

There are also some changes to make naming more consistent across comments in key bindings and command names.


rchl:
BTW. I'm still not too happy about the situation with shortcuts because both the `Features` section and `Customization` section lists/mentions key shortcuts now (although the list added here is more complete). I think those should be consolidated although I'm not entirely convinced how.

Are keyboard shortcuts a "feature"? Keyboard shortcuts trigger features but is "features" section really a good place for them?

On the other hand, is someone who is looking for a list of shortcuts gonna think to look into `Customization`? He might be looking for existing shortcuts rather than thinking about customizing anything...

When thinking about those issues I'm trying to put myself in the boot of someone new who doesn't know anything about that package and the current situation doesn't feel ideal then.

rchl:
OK, refactored more by moving keyboard shortcuts into its own page and then updated the `Features` page to not list keyboard shortcuts anymore (but link to the shortcuts page instead)

docs/src/features.md
```diff
@@ -1,33 +1,18 @@
```
rchl, line 1: This page should probably only have a high-level overview of the main features with screencasts. The existing content doesn't feel ideal here.

docs/src/keyboard_shortcuts.md
```diff
@@ -0,0 +1,31 @@
...
+Refer to the [Customization section](customization.md#keyboard-shortcuts-key-bindings) on how to modify or assign shortcuts to them.
+
+!!! Mac
+    If you using macOS, replace <kbd>ctrl</kbd> with <kbd>command</kbd>.
+
+| Feature | Shortcut | Command |
+| ------- | -------- | ------- |
+| Auto Complete | <kbd>ctrl</kbd> <kbd>space</kbd> (also on macOS) | `auto_complete`
+| Expand Selection | unbound | `lsp_expand_selection`
+| Find References | <kbd>shift</kbd> <kbd>f12</kbd> | `lsp_symbol_references`
```
predragnikolic, line 12: I just realized that this command overrides the default find reference command :)

docs/src/features.md
```diff
@@ -1,33 +1,18 @@
 
 ## Commands and shortcuts
 
-### Plugin commands
+Apart from the commands listed in the [Keyboard Shortcuts](keyboard_shortcuts.md) section, LSP also provides more generic commands in the Command Palette:
```
predragnikolic, line 4:
```suggestion
LSP provides the following commands in the Command Palette:
```

The beginning of the sentence is maybe unnecessary, because you also mention the keyboard shorcuts in the bullet point bellow:
`Preferences: LSP Key Bindings`: opens LSP key bindings configuration - see [Keyboard Shortcuts](keyboard_shortcuts.md)`

rchl: Yes but the point was that those listed below are not the only ones that LSP provides.
But I don't have very strong feeling about that either way.

docs/src/keyboard_shortcuts.md
rchl: I didn't even know that ST provides one.

docs/src/features.md
predragnikolic: I also do not have a preference here, it just felt a bit redundant.

docs/src/keyboard_shortcuts.md
predragnikolic: There is no need to change anything now as this keybinding was here in a long time... 🙂

I just never was aware of it.

thomaseizinger:
I recently adopted this package, coming from IntelliJ Rust and my primary quest was:

I want to reconfigure all the shortscut to be what I am used to.

What I found hard was to figure out, what sublime-lsp actually allows me to do. I didn't care much about existing shortcuts but a complete overview of what I can put into `command`, `args` and `context` to achieve a particular functionality would have been very helpful.

The table you added goes exactly into this direction, so thank you :)

docs/src/keyboard_shortcuts.md
```diff
@@ -0,0 +1,31 @@
...
+| Previous Diagnostic | <kbd>shift</kbd> <kbd>F4</kbd> | -
+| Rename | unbound | `lsp_symbol_rename`
+| Restart Server | unbound | `lsp_restart_server`
+| Run Code Action | unbound | `lsp_code_actions`
+| Run Source Action | unbound | `lsp_code_actions` (with args: `{"only_kinds": ["source"]}`)
+| Run Code Lens | unbound | `lsp_code_lens`
+| Signature Help | <kbd>ctrl</kbd> <kbd>alt</kbd> <kbd>space</kbd> | -
+| Hover Popup | unbound | `lsp_hover`
+| Toggle Diagnostics Panel | <kbd>ctrl</kbd> <kbd>alt</kbd> <kbd>m</kbd> | `lsp_show_diagnostics_panel`
+| Toggle Log Panel | unbound | `lsp_show_diagnostics_panel`
```
jwortmann, line 31: Should be `lsp_toggle_server_panel`

rwols (approved):

PR #1807, opened 2021-07-23
Remove implementation status from docs as it is only developer related

predragnikolic:
Remove implementation status from docs as it is only developer related.
In favor of a wiki page: https://github.com/sublimelsp/LSP/wiki/LSP-specification-implementation-status


predragnikolic:
If you do not agree with this PR, 
I will delete the wiki page also.

rchl (approved):
I think it's a good idea.

PR #1809, opened 2021-07-25
Add insert-replace support for completions

### LINKED ISSUE #1413: Look into InsertReplaceEdit for completions
https://github.com/microsoft/language-server-protocol/commit/7108dd15b67212b219beb715384c9eb8cea0d0c3

predragnikolic:
output

New: 
- setting for determining the `completion_insert_mode`(defaults to `"insert"`) Available: `"insert"` or `"replace"`.
- `enter`/`tab` will commit the completion item with the default insert mode.
- `shift+enter`/`shift+tab` will commit the completion item with the opposite insert mode.

I have tested this only with the LSP-angular server.

Closes #1413


jwortmann:
It's pretty cool, but at least <kbd>Shift</kbd>+<kbd>Tab</kbd> conflicts with a default keybinding when the autocomplete popup is visible (described in the ST preferences under "tab_completion"). Maybe it's better to leave the keybinding commented out by default, like most of the other keybindings? Additionally I would add another context `{"key": "lsp.session_with_capability", "operand": "completionProvider"}` to it.

An alternative could be <kbd>Ctrl</kbd>+<kbd>Tab</kbd> ("primary+tab"), which indeed has a default binding too (switch to next file tab), but not specific to when the autocompletion popup is visible. But it's not as easy to type as <kbd>Shift</kbd>+<kbd>Tab</kbd>.

The possible options "insert"/"replace" should probably also be mentioned in the comment which describes the setting.

rwols:
I was thinking of adding additional text in the st_detail field to distinguish insert/replace variants, instead of a command. The current diff isn’t very discoverable. I wonder how many people will really use this feature.

rchl:
> I was thinking of adding additional text in the st_detail field to distinguish insert/replace variants, instead of a command. The current diff isn’t very discoverable. I wonder how many people will really use this feature.

Even knowing about this feature, I probably wouldn't remember to pick the right insert mode by using a different shortcut. Especially if we don't indicate anywhere that the "replace" is supported by the given completion. Without indication, the user would never know if both the server and the given completion supports it and it would be a guessing game.

jwortmann:
I would maybe even remove the setting and always use the "insert" mode with the regular Tab/Enter binding for completions - just like it is now, and because it's the way how completions work in general in ST.

> Without indication, the user would never know if both the server and the given completion supports it and it would be a guessing game.

I agree, this should be indicated in some way. Otherwise it would be confusing why it doesn't work most of the time (probably most servers don't support it).

> I was thinking of adding additional text in the st_detail field to distinguish insert/replace variants

If I understand it correctly, that would mean to duplicate the entries which use InsertReplaceEdit? I don't like this approach a lot, because multiple entries with the same label is the thing I dislike the most in autocompletions. It requires to read more info (from the annotation hint or details field) to pick the correct one, and it breaks the typing flow if you need to scroll down first in case the order is wrong (or if you want to see the details for the other entries). So I think I would prefer a keybinding for the "replace" mode, provided that it doesn't conflict with an existing default binding.

LSP.sublime-settings
```diff
@@ -120,6 +120,10 @@
   // sorting algorithm and instead uses the sorting defined by the relevant language server.
   "inhibit_word_completions": true,
 
+  // Controls whether words are overwritten when accepting completions.
+  // Note that a language server needs to support this feature to work.
```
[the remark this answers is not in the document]
predragnikolic: >  // Note that the **"replace"** mode only works if the server supports it. Otherwise the mode fall backs to "insert".

The sentence at the end is not entirely correct. I would just replace `"replace"` with  "completion_insert_mode". and delete the second sentence.

>  // Note that the **"completion_insert_mode"** will only works if the server supports it.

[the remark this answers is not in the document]
rchl: But maybe we should think how can we simplify this option for the average joe.

You seem to be saying that there are basically three ways to insert completions - using the old TextEdit and then the "insert" and "replace" from the new `InsertReplaceEdit`. Isn't that a bit much to understand for the average user? Can we make the old TextEdit completion fall into the "insert" case? Otherwise I feel it would be very hard for the user to make a decision here without having deep understanding of the spec and implications of each behavior.

TheSecEng:
This feature also looks really slick.

predragnikolic:
I tried to mimic VS code behavior with this PR.

https://m.youtube.com/watch?v=QIze0qkugNc

In the PR description I haven't explained in detail why I implemented this the way I did
and that is simply that users switching from VS Code to ST could easily transfer their knowledge.

----
I wanted to create a second draft, that would address all the feedback from this draft and then I would let people decide between the 2 drafts.

Currently I do not think that I will do such a draft in the near time.

rchl:
In case it makes it easier, I'm gonna state now that I'm fine with how it is now if you just update the setting description. I will look once again through the changes after it's rebased and addressed.

As far as disabling key bindings by default, I'm not sure... That would make the feature very niche. I think that it would be fine to have them enabled by default as long the behavior fallbacks to ST default if completion doesn't use `InsertTextReplace`. Not sure how easy that would be though...

When it comes to marking completion details with something to indicate that insert/replace is supported, this could be done without duplicating completions. The challenge would be to figure out what to show as it might be hard to find a short text/symbol that would be easily recognizable and meaningful to everyone.

predragnikolic:
I removed shift+tab to not cause conflict with ST default behavior.

predragnikolic:
I added the following text next to the `More` link in the autcomplete popup, to make it more noticeable.

But I am still unsure if I would like to clutter the autocomplete popup with this text...  

image
> Note: I had `"completion_insert_mode": "replace"` in LSP.sublimse-settings, and thus I saw 	`"⇧ + ↵ to insert"`

Default.sublime-keymap
```diff
@@ -25,6 +25,24 @@
             }
         ]
     },
+    {
+        "command": "lsp_commit_completion_with_opposite_insert_mode",
```
rchl, line 29: There is a trailing space here.
Also I wouldn't put this key binding so high in the file. There are more relevant ones that probably deserve more attention.

plugin/completion.py
```diff
@@ -77,12 +90,21 @@ def _on_navigate(self, url: str) -> None:
         webbrowser.open(url)
 
 
+class LspCommitCompletionWithOppositeInsertMode(LspTextCommand):
+    shift_pressed = False
```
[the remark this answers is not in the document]
predragnikolic: `modifier_pressed` ? :)

plugin/core/views.py
```diff
@@ -1002,6 +1006,15 @@ def format_completion(
     return completion
 
 
+def get_insert_replace_support_html(item: CompletionItem) -> Optional[str]:
+    text_edit = item.get('textEdit')
+    if text_edit and 'insert' in text_edit and 'replace' in text_edit:
+        insert_mode = userprefs().completion_insert_mode
+        oposite_insert_mode = 'replace' if insert_mode == 'insert' else 'insert'
+        return '<code>⇧ + ↵</code> to {}'.format(oposite_insert_mode)
```
rchl, line 1014: Not ideal if someone customizes the key binding...
Unless we don't really intend people to customize it but then we should bury keybinding at the bottom of keybindings settings.

plugin/completion.py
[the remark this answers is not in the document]
rchl: Or just `active`? Since it's used as `LspCommitCompletionWithOppositeInsertMode.active`, it should be clear enough.

predragnikolic:
Should we rename
`lsp_commit_completion_with_opposite_insert_mode` to something more `short:

a) `lsp_complete_with_opposite_insert_mode`
b) `lsp_insert_replace_complete`

Default.sublime-keymap
```diff
@@ -25,6 +25,24 @@
...
+        "keys": [
+            "shift+enter"
```
jwortmann, line 31: This also conflicts with the built-in key binding to insert a newline if the completion popup is visible.

It doesn't affect me personally, because I activated the `"auto_complete_commit_on_tab"` setting, but I would expect this to disturb many users if the built-in functionality suddenly isn't available anymore. It might be acceptable to provide key bindings which are direct replacements of built-in functionality, but with an override like this one here, I doubt that the LSP package would get accepted to Package Control if it had to re-apply for it.

predragnikolic: Can you point to the conflicting keybinding?

There are of `shift+enter` keybindings in ST, 
but none with a context when auto_complete_visible is visible.
```
            {
                "key": "auto_complete_visible",
                "operator": "equal",
                "operand": true
            },
```
So I did not see it conflicting.

jwortmann: Unbenannt


The "context" only applys *restrictions* to a key binding. This one doesn't have a "context" restriction, so it is always activated. If the key binding from this PR gets added, it will be loaded after the default one, and therefore the default binding will be shadowed and won't work anymore.

Or do I have some misconception somewhere in my thoughts?

predragnikolic: Nope, you are correct. In this case the `lsp_commit_completion_with_opposite_insert_mode` keybinding will have precedence. :)

But is that bad? 
I mean other "shift+enter" ST keybinding also have contexts
and could also be seen as conflicting with this "shift+enter" insert `\n` keybinding, right?

In our case, 
the `lsp_commit_completion_with_opposite_insert_mode` will conflict only if the auto complete popup is visible and the user is in a file supported by LSP.

jwortmann: Maybe this will make it more clear:
If you start ST in safe mode (all default settings), then you can commit a completion with <kbd>Enter</kbd>. But if you instead want to add a linebreak when the completion popup is visible, you can press <kbd>Shift</kbd>+<kbd>Enter</kbd>.
Now someone installs the LSP package and suddenly and without any warning <kbd>Shift</kbd>+<kbd>Enter</kbd> doesn't work anymore if a language server is running. Because it will then still commit the completion (with the opposite LSP insert mode), instead of the usual linebreak.

> But is that bad? I mean other "shift+enter" ST keybinding also have contexts and could also be seen as conflicting with this "shift+enter" insert `\n` keybinding, right?

But the other ST key bindings are for when the command palette or find or replace panel has focus. So they are not relevant regarding auto completions.
If we start overriding built-in key bindings for different functionality, then I guess we could also hijack for example <kbd>Ctrl</kbd>+<kbd>R</kbd> for ... let's say e.g. LSP's formatting feature. And if users want to keep using ST's "Goto Symbol" they can just reassign another key binding for it :D
Or maybe let's reassing ST's key binding for `toggle_comment` to something different, because who needs that anyway... :)

> In our case, the `lsp_commit_completion_with_opposite_insert_mode` will conflict only if the auto complete popup is visible and the user is in a file supported by LSP.

Well, but this is exactly the critical point. If the auto complete popup is not visible, then the built-in <kbd>Shift</kbd>+<kbd>Enter</kbd> binding is not useful anyway, because you can just press regular <kbd>Enter</kbd> without the <kbd>Shift</kbd> key to insert a linebreak.

predragnikolic: You explained that well, thanks!
OK, I will remove the keybinding, and let the user configure it :)

predragnikolic: [c6fac89](https://github.com/sublimelsp/LSP/pull/1809/commits/c6fac89ec0d718f6f799e948efcd78a1e7e7cf8f)

predragnikolic:
Closes #1413

rchl:
> Closes #1413

I think this needs to be added to the initial comment to actually have an effect on merging.

tests/test_completion.py
```diff
@@ -603,6 +607,36 @@ def test_text_edit_plaintext_with_multiple_lines_indented(self) -> Generator[Non
...
+           'label': 'title',
+           'textEdit': {
+                'newText': 'title',
+                'insert': {'start': {'line': 0, 'character': 3}, 'end': {'line': 0, 'character': 5}},
+                'replace': {'start': {'line': 0, 'character': 3}, 'end': {'line': 0, 'character': 8}}
+            }
+        }])
+        yield from self.shift_select_completion()  # commit the opposite insert mode
+        yield from self.await_message("textDocument/completion")
+        self.assertEqual(self.read_file(), '{{ title }}')
```
rchl, line 638: Can you change `newText` to something else than the initial text to ensure that we don't pass if this just doesn't do anything?

sublime-package.json
```diff
@@ -452,6 +452,12 @@
               "default": true,
               "markdownDescription": "Disable Sublime Text's word completions. \"word\" completion means Sublime Text's internal completer that takes words from the current buffer you're editing and presents them in the auto-complete widget."
             },
+            "completion_insert_mode": {
+              "type": "string",
+              "default": "insert",
+              "enum": ["insert", "replace"],
+              "markdownDescription": "Controls whether words are overwritten when accepting completions.\nNote that a language server needs to support this feature to work."
```
rchl, line 459: Maybe:

> For servers that support it, decides the primary mode used for inserting completions. For example, when triggering completion in a middle of a word, the `insert` mode would insert the completion text in the middle of the word while the `replace` mode would replace the existing word with a new completion text.\nAn LSP keybinding `lsp_commit_completion_with_opposite_insert_mode` can be enabled to insert completion using the opposite mode to the one selected here.

LSP.sublime-settings
```diff
@@ -163,6 +163,10 @@
...
+  "completion_insert_mode": "insert",
```
rchl, line 168: Sync description.

sublime-package.json
```diff
@@ -452,6 +452,12 @@
               "default": true,
               "markdownDescription": "Disable Sublime Text's word completions. \"word\" completion means Sublime Text's internal completer that takes words from the current buffer you're editing and presents them in the auto-complete widget."
             },
+            "completion_insert_mode": {
+              "type": "string",
+              "default": "insert",
+              "enum": ["insert", "replace"],
+              "markdownDescription": "For servers that support it, chooses the primary mode used for inserting completions. For example, when triggering completion in a middle of a word, insert would insert the completion text in a middle of the word, while replace would replace the existing word with a new completion text.\nAn LSP keybinding lsp_commit_completion_with_opposite_insert_mode can be used to insert completion using the opposite mode to the one selected here."
```
rchl, line 459: My suggestion also included some backticks that are lost here.

LSP.sublime-settings
```diff
@@ -163,8 +163,9 @@
   // sorting algorithm and instead uses the sorting defined by the relevant language server.
   "inhibit_word_completions": true,
 
-  // Controls whether words are overwritten when accepting completions.
-  // Note that a language server needs to support this feature to work.
+  // For servers that support it, chooses the primary mode used for inserting completions.
+  // For example, when triggering completion in a middle of a word, `insert` would insert the completion text in a middle of the word, while `replace` would replace the existing word with a new completion text.
+  // An LSP keybinding `lsp_commit_completion_with_opposite_insert_mode` can be used to insert completion using the opposite mode to the one selected here.
```
rchl, line 168: Very long lines. Can you make them more close to convention in this file?

predragnikolic: Will try 🙂

LSP.sublime-settings
```diff
@@ -163,9 +163,12 @@
...
 
-  // For servers that support it, chooses the primary mode used for inserting completions.
-  // For example, when triggering completion in a middle of a word, `insert` would insert the completion text in a middle of the word, while `replace` would replace the existing word with a new completion text.
-  // An LSP keybinding `lsp_commit_completion_with_opposite_insert_mode` can be used to insert completion using the opposite mode to the one selected here.
+  // The mode used for inserting completions:
+  // - `insert` would insert the completion text in a middle of the word
+  // - `replace` would replace the existing word with a new completion text
+  // An LSP keybinding `lsp_commit_completion_with_opposite_insert_mode`
+  // can be used to insert completion using the opposite mode to the one selected here.
+  // Note: Must be supported by the language server.
```
rchl, line 171: Sorry if I've confused you but I was happy with the text (I came up with it after all ;)). It was just the length of the lines that I had problem with. Just needed some more line breaks.

rchl: But it's fine now also.

predragnikolic: Ah, ok I missinterpreted what I had to do.
But if you are ok with this, I will not revert the commit.

rchl (approved):
Given that it's a functionality that is not bound to a key binding by default and even when bound the discoverability is low (not clear when "replace" can be used), I find it a bit hard to justify.

But who am I judge, maybe someone will find it useful.
Let's merge after one more person gives an approval.

predragnikolic:
How about `"alt+enter"`?

rchl:
Seems fine to me. Someone might still be using it to insert a line break but since shift+enter works it might not be a problem.

rchl:
I'm still thinking how we could visualize if completion supports insert/replace...
- It shouldn't mention specific key binding as we can't tell which one is defined.
- It shouldn't take too much space.

Some ideas:
 - add `[I/R]` in the details field -- a bit too cryptic perhaps?
 - say "(supports replace)" if "insert" is the default mode and "(supports insert)" if "replace" is the default mode -- a bit long?
 - ???

predragnikolic:
How about the following?

Provide a default keybinding "alt+enter", 
"Replace" or "Insert" will be show at the bottom of the AC popup. So that can give an indication that the server supports insert/replace.

output1

plugin/completion.py
```diff
@@ -93,7 +93,7 @@ def _on_navigate(self, url: str) -> None:
 class LspCommitCompletionWithOppositeInsertMode(LspTextCommand):
     active = False
 
-    def run(self, edit: sublime.Edit) -> None:
+    def run(self, edit: sublime.Edit, event=None) -> None:
```
rchl, line 96:
```suggestion
    def run(self, edit: sublime.Edit, event: Any = None) -> None:
```

plugin/core/views.py
```diff
@@ -1019,3 +1023,13 @@ def format_code_actions_for_quick_panel(
         if code_action.get('isPreferred', False):
             selected_index = idx
     return items, selected_index
+
+
+def get_insert_replace_support_html(item: CompletionItem) -> Optional[str]:
+    text_edit = item.get('textEdit')
+    if text_edit and 'insert' in text_edit and 'replace' in text_edit:
```
rchl, line 1030: Sorry, don't have time to check the spec now, but is it necessary to check for `textEdit`?

rchl (commented):
I like the idea.

plugin/core/views.py
predragnikolic: yes, there is a type error otherwise :)

plugin/completion.py
predragnikolic: pushed a fix for this, but thanks :)

docs/src/features.md
```diff
@@ -15,6 +15,11 @@ The LSP package enhances the auto-complete list with results provided by the lan
 To show the documentation popup you can click the **More** link in the bottom of the autocomplete,
 or you can use the default sublime keybinding <kbd>F12</kbd> to trigger it.
 
+To inser or replace a completion item using the opposite "completion_insert_mode" setting value, the following keybinding can be used <kbd>alt+enter</kbd>.
```
predragnikolic, line 18: typo `inser`

rchl (approved):

PR #1813, opened 2021-07-31
Re-evaluate what langservers to attach when the URI scheme changes

rwols:
This fixes a regression where previously a langserver would attach to a scratch
buffer once it was saved to disk. That functionality got lost in the URI
refactorings.

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -380,8 +380,8 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         # Re-determine the URI; this time it's guaranteed to be a file because ST can only save files to a real
         # filesystem.
         uri = view_to_uri(self.view)
-        new_scheme = parse_uri(uri)[0]
-        old_scheme = parse_uri(self._uri)[0]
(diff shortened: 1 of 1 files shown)


plugin/documents.py
```diff
@@ -378,10 +379,20 @@ def on_selection_modified_async(self) -> None:
     def on_post_save_async(self) -> None:
         # Re-determine the URI; this time it's guaranteed to be a file because ST can only save files to a real
         # filesystem.
-        self.set_uri(view_to_uri(self.view))
-        if self.view.is_primary():
-            for sv in self.session_views_async():
-                sv.on_post_save_async(self._uri)
+        uri = view_to_uri(self.view)
+        new_scheme = parse_uri(uri)[0]
+        old_scheme = parse_uri(self._uri)[0]
```
rchl, line 384: Something like this would look nicer (to me at least):
```suggestion
        new_scheme, _ = parse_uri(uri)
        old_scheme, _ = parse_uri(self._uri)
```

rchl (approved):

PR #1814, opened 2021-07-31
Remember the cwd for plugins

### LINKED ISSUE #1788: Could the result from AbstractPlugin.on_pre_start be stored in the session?
### Is your feature request related to a problem? Please describe.

The Julia server operates based on a Julia environment (like Python virtual environments) and it can use the working directory where it was started in, to determine the currently active environment. If no environement can be found, it falls back to a default environment.

Using the [LSP-julia](https://github.com/sublimelsp/LSP-julia) helper package, I would like to show the active environment in the status bar after the server initializes, and I currently have a workaround for that in https://github.com/sublimelsp/LSP-julia/blob/96dde4a3443f8bf45c45e9aa4e6afc92014e50e0/plugin.py#L100-L106, but it doesn't work in all cases (for example single file mode), because there is not enough information available in `AbstractPlugin.__init__`. I need access to the result from `on_pre_start`, but I can't save that result into an instance variable from a classmethod.

### Describe the solution you'd like

If the result from `on_pre_start` is stored in the session, it can be retrieved in `__init__`. I would need the exact result from `on_pre_start`, not the actual working directory (in case `None` is returned and there exists also a workspace folder). Maybe this is too specific for my usecase to be useful in general?

### Describe alternatives you've considered

(a) If the initiating view was also given to `__init__`, then I could alternatively execute the same logic from `on_pre_start` again.

(b) If all the methods from AbstractPlugin were instance methods instead of classmethods, and what is currently `__init__` would be another instance method `on_initialize`, then data could be exchanged between methods via instance variables.

(c) I also considered to save the result as a "hidden" setting to the view via `initiating_view.settings().set("_lsp_initial_environment", result_value)`, before returning from the `on_pre_start` function. And then in `__init__` I can iterate through all views to find the initiating view and retrieve the value:
<details>
  <summary>Click to expand</summary>

  ```python
      def __init__(self, weaksession) -> None:
          super().__init__(weaksession)
          if sublime.load_settings(SETTINGS_FILE).get("show_environment_status"):
              sublime.set_timeout_async(self.set_initial_environment_status_async)

      def set_initial_environment_status_async(self) -> None:
          session = self.weaksession()
          if session:
              for session_view in session.session_views_async():
                  env_path = session_view.view.settings().get("_lsp_initial_environment", False)
                  if env_path is not False:
                      env_name = os.path.basename(env_path) if env_path else JuliaLanguageServer.default_julia_environment()
                      session.set_window_status_async(STATUS_BAR_KEY, "Julia env: {}".format(env_name))
                      return
  ```
</details>

But it doesn't work; `session.session_views_async()` doesn't return any views there.

Perhaps someone sees why it isn't working or has an idea for a better solution?

rwols:
Resolves #1788

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -870,6 +870,9 @@ class Session(TransportCallbacks):
     def get_workspace_folders(self) -> List[WorkspaceFolder]:
         return self._workspace_folders
 
+    def get_working_directory(self) -> Optional[str]:
+        return self.transport.get_working_directory() if self.transport else None
+
     def uses_plugin(self) -> bool:
         return self._plugin is not None
 
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -832,6 +832,7 @@ class Session(TransportCallbacks):
     def __init__(self, manager: Manager, logger: Logger, workspace_folders: List[WorkspaceFolder],
                  config: ClientConfig, plugin_class: Optional[Type[AbstractPlugin]]) -> None:
         self.transport = None  # type: Optional[Transport]
+        self.working_directory = None  # type: Optional[str]
         self.request_id = 0  # Our request IDs are always integers.
         self._logger = logger
         self._response_handlers = {}  # type: Dict[int, Tuple[Request, Callable, Optional[Callable[[Any], None]]]]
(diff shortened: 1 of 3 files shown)


rwols:
@randy3k for some reason the macOS workflow cannot install package control?

jwortmann:
Thanks! Despite what I wrote in the linked issue, I think it should be possible for my usecase to derive the desired environment even with the actual working directory, i.e. `workspace_folders[0].path` or `os.getcwd()` if the `on_pre_start` result is `None` (the only thing I have to do is to check whether a corresponding metadata file exists in the directory). So if you think that it would be more useful to store the actual working directory, I guess that should be fine too.

rwols:
> So if you think that it would be more useful to store the actual working directory, I guess that should be fine too.

I think the current diff will do fine :)

rchl (approved):

PR #1815, opened 2021-07-31
Allow lists in `env` dict of client configuration

### LINKED ISSUE #1811: env handling in client settings not portable
There is no way to alter the env paths in a way that would work across all systems.

For example `LSP-pylsp` defines some env variables (https://github.com/sublimelsp/LSP-pylsp/blob/75a07c497fcca88cb2780230e559ae0e7402c361/LSP-pylsp.sublime-settings#L37-L37) that are PATH-like and are using path separator `:` but that is obviously not correct on Windows.

The only way to handle that comes to my mind right now is to have a template variable for the separator. The only small downside would be that it would not be easily discoverable.

rwols:
The lists are joined into a string with the native file system's path
separator.

Resolves #1811

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -553,7 +553,7 @@ class ClientConfig:
                  enabled: bool = True,
                  init_options: DottedDict = DottedDict(),
                  settings: DottedDict = DottedDict(),
-                 env: Dict[str, str] = {},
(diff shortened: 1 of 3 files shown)


rchl:
Also update schema.

rchl (approved):

PR #1816, opened 2021-07-31
Discard diagnostics in a few cases

### LINKED ISSUE #1727: Do not render diagnostics in certain cases

There are certain files for which I don't want to see diagnostics:

- Any file that matches `index_exclude_patterns`
- Any file that matches `binary_file_patterns`
- Files outside the workspace (e.g., when jumping to a standard-library file)

Don't show diagnostics for files that match one of the criteria above.

Disable diagnostics globally.

**Additional context**
One major problem is that certain language servers are "projectless" servers: vscode-json-language-server doesn't care about project structure, and so you can open it with any file. We have to distinguish these two kinds of servers somehow.

rwols:
- When the file is outside the window folders, ignore diagnostics
- When the file matches binary_exclude_patterns, ignore it
- When the file matches file_exclude_patterns, ignore it
- When the file matches folder_exclude_patterns, ignore it

This doesn't take care of the index_exclude_patterns and
index_exclude_gitignore settings.

Resolves #1727

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1283,16 +1283,13 @@ class Session(TransportCallbacks):
         uri = params["uri"]
         mgr = self.manager()
         if not mgr:
-            # debug("ignoring diagnostics for", uri, "due to missing window manager")
             return
         reason = mgr.should_present_diagnostics(uri)
         if isinstance(reason, str):
             return debug("ignoring unsuitable diagnostics for", uri, "reason:", reason)
         sb = self.get_session_buffer_for_uri_async(uri)
-        if not sb:
-            # debug("ignoring diagnostics for", uri, "due to missing session buffer")
-            return
-        sb.on_diagnostics_async(params["diagnostics"], params.get("version"))
+        if sb:
+            sb.on_diagnostics_async(params["diagnostics"], params.get("version"))
 
     def m_client_registerCapability(self, params: Any, request_id: Any) -> None:
         """handles the client/registerCapability request"""
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -95,6 +95,12 @@ class Manager(metaclass=ABCMeta):
         """
         pass
 
+    @abstractmethod
+    def should_present_diagnostics(self, uri: DocumentUri) -> Optional[str]:
+        """
+        Should the diagnostics for this URI be shown in the view? Return a reason why not
+        """
+
     # Mutators
 
     @abstractmethod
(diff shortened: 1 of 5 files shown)


plugin/core/windows.py
```diff
@@ -437,6 +440,34 @@ def get_project_path(self, file_path: str) -> Optional[str]:
...
+
+    def _matches_glob(self, path: str, patterns: Any) -> bool:
+        if not isinstance(patterns, list):
+            return False
+        for pattern in patterns:
+            if not isinstance(pattern, str):
+                continue
+            if fnmatch.fnmatch(path, pattern):
+                return True
+        return False
```
rchl, line 469: - `fnmatch` is not really a 1:1 implementation of the ST pattern matching. Are we fine with that?
- We could have it as a global utility function since I will need it for file watcher also.
- Those patterns are not "glob"'s technically so the name is a bit misleading.

plugin/core/sessions.py
```diff
@@ -1275,9 +1281,18 @@ def m_workspace_applyEdit(self, params: Any, request_id: Any) -> None:
...
+            # debug("ignoring diagnostics for", uri, "due to missing window manager")
+            return
+        reason = mgr.should_present_diagnostics(uri)
+        if isinstance(reason, str):
+            return debug("ignoring unsuitable diagnostics for", uri, "reason:", reason)
         sb = self.get_session_buffer_for_uri_async(uri)
-        if sb:
-            sb.on_diagnostics_async(params["diagnostics"], params.get("version"))
+        if not sb:
+            # debug("ignoring diagnostics for", uri, "due to missing session buffer")
```
rchl, line 1293: Remove those commented out debug logs maybe?

plugin/core/windows.py
```diff
@@ -437,6 +440,34 @@ def get_project_path(self, file_path: str) -> Optional[str]:
...
+        scheme, path = parse_uri(uri)
+        if scheme != "file":
+            return None
+        if not self._workspace.contains(path):
+            return "not inside window folders"
+        view = self._window.active_view()
+        if not view:
+            return None
+        settings = view.settings()
+        if self._matches_glob(path, settings.get("binary_file_patterns")):
```
rchl, line 453: I have this in my project:

Screenshot 2021-08-01 at 10 24 10

and this are the view settings (the view is within the workspace):

```
>>> view.settings().get('folder_exclude_patterns')
['.svn', '.git', '.hg', 'CVS', '.mypy_cache', 'node_modules']
```

It looks like the `.nuxt` is missing (`node_modules` comes from global override). So are the view settings not actually updated with project overrides?

rwols: Hmm, I can reproduce this behavior here as well. Seems broken to me.

plugin/core/windows.py
rwols: > fnmatch is not really a 1:1 implementation of the ST pattern matching. Are we fine with that?

For now I would say yes. Seeing as we can't really filter `index_exclude_patterns` correctly anyway due to the `index_exclude_gitignore` setting adding a lot of complexity.

> Those patterns are not "glob"'s technically so the name is a bit misleading.

I thought they were... just lacking features.

rchl: > I thought they were... just lacking features.

I suppose you are right. Though typically `glob` refers to patterns that support `**` but I guess it's not a rule.

rchl: > > fnmatch is not really a 1:1 implementation of the ST pattern matching. Are we fine with that?
> 
> For now I would say yes. Seeing as we can't really filter `index_exclude_patterns` correctly anyway due to the `index_exclude_gitignore` setting adding a lot of complexity.

But the difference in parsing `/` could make this solution very error with some patterns, potentially.

rwols: > We could have it as a global utility function since I will need it for file watcher a

Made it into a public function.

plugin/core/windows.py
rwols: https://github.com/sublimehq/sublime_text/issues/4717

rwols:
I'll go ahead and merge this as this has some benefit already. The index_exclude_patterns and project-specific exclude patterns are really more of an ST issue.

PR #1823, opened 2021-08-14
Pass ignore patterns obtained from ST to the file watcher

rchl:


plugin/core/sessions.py
```diff
@@ -1104,6 +1105,17 @@ def _handle_initialize_error(self, result: Any) -> None:
...
+    def _get_global_ignore_globs(self, root_path: str) -> List[str]:
+        folder_excludes = [
+            sublime_pattern_to_glob(pattern, is_directory_pattern=True, root_path=root_path)
+            for pattern in globalprefs().get('folder_exclude_patterns')
+        ]
+        file_excludes = [
+            sublime_pattern_to_glob(pattern, is_directory_pattern=False, root_path=root_path)
+            for pattern in globalprefs().get('file_exclude_patterns')
+        ]
+        return folder_excludes + file_excludes + ['**/node_modules/**']
```
rchl, line 1117: I guess it's kinda weird that I'm forcing `node_modules` to be ignored but it's kinda a "right" thing to do as nothing should need to watch it and there would be a ton of events triggers on running "npm i" or similar.

plugin/core/sessions.py
```diff
@@ -1091,8 +1092,8 @@ def _handle_initialize_success(self, result: Any) -> None:
             pattern = config.get('pattern')
             if pattern:
                 events = config.get('events') or ['create', 'change', 'delete']
-                ignores = config.get('ignores') or DEFAULT_IGNORES
                 for folder in self.get_workspace_folders():
+                    ignores = config.get('ignores') or self._get_global_ignore_globs(folder.path)
```
rchl, line 1096: I'm not sure how to exactly handle the situation when the `ignores` is specified in the config. Whether all of those should be combined or not. If we combine then there will be no way to opt-out of the excluded patterns.

predragnikolic: I would leave as is. (I would just add a comment above with the explanation why they are not combined)

plugin/core/types.py
```diff
@@ -71,6 +72,37 @@ def matches_pattern(path: str, patterns: Any) -> bool:
     return False
 
 
+def sublime_pattern_to_glob(pattern: str, is_directory_pattern: bool, root_path: Optional[str] = None) -> str:
```
predragnikolic, line 75: Do you see a benefit to split this function into two separate functions? (to remove the is_directory_pattern flag)
- sublime_pattern_to_folder_glob
- sublime_pattern_to_file_glob

rchl: That would result in duplication of logic.

predragnikolic (approved):

plugin/core/types.py
```diff
@@ -71,6 +72,37 @@ def matches_pattern(path: str, patterns: Any) -> bool:
...
+    """
+    Convert a Sublime Text pattern (http://www.sublimetext.com/docs/file_patterns.html)
+    to a glob pattern that utilizes globstar extension.
+    """
```
rwols, line 79: Should this be used in windows.py as well?

rwols (approved):

plugin/core/types.py
rchl: How would you use it there? Convert to glob and then how match the glob to file path?

PR #1824, opened 2021-08-16
Always clear undo stack after a mutation on any panel

rwols:
This prevents memory growing infinitely due to Sublime Text maintaining an
infinite undo stack, even for panels.

### DIFF
--- a/plugin/core/panels.py
+++ b/plugin/core/panels.py
@@ -37,9 +37,19 @@ class PanelName:
 
 @contextmanager
 def mutable(view: sublime.View) -> Generator:
+    # Unset the read-only flag temporarily
     view.set_read_only(False)
-    yield
-    view.set_read_only(True)
+    try:
+        yield
+    finally:
+        # Compatible with v4113 and below
+        # todo: call directly some day
+        clear_undo_stack = getattr(view, "clear_undo_stack", None)
+        if callable(clear_undo_stack):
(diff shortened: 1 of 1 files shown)


plugin/core/panels.py
```diff
@@ -37,9 +37,19 @@ class PanelName:
 
 @contextmanager
 def mutable(view: sublime.View) -> Generator:
+    # Unset the read-only flag temporarily
     view.set_read_only(False)
-    yield
-    view.set_read_only(True)
+    try:
+        yield
+    finally:
```
rchl, line 44: Why within an exception handler now? Do you of any case when this could happen?

rchl (approved):

plugin/core/panels.py
rchl: yes but I'm not sure why would we get exceptions triggered within "with" and whether we want to hide them like that

rwols: the exception is not consumed, only the `finally` block is executed to do cleanup and the exception will then keep going up the call stack.

jfcherng:
https://github.com/sublimelsp/LSP/blob/c3075e979b3ced89ba2a4a4417f1f101157eaaf2/plugin/core/panels.py#L45-L50
Unfortunately, the console says `Warning: Can't clear undo stack from text command`.

PR #1827, opened 2021-08-17
Implement labelDetailsSupport

### LINKED ISSUE #1570: Look into detailedLabelSupport
The upcoming 3.17 spec allows servers to provide a more detailed completion label:

```typescript
export interface CompletionItemLabel {
	/**
	 * The name of a function or variable.
	 */
	name: string;

	/**
	 * The parameters without the return type.
	 */
	parameters?: string;

	/**
	 * The fully qualified name, like package name or file path.
	 */
	qualifier?: string;

	/**
	 * The return-type of a function or type of a property/variable.
	 */
	type?: string;
}
```

Because ST4 allows us to use a limited form of minihtml in the `details` slot, it should be possible to render this additional information with some coloring.

N/A

**Additional context**
See the upcoming 3.17 spec: https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/

rwols:
There isn't actually any language server that I know of that implements labelDetails. So I've written extensive tests. The most likely candidates that will are jdtls and omnisharp.

close #1570

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -802,33 +802,27 @@ def format_completion(
     st_details = ""
     if can_resolve_completion_items or item.get("documentation"):
         st_details += make_command_link("lsp_resolve_docs", "More", {"index": index, "session_name": session_name})
-
-    if lsp_filter_text and lsp_filter_text != lsp_label:
-        st_trigger = lsp_filter_text
-        if st_details:
-            st_details += " | "
-        if lsp_label_details:
-            lsp_label_detail = lsp_label_details.get("detail")
-            lsp_label_description = lsp_label_details.get("description")
-            st_details += "<p>"
-            # `label` should be rendered most prominent.
-            st_details += _wrap_in_tags("b", lsp_label)
-            if isinstance(lsp_label_detail, str):
-                # `detail` should be rendered less prominent than `label`.
-                # The string is appended directly after `label`, with no additional white space applied.
-                st_details += html.escape(lsp_label_detail)
-            if isinstance(lsp_label_description, str):
-                # `description` should be rendered less prominent than `detail`.
-                # Additional separation is added.
-                st_details += " - {}".format(_wrap_in_tags("i", lsp_label_description))
-            st_details += "</p>"
-        else:
-            st_details += _wrap_in_tags("p", lsp_label)
+        st_details += " | "
+    if lsp_label_details:
+        lsp_label_detail = lsp_label_details.get("detail")
+        lsp_label_description = lsp_label_details.get("description")
+        st_details += "<p>"
+        # `label` should be rendered most prominent.
+        st_details += _wrap_in_tags("b", lsp_label)
+        if isinstance(lsp_label_detail, str):
+            # `detail` should be rendered less prominent than `label`.
+            # The string is appended directly after `label`, with no additional white space applied.
+            st_details += html.escape(lsp_label_detail)
+        if isinstance(lsp_label_description, str):
+            # `description` should be rendered less prominent than `detail`.
+            # Additional separation is added.
+            st_details += " - {}".format(_wrap_in_tags("i", lsp_label_description))
+        st_details += "</p>"
     else:
-        st_trigger = lsp_label
+        st_details += _wrap_in_tags("p", lsp_label)
 
     completion = sublime.CompletionItem.command_completion(
-        trigger=st_trigger,
+        trigger=lsp_filter_text if lsp_filter_text else lsp_label,
         command="lsp_select_completion_item",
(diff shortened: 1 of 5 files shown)


plugin/core/views.py
```diff
@@ -802,7 +809,23 @@ def format_completion(
...
+            st_details += _wrap_in_tags(
+                "p",
+                "{label}{detail}{description}".format(
+                    # `label` should be rendered most prominent.
+                    label=_wrap_in_tags("b", lsp_label, escape=True),
+                    # `detail` should be rendered less prominent than `label`.
+                    detail=html.escape(lsp_label_detail) if isinstance(lsp_label_detail, str) else "",
+                    # `description` should be rendered less prominent than `detail`.
+                    description=_wrap_in_tags("i", lsp_label_description, escape=True)
+                ),
```
rchl, line 824: Might be hard to argue about it now, since there are now servers supporting it but the `description` most likely needs to have to be separated with a space at least from the `detail`.

I imagine the possible values could be:

label = "trim"
detail = "(str)"
description = "Trims a string"

in which without a space it would look like:

```
trim(str)Trims a string
```

Should probably use `|` or `-` to separate it.

rchl (changes requested):

plugin/core/views.py
rwols: You're probably right. For the `detail` it says there should not be spacing applied. For the `description` it says nothing about space separation. https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-17.md#textDocument_completion

rchl: Yes.
And I see that in tests you have assumed that the field itself would have a space but that's not how I think it would work. The spec itself shouldn't concern itself with presentation (aside from giving suggestions).

plugin/core/views.py
```diff
@@ -802,7 +807,23 @@ def format_completion(
...
+                # `detail` should be rendered less prominent than `label`.
+                # The string is appended directly after `label`, with no additional white space applied.
+                st_details += html.escape(lsp_label_detail)
+            if isinstance(lsp_label_description, str):
+                # `description` should be rendered less prominent than `detail`.
+                # Additional separation is added.
+                st_details += " - {}".format(_wrap_in_tags("i", lsp_label_description))
+            st_details += "</p>"
+        else:
+            st_details += _wrap_in_tags("p", lsp_label)
```
rchl, line 826: One more thing... This logic is only executed in a quite specific case when "lsp_filter_text and lsp_filter_text != lsp_label". I think that this should be supported even when there is no filter text or filter text is not different than label.

plugin/core/views.py
rwols: fixed in a8a4425

plugin/core/views.py
rwols: It looks like we always want to put it into the st_detail slot. The st_annotation slot is reserved for lsp_details.

If there is no lsp_details information, but there are labelDetails, should we put the label details in the st_annotation? I could imagine this to be somewhat confusing if some completions do have an lsp_details and some don't. But, I'm open to any kind of improvement.

plugin/core/views.py
```diff
@@ -791,23 +795,34 @@ def format_completion(
...
+            # The string is appended directly after `label`, with no additional white space applied.
+            st_details += html.escape(lsp_label_detail)
+        if isinstance(lsp_label_description, str):
+            # `description` should be rendered less prominent than `detail`.
+            # Additional separation is added.
+            st_details += " - {}".format(_wrap_in_tags("i", lsp_label_description))
+        st_details += "</p>"
     else:
-        st_trigger = lsp_label
+        st_details += _wrap_in_tags("p", lsp_label)
```
rchl, line 822: Always putting the `label` in `st_details`, even if it's no different than `lsp_filter_text` is not that great of an idea. It adds unnecessary duplication:

Screenshot 2021-08-17 at 21 55 53

plugin/core/views.py
```diff
@@ -791,23 +795,34 @@ def format_completion(
...
 
     completion = sublime.CompletionItem.command_completion(
-        trigger=st_trigger,
+        trigger=lsp_filter_text if lsp_filter_text else lsp_label,
```
rchl, line 825: could be: `lsp_filter_text or lsp_label`

plugin/core/views.py
rchl: In short, I think it's fine as it is now (minus the issue below).

> If there is no lsp_details information, but there are labelDetails, should we put the label details in the st_annotation?

With `st_annotation` we can't avoid the spacing between the label and `labelDetails.detail` so I think that is not a good fit.

rchl (approved):

PR #1828, opened 2021-08-18
Take care of InsertTextMode for completions

rwols:
This advertises that we only support "adjust indentation"-style completions.

When a completion contains a newline, say `a\nb`, then the question becomes whether the language server should account for the current indentation the cursor is at, or whether the editor should take care of that. For this package the editor takes care of that.

While adding this property in the client capabilities I discovered that we actually had an inconsistency with this. In the case of a plaintext textEdit, we did in fact not account for indentation level. This is now fixed as well in 0b1d6d3. The two commits can be rebased as-is.


rwols:
See also: https://github.com/microsoft/language-server-protocol/issues/83

rchl (approved):

PR #1829, opened 2021-08-18
Update client configuration docs

rwols:
- Instead of a typescript example I used phpactor as example. This should minimize confusion as to whether you're supposed to write this yourself or whether a helper package "provides"/"registers" a config.
- Update client config keys
- Explain transport modes better


rchl (approved):

rchl:
I wonder if it's the fact that those check are required to pass in settings makes them hang here...

PR #1832, opened 2021-08-21
Add AbstractPlugin.on_before_formatting API for tweaking formatting options

rchl:
Not entirely convinced on naming of things, of course.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -648,17 +648,16 @@ class AbstractPlugin(metaclass=ABCMeta):
         """
         pass
 
-    def on_before_formatting(self, view: sublime.View, formatting_options: Dict[str, Any]) -> Dict[str, Any]:
+    def additional_formatting_options(self, view: sublime.View) -> Dict[str, Any]:
         """
-        Called when a language server is about to perform a document or selection formatting.
+        Called when a language server is about to perform a document or range formatting. The returned formatting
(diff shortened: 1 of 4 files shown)


plugin/core/sessions.py
```diff
@@ -647,6 +648,18 @@ def on_pre_send_notification_async(self, notification: Notification) -> None:
...
+        """
+        Called when a language server is about to perform a document or selection formatting.
+
+        :param view: The view on which formatting is about to be performed.
+        :param formatting_options: The initial formatting options. The object can be modified and returned or a new
+                                   object can be returned.
+
+        :returns: The object with formatting options.
+        """
+        return formatting_options
```
rwols, line 661: Can we:

1. Rename it to `on_formatting_options`
2. Return `None` always and let the overrider modify the dict in place? My guess is anyone overriding this will modify the passed-in dict anyway, and then return that. But then the dict is modified.

rwols: Actually it may be better to call it `additional_formatting_options(self, view) -> Dict[str, Any]` akin to `additional_variables()`. We would then use it as `options.update(self._plugin.additional_formatting_options(view))`

rchl: I was going back and forth on both of those points myself. Happy to align with whatever you think is better as I couldn't convince myself.

rwols (approved):

PR #1834, opened 2021-08-30
Fix LSP boot issue

husanjun:
If other plugins are updating when the LSP is started, an error will be thrown, so make sure the plugin exists

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -693,9 +693,10 @@ def _register_plugin_impl(plugin: Type[AbstractPlugin], notify_listener: bool) -
         return
     try:
         settings, base_file = plugin.configuration()
-        if client_configs.add_external_config(name, settings, base_file, notify_listener):
-            on_change = functools.partial(client_configs.update_external_config, name, settings, base_file)
-            _plugins[name] = (plugin, SettingsRegistration(settings, on_change))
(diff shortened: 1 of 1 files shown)


plugin/core/sessions.py
```diff
@@ -693,9 +693,10 @@ def _register_plugin_impl(plugin: Type[AbstractPlugin], notify_listener: bool) -
         return
     try:
         settings, base_file = plugin.configuration()
-        if client_configs.add_external_config(name, settings, base_file, notify_listener):
-            on_change = functools.partial(client_configs.update_external_config, name, settings, base_file)
-            _plugins[name] = (plugin, SettingsRegistration(settings, on_change))
+        if base_file in sublime.find_resources('{}.*'.format(name)):
```
rchl, line 696: Would we still want to trigger some descriptive exception in that case instead of silently failing to register the plugin?

Also, could be more specific with the pattern: `sublime.find_resources(os.path.basename(base_file))`

husanjun: There are no pop-ups, just errors in the console

rchl: Still, the error in the console is better than silent ignore.
This can can also happen during development that the specified path is invalid.

husanjun: Well, that's not necessary

PR #1842, opened 2021-09-09
Docs add plugins

predragnikolic:
These plugins are available on PC but were not documented.

### DIFF
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -309,6 +309,10 @@ Follow installation instructions on [LSP-TexLab](https://github.com/sublimelsp/L
     }
     ```
 
+## Leo
+
+Follow installation instructions on [LSP-leo](https://github.com/sublimelsp/LSP-leo).
+
 ## Lua
 
 Follow installation instructions on [LSP-lua](https://github.com/sublimelsp/LSP-lua).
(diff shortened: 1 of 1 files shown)


predragnikolic:
Because I updated only the documentation, Github Actions will not be run. (so ignore the waiting status)

docs/src/language_servers.md
```diff
@@ -290,8 +294,16 @@ Follow installation instructions on [LSP-json](https://github.com/sublimelsp/LSP
 
 ## LaTeX
 
+There are multiple options:
+
+### TexLab
+
 Follow installation instructions on [LSP-TexLab](https://github.com/sublimelsp/LSP-TexLab).
 
+### LTeX
```
rwols, line 303: "LTeX" is not a language, no? We seem to be ordering these sections on language, not language server?

rwols (changes requested):
Serenata doesn't work for ST4. So you should revert 8483633.

See https://github.com/sublimelsp/LSP/pull/1792

docs/src/language_servers.md
predragnikolic: "LTeX" is the name of the language server.

When there is "There are multiple options" I put the name of the language server for each option.

rwols: But the other sections are (mainly) languages. The exception seems to be ESLint. LTeX-LS serves latex and markdown.

jwortmann: It's still under the LaTeX language section. This heading with the language server name is one level smaller, similar to the language server sections under Python or Ruby for example.

PR #1845, opened 2021-09-12
Send hover request to all sessions that supports hoverProvider

predragnikolic:
closes #https://github.com/sublimelsp/LSP-tailwindcss/issues/19

Currently LSP sends only one hover request, which doesn't work nicely when there are multiple sessions running in a file.
LSP should send hover requests to all sessions that supports the `hoverProvider`.(like we do when requesting completions)

### DIFF
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -212,6 +212,11 @@ CompletionList = TypedDict('CompletionList', {
     'items': List[CompletionItem],
 }, total=True)
 
+Hover = TypedDict('Hover', {
+    'contents': Union[str, Dict[str, str], list],
+    'range': RangeLsp,
+}, total=False)
+
 PublishDiagnosticsParams = TypedDict('PublishDiagnosticsParams', {
     'uri': DocumentUri,
     'version': Optional[int],
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -213,7 +213,7 @@ CompletionList = TypedDict('CompletionList', {
 }, total=True)
 
 Hover = TypedDict('Hover', {
-    'contents': Union[str, Dict[str, str], list],
+    'contents': Union[str, Dict[str, str], List[str]],
     'range': RangeLsp,
 }, total=False)
 
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -212,8 +212,12 @@ CompletionList = TypedDict('CompletionList', {
     'items': List[CompletionItem],
 }, total=True)
 
+MarkedString = Union[str, Dict[str, str]]
+
+MarkupContent = Dict[str, str]
+
 Hover = TypedDict('Hover', {
-    'contents': Union[str, Dict[str, str], List[Union[str, dict]]],
+    'contents': Union[MarkedString, MarkupContent, List[MarkedString]],
     'range': RangeLsp,
 }, total=False)
 
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -213,7 +213,7 @@ CompletionList = TypedDict('CompletionList', {
 }, total=True)
 
 Hover = TypedDict('Hover', {
-    'contents': Union[str, Dict[str, str], List[str]],
+    'contents': Union[str, Dict[str, str], List[Union[str, dict]]],
     'range': RangeLsp,
 }, total=False)
 
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -194,7 +194,7 @@ class LspHoverCommand(LspTextCommand):
         for hover_response in self._hover_responses:
             content = (hover_response.get('contents') or '') if isinstance(hover_response, dict) else ''
             contents.append(minihtml(self.view, content, allowed_formats=FORMAT_MARKED_STRING | FORMAT_MARKUP_CONTENT))
-        return ''.join(contents)
+        return '<hr>'.join(contents)
 
     def show_hover(self, listener: AbstractViewListener, point: int, only_diagnostics: bool) -> None:
         sublime.set_timeout(lambda: self._show_hover(listener, point, only_diagnostics))
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -1,7 +1,10 @@
 from .code_actions import actions_manager
 from .code_actions import CodeActionOrCommand
 from .core.logging import debug
+from .core.promise import Promise
 from .core.protocol import Diagnostic
+from .core.protocol import Error
+from .core.protocol import Hover
 from .core.protocol import Position
 from .core.protocol import RangeLsp
 from .core.protocol import Request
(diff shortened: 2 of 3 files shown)


predragnikolic:
Note: There is one test that was failing even before this change:
```
ERROR: test_publish_diagnostics (test_server_notifications.ServerNotifications)
  File "/home/predragnikolic/.config/sublime-text/Packages/UnitTesting/unittesting/core/py33/case.py", line 28, in _executeTestPart
    yield from deferred
  File "/home/predragnikolic/.config/sublime-text/Packages/LSP/tests/test_server_notifications.py", line 41, in test_publish_diagnostics
    yield lambda: len(self.view.get_regions("lspTESTds3")) > 0
TimeoutError: Condition not fulfilled within 4.00 seconds

Ran 228 tests in 26.405s
```

plugin/hover.py
```diff
@@ -114,12 +119,33 @@ def run_async() -> None:
...
+        sessions = list(listener.sessions_async('hoverProvider'))
+        hover_promises = []  # type: List[Promise[ResolvedHover]]
+        document_position = text_document_position_params(self.view, point)
+        for session in sessions:
+            def hover_request() -> Promise[ResolvedHover]:
+                return session.send_request_task(
+                    Request("textDocument/hover", document_position, self.view),
+                )
+
+            hover_promises.append(hover_request())
```
rwols, line 131: Why can’t you call `hover_promises.append(session.send_request_task(…))` ?

plugin/hover.py
```diff
@@ -114,12 +119,33 @@ def run_async() -> None:
...
+        errors = []  # type: List[Error]
+        for response in responses:
+            if isinstance(response, Error):
+                errors.append(response)
+                continue
+            if response:
+                hovers.append(response)
+        if errors:
+            error_messages = ", ".join(str(error) for error in errors)
+            sublime.status_message('Completion error: {}'.format(error_messages))
```
rwols, line 146: Copy-paste mistake here.

plugin/hover.py
```diff
@@ -172,8 +191,11 @@ def code_actions_content(self) -> str:
         return "".join(formatted)
 
     def hover_content(self) -> str:
-        content = (self._hover.get('contents') or '') if isinstance(self._hover, dict) else ''
-        return minihtml(self.view, content, allowed_formats=FORMAT_MARKED_STRING | FORMAT_MARKUP_CONTENT)
+        contents = []
+        for hover_response in self._hover_responses:
+            content = (hover_response.get('contents') or '') if isinstance(hover_response, dict) else ''
+            contents.append(minihtml(self.view, content, allowed_formats=FORMAT_MARKED_STRING | FORMAT_MARKUP_CONTENT))
+        return ''.join(contents)
```
jwortmann, line 198: Would it be better to have some kind of visual separator between the contents of multiple responses? Like `'<hr>'.join(contents)` (but the append content in the line above only if not empty).

plugin/hover.py
```diff
@@ -114,12 +119,30 @@ def run_async() -> None:
...
-                lambda response: self.handle_response(listener, response, point))
+        sessions = list(listener.sessions_async('hoverProvider'))
+        hover_promises = []  # type: List[Promise[ResolvedHover]]
+        document_position = text_document_position_params(self.view, point)
+        for session in sessions:
+            hover_promises.append(session.send_request_task(
+                Request("textDocument/hover", document_position, self.view)
+            ))
+
+        Promise.all(hover_promises).then(lambda responses: self._on_all_settled(responses, listener, point))
```
jwortmann, line 130: I'm not sure how exactly this "Promise" line works and haven't tested it, but I assume that will show the hover popup only after the last of the responses has arrived. The `_show_hover` method already seems to have some logic to update the popup if already visible, so would it maybe be better to already show the popup after the first response and then just update the popup content when new responses arrive? This would mean to for example store the hover content in an instance variable, update it with each new response and clear the variable after the last response arrived. That would decrease the delay if a certain server takes a longer time than other servers to answer. (But not sure whether it would be worth the additional logic, and how many users use multiple servers on a single language in general.)

rchl: On the other side, it would provide worse experience because content would be changing and jumping around after the hover popup is shown. More than it already can at least.

plugin/core/protocol.py
```diff
@@ -212,6 +212,11 @@ class InsertTextMode:
     'items': List[CompletionItem],
 }, total=True)
 
+Hover = TypedDict('Hover', {
+    'contents': Union[str, Dict[str, str], list],
```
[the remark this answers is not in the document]
predragnikolic: `List[str]` is not entirely correct. A more "correct" representation would be `List[Union[str, dict]]` (because `MarkedString` can also be a `dict`)

reference from spec:
```
export interface Hover {
	contents: MarkedString | MarkedString[] | MarkupContent;
	range?: Range;
}

type MarkedString = string | { language: string; value: string };

export interface MarkupContent {
	kind: MarkupKind;
	value: string;
}
```

I wanted to create types for `MarkedString` and `MarkupContent` but that change would affect other files... so I do not want to do it in this PR. I want to keep this PR concise and just fix the issue for LSP-tailwindcss :)

plugin/hover.py
predragnikolic: For now I would not do the update of the popup as soon as one response is received
Only if someone reports this as an issue, then I would see if I need to change the logic.

plugin/hover.py
predragnikolic: no `<hr>`:
Screenshot from 2021-09-13 00-11-00

with `<hr>`:
Screenshot from 2021-09-13 00-12-18

This depends on personal preference.
I for example like the "no `<hr>`" option more.

At the end, this doesn't matter.
In practice, there will rarely be two servers that will provide hover responses for the given point in view.
Usually one server will return `None`. (but in theory it is possible that something like that happens)

plugin/hover.py
jwortmann: Okay, sounds good to me.

plugin/hover.py
jwortmann: I like the "with `<hr>`" version a bit more, but I would leave it to you to decide. Or maybe someone can add a third opinion here.

rwols: I prefer the `<hr>` version as well. Also, how would I know from which server the content came from in the popup?

rchl: Also prefer `<hr>`. Depending on the content that two or more servers provide, it could get weird without some sort of divider.

> Also, how would I know from which server the content came from in the popup?

I also thought about that but then I've realized that it's probably not the kind of information that the user needs to know.

predragnikolic: I added the `<hr>` divider

>  I also thought about that but then I've realized that it's probably not the kind of information that the user needs to know.

+1

rwols (approved):

PR #1847, opened 2021-09-13
Generalize implementation of the transport

rchl:
Extracted the JSON RPC-specific handling into a separate processor
and used composition concept to pass and use it within the transport.

That should allow re-using the transport implementation in the file
watcher to implement a transport that uses plain lines of text rather
than JSON RPC.

Not using abstract classes anymore as that is not compatible with
generic types.

### DIFF
--- a/plugin/core/transports.py
+++ b/plugin/core/transports.py
@@ -18,6 +18,7 @@ import weakref
 
 
 T = TypeVar('T')
+T_contra = TypeVar('T_contra', contravariant=True)
 
 
 class Transport(Generic[T]):
@@ -29,12 +30,12 @@ class Transport(Generic[T]):
         raise NotImplementedError()
 
 
-class TransportCallbacks(Protocol):
+class TransportCallbacks(Protocol[T_contra]):
 
     def on_transport_close(self, exit_code: int, exception: Optional[Exception]) -> None:
         ...
 
-    def on_payload(self, payload: Dict[str, Any]) -> None:
+    def on_payload(self, payload: T_contra) -> None:
         ...
 
     def on_stderr_message(self, message: str) -> None:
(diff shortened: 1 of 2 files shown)


plugin/core/transports.py
```diff
@@ -87,24 +129,16 @@ def __del__(self) -> None:
...
-                except Exception as ex:
-                    exception_log("JSON decode error", ex)
+                payload = self._processor.read_data(self._reader)
+                if payload is None:
                     continue
-                finally:
-                    # We don't need these anymore
-                    del body
-                    del headers
+
```
rchl, line 135: This is probably not very generic.

I've noticed that this depends on an exception being thrown to exit the loop when the process has exited. But that logic seems to be fairly specific to JSON RPC since with a processor that just reads lines as text there won't normally be any exception triggered.

Ideally there should probably be a more explicit way to shutdown the thread...

rwols:
I probably want to rewrite this part of the code once we can use py38. How would that work when other packages use this code?

rchl:
> I probably want to rewrite this part of the code once we can use py38. How would that work when other packages use this code?

The LSP-file-watcher-chokidar is tightly tied to LSP so I imagine those would have to be updated at the same time.

plugin/core/transports.py
```diff
@@ -18,49 +17,100 @@
...
+        body = self._encode(data)
+        writer.writelines(("Content-Length: {}\r\n\r\n".format(len(body)).encode('ascii'), body))
+
+    def read_data(self, reader: IO[bytes]) -> Optional[Dict[str, Any]]:
+        headers = http.client.parse_headers(reader)  # type: ignore
+        try:
+            body = reader.read(int(headers.get("Content-Length")))
+        except TypeError:
+            # Expected error on process stopping. Stop the read loop.
+            raise StopLoopError()
```
rchl, line 70: Change so that explicit action is used to stop the read loop instead of relying on `TypeError` doing it.

We could maybe just raise `StopLoopError()` earlier by checking if `headers.get("Content-Length")` is `None`.

rwols (approved):

PR #1852, opened 2021-09-16
Ensure lsp_save actually runs when context is active for keybinding

### LINKED ISSUE #1791: Unable to save modified file when dragged to a new window
Description
Editing typescript project, dragged a tab out to a separate window then modified some lines and tried to save. Sublime won't save unless I drag the tab back into the open project. (see sublime issue - https://github.com/sublimehq/sublime_text/issues/4623)

Steps to reproduce
Start Sublime Text, open a directory containing typescript files. (make sure to have the LSP plugin installed)
open multiple files in tabs
drag one of the tabs out to a separate window
modify the tab that's been dragged out, then try to save
Expected behavior
Expect the separate window/modified file to save.

Actual behavior
125587999-89de6ff6-454e-4746-ba01-375f066a4c3b

The separate window doesn't save with cmd+s (can still see the modified indication on top (circle icon/symbol)

Environment
Sublime Build: 4112
Operating system and version: macOS 11.4,

jwortmann:
Fixes #1791.

Not sure if it is a proper fix, but the logic to determine whether the keybinding is active or not, is already present in documents.py/on_query_context(). Otherwise, the `is_enabled` evaluating to False from LspTextCommand resulted in preventing <kbd>Ctrl</kbd>+<kbd>S</kbd> from doing anything, because that binding was active, but not actually run.

I would assume that the code actions on save still won't work in the case described in #1791, but I haven't tested it.

### DIFF
--- a/plugin/save_command.py
+++ b/plugin/save_command.py
@@ -91,6 +91,12 @@ class LspSaveCommand(LspTextCommand):
             self._trigger_native_save()
 
     def is_enabled(self, event: Optional[dict] = None, point: Optional[int] = None) -> bool:
+        # Workaround to ensure that the command will run, even if a view was dragged out to a new window,
+        # see https://github.com/sublimelsp/LSP/issues/1791.
+        # The check to determine whether the keybinding for lsp_save is applicable already happens in
+        # DocumentSyncListener.on_query_context and should not be required here, if lsp_save is only used for the
+        # keybinding. A proper fix should ensure that LspTextCommand.is_enabled returns the correct value even for
+        # dragged out views and that LSP keeps working as expected.
         return True
 
     def _trigger_on_pre_save_async(self) -> None:
--- a/plugin/save_command.py
+++ b/plugin/save_command.py
@@ -1,6 +1,6 @@
 from .core.registry import LspTextCommand
 from .core.settings import userprefs
-from .core.typing import Callable, List, Type
+from .core.typing import Callable, List, Optional, Type
 from abc import ABCMeta, abstractmethod
 import sublime
 import sublime_plugin
@@ -90,7 +90,7 @@ class LspSaveCommand(LspTextCommand):
         else:
             self._trigger_native_save()
 
-    def is_enabled(self) -> bool:
+    def is_enabled(self, event: Optional[dict] = None, point: Optional[int] = None) -> bool:
         return True
 
     def _trigger_on_pre_save_async(self) -> None:
--- a/plugin/save_command.py
+++ b/plugin/save_command.py
@@ -90,6 +90,9 @@ class LspSaveCommand(LspTextCommand):
         else:
             self._trigger_native_save()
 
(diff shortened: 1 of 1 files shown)


plugin/save_command.py
```diff
@@ -90,6 +90,9 @@ def run(self, edit: sublime.Edit) -> None:
         else:
             self._trigger_native_save()
 
+    def is_enabled(self, event: Optional[dict] = None, point: Optional[int] = None) -> bool:
+        return True
```
rwols, line 94: I don’t understand how this solves the linked issue. Can you add some comments in code to explain why this is needed?

rchl: Looks like after detaching a document, the `on_query_context` in:
https://github.com/sublimelsp/LSP/blob/0153663813015100013bbe44e9903d81b0f1b0a5/plugin/documents.py#L408-L408
will return `True` but `LspTextCommand.is_enabled`:
https://github.com/sublimelsp/LSP/blob/0153663813015100013bbe44e9903d81b0f1b0a5/plugin/core/registry.py#L68-L68
return `False`.

So it looks like those places check for sessions in a different one and one way is affected by the detach bug and the other isn't.

rchl (approved):
I like the fix but also agree that a comment (in the code) clarifying why it's needed would be good.

plugin/save_command.py
jwortmann: `DocumentSyncListener.on_query_context` returns `True` if a view is dragged to a new window, while `LspTextCommand.is_enabled` returns `False`. I don't know why this divergence happens, but there must be something fundamentally wrong with how applicable views are managed in a session or so. Probably due to the mentioned ST core issue that `on_post_move_async` isn't triggered. Anyway, `on_query_context` will cause the keybinding to be active, while `is_enabled` prevents the `run` method of `lsp_save` to be executed. This turns <kbd>Ctrl</kbd>+<kbd>S</kbd> into "noop". By enforcing `is_enabled` to be `True` is it at least ensured that the command is run, and the file can be saved. The bug affected users even if they don't use any code actions on save at all.

It should be pretty fine to have `is_enabled` always be `True`, because it is only used for that keybinding and not in code via `run_command("lsp_save")`. In general, it should not make much sense to do the same check twice whether the keybinding is applicable or not, both in `on_query_context` and in `is_enabled`. Eventually a proper fix would ensure that both will return the same value for `lsp_save`, so that the code actions on save also work as expected (not sure if they already work or not - I just wanted to fix the linked issue here).

I don't really know if the actual bug is in `DocumentSyncListener.on_query_context` or in `LspTextCommand.is_enabled` or somewhere else.

jwortmann: By the way, this bug affects other things too. For example DocumentHighlights still seem to work fine, while most other things like hover won't work anymore with such a view.

rwols (approved):

PR #1853, opened 2021-09-16
Remove obsolete dependencies

deathaxe:
As mdpopups vendors all required libraries as of 4.2.1 its dependencies can be removed.

### DIFF
--- a/dependencies.json
+++ b/dependencies.json
@@ -3,12 +3,7 @@
         ">=3124": [
             "backrefs",
             "bracex",
-            "markupsafe",
             "mdpopups",
-            "pygments",
-            "pymdownx",
-            "python-jinja2",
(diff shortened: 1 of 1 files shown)


rchl (approved):
Note that PC is notoriously bad at updating dependencies. Even though the latest `mdpopups` release is 9 days old, my two ST instances that I restart regularly still haven't picked it up. So if PC removes those dependencies and doesn't update `mdpopups` at the same time, things will break.

Not saying this because I'm against this change but just something to be aware of. The user might have to manually run `Satisfy Dependencies` to fix it.

rwols (approved):

deathaxe:
Works like a charm here with a slightly tweaked PC 4.0 alpha

Issue #1729, opened 2021-06-08
show_diagnostics_highlights does not allow styling


eproxus:
**Describe the bug**
The new option `show_diagnostics_highlights` that replaced `diagnostics_highlight_style` does not allow styling the highlights. This is a regression from the earlier feature set where inline diagnostic highlights could be styled to be more easily visible. The setting `document_highlight_style` only affect other highlights, not diagnostics highlights.

**To Reproduce**
Use the latest version.

**Expected behavior**
It should be possible to style diagnostic highlights.

**Environment (please complete the following information):**
- OS: macOS 11.4
- Sublime Text version: 4106
- LSP version: v1.5.0
- Language servers used: erlang_ls

jwortmann:
This is a necessary consequnce of a recent refinement for the diagnostics styles (#1710), i.e. errors and warnings always use squiggly style, whereas infos and hints always use dotted style. And this allows to draw multi-line diagnostics as a box without much confusion. It also ensures diagnostics to have a different style than the *document highlights* (solid underline by default). I guess the squiggly for errors/warnings is what most users would expect/use, so the removal of configuration options here seems to be acceptable in favor of the improvements.

> The setting document_highlight_style only affect other highlights, not diagnostics highlights.

Yes, this is only for the `textDocument/documentHighlight` request and has nothing to do with the diagnostic styles. For comparison, VSCode draws them similar to the "fill" style, but in ST that only looks well if it is supported/tweaked by the color scheme (as described in the docs under "Customization").

eproxus:
Ah, I understand. Is it planned to repair the ability to customize the error/warning (and info/hint) styles?

I’m having trouble seeing the various underlines and prefer the box styles with different colors for all diagnostics.

peterkos:
I don't personally understand the difficulty of adding a customization option that uses the defaults added in #1710 but allows for customization in a setting as before.
The code below is where I found the defaults are being set:
https://github.com/sublimelsp/LSP/blob/12de1dae52fe9664fbaf885d3a687e6cf1dc2d01/plugin/core/views.py#L32-L38

bakerkretzmar:
Would also really appreciate it if the ability to customize these were restored, I find `stippled` much easier to see. I have lots of the other 'boxes' and stuff turned off.

iamkroot:
I would also like to have this option as I am used to seeing `box` everywhere.
[...]

PR #1856, opened 2021-09-19
Allow customizing diagnostics styles

iamkroot:
This change brings back the "diagnostics_highlight_style" preference to allow customization of diagnostic styles, albeit only for single-lines.

See #1729

I am not sure how to update the `sublime-package.json` file with the new schema. Specifically, I want to use the `$ref` feature to not have to repeat the same schema. I have something like this:

```json
"diagnostics_highlight_style": {
  "oneOf": [
    {
      "enum": ["default", "box", "underline", "stippled", "squiggly", ""],
      "default": "default"
    },
    {
      "properties": {}  // way too much repetition here, need to reuse the previous schema
    }
  ],
  "default": "default",
  "markdownDescription": ""  // how to deal with code snippet inside this?
},
```


jwortmann:
I would drop the value "default" and explicitly use the mapping to severities as the default setting value. This makes it easier for users to see what the setting is expected to look like, if they want to adjust styles per severity.

In "markdownDescription" you can use Markdown as its name says ;) So fenced code blocks or indented code blocks should both work. I think sometimes the Markdown parser expects double linebreaks for correct formatting. For example:
`````
"markdownDescription": "Highlight style of code diagnostics.\nThis is a fenced code block:\n\n```json\n{\n    "hello": "world"\n}\n```\n"
`````

There is also a copy/paste error from "document highlights" in the settings description, and the name for Python's "dict" in JSON would be "object" (or you could use a generic word like "mapping" instead).

File: plugin/core/types.py [hunk 1]
```diff
@@ -187,6 +187,7 @@ class Settings:
     diagnostics_additional_delay_auto_complete_ms = None  # type: int
     diagnostics_delay_ms = None  # type: int
     diagnostics_gutter_marker = None  # type: str
+    diagnostics_highlight_style = None  # type: Union[str, Dict[str, str]]
```
iamkroot, line 190: Should this be `""` instead of `None`? Should fix the mypy lint complaints.

plugin/core/types.py
```diff
@@ -275,11 +276,14 @@ def r(name: str, default: Union[bool, int, str, list, dict]) -> None:
...
-        if isinstance(diagnostics_highlight_style, str):
-            if not diagnostics_highlight_style:
-                self.show_diagnostics_highlights = False
+        # correctness checking will happen inside diagnostics_highlight_style_flags method
+        self.diagnostics_highlight_style = s.get("diagnostics_highlight_style")
+
+        # Backwards-compatible with "show_diagnostics_highlights"
+        if self.diagnostics_highlight_style is None:
+            show_diagnostics_highlights = s.get("show_diagnostics_highlights")
+            self.diagnostics_highlight_style = "default" if show_diagnostics_highlights else ""
```
iamkroot, line 285: Not sure if it makes sense to have `"default"` here, it is not a special value. The alternative is to specify the full dict, but that would mean duplicating logic between this and `DIAGNOSTIC_SEVERITY` inside `views.py`.

plugin/core/types.py
```diff
@@ -296,6 +300,39 @@ def document_highlight_style_region_flags(self) -> Tuple[int, int]:
         else:
             return sublime.DRAW_NO_FILL, sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE | sublime.DRAW_SOLID_UNDERLINE
 
+    @staticmethod
+    def _style_str_to_flag(style_str: str) -> int:
+        # This method could be a dict or lru_cache
```
iamkroot, line 305: Should I make this a dict?

plugin/core/types.py
rwols: It must remain `None` because from what I understand, this is the way to annotate class member variables in old versions of python.

plugin/core/types.py
```diff
@@ -296,6 +300,39 @@ def document_highlight_style_region_flags(self) -> Tuple[int, int]:
...
+        """Returns flags for highlighting diagnostics on single lines per severity"""
+        if isinstance(self.diagnostics_highlight_style, str):
+            # same style for all severity levels
+            return [self._style_str_to_flag(self.diagnostics_highlight_style)] * 4
+        elif isinstance(self.diagnostics_highlight_style, dict):
+            flags = []
+            for sev in ("error", "warning", "info", "hint"):
+                user_style = self.diagnostics_highlight_style.get(sev)
+                if user_style is None:  # user did not provide a style
+                    flags.append(0)
```
rwols, line 329: How is the flag `0` rendered? Is `0` an appropriate default?

sublime-package.json
```diff
@@ -309,8 +309,40 @@
...
+                  "additionalProperties": false
+                }
+              ],
+              "default": {
+                "error": "squiggly",
+                "warning": "squiggly",
+                "info": "stippled",
+                "hint": "stippled"
+              },
+              "markdownDescription": "Highlight style of code diagnostics.\nCan be a string, or a mapping of string->string for severity-based styling.\nValid string values are `\"box\"`, `\"underline\"`, `\"stippled\"`, `\"squiggly\"` or the empty string (`\"\"`).\nWhen set to a valid string value, all severities will use that style.\nWhen set to the empty string (`\"\"`), no document highlighting is requested.\nDiagnostics which span multiple lines are always drawn with the `box` style (See `\"show_multiline_diagnostics_highlights\"`)."
```
rwols, line 345: This mentions "document highlighting is requested", but we're dealing with diagnostics, not document highlights.

plugin/session_view.py
```diff
@@ -202,11 +202,12 @@ def diagnostics_tag_scope(self, tag: int) -> Optional[str]:
         return None
 
     def present_diagnostics_async(self) -> None:
-        flags = 0 if userprefs().show_diagnostics_highlights else sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE
+        flags = userprefs().diagnostics_highlight_style_flags()  # for single lines
+        multiline_flags = 0 if userprefs().show_multiline_diagnostics_highlights else sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE
```
rwols, line 206: I don't get this `show_multiline_diagnostics_highlights`, do you also want to add the option to not render multiline diagnostics at all?

plugin/core/types.py
rwols: see https://mypy.readthedocs.io/en/stable/class_basics.html

plugin/core/types.py
jwortmann: I guess it is used because it evaluates to "false", and that allows to use a convenient way to get the actual value with an `or` or `if ... else` in the other function below. But maybe it would be more readable to do this logic right here and just append the actual default value for each severity instead of `0`.

plugin/core/types.py
jwortmann: I think the only thing you need to do here is to set all styles to `""` if "show_diagnostics_highlights" is `False`. Otherwise the default values are already defined by the default setting for "diagnostics_highlight_style".

iamkroot: Oh right, the settings file we ship already has the proper defaults.

plugin/core/types.py
iamkroot: Ah I see. Mypy is complaining later (line 280) when I set the var to user input directly, and leave the verification for `_style_str_to_flag`. Should I add a directive for mypy to ignore that line?

plugin/core/types.py
iamkroot: I guess we can leave this for another refactor.

plugin/core/types.py
iamkroot: @jwortmann You are correct. The scheme takes advantage of the short-circuiting behaviour of python's `or` operator to make the code less verbose. 

Regarding moving the logic here, I considered that, but decided against it. The reason being, this class/file is only supposed to model _user settings_, at least from what I understood. If we start adding custom logic here, it might open the floodgates to this class being a dumping ground for more logic.

sublime-package.json
iamkroot: Yup, my error, will fix.

plugin/session_view.py
iamkroot: Yeah, I believe the user should have the option to disable all highlights. If they just set `diagnostics_highlight_style` to `""`, it will (currently) only disable the single-line highlights. So this option was added to explicitly toggle multiline highlights.

The reasons I did not want to infer this option from the single setting are the following:
1. Checking that the value `""` is fine, but we would also need to handle the case when all severities are set to `""`. If we choose not to, we would create a mismatch between the meaning of value `""` and an object with all the values set to `""`.
2. Relatedly, it would be better if the user expects `diagnostics_highlight_style` to either deal with all highlights, or only single line (as they are now, maybe we should rename it?). The former gives a wrong expectation that the user can customize multi-line highlight styles. So I chose to create a separate setting.

PR #1861, opened 2021-09-27
Add documentation for terraform-ls

mconigliaro:


rchl:
Actually, we should recommend https://github.com/sublimelsp/LSP-terraform here instead

mconigliaro:
That would be nice if the package actually existed in packagecontrol. 😁

PR #1862, opened 2021-09-28
Add a couple of types here and there

rchl:


plugin/core/protocol.py
```diff
@@ -156,7 +165,6 @@ class InsertTextMode:
 LocationLink = TypedDict('LocationLink', {
     'originSelectionRange': Optional[RangeLsp],
     'targetUri': DocumentUri,
-    'targetRange': RangeLsp,
```
predragnikolic, line 159: I would not remove `targetRange`, because it exist in the spec.

But I do not see it as a bad thing to remove it for now, if it is not used.

rchl: I agree that it should not be removed. Somehow I didn't realize that it's a type from the spec.

tests/test_single_document.py
```diff
@@ -40,7 +40,6 @@
     {
         'originSelectionRange': {'start': {'line': 0, 'character': 0}},
         'targetUri': GOTO_RESPONSE[0]['uri'],
-        'targetRange': GOTO_RESPONSE[0]['range'],
```
predragnikolic, line 43: forgot to revert this line? :)

rchl: thanks

predragnikolic (approved):

plugin/references.py
```diff
@@ -30,17 +30,17 @@ class LspSymbolReferencesCommand(LspTextCommand):
...
-    def __init__(self, view: sublime.View) -> None:
-        super().__init__(view)
-        self._picker = None  # type: Optional[LocationPicker]
-
     def run(self, _: sublime.Edit, event: Optional[dict] = None, point: Optional[int] = None) -> None:
         session = self.best_session(self.capability)
         file_path = self.view.file_name()
         pos = get_position(self.view, event, point)
         if session and file_path and pos is not None:
-            params = text_document_position_params(self.view, pos)
```
rwols, line 42: What was wrong with the old way here though?

predragnikolic: Nothing, 
but I found the new way more explicit.

rchl: The return type of was updated from `Dict[str, Any]` to `TextDocumentPositionParams` and since `TextDocumentPositionParams` is a "sealed" object (total=True), you can't add more properties to it in a separate statement.

So with updated return type, doing it like before would be a type error.

jwortmann: Does that mean we can't use the `text_document_position_params` function anymore if we want to add more properties to the returned dict?

See https://github.com/sublimelsp/LSP-julia/blob/ad7415e040464afac39796fe07fd58c73b7f9120/plugin.py#L61-L69

rchl: Correct, that will be a type error now and would need to handle it the same as here.

jwortmann: Ok no problem, the alternative handling is quite simple too and I can easily update it.

plugin/references.py
```diff
@@ -30,17 +30,17 @@ class LspSymbolReferencesCommand(LspTextCommand):
...
-            params['context'] = {"includeDeclaration": False}
+            textDocumentPositionParams = text_document_position_params(self.view, pos)
```
rwols, line 38: Sorry for being late with replying. My OCD is telling me this variable should be snake_case :)

rchl: I agree. :)

rwols (approved):

PR #1865, opened 2021-10-02
Don't allow old nested dict style keys anymore

### LINKED ISSUE #1858: Unwanted splitting on dot of the setting keys
**Describe the bug**

In LSP-yaml, the user can specify schema mapping like so:

```
yaml.schemas: {
    "https://json.schemastore.org/composer": "/*"
}
```

This doesn't work in ST because LSP splits the URL key on every dot, creating an object that the server doesn't understand.

**Expected behavior**

Should send the value of the `yaml.schemas` object without altering it

**Additional context**

I'm not sure how VSCode manages to handle it but I would guess it is somewhat taking into consideration the schema itself. We could try to check that first.

rwols:
Fixes #1858

### DIFF
--- a/plugin/core/collections.py
+++ b/plugin/core/collections.py
@@ -135,13 +135,10 @@ class DottedDict:
         """
         Overwrite and/or add new key-value pairs to the collection.
 
-        :param      d:    The overriding dictionary. Can contain nested dictionaries.
+        :param      d:    The overriding dictionary. Keys must be in the new-style dotted format.
         """
         for key, value in d.items():
-            if isinstance(value, dict):
-                self._update_recursive(value, key)
-            else:
-                self.set(key, value)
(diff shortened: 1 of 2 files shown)


rchl:
Have to be more careful when using `update()` now.

For example this is no longer entirely correct https://github.com/sublimelsp/LSP-vue/blob/d831ccd89b476f76b42d52a77bc1b8693dc85781/plugin.py#L24-L30 
It will still work fine because it just happens that the `vetur.format.options` doesn't contain any other keys than those that are updated but it's not really correct anymore.

rchl (approved):
Nice and clean solution. Thought the problem is more difficult to solve.

rwols:
Yeah, it took me quite a while to debug this.

PR #1866, opened 2021-10-02
Add a comment to guide new users how to change the keybind of a command

RUSshy:
I got confused because i didn't know about the context part

I think this comment could be helpful to guide new users

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -1,5 +1,6 @@
 [
-    // To change the keybind of a command, copy its section, and edit the 'keys' part
+    // To enable or change an existing key binding, copy it to the right view and uncomment, if needed.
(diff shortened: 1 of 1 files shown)


rwols (approved):

Default.sublime-keymap
```diff
@@ -1,4 +1,6 @@
 [
+    // To change the keybind of a command, copy its section, and edit the 'keys' part
```
rchl, line 2: If we follow the ST nomenclature then each object is called a "key binding" and we are calling it "command" here which is rather just a part of a key binding. So this sentence can be confusing.

Maybe:

```suggestion
    // To enable or change an existing key binding, copy it to the right view.
    // To change the associated key press, edit the "keys" value.
```

rchl (approved):

PR #1867, opened 2021-10-03
Fix console exception with panel logger enabled

rchl:
After window is closed, the server messages can keep coming for a while
and the "is_valid()" check doesn't really work as expected for closed
Window so we need to do the "is in" check instead.


rchl:
When checking the "Window.is_valid()" in the console it seems to work but I guess there is a small time frame after the tab is removed where it will still be reporting is_valid() == True

Issue #1721, opened 2021-06-04
The `next_result`/`prev_result` hooks don't jump to the next/prev diagnostic if there are build results


timfjord:
**Describe the bug**
The `next_result`/`prev_result` hooks don't jump to the next/prev diagnostic if there are build results

**To Reproduce**
Steps to reproduce the behaviour:
1. Initiate a build
2. Introduce an LSP error in some file
3. Press `f4`(the default `next_result` shortcut)
4. Observe nothing happens

**Expected behaviour**
The error should be highlighted

**Environment (please complete the following information):**
- OS: macOs 0.15.7 
- Sublime Text version: 4107
- LSP version: 1.4.0
- Language servers used: ruby(solargraph, sorbet)

mikewilkerson-solas:
It appears that this also happens when generating search results (the `next_result`/`previous_result` functionality applies to searches, not LSP diagnostics). The only way I know to get back to navigating LSP results is to quit and restart Sublime Text.

(I'm also on build 4107)

wilkystyle:
It appears that this also happens when generating search results (the `next_result`/`previous_result` functionality applies to searches, not LSP diagnostics). The only way I know to get back to navigating LSP results is to quit and restart Sublime Text.

(I'm also on build 4107)

timfjord:
I think closing the search result page(or the search panel) should fix the problem. But there is no such thing for builds, there is no way to reset the build without restarting Sublime

wilkystyle:
> I think closing the search result page(or the search panel) should fix the problem. But there is no such thing for builds, there is no way to reset the build without restarting Sublime

At least for me, closing the search page doesn't restore my ability to navigate between LSP diagnostics with `next_result` or `previous_result`—it simply disables navigating between search results, with `next_result`/`previous_result` no longer performing any action.

rwols:
Fixed by #1884

timfjord:
@rwols I think the issue is only partly fixed, because in the References search the `next_result/prev_result` hooks are still used.

It might make sense to introduce the same `Goto` thing but for references

rwols:
You can use a quick panel for references, see LSP.sublime-settings.

timfjord:
Indeed, thanks for the tip @rwols

frou:
[...]

Issue #1696, opened 2021-05-28
Filter errors in the diagnostics panel to files in view


Rapptz:
**Is your feature request related to a problem? Please describe.**
Some projects have a lot of lints that cause the diagnostic panel to be spammed with warnings/hints/errors. 

**Describe the solution you'd like**
For focusing on the problem I think it'd be desirable to only have the currently open file's diagnostics displayed.

**Describe alternatives you've considered**
Disabling the panel.

**Additional context**
N/A

rwols:
You could also:

- Right-click on the Tab and select Close Other Tabs
- Disable various linter settings in the server settings, if your langserver has those settings

rwols:
Fixed by #1884

PR #1884, opened 2021-10-28
"Goto Diagnostic" quick panels.

htmue:
@rchl (https://github.com/sublimelsp/LSP/issues/1721#issuecomment-951725176): This is what I had in mind.

With `lsp_goto_diagnostic` / `lsp_goto_diagnostic_in_project` you can launch a quick panel with either the diagnostics for the current file or all diagnostics currently in store. It shows the selected diagnostic in a view behind the panel and opens transient views when necessary. With ENTER you go to the corresponding location. With ALT+ENTER on the project panel you switch to the corresponding per-file panel.

It doesn't interfere with `next_result` / `previous_result`.

It is already quite usable, especially when bound to shortcuts, but definitely needs some polishing:

* activate in-view diagnostics for transient views (should be possible without pinging the language server, the diagnostics are in store)
* show in-view popup for selected diagnostic (maybe with on/off user setting)
* center selected diagnostic in main diagnostics panel (maybe with on/off user setting)
* launch code actions with CMD+ENTER
* ...

Addresses #1696, #1721.


htmue:
One more thing about the failing tests: When I try to use Pylsp or Pyright with the LSP code itself, I get this exception, no matter whether on this PR or on main:
```
LSP: starting ['/Users/htmue/Library/Caches/Sublime Text 3/Package Storage/LSP-pylsp/bin/pylsp'] in /Users/htmue/Library/Application Support/Sublime Text 3/Packages/LSP
failed to register session LSP-pylsp to listener ViewListener(20)
  File "/Users/htmue/Library/Application Support/Sublime Text 3/Packages/LSP/plugin/core/windows.py", line 271, in _publish_sessions_to_listener_async
    listener.on_session_initialized_async(session)
  File "/Users/htmue/Library/Application Support/Sublime Text 3/Packages/LSP/plugin/documents.py", line 218, in on_session_initialized_async
    self._session_views[session.config.name] = SessionView(self, session, self._uri)
  File "/Users/htmue/Library/Application Support/Sublime Text 3/Packages/LSP/plugin/session_view.py", line 60, in __init__
    self._clear_auto_complete_triggers(settings)
  File "/Users/htmue/Library/Application Support/Sublime Text 3/Packages/LSP/plugin/session_view.py", line 85, in _clear_auto_complete_triggers
    triggers = [t for t in triggers if self.session.config.name != t.get("server", "")]
  File "/Users/htmue/Library/Application Support/Sublime Text 3/Packages/LSP/plugin/session_view.py", line 85, in <listcomp>
    triggers = [t for t in triggers if self.session.config.name != t.get("server", "")]
AttributeError: 'str' object has no attribute 'get'

Exception AttributeError: "'str' object has no attribute 'get'" in <bound method SessionView.__del__ of <LSP.plugin.session_view.SessionView object at 0x7f9a4e394e50>> ignored
```
Stack trace only for Pylsp, Pyright startup crashes at the same point.

Removing the cache and restarting Sublime doesn't help.

With my main project (big multi-folder mixed-languages project) both run fine using this PR.

Any idea?

htmue:
Screenshot 2021-10-29 at 22 01 50
Screenshot 2021-10-29 at 23 20 25

htmue:
Screenshot 2021-10-30 at 19 20 29
Screenshot 2021-10-30 at 19 20 54


> ... When I try to use Pylsp or Pyright with the LSP code itself, I get this exception, ...

After removing the UnitTesting package, it works again.

htmue:
> It perhaps would look better to use a ListInputHandler instead of the quick panel, because it's newer and would allow more styling, i.e. the long diagnostic messages could be put into the "details" panel at the bottom then.

For now I have removed the long messages. To have one "details" panel at the bottom, for the selected item, would be much better, I agree. This would also be better than opening the popup in the view behind the panel.

Would it work like this with ListInputHandler? At first glance it seems to have a totally different interface. Maybe this could be added at a later time, separate from this PR, by reimplementing the LocationPicker? From this the other uses of the LocationPicker would benefit as well.

rchl:
I have also minor comments about some things:
 - the row:column info looks quite messy. Do we really need it? I know that you are probably trying to match the diagnostics panel but I'm not really sure if this provides meaningful benefit
 - Does the server name have to be prefixed with `#`? We don't use that kind of presentation anywhere else
 - Can we also include the diagnostic code like we do in the panel? Some servers just show a number here, some more meaningful and useful info like linting rule name.

Default.sublime-commands
```diff
@@ -81,6 +81,14 @@
         "caption": "LSP: Goto Declaration",
         "command": "lsp_symbol_declaration",
     },
+    {
+        "caption": "LSP: Goto Diagnostic",
```
rchl, line 85: The name doesn't make it clear that it only shows diagnostics for the current file. Maybe `LSP: Goto Diagnostic in File`?

htmue: Yes, I agree. I took it from "Goto Symbol..." and "Goto Symbol in Project...", but adding "for File' makes it clear immediately.

jwortmann:
> 
> For now I have removed the long messages. To have one "details" panel at the bottom, for the selected item, would be much better, I agree. This would also be better than opening the popup in the view behind the panel.
> 
> Would it work like this with ListInputHandler? At first glance it seems to have a totally different interface. Maybe this could be added at a later time, separate from this PR, by reimplementing the LocationPicker? From this the other uses of the LocationPicker would benefit as well.

Yes, no problem, maybe I can give it a try to change from the LocationPicker class to a ListInputHandler after this PR is merged. We would need to see if this would work and be useful for the other overlays too. But for the diagnostics, I could imagine that the ListInputHandler would have another great advantage: the ListInputHandlers work like a stack in the command palette, i.e. if you need multiple inputs, it is possible to open multiple InputHandlers after another, and you can always navigate back with <kbd>backspace</kbd>. I guess this could work well for the "Goto Diagnostic in Project" command, which would first show a ListInputHandler with the filenames, and then the one from "Goto Diagnostic" with the list of diagnostics. ("Goto Diagnostic in Project" did only opened the file when I tried it yesterday)

---

Two other comments:
- there is a convention that the names of command classes end with "Command", so for example `LspGotoDiagnosticCommandBase` should probably be `LspGotoDiagnosticBaseCommand`
- can you add the popup text for the icon:
```python
DIAGNOSTIC_KIND_TUPLE = {
    DiagnosticSeverity.Error: (sublime.KIND_ID_COLOR_REDISH, "e", "Error"),
    DiagnosticSeverity.Warning: (sublime.KIND_ID_COLOR_YELLOWISH, "w", "Warning"),
    DiagnosticSeverity.Information: (sublime.KIND_ID_COLOR_BLUISH, "i", "Information"),
    DiagnosticSeverity.Hint: (sublime.KIND_ID_COLOR_BLUISH, "h", "Hint"),
}
```

Default.sublime-commands
htmue: Hm, I just tried it and I wouldn't do this. It doesn't feel right. My understanding is the non-qualified "Goto (whatever)" implies the active view as context.

With "Goto Diagnostic for File" there is immediately the question "which file?" and I would expect a file chooser, but this is exactly what the other command does.

htmue:
> I have also minor comments about some things:

Yes to all the suggestions.

> command classes end with "Command", so for example `LspGotoDiagnosticCommandBase` should probably be `LspGotoDiagnosticBaseCommand`

I meant to make clear which of the classes are in fact commands and which are support classes that won't work as command themselves. What's the convention for that?

> ... the ListInputHandlers work like a stack in the command palette, i.e. if you need multiple inputs, it is possible to open multiple InputHandlers after another, and you can always navigate back with backspace.

Top.

> ("Goto Diagnostic in Project" did only opened the file when I tried it yesterday)

I changed it to shift+enter, maybe it was that? Alt+enter felt too heavy. For testing I have bound the commands to f8 / shift+f8.

Screenshot 2021-10-30 at 21 26 24

Screenshot 2021-10-30 at 21 24 53

jwortmann:
> I changed it to shift+enter, maybe it was that?

Oh, I wasn't aware of the modifier key. I just tried it again and it works fine, even when selecting an entry with the mouse. Your EnhancedLocationPicker is pretty impressive! But I wonder if the behavior from <kbd>shift+enter</kbd> shouldn't be the default (on <kbd>enter</kbd> without modifier key)? I don't think I've seen this possibility to use modifier keys in the command palette / list overlays in ST before, so I guess most of the users wouldn't know about this ability.

I have a feeling that the ListInputHandler could simplify things a lot here regarding the implementation, and the additional "Base" commands probably wouldn't be needed then. But I must say that it already works pretty well and looks neat.

htmue:
> But I wonder if the behavior from shift+enter shouldn't be the default (on enter without modifier key)?

Yes. The modifier key approach came from my first take (one entry per diagnostic in the project). It was just easier to try out the simplified version (one entry per file) without going into the location picker code again.

Plain `enter` without modifier is better, I agree. Then `enter` twice in a row will take you to the first diagnostic in that file anyway. (But I would welcome `cmd+enter` for code actions, which could be added later.)

> I have a feeling that the ListInputHandler could simplify things a lot here regarding the implementation.

Excellent! Also this stack behaviour you mentioned would be very useful for quickly browsing through the code breakage one would cause with a certain change, in a big project.

You said you could maybe give the ListInputHandler a try?

jwortmann:
> You said you could maybe give the ListInputHandler a try?

Yeah maybe. But your code looks a bit complicated with all this OOP stuff and generators everywhere :wink:
Here would be something to start with:

- for `Default.sublime-commands`:
```json
    {
        "caption": "LSP: Goto Diagnostic",
        "command": "lsp_goto_diagnostic",
        "args": {"uri": "$view_uri"}
    },
    {
        "caption": "LSP: Goto Diagnostic in Project",
        "command": "lsp_goto_diagnostic"
    },
```

- for `goto_diagnostic.py`:
<details>

```python
import os

def get_sessions(window: sublime.Window) -> Iterator[Session]:
    wm = windows.lookup(window)
    if wm is not None:
        for session in wm._sessions:
            yield session


class LspGotoDiagnosticCommand(sublime_plugin.WindowCommand):
    def run(self) -> None:
        pass

    def is_enabled(self) -> bool:
        view = self.window.active_view()
        if view is None:
            return False
        # TODO check if there are any diagnostics or not
        return True

    def input(self, args: dict) -> Optional[sublime_plugin.CommandInputHandler]:
        if "uri" not in args:
            return DiagnosticUriInputHandler(self.window)
        uri = args["uri"]
        if uri == "$view_uri":
            view = self.window.active_view()
            uri = parse_uri(uri_from_view(view))[1]  # type: ignore  # view cannot be None due to self.is_enabled()
        if "diagnostic" not in args:
            return DiagnosticInputHandler(self.window, uri)

    def input_description(self) -> str:
        return "Goto Diagnostic"


class DiagnosticUriInputHandler(sublime_plugin.ListInputHandler):
    uri = ""

    def __init__(self, window: sublime.Window) -> None:
        self.window = window

    def name(self) -> str:
        return "uri"

    def list_items(self) -> List[sublime.ListInputItem]:
        file_items = list()
        max_severity = userprefs().diagnostics_panel_include_severity_level
        items_per_path = OrderedDict(
        )  # type: OrderedDict[ParsedUri, List[Tuple[weakref.ref[Session] ,Union[Location, LocationLink], int]]]
        for session in get_sessions(self.window):
            weaksession = weakref.ref(session)
            for parsed_uri, (location, item) in session.diagnostics_manager.filter_map_diagnostics_flat_async(
                    is_severity_included(max_severity), _location_severity):
                seen = parsed_uri in items_per_path
                items_per_path.setdefault(parsed_uri, []).append((weaksession, location, item))
                if not seen:
                    items_per_path.move_to_end(parsed_uri)
        for parsed_uri, location_items in items_per_path.items():
            weaksession, location, _ = location_items[0]  # non-empty list
            counts = Counter(severity for _, _, severity in location_items)
            text = "{}: {}".format(format_severity(min(counts)), self._simple_project_path(parsed_uri))
            annotation = "E: {}, W: {}".format(counts.get(DiagnosticSeverity.Error, 0),
                                               counts.get(DiagnosticSeverity.Warning, 0))
            kind = DIAGNOSTIC_KIND_TUPLE[min(counts)]
            file_items.append(sublime.ListInputItem(text, parsed_uri[1], annotation=annotation, kind=kind))
        return file_items

    def placeholder(self) -> str:
        return "Select file"

    def confirm(self, value: DocumentUri) -> None:
        self.uri = value
        # self.window.open_file(value, flags=sublime.TRANSIENT)

    def next_input(self, args: dict) -> Optional[sublime_plugin.CommandInputHandler]:
        return DiagnosticInputHandler(self.window, self.uri)

    def description(self, value: DocumentUri, text: str) -> str:
        return os.path.basename(value)

    def _simple_project_path(self, parsed_uri: ParsedUri) -> str:
        scheme, path = parsed_uri
        if scheme == "file":
            path = str(simple_project_path(map(Path, self.window.folders()), Path(path))) or path
        return path


class DiagnosticInputHandler(sublime_plugin.ListInputHandler):
    def __init__(self, window: sublime.Window, uri: DocumentUri) -> None:
        self.window = window
        self.uri = uri

    def name(self) -> str:
        return "diagnostic"

    def list_items(self) -> List[sublime.ListInputItem]:
        items = list()
        # add diagnostics for self.uri here
        items.append("foo")
        items.append("bar")
        return items

    def placeholder(self) -> str:
        return "Select diagnostic"

    def preview(self, value) -> str:
        return ""
        # return description of diagnostic for details pane. "value" can be defined in each ListInputItem as desired
        # and scroll view to location of the diagnostic (should be included in "value")

```
</details>

If you can return a `ListInputItem` from `session.diagnostics_manager.diagnostics_by_parsed_uri()`, then this code should already cover a big part of the functionality. And it only requires a single command.
The URI/filepath handling probably doesn't work in the snippet above, as this was just hacked together quickly.

htmue:
> But your code looks a bit complicated with all this OOP stuff and generators everywhere 😉 

It seems it's always getting a mess when I try OOP. And after some years of Haskell I'm thinking in functors, traversals and folds, and curried functions... This doesn't always translate nicely to Python.

> Here would be something to start with: [...]

Thanks! Now I see how it is meant to work with input handlers. And this leads to much nicer code indeed. I'm filling in the blanks now.

rchl:
Not entirely on topic of this PR but I've just realized that with #1881 the diagnostics are no longer sorted by location in file. Even more noticeable with this PR since you'd expect sorted order when navigating diagnostics in current file.

plugin/goto_diagnostic.py
```diff
@@ -0,0 +1,206 @@
...
+from .core.registry import windows
+from .core.sessions import Session
+from .core.settings import userprefs
+from .core.typing import Iterable, Iterator, List, Optional, Tuple, Union
+from .core.url import parse_uri, unparse_uri
+from .core.views import MissingUriError
+from .core.views import diagnostic_severity, format_diagnostic_source_and_code, format_severity, uri_from_view
+from .locationpicker import EnhancedLocationPicker, OnModifierKeysAcitonMap
+from collections import Counter, OrderedDict
+from pathlib import Path
```
rchl, line 12: `pathlib` is not available in the 3.3 runtime that this package runs within. I think that's the reason for tests failing.

Though I believe there is a PC dependency available and we have it installed through another package so it doesn't fail locally.

We could either add `pathlib` to `dependencies.json` or not use `Path`.

htmue:
> Not entirely on topic of this PR but I've just realized that with #1881 the diagnostics are no longer sorted by location in file. Even more noticeable with this PR since you'd expect sorted order when navigating diagnostics in current file.

Actually, I don't. Type checkers have to walk their way through the (reverse) dependency tree, therefore chances are good that a code issue that causes subsequent problems in the first place is also reported first. This root issue doesn't have to be an error, it can also show up as a warning. Ordering by severity and/or line number seems to me a bit arbitrary in comparison.

That's why I was careful to preserve incoming order. This is the order one also gets when running a compiler on the command line (more or less). 

But preferences differ, sorting could be offered as a user setting (see my commit message cc2948c).

rchl:
I'm not sure how much of that applies in LSP. Just the [last example](https://github.com/sublimelsp/LSP/pull/1884#issuecomment-956618124) shows the case where ordering by location would be more useful since all issues are unrelated.

But, you don't have to do anything about that now. We can think about that later since we haven't sorted by location before either, only by severity.

htmue:
It absolutely applies to LSP. Not that much in single-file mode, maybe, but when you use it with a project-aware language server and a language with an advanced type system, LSP can really shine. After all, I'm here because awareness of the latter apparently was lost here. I want to bring LSP a bit closer to its potential for languages like that.

Sorting as an option I suggested already in my first commit, maybe open a new issue?

@jwortmann: Btw. The list input handler rewrite is kind of done, I'm just giving it a spin before committing.

predragnikolic:
> But, you don't have to do anything about that now. 
We can think about that later since we haven't sorted by location before either, only by severity.

I just wanted to confirm that we did sort the diagnostics by the location, but it was removed with:
https://github.com/sublimelsp/LSP/commit/f41c102ccc4117cf27aca6570e90951f7fb34224#diff-982c110867d9809c978a5597e4da66ce7538b327675a66b8f27214413b7c0ec1L253

jvican:
The order in which diagnostics are sent is best understood by the language server that has semantic information about each diagnostic and can do a better job than the editor can. Thus, the language server should have a guarantee that the order in which messages are sent ends up being the same order that users see the diagnostics in.

Sorting diagnostics in the editor can be seen as an improvement, but I'd caution against it because many language servers already send diagnostics in a particular order on purpose. Looks like the default LSP for Python isn't smart enough to establish a sensible order, but there are compilers and LSPs that can. As it happens, I've written such logic in the Scala compiler toolchain in the past and that's the current logic used to order Scala diagnostics.

I hope I've made the case to not make that a default again. I'm not sure if it's worth having it as an option or just asking the downstream LSP servers such as Python LSP to send messages in a better order.

predragnikolic:
> Sorting as an option I suggested already in my first commit, maybe open a new issue?

Because sorting is out of scope for this PR, I will not try to start a conversation.

> The order in which diagnostics are sent is best understood by the language server that has semantic information about 
each diagnostic and can do a better job than the editor can. Thus, the language server should have a guarantee that the order in which messages are sent ends up being the same order that users see the diagnostics in.

Not all language servers can guarantee that. 

If 3 language servers are running in one file.
When navigating diagnostics that are unsorted.
The view may jumps up and down.

Which for me is acceptable.
But when navigating diagnostics that are sorted.
I can either go only forward, or only backward in the view.
Which for me seems like a more nicer experience.

---- 

I haven't seen Scala diagnostics, so maybe for you having unsorted diagnostics is better.

Having an option where like `"diagnotics_order": "sorted" | "preserve_order"` would make sense... but having too much options affects maintenance.

htmue:
The last commit adds preview pane and back button.
Screenshot 2021-11-03 at 22 00 39

As far as I'm concerned, the feature is finished (I stopped using `f4`).

I suggest to bind "Goto Diagnostic" to `f8` and "Goto Diagnostic in Project" to `shift+f8` (next/prev key from LSP/ST3).

Future improvements could be:

- squiggly lines in transient file previews
- highlight selected item in file previews
- scroll to and highlight selected item in diagnostics panel
- full content and styling of LSP hover popup in preview pane
- `cmd+enter` to launch code actions

Thanks to @jwortmann and @rchl for suggestions and support!

jwortmann:
Looks and works nice!
Some little tweaks:
* could you add `a {color: var(--bluish)}` to PREVIEW_PANE_CSS
* do we really need the "back" option? It already comes for free when just pressing <kbd>backspace</kbd>, so I would remove it
* I guess you can add the keybindings to Default.sublime-keymap, but I would rather leave them commented out because they might conflict with user keybindings
* if there is a `codeDescription` for a diagnostic, then a colon will be appended to the server name (annotation) from the `diagnostic_source_and_code()` function
* if the filepath is very long, then the description "blob" in the input field will span the whole width of the text field or even be truncated at the right side. That's why I suggested to only show the filename via `os.path.basename()`. I still think it would be the better option, because you already saw the (project) filepath when you selected the file.

screenshot

---

Regarding the diagnostics sorting; I also prefer them sorted by location, so I would be in favor of having a setting for it (in another PR). I would maybe add another option to sort them by severity (and then by location for diagnostics of same severity):
```jsonc
// Sort order of diagnostics for each file in the diagnostics panel and the Goto Diagnostics command
"diagnostics_sort_order": "none" | "location" | "severity"
```

rchl:
It would be nice to add keybindings but commented out by default. Would help with visibility of that feature.

jvican:
I've just checked out these changes to try this feature out. The shift+f8 action works well and I can iterate across diagnostics of different buffers, but the f8 action that takes the `$view_uri` as an argument doesn't work for me. I don't have f8 mapped to anything so I'm inclined to think it's a bug.

htmue:
The "`f8` diagnostics" (for the current buffer) only come up when there are diagnostics for this file. This is kind of a new feature I actually already use all the time. First I check with `f8` if I'm done with a file and then I move on with `shift+f8`.

But I'm not sure if I understand you correctly. There are now the new "LSP: Goto Diagnostic..." / "LSP: Goto Diagnostic in Project..." items in the "Goto" menu. The first one should be enabled when there are diagnostics available for the current file, the second one when there are any diagnostics in the project. Could you please check if this is the case? Do they work?

jwortmann:
For me, the keybindings seem to work correctly. Maybe there should be a short status message "No diagnostics for active view" if there are no diagnostics, to make it clear. But I don't know how to achieve this, because to be shown for the keybinding, the status message would have to be set from within the `is_enabled()` method, but then I think it would also appear everytime when the command palette is opened. It might work if some of the logic is moved from `is_enabled()` to `input()` and then call the status message from there and just return `None` if there are no diagnostics.

Also, for me the padding between the list items is different dependent on whether the command was triggered by a keybinding / from the menu, or from the command palette. But I assume this must be a ST core bug and nothing we could control. Here with the adaptive theme:

* from command palette:
commandpalette

* from keybinding & menu:
keybinding

Sometimes the description in the details pane at the bottom is cut off too, if it spans multiple lines, but iirc I've already seen a report in the ST issue tracker for that.

And sorry to bother again, but would it look better to remove the codeDescription from the annotation of the list items and only show the server name (source)? Currently the content of the details pane is basically the same of what is shown already in the list items, and one reason for the ListItemHandler was to save some space in the items. But if you/others prefer how it is now, feel free to leave it there.

rchl (commented):
What I would really like to see implemented is restoration of the original selection on canceling the list. Right now the cursor stays on last previewed location.

Default.sublime-keymap
```diff
@@ -252,6 +252,33 @@
...
+    // {
+    //     "command": "lsp_goto_diagnostic",
+    //     "keys": [
+    //         "shift+f8"
+    //     ],
+    //     "context": [
+    //         {
+    //             "key": "setting.lsp_active"
+    //         }
+    //     ]
```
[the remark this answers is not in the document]
htmue: It is supposed to work for all files in project, even for those without a session, and it does.

htmue:
The annotation is actually the `source` (name of server or server plugin, depending on the server) and the `code`, not the `codeDescription`. I would leave it as it is. This is helpful overview information and when there is not enough space it's moved out of sight anyway.

The padding bug I don't see on the Mac, but sometimes the preview pane partly covers the items. The layout engine seems to be a little buggy.

As I already mentioned: As far as I'm concerned, the feature is finished. There is certainly room for improvements (I have mentioned a few points myself), but for now I'm done with it.

This is a voluntary contribution, take it or leave it.

Default.sublime-keymap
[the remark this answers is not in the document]
rchl: It does work indeed but this context check is wrong because, if it would work it would only allow this command to run when active file has active session.

The reason it doesn't cause problems is that this is a `WindowCommand` so our custom `on_query_context` handler is not even queried and the check passes.

rchl:
It's a great change and thanks for you contribution. We can improve some small things after this is merged if you don't have more resources to work on it.

Default.sublime-keymap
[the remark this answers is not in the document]
rchl: Or more precisely, the `setting.lsp_active` key is internally matched against the view's settings object but since this is not a TextCommand, it doesn't do what it looks like.

plugin/core/diagnostics_manager.py
```diff
@@ -17,44 +20,83 @@ class DiagnosticsManager(OrderedDict):
...
+        Returns possibly empty list of diagnostic for `uri`.
+        """
+        return self.get(uri, [])
+
+
+def severity_count(severity: int) -> Callable[[List[Diagnostic]], int]:
+    def severity_count(diagnostics: List[Diagnostic]) -> int:
+        return len(list(filter(has_severity(severity), diagnostics)))
+
+    return severity_count
```
rwols, line 88: Can't this be a lambda?

```suggestion
def severity_count(severity: int) -> Callable[[List[Diagnostic]], int]:
    return lambda diagnostics: len(list(filter(has_severity(severity), diagnostics)))
```

plugin/core/diagnostics_manager.py
```diff
@@ -17,44 +20,83 @@ class DiagnosticsManager(OrderedDict):
...
         return diagnostic_severity(diagnostic) == severity
+
     return has_severity
+
+
+def is_severity_included(max_severity: int) -> Callable[[Diagnostic], bool]:
+    def severity_included(diagnostic: Diagnostic) -> bool:
+        return diagnostic_severity(diagnostic) <= max_severity
+
+    return severity_included
```
rwols, line 102: Lambda?

```suggestion
def is_severity_included(max_severity: int) -> Callable[[Diagnostic], bool]:
    return lambda diagnostics: diagnostic_severity(diagnostic) <= max_severity
```

plugin/core/diagnostics_manager.py
```diff
@@ -17,44 +20,83 @@ class DiagnosticsManager(OrderedDict):
...
+        return len(list(filter(has_severity(severity), diagnostics)))
+
+    return severity_count
+
 
 def has_severity(severity: int) -> Callable[[Diagnostic], bool]:
     def has_severity(diagnostic: Diagnostic) -> bool:
         return diagnostic_severity(diagnostic) == severity
+
     return has_severity
```
rwols, line 95: Maybe turn this into a lambda as well?

```suggestion
def has_severity(severity: int) -> Callable[[Diagnostic], bool]:
    return lambda diagnostic: diagnostic_severity(diagnostic) == severity
```

plugin/goto_diagnostic.py
```diff
@@ -0,0 +1,312 @@
...
+    .diagnostics.info {background-color: color(var(--bluish) alpha(0.25))}
+    .diagnostics.hint {background-color: color(var(--bluish) alpha(0.25))}
+    """
+
+
+def get_sessions(window: sublime.Window) -> Iterator[Session]:
+    wm = windows.lookup(window)
+    if wm is not None:
+        for session in wm._sessions:
+            yield session
```
rwols, line 38:
```suggestion
        yield from wm._sessions
```

plugin/core/diagnostics_manager.py
htmue: These I would leave as they are because it's immediately clear from the layout that they are just curried functions, with lambdas this would be a little harder to get.

A proper `@curry` decorator would be nice, though:
```python
@curry
def is_severity_included(max_severity: ..., diagnostic: ...) -> ...:
    return diagnostic_severity(diagnostic) <= max_severity
```
There is one in the `toolz` package, but it seems overkill here and I don't know how well it plays with `typing`.

rwols: I'm personally not accustomed to the use of curried functions. I see a callable as return type and would expect a `lambda` if it's a single line. I think this is true for a lot of Python (or imperative) developers. But, I guess your argumentation makes sense as well.

rwols (approved):
I was a bit weary of yet another dependency (pathlib), but it's included in the std library for py38 anyway.

plugin/core/diagnostics_manager.py
rchl: One downside of lambdas is that those can't be type-annotated properly.

FichteFoll: I would recommend `functools.partial`, but iirc mypy also had problems with that. No idea about pyright.

Currying is really cool, but it's not native to Python and will forever feel alien unless it becomes integrated and more Python developers get to know it.

PR #1888, opened 2021-11-02
Don't resolve code lenses if server doesn't support resolving

### LINKED ISSUE #1886: Error handling None in Console
**Describe the bug**
Console gets filed with error messages

**To Reproduce**
1. Install LSP, LSP-jdtls
2. Open Java file

**Expected behavior**
Nothing

**Screenshots**
none

- OS: Manjaro Linux
- Sublime Text version: 4121
- LSP version: 1.13

**Additional context**
I installed LSP-file-watcher-chokidar

**Log**
[log.txt](https://github.com/sublimelsp/LSP/files/7453704/log.txt)

rchl:
Also show the error text when resolving code lens fails.

We could also make the code lens text red when resolving fails but not sure it's worth it for that corner case.

Fixes #1886

### DIFF
--- a/plugin/code_lens.py
+++ b/plugin/code_lens.py
@@ -1,5 +1,5 @@
-from .core.protocol import CodeLens, Range
-from .core.typing import List, Tuple, Dict, Iterable, Generator
+from .core.protocol import CodeLens, Error, Range
+from .core.typing import List, Tuple, Dict, Iterable, Generator, Union
 from .core.registry import LspTextCommand
 from .core.registry import windows
 from .core.views import make_command_link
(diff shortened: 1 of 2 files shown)


rwols (approved):

rchl:
For the record I'll just add that the main fix that actually fixes the reported issue is this:

```py
    def _on_code_lenses_resolved_async(self, mode: str) -> None:
        if self._is_listener_alive():
            sublime.set_timeout(lambda: self._code_lenses.render(mode))
```

So basically checking if the view listener is still alive after long-running request, and not handling the response if it isn't.

This is a pretty generic problem that we might want to find a generic solution for.

PR #1892, opened 2021-11-06
report didChangeWatchedFiles capability in a correct place

rchl:


predragnikolic (approved):

rwols (approved):
d'oh! Something similar is probably wrong with client-initiated progress somewhere. I should look into that...

rchl:
That one seems to match the spec, if that is what you mean:

Screenshot 2021-11-06 at 20 05 53

rchl:
And for the client initiated there isn't a client capability.

rwols:
That's **server-initiated** progress you're quoting

rwols:
for client-initiated progress we're sending a token along with each request, and pyright is supposed to use that token, but it starts using server-initiated progress anyway for textDocument/references. It should use the provided client token but it doesn't for some reason.

rchl:
I believe I found the issue in pyright https://github.com/microsoft/pyright/issues/2545

PR #1893, opened 2021-11-06
Don't restart all servers on changing LSP-* plugin settings

rchl:
Ensure that only the server who's settings were saved is restarted.

This of course only addresses changes in LSP-* settings, not
client settings within the main LSP settings.

Also consolidated almost identical `WindowManager._end_sessions_async`
and `WindowManager.end_config_sessions_async` into one.

### DIFF
--- a/plugin/configuration.py
+++ b/plugin/configuration.py
@@ -47,7 +47,7 @@ class LspDisableLanguageServerGloballyCommand(sublime_plugin.WindowCommand):
             config_name = self._items[index]
             client_configs.disable(config_name)
             wm = windows.lookup(self.window)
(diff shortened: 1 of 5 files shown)


rwols (approved):
Nice 👍

plugin/configuration.py
```diff
@@ -47,7 +47,7 @@ def _on_done(self, index: int) -> None:
             config_name = self._items[index]
             client_configs.disable(config_name)
             wm = windows.lookup(self.window)
-            sublime.set_timeout_async(lambda: wm.end_config_sessions_async(config_name))
+            sublime.set_timeout_async(lambda: wm._end_sessions_async(config_name))
```
rchl, line 50: Ops, I guess it should have been a public method.

PR #1895, opened 2021-11-07
Simplify Vala Language Server instructions

colinkiama:


docs/src/language_servers.md
```diff
@@ -561,9 +562,6 @@ Follow installation instructions on [LSP-volar](https://github.com/sublimelsp/LS
     }
     ```
 
-4. Enable the Vala Langauge Server for the project by going to `Tools > LSP > Enable Language Server In Project...`
-5. For the server to fully understand your code, you will need to generate a `compile_commands.json` file or build your project with [meson](https://mesonbuild.com/).
```
rchl, line 565: What about this point? Is it not relevant (anymore)?

colinkiama: Step 4 isn't relevant anymore. I wrote that based on the instructions for other language servers I was using at time. 

Over time I learned that Vala Language Server detects build files automatically and work just fine by enabling it globally once. It saves so much time compared to enabling it for every new project that you create.

However, you are right about step 5. My mistake. This is still necessary for it to function correctly. 

I'm not sure that I would consider it a step anymore though. It's seems more like important info or a warning to me. Maybe this should be added as a warning call-out instead?

rwols: A call-out (admonition?) sounds good to me.

rwols (approved):

PR #1896, opened 2021-11-07
Include diagnostics when triggering codeAction request on save

rchl:
Some servers might apply different logic depending on provided
diagnostics. For example theia server might not want to apply
destructive action like removing import statements when there are
error diagnostics.

While testing this new logic, uncovered a bug where we didn't include
diagnostics "touching" the selection region which also affected the
case when getting diagnostics for the whole document range on save,
which didn't include diagnostics that were at the end of the document.

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -24,7 +24,7 @@ from .core.types import SettingsRegistration
 from .core.typing import Any, Callable, Optional, Dict, Generator, Iterable, List, Tuple, Union
 from .core.url import parse_uri
 from .core.url import view_to_uri
-from .core.views import diagnostic_severity, region_to_range
+from .core.views import diagnostic_severity
 from .core.views import document_color_params
 from .core.views import first_selection_region
 from .core.views import format_completion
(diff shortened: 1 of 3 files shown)


plugin/documents.py
```diff
@@ -252,10 +252,13 @@ def diagnostics_intersecting_region_async(
...
         for sb, diagnostics in self.diagnostics_async():
             intersections = []  # type: List[Diagnostic]
             for diagnostic, candidate in diagnostics:
-                if region.intersects(candidate):
+                # Perform intersection checks on Range since it's inclusive unlike Region which is exclusive.
+                diagnostic_range = region_to_range(self.view, candidate)
+                if requested_range.intersects(diagnostic_range):
                     covering = covering.cover(candidate)
                     intersections.append(diagnostic)
             if intersections:
```
rchl, line 264: This whole function (and related) could probably be refactored to work on `Range` exclusively but that's too much of a job for now.

rwols (commented):
My idea was to convert LSP ranges to ST regions as soon as possible, and work with ST regions as much as possible. And then remove the Range/Point classes. I think this can also be solved by calling `diagnostics_touching_point` somewhere alongside `diagnostics_intersecting_region` if the region turns out to be (n, n) instead of (n, m) with n < m.

It seems like you prefer working with LSP ranges instead. That's fine, but we should make a decision going forward. I think both have their pros and cons.

Pros of using ST regions:
- Less maintenance
- ST API functions expect ST regions
- Converting an LSP range into an ST region depends on the change_count of the view... so we must do that at the point in time where the LSP range's change_count is still equal to the view's change_count

Cons of using ST regions:
- Has different behavior for single points and regions (n, m) with n < m

Pros of using LSP ranges:
- Has nice behavior whether regions are points or not
- When sending back diagnostics for code actions, there is no need to convert ST regions to LSP ranges

Cons of using LSP ranges:
- A bit more maintenance
- Requires conversion to an ST region when wanting to use it in an ST API function. The conversion depends on the view's change_count

There is much confusion about the implementation of ST regions, and I don't totally agree with the ideas/concepts, but it is what it is. https://github.com/sublimehq/sublime_text/issues/3837

rwols:
See also https://en.wikipedia.org/wiki/Off-by-one_error#Fencepost_error

rchl:
> My idea was to convert LSP ranges to ST regions as soon as possible, and work with ST regions as much as possible. And then remove the Range/Point classes. I think this can also be solved by calling `diagnostics_touching_point` somewhere alongside `diagnostics_intersecting_region` if the region turns out to be (n, n) instead of (n, m) with n < m.

Before I think about it more, are you suggesting that this is only an issue when region we are checked against is (n, n)? Because it's not. If we have selection region (1, 5) then this doesn't intersect with diagnostic at (5, 6) if we use Region for checking.

rchl:
Changed to use `contains` check against region points.

rwols (approved):

PR #1897, opened 2021-11-12
Use new UnitTesting github actions

rchl:
Not sure about coverage since I don't see an option to pass the token. But then I don't think we actually care about coverage.

@randy3k


rchl:
Well, coverage works without the token. Token is only needed if we want to upload it and have the app make a comment. Do we want to make that work @rwols ?

randy3k:
Codecov token is not needed for public repo. With the said, we should have the options to specify it (in case there are private Sublime Text package).

.github/workflows/main.yml
```diff
@@ -16,62 +16,22 @@ on:
...
-      - run: curl -OL https://raw.githubusercontent.com/SublimeText/UnitTesting/master/sbin/github.sh
-      - run: python3 tests/server.py --version
-      - run: python3 tests/adjust_unittesting_config_for_ci.py && cat unittesting.json
-      - run: |
-          sh github.sh bootstrap
-          sh github.sh install_package_control
-          sh github.sh run_tests
-        name: Run UnitTesting tests
+          coverage: true
+          odecov-upload: true
```
rwols, line 34: Is this a typo?

rwols:
I guess it's okay to work without coverage. It doesn't work well anyway because the coverage is not computed from code running in the worker thread from what I remember.

randy3k:
Looks like there is a bug in UnitTesting preventing upload: https://github.com/SublimeText/UnitTesting/commit/4ba3426b87490a2a197c4ab64ca4868a7753ed99

rchl:
The `codecov-upload` didn't seem to result in codecov comment being added here but we don't want it anyway so just removing it (and the codecov ENV).

rchl:
> I guess it's okay to work without coverage. It doesn't work well anyway because the coverage is not computed from code running in the worker thread from what I remember.

Yeah, the coverage numbers for sessions.py are pretty low:

>  plugin/core/sessions.py                    843    745    12%

so I guess you are right.

We still have the report in the output but won't be using the upload. Thanks @randy3k for fixing anyway.

randy3k:
`coverage.py` could only trace high level threads created by `threading` library but not lower level threads used by ST worker.

randy3k:
Just pushed a [workaround](https://github.com/SublimeText/UnitTesting/commit/1397b1cfd546171238c16369143ccc502e33ba36) to detect coverage in the worker thread (inspired by https://github.com/nedbat/coveragepy/issues/686).

rchl:
It doesn't seem like the coverage has improved - https://github.com/sublimelsp/LSP/runs/4197708709?check_suite_focus=true

Assuming that I've tested it properly.

randy3k:
> It doesn't seem like the coverage has improved - [sublimelsp/LSP/runs/4197708709?check_suite_focus=true](https://github.com/sublimelsp/LSP/runs/4197708709?check_suite_focus=true)
> 
> ~Assuming that I've tested it properly.~ EDIT: I see, there is a new setting for that.

Um, it is because I am still not very understand about the workaround. So better to make it off first.

rchl:
> So better to make it off first.

Sure, just checking it out for fun.

randy3k:
Definitely a coverage increment here.

rchl:
Now it's better:

> plugin/core/sessions.py                    843    482    43%

Not sure how accurate it is now but certainly better.

randy3k:
The [codecov](https://codecov.io/gh/sublimelsp/LSP/commit/9e5c3d966b1189033e61e05286a81de58d9c9d8d) report is more readable.

rchl:
Strange to me that the lines that declare a function or method (`def foo(...):`) and imports are always not covered.

But still could be useful to track relative changes, of course.

randy3k:
You have [turned off](https://github.com/sublimelsp/LSP/blob/main/unittesting.json#L6) `reload_package_on_testing`.

The code is loaded by Sublime before coverage starts. We'll need to reload the package to get those lines covered.

rchl:
I think reloading the package makes things a bit unstable...

randy3k:
> I think reloading the package makes things a bit flaky...

It could be. However, I think the Windows failure is kind of random error.

rchl:
> It could be. However, I think the Windows failure is kind of random error.

but then I've re-run and mac failed

randy3k:
It could be the new workaround for worker thread. It will definitely make things slower.

rwols (approved):

PR #1839, opened 2021-09-03
Semantic highlighting (v2)

jwortmann:
This is an alternative to #1751, but with support for more request types and additional features.

There is still a lot to optimize, for example there is a short flickering because all regions are redrawn after each request. Also I'm not happy yet with how token modifiers are taken into account - there probably should be a way for the helper plugins to consider them too, and for color schemes to target them. I left some "TODO" comments in the code where I got stuck a bit; suggestions or help appreciated.
I haven't yet considered the region draw order (so that it works together with diagnostics etc.) and there might be some other bugs around - just wanted to open it here already, so that those who are interested could take a look or provide suggestions. And in general, I'm not sure if this add_regions approach for semantic highlighting is something we want, or if we shouldn't better wait for a dedicated API (might possibly never be implemented though). 

In general it seems to work with clangd, but I have only tested with very simple examples because I'm not very familiar with C++. Maybe some tweaks for the default scopes are needed too, probably requires some experience with more servers about how token types are used.
I couldn't get it really to work with the Lua server, because that server immediately unregisters the semantic token capability after it initializes.

PR #1899, opened 2021-11-16
Fix color boxes not updating in some cases

rchl:
As was noticed in https://github.com/sublimelsp/LSP/pull/1839#discussion_r750288905, it was incorrect to use `_when_selection_remains_stable_async` for the features that relied on view changes being reported first.

The `_when_selection_remains_stable_async` would cause the request to be skipped if the selection has changed since it was called. That means that changing the document and then quickly moving the cursor would result in color boxes not being requested/updated.

And on top of that, we might have also requested color boxes before the `didChange` notification has been triggered thus resulting in out of sync positioning of boxes.

Moved the logic to `SessionBuffer` where it can be called after the `didChange` notification and added checks if view is still alive and whether the document has remained the same on getting results from session.

We also shouldn't need any extra debouncing as `didChange` notifications are already debounced.

As a side effect, we'll request color boxes from all active sessions that support them rather then only the first one. This might be good or bad but if it's an issue, the user can disable relevant capability for one of the sessions.

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -25,10 +25,8 @@ from .core.typing import Any, Callable, Optional, Dict, Generator, Iterable, Lis
 from .core.url import parse_uri
 from .core.url import view_to_uri
 from .core.views import diagnostic_severity
-from .core.views import document_color_params
 from .core.views import first_selection_region
 from .core.views import format_completion
-from .core.views import lsp_color_to_phantom
 from .core.views import make_command_link
 from .core.views import range_to_region
 from .core.views import show_lsp_popup
@@ -176,7 +174,6 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         self._manager = None  # type: Optional[WindowManager]
         self._session_views = {}  # type: Dict[str, SessionView]
         self._stored_region = sublime.Region(-1, -1)
-        self._color_phantoms = sublime.PhantomSet(self.view, "lsp_color")
         self._sighelp = None  # type: Optional[SigHelp]
         self._registered = False
 
@@ -186,7 +183,6 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         triggers = [trigger for trigger in triggers if 'server' not in trigger]
         settings.set("auto_complete_triggers", triggers)
         self._stored_region = sublime.Region(-1, -1)
-        self._color_phantoms.update([])
         self.view.erase_status(AbstractViewListener.TOTAL_ERRORS_AND_WARNINGS_STATUS_KEY)
         self._clear_highlight_regions()
         self._clear_session_views_async()
@@ -224,7 +220,6 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
             self.view.settings().set("lsp_active", True)
             added = True
         if added:
-            self._do_color_boxes_async()
             self._do_code_lenses_async()
 
     def on_session_shutdown_async(self, session: Session) -> None:
@@ -322,8 +317,6 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         if userprefs().document_highlight_style:
             self._when_selection_remains_stable_async(self._do_highlights_async, current_region,
                                                       after_ms=self.highlights_debounce_time)
-        self._when_selection_remains_stable_async(self._do_color_boxes_async, current_region,
-                                                  after_ms=self.color_boxes_debounce_time)
         self.do_signature_help_async(manual=False)
         self._when_selection_remains_stable_async(self._do_code_lenses_async, current_region,
                                                   after_ms=self.code_lenses_debounce_time)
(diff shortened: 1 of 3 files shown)


plugin/session_buffer.py
```diff
@@ -265,6 +273,32 @@ def some_view(self) -> Optional[sublime.View]:
             return sv.view
         return None
 
+    def _if_view_version_stable(self, f: Callable[[sublime.View, Any], None], version: int) -> Callable[[Any], None]:
```
predragnikolic, line 276: `_view_remained_unchanged`
rename if you find the suggested name more suitable

plugin/documents.py
```diff
@@ -322,8 +317,6 @@ def on_text_changed_async(self, change_count: int, changes: Iterable[sublime.Tex
         if userprefs().document_highlight_style:
             self._when_selection_remains_stable_async(self._do_highlights_async, current_region,
                                                       after_ms=self.highlights_debounce_time)
-        self._when_selection_remains_stable_async(self._do_color_boxes_async, current_region,
-                                                  after_ms=self.color_boxes_debounce_time)
         self.do_signature_help_async(manual=False)
         self._when_selection_remains_stable_async(self._do_code_lenses_async, current_region,
```
predragnikolic, line 321: The same fix can be applied on `_do_code_lenses_async` but that can be done in a separate PR.

plugin/session_buffer.py
rchl: I think it reads better at use-site with "if" but I've adapted part of your suggestion.

plugin/session_buffer.py
```diff
@@ -265,6 +273,32 @@ def some_view(self) -> Optional[sublime.View]:
...
+
+        return handler
+
+    # --- textDocument/documentColor -----------------------------------------------------------------------------------
+
+    def _do_color_boxes_async(self, view: sublime.View, version: int) -> None:
+        if self.session.has_capability("colorProvider"):
+            self.session.send_request_async(
+                Request.documentColor(document_color_params(view), view),
+                self._if_view_unchanged(self._on_color_boxes_async, version)
```
predragnikolic, line 293: Currently the request will be send regardless if the view is unchanged. 
I think it would be a better to send the `documentColor` request only when the view is unchanged. (to avoid causing unnecessary tragic to the server)

----
Currently `_do_color_boxes_async` is called from `purge_changes_async`,
`purge_changes_async` is called in `on_text_changed_async`, 
and `purge_changes_async` is debounced there, 
so we won't hit the problem I described above.

predragnikolic (changes requested):

plugin/session_buffer.py
rchl: `_do_color_boxes_async` is only triggered after didOpen and didChange notifications so that implies that the view just changed and is stable. Isn't it?

predragnikolic: _do_color_boxes_async is also called from purge_changes_async.

purge_changes_async is called in on_text_changed_async,
But purge_changes_async is debounced there, so we won't hit the problem I described above...
Still I think it would be preferred to not send the request at all until the vue remains unchanged.

What do you think?

rchl: > _do_color_boxes_async is also called from purge_changes_async.

That's where the `didChange` notification is sent and it's only sent when there were pending changes that we just purged and at this point the change_count() is at latest state.

So again, I'm not getting what you mean really...

predragnikolic: Hahaha, I wrote vue instead of view... what a brain bug 😀

predragnikolic: > So again, I'm not getting what you mean really...

Is there a need to handle the response only when the view remained unchanged? 
Can this line could be changed  to:
```diff
            self.session.send_request_async(
                Request.documentColor(document_color_params(view), view),
-                self._if_view_unchanged(self._on_color_boxes_async, version)
+                self._on_color_boxes_async
```

It seems unnecessary to fire the request and to handle response **only** when the view is unchanged:

```py
            self.session.send_request_async(
                Request.documentColor(document_color_params(view), view),
                self._if_view_unchanged(self._on_color_boxes_async, version) #  handle the response only if the view remained unchanged
```

rchl: The view is unchanged at the point of making the request. That can change by the time the response arrives so we have to check.

predragnikolic (approved):

rwols (changes requested):
Shouldn’t the PhantomSets go into the SessionView? I think the color regions are correctly stored in SessionBuffer but now clones won’t show the same phantoms right? Maybe they didn’t before as well.

rchl:
> Shouldn’t the PhantomSets go into the SessionView? I think the color regions are correctly stored in SessionBuffer but now clones won’t show the same phantoms right? Maybe they didn’t before as well.

How do we create clones again? Was it "Split View"? If so then it works as is. My best guess is that even though ST provides the api on the `View`, it internally ties phantoms to buffers and we are doing the right thing by creating them once instead of for each view clone.

rwols (approved):
In that case lgtm

PR #1900, opened 2021-11-18
Send range with textDocument/hover when possible

ayoub-benali:
This relies on `experimental.rangeHoverProvider` which so far only Metals implement. But this workaround would be removed once https://github.com/microsoft/language-server-protocol/issues/377 is part of the LSP specification.

Peek 2021-11-18 14-42

Fixes https://github.com/scalameta/metals-sublime/issues/44

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -272,8 +272,13 @@ def text_document_position_params(view: sublime.View, location: int) -> TextDocu
     return {"textDocument": text_document_identifier(view), "position": position(view, location)}
 
 
-def text_document_range_params(view: sublime.View, region: sublime.Region) -> ExperimentalTextDocumentRangeParams:
-    return {"textDocument": text_document_identifier(view), "range": region_to_range(view, region).to_lsp()}
+def text_document_range_params(view: sublime.View, location: int,
+                               region: sublime.Region) -> ExperimentalTextDocumentRangeParams:
+    return {
+        "textDocument": text_document_identifier(view),
+        "position": position(view, location),
+        "range": region_to_range(view, region).to_lsp()
+    }
 
 
 def did_open_text_document_params(view: sublime.View, language_id: str) -> Dict[str, Any]:
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -270,6 +270,8 @@ def versioned_text_document_identifier(view: sublime.View, version: int) -> Dict
 def text_document_position_params(view: sublime.View, location: int) -> TextDocumentPositionParams:
     return {"textDocument": text_document_identifier(view), "position": position(view, location)}
 
+def text_document_range_params(view: sublime.View, region: sublime.Region) -> Dict[str, Any]:
+    return {"textDocument": text_document_identifier(view), "range": region_to_range(view, region).to_lsp()}
 
 def did_open_text_document_params(view: sublime.View, language_id: str) -> Dict[str, Any]:
     return {"textDocument": text_document_item(view, language_id)}
@@ -404,6 +406,8 @@ def text_document_code_action_params(
 # Workaround for a limited margin-collapsing capabilities of the minihtml.
 LSP_POPUP_SPACER_HTML = '<div class="lsp_popup--spacer"></div>'
 
+def hide_lsp_popup(view: sublime.View) -> None:
+    mdpopups.hide_popup(view)
 
 def show_lsp_popup(view: sublime.View, contents: str, location: int = -1, md: bool = False, flags: int = 0,
                    css: Optional[str] = None, wrapper_class: Optional[str] = None,
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -406,9 +406,6 @@ def text_document_code_action_params(
 # Workaround for a limited margin-collapsing capabilities of the minihtml.
 LSP_POPUP_SPACER_HTML = '<div class="lsp_popup--spacer"></div>'
 
-def hide_lsp_popup(view: sublime.View) -> None:
-    mdpopups.hide_popup(view)
-
 def show_lsp_popup(view: sublime.View, contents: str, location: int = -1, md: bool = False, flags: int = 0,
                    css: Optional[str] = None, wrapper_class: Optional[str] = None,
                    on_navigate: Optional[Callable] = None, on_hide: Optional[Callable] = None) -> None:
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -270,8 +270,10 @@ def versioned_text_document_identifier(view: sublime.View, version: int) -> Dict
 def text_document_position_params(view: sublime.View, location: int) -> TextDocumentPositionParams:
     return {"textDocument": text_document_identifier(view), "position": position(view, location)}
 
+
 def text_document_range_params(view: sublime.View, region: sublime.Region) -> Dict[str, Any]:
     return {"textDocument": text_document_identifier(view), "range": region_to_range(view, region).to_lsp()}
+
 
 def did_open_text_document_params(view: sublime.View, language_id: str) -> Dict[str, Any]:
     return {"textDocument": text_document_item(view, language_id)}
@@ -405,6 +407,7 @@ def text_document_code_action_params(
 
 # Workaround for a limited margin-collapsing capabilities of the minihtml.
 LSP_POPUP_SPACER_HTML = '<div class="lsp_popup--spacer"></div>'
+
 
 def show_lsp_popup(view: sublime.View, contents: str, location: int = -1, md: bool = False, flags: int = 0,
                    css: Optional[str] = None, wrapper_class: Optional[str] = None,
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -271,10 +271,6 @@ def text_document_position_params(view: sublime.View, location: int) -> TextDocu
     return {"textDocument": text_document_identifier(view), "position": position(view, location)}
 
 
-def text_document_range_params(view: sublime.View, region: sublime.Region) -> Dict[str, Any]:
-    return {"textDocument": text_document_identifier(view), "range": region_to_range(view, region).to_lsp()}
-
-
 def did_open_text_document_params(view: sublime.View, language_id: str) -> Dict[str, Any]:
     return {"textDocument": text_document_item(view, language_id)}
 
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -5,6 +5,7 @@ from .protocol import Diagnostic
 from .protocol import DiagnosticRelatedInformation
 from .protocol import DiagnosticSeverity
 from .protocol import DocumentUri
+from .protocol import ExperimentalTextDocumentRangeParams
 from .protocol import Location
 from .protocol import LocationLink
 from .protocol import MarkedString
@@ -269,6 +270,10 @@ def versioned_text_document_identifier(view: sublime.View, version: int) -> Dict
 
 def text_document_position_params(view: sublime.View, location: int) -> TextDocumentPositionParams:
     return {"textDocument": text_document_identifier(view), "position": position(view, location)}
+
+
+def text_document_range_params(view: sublime.View, region: sublime.Region) -> ExperimentalTextDocumentRangeParams:
+    return {"textDocument": text_document_identifier(view), "range": region_to_range(view, region).to_lsp()}
 
 
 def did_open_text_document_params(view: sublime.View, language_id: str) -> Dict[str, Any]:
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -93,9 +93,6 @@ class LspHoverCommand(LspTextCommand):
         if not window:
             return
         hover_point = temp_point
-        window = self.view.window()
-        if not window:
-            return
         wm = windows.lookup(window)
         self._base_dir = wm.get_project_path(self.view.file_name() or "")
         self._hover_responses = []  # type: List[Hover]
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -20,9 +20,9 @@ from .core.views import FORMAT_MARKED_STRING, FORMAT_MARKUP_CONTENT, minihtml
 from .core.views import is_location_href
 from .core.views import make_command_link
 from .core.views import make_link
+from .core.views import region_to_range
 from .core.views import show_lsp_popup
 from .core.views import text_document_position_params
-from .core.views import text_document_range_params
 from .core.views import unpack_href_location
 from .core.views import update_lsp_popup
 from .core.windows import AbstractViewListener
@@ -129,7 +129,8 @@ class LspHoverCommand(LspTextCommand):
                 if region is not None:
                     if region.contains(point):
                         # hovering selection or only text was selected
-                        document_position = text_document_range_params(self.view, region)
+                        document_position.pop('position', None)
+                        document_position['range'] = region_to_range(self.view, region).to_lsp()
             hover_promises.append(session.send_request_task(
                 Request("textDocument/hover", document_position, self.view)
             ))
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -121,8 +121,8 @@ class LspHoverCommand(LspTextCommand):
 
     def request_symbol_hover_async(self, listener: AbstractViewListener, point: int) -> None:
         hover_promises = []  # type: List[Promise[ResolvedHover]]
-        document_position = text_document_position_params(self.view, point)
         for session in listener.sessions_async('hoverProvider'):
+            document_position = text_document_position_params(self.view, point)
             range_hover_provider = session.get_capability('experimental.rangeHoverProvider')
             if range_hover_provider:
                 region = first_selection_region(self.view)
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -20,8 +20,10 @@ from .core.views import FORMAT_MARKED_STRING, FORMAT_MARKUP_CONTENT, minihtml
 from .core.views import is_location_href
 from .core.views import make_command_link
 from .core.views import make_link
+from .core.views import hide_lsp_popup
 from .core.views import show_lsp_popup
 from .core.views import text_document_position_params
+from .core.views import text_document_range_params
 from .core.views import unpack_href_location
 from .core.views import update_lsp_popup
 from .core.windows import AbstractViewListener
@@ -35,7 +37,7 @@ import webbrowser
 SUBLIME_WORD_MASK = 515
 SessionName = str
 ResolvedHover = Union[Hover, Error]
-
+PointOrRegion = Union[int, sublime.Region]
 
 _test_contents = []  # type: List[str]
 
@@ -71,6 +73,8 @@ link_kinds = [
 
 class LspHoverCommand(LspTextCommand):
 
+    range_hover_capability = 'experimental.rangeHoverProvider'
+
     def __init__(self, view: sublime.View) -> None:
         super().__init__(view)
         self._base_dir = None   # type: Optional[str]
@@ -82,17 +86,28 @@ class LspHoverCommand(LspTextCommand):
         point: Optional[int] = None,
         event: Optional[dict] = None
     ) -> None:
-        temp_point = point
-        if temp_point is None:
+        range_hover_provider = False
+        session = self.best_session(self.range_hover_capability)
+        if session:
+            capability = session.get_capability(self.range_hover_capability)
+            if isinstance(capability, bool):
+                range_hover_provider = capability
+
+        # ignore selection if there a point is hovered
+        point_or_region = point # type: Optional[PointOrRegion]
+        if point_or_region is None:
             region = first_selection_region(self.view)
-            if region is not None:
-                temp_point = region.begin()
-        if temp_point is None:
+            if region is not None and not range_hover_provider:
+                point_or_region = region.begin()
+            elif region is not None and range_hover_provider and not region.empty():
+                point_or_region = region
+
+        if point_or_region is None:
             return
+
         window = self.view.window()
         if not window:
             return
-        hover_point = temp_point
         wm = windows.lookup(window)
         self._base_dir = wm.get_project_path(self.view.file_name() or "")
         self._hover_responses = []  # type: List[Hover]
@@ -106,21 +121,29 @@ class LspHoverCommand(LspTextCommand):
             if not listener:
                 return
             if not only_diagnostics:
-                self.request_symbol_hover_async(listener, hover_point)
-            self._diagnostics_by_config, covering = listener.diagnostics_touching_point_async(
-                hover_point, userprefs().show_diagnostics_severity_level)
-            if self._diagnostics_by_config:
-                self.show_hover(listener, hover_point, only_diagnostics)
-            if not only_diagnostics and userprefs().show_code_actions_in_hover:
-                actions_manager.request_for_region_async(
-                    self.view, covering, self._diagnostics_by_config,
-                    functools.partial(self.handle_code_actions, listener, hover_point))
+                self.request_symbol_hover_async(listener, point_or_region)
+            if isinstance(point_or_region, int):
+                hover_point = point_or_region
+                self._diagnostics_by_config, covering = listener.diagnostics_touching_point_async(
+                    hover_point, userprefs().show_diagnostics_severity_level)
+                if self._diagnostics_by_config:
+                    self.show_hover(listener, hover_point, only_diagnostics)
+                if not only_diagnostics and userprefs().show_code_actions_in_hover:
+                    actions_manager.request_for_region_async(
+                        self.view, covering, self._diagnostics_by_config,
+                        functools.partial(self.handle_code_actions, listener, hover_point))
 
         sublime.set_timeout_async(run_async)
 
-    def request_symbol_hover_async(self, listener: AbstractViewListener, point: int) -> None:
+    def request_symbol_hover_async(self, listener: AbstractViewListener, point_or_region: PointOrRegion) -> None:
         hover_promises = []  # type: List[Promise[ResolvedHover]]
-        document_position = text_document_position_params(self.view, point)
+        if isinstance(point_or_region, int):
+            document_position = text_document_position_params(self.view, point_or_region)
+            point = point_or_region
+        else:
+            document_position = text_document_range_params(self.view, point_or_region)
+            point = point_or_region.begin()
+
         for session in listener.sessions_async('hoverProvider'):
             hover_promises.append(session.send_request_task(
                 Request("textDocument/hover", document_position, self.view)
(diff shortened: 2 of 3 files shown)


plugin/hover.py
```diff
@@ -209,15 +232,15 @@ def _show_hover(self, listener: AbstractViewListener, point: int, only_diagnosti
...
             if self.view.is_popup_visible():
-                update_lsp_popup(self.view, contents)
-            else:
-                show_lsp_popup(
-                    self.view,
-                    contents,
-                    flags=sublime.HIDE_ON_MOUSE_MOVE_AWAY,
-                    location=point,
-                    on_navigate=lambda href: self._on_navigate(href, point))
+                hide_lsp_popup(self.view)
```
rchl, line 237: Shouldn't do that as that can result in popup position jumping while the content is arriving late.

I don't think you need this with the current implementation anyway?

plugin/core/views.py
```diff
@@ -270,6 +270,8 @@ def versioned_text_document_identifier(view: sublime.View, version: int) -> Dict
 def text_document_position_params(view: sublime.View, location: int) -> TextDocumentPositionParams:
     return {"textDocument": text_document_identifier(view), "position": position(view, location)}
 
+def text_document_range_params(view: sublime.View, region: sublime.Region) -> Dict[str, Any]:
+    return {"textDocument": text_document_identifier(view), "range": region_to_range(view, region).to_lsp()}
```
rchl, line 274: Did scala implementation change and now it accepts range in `range` parameter?
Also, if that's the case, you probably could also send `position` per normal?

plugin/core/views.py
```diff
@@ -404,6 +406,8 @@ def text_document_code_action_params(
 # Workaround for a limited margin-collapsing capabilities of the minihtml.
 LSP_POPUP_SPACER_HTML = '<div class="lsp_popup--spacer"></div>'
 
+def hide_lsp_popup(view: sublime.View) -> None:
+    mdpopups.hide_popup(view)
```
rchl, line 410: Shouldn't be needed.

plugin/hover.py
```diff
@@ -106,21 +121,29 @@ def run_async() -> None:
...
+        if isinstance(point_or_region, int):
+            document_position = text_document_position_params(self.view, point_or_region)
+            point = point_or_region
+        else:
+            document_position = text_document_range_params(self.view, point_or_region)
+            point = point_or_region.begin()
+
         for session in listener.sessions_async('hoverProvider'):
             hover_promises.append(session.send_request_task(
                 Request("textDocument/hover", document_position, self.view)
```
rchl, line 149: The capability check should be here so that with multiple servers we send range only to those that support it.

plugin/hover.py
```diff
@@ -82,17 +86,28 @@ def run(
...
         event: Optional[dict] = None
     ) -> None:
-        temp_point = point
-        if temp_point is None:
+        range_hover_provider = False
+        session = self.best_session(self.range_hover_capability)
+        if session:
+            capability = session.get_capability(self.range_hover_capability)
+            if isinstance(capability, bool):
+                range_hover_provider = capability
```
rchl, line 94: It's not correct to just get the "best_session" since it won't be accurate with multiple session running. The check should be later, when iterating sessions.

rchl (changes requested):

plugin/core/views.py
ayoub-benali: Only range or position should be sent https://scalameta.org/metals/docs/integrations/new-editor#textdocumenthover

rchl: OK.
I would think that having data on where the cursor is exactly could be useful but it's up to their implementation...

plugin/hover.py
ayoub-benali: We need this because the hover and the actual selection could at different places see: Peek 2021-11-18 14-42

rchl: How is that with this selection:

```
|toFloat.toString()|
```

You are getting different results when hovering `toFloat` and `toString()`. Wouldn't the implementation send the same range in both cases?

plugin/hover.py
```diff
@@ -106,22 +109,35 @@ def run_async() -> None:
...
-            if self._diagnostics_by_config:
-                self.show_hover(listener, hover_point, only_diagnostics)
-            if not only_diagnostics and userprefs().show_code_actions_in_hover:
-                actions_manager.request_for_region_async(
-                    self.view, covering, self._diagnostics_by_config,
-                    functools.partial(self.handle_code_actions, listener, hover_point))
+                self.request_symbol_hover_async(listener, point_or_region)
+            if isinstance(point_or_region, int):
+                hover_point = point_or_region
+                self._diagnostics_by_config, covering = listener.diagnostics_touching_point_async(
```
rchl, line 115: When having a region you should be using `diagnostics_intersecting_region_async` instead of disabling this code completely.

plugin/hover.py
```diff
@@ -106,22 +109,35 @@ def run_async() -> None:
...
+    def request_symbol_hover_async(self, listener: AbstractViewListener, point_or_region: PointOrRegion) -> None:
         hover_promises = []  # type: List[Promise[ResolvedHover]]
-        document_position = text_document_position_params(self.view, point)
+        point = point_or_region if isinstance(point_or_region, int) else point_or_region.begin()
+
         for session in listener.sessions_async('hoverProvider'):
+            capability = session.get_capability('experimental.rangeHoverProvider')
+            range_hover_provider = False
+            if isinstance(capability, bool):
+                range_hover_provider = capability
```
rchl, line 134: This should be enough:
```suggestion
            range_hover_provider = session.get_capability('experimental.rangeHoverProvider')
```
In the spec it's normal to also have objects as values which also should evaluate to capability being supported.

plugin/hover.py
```diff
@@ -106,22 +109,35 @@ def run_async() -> None:
...
+
+            if isinstance(point_or_region, sublime.Region) and range_hover_provider:
+                document_position = text_document_range_params(self.view, point_or_region)
+            else:
+                document_position = text_document_position_params(self.view, point_or_region)
+
```
rchl, line 140: Please don't add extra blank lines.
The style of the LSP code generally avoid blank lines within methods.

plugin/hover.py
rchl: Looking close, it seems like your implement almost never sends range actually since you only take a range if "point" is not provided in LspHoverCommand but it almost always is.

So once you fix that we can investigate more but I'm still claiming that you shouldn't be hiding the popup as that will also affect the normal behavior negatively.

ayoub-benali: Since there isn't an LSP spec the user behavior isn't defined. 
The current implementation sends only a hover point even when a selection is hovered. If you want to send the range you have to use the key binding. 
I guess you want the range to be sent if the user hovers a selection, right ?

@ckipp01 @tgodzik what is the user behavior in vscode and vim ?

tgodzik: > I guess you want the range to be sent if the user hovers a selection, right ?

That's right, currently VS Code plugin always sends a range even with no selection, but this should work the same if point is sent.

plugin/hover.py
ayoub-benali: Metals is sending a boolean for that field https://github.com/scalameta/metals/pull/3293/files#diff-51fd3540bc876e6301b46ef557c4be26f3cba4705ddf2e649520124f81abd02bR10
That is why I am doing the check

plugin/hover.py
rchl: Yeah, I thought that's the point of the future.

And IMO this should work as expected first and foremost for the mouse use case. That is, if selection is hovered it should send a range to the server. If there is no selection under the hover point then I guess it doesn't really matter what is sent (point or range), the result should be the same.

plugin/hover.py
rchl: Same in other places. Have a look at the complete diff here on GH.

plugin/hover.py
```diff
@@ -106,22 +107,38 @@ def run_async() -> None:
...
+            range_hover_provider = session.get_capability('experimental.rangeHoverProvider')
+            if range_hover_provider and isinstance(point_or_region, int):
+                for region in self.view.sel():
+                    if region.contains(point):
+                        # when hovering selection send it as range
+                        document_position = text_document_range_params(self.view, region)
+                    else:
+                        document_position = text_document_position_params(self.view, point_or_region)
+            else:
+                document_position = text_document_range_params(self.view, PointOrRegion)
```
rchl:
```suggestion
                document_position = text_document_position_params(self.view, PointOrRegion)
```

plugin/hover.py
rchl: So why would it be necessary to hide the popup manually now? ST should be hiding it iself when hovering new point.

ayoub-benali: Because then you get the pop at the wrong place.
1. Hover "App"
2. invoke `lsp_hover` for selection
3. pop up showing up in the wrong place

Peek 2021-11-18 16-10

ayoub-benali: It is an issue when using both. Generate a popup via mouse, keep the cursor in position and then invoke a hover via keyboard which leads to a popup showing in the wrong place. Hiding the popup fixes the problem because then we show the popup in the correct position

ayoub-benali: OK so `mdpopups.update_popup` https://facelessuser.github.io/sublime-markdown-popups/api/#update-popup doesn't take location as argument so we have to hide it in this scenario

rchl: So yes, it's an unrelated issue.
It should be fixed separately and in a different way.

ayoub-benali: It is even a sublime API issue :(  https://www.sublimetext.com/docs/api_reference.html#ver-dev `update_popup` takes only `content` as arugment

ckipp01: > what is the user behavior in vscode and vim ?

In `nvim-metals` if the user is in visual mode, it sends the range, if they are in normal mode it sends the point.

rchl: That's fine because you can close the popup before showing it. The issue is not with closing the popup manually but with **when** it's closed. We don't want to close it when we've shown the popup with content from one session and we are waiting for the rest of the sessions. So this needs to be solved differently. Possibly by just closing the popup at the beginning of `LspHoverCommand`.

rchl: But I don't want this unrelated fix to be mixed into this PR.

ayoub-benali: ok will remove `hide_lsp_popup` and will create a separate issue once this is merged

jwortmann: > 1. Hover "App" with mouse, popup shows up, keep cursor in place.
> 2. invoke `lsp_hover` for selection via key binding
> 3. Notice pop up showing up in the wrong place. The popup is still on "App" while it should move to "|toFloat.toString()|"

Shouldn't the popup for `lsp_hover` always be at the place where the mouse cursor is? So I would say when doing these steps, `lsp_hover` should ignore the selected range and still send the point from "App", regardless whether triggered by keybinding or not. Otherwise it wouldn't be "hover" anymore. Instead there should better be a new command for it (and ideally it shouldn't be handled by the `textDocument/hover` request if it's unrelated to mouse hover).

jwortmann: Maybe I'm wrong because I interpreted the "hover" to always be related to the mouse hovering over a position, but what actually might be meant is that the popup is displayed hovering over the text ;)

ayoub-benali: Please see https://github.com/microsoft/language-server-protocol/issues/377 the spec will most likely use `textDocument/hover`

jwortmann: But from that discussion it seems my assumption was correct:

> What does the client send as position when a range is selected but cursor is not on the range? Just a Position, I presume?

> The current implementation just sends the position, yes.

So when doing the mentioned steps, `lsp_hover` should ignore the range and only send the position. And we don't need to care about updating popup locations.

plugin/hover.py
```diff
@@ -106,22 +106,41 @@ def run_async() -> None:
...
 
         sublime.set_timeout_async(run_async)
 
-    def request_symbol_hover_async(self, listener: AbstractViewListener, point: int) -> None:
+    def request_symbol_hover_async(self, listener: AbstractViewListener, point_or_region: PointOrRegion) -> None:
         hover_promises = []  # type: List[Promise[ResolvedHover]]
-        document_position = text_document_position_params(self.view, point)
+        point = point_or_region if isinstance(point_or_region, int) else point_or_region.begin()
         for session in listener.sessions_async('hoverProvider'):
+            range_hover_provider = session.get_capability('experimental.rangeHoverProvider')
```
rwols, line 131: Sorry, but I still think this is way too complicated for its own good?

I was imagining a server capability path `hoverProvider.rangeSupport`, which is a boolean, which is `true` when the server has support for a range in the request.

Client side request:

- If the server has `hoverProvider.rangeSupport`, and
- If the position is contained in the current selection, and
- If the current selection is not empty,
- then add `params["range"] = region_to_range(first_selection_region(view), view)`

(I'm unsure if all these additional checks other than the `hoverProvider.rangeSupport` are needed. I could imagine other servers wanting to have the information about the current position in any case, whether it's contained in the range or not)

Server side:

- Check whether the request contains a `"range"` and do something clever with that

Client side response:

- Show whatever the server returns

This whole scheme requires no more than 8 lines of diff, no? We're in a position to cook up an additional server capability and enhancement to the `textDocument/hover` request and I think the above scheme is the least intrusive for the spec.

rwols (approved):

rwols (changes requested):
I accidentally approved it :) see above comment.

plugin/hover.py
rwols: By the way, whether the server capability is considered experimental or not (`experimental.rangeHoverProvider`) is an implementation detail for LSP-metals I think.

ayoub-benali:
```
Client side request:

If the server has hoverProvider.rangeSupport, and
If the position is contained in the current selection, and
If the current selection is not empty,
then add params["range"] = region_to_range(first_selection_region(view), view)
(I'm unsure if all these additional checks other than the hoverProvider.rangeSupport are needed. I could imagine other servers wanting to have the information about the current position in any case, whether it's contained in the range or not)
```
Two scenarios are missing:
 - `lsp_hover` is invoked via keyboard and the mouse isn't hovering. `textDocument/hover` should be sent with selection
 -  there is a selection but mouse is hovering an other location. `textDocument/hover` should be sent with mouse position
 
 This is why there are four cases

plugin/hover.py
```diff
@@ -209,6 +228,7 @@ def _show_hover(self, listener: AbstractViewListener, point: int, only_diagnosti
         _test_contents.append(contents)  # for testing only
 
         if contents:
+            # The previous popup could be in a different location from the next one
```
rchl, line 231: remove

plugin/core/views.py
```diff
@@ -404,7 +406,6 @@ def text_document_code_action_params(
 # Workaround for a limited margin-collapsing capabilities of the minihtml.
 LSP_POPUP_SPACER_HTML = '<div class="lsp_popup--spacer"></div>'
 
-
```
rchl, line 407: random change

rchl (commented):
I personally am pretty OK with those changes.
Just make sure to fix all the type/test issues.

plugin/hover.py
rwols: I mean, maybe I want to see both the expression type of the range as well as type info of the position outside of the range using the mouse? Why must that be excluded?

ayoub-benali: yes and it is possible with the current implementation. We just can't send a request with both fields set: https://scalameta.org/metals/docs/integrations/new-editor#textdocumenthover

>  Why must that be excluded?
it is not excluded but as I said there are multiple combinations and these four checks are there for that. 
Are you saying you can handle all those cases with fewer checks ?

jwortmann: > `lsp_hover` is invoked via keyboard and the mouse isn't hovering. `textDocument/hover` should be sent with selection

This still feels wrong to me and I would argue to not send the selection if it is entirely unrelated to mouse hover. Shouldn't it be a separate command which can be used to assign a key binding then? For example `lsp_selection_info` or so.

jwortmann: Well, actually the `lsp_hover` already takes the optional `point` argument, and as implemented here it is just extending this concept. So feel free to ignore my concern if it's useful to combine them all into `lsp_hover`, even though not really related to "hover".

rchl: > > `lsp_hover` is invoked via keyboard and the mouse isn't hovering. `textDocument/hover` should be sent with selection
> 
> This still feels wrong to me and I would argue to not send the selection if it is entirely unrelated to mouse hover. Shouldn't it be a separate command which can be used to assign a key binding then? For example `lsp_selection_info` or so.

When using the keyboard the "point" has to follow the caret. It wouldn't make sense to follow the mouse pointer when using keyboard.

Yes, the naming of `textDcoument/hover` request makes it confusing to think about in terms of keyboard usage but if we want to allow triggering the hover info from the keyboard (which we are and should) then it only makes sense to follow the caret position. It would be unusable otherwise.

jwortmann: Right, I think I had this confusion because I don't have this keybinding set up, and because `lsp_hover` does different things depending on how it was invoked. Should be fine then as it is!

plugin/hover.py
```diff
@@ -82,17 +83,16 @@ def run(
...
             if region is not None:
-                temp_point = region.begin()
-        if temp_point is None:
+                point_or_region = region
+        if point_or_region is None:
             return
         window = self.view.window()
         if not window:
             return
-        hover_point = temp_point
```
rchl, line 95: Actually changes here seem unnecessary.

We don't need to pass range to `request_symbol_hover_async`. It can figure out what to do itself based on the hover point.

If your intention was to let `lsp_hover` be passed a range then I don't see the point of that due to above.

rchl: Well, almost, if not the fact that we want to get relevant diagnostics.

rchl: But then you are not even using `diagnostics_intersecting_region_async` so it doesn't work as I would expect anyway...

ayoub-benali: I removed the usage of `diagnostics_intersecting_region_async` when it is a region because one test is failing and I couldn't figure out what was happening to even fix it. Maybe the test case was wrong ?

rchl: Test server doesn't support range request so it should be unaffected by these changes if the changes are done correctly. It's probably related to this issue since those changes seem to change the behavior in some cases.

plugin/hover.py
```diff
@@ -120,8 +121,19 @@ def run_async() -> None:
...
-        document_position = text_document_position_params(self.view, point)
         for session in listener.sessions_async('hoverProvider'):
+            range_hover_provider = session.get_capability('experimental.rangeHoverProvider')
+            if range_hover_provider:
+                region = first_selection_region(self.view)
+                if region is not None:
+                    if region.contains(point):  # when hovering selection send the selection as range
+                        document_position = text_document_range_params(self.view, region)
+                    else:  # there is selection but ignored
+                        document_position = text_document_range_params(self.view, region)
```
rchl, line 132: This case should send point, no?

plugin/hover.py
```diff
@@ -122,6 +123,13 @@ def request_symbol_hover_async(self, listener: AbstractViewListener, point: int)
         hover_promises = []  # type: List[Promise[ResolvedHover]]
         document_position = text_document_position_params(self.view, point)
         for session in listener.sessions_async('hoverProvider'):
+            range_hover_provider = session.get_capability('experimental.rangeHoverProvider')
+            if range_hover_provider:
+                region = first_selection_region(self.view)
+                if region is not None:
+                    if region.contains(point):
+                        # hovering selection or only text was selected
+                        document_position = text_document_range_params(self.view, region)
```
rchl, line 132: This is not correct. `document_position` will be set here and then we'll send it to other sessions also on next loop iterations.

ayoub-benali: @rchl you prefer [this](https://github.com/sublimelsp/LSP/pull/1900/commits/7e5e0450e2aed35ca8d71297f58934e281af2ae5) or override the variable ?

plugin/hover.py
ayoub-benali: @rwols I simplified it now and all cases work. Sorry for missing your point last time.

plugin/hover.py
```diff
@@ -120,8 +121,16 @@ def run_async() -> None:
...
         for session in listener.sessions_async('hoverProvider'):
+            document_position = text_document_position_params(self.view, point)
+            range_hover_provider = session.get_capability('experimental.rangeHoverProvider')
+            if range_hover_provider:
+                region = first_selection_region(self.view)
+                if region is not None:
+                    if region.contains(point):
+                        # hovering selection or only text was selected
+                        document_position.pop('position', None)
+                        document_position['range'] = region_to_range(self.view, region).to_lsp()
```
ayoub-benali, line 133: OK mypy isn't happy about these two lines.
```
plugin/hover.py:132: error: Key 'position' of TypedDict "TextDocumentPositionParams" cannot be deleted
plugin/hover.py:133: error: TypedDict "TextDocumentPositionParams" has no key 'range'
```
From my I see this one is a known issue in mypy https://github.com/python/mypy/issues/8887
Should simply disable the checks or any other workaround ?

rchl, line 133: This looks very messy. You could make it simpler and fix types with something like:

```suggestion
            params = self._create_hover_request(session, point)
            hover_promises.append(session.send_request_task(Request("textDocument/hover", params, self.view)))
```

and define

```py
    def _create_hover_request(
        self, session: Session, point: int
    ) -> Union[TextDocumentPositionParams, ExperimentalTextDocumentRangeParams]:
        if session.get_capability('experimental.rangeHoverProvider'):
            region = first_selection_region(self.view)
            if region is not None and region.contains(point):
                return text_document_range_params(self.view, region)
        return text_document_position_params(self.view, point)
```

I would also create proper type for the `text_document_range_params`:

```py
ExperimentalTextDocumentRangeParams = TypedDict('ExperimentalTextDocumentRangeParams', {
    'textDocument': TextDocumentIdentifier,
    'range': RangeLsp,
}, total=True)
```

plugin/hover.py
```diff
@@ -119,18 +122,19 @@ def run_async() -> None:
...
 
+    def _create_hover_request(
+        self, session: Session, point: int
+    ) -> Union[TextDocumentPositionParams, ExperimentalTextDocumentRangeParams]:
+        if session.get_capability('experimental.rangeHoverProvider'):
+            region = first_selection_region(self.view)
+            if region is not None and region.contains(point):
+                return text_document_range_params(self.view, region)
+        return text_document_position_params(self.view, point)
+
```
rchl, line 133: I'm nit picked but it's better if the methods are ordered in the order they are used. It help when reading the code from top to bottom. So would ask you to move this one below `request_symbol_hover_async`

rchl (approved):

predragnikolic (approved):

rchl:
I've just noticed that rust-analyzer supports the same experimental feature but implemented differently and with different capability. https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/lsp-extensions.md#hover-range

It would be ideal if both server's implementation would be unified :)

ayoub-benali:
> It would be ideal if both server's implementation would be unified :)

TBH I have no involvement with the server development so it is gonna be very difficult getting the two servers to agree on the capability and implementation without an official spec :/

rwols (changes requested):
Am I the only one nagging too much?

These two comments describe `position OR range`:

https://github.com/microsoft/language-server-protocol/issues/377#issuecomment-953753561
https://github.com/microsoft/language-server-protocol/issues/377#issuecomment-969151283

But both metals and rust-analyzer now want `position XOR range`. It doesn't make sense to me to necessarily omit the position when there is a range. @tgodzik can't metals be modified to take `position OR range`? I don't want to get into a situation where we have to maintain an old version of metals for `position OR range` while a future standardized version of VSCode uses `position XOR range`. Let us figure out what the correct params look like before committing to this kind of code.

ckipp01:
> can't metals be modified to take position OR range

Just to be super clear, you're asking for it to be changed to what is listed in this comment right?

https://github.com/microsoft/language-server-protocol/issues/377#issuecomment-969151283

If so, it's actually pretty easy to change this on our end, we just need to make sure the clients that have implemented this are aware of it. If so, it'd be as simple as just changing the logic in this single method:

https://github.com/scalameta/metals/blob/3f211b03d6c2d1ff4d5fa936d50a4aab013100fb/metals/src/main/scala/scala/meta/internal/metals/HoverExtension.scala#L14-L16

We could just check to see if range is there, if so use it, if not, use the fallback to the position. I don't think this would be an issue unless @tgodzik sees something I'm missing.

ayoub-benali:
> I don't want to get into a situation where we have to maintain an old version of metals for position OR range while a future standardized version of VSCode uses position XOR range

Metals haven't reached 1.0 so breaking changes are expected and even more for features under experimental. So I wouldn't worry about that. 

If Metals' implementation would be to change then I prefer the first option: 

```
interface HoverParams extends WorkDoneProgressParams {
    textDocument: TextDocumentIdentifier;
    position: Position;
    range?: Range;
}
```

tgodzik:
Currently Metals has HoverParams specified as:

```
case class HoverExtParams(
    textDocument: TextDocumentIdentifier,
    @Nullable position: Position = null,
    @Nullable range: Range = null
) {
  def getPosition: Position =
    if (position != null) position
    else range.getStart()
}
```
 
so it can have both `position` and `range`, where range will only be used for specific functionality (type on selection).

https://scalameta.org/metals/docs/integrations/new-editor#textdocumenthover

So I don't think any change is really needed on the server side? If I am following the discussion correctly?

ayoub-benali:
Thanks for the info @tgodzik ! I wasn't aware that both position and range could be set. 
The documentation you sent says otherwise 
```
 /** Either `position` or `range` should be specified */
 ```
 On the Server side there is nothing to change expect fixing the doc.
 
 @rwols is it enough if the implementation is updated to simply include the range when possible to get this PR merged ?

**Comment by tgodzik** [comment]
> The documentation you sent says otherwise 

The comment is wrong I think, the spec underneath it doesn't  say that only one could be sent. And from the implementation it seem both are fine.

**Comment by ckipp01** [comment]
I think there are multiple different conversations going on that are sort of mixing. If I'm understanding what @rwols was pointing out, the conversations are referring to `position` _always_ being sent with an optional `range`. We could still do this, but then the logic we have would still need to change since we default to position, which would break in the examples above.

```scala
def getPosition: Position =
    if (position != null) position
    else range.getStart()
```

>  the spec underneath it doesn't say that only one could be sent.

The spec doesn't say it, but we sort of assume it since if the position is there we take it no matter what.

tgodzik:
> Not really since if a user sends both, they'll never get the range.

Though in theory this should not be an issue, no? `getPosition` is used for features that do not use range, so they would not have a use for that anyway. Range is still used for expression type on selection feature

ckipp01:
> Though in theory this should not be an issue, no? getPosition is used for features that do not use range, so they would not have a use for that anyway. Range is still used for expression type on selection feature

Ahhh you're totally right. I was mistaken on how that method was actually used.

ayoub-benali:
@rwols I modified the request to send position and range when both are available.

rwols:
Sorry for the lag. I’ll get to this in the weekend!

rwols (approved):

PR #1901, opened 2021-11-20
Fix LSP status field missing randomly

rchl:
The status fields for indicating that the server is running would sometimes not be shown. This has happened randomly when sessions were initializing.

The issue was that the [view settings listener](https://github.com/sublimelsp/LSP/blob/a8e67857126b2c9914d556850ebfc32526f650c6/plugin/documents.py#L161-L161) set up in `DocumentViewListener` would always trigger initially since we didn't initialize `self._current_syntax` from `__init__` so the [check if something has changed](https://github.com/sublimelsp/LSP/blob/a8e67857126b2c9914d556850ebfc32526f650c6/plugin/documents.py#L786-L788) would evaluate to `True` and call `self._reset()`.

At the point the `self._reset()` is called the `SessionViews` might already have been created (and in fact almost always are since the settings change is triggered by us adding new keys on starting the session) and the reset call will result in [those being destroyed](https://github.com/sublimelsp/LSP/blob/a8e67857126b2c9914d556850ebfc32526f650c6/plugin/documents.py#L773-L780) and new `SessionViews` created.

Now, this doesn't fix the root issue. I think the root issue is the fact that the `SessionView.__del__` of those cleared instances can get called after the new `SessionView`s are created thus resulting in the old one erasing the view status set by the new one.

As nice as relying on `__del__` might seem, I think it would be better to have explicit public methods for clearing the state and not relying on `__del__` in general since it's not really guaranteed when those will run.

The fix ensures that we don't unnecessarily reset the sessions and trigger this issue on start.

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -225,7 +225,7 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
     def on_session_shutdown_async(self, session: Session) -> None:
         removed_session = self._session_views.pop(session.config.name, None)
         if removed_session:
-            removed_session.on_before_destroy()
+            removed_session.on_before_remove()
             if not self._session_views:
                 self.view.settings().erase("lsp_active")
                 self._registered = False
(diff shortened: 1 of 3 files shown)


rchl:
Also added a fix for the root issue. Let me know if this should be included or separate.

@jwortmann this will potentially conflict with your task unfortunately.

rchl:
Another advantage of being explicit is that we can enforce the cleanup to run on async thread.

rwols (approved):
This adds more maintenance by asking everyone to remember to call cleanup code when refactoring stuff. On the other hand it's hard to ensure these destructors run from the right thread.

PR #1903, opened 2021-11-23
fix types for protocols to add type safety

rchl:
Fixes types for pyright in cases like below where we've blatantly lied about types.

Screenshot 2021-11-23 at 10 41 10

The way it was done before, apart from causing type issues with Pyright, allowed users to just access `session_view.session_buffer` from `Session`, for example, and get away with accessing anything on `SessionBuffer` because the type was `Any`.

@jwortmann sorry to cause troubles for you again. I don't mean bad. :)

### DIFF
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -6,13 +6,13 @@ from .core.protocol import DiagnosticTag
 from .core.protocol import DocumentUri
 from .core.protocol import Notification
 from .core.protocol import Request
+from .core.sessions import AbstractViewListener
 from .core.sessions import Session
 from .core.settings import userprefs
 from .core.types import debounced
 from .core.typing import Any, Iterable, List, Tuple, Optional, Dict, Generator
 from .core.views import DIAGNOSTIC_SEVERITY
 from .core.views import text_document_identifier
-from .core.windows import AbstractViewListener
 from .session_buffer import SessionBuffer
 from weakref import ref
 from weakref import WeakValueDictionary
@@ -88,7 +88,7 @@ class SessionView:
         return self._view
 
     @property
-    def listener(self) -> ref[AbstractViewListener]:
+    def listener(self) -> 'ref[AbstractViewListener]':
         return self._listener
 
     @property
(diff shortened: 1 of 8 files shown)


plugin/session_view.py
```diff
@@ -37,47 +37,63 @@ class SessionView:
     _session_buffers = WeakValueDictionary()  # type: WeakValueDictionary[Tuple[int, int], SessionBuffer]
 
     def __init__(self, listener: AbstractViewListener, session: Session, uri: DocumentUri) -> None:
-        self.view = listener.view
```
rchl, line 40: To avoid changing too much code, decided to use the "underscore" property in `__init__` and access through the getter elsewhere. This is maybe not super consistent but makes the diff smaller.

rchl:
Hmm, looks like I've triggered some import loop...

rwols:
The `= None` syntax is the correct syntax for python < 3.6. Because plugin_host uses python 3.3, we cannot annotate properties of classes so this syntax is used. Mypy understands it.

https://mypy.readthedocs.io/en/stable/class_basics.html#instance-and-class-attributes

> Type comments work as well, if you need to support Python versions earlier than 3.6:
> ```
> class A:
>     x = None  # type: List[int]  # Declare attribute 'x' of type List[int]
> ```

Do we really need accessor methods? It's starting to look a bit enterprisey...

rwols:
I do appreciate that `listener` and `session_buffer` don't have type `Any` anymore, but the accessor methods are a bit too much I think.

rchl:
> Do we really need accessor methods? It's starting to look a bit enterprisey...

I don't know of another way to do it.

Changing:

```py
session_buffer = None  # type: Any
```

to

```py
session_buffer = None  # type: SessionViewProtocol
```

confuses `mypy` in places where `SessionView` is passed when `SessionViewProtocol` is expected so that can't be done either.

rchl:
I would think that type safety is more important over how it "feels"? And to me it doesn't even feel bad from the *Protocol perspective.

I guess it is a bit weird how it feels in the actual "implementations" of those protocols though but it can be seen as a positive thing since the distinction between the private properties and the public interface is clear that way.

rwols (approved):

Issue #1880, opened 2021-10-23
Disable LSP-intelephense globally doesn't work, in projects it does


braver:
**Describe the bug**
When I use the command to disable the LSP server I'm using (Intelephense) globally, it makes the required entry in the settings file, but Intelephense still runs and will continue to run and initialise for any new window I open. Even after restarting ST, etc. It appears this setting does not work at all. There is a note in the example settings that other settings are required as well (command, selector), but adding those makes no difference. The debug log has no error messages, it does clearly log the initialisations that still happen.

If I use "disable language server in project" it creates the same settings, now in the project file. This immediately stops the server for that project and effectively stops it from running in future instances, e.g. after restarting ST. The debug logs indicate that the project override is applied. This seems to work exactly as it says on the tin.

**To Reproduce**
1. Open a project with some PHP in it, and have LSP-intelephense set up.
1. Note that it runs and appears in the status bar.
2. Run the command "disable language server globally".
3. Note that it is however still running and displayed on the status bar.
4. Run the command "disable language server in project".
4. Note that it stops running and doesn't come back.

**Expected behavior**
LSP-intelephense should stop running when disabled globally.

**Environment (please complete the following information):**
- OS: macOS Big Sur 11.6
- Sublime Text version: 4113
- LSP version: 1.12.1
- Language servers used: LSP-intelephense

rwols:
Was this fixed by #1907?

predragnikolic:
Yes.

PR #1907, opened 2021-11-28
Fix "disable server globally" command for LSP packages

predragnikolic:
This *should* fix `LSP: Enable / Disable Language Server Globally` for LSP-* plugins.

Previously we suggeted users to disable `LSP-*` with `Package Control: Disable package`. 
The downside of that is that we clean up the `Package Storage` folder. 
And when we run `Package Control: Enable package`, we have to download the servers bins from the internet again.

Which is ok, unless the user do not have access to the internet at the moment he ran `Package Control: Enable package`, then we wont be able to download what is necessary.

----

But if we use `LSP: Enable / Disable Language Server Globally`, 
we will just switch the `enabled` bool  setting, 
which wont delete anything inside  the `Package Storage` folder. 

----

The only thing is that my fix doesn't work, because of this import error:

```
  File "/Applications/Sublime Text.app/Contents/MacOS/Lib/python33/sublime_plugin.py", line 308, in reload_plugin
    m = importlib.import_module(modulename)
  File "./python3.3/importlib/__init__.py", line 90, in import_module
  File "<frozen importlib._bootstrap>", line 1584, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1565, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1532, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 584, in _check_name_wrapper
  File "<frozen importlib._bootstrap>", line 1022, in load_module
  File "<frozen importlib._bootstrap>", line 1003, in load_module
  File "<frozen importlib._bootstrap>", line 560, in module_for_loader_wrapper
  File "<frozen importlib._bootstrap>", line 868, in _load_module
  File "<frozen importlib._bootstrap>", line 313, in _call_with_frames_removed
  File "/Users/codetribe/Library/Application Support/Sublime Text/Packages/LSP-json/plugin.py", line 2, in <module>
    from LSP.plugin import DottedDict
  File "/Users/codetribe/Library/Application Support/Sublime Text/Packages/LSP/plugin/__init__.py", line 12, in <module>
    from .core.sessions import AbstractPlugin
  File "/Users/codetribe/Library/Application Support/Sublime Text/Packages/LSP/plugin/core/sessions.py", line 2, in <module>
    from .diagnostics_manager import DiagnosticsManager
  File "/Users/codetribe/Library/Application Support/Sublime Text/Packages/LSP/plugin/core/diagnostics_manager.py", line 4, in <module>
    from .views import diagnostic_severity
  File "/Users/codetribe/Library/Application Support/Sublime Text/Packages/LSP/plugin/core/views.py", line 20, in <module>
    from .settings import userprefs
  File "/Users/codetribe/Library/Application Support/Sublime Text/Packages/LSP/plugin/core/settings.py", line 3, in <module>
    from .sessions import get_plugin
ImportError: cannot import name get_plugin
```

related issue -> https://github.com/sublimelsp/LSP/issues/1880


plugin/core/settings.py
```diff
@@ -85,6 +86,14 @@ def update_configs(self) -> None:
     def _set_enabled(self, config_name: str, is_enabled: bool) -> None:
         settings = sublime.load_settings("LSP.sublime-settings")
         clients = settings.get("clients")
+
+        plugin = get_plugin(config_name)
+        if plugin:
+            plugin_settings, plugin_settings_name = plugin.configuration()
+            plugin_settings.set("enabled", is_enabled)
+            sublime.save_settings(plugin_settings_name)
+            return
```
rchl, line 95: This can be the first thing in this function as it doesn't use variables above.

Also, we tend to avoid unnecessary blank lines.

rchl (commented):
> The downside of that is that we clean up the `Package Storage` folder.

While I agree with the fix, this should no longer be true. The storage should only be removed on uninstalling the package now.

> The only thing is that my fix doesn't work, because of this import error:

Looks like a recursive import loop.
Not that trivial to fix. Might require some refactoring.

predragnikolic:
I fixed the import loop,
I expect only `"enabled": false` to be inserted in `LSP-volar.sublime-settings`

But currently all settings keys are inserted. Which is not what I want.

https://user-images.githubusercontent.com/22029477/143776290-f8cde7c5-71cd-4dcf-b030-b39067a53e69.mp4

predragnikolic:
When I call `sublime.save_settings(settings_basename)`,

```py
            plugin_settings, plugin_settings_name = plugin.configuration()
            settings_basename = os.path.basename(plugin_settings_name)
            plugin_settings.set("enabled", is_enabled)
            sublime.save_settings(settings_basename)
```

ST will save create a key in the setting file for each set field. (example `plugin_settings.set("enabled", is_enabled)`)

It turns out that 
lsp_utils/generic_client_handler.py set some settings, 
but never really saves the loaded_settings with `sublime.save_settings(loaded_settings_basename)` file so we never see it.

https://github.com/sublimelsp/lsp_utils/blob/master/st3/lsp_utils/generic_client_handler.py#L124

rchl:
Your need to manually load the user overrides file (and create if missing).

predragnikolic:
I took a different approach.

I overwrote the:
/Users/codetribe/Library/Application Support/Sublime Text/Packages/lsp_utils/st3/lsp_utils/_client_handler/abstract_plugin.py

```diff
class ClientHandler(AbstractPlugin, ClientHandlerInterface):
    """
    The base class for creating an LSP plugin.
    """

    # --- AbstractPlugin handlers -------------------------------------------------------------------------------------

    @classmethod
    def name(cls) -> str:
        return cls.get_displayed_name()

    @classmethod
    def configuration(cls) -> Tuple[sublime.Settings, str]:
+        name = cls.name()
+        basename = "{}.sublime-settings".format(name)
+        filepath = "Packages/{}/{}".format(name, basename)
+        return sublime.load_settings(basename), filepath
-     @classmethod
-     def configuration(cls) -> Tuple[sublime.Settings, str]:
-         return cls.read_settings()
```

and that fixed the issue.

https://user-images.githubusercontent.com/22029477/143777387-4402cd33-6ea0-4543-ba0b-1e24912c0de4.mp4

predragnikolic:
> Your need to manually load the user overrides file (and create if missing).

Can you tell me how to read the settings and override it?
 
I tried with:
```
settings = sublime.load_settings(basename)`
settings.set('enabled', is_enabled)
sublime.save_settings(basename)
```

but that approach doesn't work as it will give the same result as in the [video](https://github.com/sublimelsp/LSP/pull/1907#issuecomment-981110752) above.

So I am not sure on to manually load the user overrides file.

rchl:
> Can you tell me how to read the settings and override it?

Manually read the file like in https://github.com/sublimelsp/LSP/blob/c608b434bd04372cb31ae9a71893ff5901a2700d/plugin/core/types.py#L668-L668 for example.

But since the issue is more or less caused by lsp_utils, maybe it should be fixed there (though can't immediately tell how).

rchl:
Searching github for all uses of `on_client_configuration_ready` seems to indicate that nothing else is using that API so it would be possible to just change it in `lsp_utils` without breaking anything.

predragnikolic:
Because you probably know lsp_utils more than I do,

I still will try to give my assumptions to this: (but keep in mind that I might be very wrong).

<details>

To me the main cause to me seems to be `read_settings`,
because it mutates the settings object here, without saving:

```py
    @classmethod
    def read_settings(cls) -> Tuple[sublime.Settings, str]:
        ...
        for key in cls.get_default_settings_schema().keys():
            loaded_settings.set(key, settings_copy[key]) # mutates the settings object here, without saving
        filepath = "Packages/{}/{}".format(cls.package_name, filename)
        return (loaded_settings, filepath)
```

`read_settings` is used in 4 places:

```
Searching 26 files for "read_settings"

/Users/codetribe/Library/Application Support/Sublime Text/Packages/lsp_utils/st3/lsp_utils/generic_client_handler.py:
  106  
  107      @classmethod
  108:     def read_settings(cls) -> Tuple[sublime.Settings, str]:
  109          filename = "{}.sublime-settings".format(cls.package_name)
  110          loaded_settings = sublime.load_settings(filename)

/Users/codetribe/Library/Application Support/Sublime Text/Packages/lsp_utils/st3/lsp_utils/_client_handler/abstract_plugin.py:
   96      @classmethod
   97      def configuration(cls) -> Tuple[sublime.Settings, str]:
   98:         return cls.read_settings()
   99  
  100      @classmethod

/Users/codetribe/Library/Application Support/Sublime Text/Packages/lsp_utils/st3/lsp_utils/_client_handler/interface.py:
   76  
   77      @classmethod
   78:     def read_settings(cls) -> Tuple[sublime.Settings, str]:
   79          ...
   80  

/Users/codetribe/Library/Application Support/Sublime Text/Packages/lsp_utils/st3/lsp_utils/_client_handler/language_handler.py:
   79      @property
   80      def config(self) -> ClientConfig:
   81:         settings, filepath = self.read_settings()
   82          settings_dict = {}
   83          for key, default in self.get_default_settings_schema().items():

4 matches across 4 files
```

While I think that we can fix it in `abstract_plugin.py` with:
```diff
class ClientHandler(AbstractPlugin, ClientHandlerInterface):
    @classmethod
    def configuration(cls) -> Tuple[sublime.Settings, str]:
+        name = cls.name()
+        basename = "{}.sublime-settings".format(name)
+        filepath = "Packages/{}/{}".format(name, basename)
+        return sublime.load_settings(basename), filepath
-     @classmethod
-     def configuration(cls) -> Tuple[sublime.Settings, str]:
-         return cls.read_settings()
```

I am not sure how to fix it for `language_handler.py`:
```
   79      @property
   80      def config(self) -> ClientConfig:
   81:         settings, filepath = self.read_settings()
   82          settings_dict = {}
   83          for key, default in self.get_default_settings_schema().items():
```

As I think it is used for ST 3 only. Which would mean that this wont be fixed for ST3 users.

</details>

rchl:
> To me the main cause to me seems to be `read_settings`,
> because it mutates the settings object here, without saving:

As far as I can see, the problem is not so much with it not saving the settings (since that would trigger the same problem that you have, just earlier), but the fact that it mutates the settings. We can't just skip that part, as you did with your change, as we need to extend `env` in [one place](https://github.com/sublimelsp/lsp_utils/blob/2ec5b92f07e48c74b92a9395436248cfc1f9cdcc/st3/lsp_utils/npm_client_handler.py#L105-L108).

I thin that a proper solution would require some new API to be able to modify the config without actually modifying the Settings object. So basically modifying the ClientConfig rather than Settings. We'd need to make sure then that the changes are also applied when original settings are reloaded.

While at it, maybe should also think about how to handle 

I guess I will try to come up with some solution for that.

predragnikolic:
The [PR](https://github.com/sublimelsp/lsp_utils/pull/85) probably fixed on the lsp_utils side,
but it seems that there is a `settings.set('languages', value)` call left somewhere. (maybe on the lsp side, maybe on the lsp_utils... could not locate it).

If I have a LSP-json.sublime-settings file:
```
{
	"enabled": true,
}
```
when I run the disable language server command and select LSP-json.

I expect the  LSP-json.sublime-settings to be:
```
{
	"enabled": true,
}
```

but this gets "languages" key gets inserted as well. (and note that the `languageId` is null)

```
{
	"enabled": false,
	"languages":
	[
		{
			"document_selector": "source.json",
			"languageId": null
		}
	],
}
```

plugin/core/settings.py
```diff
@@ -1,3 +1,4 @@
+import os
```
rchl, line 1: order :)

predragnikolic:
I believe that PC still has the issue where it doesn't update dependencies regularly.

Even when the fix is merged to lsp_utils,
should we merge this in the LSP main immediately? 

Or wait a month before we merge this, just to make sure that users do not run into this [issue](https://github.com/sublimelsp/LSP/pull/1907#issuecomment-981110752)

rchl:
> Even when the fix is merged to lsp_utils,
> should we merge this in the LSP main immediately?

I think that the time makes no difference here.

I haven't verified but I believe the dependencies are only updated when the package that uses given dependency gets updated.

rchl (approved):

rchl:
Looks good to me.
There was a failing test but that was just a fluke.

PR #1908, opened 2021-11-28
POC: Fix LSP-pyright code highlighting

deathaxe:
This commit ...

1. adds a Popups-Python.sublime-syntax and
2. assigns it to mdpopups using `language_map` in frontmatter.

The goal is to fix the highlighting issue of the leading `(type)` parentheses in code fences returned by LSP-pyright.

This is a proof of concept.

Maybe a support plugin such as LSP-pyright should be able to register a syntax definition.

Without this commit

grafik

With this commit

grafik

### DIFF
--- a/Syntaxes/Popups-Python.sublime-syntax
+++ b/Syntaxes/Popups-Python.sublime-syntax
@@ -1,0 +1,29 @@
+%YAML 1.2
+---
+name: Python (for LSP)
+scope: source.python.lsp
+hidden: true
+
+extends: Packages/Python/Python.sublime-syntax
+
+contexts:
+  main:
+    - meta_include_prototype: false
+    - match: ''
+      push: [statements, lsp-type]
+
+  lsp-type:
+    - meta_include_prototype: false
+    - match: \(
+      scope: punctuation.section.parens.begin.python
+      set: lsp-type-body
+    - match: (?=\S)
+      pop: true
+
+  lsp-type-body:
+    - meta_scope: meta.parens.python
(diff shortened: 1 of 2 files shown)


rchl:
I like this. Hopefully there is no drawbacks to that also.

Note that I've been experimenting with fixing that on the LSP side (with typescript server rather than pyright) by just altering the server output to remove that prefix (or rather, put it as a comment on the line above). While that helps in many cases, at least with the typescript server it doesn't guarantee that the output will still be a valid syntax since it's not always technically a code but often just a complex type information.

deathaxe:
The illustrated approach of this POC would even enable writing custom syntax definitions and tell mdpopups to use them. We don't neccesarily inherit from an existing syntax.

What we'd need would be some interface so supporting packages such as LSP-pyright or LSP-typescript can provide custom sublime-syntax definitions and the required language map entry to be injected into the frontmatter.

We could of course parse content and convert `(class)` into an `` tag or unicode.

I don't feel like I'll find the time to implement a full fledged solution though. Just want to share an idea for a possible solution.

rwols:
I think this is definitely the right approach, let me try to write some extension point

rchl:
Just for the record gonna illustrate my point. Both syntax highlighting are invalid below.

Screenshot 2021-12-04 at 22 19 40
Screenshot 2021-12-04 at 22 29 23

deathaxe:
And with this POC realized a helper plugin can provide its own syntax for it.

rwols:
superseded by #1914

PR #1909, opened 2021-11-30
Add setting for diagnostics sort order

jwortmann:
This is a suggestion to make the sort order for diagnostics (per file) configurable by the user. It allows for example to restore the former sorting by location, which was removed when the DiagnosticsManager / Goto Diagnostic quick panel was added.

I've left the default value to be unsorted (current behavior).

For some context see https://github.com/sublimelsp/LSP/pull/1884#issuecomment-960159248

---

I've also move the "show_multiline_diagnostics_highlights" setting up a little bit in LSP.sublime-settings, so that all settings related to diagnostics are grouped together.


plugin/core/diagnostics_manager.py
```diff
@@ -9,6 +10,19 @@
...
 
 
+def by_location(diagnostic: Diagnostic) -> Tuple[int, int]:
+    line = diagnostic.get("range", {})["start"]["line"]
+    column = diagnostic.get("range", {})["start"]["character"]
+    return line, column
+
+
+def by_severity(diagnostic: Diagnostic) -> Tuple[int, int, int]:
+    severity = diagnostic.get("severity", -1)
```
jwortmann, line 20: Severity can be omitted, and according to the specs it is up to the client how to interpret the diagnostic in that case.
`-1` here would show such diagnostics at the top (above errors), maybe we rather should move them to the bottom instead?

jwortmann: I changed it to list them at the bottom, which I think is a better order.

rchl:
Even though I've originally raised this need, I don't think anymore that it's better to sort by location after using the current logic for a while.

Nonetheless, this looks like potentially useful and simple feature.

rchl:
@jwortmann are you fine with dropping this PR? Nobody else complained, I personally don't miss this feature either so maybe it's probably not worth complicating the code.

jwortmann:
Well, I still find it weird to use just the arbitrary and sometimes random ordering from the server, rather than sorting the diagnostics. Especially for the "Goto diagnostic" in the command palette it seems confusing to me that the view scrolls back and forth when you repeatedly press the arrow-down key in the list.

Unbenannt

VSCode for example seems to sort the diagnostics by line number in its problem panel. But the matter is not that important to me and I rarely use the diagnostics panel anyway, so I guess we can close this.

> Nobody else complained

That's true, but not really an argument that it is the best solution what is currently used, because before the PR which removed the sorting I think also nobody complained over a time of several years :)

PR #1914, opened 2021-12-05
Allow plugins to specify a custom syntax for code blocks in markdown

rwols:
Real implementation for the POC in #1908

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -38,11 +38,11 @@ class LspResolveDocsCommand(LspTextCommand):
     def _format_documentation(
         self,
         content: Union[str, Dict[str, str]],
-        language_map: MarkdownLangMap
+        language_map: Optional[MarkdownLangMap]
     ) -> str:
         return minihtml(self.view, content, FORMAT_STRING | FORMAT_MARKUP_CONTENT, language_map)
 
-    def _handle_resolve_response_async(self, language_map: MarkdownLangMap, item: CompletionItem) -> None:
+    def _handle_resolve_response_async(self, language_map: Optional[MarkdownLangMap], item: CompletionItem) -> None:
         detail = ""
         documentation = ""
         if item:
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -1,15 +1,17 @@
-import sublime
-import webbrowser
-from .core.logging import debug
 from .core.edit import parse_text_edit
+from .core.logging import debug
 from .core.protocol import Request, InsertTextFormat, Range, CompletionItem
 from .core.registry import LspTextCommand
 from .core.typing import List, Dict, Optional, Generator, Union
-from .core.views import FORMAT_STRING, FORMAT_MARKUP_CONTENT, minihtml
+from .core.views import FORMAT_STRING, FORMAT_MARKUP_CONTENT
+from .core.views import MarkdownLangMap
+from .core.views import minihtml
 from .core.views import range_to_region
 from .core.views import show_lsp_popup
 from .core.views import update_lsp_popup
 import functools
+import sublime
+import webbrowser
 
 SessionName = str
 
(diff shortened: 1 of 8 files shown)


plugin/core/views.py
```diff
@@ -33,6 +33,8 @@
 import sublime_plugin
 import tempfile
 
+MarkdownLangMap = Optional[Dict[str, Tuple[Sequence[str], Sequence[str]]]]
```
rchl, line 36: `Optional` should IMO not be a part of this type itself. It can be confusing to see a type name indicating that it's a map without being clear that it can also be `None`.

plugin/completion.py
```diff
@@ -25,23 +27,31 @@ def run_async() -> None:
...
 
         sublime.set_timeout_async(run_async)
 
-    def _format_documentation(self, content: Union[str, Dict[str, str]]) -> str:
-        return minihtml(self.view, content, allowed_formats=FORMAT_STRING | FORMAT_MARKUP_CONTENT)
+    def _format_documentation(
+        self,
+        content: Union[str, Dict[str, str]],
+        language_map: MarkdownLangMap
+    ) -> str:
```
rchl, line 42: Still fits one line.

plugin/documents.py
```diff
@@ -465,8 +467,13 @@ def _get_signature_help_session(self) -> Optional[Session]:
...
         return self.session_async("signatureHelpProvider", pos)
 
-    def _on_signature_help(self, response: Optional[SignatureHelp], point: int) -> None:
-        self._sighelp = SigHelp.from_lsp(response)
+    def _on_signature_help(
+        self,
+        response: Optional[SignatureHelp],
+        point: int,
+        language_map: MarkdownLangMap
+    ) -> None:
```
rchl, line 475: Can still be one line.

rchl (commented):
Will have a proper look later.

rchl (commented):
Tried with @deathaxe's POC syntax and highlighting can still be broken.

Screenshot 2021-12-05 at 19 10 56

So the custom syntax would need some more work. Though not sure if pyright/ts-ls follows any specific convention so this might be a bit tricky.

plugin/core/sessions.py
```diff
@@ -599,6 +600,17 @@ def __init__(self, weaksession: 'weakref.ref[Session]') -> None:
         """
         self.weaksession = weaksession
 
+    def markdown_language_id_to_st_syntax_map(self) -> MarkdownLangMap:
+        """
+        Override this method to tweak the syntax highlighting of code blocks in popups from your language server.
+        The returned object should be a dictionary exactly in the form of mdpopup's language_map setting.
+
+        See: https://facelessuser.github.io/sublime-markdown-popups/settings/#mdpopupssublime_user_lang_map
```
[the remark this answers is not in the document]
rchl: Ah, the documentation talks in context of the JSON file so it's correct with that in mind.

So we should give an example since our use is within the Python context.

rchl:
Will check properly tomorrow but the type of MarkdownLangMap should still be changed to use tuples.

plugin/core/views.py
```diff
@@ -33,6 +33,8 @@
 import sublime_plugin
 import tempfile
 
+MarkdownLangMap = Dict[str, Tuple[Sequence[str], Sequence[str]]]
```
deathaxe, line 36:
```suggestion
MarkdownLangMap = Dict[str, Tuple[Tuple[str], Tuple[str]]]
```

@rchl  is right about tuples. They are crucial to make the langmap to be accepted by mdpopups.

deathaxe:
> Tried with @deathaxe's POC syntax and highlighting can still be broken.

I guess this will require some improvements with regards to type hint highlighting in python and/or a dedicated pyright syntax, especially for the `Self@Something` syntax.

PR #1915, opened 2021-12-06
Handle window/showMessageRequest with quick panel

ayoub-benali:
https://github.com/sublimelsp/LSP/pull/974 was introduced as a workaround because ST had no way of presenting a message to the user. Now that it is possible to set a place holder. The popup logic isn't needed.

You can see the workflow improvement bellow:
Before
Peek 2021-12-05 22-48

After
Peek 2021-12-06 11-09


tests/test_message_request_handler.py
```diff
@@ -17,6 +17,6 @@ def test_show_popup(self):
                 {'title': "def"}
             ]
         }
-        handler = MessageRequestHandler(view, session, "1", params, 'lsp server')
-        handler.show()
-        self.assertTrue(view.is_popup_visible())
+        manager = WindowManager(window, None, None)
+        manager.handle_message_request(session, params, "1")
+        self.assertEquals(view.element(), "input:input")
```
ayoub-benali, line 22: `view.element()` is always returning None. I am not sure what is wrong here

jwortmann: Looks like `Window.active_view()` always returns the active editor view (= tab), even if the input focus is in a widget / input field. Not sure whether it is intentional or a bug.

plugin/core/windows.py
```diff
@@ -412,9 +412,29 @@ def _create_logger(self, config_name: str) -> Logger:
...
+            2: '⚠️',
+            3: 'ℹ️',
+            4: '📝'
+        }
+        actions = params.get("actions", [])
+        titles = list(action.get("title") for action in actions)
+
+        def send_user_choice(index: int) -> None:
+            # when noop; nothing was selected e.g. the user pressed escape
+            result = {}
```
jwortmann, line 426: Shouldn't this be `None`, like in the former implementation? Specs say
> result: the selected `MessageActionItem` | `null` if none got selected.

ayoub-benali: `None` was't serialized as `null`, this client sending `None` which was causing errors on the server.
`{}` translated to `null` with Metals at least.
Maybe Response should be updated then here: https://github.com/sublimelsp/LSP/blob/main/plugin/core/protocol.py#L393

tests/test_message_request_handler.py
rchl: You probably need to remove all references to the removed notifications.css since it's currently breaking the popups.

plugin/core/windows.py
rchl: > this client sending None which was causing errors on the server.

That's not possible since `None` doesn't exist in JSON. Can you show communication log?

tests/test_message_request_handler.py
rchl: No idea if related to this thread actually but the popups are broken. :)

ayoub-benali: > You probably need to remove all references to the removed notifications.css since it's currently breaking the popups.

I just did but same problem

plugin/core/windows.py
ayoub-benali: This is the LSP log after dismissing the quick panel

```
:: <-- metals window/showMessageRequest(2): {'message': 'Select the kind of file to create', 'actions': [{'title': 'class'}, {'title': 'case-class'}, {'title': 'object'}, {'title': 'trait'}, {'title': 'package-object'}, {'title': 'worksheet'}, {'title': 'ammonite'}, {'title': 'enum'}]}
:: >>> metals 2: None
:: <~~ metals 3: {'data': 'java.util.concurrent.CompletionException: java.lang.NullPointerException\n\tat java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:292)\n\tat java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:308)\n\tat java.util.concurrent.CompletableFuture.uniAccept(CompletableFuture.java:661)\n\tat java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:646)\n\tat java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:488)\n\tat java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:1990)\n\tat scala.concurrent.java8.FuturesConvertersImpl$CF.apply(FutureConvertersImpl.scala:29)\n\tat scala.concurrent.java8.FuturesConvertersImpl$CF.apply(FutureConvertersImpl.scala:26)\n\tat scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)\n\tat scala.concurrent.BatchingExecutor$Batch.processBatch$1(BatchingExecutor.scala:67)\n\tat scala.concurrent.BatchingExecutor$Batch.$anonfun$run$1(BatchingExecutor.scala:82)\n\tat scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)\n\tat scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:85)\n\tat scala.concurrent.BatchingExecutor$Batch.run(BatchingExecutor.scala:59)\n\tat scala.concurrent.Future$InternalCallbackExecutor$.unbatchedExecute(Future.scala:875)\n\tat scala.concurrent.BatchingExecutor.execute(BatchingExecutor.scala:110)\n\tat scala.concurrent.BatchingExecutor.execute$(BatchingExecutor.scala:107)\n\tat scala.concurrent.Future$InternalCallbackExecutor$.execute(Future.scala:873)\n\tat scala.concurrent.impl.CallbackRunnable.executeWithValue(Promise.scala:72)\n\tat scala.concurrent.impl.Promise$DefaultPromise.$anonfun$tryComplete$1(Promise.scala:288)\n\tat scala.concurrent.impl.Promise$DefaultPromise.$anonfun$tryComplete$1$adapted(Promise.scala:288)\n\tat scala.concurrent.impl.Promise$DefaultPromise.tryComplete(Promise.scala:288)\n\tat scala.concurrent.Promise.complete(Promise.scala:53)\n\tat scala.concurrent.Promise.complete$(Promise.scala:52)\n\tat scala.concurrent.impl.Promise$DefaultPromise.complete(Promise.scala:187)\n\tat scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:33)\n\tat scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)\n\tat java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)\n\tat java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)\n\tat java.lang.Thread.run(Thread.java:748)\nCaused by: java.lang.NullPointerException\n\tat scala.meta.internal.metals.ConfiguredLanguageClient.$anonfun$metalsQuickPick$1(ConfiguredLanguageClient.scala:118)\n\tat scala.util.Success.$anonfun$map$1(Try.scala:255)\n\tat scala.util.Success.map(Try.scala:213)\n\tat scala.concurrent.Future.$anonfun$map$1(Future.scala:292)\n\tat scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:33)\n\t... 5 more\n', 'code': -32603, 'message': 'Internal error.'}

rchl: Looks to me like `null` is actually sent and the server is crashing on it.

ayoub-benali: You are right, Metals lsp trace shows that `null` was received. I was fooled by the logged `None` in the LSP console

tests/test_message_request_handler.py
ayoub-benali: > No idea if related to this thread actually but the popups are broken. :)

After https://github.com/sublimelsp/LSP/pull/1915/commits/405cd7b4899f83abde93772384d18100d9116b68 they should be unbroken now

rchl:
Not ideal that the message might not be fully visible

Screenshot 2021-12-06 at 17 09 28

ayoub-benali:
> Not ideal that the message might not be fully visible

If the message is longer than a certain size we could log it or switch the popup logic ?

rchl:
> > Not ideal that the message might not be fully visible
> 
> If the message is longer than a certain size we could log it or switch the popup logic ?

There is no way to reliably check if the message fits the input.

The panels can include content at the bottom of the list which would fit more but not sure if it would be a good fit.

ayoub-benali:
Would logging the message each time be enough since most the time messages would fit ?
Otherwise I can put the two implementation behind a flag like `show_references_in_quick_panel`

rchl:
> Would logging the message each time be enough since most the time messages would fit ? Otherwise I can put the two implementation behind a flag like `show_references_in_quick_panel`

Logging is not really discoverable by users.

I like the presentation and the fact that it's keyboard accessible but the fact that the text might not fit and that it takes focus away from typing is not ideal.

But ST just doesn't have a proper solution for that...

ayoub-benali:
>  it takes focus away from typing is not ideal.

The popup also takes focus and can be dismissed accidentally :/

What about this approach ?

Peek 2021-12-06 20-40

rchl:
> The popup also takes focus and can be dismissed accidentally :/

Dismissed accidentally yes, but doesn't take focus away. If I'm typing something, the text will not go to some different place if it suddenly pops up.

rchl:
What I meant earlier is that we could utilize this space at the bottom of new quick panels:

Screenshot 2021-12-06 at 20 46 43

From the implementation side it might be a bit weird though as I believe that message is per item so we would need to assign it to all the items.

ayoub-benali:
I think this approach can't work because sublime doesn't show CommandInputHandler from API calls. It just tested gain the command is called but the input field isn't shown.

rchl:
> I think this approach can't work because sublime doesn't show CommandInputHandler from API calls. It just tested gain the command is called but the input field isn't shown.

I'm not sure what you mean here but I think it's doable. We'd just need to do some bigger refactoring instead of using the convenient `show_quick_panel` API.
Unless I'm wrong and there is something that would prevent that from working...

One more thing that is not ideal is that the session's name is missing in the new panel. I have 3 servers running on a Vue file and one of those is showing the message below. Without the session name you'd be guessing sometimes where is the message coming from.

Screenshot 2021-12-07 at 09 42 35

jwortmann:
> We'd just need to do some bigger refactoring instead of using the convenient `show_quick_panel` API.

You can use the details pane at the bottom even with the `show_quick_panel` API. The only thing you have to do is to replace the list of string items with a list of `QuickPanelItem`s and assign the message to the `details` field of each QuickPanelItem. And the session name could also be added there as a heading, with the help of limited [minihtml formatting](https://www.sublimetext.com/docs/minihtml.html#inline_formatting).

But I don't know if this is a good solution, because the actual purpose of the details pane is to give information about the selected item, and also it tends to be buggy if the content gets too large.

TheSecEng:
This is really slick looking. Hoping this moves forward

ayoub-benali:
I completely forgot about this PR, sorry. 

Given sublime limitations, I think the only way to get this merged is if we hide the behavior change behind a setting, but that is not up to me.

rwols:
Sorry but I really don't like supporting more settings. I'd say either find a strictly better presentation for all types of users or don't merge this.

ayoub-benali:
I am closing this one then :)

rwols:
I wasn't specifically commenting on the content PR. It's certainly a good attempt. But the mood of the conversation went towards "let's please merge this", which I don't agree with :)

rwols:
More philosophically speaking the `window/showMessageRequest` request is somewhat of a VSCode user flow detail leaking into the protocol in my opinion. An editor like ST is more oriented to not disturbing the user while he/she's working. But a popup dialog kind of breaks that flow. The main author of the protocol also doesn't try to add UI flow lock-in but `window/showMessageRequest` seems to be an exception.

Having said that, `window/showMessageRequest` can really be useful when a language server detects some kind of workspace but wants confirmation to do a thing. The idea of ST is that the user "always knows best" so doesn't need hand-holding I guess.

PR #1916, opened 2021-12-08
Call can_start() and on_pre_start() in the troubleshooting flow

rchl:
Follow to real flow more closely for better error reporting.

### DIFF
--- a/plugin/tooling.py
+++ b/plugin/tooling.py
@@ -500,8 +500,8 @@ class ServerTestRunner(TransportCallbacks):
                 if cannot_start_reason:
                     raise Exception('Plugin.can_start() prevented the start due to: {}'.format(cannot_start_reason))
                 cwd = plugin_class.on_pre_start(window, initiating_view, workspace_folders, config)
-            if not cwd and window.folders():
-                cwd = window.folders()[0]
+            if not cwd and workspace_folders:
+                cwd = workspace_folders[0].path
             transport_config = config.resolve_transport_config(variables)
             self._resolved_command = transport_config.command
             self._transport = create_transport(transport_config, cwd, self)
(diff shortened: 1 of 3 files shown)


rchl:
Am I blind? What the hell is mypy complaining about (and why I don't see that through `LSP-pylsp`...?

plugin/tooling.py
```diff
@@ -290,10 +290,11 @@ class LspTroubleshootServerCommand(sublime_plugin.WindowCommand, TransportCallba
 
     def run(self) -> None:
         window = self.window
-        active_view = window.active_view()
-        if not active_view:
+        view = window.active_view()
+        if not view:
             sublime.message_dialog('Troubleshooting must be run with a file opened')
             return
+        active_view = view
```
predragnikolic, line 297: Hahahaha, this was the issue!? 😂

rchl: mypy is too stupid to figure out that the variable that we just None-checked is definitively not gonna be `None` so we need to assign it to a new variable.

rwols: Actually it _may_ change back to None in the closure, right? The variables are evaluated once the closure is invoked, which is at a later point in time. I'm sure that if you use `functools.partial`, then mypy won't complain. I'm unsure whether mypy is truly correct here. But we did see something similar in the TransportCallbacks in plugin/core/transports.py

rchl: I get tripped by similar issue once in a while but I'm not sure that it's the case here.

I don't think that the value of the variable can change in that specific case when lambda before lambda is evaluated. Pyright is smart enough to realize that but not mypy.

If there would be a code after the lambda is created that changes it then I would understand.

rwols (approved):

rwols:
Please resolve the merge conflicts :)

PR #1917, opened 2021-12-10
Don't call the callback if the transport was closed

rchl:
When saving an LSP package's `plugin.py`, the plugin would unload and the chokidar file watcher would be destroyed but the callback would still be attempted. The weak reference is not gone because it's referenced by the `ProcessTransport` itself.

### DIFF
--- a/plugin/core/transports.py
+++ b/plugin/core/transports.py
@@ -142,6 +142,8 @@ class ProcessTransport(Transport[T]):
                     continue
 
                 def invoke(p: T) -> None:
+                    if self._closed:
+                        return
                     callback_object = self._callback_object()
                     if callback_object:
                         callback_object.on_payload(p)


plugin/core/transports.py
```diff
@@ -142,6 +142,8 @@ def _read_loop(self) -> None:
                     continue
 
                 def invoke(p: T) -> None:
+                    if self._closed:
+                        return
```
rwols, line 146: Is this really correct? What if there's a final error message on stderr just before the transport closes? I'm not convinced this doesn't cause more trouble.

rchl: This is the stdout, not stderr, loop. That said, stderr loop also already has the same check.

rwols (approved):

PR #1918, opened 2021-12-10
Handle Code Lens Refresh Request

ayoub-benali:
https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#codeLens_refresh

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1328,7 +1328,7 @@ class Session(TransportCallbacks):
         if self.uses_plugin():
             for sv in self.session_views_async():
                 sv.start_code_lenses_async()
-            self.send_response(Response(request_id, None))
+        self.send_response(Response(request_id, None))
 
     def m_textDocument_publishDiagnostics(self, params: Any) -> None:
         """handles the textDocument/publishDiagnostics notification"""
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -384,6 +384,8 @@ class SessionViewProtocol(Protocol):
     def get_resolved_code_lenses_for_region(self, region: sublime.Region) -> Generator[CodeLens, None, None]:
         ...
 
+    def start_code_lenses_async(self) -> None:
+        ...
 
 class SessionBufferProtocol(Protocol):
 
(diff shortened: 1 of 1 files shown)


plugin/core/sessions.py
```diff
@@ -1317,6 +1320,12 @@ def m_workspace_applyEdit(self, params: Any, request_id: Any) -> None:
         self._apply_workspace_edit_async(params.get('edit', {})).then(
             lambda _: self.send_response(Response(request_id, {"applied": True})))
 
+    def m_workspace_codeLens_refresh(self, _: Any, request_id: Any) -> None:
+        """handles the workspace/codeLens/refresh request"""
+        for sv in self.session_views_async():
+            sv.start_code_lenses_async()
```
ayoub-benali, line 1326: @rwols is it safe to call `start_code_lenses_async` each time ?

ayoub-benali: Also any idea how to fix `error: "SessionViewProtocol" has no attribute "start_code_lenses_async"` ?

rchl: Should be fine to call it each time. This already happens on editing the document.

Though you should guard the calls with `session.uses_plugin()` like the existing places do:

https://github.com/sublimelsp/LSP/blob/02d0f324142022b247bf4c6a0c3d27eaec7bc364/plugin/documents.py#L547-L547

The type issue can be fixed by declaring `start_code_lenses_async` method on the `SessionViewProtocol` interface.

plugin/core/sessions.py
```diff
@@ -1317,6 +1323,13 @@ def m_workspace_applyEdit(self, params: Any, request_id: Any) -> None:
         self._apply_workspace_edit_async(params.get('edit', {})).then(
             lambda _: self.send_response(Response(request_id, {"applied": True})))
 
+    def m_workspace_codeLens_refresh(self, _: Any, request_id: Any) -> None:
+        """handles the workspace/codeLens/refresh request"""
+        if self.uses_plugin():
+            for sv in self.session_views_async():
+                sv.start_code_lenses_async()
+            self.send_response(Response(request_id, None))
```
rchl, line 1331: We have to send the response regardless if handled.

rchl (approved):

Issue #1922, opened 2021-12-24
Add setting to disable global rename confirmation prompt


jonpalmisc:
**Is your feature request related to a problem? Please describe.**
I've been doing a lot of refactoring lately and have found the "Rename N occurrences across N files" prompt to be very annoying.

**Describe the solution you'd like**
It would be nice to be able to disable this prompt via a setting.

**Describe alternatives you've considered**
I have considered trying to be more accepting of the dialog. It is trying its best.

**Additional context**
N/A

jonpalmisc:
I've actually tried adding this myself before filing this issue and I hit a wall — any guidance is welcome and I'll happily submit a PR to address this.

I uninstalled LSP package via package control then cloned the repo into my packages folder with Git. I made these changes, which I expected to be sufficient:

```diff
diff --git a/LSP.sublime-settings b/LSP.sublime-settings
index 5903f32..869f714 100644
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -10,6 +10,9 @@
   // "settings" section of project files.
   "lsp_format_on_save": false,
 
+  // Confirm rename operations that will affect more than one file.
+  "confirm_global_rename": true,
+
   // A dictionary of code action identifiers that should be triggered on save.
   //
   // Code action identifiers are not officially standardized so refer to specific
diff --git a/plugin/rename.py b/plugin/rename.py
index 39ed5ad..75c4ae8 100644
--- a/plugin/rename.py
+++ b/plugin/rename.py
@@ -122,11 +122,12 @@ class LspSymbolRenameCommand(LspTextCommand):
 
     def on_rename_result(self, response: Any) -> None:
         window = self.view.window()
+        settings = self.view.settings()
         if window:
             if response:
                 changes = parse_workspace_edit(response)
                 file_count = len(changes.keys())
-                if file_count > 1:
+                if file_count > 1 and settings.get('confirm_global_rename'):
                     total_changes = sum(map(len, changes.values()))
                     message = "Replace {} occurrences across {} files?".format(total_changes, file_count)
                     choice = sublime.yes_no_cancel_dialog(message, "Replace", "Dry Run")
```

There's probably something I'm missing here, but this didn't work.

jwortmann:
Your diff doesn't work because the setting value is set in the LSP.sublime-settings file, but later you try to read it from the settings of the current view. It should work like this (untested):

```diff
diff --git a/LSP.sublime-settings b/LSP.sublime-settings
[...]

PR #1923, opened 2021-12-25
Add `confirm_global_rename` option to ease rename confirmation prompt

jonpalmisc:
This PR resolves my recently-created issue: https://github.com/sublimelsp/LSP/issues/1922

This worked in my own testing and I believe I followed all the appropriate conventions. Please let me know if I missed anything.

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -184,6 +184,7 @@ def read_list_setting(settings_obj: sublime.Settings, key: str, default: list) -
 class Settings:
 
     # This is only for mypy
+    confirm_global_rename = None # type: bool
     diagnostics_additional_delay_auto_complete_ms = None  # type: int
     diagnostics_delay_ms = None  # type: int
     diagnostics_gutter_marker = None  # type: str
@@ -197,7 +198,7 @@ class Settings:
     log_max_size = None  # type: int
     log_server = None  # type: List[str]
     lsp_code_actions_on_save = None  # type: Dict[str, bool]
-    lsp_format_on_save = None  # type: bool
+    lsp_format_on_save = None  # type: bool    
     on_save_task_timeout_ms = None  # type: int
     only_show_lsp_completions = None  # type: bool
     popup_max_characters_height = None  # type: int
@@ -223,6 +224,7 @@ class Settings:
             val = s.get(name)
             setattr(self, name, val if isinstance(val, default.__class__) else default)
 
+        r("confirm_global_rename", True)
         r("diagnostics_additional_delay_auto_complete_ms", 0)
         r("diagnostics_delay_ms", 0)
         r("diagnostics_gutter_marker", "dot")
@@ -232,7 +234,7 @@ class Settings:
         r("log_debug", False)
         r("log_max_size", 8 * 1024)
         r("lsp_code_actions_on_save", {})
-        r("lsp_format_on_save", False)
+        r("lsp_format_on_save", False)        
         r("on_save_task_timeout_ms", 2000)
         r("only_show_lsp_completions", False)
         r("popup_max_characters_height", 1000)
--- a/sublime-package.json
+++ b/sublime-package.json
@@ -280,6 +280,11 @@
             "lsp_code_actions_on_save": {
               "$ref": "sublime://settings/LSP#/definitions/lsp_code_actions_on_save",
             },
+            "confirm_global_rename": {
+              "type": "boolean",
+              "default": true,
+              "markdownDescription": "Display a confirmation prompt before performing global rename operations that will modify more than one file."
+            },
             "show_diagnostics_in_view_status": {
               "type": "boolean",
               "default": true,
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -10,6 +10,10 @@
   // "settings" section of project files.
   "lsp_format_on_save": false,
 
+  // Display a confirmation prompt before performing global rename operations
+  // that will modify more than one file.
+  "confirm_global_rename": true,
+
   // A dictionary of code action identifiers that should be triggered on save.
   //
   // Code action identifiers are not officially standardized so refer to specific
--- a/plugin/rename.py
+++ b/plugin/rename.py
@@ -8,6 +8,7 @@ from .core.protocol import Request
 from .core.registry import get_position
 from .core.registry import LspTextCommand
 from .core.registry import windows
+from .core.settings import userprefs
 from .core.types import PANEL_FILE_REGEX, PANEL_LINE_REGEX
 from .core.typing import Any, Optional, Dict, List
 from .core.views import first_selection_region, range_to_region, get_line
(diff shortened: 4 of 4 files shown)


sublime-package.json
```diff
@@ -11,6 +11,11 @@
               "default": false,
               "markdownDescription": "Run the server's formatProvider (if supported) on a document before saving. This option is also supported in syntax-specific settings and/or in the `\"settings\"` section of project files."
             },
+            "confirm_global_rename": {
```
predragnikolic, line 14: This is not in the right place for the `"confirm_global_rename"`.

Currently it is put in JSONSchema definitions,
but you want to put it here https://github.com/sublimelsp/LSP/blob/60d725264e8fafe9111cb1ab8373f9829cc63f79/sublime-package.json#L288

predragnikolic (changes requested):
Looks good, but requires a small change.

sublime-package.json
jonpalmisc: This should be fixed now.

predragnikolic (approved):

rchl:
I don't quite see how a dialog that can be confirmed with a single press of an enter key warrants a new option.

It's imo useful to know how many files are going to get changed, especially with many files open since those have to be found and saved manually.

Maybe you could describe what makes the dialog so annoying?

I'd like to figure out solution that is less or not annoying that can work for everybody rather than handling that with a new option.

For example maybe we could proceed with renaming without showing the dialog but then we'd show how many files were affected through a message in a panel. That could also act as a quick way of switching to each modified file.

jonpalmisc:
I realize I could have provided more context in my original issue. Hopefully I can address some of these concerns.

> Maybe you could describe what makes the dialog so annoying?

Not including auto-complete, renaming is hands-down my most used feature. I easily use it 50+ times per day. In the ~2 months I've been using the LSP plugin, I have _never_ wanted to abort a rename operation. "Staying out of the way" is part of the spirit of Sublime Text, and I find this dialog — which interrupts my work on every rename operation — to not be aligned with this philosophy. 

> I don't quite see how a dialog that can be confirmed with a single press of an enter key warrants a new option.

This is my first contribution to this project, so I'm in unfamiliar territory here. I'll say this:
- The current behavior of the rename operation is unchanged with the addition of this option — I made the default value of the setting `true`, meaning "confirm rename operations".
- I imagine I am not the only one bothered by this behavior. I was actually surprised to find that this option didn't exist at first, as I thought "surely there is a way to disable this popup".

> It's imo useful to know how many files are going to get changed, especially with many files open since those have to be found and saved manually.

I disagree — I am never interested in how many files changed and I always hit "save all" afterwards. The changes will be clearly visible with a quick `git diff`. I imagine there are many users like you who do want to know, and many other users like me who don't want to know — as such, it makes sense for this to be a configurable option, meaning everyone can have their preferred experience.

> I'd like to figure out solution that is less or not annoying that can work for everybody rather than handling that with a new option.

Could you explain what's annoying about handling a new option? As I said before, this change will be transparent to existing plugin users as the default value of the option yields the current behavior. I don't see how adding this option complicates things or would have any negative effects.

> For example maybe we could proceed with renaming without showing the dialog but then we'd show how many files were affected through a message in a panel. That could also act as a quick way of switching to each modified file.

If this were the case, I'd still want the ability to disable it. A pop-up panel also sounds annoying.

jwortmann:
I guess this confirm dialog is inspired by the built-in "Replace" functionality from the Find in Files panel, because it uses the exact same text message.

But our rename command will open a TextInputHandler in all cases, before the dialog prompt, right?

How about we use the following text in `preview()` for the preview area under the input field:
"<kbd>Enter</kbd> to confirm, <kbd>Ctrl</kbd>+<kbd>Enter</kbd> to preview the changes"

Then we drop the confirm dialog, and for the normal execution (Enter) we show a `window.status_message()` "Made 42 replacements in 5 files". If something was unexpected for the user, they can stil just undo it with <kbd>Ctrl</kbd>+<kbd>Z</kbd>.

And with <kbd>Ctrl</kbd>+<kbd>Enter</kbd> (or the corresponding key for macOS) we can still do the functionality from the "Dry Run" button. I think something similar to this with the modifier key was used in the initial version of the recent "Goto Diagnostic in quick panel" PR. I'm not exactly sure whether it also works with TextInputHandlers though.

predragnikolic:
The confirm dialog was added to fix a case where someone renames a built-in type in pyls, like `str`. 
That caused more than 1000 files to be changed. 

The user couldn't not just press `ctrl+z` to undo that, because each file was edited with a different edit token. So the user must close each file manually and discard changes.

So the confirm modal was added to prevent cases like that.

Most editors do not have such a confirm modal, but rely on the textDocument/prepareRename which should confirm if a rename is valid for that range.

Most language servers wont even allow renaming of built in types. 

I find the confirmation modal useful. But I wouldn't mind if someone wants to disable the modal, given that LSP for ST implemented the textDocument/prepareRename.

jwortmann:
> The user couldn't not just press `ctrl+z` to undo that, because each file was edited with a different edit token. So the user must close each file manually and discard changes.

Ah, you're right. I guess then we either need to wait for https://github.com/sublimehq/sublime_text/issues/4604 to be implemented in ST, or rely on the current confirmation dialog.

plugin/core/types.py
```diff
@@ -198,6 +198,7 @@ class Settings:
     log_server = None  # type: List[str]
     lsp_code_actions_on_save = None  # type: Dict[str, bool]
     lsp_format_on_save = None  # type: bool
+    confirm_global_rename = None # type: bool
```
jwortmann, line 201: Seems that these entries are ordered alphabetically, and this one is inserted at a wrong place. The same applies to next next changed line below.

jonpalmisc:
> The confirm dialog was added to fix a case where someone renames a built-in type in pyls, like str. That caused more than 1000 files to be changed.

Perhaps the confirmation dialog could simply be made less eager? Rather than alerting on >1 files modified, only alert on >10? I can see the benefit in that. I think that >1 files just seems like a low threshold as that is easily triggered when renaming a C++ class that is defined in a header and implemented in a source file.

rwols:
> Perhaps the confirmation dialog could simply be made less eager? Rather than alerting on \>1 files modified, only alert on \>10? \>15? (The limit is arbitrary, really.) I can see the benefit in that.

I think this is a good approach. Can you change the setting into something like `rename_dialog_threshold` with a default value of `1`?

jonpalmisc:
For unrelated reasons, I'm not going to be using Sublime Text as my editor moving forward, and will not be continuing work on this PR — this can be closed if nobody else wants to take up the work.

predragnikolic:
@jonpalmisc thanks for taking the time to create the PR!

But I would close this PR tomorrow, if no one has nothing against.
We can revisit this, if other people also report that the rename popup is distracting.

PR #1927, opened 2022-01-06
Ensure view gets registered when tab is moved to create new window

jwortmann:
This is a workaround for the following ST bug: https://github.com/sublimehq/sublime_text/issues/4630

With this change, LspTextCommands still work after a tab is dragged out of a window to create a new window. For example lsp_hover will work again in such a case :)

The 1ms delay appears to be sufficient, so that this check still works:
https://github.com/sublimelsp/LSP/blob/19a01aa045de04bcc805e56043923656548050e0/plugin/documents.py#L205

### DIFF
--- a/boot.py
+++ b/boot.py
@@ -199,13 +199,18 @@ class Listener(sublime_plugin.EventListener):
     def on_pre_close_window(self, w: sublime.Window) -> None:
         windows.discard(w)
 
-    def on_post_move_async(self, view: sublime.View) -> None:
+    # Note: EventListener.on_post_move_async does not fire when a tab is moved out of the current window in such a way
+    # that a new window is created: https://github.com/sublimehq/sublime_text/issues/4630
+    # Hence, as a workaround we use on_pre_move, which still works in that case.
+    def on_pre_move(self, view: sublime.View) -> None:
         listeners = sublime_plugin.view_event_listeners.get(view.id())
         if not isinstance(listeners, list):
             return
         for listener in listeners:
             if isinstance(listener, DocumentSyncListener):
-                return listener.on_post_move_window_async()
+                # we need a small delay here, so that the DocumentSyncListener will recognize a possible new window
+                sublime.set_timeout_async(listener.on_post_move_window_async, timeout_ms=1)
+                return
 
     def on_load(self, view: sublime.View) -> None:
         file_name = view.file_name()
--- a/plugin/save_command.py
+++ b/plugin/save_command.py
@@ -1,6 +1,6 @@
 from .core.registry import LspTextCommand
 from .core.settings import userprefs
-from .core.typing import Callable, List, Optional, Type
+from .core.typing import Callable, List, Type
 from abc import ABCMeta, abstractmethod
 import sublime
 import sublime_plugin
--- a/plugin/save_command.py
+++ b/plugin/save_command.py
@@ -90,15 +90,6 @@ class LspSaveCommand(LspTextCommand):
         else:
             self._trigger_native_save()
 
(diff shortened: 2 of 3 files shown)


rchl (approved):

boot.py
```diff
@@ -199,13 +199,18 @@ def on_new_window_async(self, w: sublime.Window) -> None:
...
+    # Hence, as a workaround we use on_pre_move, which still works in that case.
+    def on_pre_move(self, view: sublime.View) -> None:
         listeners = sublime_plugin.view_event_listeners.get(view.id())
         if not isinstance(listeners, list):
             return
         for listener in listeners:
             if isinstance(listener, DocumentSyncListener):
-                return listener.on_post_move_window_async()
+                # we need a small delay here, so that the DocumentSyncListener will recognize a possible new window
+                sublime.set_timeout_async(listener.on_post_move_window_async, timeout_ms=1)
```
rchl, line 212: We usually just skip the timeout argument. Should be equivalent.

jwortmann: Should I remove or leave it? I figured it would be nice to see at first glance what this parameter "1" stands for.

rchl: You could skip the timeout value completely, including the number. I feel that that would be more meaningful since it would mean "as soon as possible after current message loop iteration finishes" instead of "after some arbitrary time".

In practice it would of course work the same pretty much.

jwortmann: No, that is not possible, see my comment in the line above. If there is no delay, the `view.window()` will still point to the old window, which will result the logic in `DocumentSyncListener.on_post_move_window_async` not working.
ST needs a little time (>0, but <1ms) after on_pre_move was fired, to create the new window (or to attach the tab to an existing window). At least it didn't work when I tested it without the 1ms delay.

jwortmann: With this 1ms delay we basically craft our own `on_post_move_async`, but without the bug from the same ST API function :)

rchl: If those two behave differently then it doesn't bode well for the reliability of this solution. It means it relies on some arbitrary time period that might not be guaranteed in all cases.

jwortmann: So what do you suggest then? Increase the delay to for example 5ms? Or keep the current behavior of LSP being mostly broken whenever a tab is dragged out of the window to create a new one?

rchl: Tried this in safe mode on Mac and "no timeout" works the same as 1ms timeout in this simple case.
(Though I can imagine that different platforms might behave differently)

```
import sublime
import sublime_plugin

class Listener(sublime_plugin.EventListener):
    def on_pre_move(self, view):
    	print('on_pre_move before', view.window())
    	sublime.set_timeout_async(lambda: print('on_pre_move after timeout (0ms)', view.window()))
    	sublime.set_timeout_async(lambda: print('on_pre_move after timeout (1ms)', view.window()), 1)
```

on_pre_move before Window(3)
on_pre_move after timeout (0ms) Window(6)
on_pre_move after timeout (1ms) Window(6)

jwortmann: I can confirm your results in safe mode on Windows, and surprisingly it seems to work for me in LSP without the delay as well now. But I can swear that when I tested this yesterday, that it didn't work for me without the delay. While I was debugging this, I even added a print statement in https://github.com/sublimelsp/LSP/blob/19a01aa045de04bcc805e56043923656548050e0/plugin/documents.py#L205-L206 and it would print everytime if, and only if I had not specified the delay.

When I use *Tools > Developer > Profile plugins*, it shows no other package under on_pre_move, which could possibly block the operation.

Not sure how we should proceed now. If we remove the delay, one option could be to keep the `is_enabled: return True` in the subclasses of LspTextCommand, for safety in case it doesn't work sometimes or for some users without the delay.

PR #1929, opened 2022-01-10
Make code actions lightbulb icon clickable

jwortmann:
I always wondered why this wasn't implemented yet, I would expect the dynamic lightbulbs to be interactive

I think it is a nice little addition, but feel free to close if popups from mouse hover over the gutter are not desired.

Most of the code is copied from hover.py, so maybe it would be possible / better to refactor the code a bit to reduce duplication, if we want to proceed to merge this.

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -117,8 +117,6 @@
   // "annotation" - show an annotation on the right when code actions are available.
   // "bulb" - show a bulb in the gutter when code actions are available.
   // "" - don't show code actions.
-  // Note: Due to API limitations, the "bulb" icon can not be clicked so the code actions can only be triggered
-  // using a keyboard shortcut or the context menu.
   "show_code_actions": "annotation",
 
   // Show code lens:
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -1,4 +1,5 @@
 from .code_actions import actions_manager
+from .code_actions import CodeActionOrCommand
 from .code_actions import CodeActionsByConfigName
 from .completion import LspResolveDocsCommand
 from .core.logging import debug
@@ -29,6 +30,7 @@ from .core.views import diagnostic_severity
 from .core.views import first_selection_region
 from .core.views import format_completion
 from .core.views import make_command_link
+from .core.views import make_link
 from .core.views import MarkdownLangMap
 from .core.views import range_to_region
 from .core.views import show_lsp_popup
@@ -40,6 +42,7 @@ from .session_view import SessionView
 from functools import partial
 from weakref import WeakSet
 from weakref import WeakValueDictionary
+import functools
 import itertools
 import sublime
 import sublime_plugin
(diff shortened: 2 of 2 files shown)


predragnikolic:
Hello,
I tried this, but it seems like it doesn't work.

When I click on the bulb in the gutter
I expect code actions to be shown,
but currently nothing is being shown:

output

jwortmann:
Well, the icon is not really clickable, sorry for the clickbait title ;)

It should show a small hover popup when you hover over the gutter icon.

lightbulb

Doesn't it work for you?

I think what happened in your GIF, is that when you clicked on the icon, it selected the full line and the caret moved to the next line. Because of that, the lightbulb went away (the popup should only show when there is currently a lightbulb at the same place where you hover over the gutter).

I've tested also with multiple code actions on a line, and then it should open the command palette when you click the link, just like how it works for the annotations and the "normal" hover popups.

predragnikolic:
Yeah, the title got me confused.

> Doesn't it work for you?

It does work when I hover over the icon. :)

jwortmann:
By the way, to give a bit more context; initially I intended to show *all* available code actions for the entire line when hovering over the lightbulb icon. But then I realized that the icon is not necessarily shown in the line where the diagnostics are, but rather where the selection starts. This is a result of how the code actions are requested for the combined covering region of the selection & diagnostic: https://github.com/sublimelsp/LSP/blob/19a01aa045de04bcc805e56043923656548050e0/plugin/documents.py#L527-L529

If the selection starts at a previous line, then the lightbulb will be shown at the line where the selection begins, and not at the line where the diagnostic is / begins. Therefore I used the exact same logic when showing the available code actions on gutter hover.

But I wonder whether LSP is doing the correct thing in general by combining the regions, when the code actions are requested from the server. It seems that this can cause possible issues when a code action is executed, too. Look at the following example:

This is when I execute a code action with an empty selection, using the Lua server (everything works as expected):

line

Now I execute the same code action, but with a selection starting at a previous line. Notice how the line comment from the code action gets inserted at a wrong position (at the line in front of the selection).

selection

Is this a bug in the Lua server, or should LSP better use only the region where the diagnostics are, when requesting code actions?

---

What do you think in general about the popup when hovering the gutter icon? We could also add the same for the diagnostic icons, i.e. a popup could show all the diagnostics which are contained in that line, but of course without all the additional symbol information which is often there in the normal hover popups.

rwols (changes requested):
I do think the code for building the minihtml should be shared between the two features, can you make that happen?

jwortmann:
Yes, I will try to do that.

rchl:
> Is this a bug in the Lua server, or should LSP better use only the region where the diagnostics are, when requesting code actions?

Looks like as server issue to me. It should know where the diagnostic is located and where the comment should be inserted so there it doesn't make sense that it chooses to insert it at the first line of the selection.

And for reference, it returns the edit in the initial code action response already. For example:

```
  {
    "edit": {
      "changes": {
        "file:///Users/rafal/workspace/temp/LUA-Projects/projects/1.Basics.lua": [
          {
            "newText": "---@diagnostic disable-next-line: unused-local\n",
            "range": {
              "end": {
                "character": 0,
                "line": 56
              },
              "start": {
                "character": 0,
                "line": 56
              }
            }
          }
        ]
      }
    },
    "kind": "quickfix",
    "title": "Disable diagnostics on this line (unused-local)."
  },
```

jwortmann:
The code to create the minihtml for the code actions is reused from hover.py now, and I removed the additional request when hovering over the lightbulb, because it wasn't really necessary.

There is still a bit of duplication of the code for handling of clicking on the link, but I think sharing this code would need some bigger refactoring of LspHoverCommand.

rwols (approved):

PR #1936, opened 2022-01-18
Remove registration of the legacy LanguageHandler class

rwols:
I can't find any package any more that still uses this.

### DIFF
--- a/boot.py
+++ b/boot.py
@@ -1,7 +1,6 @@
 import os
 import sublime
 import sublime_plugin
-import weakref
 
 # Please keep this list sorted (Edit -> Sort Lines)
 from .plugin.code_actions import LspCodeActionsCommand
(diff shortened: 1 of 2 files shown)


predragnikolic:
I guess only these 2 plugins that are published on PC still use LanguageHandler:

https://github.com/prometheus-community/sublimelsp-promql/blob/master/plugin.py#L130 (not part of the LSPSublimeOrg)
https://github.com/sublimelsp/LSP-Serenata/blob/master/plugin.py#L10

predragnikolic:
Will you also delete the `/LSP/plugin/core/handlers.py` file?

jfcherng:
> I guess only these 2 plugins that are published on PC still use LanguageHandler:
> 
> [prometheus-community/sublimelsp-promql@`master`/plugin.py#L130](https://github.com/prometheus-community/sublimelsp-promql/blob/master/plugin.py?rgh-link-date=2022-01-19T13%3A40%3A49Z#L130) (not part of the LSPSublimeOrg) [sublimelsp/LSP-Serenata@`master`/plugin.py#L10](https://github.com/sublimelsp/LSP-Serenata/blob/master/plugin.py?rgh-link-date=2022-01-19T13%3A40%3A49Z#L10)

LSP-Serenata hasn't updated for 2 years and the server, [Serenata](https://gitlab.com/Serenata/Serenata) was last active commit in 10 months ago. Consider PHP is evolving every year, I may assume both of LSP-Serenata and Serenata are gradually outdated...

rchl (approved):
In `boot.py` those are unused:

- import weakref
- `Callable` and `Tuple` from `.plugin.core.typing`

Also maybe we should be defining `__all__` to clearly indicate which imports are "public". That would also get rid of all false positive warnings that import is "unused".

PR #1944, opened 2022-02-15
Add a few examples for disabled_capabilities

### LINKED ISSUE #1840: Add ability to disable popup on hover

When you work with code, the popup often appears, even if you don't want it. I.e. in 90% cases you think how to write your program, cursor stops at random area, then LSP popup appears with some info about this area. But you don't need it. And this redundant flickering happens very often. For me is more convenient, when i have no random hover popups at all. If i need to know LSP info about variable, i manually invoke "lsp_hover" command via keybinding.

Currently i just overwrited source file, commented those lines https://github.com/sublimelsp/LSP/blob/main/plugin/documents.py#L430-L433
And added keybinding with "lsp_hover" command. But it would be cool if such thing available through settings. Something like
```
// Show hover popup. If disabled, it can be manually invoked via "lsp_hover" command in keybindings.
"show_popup_on_hover": true,
```

jwortmann:
~~Close~~ Related: #1840.

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -231,8 +231,12 @@
   //     // Sent once to server in initialize request
   //     "initializationOptions": { },
   //
-  //     // Disable providers if so desired
-  //     "disabled_capabilities": { },
+  //     // Disable providers if so desired - https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#serverCapabilities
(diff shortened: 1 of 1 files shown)


jwortmann:
Ah, actually the proposal in that issue was that the command still works if invoked via a key binding. Probably need a few small adjustments to still allow that.

PR #1948, opened 2022-03-02
Reorganize documentation for JavaScript/TypeScript

strager:
The language servers for JavaScript are scattered throughout the
documentation. Most other language servers are grouped by programming
language. Reorganize the documentation for JavaScript language servers
so it's easy to see what options are available for JS developers.

### DIFF
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -112,10 +112,6 @@ Follow installation instructions on [LSP-elm](https://github.com/sublimelsp/LSP-
     }
     ```
 
-## ESLint
-
-Follow installation instructions on [LSP-eslint](https://github.com/sublimelsp/LSP-eslint).
-
 ## F\#
 
 1. Install the [F#](https://packagecontrol.io/packages/F%23) package from Package Control for syntax highlighting.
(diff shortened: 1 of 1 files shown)


docs/src/language_servers.md
```diff
@@ -221,6 +213,28 @@ Follow installation instructions on [LSP-html](https://github.com/sublimelsp/LSP
...
+
+### ESLint
+
+Follow installation instructions on [LSP-eslint](https://github.com/sublimelsp/LSP-eslint).
+
+### Flow
+
+Follow installation instructions on [LSP-flow](https://github.com/sublimelsp/LSP-flow).
+
+### TypeScript
```
rchl, line 234: Maybe this header should also repeat the parent header (`JavaScript/TypeScript`)? It works for both.

rchl (approved):

docs/src/language_servers.md
strager: The tool used by LSP-typescript is called "TypeScript Language Server": https://github.com/typescript-language-server/typescript-language-server

If we rename the TypeScript subsection to JavaScript/TypeScript, should we also rename the Flow subsection to JavaScript/Flow?

rchl: Well, it's hard to be fully precise and specific here so I guess it's not a big deal and we can leave as is.

`eslint` can also run on Vue files and possibly other with dedicated processors so it's not really possible to specify all the languages.

rwols (approved):

PR #1950, opened 2022-03-11
Fix "side by side go-to" if file is already open

### LINKED ISSUE #1946: Side by side option for symbol action links in hover popup doesn't work if location is in same file
**Describe the bug**
The side by side icon link for "Definition" / "Type Definition" / "Declaration" from the hover popup doesn't work if the location of the definition/declaration is in the same file.

**To Reproduce**
1. Have `"show_symbol_action_links": true` in the settings (this is the default value)
2. Hover over symbol (e.g. function call) which has a definition in the same file
3. Click on ◨ next to "Definition", or use <kbd>Ctrl</kbd> + click on the text link
4. See that the view scrolls to the location, instead of opening the location in a new tab to the right

**Expected behavior**
LSP should open the definition in a new new to the right, similar to how the built-in definitions popup from ST does

- OS: Win 10
- LSP version: main

**Additional context**

Seems like the `flags` argument which includes the "side_by_side" information is lost/ignored here:
https://github.com/sublimelsp/LSP/blob/1bcd518102c1516c9d808c974b7d2a5eba7d0b80/plugin/core/open.py#L30-L31

rchl:
Resolves #1946


rchl:
I'm noticing now that views opened with this code branch don't have LSP initialized on them...

rwols (approved):

rwols (changes requested):
views opened with this code branch don't have LSP initialized on them

rchl:
I guess it should be reported as an upstream bug because I don't see how we can get this working.
The sheet opened like that appears to:
- have the samve View ID as the already open view
- not trigger `ViewEventListener` listeners (for example `on_hover`)

It's possibly a bug in ST or just some limitation without a way around.

rwols:
Ah okay :) I thought you meant that this branch introduced a bug which isn't present on the main branch.

rwols (approved):

Issue #1951, opened 2022-03-17
LSP's `rename` fails in a particular scenario.


Ultra-Instinct-05:
**Describe the bug**
Apologies for the vague issue title but I am not sure what else to call it other than the above 🙂 

It seems that LSP's `rename` feature can't handle a particular scenario, when doing a rename in a Prisma schema file. The scenario is detailed as follows.

**To Reproduce**
1. Install the [LSP-prisma](https://github.com/Sublime-Instincts/LSP-prisma) package.
2. Install the [Prisma](https://packagecontrol.io/packages/Prisma) package.
3. Create a `schema.prisma` file somewhere and paste the following

```
model Model {
  id Int @id
  role Role @default(A_VARIANT_WITH_UNDERSCORES)
}

enum Role {
  ADMIN
  MODERATOR
  A_VARIANT_WITH_UNDERSCORES
}
```
4. Right click on `A_VARIANT_WITH_UNDERSCORES` and use `LSP -> Rename`. Type anything for the new name and press <kbd>enter</kbd>

**Expected behavior**
It seems VS Code is able to properly rename it.

**Actual Behavior**
ST's `LSP` package seems to fail with the following traceback in the console.

```
Traceback (most recent call last):
  File "C:\Sublime Text\Lib\python33\sublime_plugin.py", line 1488, in run_
    return self.run(edit, **args)
  File "C:\Users\DELL\AppData\Roaming\Sublime Text\Installed Packages\LSP.sublime-package\plugin/edit.py", line 38, in run
  File "C:\Sublime Text\Lib\python33\sublime.py", line 1493, in text_point_utf16
    return sublime_api.view_text_point_utf16(self.view_id, row, col_utf16, clamp_column)
OverflowError: int too big to convert
``` 

**Screenshots**

![sublime_text](https://user-images.githubusercontent.com/34304644/158831055-1f9a5339-4f64-4d7e-944a-9b0d0fedcd82.gif)

**Environment (please complete the following information):**
- OS: Windows 11
- Sublime Text version: 4129
- LSP version: 1.16.1
- Language servers used: https://github.com/prisma/language-tools/tree/main/packages/language-server

**Additional context**
No changes are made in `LSP-prisma.sublime-settings`. It's just the default what is shipped.

rwols:
I’m guessing this is the issue

[...]

PR #1952, opened 2022-03-17
Guard for overflow of the column offset when applying edits

rwols:
We could also guard for overflow from the client side. This is all totally hypothetical and I haven't actually tried using LSP-Prisma yet.

Related: #1951


predragnikolic (approved):
I can confirm that the issue is fixed.

PR #1957, opened 2022-04-03
Check for `:` and `/` while updating nested dict structures in DottedDict

### LINKED ISSUE #1858: Unwanted splitting on dot of the setting keys

rwols:
This is another attempt at fixing #1858. This time we keep the old behavior of
merging dicts only for the custom client dicts.

Fixes #1858

### DIFF
--- a/plugin/core/collections.py
+++ b/plugin/core/collections.py
@@ -113,7 +113,9 @@ class DottedDict:
         """
         return bool(self._d)
 
-    def __contains__(self, path: str) -> bool:
+    def __contains__(self, path: object) -> bool:
+        if not isinstance(path, str):
+            return False
         value = self.get(path)
         return value is not None and value is not False
 
--- a/plugin/core/collections.py
+++ b/plugin/core/collections.py
@@ -154,7 +154,7 @@ class DottedDict:
         return sublime.expand_variables(self._d, variables)
 
     def _update_recursive(self, current: Dict[str, Any], prefix: str) -> None:
-        if not current:
+        if not current or any(filter(lambda key: isinstance(key, str) and (":" in key or "/" in key), current.keys())):
             return self.set(prefix, current)
         for key, value in current.items():
             path = "{}.{}".format(prefix, key)
--- a/tests/test_collections.py
+++ b/tests/test_collections.py
@@ -137,3 +137,45 @@ class DottedDictTests(TestCase):
         self.assertEqual(d.get(), {"a": {}})
         d.update({"a": {"b": {}}})
         self.assertEqual(d.get(), {"a": {"b": {}}})
+
+    def test_from_base_and_override(self) -> None:
+        base = DottedDict({
+            "yaml.schemas": {}
+        })
+        override = {
+            "yaml.schemas": {
+                "http://foo.com/bar.json": "**/*.json"
+            }
+        }
+        result = DottedDict.from_base_and_override(base, override)
+        self.assertEqual(
+            result.get(None),
+            {
+                "yaml": {
+                    "schemas": {
+                        "http://foo.com/bar.json": "**/*.json"
+                    }
+                }
+            }
+        )
+
+    def test_update_with_dicts(self) -> None:
+        base = {
+            "settings": {
+                "yaml.schemas": {}
+            }
+        }
+        overrides = {
+            "yaml.schemas": {
+                "http://foo.com/bar.json": "**/*.json"
+            }
+        }
+        settings = DottedDict(base.get("settings", {}))
+        settings.update(overrides)
+        self.assertEqual(settings.get(), {
+            "yaml": {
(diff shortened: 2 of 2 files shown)


plugin/core/collections.py
```diff
@@ -135,7 +135,17 @@ def update(self, d: Dict[str, Any]) -> None:
...
+        """
+        for key, value in d.items():
+            self.set(key, value)
+
+    def update_recursive(self, d: Dict[str, Any]) -> None:
+        """
+        Overwrite and/or add new key-value pairs to the collection.
+
         :param      d:    The overriding dictionary. Can contain nested dictionaries.
+                          The key is dynamically built up from pieces.
```
rchl, line 148: When we have default settings (in this example LSP-eslint):

```js
{
	"settings": {
		"codeAction.disableRuleComment": {
			"enable": true,
			"location": "separateLine"
		},
}
```

and we want to override just the `enable` setting with:

```js
{
    "settings": {
        "codeAction.disableRuleComment": {
            "enable": true
        },
}
```

then the `settings.codeAction.disableRuleComment.location` will be gone in the resulting object.

I'm guessing that it's not how it's supposed to work? In VSCode the settings are merged correctly.

rchl (changes requested):

plugin/core/collections.py
rchl: Overriding with `"codeAction.disableRuleComment.enable": true,` works but it wouldn't match the schema so would trigger a warning diagnostic.

rwols: So I looked into this and we can't have both your above described behavior _and_ fix the `http://foo.com -> http://foo  com` bug, unless we start requiring a schema and parsing that schema.

rchl: I'm still quite confused about how VSCode handles it. I don't think it actually uses schema for that. At least haven't found any evidence of that when checking out its code before.

rchl: It looks to me like VSCode just doesn't do expansion recursively.

I've tested a bit on the `yaml` server in VSCode. Defined settings like:

```js
  "yaml.schemas": {
    "test.me": "foo"
  },
  "yaml.schemas.foo": {
    "test.me": "foo"
  },
  "yaml.foo": {
    "test.me": "abc"
  },
```

Where only the first setting matches the schema. This resulted in server receiving these settings in `workspace/configuration` request:

```js
        "schemas": {
            "test.me": "foo",
            "foo": {
                "test.me": "foo"
            }
        },
        "foo": {
            "test.me": "abc"
        }
```

So schema doesn't seem to matter here. What it seems to do is that it only expands the keys at the root of the `settings` object.

rchl: I'm thinking that we could possibly **only** support dotted keys at the root level of the `settings` object. Possibly not even in `initializationSettings`. Though ideally one would go through all LSP(-*) configuration files to see if that is likely to break anything.

rwols:
I gave up looking for a pretty solution. We can check whether there's a `:` or `/` character in the key. In that case it's likely some URL.

The problem was that at startup we create a `ClientConfigs` in settings.py. Plugins then call `add_external_config` on it. That in turn calls `ClientConfig.from_sublime_settings`. That in turn calls `settings.update(read_dict_setting(s, "settings", {}))  # overrides from the user`. That in turn starts recursively building up the keys in nested dicts. We now check in that method whether a key candidate contains "forbidden" characters.

rchl:
> I gave up looking for a pretty solution. We can check whether there's a `:` or `/` character in the key. In that case it's likely some URL.

I really don't like this magic approach. So much in fact that I've spent some time trying out my suggested solution where we would only parse dotted keys at the root of the `settings` object (and for backwards-compat potentially also at the root of `initializationOptions`).

I have horrific hack at https://github.com/sublimelsp/LSP/commits/fix/dotted-keys-level that does that in an incomplete fashion (no correct handling of project settings) but I think it's doable. Not sure when I would have enough time to complete it though.

If you insist, you could get your fix in anyway and it could potentially be replaced with my suggestion eventually.

rwols:
> If you insist, you could get your fix in anyway and it could potentially be replaced with my suggestion eventually.

That is fine for me.

rwols:
Any objections left? :)

rchl:
No. Should be fine for now.

PR #1961, opened 2022-04-11
Fix LspShowScopeNameCommand for the new ContextStackFrame type

### LINKED ISSUE #1943: Update lsp_show_scope_name command for ST 4127
It looks like the `View.context_backtrace()` API function had a backward incompatible change on ST 4127, so I guess the scope name popup from LSP is broken for new ST versions now. This only impacts active language servers which support semantic tokens, and only if the "semantic_highlighting" setting is enabled.

A solution would be to add something like `if sublime.version() > "4126"` and copy the relevant new code from *Default/show_scope_name.py* into this if-branch. Or probably even better would be to add a new tag at https://github.com/sublimelsp/repository/blob/ccc25996a2ee05d70457344219a6dd62130ad2ac/repository.json#L59-L68 for the LSP package. The current version of LSP does no longer work on ST 4070 anyway.

rwols:
Resolves #1943

### DIFF
--- a/plugin/semantic_highlighting.py
+++ b/plugin/semantic_highlighting.py
@@ -1,5 +1,5 @@
 from .core.registry import LspTextCommand
-from .core.typing import List, Tuple, cast
+from .core.typing import Any, List, Tuple, cast
 import sublime
 import os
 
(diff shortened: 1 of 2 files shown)


rchl (approved):
Haven't looked much at the changes but works fine. :)

PR #1962, opened 2022-04-14
Fix WSL path URI decoding under Windows

jfcherng:
```
Python 3.3>>> from LSP.plugin.core.url import filename_to_uri, parse_uri, urlparse

Python 3.3>>> filename_to_uri('\\\\wsl$\\Ubuntu-20.04\\home\\jfcherng\\Desktop\\docs\\a.py')
'file://wsl%24/Ubuntu-20.04/home/jfcherng/Desktop/docs/a.py'
           
Python 3.3>>> parse_uri('file://wsl%24/Ubuntu-20.04/home/jfcherng/Desktop/docs/a.py')
('file', '\\\\wsl%24\\Ubuntu-20.04\\home\\jfcherng\\Desktop\\docs\\a.py')
                 ^^^ this is wrong. should be $

Python 3.3>>> urlparse('file://wsl%24/Ubuntu-20.04/home/jfcherng/Desktop/docs/a.py')
ParseResult(scheme='file', netloc='wsl%24', path='/Ubuntu-20.04/home/jfcherng/Desktop/docs/a.py', params='', query='', fragment='')
                                      ^^^ this is wrong. should be $
```

Notice that `netloc` is `wsl%24` after `urlparse()`. We need to decoding that before doing `"\\\\{}\\{}".format(netloc, path)` just like what we do for `path`.

Related issue: https://discord.com/channels/280102180189634562/280157083356233728/963549081886666784

### DIFF
--- a/plugin/core/url.py
+++ b/plugin/core/url.py
@@ -49,8 +49,8 @@ def parse_uri(uri: str) -> Tuple[str, str]:
     parsed = urlparse(uri)
     if parsed.scheme == "file":
         path = url2pathname(parsed.path)
-        netloc = url2pathname(parsed.netloc)
         if os.name == 'nt':
+            netloc = url2pathname(parsed.netloc)
             path = path.lstrip("\\")
             path = re.sub(r"^([a-z]):", _uppercase_driveletter, path)
             if netloc:
--- a/plugin/core/url.py
+++ b/plugin/core/url.py
@@ -49,12 +49,13 @@ def parse_uri(uri: str) -> Tuple[str, str]:
     parsed = urlparse(uri)
     if parsed.scheme == "file":
         path = url2pathname(parsed.path)
+        netloc = url2pathname(parsed.netloc)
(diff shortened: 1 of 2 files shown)


plugin/core/url.py
```diff
@@ -49,12 +49,13 @@ def parse_uri(uri: str) -> Tuple[str, str]:
     parsed = urlparse(uri)
     if parsed.scheme == "file":
         path = url2pathname(parsed.path)
+        netloc = url2pathname(parsed.netloc)
```
rwols, line 52: this variable can go into the `if os.name == 'nt'` block

rwols (changes requested):

rwols (approved):

rwols (changes requested):
Can you add a test for this so we don’t regress in the future? There are windows-specific tests that run in the CI

rwols (approved):

PR #1964, opened 2022-04-27
Follow global setting whether to show snippet completions

### LINKED ISSUE #1939: Snippets Clangd C++
**Describe the bug**
C++ completion using clangd is giving snippets despite these settings:

Main User Preferences
```json
"auto_complete_include_snippets": false,
"auto_complete_include_snippets_when_typing": false,
"ignored_snippets": [
    "C++/*",		
],

```

LSP Settings
```json
"inhibit_snippet_completions": false,
"clients": {
    "clangd": {
	"enabled": true,
	"command": [
			"clangd",
			"--background-index"
	],
    },
}
```

**To Reproduce**
You could for example type: `for<tab>` and you will be snippetted! (<- that's should be correct English! :)

**Expected behavior**
Not heaving snippets in the auto completion list!

**Screenshots**
image

**Environment**
- OS: Windows 10 x64
- Sublime Text version: 4126
- LSP version: 1.15.0
- Language servers used: clangd: 13.0.0

**Additional context**
If I would turn of clangd in the clients
```json
"clangd": {
    "enabled": false,
}
```
then these specific snippets will not show up. At this point I am assuming that they are provided by clangd. The default c++ snippets provided in the default Sublime packages - if turned on - have a different layout.

jwortmann:
Closes #1939 by using the *"auto_complete_include_snippets"* setting from ST to decide whether completions with the "snippet" kind should be included.

Addendum to what I wrote in https://github.com/sublimelsp/LSP/issues/1939#issuecomment-1109749912:

As far as I understand, the "snippet" *kind* should more or less represent the same as what a snippet is in ST, and whether a CompletionItem includes "snippet syntax" with `$0`, etc. tab stops & placeholders is designated by the "insertTextFormat" field (even though this distinction isn't described clearly in the LSP specs and they probably overlap in most cases):

```ts
/**
 * Defines whether the insert text in a completion item should be interpreted as
 * plain text or a snippet.
 */
export namespace [InsertTextFormat](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#insertTextFormat) {
	/**
	 * The primary text to be inserted is treated as a plain string.
	 */
	export const PlainText = 1;

	/**
	 * The primary text to be inserted is treated as a snippet.
	 *
	 * A snippet can define tab stops and placeholders with `$1`, `$2`
	 * and `${3:foo}`. `$0` defines the final tab stop, it defaults to
	 * the end of the snippet. Placeholders with equal identifiers are linked,
	 * that is typing in one will update others too.
	 */
	export const Snippet = 2;
}
```

### DIFF
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -23,6 +23,34 @@ class DiagnosticTag:
     Deprecated = 2
 
 
+class CompletionItemKind:
+    Text = 1
+    Method = 2
+    Function = 3
+    Constructor = 4
+    Field = 5
+    Variable = 6
+    Class = 7
+    Interface = 8
+    Module = 9
+    Property = 10
+    Unit = 11
+    Value = 12
+    Enum = 13
+    Keyword = 14
+    Snippet = 15
+    Color = 16
+    File = 17
+    Reference = 18
+    Folder = 19
+    EnumMember = 20
+    Constant = 21
+    Struct = 22
+    Event = 23
+    Operator = 24
+    TypeParameter = 25
+
+
 class CompletionItemTag:
     Deprecated = 1
 
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -5,6 +5,7 @@ from .completion import LspResolveDocsCommand
 from .core.logging import debug
 from .core.promise import Promise
 from .core.protocol import CompletionItem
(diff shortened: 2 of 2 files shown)


rwols (approved):
This looks like a good solution, thanks!

PR #1971, opened 2022-05-30
docs: add info about enabling clangd server

istudyatuni:
First of all, I'm not sure of the wording. Also, I'm not sure if this information is really needed, but I'm having trouble finding information about activation of the server. And also, maybe it's better to write about this after

> You will thus have to adjust your `"clients"` dictionary in your user preferences.

and not at the end of page.


rwols (approved):

rwols:
Thank you for the clarification!

PR #1974, opened 2022-06-08
Add support for textDocument/documentLink request

jwortmann:
Is this useful? Tested with LSP-TexLab:

lsp

A strange (or maybe just inaccurate) thing in the LSP specs is that a "document link"
> links to an internal or external resource, like another text document or a web site

but the type for `DocumentLink.target` is `DocumentUri`, and not the more general `URI` (they both are defined as `string` though).

---

Initially I figured it might be good to make the links available in the right click context menu, similar to what `Default/open_context_url.py` does for website links, and to run the request only on demand when you open the context menu. But then we can't show the underlines for the links, and it would probably make the context menu very slow (I don't know if it would even be possible to run the request from the `sublime_plugin.TextCommand.is_enabled()` and wait for a response). So I guess to run the request in the background after text changed and show the link in a hover popup (like what VS Code does) is better and looks nicer.

---

### TODO

- [x] Add support for `documentLink/resolve` (when the "target" is not provided in the initial response)
- [ ] Maybe add some tests

### Questions

- Are the style options useful? Is the default value good?
- Do we need an additional mouse binding (Alt + click) like in VS Code to directly open the link?

### DIFF
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -295,6 +295,13 @@ CompletionList = TypedDict('CompletionList', {
     'items': List[CompletionItem],
 }, total=True)
 
+DocumentLink = TypedDict('DocumentLink', {
+    'range': RangeLsp,
+    'target': DocumentUri,
+    'tooltip': str,
+    'data': Any
+}, total=False)
+
 MarkedString = Union[str, Dict[str, str]]
 
 MarkupContent = Dict[str, str]
@@ -385,6 +392,10 @@ class Request:
         return Request("textDocument/documentHighlight", params, view)
 
     @classmethod
+    def documentLink(cls, params: Mapping[str, Any], view: sublime.View) -> 'Request':
+        return Request("textDocument/documentLink", params, view)
+
+    @classmethod
     def semanticTokensFull(cls, params: Mapping[str, Any], view: sublime.View) -> 'Request':
         return Request("textDocument/semanticTokens/full", params, view)
 
@@ -399,6 +410,10 @@ class Request:
     @classmethod
     def resolveCompletionItem(cls, params: CompletionItem, view: sublime.View) -> 'Request':
         return Request("completionItem/resolve", params, view)
+
+    @classmethod
+    def resolveDocumentLink(cls, params: DocumentLink, view: sublime.View) -> 'Request':
+        return Request("documentLink/resolve", params, view)
 
     @classmethod
     def shutdown(cls) -> 'Request':
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -261,7 +261,7 @@ class LspHoverCommand(LspTextCommand):
         contents = self.diagnostics_content() + hover_content + code_actions_content(self._actions_by_config)
         link_content, link_has_standard_tooltip = self.document_link_content(listener, point)
         if userprefs().show_symbol_action_links and contents and not only_diagnostics and hover_content:
-            symbol_actions_content =  self.symbol_actions_content(listener, point)
+            symbol_actions_content = self.symbol_actions_content(listener, point)
             if link_content and link_has_standard_tooltip:
                 symbol_actions_content += ' | ' + link_content
             elif link_content:
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -3,6 +3,7 @@ from .code_actions import CodeActionOrCommand
 from .core.logging import debug
 from .core.promise import Promise
 from .core.protocol import Diagnostic
+from .core.protocol import DocumentLink
 from .core.protocol import Error
 from .core.protocol import ExperimentalTextDocumentRangeParams
 from .core.protocol import Hover
@@ -223,15 +224,24 @@ class LspHoverCommand(LspTextCommand):
             if sv.has_capability_async("documentLinkProvider"):
                 link = sv.session_buffer.get_document_link_at_point(sv.view, point)
                 if link is not None:
-                    title = link.get("tooltip") or "Follow link"
                     target = link.get("target")
-                    if target is not None:
-                        contents.append('<a href="{}">{}</a>'.format(target, title))
-                    else:
-                        # TODO send documentLink/resolve request
-                        # TODO maybe figure out how "Promise" works and if it could help here
-                        pass
+                    if not target:
+                        if not sv.has_capability_async("documentLinkProvider.resolveProvider"):
+                            continue
+                        sv.session.send_request_task(Request.resolveDocumentLink(link, sv.view)).then(
+                            lambda result: self._on_resolved_link(sv.session_buffer, result))
+                        link = sv.session_buffer.get_document_link_at_point(sv.view, point)  # resolved link ???
+                        if not link:
+                            continue
+                        target = link.get("target")
+                        if not target:
+                            continue
+                    title = link.get("tooltip") or "Follow link"
+                    contents.append('<a href="{}">{}</a>'.format(target, title))
         return '<div class="link">' + '<br>'.join(contents) + '</div>' if contents else ''
+
+    def _on_resolved_link(self, session_buffer: SessionBufferProtocol, link: DocumentLink) -> None:
+        session_buffer.update_document_link(link)
 
     def hover_content(self) -> str:
         contents = []
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -216,21 +216,21 @@ class LspHoverCommand(LspTextCommand):
         return "".join(formatted)
 
     def document_link_content(self, listener: AbstractViewListener, point: int) -> str:
-        if userprefs().link_highlight_style in ("underline", "none"):
-            for sv in listener.session_views_async():
-                if sv.view == self.view and sv.has_capability_async("documentLinkProvider"):
-                    sb = sv.session_buffer
-                    for link in sb.get_document_links():
-                        if link.contains(point):
-                            title = link.tooltip or "Follow link"
-                            if link.target is not None:
-                                return '<div class="link"><a href="{}">{}</a></div>'.format(link.target, title)
-                            else:
-                                # TODO send documentLink/resolve request
-                                # TODO maybe figure out how "Promise" works and if it could help here
-                                return ""
-                    break
-        return ""
+        if userprefs().link_highlight_style not in ("underline", "none"):
+            return ""
+        contents = []
+        for sv in listener.session_views_async():
+            if sv.has_capability_async("documentLinkProvider"):
+                for link in sv.session_buffer.get_document_links():
+                    if link.contains(point):
+                        title = link.tooltip or "Follow link"
+                        if link.target is not None:
+                            contents.append('<a href="{}">{}</a>'.format(link.target, title))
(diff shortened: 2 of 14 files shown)


plugin/hover.py
```diff
@@ -215,6 +215,23 @@ def diagnostics_content(self) -> str:
...
+                    for link in sb.get_document_links():
+                        if link.contains(point):
+                            title = link.tooltip or "Follow link"
+                            if link.target is not None:
+                                return '<div class="link"><a href="{}">{}</a></div>'.format(link.target, title)
+                            else:
+                                # TODO send documentLink/resolve request
+                                # TODO maybe figure out how "Promise" works and if it could help here
+                                return ""
+                    break
```
rchl, line 232: Any reason for breaking the loop instead of collecting links from all sessions?

plugin/core/protocol.py
```diff
@@ -295,6 +295,13 @@ class SemanticTokenModifiers:
     'items': List[CompletionItem],
 }, total=True)
 
+DocumentLink = TypedDict('DocumentLink', {
+    'range': RangeLsp,
+    'target': DocumentUri,
+    'tooltip': str,
+    'data': Any
+}, total=False)
```
rchl, line 303: Would it be better to have `total=True` and mark `target`, `tooltip` and `data` as optional?
With the way it is now `range` triggers a type error in the part of the code where it's accessed.

Neither `total=False` nor `total=True` is fully accurate. The former because it makes all properties optional while `range` isn't optional, the latter because missing property is not strictly the same as `Optional` property. Thus I would pick whatever is more functional for us.

plugin/hover.py
```diff
@@ -215,6 +215,23 @@ def diagnostics_content(self) -> str:
             formatted.append("</div>")
         return "".join(formatted)
 
+    def document_link_content(self, listener: AbstractViewListener, point: int) -> str:
+        if userprefs().link_highlight_style in ("underline", "none"):
+            for sv in listener.session_views_async():
+                if sv.view == self.view and sv.has_capability_async("documentLinkProvider"):
+                    sb = sv.session_buffer
+                    for link in sb.get_document_links():
```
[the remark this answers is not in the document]
jwortmann: Yeah, I've already seen this. This part of the code is basically a first attempt to make it work and should certainly be improved or rewritten. Will do later.

plugin/hover.py
```diff
@@ -215,6 +215,23 @@ def diagnostics_content(self) -> str:
             formatted.append("</div>")
         return "".join(formatted)
 
+    def document_link_content(self, listener: AbstractViewListener, point: int) -> str:
+        if userprefs().link_highlight_style in ("underline", "none"):
+            for sv in listener.session_views_async():
+                if sv.view == self.view and sv.has_capability_async("documentLinkProvider"):
```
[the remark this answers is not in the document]
jwortmann: I think you are correct and it is not needed.

plugin/core/protocol.py
jwortmann: I would think that `total=False` is better, because I directly store the links from the response as `DocumentLink` now, and otherwise the types for links with missing properties wouldn't really be correct as well. This error for `link["range"]` is reported from Pyright, but mypy doesn't seem to have a problem with it. And we already use the same a few lines below for `diagnostic["range"]`, with the same error from Pyright.

Since we know from the specs that "range" is a mandatory property, we could just leave it as it is, or alternatively use the less clear (imo) `link.get("range")`, or add a `# type: ignore` for Pyright.

rwols:
Awesome work @jwortmann. I do see that the "follow link" link is being positioned incorrectly with respect to the CSS? See screenshot below.

image

(for clangd one doesn't really need textDocument/documentLink because "goto definition" also opens the link to the #include)

Couple of boring questions:

- Should "Follow Link" also go into the "Goto" menu of the menu bar?
- Should "Follow Link" also go into the context menu when you right click a link?
- I think some users may want to key-bind this functionality. Maybe add a commented-out example keybinding?
- I fear that some users may get confused by the (default) blue underlines because those are also used (by default) as hints/info diagnostics. The [OpenUri](https://packagecontrol.io/packages/OpenUri) package uses an arrow emoji for an additional hint that the text is a link that you can follow. Do you think that such an arrow emoji would help? cc @jfcherng

plugin/core/protocol.py
rchl: I'll resolve this but I would still prefer the other way around.

There would be an issue with `total=True` had we have to create such typed dict manually because we couldn't create it without providing the optional properties. But since we don't create it manually and rather just use what the LSP server sends us, there wouldn't be an issue.

Proper support for defining such objects seems to be implemented in Python 3.11+ (https://peps.python.org/pep-0655/) and is available in older versions through the https://github.com/python/typing_extensions/ module. We could vendor that module but that's a separate discussion.

jwortmann:
Thanks for the feedback. I will try to fix the CSS alignment in the popup, but it is a bit inconvenient, because if I add padding, it doesn't look nice if the link is the only content of the popup. And probably needs some tweaking dependent on whether users have disabled the "show_symbol_action_links".

> (for clangd one doesn't really need textDocument/documentLink because "goto definition" also opens the link to the #include)

Yes, it is the same case for LSP-TexLab. In general "document links" and "goto definition" seem to have overlapping functionality. I would guess that the "document links" are supposed to allow a distinct UI presentation for links in the source code, like the underlining or a popup on mouse hover.

>     * Should "Follow Link" also go into the "Goto" menu of the menu bar?

Added.

>     * Should "Follow Link" also go into the context menu when you right click a link?

Hm, not sure if this is necessary, because there is already the popup on mouse hover and I think navigation via this popup is faster than using right click and then selecting a sub-menu entry of the context menu.

>     * I think some users may want to key-bind this functionality. Maybe add a commented-out example keybinding?

Also done. If you want, you could add a mouse binding too:
```jsonc
// Default.sublime-mousemap
[
    {
        "button": "button1", "count": 1, "modifiers": ["alt"],
        "press_command": "lsp_open_link"
    },
]
```
It works fine for the links, but unfortunately this example with "Alt + click" would take priority over (and therefore disable) the built-in ability of subtractive selections (Windows & Linux) or column-wise selections (macOS).

>     * I fear that some users may get confused by the (default) blue underlines because those are also used (by default) as hints/info diagnostics.

It is not necessarily blue, the color depends on whatever the color scheme uses for the `markup.underline.link` scope. And the underlining with this color just reflects the standard highlighting of internet links (and some other links) in various syntaxes and color schemes, for example in Markdown, JavaDoc, C# comments etc.
The built-in color schemes all have them underlined.

And the diagnostics for info & hint are dotted, so it should be easy to distinguish those. What we perhaps could do, is to use no underline ("none") as default value for the "link_highlight_style" setting?

plugin/core/protocol.py
jwortmann: I mean we could change it, but then we should also change all of the other TypedDict's which currently use `total=False` to `total=True`, e.g. Diagnostic, CompletionItem, etc.
There are countless `reportTypedDictNotRequiredAccess" errors reported from Pyright all over the source code of this repo.

But this could probably be done in another pull request, I guess.

plugin/hover.py
```diff
@@ -215,6 +216,33 @@ def diagnostics_content(self) -> str:
...
+            if sv.has_capability_async("documentLinkProvider"):
+                link = sv.session_buffer.get_document_link_at_point(sv.view, point)
+                if link is not None:
+                    target = link.get("target")
+                    if not target:
+                        if not sv.has_capability_async("documentLinkProvider.resolveProvider"):
+                            continue
+                        sv.session.send_request_task(Request.resolveDocumentLink(link, sv.view)).then(
+                            lambda result: self._on_resolved_link(sv.session_buffer, result))
+                        link = sv.session_buffer.get_document_link_at_point(sv.view, point)  # resolved link ???
```
jwortmann, line 233: I've added support to resolve unresolved links via `documentLink/resolve` now, tested and confirm that it works with LSP-Dockerfile :slightly_smiling_face:

There are no problems when called via the command from the menu, but the hover popup will only work after I hover a second time over the link. Could someone help me and explain how I could run the request from the highlighted lines in a blocking way? I've tried with `Promise.then()` as shown here, which still seems to run async, and I've tried `Promise.resolve()` which doesn't work at all.

rchl: There are two ways to handle it:
 - wait for all requests to finish and then show a popup
 - show the popup as soon as possible and then update it when document link + resolve request finishes

I suppose we want the latter as this functionality is not essential enough to delay the popup?
The downside of that is that the content will visibly change (potentially shift) after showing but that's not a big deal IMO and if the link element is the one at the very bottom then it's a complete non-issue.

plugin/hover.py
```diff
@@ -229,8 +259,16 @@ def show_hover(self, listener: AbstractViewListener, point: int, only_diagnostic
     def _show_hover(self, listener: AbstractViewListener, point: int, only_diagnostics: bool) -> None:
         hover_content = self.hover_content()
         contents = self.diagnostics_content() + hover_content + code_actions_content(self._actions_by_config)
-        if contents and not only_diagnostics and hover_content:
-            contents += self.symbol_actions_content(listener, point)
+        link_content, link_has_standard_tooltip = self.document_link_content(listener, point)
```
rchl, line 262: Doesn't seem right to support multiple links but have just a single `link_has_standard_tooltip`. I feel that each link should have its own. So I think it would make more sense if the `document_link_content` would return a list of tuples instead.

That said, having multiple links is probably gonna be such a rare case that I'm not sure how much we should care.

rchl: Now I've noticed the `len(contents) > 1` logic that sets it to `false` when there are multiple links so I guess it's OK.

jwortmann: Yes, the reason is that I append the link to the "symbol action links" now, but only if it is a single link (i.e. from one session) and if it doesn't have a custom title. Just so we know that the title isn't too long, because perhaps there is a server which uses the URL or some other long text as the tooltip/title. Maybe another variable name "can_be_shown_in_symbol_action_links" would be more clearly. But it's a bit long, so...

Unbenannt

plugin/hover.py
jwortmann: Thanks, it works great!

jwortmann:
Any opinions about the default value for the link style? I'm undecided now, or even slightly lean to remove the underlining ("none"). I think as soon as you discover what your language server treats as links, the underlining provides little value. But maybe I'm biased, because it doesn't look very nice in my own color scheme.

---

By the way, Dockerfiles are an example where the links provide an actual feature not covered by "Goto Definition". For example in
```dockerfile
FROM debian:latest
```
the link on `debian` will take you to the corresponding Docker website, and there is no "Goto Definition" available.

rchl:
Since the work to implement this feature was made, I think the results should be easily discoverable by the user. Otherwise it's kind of a dead feature. So personally I think we should keep the default underline.

rwols:
I just noticed another example is LSP-json for schema files, where you can do a "follow link" on a `$ref`, very nice!

PR #1986, opened 2022-07-04
Optionally fallback to goto_definition in lsp_symbol_definition

timfjord:
Unfortunatelly not all LSPs have good symbol definition support. For example, `solargraph`(ruby) has limited support and quite offten returns `No results found`, so you have to keep and remember the built-in Sublime `goto_definition` key bining to be able to jump to at least something.  
It becomes especially inconvenient when you switch between different programming langages where one has a better symbol defition support than another so you always have to keep in mind what exact key to use.

This PR adds an optional `fallback` parameter to the `lsp_symbol_definition` command that allows to fallback to the built-in `goto_definition` command in case no results were found.  

```JSON
{ "keys": ["super+alt+down"], "command": "goto_definition" },
{
  "keys": ["super+alt+down"],
  "command": "lsp_symbol_definition",
  "args": { "side_by_side": false, "fallback": true },
  "context": [
    {
      "operator": "equal",
      "operand": "definitionProvider"
    },
    {
      "key": "auto_complete_visible",
      "operator": "equal",
      "operand": false
    }
  ]
},
```
This option allows you to have more consistant key bining and better integration with Sublime, because you don't need to worry if LSP is enable or no, as every time it would try to find at least something.

### DIFF
--- a/plugin/goto.py
+++ b/plugin/goto.py
@@ -65,7 +65,7 @@ class LspGotoCommand(LspTextCommand):
         else:
             self._handle_no_results(fallback, side_by_side)
 
-    def _handle_no_results(self, fallback: bool = False, side_by_side: bool = False):
+    def _handle_no_results(self, fallback: bool = False, side_by_side: bool = False) -> None:
         window = self.view.window()
 
         if not window:
--- a/plugin/goto.py
+++ b/plugin/goto.py
@@ -24,7 +24,7 @@ class LspGotoCommand(LspTextCommand):
         side_by_side: bool = False,
         fallback: bool = False
     ) -> bool:
-        return super().is_enabled(event, point)
+        return fallback or super().is_enabled(event, point)
 
     def run(
         self,
@@ -42,6 +42,8 @@ class LspGotoCommand(LspTextCommand):
             session.send_request(
                 request, functools.partial(self._handle_response_async, session, side_by_side, fallback)
             )
+        else:
+            self._handle_no_results(fallback, side_by_side)
 
     def _handle_response_async(
         self,
--- a/plugin/goto.py
+++ b/plugin/goto.py
@@ -15,8 +15,15 @@ import sublime
 class LspGotoCommand(LspTextCommand):
 
     method = ''
+    fallback_command = ''
 
-    def is_enabled(self, event: Optional[dict] = None, point: Optional[int] = None, side_by_side: bool = False) -> bool:
+    def is_enabled(
+        self,
+        event: Optional[dict] = None,
+        point: Optional[int] = None,
+        side_by_side: bool = False,
+        fallback: bool = False
+    ) -> bool:
         return super().is_enabled(event, point)
 
     def run(
@@ -24,19 +31,23 @@ class LspGotoCommand(LspTextCommand):
         _: sublime.Edit,
         event: Optional[dict] = None,
         point: Optional[int] = None,
-        side_by_side: bool = False
+        side_by_side: bool = False,
+        fallback: bool = False
     ) -> None:
         session = self.best_session(self.capability)
         position = get_position(self.view, event, point)
         if session and position is not None:
             params = text_document_position_params(self.view, position)
             request = Request(self.method, params, self.view, progress=True)
-            session.send_request(request, functools.partial(self._handle_response_async, session, side_by_side))
+            session.send_request(
+                request, functools.partial(self._handle_response_async, session, side_by_side, fallback)
+            )
 
     def _handle_response_async(
         self,
         session: Session,
         side_by_side: bool,
+        fallback: bool,
         response: Union[None, Location, List[Location], List[LocationLink]]
     ) -> None:
         if isinstance(response, dict):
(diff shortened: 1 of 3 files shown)


plugin/goto.py
```diff
@@ -15,53 +15,72 @@
...
     ) -> None:
         if isinstance(response, dict):
             self.view.run_command("add_jump_record", {"selection": [(r.a, r.b) for r in self.view.sel()]})
             open_location_async(session, response, side_by_side)
         elif isinstance(response, list):
             if len(response) == 0:
-                window = self.view.window()
-                if window:
-                    window.status_message("No results found")
+                self._handle_no_results(fallback=fallback, side_by_side=side_by_side)
```
rchl, line 58: Shouldn't be needed to use keyword arguments here? It's a bit redundant.

rchl (commented):
- Would be good to mention something about that in https://lsp.sublimetext.io/features/#goto-definition also.
- The default keybindings should be updated to include this argument (disabled by default) for visibility.

timfjord:
@rchl I've addressed your suggestions.  
Please let me know if the doc update makes sense for you, otherwise I will come up with something else

rchl (approved):

plugin/goto.py
```diff
@@ -15,53 +15,72 @@
...
         _: sublime.Edit,
         event: Optional[dict] = None,
         point: Optional[int] = None,
-        side_by_side: bool = False
+        side_by_side: bool = False,
+        fallback: bool = False
     ) -> None:
         session = self.best_session(self.capability)
         position = get_position(self.view, event, point)
         if session and position is not None:
```
timfjord, line 39: @rchl I was thininkig about one thing. What if we change this to something like
```python
if session and position is not None:
    # ...
elif position is not None:
    self._handle_no_results(True, side_by_side)
```
This way the `lsp_symbol_definition` with `"fallback": true` can be a drop-in replacement for the `goto_definition`(I mean we wouldn't need the context with the capatibility check) because if there is no capability(or no LSP at all) it would be just an equivalent for the `goto_definition`

jwortmann:
I wonder whether this shouldn't be made the default behavior? (I.e. use default value `fallback: bool = True` in the argument list.)
Then we could even consider to enable the F12 keybinding for `lsp_symbol_definition`, which is currently commented out.

Otherwise, users need to override the resource files `Context.sublime-menu` & `Main.sublime-menu` if they want to make use of this improvement. And probably not many people read the documentation or keep up to date with the newly added features.

timfjord:
If we want to make the `fallback` `true` by default, we might need to update the status with something like `No results found, falling back to the built-in Goto Definition` to let the user know what happens under the hood

plugin/goto.py
rchl: Technically it's not necessary and it's a bit safer to not use the LSP command when there are no active sessions.

rchl:
I'm not 100% sure but would lean towards enabling `fallback` and the LSP key binding for F12 by default.

Adding the status message on fallback is not a bad idea (although myself I probably wouldn't notice it anyway).

plugin/goto.py
timfjord: Hm, what if we do this only when the `fallback` is set to `true`
```python
if session and position is not None:
    # ...
else:
    self._handle_no_results(fallback, side_by_side)
```
This way, users who have `{"key": "lsp.session_with_capability", "operator": "equal", "operand": "definitionProvider"}` in their key binding notice a difference **only** if for some reasons `position is None`(they would see `No results found` in the status bar, which is perfectly fine I think).

But at the same, users who want to have one command(`lsp_symbol_definition`) everywhere can achieve that by having it with `"fallback": true` argument without any additional context(e.g. `{"keys": [<key>], "command": "lsp_symbol_definition", "args": { "fallback": true }`)

It can be useful for cases where there is no context support, like, if you want to go to the definition with CMD + mouse click. Currently there is no way to achieve this, because `*.sublime-mousemap` files don't support context, so you either use the built-in `goto_definition`(which is not always accurate) or `lsp_symbol_definition`(which is not always available)

rchl: right, mouse maps are a good argument for doing that I guess

timfjord: Ok, here we go 5a970b492ce731a6821b9bc1509649f24ba23894

Please let me know what do you think?

I tested it and it works great.

rchl (approved):

plugin/goto.py
```diff
@@ -15,53 +15,74 @@
...
+        if fallback and self.fallback_command:
+            window.run_command(self.fallback_command, {"side_by_side": side_by_side})
+        else:
+            window.status_message("No results found")
 
 
 class LspSymbolDefinitionCommand(LspGotoCommand):
     method = "textDocument/definition"
     capability = method_to_capability(method)[0]
+    fallback_command = "goto_definition"
```
rwols, line 85: I haven’t followed the conversation in detail, but why shouldn’t the other goto variants have this fallback?

timfjord: I added it only for the `LspSymbolDefinitionCommand` because other commands don't have a Sublime equivalent.
In the current implementation, all the LSP specific commands(except `LspSymbolDefinitionCommand`)  will return `No results found` which makes sense to me

PR #1991, opened 2022-07-06
Add LspWindowCommand

### LINKED ISSUE #1988: Add LspWindowCommand
It would be good to have a LspWindowCommand, which is enabled when there is any active session with a certain name/capability in this window (i.e. attached to any existing view of the window).

An example use case is to show documentation provided by the language server in a separate tab as `HtmlSheet` (rendered minihtml). It might be desired to run a request to the server, when a link within the HtmlSheet is clicked, but this is not possible via a LspTextCommand, because a HtmlSheet is not a `View`.

I have an implementation for the Julia language server, currently with an [ugly workaround](https://github.com/sublimelsp/LSP-julia/blob/eba9ea3bb9ab043dd21a422ed62a2b1e9fc30bc6/plugin.py#L34) to store a reference to the session in a global variable when a certain LspTextCommand is run. Later, this global variable is used from a normal `sublime_plugin.WindowCommand` to run the server request. This has several drawbacks:
- The WindowCommand only works after the LspTextCommand has been invoked at least one time
- By using a single global variable for the session, the WindowCommand is not actually bound to the particular window. It is also active for other, entirely unrelated windows, and also it would be problematic if there exist multiple sessions of the language server at the same time.
- There is no way to check if the session is still alive

jwortmann:
Resolves #1988.

### DIFF
--- a/plugin/core/registry.py
+++ b/plugin/core/registry.py
@@ -38,18 +38,46 @@ def get_position(view: sublime.View, event: Optional[dict] = None, point: Option
         return None
 
 
+class LspWindowCommand(sublime_plugin.WindowCommand):
+    """
+    Inherit from this class to define requests which are not bound to a particular view. This allows to run requests
+    for example from links in HtmlSheets or when an unrelated file has focus.
+    """
+
+    # When this is defined in a derived class, the command is enabled only if there exists a session with the given
+    # capability attached to a view in the window.
+    capability = ''
+
+    # When this is defined in a derived class, the command is enabled only if there exists a session with the given
+    # name attached to a view in the window.
+    session_name = ''
+
+    def is_enabled(self) -> bool:
+        return self.session() is not None
+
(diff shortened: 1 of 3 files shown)


plugin/core/registry.py
```diff
@@ -38,18 +38,46 @@ def get_position(view: sublime.View, event: Optional[dict] = None, point: Option
...
+
+    # When this is defined in a derived class, the command is enabled only if there exists a session with the given
+    # name attached to a view in the window.
+    session_name = ''
+
+    def is_enabled(self) -> bool:
+        return self.session() is not None
+
+    def session(self) -> Optional[Session]:
+        for session in windows.lookup(self.window)._sessions:
```
rchl, line 59: Would prefer adding a getter for sessions rather than accessing private property.

rchl (approved):

PR #1992, opened 2022-07-08
Allow plugins to modify server response messages

jwortmann:
This method for `AbstractPlugin` allows LSP-* plugins to modify the "result" in (successful) response messages from the server, before it gets further processed by the response handler function in LSP.

It is useful especially for problematic content in hover responses, but the implementation is generic enough to allow plugins to do any shenanigans they like with the payload ;)

### Examples:

Hover popups from LSP-pyright have some minor issues (https://github.com/sublimelsp/LSP-pyright/issues/42), for example there is occasionally only a single `\n` between the end of a fenced code block and `---` (horizontal rule). This is absolutely fine in CommonMark, but mdpopups (or the Python-Markdown parser or whatever it uses) mistakenly interprets it as a Setext header, i.e. currently the code will be bold and the horizontal rule is omitted:

before
 
Add the following lines to LSP-pyright in `plugin.py`, and it will be displayed correctly:
```python
    def on_server_response(self, method: str, result: Any) -> None:
        if method == "textDocument/hover" and result.get("contents", {}).get("kind") == "markdown":
            result["contents"]["value"] = re.sub("\n---", "\n\n---", result["contents"]["value"])
``` 

after

---

I would like to use it for the hover popups from the Julia language server, which have several similar issues in their Markdown content. In particular they often use links formatted in the form ```[`title`](@ref)```, and as soon as you click on the link, it will start a webbroser and try to open `@ref` as URL, which of course doesn't work. But with small modifications I could make these links to open a corresponding documentation page from the server in a HtmlSheet - https://github.com/sublimelsp/LSP-julia/blob/50f978826407faf0d1aaebcd3c2f8e5cb53f8136/plugin.py#L543-L562

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -65,6 +65,7 @@ from .views import SYMBOL_KINDS
 from .workspace import is_subpath_of
 from abc import ABCMeta
 from abc import abstractmethod
+from copy import deepcopy
 from weakref import WeakSet
 import functools
 import mdpopups
@@ -872,6 +873,17 @@ class AbstractPlugin(metaclass=ABCMeta):
         This API is triggered on async thread.
 
         :param    notification:  The notification object. The notification params can be modified by the plugin.
+        """
+        pass
+
+    def on_server_response(self, method: str, result: Any) -> None:
+        """
+        Notifies about a response message that has been received from the language server.
+        Only successful responses are passed to this method.
+
+        :param    method:  The method of the request.
+        :param    result:  The result to the request. It can be modified by the plugin, before it gets further handled
+                           by the LSP package.
         """
         pass
 
@@ -1898,10 +1910,11 @@ class Session(TransportCallbacks):
                 return res
         elif "id" in payload:
             response_id = int(payload["id"])
-            handler, result, is_error = self.response_handler(response_id, payload)
-            response_tuple = (handler, result, None, None, None)
-            self._logger.incoming_response(response_id, result, is_error)
-            return response_tuple
+            handler, method, result, is_error = self.response_handler(response_id, payload)
+            self._logger.incoming_response(response_id, deepcopy(result), is_error)
+            if self._plugin and not is_error:
+                self._plugin.on_server_response(method, result)  # type: ignore
+            return handler, result, None, None, None
         else:
             debug("Unknown payload type: ", payload)
         return (None, None, None, None, None)
@@ -1925,21 +1938,25 @@ class Session(TransportCallbacks):
             except Exception as err:
                 exception_log("Error handling {}".format(typestr), err)
 
-    def response_handler(self, response_id: int, response: Dict[str, Any]) -> Tuple[Optional[Callable], Any, bool]:
+    def response_handler(
+        self,
+        response_id: int,
+        response: Dict[str, Any]
+    ) -> Tuple[Optional[Callable], Optional[str], Any, bool]:
         request, handler, error_handler = self._response_handlers.pop(response_id, (None, None, None))
         if not request:
             error = {"code": ErrorCode.InvalidParams, "message": "unknown response ID {}".format(response_id)}
-            return (print_to_status_bar, error, True)
+            return (print_to_status_bar, None, error, True)
         self._invoke_views(request, "on_request_finished_async", response_id)
         if "result" in response and "error" not in response:
-            return (handler, response["result"], False)
+            return (handler, request.method, response["result"], False)
         if not error_handler:
             error_handler = print_to_status_bar
         if "result" not in response and "error" in response:
             error = response["error"]
         else:
             error = {"code": ErrorCode.InvalidParams, "message": "invalid response payload"}
-        return (error_handler, error, True)
+        return (error_handler, request.method, error, True)
 
     def _get_handler(self, method: str) -> Optional[Callable]:
         return getattr(self, method2attr(method), None)


plugin/core/sessions.py
```diff
@@ -1898,10 +1910,11 @@ def deduce_payload(
...
         elif "id" in payload:
             response_id = int(payload["id"])
-            handler, result, is_error = self.response_handler(response_id, payload)
-            response_tuple = (handler, result, None, None, None)
-            self._logger.incoming_response(response_id, result, is_error)
-            return response_tuple
+            handler, method, result, is_error = self.response_handler(response_id, payload)
+            self._logger.incoming_response(response_id, deepcopy(result), is_error)
+            if self._plugin and not is_error:
+                self._plugin.on_server_response(method, result)  # type: ignore
```
rchl, line 1916: The result could be a primitive like a string or number in which case it won't be possible to modify it when passed through an argument like that. It should be either wrapped into some object or returned through the return value.

plugin/core/sessions.py
```diff
@@ -875,6 +876,17 @@ def on_pre_send_notification_async(self, notification: Notification) -> None:
         """
         pass
 
+    def on_server_response(self, method: str, result: Any) -> None:
```
rchl, line 879: I don't have the mind for it right now but I would want to think about whether this is generic enough approach and how would it fit in the future if we'd potentially want to also hook into other things like:
- hooking server initiated notification/request
- hooking client initiated notification/request
- hooking client response to server initiated request.

plugin/core/sessions.py
```diff
@@ -1898,10 +1910,11 @@ def deduce_payload(
                 return res
         elif "id" in payload:
             response_id = int(payload["id"])
-            handler, result, is_error = self.response_handler(response_id, payload)
-            response_tuple = (handler, result, None, None, None)
-            self._logger.incoming_response(response_id, result, is_error)
-            return response_tuple
+            handler, method, result, is_error = self.response_handler(response_id, payload)
+            self._logger.incoming_response(response_id, deepcopy(result), is_error)
```
rchl, line 1914: Is it necessary to deepcopy? It can be a significant performance penalty since when thinking about typescript returning potentially megabytes of completions.

plugin/core/sessions.py
jwortmann: Ah, good point. I think it's probably better to keep the return value of this function to be `None`, because otherwise plugins need to remember to add something like `else: return result` at the bottom if they choose to implement this method.

I could wrap it in a dict `{"result": result}` and rename the argument to `response`, then plugins could modify `response["result"]`. Or I could use the Response class, then plugins could modify the `response.result` field.
I guess I'll go with the latter option, if there are no objections.

plugin/core/sessions.py
jwortmann: If the result is a dict, and you use normal `copy` and later modify `result["somekey"]`, then the copy will also see the new value. Therefore we need to use deepcopy here.

I guess the only alternative would be to not copy at all and do the logging on the main thread instead of async, but that would probably have an even bigger performance impact.

plugin/core/sessions.py
jwortmann: So do you mean to only have a single method for all different kinds, like `on_payload`? I think that would make it more complicated with more parameters like "direction", and plugins would then need to do everything what LSP already does in `Session.deduce_payload` themselves. And there are already a few methods in AbstractPlugin:

> * hooking client-initiated notification/request

This should already be possible with the `on_pre_send_*` methods.

I guess if we need more methods e.g. to modify incoming notifications before passing them on to LSP, we could add it on demand when we find a usecase for it.

plugin/core/sessions.py
jwortmann: Actually I'm not sure now. When I use `from copy import copy` and modify `deepcopy(result)` to `copy(result)` it does not work, i.e. the modified payload gets logged into the console. But when I run the following in the Python REPL, it does work:
```
>>> from copy import copy
>>> old = {"x": "foo"}
>>> copied = copy(old)
>>> old["x"] = "bar"
>>> copied
{'x': 'foo'}
```

Strange

plugin/core/sessions.py
jwortmann: Should be fixed by 16024e3e122463c22b0324de594f933aa9c856cf

jwortmann: By the way, the type annotation for the result field in both our Response class, as well as in the LSP specs is wrong. The LSP specs misses `array`, and we miss `str`, `float`, `bool`:
https://github.com/sublimelsp/LSP/blob/1289fe9f3a522deac7b554a78b73ba2b8cbe095e/plugin/core/protocol.py#L486

plugin/core/sessions.py
jwortmann: Another idea would be to gate that whole line behind a `if userprefs().log_server`, so it would only impact performance if logging is enabled. But from the code I can only see the PanelLogger to honor this setting inside its logging methods, and idk how the remote logger works. Would it break anything to not call this method at this place here, if the setting is disabled?

plugin/core/sessions.py
rchl: Spec lists an `object` which in Javascript means almost any non-literal type so not sure. Depends how it's meant to be interpreted.

plugin/core/sessions.py
rchl: I'm not sure yet. I would have to think about it and maybe prototype a bit. Would not want to end up with inconsistent and confusing API if we need to add stuff later.

Since we have `on_pre_send_request_async` and `on_pre_send_notification_async` already, we should probably either be as consistent with those as possible or deprecate those and create something more generic.

Not sure yet.

plugin/core/sessions.py
rchl: I always run with logging enabled so I wouldn't be happy with performance being affected in that case.

I guess the question can also be asked whether we have to log the value as received from the server or it's OK to log it after object being potentially modified. I guess either could cause confusion, depending on what one is looking for.

plugin/core/sessions.py
jwortmann: Hm, just below that, the specs list `data?: string | number | boolean | array | object | null;` for the ResponseError type. So the missing `array` for the type of `result` just looks like an oversight to me. I would guess the `object` refers to a JSON "object" here, i.e. what in Python would be a "dict".

In general there are many inaccuracies and inconsistencies in the LSP specs. Another strange example is that the id of a request is `id: integer | string;`, but for a response message it can also be null: `id: integer | string | null;`.
What should the client do if it receives a response with id `null`? It can't really assign it to any request without having an id...

plugin/core/sessions.py
rwols: Basically if I don't see an `from copy import deepcopy` line anymore I'll merge this :)

jwortmann: deepcopy is still there.

jwortmann: So `on_server_response_async`, or maybe `on_receive_response_async` to be closer to the existing naming scheme? ("pre" isn't really applicable here I guess)

rwols (approved):

plugin/core/sessions.py
jwortmann: It is indeed on the async thread.

But why is there a `set_timeout_async` then in
https://github.com/sublimelsp/LSP/blob/418d9931e62e63d0484279995b79843dbdfc051d/plugin/core/windows.py#L505-L517

... when all the logging methods and therefore `PanelLogger.log()` is already called from the async thread anyways? This seems to be the reason why a following modification to the payload would currently end up in the log panel too. Shouldn't the logging work as expected, (i.e. original payload), and without the need for deepcopy if we remove the `set_timeout_async` in `log()`?

rwols: It’s been a long time since I looked at that part of the code, but I think the set_timeout_async is indeed not necessary.

rchl: Yes, having `receive` would be more in line though it gets pretty wordy.

Though `on_receive_response_async` doesn't sound right as it could then be misunderstood as client sending the response. So it would have to be `on_receive_server_response_async`.

Let's see how the names could look like if we'd follow the current convention...

## Existing
 - on_pre_send_request_async
 - on_pre_send_notification_async
## This one
 - on_receive_server_response_async
## Potential
 - on_receive_client_response_async
 - on_pre_server_request_async
 - on_pre_client_request_async

(The `*_client_*` ones could still be useful IMO as one might want to let the client use the default logic and then still modify something).

I suppose this would work after all?

rchl: Well, that's not consistent either as existing APIs don't contain the `*_client_*` part...

rchl: Proposed simplification below. The current APIs could be deprecated and replaced with new ones eventually (I might be the only one using one of them anyway).

## Existing
 - on_client_notification_async
 - on_client_request_async
## This one
 - on_server_response_async
## Potential
 - on_client_response_async
 - on_server_request_async

rchl: > It’s been a long time since I looked at that part of the code, but I think the set_timeout_async is indeed not necessary.

I agree

rchl (approved):

PR #1993, opened 2022-07-09
Don't use actual linebreaks in log panel if payload is string literal

jwortmann:
Before:

before

After:

after

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -506,8 +506,7 @@ class PanelLogger(Logger):
 
         def run_on_async_worker_thread() -> None:
             nonlocal message
-            params_str = "'{}'".format(params.replace("\n", "\\n").replace("'", "\\'")) if isinstance(params, str) \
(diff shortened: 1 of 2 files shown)


rchl (approved):

plugin/core/windows.py
```diff
@@ -506,7 +506,7 @@ def log(self, message: str, params: Any) -> None:
 
         def run_on_async_worker_thread() -> None:
             nonlocal message
-            params_str = str(params)
+            params_str = repr(params) if isinstance(params, str) else str(params)
```
rchl, line 509: It seems like we could use `repr` in all cases. At least objects and lists seem to render the same with both `repr` and `str`.

jwortmann: Yeah I thought the same. The only difference between repr and str seems to be for floating point numbers, but they are very rare in LSP anyway.

rchl (approved):

Issue #886, opened 2020-02-21
Question: How to go to definition in already open tab?


RafalSkolasinski:
Each time I use `go to definition` functionality it opens a new tab.
It would be nice if it could reuse already open tab and just set the focus in the right place in similar fashion as going through search results do.

jfcherng:
Another thing that I found is the prompt menu sometimes has duplicated items

![image](https://user-images.githubusercontent.com/6594915/75041174-0540a780-54f7-11ea-81af-4cd1427c07cd.png)

RafalSkolasinski:
What I notice is that it is actually reusing the open tab but only if it is open in the same pane.
If the tab with target file is open in different pane `go to definition` will open it again in the same pane.

rwols:
The built-in "Goto Definition" from ST has the same behavior, and we pretty much copy that behavior. There's not much more we can do. See: https://github.com/sublimehq/sublime_text/issues/813

rwols:
I spoke too soon. It's actually fixable by removing a FORCE_GROUP flag!

jfcherng:
I would actually like have the same behavior (or make it configurable) since it's not obvious that which tab in which group is currently getting focused.

rwols:
Well, FORCE_GROUP is an [undocumented flag](https://github.com/sublimehq/sublime_text/issues/2290) anyway, so we shouldn't use it.

PR #1994, opened 2022-07-13
Keep active group when using Goto commands

### LINKED ISSUE #1990: Goto definition in the same file should stay in the same tab
With more than one instance of the same file open, e.g. in 2 columns mode, goto definition always jumps to the first file instance instead of the currently focused file instance.

**To Reproduce**
1. Open 2 columns to view the same file
2. Go to the right one, goto definition on something
3. Focus jumps to the left file instance

**Expected behavior**
Focus stays on the right instance

- OS: macOS
- Sublime Text version: 4126
- LSP version: 1.16.3
- Language servers used: LSP-typescript

"Find references" opens a pane whose links always jump to the currently focused file instance, so I thought goto definition doing the same would be consistent.

jwortmann:
Fixes #1990

More general, this also aligns the behavior of LSP Goto commands with the behavior of the native commands in the following, related case: if the location is in another file, which is already opened in another group, then the LSP commands would focus the already opened tab. After this PR, it will instead create a new view for the file in the active group.

There was a logic bug in the open_file function, which caused the current behavior (assuming that the `group = -1` argument should have the same meaning as in all the Sublime API functions).

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -194,7 +194,8 @@
     // {
     //     "command": "lsp_symbol_type_definition",
     //     "args": {
-    //         "side_by_side": false
+    //         "side_by_side": false,
+    //         "force_group": true
     //     },
     //     "keys": [
     //         "UNBOUND"
@@ -216,7 +217,8 @@
     // {
     //     "command": "lsp_symbol_declaration",
     //     "args": {
-    //         "side_by_side": false
+    //         "side_by_side": false,
+    //         "force_group": true
     //     },
     //     "keys": [
     //         "UNBOUND"
@@ -238,7 +240,8 @@
     // {
     //     "command": "lsp_symbol_implementation",
     //     "args": {
-    //         "side_by_side": false
+    //         "side_by_side": false,
+    //         "force_group": true
     //     },
     //     "keys": [
     //         "UNBOUND"
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -171,7 +171,8 @@
     // {
     //     "command": "lsp_symbol_definition",
     //     "args": {
-    //         "side_by_side": false
+    //         "side_by_side": false,
+    //         "force_group": true
     //     },
     //     "keys": [
     //         "f12"
--- a/plugin/core/open.py
+++ b/plugin/core/open.py
@@ -28,10 +28,10 @@ def open_file(
     # to open as a separate view).
     view = window.find_open_file(file)
     if view:
-        view_group = window.get_view_index(view)[0]
-        opens_in_desired_group = not bool(flags & sublime.FORCE_GROUP) or view_group == window.active_group()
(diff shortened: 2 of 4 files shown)


jwortmann:
Well, I see now that this behavior was changed some time ago, as a result of the issue/question #886.

So the question is, do we want to use the same behavior as native ST commands do (i.e. *with* the FORCE_GROUP flag - which *is* documented now btw, and seems to be used intentionally in Default/symbol.py), or should files be focused even if they are in different groups. Apparently, there will be issues reported from users in any of the cases, because some prefer it the one or the other way...

But I guess if the "LSP Goto..." commands are meant to be a possible replacement for the native commands, then it is probably better to follow the behavior of native ST?

rwols (approved):
lgtm, please fix the merge conflicts

jyounce-aim:
Hi guys, I recently upgraded to LSP v1.17.0.

My question is, how do I get this functionality back? https://github.com/sublimelsp/LSP/issues/886

I want to be able to `go to definition` and it open up in an existing open file regardless of what pane the file is in, how it used to be before this package upgrade.

Thank you,

Judd

rchl:
If you'll look at the changes here you'll notice that the keybindings got a new argument `force_group` that defaults to `true`. I think that defining your own keybinding with value `false` should do the trick.

jwortmann:
If you use a key binding for it, see the example but use `"force_group": false`

https://github.com/sublimelsp/LSP/blob/4b0d5fd214234013cd2055ac799b6212af1cd4b5/Default.sublime-keymap#L171-L193

For the right-click context menu you could create a new entry for it in a `Packages/User/Context.sublime-menu` file with content
```json
[
    {
        "caption": "LSP Goto Definition",
        "command": "lsp_symbol_definition",
        "args": {"force_group": false}
    },
]
```

If you use the link from the hover popup, then it is not easily possible. But I would say that LSP should follow ST's behavior, which has been like this for quite some time, and it seems to be intentional this way in ST. There is some discussion about it in https://github.com/sublimehq/sublime_text/issues/813 and in case it gets changed in the future or a setting will be introduced for it, then I guess LSP can quickly adopt that.

maegul:
Hmmm ... changing to `"force_group": false` as @jwortmann and @rchl suggest works well ... 

... except for when there are multiple definitions of the same function/symbol and a quick panel arises.  This is quite common if you're using `overloads` in python for instance.  It's not clear to me whether this is a bug or another setting I can manipulate.

jwortmann:
> ... except for when there are multiple definitions of the same function/symbol and a quick panel arises.

Looks like the "force_group" argument is not passed at
https://github.com/sublimelsp/LSP/blob/87e072cd2e9cb36888b99d1b07fdd3970eb408ca/plugin/goto.py#L76
and it needs to be added as another argument for the LocationPicker.

rchl:
Addressing in https://github.com/sublimelsp/LSP/pull/2134

PR #1997, opened 2022-07-16
Restore selections after location picker panel

### LINKED ISSUE #1977: goto commands don't restore selection when location picking is canceled
**Describe the bug**
when there's more than one location available for a `goto*` command, a quick panel is shown to pick.
highlighting entries modifies the selection, canceling the operation doesn't restore the initial selection.

**Expected behavior**
it should restore the selection, like ST's built-in

**Screenshots**
lsp_goto_cancel_bug

- OS: Windows 10
- Sublime Text version: 4126
- LSP version: 1.16.3
- Language servers used: LSP-rust-analyzer

**Additional context**
Add any other context about the problem here. For example, whether you're using a helper
package or your manual server configuration in LSP.sublime-settings. When using
a manual server configuration please include it here if you believe it's applicable.

jwortmann:
Fixes #1977

### DIFF
--- a/plugin/locationpicker.py
+++ b/plugin/locationpicker.py
@@ -53,6 +53,7 @@ class LocationPicker:
         side_by_side: bool
     ) -> None:
         self._view = view
+        self._view_states = ([r.to_tuple() for r in view.sel()], view.viewport_position())
         window = view.window()
         if not window:
             raise ValueError("missing window")
@@ -76,6 +77,9 @@ class LocationPicker:
         return self._weaksession(), location, uri, position
 
     def _select_entry(self, index: int) -> None:
+        if self._view.is_valid() and not self._side_by_side:
+            self._view.set_viewport_position(self._view_states[1])
+            self._view.run_command('lsp_selection_set', {'regions': self._view_states[0]})
         if index >= 0 and self._view.is_valid():
             session, location, uri, position = self._unpack(index)
             if not session:


plugin/locationpicker.py
```diff
@@ -76,6 +77,9 @@ def _unpack(self, index: int) -> Tuple[Optional[Session], Union[Location, Locati
         return self._weaksession(), location, uri, position
 
     def _select_entry(self, index: int) -> None:
+        if self._view.is_valid() and not self._side_by_side:
+            self._view.set_viewport_position(self._view_states[1])
+            self._view.run_command('lsp_selection_set', {'regions': self._view_states[0]})
```
rchl, line 82: Would it not make sense to do it only when index is `-1`? Otherwise it seems wasteful to run it if we're gonna override it below anyway. And it could cause some visual issues in theory (although I don't see any when testing here).

jwortmann: I had tested this yesterday and the problem also exists in some cases when you select a valid entry. It happens when there is an entry with a location from the current file, and another entry with a location in different file. If you first highlight the location from the same file, it will scroll the viewport and update the selection, and if you then select the location in the other file, it will be opened in a new tab. But then the original file is still scrolled to the symbol position and the original selection is lost.

rwols (approved):

PR #2000, opened 2022-07-20
Refactor symbol and completion item kinds

jwortmann:
- This defines [sublime.Kind](https://www.sublimetext.com/docs/api_reference.html#sublime.Kind) tuples, so that they can be reused for LSP's `SymbolKind`s and `CompletionItemKind`s, which have a lot of overlap and even had a few inconsistencies before.
- Also they could easily be used for sublime.QuickPanelItem now.
- No more awkward indexing with `kind - 1`
- Moved the small lookup dicts for DocumentHighlightKind to core/views.py as well, because that seems to be the place where most of such things are defined.
- Fixes a bug in views.py#format_completion, which would use an `int` instead of a kind tuple for the kind in `sublime.CompletionItem.command_completion`, in case there is a kind id in the language server response, which is out of range for LSP's CompletionItemKind.

Changes:
- Removed "readwrite" from "variable.other.readwrite" scope of SymbolKind.Variable (dubious scope only used in JavaScript syntax - `https://github.com/sublimehq/Packages/issues/1336`).
- Adjusted scope of SymbolKind.Boolean from "constant.language" to "constant.language.boolean" (used in various syntaxes).
- Adjusted scope of SymbolKind.Null from "constant.language" to "constant.language.null" (used in various syntaxes).
- Changed scope of SymbolKind.Event from "storage.modifier" to "entity.name.function" and kind id from sublime.KIND_ID_TYPE to sublime.KIND_ID_FUNCTION (I'm not entirely sure, but I think "Event" is meant to be things like `onclick` or `onchange`).
- Changed scope of SymbolKind.TypeParameter from "storage.type" to "variable.parameter.type" (but keep the kind id as sublime.KIND_ID_TYPE). Ass far as I understand, the following is a type parameter (Java example):
    ```java
    class ClassWithGenericType<T> {}
    //                         ^ Type parameter
    ```
- Changed kind id of CompletionItemKind.Folder from sublime.KIND_ID_NAMESPACE to sublime.KIND_ID_NAVIGATION, so that it is consistent with CompletionItemKind.File
- Changed kind id of SymbolKind.Interface from sublime.KIND_ID_VARIABLE to sublime.KIND_ID_TYPE, so that it is consistent with CompletionItemKind.Interface.
- Changed kind id of SymbolKind.Operator from sublime.KIND_ID_FUNCTION to sublime.KIND_ID_KEYWORD, so that it is consistent with CompletionItemKind.Operator.
- Changed kind id of SymbolKind.String from sublime.KIND_ID_MARKUP to sublime.KIND_ID_VARIABLE, so that it is consistent with SymbolKind.Number, SymbolKind.Boolean, SymbolKind.Null.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -186,7 +186,7 @@ class Manager(metaclass=ABCMeta):
         pass
 
 
-def _enum_like_class_to_list(c: Type[object]) -> Union[List[int], List[str]]:
+def _enum_like_class_to_list(c: Type[object]) -> List[Union[int, str]]:
     return [v for k, v in c.__dict__.items() if not k.startswith('_')]
 
 
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -186,7 +186,7 @@ class Manager(metaclass=ABCMeta):
         pass
 
 
-def _enum_like_class_to_list(c: Type[object]) -> List[Union[int, str]]:
+def _enum_like_class_to_list(c: Type[object]) -> Union[List[int], List[str]]:
     return [v for k, v in c.__dict__.items() if not k.startswith('_')]
 
 
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -186,7 +186,7 @@ class Manager(metaclass=ABCMeta):
         pass
 
 
-def _enum_like_class_to_list(c: Type[object]) -> List[Union[int, str]]:
+def _enum_like_class_to_list(c: Type[object]) -> Union[List[int], List[str]]:
     return [v for k, v in c.__dict__.items() if not k.startswith('_')]
 
 
@@ -197,7 +197,7 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
     diagnostic_tag_value_set = _enum_like_class_to_list(DiagnosticTag)
     completion_tag_value_set = _enum_like_class_to_list(CompletionItemTag)
     symbol_tag_value_set = _enum_like_class_to_list(SymbolTag)
-    semantic_token_types = _enum_like_class_to_list(SemanticTokenTypes)
+    semantic_token_types = _enum_like_class_to_list(SemanticTokenTypes)  # type: List[str]
     if config.semantic_tokens is not None:
         for token_type in config.semantic_tokens.keys():
             if token_type not in semantic_token_types:
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -186,7 +186,7 @@ class Manager(metaclass=ABCMeta):
         pass
 
 
-def _enum_like_class_to_list(c: Type[object]) -> Union[List[int], List[str]]:
+def _enum_like_class_to_list(c: Type[object]) -> List[Union[int, str]]:
     return [v for k, v in c.__dict__.items() if not k.startswith('_')]
 
 
@@ -197,7 +197,7 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
     diagnostic_tag_value_set = _enum_like_class_to_list(DiagnosticTag)
     completion_tag_value_set = _enum_like_class_to_list(CompletionItemTag)
     symbol_tag_value_set = _enum_like_class_to_list(SymbolTag)
-    semantic_token_types = _enum_like_class_to_list(SemanticTokenTypes)  # type: List[str]
+    semantic_token_types = _enum_like_class_to_list(SemanticTokenTypes)
     if config.semantic_tokens is not None:
         for token_type in config.semantic_tokens.keys():
             if token_type not in semantic_token_types:
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -19,6 +19,7 @@ from .promise import PackagedTask
 from .promise import Promise
 from .protocol import CodeAction, CodeLens, InsertTextMode, Location, LocationLink
 from .protocol import Command
+from .protocol import CompletionItemKind
 from .protocol import CompletionItemTag
 from .protocol import Diagnostic
 from .protocol import DiagnosticSeverity
@@ -36,6 +37,7 @@ from .protocol import Request
 from .protocol import Response
 from .protocol import SemanticTokenModifiers
 from .protocol import SemanticTokenTypes
+from .protocol import SymbolKind
 from .protocol import SymbolTag
 from .protocol import WorkspaceFolder
 from .settings import client_configs
@@ -55,13 +57,11 @@ from .typing import Callable, cast, Dict, Any, Optional, List, Tuple, Generator,
 from .url import filename_to_uri
 from .url import parse_uri
 from .version import __version__
-from .views import COMPLETION_KINDS
 from .views import extract_variables
 from .views import get_storage_path
 from .views import get_uri_and_range_from_location
 from .views import MarkdownLangMap
 from .views import SEMANTIC_TOKENS_MAP
-from .views import SYMBOL_KINDS
 from .workspace import is_subpath_of
 from abc import ABCMeta
 from abc import abstractmethod
@@ -186,18 +186,14 @@ class Manager(metaclass=ABCMeta):
         pass
 
 
-def _sequence_to_lsp_indices(sequence: Sequence[Any]) -> List[int]:
-    return list(range(1, len(sequence) + 1))
-
-
-def _enum_like_class_to_list(c: Type[object]) -> List[str]:
+def _enum_like_class_to_list(c: Type[object]) -> List[Union[int, str]]:
     return [v for k, v in c.__dict__.items() if not k.startswith('_')]
 
 
 def get_initialize_params(variables: Dict[str, str], workspace_folders: List[WorkspaceFolder],
                           config: ClientConfig) -> dict:
-    completion_kinds = _sequence_to_lsp_indices(COMPLETION_KINDS)
-    symbol_kinds = _sequence_to_lsp_indices(SYMBOL_KINDS)
+    completion_kinds = _enum_like_class_to_list(CompletionItemKind)
+    symbol_kinds = _enum_like_class_to_list(SymbolKind)
     diagnostic_tag_value_set = _enum_like_class_to_list(DiagnosticTag)
     completion_tag_value_set = _enum_like_class_to_list(CompletionItemTag)
     symbol_tag_value_set = _enum_like_class_to_list(SymbolTag)
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -1,9 +1,11 @@
 from .css import css as lsp_css
 from .protocol import CompletionItem
+from .protocol import CompletionItemKind
 from .protocol import CompletionItemTag
 from .protocol import Diagnostic
 from .protocol import DiagnosticRelatedInformation
 from .protocol import DiagnosticSeverity
+from .protocol import DocumentHighlightKind
 from .protocol import DocumentUri
 from .protocol import ExperimentalTextDocumentRangeParams
 from .protocol import Location
@@ -16,6 +18,7 @@ from .protocol import Position
 from .protocol import Range
 from .protocol import RangeLsp
 from .protocol import Request
+from .protocol import SymbolKind
 from .protocol import TextDocumentIdentifier
 from .protocol import TextDocumentPositionParams
 from .settings import userprefs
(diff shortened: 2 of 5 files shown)


plugin/core/sessions.py
```diff
@@ -186,18 +186,14 @@ def on_post_exit_async(self, session: 'Session', exit_code: int, exception: Opti
         pass
 
 
-def _sequence_to_lsp_indices(sequence: Sequence[Any]) -> List[int]:
-    return list(range(1, len(sequence) + 1))
-
-
-def _enum_like_class_to_list(c: Type[object]) -> List[str]:
+def _enum_like_class_to_list(c: Type[object]) -> List[Union[int, str]]:
```
rwols, line 189: The return type should be `Union[List[str], List[int]]`, or am I misunderstanding the use case?

plugin/core/views.py
```diff
@@ -47,65 +50,140 @@
...
+KIND_SNIPPET = (sublime.KIND_ID_SNIPPET, "s", "Snippet")
+KIND_STRING = (sublime.KIND_ID_VARIABLE, "s", "String")
+KIND_STRUCT = (sublime.KIND_ID_TYPE, "s", "Struct")
+KIND_TEXT = (sublime.KIND_ID_MARKUP, "t", "Text")
+KIND_TYPEPARAMETER = (sublime.KIND_ID_TYPE, "t", "Type Parameter")
+KIND_UNIT = (sublime.KIND_ID_VARIABLE, "u", "Unit")
+KIND_VALUE = (sublime.KIND_ID_VARIABLE, "v", "Value")
+KIND_VARIABLE = (sublime.KIND_ID_VARIABLE, "v", "Variable")
+
+COMPLETION_KINDS = {
```
rwols, line 90: So the reason for using a `List` was performance :) a list item lookup is an offset into a contiguous array. But, I guess the performance doesn't matter much. Maybe the dict is even faster, who knows? From what I understand dicts are implemented as hash maps. So a number is run through a hash function before doing an array offset index.

plugin/core/views.py
```diff
@@ -874,10 +952,8 @@ def format_completion(
 ) -> sublime.CompletionItem:
     # This is a hot function. Don't do heavy computations or IO in this function.
     item_kind = item.get("kind")
-    if isinstance(item_kind, int) and 1 <= item_kind <= len(COMPLETION_KINDS):
-        kind = COMPLETION_KINDS[item_kind - 1]
-    else:
-        kind = sublime.KIND_AMBIGUOUS
+    kind = COMPLETION_KINDS[item_kind] if isinstance(item_kind, int) and item_kind in COMPLETION_KINDS \
```
rwols, line 955: Would it be possible to avoid the double dict lookup for the key here? We do a lookup for the `item_kind` key in `item_kind in COMPLETION_KINDS` as test, and then do the same lookup again in `COMLETION_KINDS[item_kind]`. Perhaps use something like `kind = COMPLETION_KINDS.get(item_kind)` and test for a `None` value, which is just **1** lookup in the dict.

plugin/symbols.py
```diff
@@ -15,29 +16,23 @@
 SUPPRESS_INPUT_SETTING_KEY = 'lsp_suppress_input'
 
 
-def unpack_lsp_kind(kind: int) -> Tuple[int, str, str, str]:
-    if 1 <= kind <= len(SYMBOL_KINDS):
-        return SYMBOL_KINDS[kind - 1]
-    return sublime.KIND_ID_AMBIGUOUS, "?", "???", "comment"
+def unpack_lsp_kind(kind: int) -> Tuple[int, str, str]:
+    return SYMBOL_KINDS[kind] if kind in SYMBOL_KINDS else (sublime.KIND_ID_AMBIGUOUS, "?", "???")
```
rwols, line 20: Would it be possible to avoid the double dict lookup for the key here? We do a lookup for the `item_kind` key in `kind in SYMBOL_KINDS` as test, and then do the same lookup again in `SYMBOL_KINDS[kind]`. Perhaps use something like `st_kind = SYMBOL_KINDS.get(kind)` and test for a `None` value, which is just **1** lookup in the dict.

plugin/symbols.py
```diff
@@ -15,29 +16,23 @@
...
 
 
 def format_symbol_kind(kind: int) -> str:
-    if 1 <= kind <= len(SYMBOL_KINDS):
-        return SYMBOL_KINDS[kind - 1][2]
-    return str(kind)
+    return SYMBOL_KINDS[kind][2] if kind in SYMBOL_KINDS else str(kind)
```
rwols, line 24: Same issue with the two dictionary lookups

plugin/symbols.py
```diff
@@ -15,29 +16,23 @@
...
 
 
 def get_symbol_scope_from_lsp_kind(kind: int) -> str:
-    if 1 <= kind <= len(SYMBOL_KINDS):
-        return SYMBOL_KINDS[kind - 1][3]
-    return 'comment'
+    return SYMBOL_KIND_SCOPES[kind] if kind in SYMBOL_KIND_SCOPES else "comment"
```
rwols, line 28: Same issue with the two dictionary lookups

rwols (changes requested):
Nice work. I do think we can optimize to use 1 dict lookup everywhere instead of 2.

plugin/core/sessions.py
jwortmann: Apparently that isn't possible, because mypy can't infer which one of the list types the return value is, when it is used later. See https://github.com/sublimelsp/LSP/runs/7490372297

jwortmann: Well, I guess I can add a type hint in this case.

rwols: You can run mypy locally by running `tox` by the way

jwortmann: Hm, seems that Pyright and mypy disagree what to do with the type hint comment in https://github.com/sublimelsp/LSP/pull/2000/commits/f95c500127fc5580cf9128082cce9a782371973b.

So I guess I could revert to the initial return type `List[Union[int, str]]`, or use `# type: ignore` for the failing line. Or do you see a better solution?

Thanks I will try to figure out how to run mypy locally.

rwols: Well, in general functions that return a Union aren’t that nice. I think we can satisfy both mypy and pyright by doing a `isinstance` check but that is clumsy. We already know what we’ll pass into the function, so it seems that we need another function that basically does the same thing but returns a List[str] (the other one returning a List[int]). I think the underlying issue is that we don’t differentiate between “string enums” and “int enums”.

rchl: We should help the type system by being explicit about what the types of those class dicts are. Below I've done some initial work for the `CompletionItemKind`. If you agree on that solution then other classes could be changed the same way.

```diff
diff --git a/plugin/core/protocol.py b/plugin/core/protocol.py
index ce2f0c7..1da857b 100644
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -24,6 +24,10 @@ class DiagnosticTag:
 
 
 class CompletionItemKind:
+    @staticmethod
+    def items() -> Dict[str, int]:
+        return CompletionItemKind.__dict__
+
     Text = 1
     Method = 2
     Function = 3
diff --git a/plugin/core/sessions.py b/plugin/core/sessions.py
index 47d407e..865dfe6 100644
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -53,7 +53,7 @@ from .types import DocumentSelector
 from .types import method_to_capability
 from .types import SettingsRegistration
 from .types import sublime_pattern_to_glob
-from .typing import Callable, cast, Dict, Any, Optional, List, Tuple, Generator, Iterable, Type, Protocol, Sequence, Mapping, Union  # noqa: E501
+from .typing import Callable, cast, Dict, Any, Optional, List, Tuple, Generator, Iterable, Type, Protocol, Sequence, TypeVar, Mapping, Union  # noqa: E501
 from .url import filename_to_uri
 from .url import parse_uri
 from .version import __version__
@@ -186,23 +186,26 @@ class Manager(metaclass=ABCMeta):
         pass
 
 
-def _enum_like_class_to_list(c: Type[object]) -> Union[List[int], List[str]]:
-    return [v for k, v in c.__dict__.items() if not k.startswith('_')]
+T = TypeVar("T")
+
+
+def dict_to_list(d: Dict[str, T]) -> List[T]:
+    return [v for k, v in d.items() if not k.startswith('_')]
 
 
 def get_initialize_params(variables: Dict[str, str], workspace_folders: List[WorkspaceFolder],
                           config: ClientConfig) -> dict:
-    completion_kinds = _enum_like_class_to_list(CompletionItemKind)
-    symbol_kinds = _enum_like_class_to_list(SymbolKind)
-    diagnostic_tag_value_set = _enum_like_class_to_list(DiagnosticTag)
-    completion_tag_value_set = _enum_like_class_to_list(CompletionItemTag)
-    symbol_tag_value_set = _enum_like_class_to_list(SymbolTag)
-    semantic_token_types = _enum_like_class_to_list(SemanticTokenTypes)  # type: List[str]
+    completion_kinds = dict_to_list(CompletionItemKind.items())
+    symbol_kinds = dict_to_list(SymbolKind)
+    diagnostic_tag_value_set = dict_to_list(DiagnosticTag)
+    completion_tag_value_set = dict_to_list(CompletionItemTag)
+    symbol_tag_value_set = dict_to_list(SymbolTag)
+    semantic_token_types = dict_to_list(SemanticTokenTypes)
     if config.semantic_tokens is not None:
         for token_type in config.semantic_tokens.keys():
             if token_type not in semantic_token_types:
                 semantic_token_types.append(token_type)
-    semantic_token_modifiers = _enum_like_class_to_list(SemanticTokenModifiers)
+    semantic_token_modifiers = dict_to_list(SemanticTokenModifiers)
     first_folder = workspace_folders[0] if workspace_folders else None
     general_capabilities = {
         # https://microsoft.github.io/language-server-protocol/specification#regExp
```

rchl: `dict_to_list` -> `dict_values_to_list` would be more precise.

jwortmann: But I wonder if it is reasonable to duplicate this function with the same function body just to satisfy mypy and have slightly more precise return type annotations. This function is used only a handful of times at this spot for the initialize options. I would suggest to revert to `List[Union[int, str]]`, okay?

rchl: And to DRY it more:

```diff
diff --git a/plugin/core/protocol.py b/plugin/core/protocol.py
index ce2f0c7..58e8514 100644
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -1,4 +1,4 @@
-from .typing import Any, Dict, Iterable, List, Mapping, Optional, TypedDict, Union
+from .typing import Any, Dict, Iterable, Generic, List, Mapping, Optional, TypedDict, TypeVar, Union
 from .url import filename_to_uri
 import os
 import sublime
@@ -10,6 +10,14 @@ TextDocumentSyncKindNone = 0
 TextDocumentSyncKindFull = 1
 TextDocumentSyncKindIncremental = 2
 
+T = TypeVar("T")
+
+
+class EnumLike(Generic[T]):
+    @staticmethod
+    def items() -> Dict[str, T]:
+        return CompletionItemKind.__dict__
+
 
 class DiagnosticSeverity:
     Error = 1
@@ -23,7 +31,7 @@ class DiagnosticTag:
     Deprecated = 2
 
 
-class CompletionItemKind:
+class CompletionItemKind(EnumLike[int]):
     Text = 1
     Method = 2
     Function = 3
```

rwols: Yeah, eventually it goes into some json anyway. For now it’s okay to use List[Union] :)

plugin/core/views.py
jwortmann: I see. As an alternative to the dict, I could use lists and add the fallback `KIND_UNSPECIFIED = (sublime.KIND_ID_AMBIGUOUS, "?", "???")` as a first entry in the list, then indexing with `kind - 1` could also be avoided. But I guess the performance difference (if any) indeed shouldn't matter here and is orders of magnitude smaller than the display refresh time. So it probably doesn't have any impact and would only be a matter of style/preference. And for the lists, there was an additional check `if 1 <= kind <= len(SYMBOL_KINDS)`, which might be even more overhead...

plugin/core/sessions.py
rchl: Hmm, might not work with inheritance...

jwortmann: Let's postpone this to another PR and use List[Union] for now ;-)

Alternatively, there is also the enum module / [IntEnum](https://docs.python.org/release/3.4.0/library/enum.html?highlight=intenum#enum.IntEnum) which could be used via https://github.com/packagecontrol/enum dependency.

rwols (approved):

plugin/core/sessions.py
rwols: `EnumLike(Generic[T])` looks sweet, but like jwortmann I imagined something like https://docs.python.org/3/library/enum.html. Maybe postpone that until python 3.8 because I don't like more dependencies. Or use the `EnumLike(Generic[T])` approach.

rchl: That more concise solution using `EnumLike` didn't really work as it wouldn't be possible to access class method without instantiating an object. And `@staticmethod` doesn't have access to `cls`. So it would have to be either the more repetitive solution or some other smart approach.

Issue #1924, opened 2021-12-28
detailedLabelSupport not correctly implemented


ryuukk:
**Describe the bug**

A screenshot worth more than words:

actual:

![image](https://user-images.githubusercontent.com/44361234/147585153-984f9902-b626-4cf8-a34f-44ae08a48919.png)

expected: (vscode)

![image](https://user-images.githubusercontent.com/44361234/147585217-62356cd6-53b1-4964-afaf-1c25bbad6a4e.png)

**To Reproduce**

Use D's LSP according to the documentation

https://github.com/Pure-D/serve-d/

https://lsp.sublimetext.io/language_servers/#d

**Expected behavior**

with detailedLabelSupport, i expect each completion item to properly strip the funciton parameters, so it doesn't show the item label twice.. 

**Screenshots**

please refer to the 1st block above

**Environment (please complete the following information):**
- OS: arch linux
- Sublime Text version: 4126
- LSP version: 1.14
- Language servers used: D

jfcherng:
There is no feasible solution.

![image](https://user-images.githubusercontent.com/6594915/147589889-f5369237-ab8b-4538-8929-4b03272e5360.png)

related discuss:
- https://discord.com/channels/280102180189634562/645268178397560865/893676807616884789

ryuukk:
Oh i see, is it being looked at by the sublime text developpers?

predragnikolic:
I am looking at the image for the expect behavior 
and IMO displaying that much info in the autocomplete panel is distracting.

For me the LSP signature popup is used to show the parameters for a given function. 

Is it helpful to see a pretty wide autocomplete popup with a bunch of information?
And if so please emphasize what information is really important.

ryuukk:
@predragnikolic 

showing all the line isn't the point of the feature, let's ignore it for the moment

it's the fact that no information is duplicated anymore

# before:

``create_texture --------------------------- create_texture(uint....)``

why is ``create_texture`` displayed twice?

# after:

``create_texture (uint...) --------------------------- Texture2D``


There is less confusing, and the duplication of information is no more, less distraction

--

Then for showing all the lines at once, it could be an option ``"show_details_for_all_items":  "true|false"``

jwortmann:
1. Are you sure that the D server has the `completionProvider.completionItem.labelDetailsSupport` capability?
[...]

PR #2002, opened 2022-07-24
Set completion trigger from label detail if present

rchl:


plugin/core/views.py
```diff
@@ -885,40 +885,45 @@ def format_completion(
...
-        # `label` should be rendered most prominent.
-        st_details += _wrap_in_tags("b", lsp_label)
+        label_detail = lsp_label
         if isinstance(lsp_label_detail, str):
             # `detail` should be rendered less prominent than `label`.
             # The string is appended directly after `label`, with no additional white space applied.
-            st_details += html.escape(lsp_label_detail)
+            label_detail += lsp_label_detail
+        # annotations.append(label_detail)
+        lsp_label = label_detail
```
predragnikolic, line 901: Can we avoid having this?
```
label_detail = lsp_label
...
lsp_label = label_detail
```

rchl: It got a bit messy after many back-and-forth. Cleaned up now.

predragnikolic (approved):
👍

predragnikolic:
Something I've noticed while testing this.
Tried it with TS 4.7.2 and the lates branch from theia ts ls.
Screenshot 2022-07-25 at 06 22 08

Do you maybe know why I got 2 completions items for `bar`?

-----

I wanted to check if I create function overloads, that I might see more completion items,
but that was not the case.

Screenshot 2022-07-25 at 06 42 32

rchl:
I believe it provides two completions on purpose since depending on user's code, one might be preferred over another . I suppose it works the same in vs code.

Also it seems to be a pretty new issue so maybe it isn't supported in all possible cases tsserver. Just guessing.

jwortmann:
I'm not really convinced that this would be a good solution. There is already the CompletionItem.filterText (or CompletionItem.label if filterText doesn't exist) to be used for the "trigger" field of ST autocompletion items. I think the example with `bar(x)` is insufficient, because usually function signatures contain multiple parameters with longer names & types. So in practice the typed text will randomly fuzzy match the characters in the label details more often, so that filtering doesn't work as expected. Due to the lack of language servers with support for label details, I haven't tested it though, but I would expect bug reports for it when more servers start to support it. And it just feels wrong to me to misuse the "trigger" field for information which are not meant for filtering. Instead, we should adopt to the abilities of the ST autocompletion popup (either "annotation" or "details"), or create a feature request for it in the ST issue tracker, and as long as it isn't implemented use one of those existing fields.

And what if next time someone creates an issue like
>**deprecated CompletionItems not correctly implemented**
>
>CompletionItems with the "deprecated" property or with CompletionItemTag.Deprecated should be rendered using strike-through in the autocompletion popup. ST doesn't have the ability to do that, and as a workaround LSP currently uses a "!" as the icon letter and an additional mark in the "details" panel at the bottom for deprecated items. But instead, we can just add a dash after every other character in the "trigger" field, which would look a little bit like strike-through!
>
>deprecated
>
> That would have even less to none impact for filtering, because usually completion items don't contain dashes.


Another solution to have the label details in the "trigger" field for a particular server, would be to use the new AbstractPlugin.on_server_response_async method to just replace the CompletionItem.filterText with label + label details for each completion item in the server response.

rchl:
I'm not convinced yet that this is proper solution (or that it's better than the old one) but if your main argument is that it affects filtering then it seems like it also does affect filtering in VSCode so maybe it's not that bad after all.

Screenshot 2022-07-25 at 19 58 11

I'd need to see some specific case where it actually is noticeably worse before I'm convinced.

rchl:
For the record here is another case. Two completions:

```js
    {
      "commitCharacters": [
        ".",
        ",",
        "("
      ],
      "data": {
        "entryNames": [
          "banner"
        ],
        "file": "/usr/local/workspace/github/typescript-language-server/rollup.config.ts",
        "line": 13,
        "offset": 16
      },
      "filterText": "banner",
      "insertText": "banner",
      "kind": 5,
      "label": "banner?",
      "sortText": "12\u0000banner\u0000"
    },
    {
      "commitCharacters": [
        ".",
        ",",
        "("
      ],
      "data": {
        "entryNames": [
          {
            "name": "banner",
            "source": "ObjectLiteralMethodSnippet/"
          }
        ],
        "file": "/usr/local/workspace/github/typescript-language-server/rollup.config.ts",
        "line": 13,
        "offset": 16
      },
      "filterText": "banner",
      "insertText": "banner() {\n    $0\n},",
      "insertTextFormat": 2,
      "kind": 5,
      "label": "banner?",
      "labelDetails": {
        "detail": "()"
      },
      "sortText": "12\u0000banner\u00001"
    },
```

VSCode shows them as:
 - `banner?`
 - `banner?()`

ST as:
 - `banner`
 - `banner`
with the current code because it prefers `filterText` over `label+label.details`. We could use the latter but that would create a bit of inconsistency as we'd show `?` in the second case but not the first one.
(Note that having `?` in the trigger should not affect filtering as completion popup is canceled on pressing `?` anyway but it's just this specific case that uses `?`, other cases might use characters that can actually use other characters)

rchl:
> Yeah, that is my main point and it surprises me that VSCode uses the label details for matching/filtering.

I believe there might be something else going on. I think it actually uses a separate `filterText` completion property for matching. In the `bar(x)` case that is set to "bar(x) {\n    $0\n}," even.
(actually you've mentioned that later also ;)).

> At least I'm not the only one who is skeptical: [#1924 (comment)](https://github.com/sublimelsp/LSP/issues/1924#issuecomment-1194484673)

But that's two ST users who haven't even properly tried in practice for a decent amount of time so I'm not convinced yet. :)

> Hm.. maybe LSP should entirely ignore `filterText`, and always prefer `label` (possibly + `label.details` depending on the outcome of this PR). `filterText` seems to be used as an additional internal filter mechanism in VSCode, but isn't shown to the user in the completion item, as we do here.

I suppose that we would have to see the effects of such change in practice while testing various servers.

> I guess that if LSP would show the full `param1=` in the "trigger", this situation would be improved.

👍 

> As a Sidenote, I believe you can/should replace the `_wrap_in_tags("p", lsp_label)` by `html.escape(lsp_label)`, because the `p` tag isn't supported in the completion item details anyway and will be ignored: [sublimetext.com/docs/minihtml.html#inline-formatting](https://www.sublimetext.com/docs/minihtml.html#inline-formatting)

👍 👍

jwortmann:
How about the following proposal:

1. If there are no `label.details`, we always use `label` as the "trigger" (i.e. `filterText` will be ignored)
2. If there are `label.details`, we use `label + label.details` as the "trigger", *unless `filterText` exists and `filterText.includes(label + label.details)` returns `False` (in which case we would only use `label`)*

rchl:
Maybe I've lost some context but can you explain the reasoning for nr. 2?

jwortmann:
> Maybe I've lost some context but can you explain the reasoning for the `filterText` part of nr. 2?

I think that would be the closest to how VSCode handles it, and it would be a compromise between "show detailed information in the trigger" and "use sensible matching algorithm". VSCode indeed doesn't seem to use the label details for matching, but the `bar(x)` completion will still be shown when you only type `x` because it is included in the "filterText". So in ST, we would show the full `bar(x)` from the TypeScript server in the "trigger", and it would also work for the D server according to https://github.com/sublimelsp/LSP/issues/1924#issuecomment-1002730265, because that server doesn't use the "filterText".

In general a server would use "filterText", if it wants to give additional information how a particular completion item should be matched (otherwise it could just omit it and according to the specs the client can fallback to "label"). So if we know that the `label+label.details` is also included in the "filterText", then we know that it is intentional that the matching algorithm should also consider the `label.details` for matching, and we can safely put it in the "trigger". And the other way, if the server wants to restrict matching by using "filterText" and not including the label details therein, then we know that ST shouldn't use it for matching (then we should probably put it into the "details" instead, like it is the case now). 

I'm a bit doubtful that it is a good way how the TypeScript server uses "filterText", by including the `\n` and tabstop, but well, it's up to them if they like to.


A more conservative (in regard to the current behavior) alternative could also be:

2\.  If there are `label.details` we use `label + label.details` as the "trigger", but only if also `filterText` exists and `label+label.details in filterText` returns `True` (and otherwise use only `label`). 

But this one wouldn't show the label details from the D server in the "trigger".

rwols:
For more context on this matter: see https://github.com/sublimelsp/LSP/pull/990#issuecomment-620114750

Also see: https://github.com/microsoft/language-server-protocol/issues/898#issuecomment-593968008

My takeaway from this is that Sublime is fundamentally different from VSCode in that it shows what the user is matching, while VSCode hides what the user is matching.

We can try using LSP-label as ST-trigger if the LSP-label starts with LSP-filterText.

I don't see how we can "dim" the LSP-label-details though.

jwortmann:
> We can try using LSP-label as ST-trigger if the LSP-label starts with LSP-filterText.

Hm, okay that could work. But then I think we could do the same with the LSP-label-details = `item["labelDetails"]["detail"]`, i.e. if LSP-label + LSP-label-details starts with LSP-filterText, then we can use LSP-label + LSP-label-details as the "trigger".

I've made some tests with it and how I would implement it is at https://github.com/jwortmann/LSP/commit/e8f6928b25ea13fb41eecba2dee035beffa36237.

In the mentioned case of LSP-label + LSP-label-details as "trigger", it will also use the LSP-label-description (if present) as the "annotation". This should yield better results with the D server https://github.com/sublimelsp/LSP/issues/1924#issuecomment-1002730265 and follows the specs in regard that it should be "rendered less prominently after" LSP-label-details.

It's quite annoying that various servers seem to use some fields differently or in unexpected ways. For example the Julia server doesn't use the labelDetails, but puts the whole function body into the regular `item["detail"]` for a function completion:

lsp

rchl:
I do like how it works (after testing for 5 minutes, mind you).

rchl:
@jwortmann do you want to create a PR with your changes?

Alternatively I could adopt those here.

PR #2003, opened 2022-07-24
Prevent failing to load all configs when one fails

rwols:
Some users may accidentally have `"lsp_format_on_save": true` in their .sublime-project like this:

```js
{
  "settings": {
    "LSP": {
      "lsp_format_on_save": true
    }
  }
}
```

Of course LSP-json flags this as invalid, but not everyone has LSP-json installed. Furthermore this would prevent loading any configuration, so LSP wouldn't run any language server.


predragnikolic (approved):
👍

PR #2006, opened 2022-07-27
Tweaks for signature help popup

jwortmann:
- Add contextSupport (see https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#signatureHelpContext). For now, "triggerKind" will always be SignatureHelpTriggerKind.Invoked in case of invoking via key binding, or otherwise SignatureHelpTriggerKind.TriggerCharacter. Maybe some more logic could be added in the future, to determine if the last pressed key was e.g. <kbd>Backspace</kbd>, for the third possible option of SignatureHelpTriggerKind.
- Add support for `activeParameter` property on `SignatureInformation` (since LSP 3.16 - see https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#signatureInformation)
- Allow color schemes to adjust color of currently highlighted parameter
- Small tweaks regarding code style

Also I believe ST has a bug with the `sublime.HIDE_ON_MOUSE_MOVE_AWAY` flag for popups, which would randomly hide the popup even when the cursor isn't moved (but only if the window is relatively small [sic!] :D - it doesn't happen for me in fullscreen or with a big window). This bug seems to happen more often on this branch, and I think HIDE_ON_MOUSE_MOVE_AWAY isn't useful anyway for the signature help popup, so I removed it.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -254,9 +254,9 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
         },
         "signatureHelp": {
             "dynamicRegistration": True,
-            "activeParameterSupport": True,
             "contextSupport": True,
             "signatureInformation": {
+                "activeParameterSupport": True,
                 "documentationFormat": ["markdown", "plaintext"],
                 "parameterInformation": {
                     "labelOffsetSupport": True
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -258,8 +258,10 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
                 "documentationFormat": ["markdown", "plaintext"],
                 "parameterInformation": {
                     "labelOffsetSupport": True
-                }
-            }
+                },
+                "activeParameterSupport": True
+            },
+            "contextSupport": True
         },
         "references": {
             "dynamicRegistration": True
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -254,14 +254,14 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
         },
         "signatureHelp": {
             "dynamicRegistration": True,
+            "activeParameterSupport": True,
+            "contextSupport": True,
             "signatureInformation": {
                 "documentationFormat": ["markdown", "plaintext"],
                 "parameterInformation": {
                     "labelOffsetSupport": True
-                },
-                "activeParameterSupport": True
-            },
-            "contextSupport": True
+                }
+            }
         },
         "references": {
             "dynamicRegistration": True
--- a/plugin/core/signature_help.py
+++ b/plugin/core/signature_help.py
@@ -87,9 +87,6 @@ class SigHelp:
         new_index = self._active_signature_index + (1 if forward else -1)
         self._active_signature_index = max(0, min(new_index, len(self._signatures) - 1))
 
-    def active_signature(self) -> SignatureInformation:
-        return self._signatures[self._active_signature_index]
-
     def _render_intro(self) -> str:
         fmt = '<p><div style="font-size: 0.9rem"><b>{}</b> of <b>{}</b> overloads ' + \
               "(use ↑ ↓ to navigate, press Esc to hide):</div></p>"
--- a/plugin/core/signature_help.py
+++ b/plugin/core/signature_help.py
@@ -44,8 +44,8 @@ class SigHelp:
         self._state = state
         self._language_map = language_map
         self._signatures = self._state["signatures"]
-        self._active_signature_index = self._state.get("activeSignature", 0)
-        self._active_parameter_index = self._state.get("activeParameter", 0)
+        self._active_signature_index = self._state.get("activeSignature") or 0
+        self._active_parameter_index = self._state.get("activeParameter") or 0
 
     @classmethod
     def from_lsp(
--- a/plugin/core/signature_help.py
+++ b/plugin/core/signature_help.py
@@ -1,6 +1,5 @@
 from .logging import debug
 from .protocol import SignatureHelp
-from .protocol import SignatureHelpContext
 from .protocol import SignatureInformation
 from .registry import LspTextCommand
 from .typing import Optional, List
@@ -45,8 +44,8 @@ class SigHelp:
         self._state = state
         self._language_map = language_map
         self._signatures = self._state["signatures"]
-        self._active_signature_index = self._state.get("activeSignature") or 0
-        self._active_parameter_index = self._state.get("activeParameter") or 0
+        self._active_signature_index = self._state.get("activeSignature", 0)
+        self._active_parameter_index = self._state.get("activeParameter", 0)
 
     @classmethod
     def from_lsp(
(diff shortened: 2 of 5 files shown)


plugin/core/sessions.py
```diff
@@ -258,8 +258,10 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
                 "documentationFormat": ["markdown", "plaintext"],
                 "parameterInformation": {
                     "labelOffsetSupport": True
-                }
-            }
+                },
+                "activeParameterSupport": True
+            },
+            "contextSupport": True
```
rchl, line 264: alphabetically would be nice

plugin/core/signature_help.py
```diff
@@ -99,15 +90,13 @@ def select_signature(self, forward: bool) -> None:
     def active_signature(self) -> SignatureInformation:
```
rchl, line 90: Looks like this one is (and was) unused.

rchl (approved):

rchl:
on related note: I've been thinking a bit on how the signature help could be improved so it doesn't disappear all the time but it seems like there are some issues with the popup that would make it a bit more difficult https://github.com/sublimehq/sublime_text/issues/5518

plugin/core/sessions.py
```diff
@@ -254,14 +254,14 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
         },
         "signatureHelp": {
             "dynamicRegistration": True,
+            "activeParameterSupport": True,
```
rchl, line 257: It's not in the correct object now (also not sorted :)).

jwortmann: Oops. BTW they are not sorted strictly alphabetically in the other capabilities too. DynamicRegistration seems to be the first entry usually, even if there is another key starting with lower alphabetical letters. Initially I used the order how it is defined in the specs.

rchl: Such arbitrary logic just creates confusion and unnecessary review noise like here IMO. But it's not worth spending more time on it :)

jwortmann: I mean you was it who suggested to order alphabetically rather than follow the order from the specs ;)

rchl: Yeah. And i wished it would be like that but it's not because the dynamic one is arbitrary :)
It's not your fault though 😉

jwortmann:
> on related note: I've been thinking a bit on how the signature help could be improved so it doesn't disappear all the time but it seems like there are some issues with the popup that would make it a bit more difficult [sublimehq/sublime_text#5518](https://github.com/sublimehq/sublime_text/issues/5518)

Would it make sense to also request signature help if the cursor position is in the `meta.function-call.arguments` scope? See also https://github.com/sublimelsp/LSP/pull/163#issuecomment-335554103. It would depend on the syntax of course, but this scope is used in various of the default syntaxes (Python, Java, Lua, ...). ST still prevents the popup when typing special characters like `"` though, which leads to some kind of flickering, but it would stay open on alphanumeric characters at least.

signature_help

```diff
diff --git a/plugin/documents.py b/plugin/documents.py
index 97a16cb..c8312ce 100644
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -485,6 +485,22 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
             language_map = session.markdown_language_id_to_st_syntax_map()
             request = Request.signatureHelp(params, self.view)
             session.send_request_async(request, lambda resp: self._on_signature_help(resp, pos, language_map))
+        elif self.view.match_selector(pos, "meta.function-call.arguments"):
+            self.purge_changes_async()
+            position_params = text_document_position_params(self.view, pos)
+            context_params = {}  # type: SignatureHelpContext
+            context_params["triggerKind"] = SignatureHelpTriggerKind.ContentChange
+            context_params["isRetrigger"] = self._sighelp is not None
+            if self._sighelp:
+                context_params["activeSignatureHelp"] = self._sighelp.active_signature_help()
+            params = {
+                "textDocument": position_params["textDocument"],
+                "position": position_params["position"],
+                "context": context_params
+            }
+            language_map = session.markdown_language_id_to_st_syntax_map()
+            request = Request.signatureHelp(params, self.view)
+            session.send_request_async(request, lambda resp: self._on_signature_help(resp, pos, language_map))
         else:
             # TODO: Refactor popup usage to a common class. We now have sigHelp, completionDocs, hover, and diags
             # all using a popup. Most of these systems assume they have exclusive access to a popup, while in
```

rchl:
I've tried this with JS first and it doesn't actually use `meta.function-call.arguments`.

And while your solution is something to consider, it would still be preferred to have the reported issues fixed as it would have the benefits of:
 - less flickering
 - respect the user choice of canceling the signature help (at least until next trigger character)

jwortmann:
I had a better idea. We just don't force to close the signature help while the user is typing in that scope. See https://github.com/sublimelsp/LSP/pull/2006/commits/9f096439e5b973a6b397c8bc63df89ef834d9749. Now you can still dismiss the popup with <kbd>Esc</kbd> and it will stay closed until the next trigger character.

> I've tried this with JS first and it doesn't actually use `meta.function-call.arguments`.

Could be an incentive to improve the JS syntax with this scope ;)

rchl:
> I had a better idea. We just don't force to close the signature help while the user is typing in that scope. See [9f09643](https://github.com/sublimelsp/LSP/commit/9f096439e5b973a6b397c8bc63df89ef834d9749). Now you can still dismiss the popup with Esc and it will stay closed until the next trigger character.

I suppose that this could prove problematic when the ST issue gets fixed as then the popup wouldn't close on moving the selection but since we don't make a request, the state could get outdated.

But since it closes on moving selection now, in theory we could keep it for now.

jwortmann:
> I suppose that this could prove problematic when the ST issue gets fixed as then the popup wouldn't close on moving the selection but since we don't make a request, the state could get outdated.

I believe ST automatically closes the popup anyway when you change the cursor position with the arrow keys or via mouse click, unless the [KEEP_ON_SELECTION_MODIFIED](https://www.sublimetext.com/docs/api_reference.html#sublime.PopupFlags.KEEP_ON_SELECTION_MODIFIED) flag is used (which we do not). So it should be a non-issue and independant of the status of the linked ST issue.

I've tested it quickly, and it seems to work quite well so far and in my opinion is an improvement to the former behavior.

jwortmann:
By the way. while looking through the code I saw that LSP also closes popups whenever a command containing "delete" in its name is invoked:
https://github.com/sublimelsp/LSP/blob/2ac87a73b3a4d95c74d10c39caca4d5820d31e0c/plugin/documents.py#L437-L439

This means that for example we close the popup when pressing <kbd>Backspace</kbd> ("left_delete" command). Is that on purpose?

And in general I think this is a bad practice using `'delete' in command_name`, because it also could affect third-party plugins in an unintended way. I haven't checked the behavior for the sighelp popup if we wouldn't hide the popup in this case, but I think we should at least restrict the commands to only the known command names where closing the popup would be intended.

And I think the comment is simple wrong, because ST automatically closes the popups on ESC and arrow keys (see my last comment - and also tested and confirmed in safe-mode).

plugin/core/signature_help.py
```diff
@@ -66,26 +65,18 @@ def render(self, view: sublime.View) -> str:
...
-        if intro:
-            formatted.append(intro)
+        if self.has_multiple_signatures():
+            formatted.append(self._render_intro())
         formatted.extend(self._render_label(view, signature))
         formatted.extend(self._render_docs(view, signature))
         return "".join(formatted)
 
-    def context(self, trigger_kind: int, trigger_character: str, is_retrigger: bool) -> SignatureHelpContext:
+    def active_signature_help(self) -> SignatureHelp:
```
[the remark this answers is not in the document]
jwortmann: I named it after the "activeSignatureHelp" key of SignatureHelpContext, and the only purpose of this function right now is to retrieve this value in preparation of a new request. While it is true that it will update `self._state` (just because it is simpler this way than to create a whole new SignatureHelp object for the output value of this function), the state isn't used anymore and discarded as soon as the new response arrives. In general I'm not very picky in regard of naming and could adjust it if you want, but I don't think it would be very useful because this function isn't used elsewhere.

predragnikolic (approved):

PR #2010, opened 2022-07-31
Completions overhaul

jwortmann:
Alternative to and discussed in #2002.

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -945,10 +945,11 @@ def format_completion(
     # This is a hot function. Don't do heavy computations or IO in this function.
 
     lsp_label = item['label']
-    lsp_label_detail = item.get('labelDetails', {}).get('detail', "")
-    lsp_label_description = item.get('labelDetails', {}).get('description', "")
-    lsp_filter_text = item.get('filterText', "")
-    lsp_detail = item.get('detail', "").replace("\n", " ")
+    lsp_label_details = item.get('labelDetails') or {}
+    lsp_label_detail = lsp_label_details.get('detail') or ""
+    lsp_label_description = lsp_label_details.get('description') or ""
+    lsp_filter_text = item.get('filterText') or ""
+    lsp_detail = (item.get('detail') or "").replace("\n", " ")
 
     kind = COMPLETION_KINDS.get(item.get('kind', 0), KIND_UNSPECIFIED)
 
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -945,7 +945,7 @@ def format_completion(
     # This is a hot function. Don't do heavy computations or IO in this function.
 
     lsp_label = item['label']
-    lsp_label_details = item.get('labelDetails', {}).get('detail', "")
+    lsp_label_detail = item.get('labelDetails', {}).get('detail', "")
     lsp_label_description = item.get('labelDetails', {}).get('description', "")
     lsp_filter_text = item.get('filterText', "")
     lsp_detail = item.get('detail', "").replace("\n", " ")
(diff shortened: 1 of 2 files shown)


plugin/core/views.py
```diff
@@ -939,66 +939,51 @@ def format_diagnostic_for_html(
...
-        st_details += _wrap_in_tags("p", lsp_label)
+
+    lsp_label = item['label']
+    lsp_label_details = item.get('labelDetails') or {}
+    lsp_label_detail = lsp_label_details.get('detail') or ""
+    lsp_label_description = lsp_label_details.get('description') or ""
+    lsp_filter_text = item.get('filterText') or ""
+    lsp_detail = (item.get('detail') or "").replace("\n", " ")
+
+    kind = COMPLETION_KINDS.get(item.get('kind', 0), KIND_UNSPECIFIED)
```
rchl, line 954: Just a style but the `0` fallback seems unnecessary here. The default `None` should work the same and IMO be more self-explanatory?

rchl (approved):

plugin/core/views.py
jwortmann: The reason for it is because Pyright doesn't like it...

pyright

rchl: Oh, OK.
Though maybe -1 would be more obvious then.

plugin/core/views.py
```diff
@@ -939,66 +939,51 @@ def format_diagnostic_for_html(
...
+        annotation = lsp_label_description or lsp_detail
+    elif lsp_label.startswith(lsp_filter_text):
+        trigger = lsp_label
+        annotation = lsp_detail
+        if lsp_label_detail:
+            details.append(html.escape(lsp_label + lsp_label_detail))
+        if lsp_label_description:
+            details.append(html.escape(lsp_label_description))
+    else:
+        trigger = lsp_filter_text
```
predragnikolic, line 971:
```
	/**
	 * A string that should be used when filtering a set of
	 * completion items. When `falsy` the label is used as the
	 * filter text for this item.
	 */
	filterText?: string;
```
filterText might not be present all the time, I would suggest `trigger = lsp_filter_text or  item.get('label')`

predragnikolic (approved):

rchl:
Uses this logic in that case:

https://github.com/sublimelsp/LSP/blob/ac9144692159dea8264f4925237debb0af69cce5/plugin/core/views.py#L963-L969

Could possibly show `labelDetails.description` in annotation but then the `detail` would have to be in the bottom field or just not shown.

jwortmann:
Maybe it would be more consistent to use `annotation = lsp_label_description or lsp_detail` in all of the cases, meaning that `labelDetails.description` will always be used as a replacement for `detail`. But it seems to me that this server doesn't really follow the descriptions from the LSP specs on how to use the different fields.

CompletionItem.detail should be used for e.g. types:
```typescript
        /**
	 * A human-readable string with additional information
	 * about this item, like type or symbol information.
	 */
	detail?: string;
```

CompletionItem.labelDetails.description should be used for e.g. file paths:
```typescript
	/**
	 * An optional string which is rendered less prominently after
	 * {@link CompletionItemLabelDetails.detail}. Should be used for fully qualified
	 * names or file path.
	 */
	description?: string;
```

I would argue that file path or fully qualified name is less important than type information, and therefore should go into ST "details" pane at the bottom (currently used logic).

But feel free to open a PR with an improvement. However, if we change the "annotation" as suggested above, and another server uses the fields according to the descriptions in the docs, then we have a problem.

Also it says that `labelDetails.description` should be rendered after `labelDetails.detail`. But in the given payload, the server doesn't provide `labelDetails.detail` for the items - so it's unclear what to do with `labelDetails.description` then or if we should just ignore it in this case.

PR #2011, opened 2022-08-04
Update F# guidance

baronfel:
The current docs install FSAC as a global tool, but invoke it as a local tool. Invoking it as a global tool means executing the globally-installed shim (which is installed to a global path by default).

These updates are based on getting @matthewcrews up and running, thanks Matthew!


rwols (approved):
Thanks!

Issue #1612, opened 2021-03-11
It doesn't seem work in  ST 4098


tamamadesu:
in .TSX file ,  Here is the startup console:
![image](https://user-images.githubusercontent.com/967832/110568419-eb1f4900-818d-11eb-9d23-134c030c098b.png)

LSP-eslint.sublime-settings:
`
{
    "enabled": true,
    "settings": {
        "validate": [
            "javascript",
            "javascriptreact",
            "vue",
            "typescript",
            "typescriptreact"
        ],
        "options": {
            "parserOptions": {
                "project": [
                  "./tsconfig.json"
                ],
              }
        },
    }
}
`

Need some help, Thx !

rchl:
Moved this to the LSP repo.

At the minimum, we should print the exception to make cases like this easier to diagnose in the future.

Possibly we can gracefully handle such cases and just prints those non-recognized messages to the LSP log panel.

tamamadesu:
> LSP-eslint doesn't seem to report any error and just crashes silently.
> 
> But if you try to run `eslint` manually then you'll see what the error is:
> 
> ```
> npx eslint --ext .tsx .
> =============
> 
> WARNING: You are currently running a version of TypeScript which is not officially supported by typescript-estree.
> 
> You may find that it works just fine, or you may not.
> 
> SUPPORTED TYPESCRIPT VERSIONS: >=3.2.1 <3.4.0
> 
> YOUR TYPESCRIPT VERSION: 3.9.9
> 
> Please only submit bug reports when using the officially supported version.
> 
> =============
> 
> /temp/eslint-test/index.tsx
>   0:0  error  Parsing error: ImportDeclaration should appear when the mode is ES6 and in the module context
> 
> ✖ 1 problem (1 error, 0 warnings)
> ```
> 
> You are using very old versions of both `eslint`, `typescript`, and `typescript-eslint` plugins. Try to upgrade them all to the latest versions.

Ok， Thx!!!

rchl:
Actually, VSCode uses IPC transport and not STDIO so it's not affected by this issue.

I'm not sure we can or should even try to fix it as the fix would be against the spec.

We can probably only just make sure that the error is surfaced.

alecmev:
Continuing from #2013.

> vscode handles it fine. Looks like it ignores anything that can't be parsed as LSP messages.

[...]

Issue #1389, opened 2020-10-22
Switch to python 3.8


rwols:
Before we can use asyncio in all of its glory, we need to switch to the Python 3.8 runtime. To do that, we have to wait for a few things:

- [x] Package Control must implement an upgrade path for existing "dependencies"
- [x] The backrefs dependency must be upgraded to 3.8
- [x] The dependency bracex must be upgraded to 3.8
- [x] The dependency markupsafe must be upgraded to 3.8
- [x] The dependency mdpopups must be upgraded to 3.8
- [x] The dependency pygments must be upgraded to 3.8
- [x] The dependency pymdownx must be upgraded to 3.8
- [x] The dependency python-jinja2 must be upgraded to 3.8
- [x] The dependency python-markdown must be upgraded to 3.8
- [x] The dependency pyyaml must be upgraded to 3.8
- [x] The dependency wcmatch must be upgraded to 3.8
- [ ] sublime_lib must be upgraded to 3.8
- [ ] lsp_utils must be upgraded to 3.8
- [x] have a pull request ready for switching to 3.8 for the LSP package
- [ ] announce deadline for switching all helper packages

As far as I understand, this is also going to be a hard backwards-incompatible break. So we have to announce it.

rwols:
Package Control WIP branch is at https://github.com/wbond/package_control/tree/four-point-oh

rchl:
Realistically also `lsp_utils` and `sublime_lib` would have to be updated or most of the LSP-* packages won't work.

I only really use `ResourcePath` from `sublime_lib` so I could vendor it but it will probably be updated anyway since it's form active ST member.

predragnikolic:
I think the time for Python 3.8 if finally here :)

Just a small note,
the list of deps that LSP currently uses is this:
```
            "bracex",
            "mdpopups",
            "pathlib",
            "wcmatch",
```

I did a quick experiment.

I have only LSP and LSP-pyright setup.
I created a  .python-version file at the root of LSP and LSP-pyright.
The file just contains `3.8`.

Once I did that, I expected that I need to change something in lsp_utils,
but I didn't. I restarted ST, both LSP and LSP-pyright were running in 3.8, which is great.

I don't know why I didn't need to change anything in lsp_utils... 
it looked like PC4 already added lsp_utils to Lib/python38 with the rest of the required LSP dependencies: 
```
~/.config/sublime-text/Lib/python38  ls
[...]

PR #2015, opened 2022-08-09
Add --node-ipc support

alecmev:
Fix https://github.com/sublimelsp/LSP/issues/1612

Lots of room for refactoring and better typing and better errors, but went for the cleanest diff possible. Can be tested with stock `LSP-eslint` by changing `command` to `["${node_bin}", "${server_path}", "--node-ipc"]`.


rchl:
It looks a bit rough code-quality wise but thanks, it can be polished before merging.

High-level question to @rwols, @predragnikolic, do you prefer explicit config option for enabling node-ipc or do it like here where we check for the `--node-ipc` argument? In theory servers might implement node-ipc without using that specific argument so I probably prefer being explicit about that.

plugin/core/transports.py
```diff
@@ -79,7 +87,6 @@ def _encode(data: Dict[str, Any]) -> bytes:
         return json.dumps(
             data,
             ensure_ascii=False,
-            sort_keys=False,
             check_circular=False,
```
rchl, line 90: Any specific reason for removing this? I think the reason we had it is to not spend time on doing something not strictly needed.

plugin/core/transports.py
```diff
@@ -48,26 +49,33 @@ def on_stderr_message(self, message: str) -> None:
...
+    def read_data(self, reader: IO[bytes], is_node_ipc: bool) -> Optional[Dict[str, Any]]:
+        if not is_node_ipc:
+            headers = http.client.parse_headers(reader)  # type: ignore
+            try:
+                body = reader.read(int(headers.get("Content-Length")))
+            except TypeError:
+                # Expected error on process stopping. Stop the read loop.
+                raise StopLoopError()
+        else:
+            body = reader.readline()
```
rchl, line 77: Since the logic is pretty much separate for node_ipc, it sounds like there should be a separate processor for that case.

plugin/core/transports.py
alecmev: [`False` appears to be the default](https://docs.python.org/3.3/library/json.html).

plugin/core/transports.py
alecmev: I agree! The current name is a bit confusing, since it's an LSP processor, and JSON RPC only describes the structure of the payload. How would you name it? And the Node.js IPC equivalent?

predragnikolic: I would suggest to not extende the AbstracProcessor class with a is_node_ipc flag argument 
and to not modify the JsonRpcProcessor class.

Instead create a separate class that extends the AbstractProcessor class, called NodeIpcProcessor (feel free to name it better).

And instantiate that NodeIpcProcessor instead of JsonRpcProcessor if an explicit property flag in the client configuration field exist.

That way we avoid if statements that were added because of the is_node_ipc flag argument.

alecmev:
How about "stdio", "tcp" and "node-ipc" (as I imagine there might be other kinds of IPC out there)? Don't think RPC makes sense, since all of them are RPC.

rchl:
I would think something like `"transport": "..."` but that would be a breaking change for the `tcp` case which currently doesn't require this new option to be set. So just adding a separate new option like "use_node_ipc" might more messy but create least friction.

plugin/core/transports.py
rwols: The defining property of this node IPC protocol seems to be that messages are separated by a newline character.  Contrast to JSON-RPC, which uses HTTP headers to separate messages. I'm surprised it's that simple. I think having it called NodeIpcProcessor makes sense. All the logic of the current NodeIpcIO class can go in there.

rwols:
Yeah agreed with the boolean `use_node_ipc` setting for the ClientConfig.

There is further separation in `"tcp"`: `"tcp-server"` and `"tcp-client"` :) Anyway, offtopic for this PR.

alecmev:
Thanks for all the comments! I'll rework the processor, but in the meantime, does somebody have access to Windows? The pipe is supposed be to be cross-platform, but we're going off-piste here a bit.

predragnikolic:
Just a note of what needs to be done.

Step 1
If the server requires the '--node-ipc' argument,
in LSP-eslint we need to override the get_binary_arguments method:

```py
class NpmClientHandler(GenericClientHandler):
	...
	@classmethod
	def get_binary_arguments(cls) -> List[str]:
		return ['--stdio']
```

```
class LspEslintPlugin(NpmClientHandler):
	@classmethod
	def get_binary_arguments(cls) -> List[str]:
		return ['--node-ipc']

```
or maybe just overriding the command in LSP-eslint/LSP-eslint.sublime-settings will work:
```diff

- "command": ["${node_bin}", "${server_path}", "--stdio"],
+ "command": ["${node_bin}", "${server_path}", "--node-ipc"],
```

----

Step 2 - Lets define "use_node_ipc"

- As I understood we want the following api:
Plugins and other Client configuration can specify the option

```
// Settings in here override those in "LSP-eslint/LSP-eslint.sublime-settings"
{
	// ...
	"use_node_ipc": true // new setting
}
```

```
// LSP.sublime-settings
{
	"clients": {
		"st-themes-ls": {
			// ...
			"use_node_ipc": true // new setting
		}
	}
}

```

In LSP.sublime-package.json,
at these 2 places define "use_node_ipc":
https://github.com/sublimelsp/LSP/blob/main/sublime-package.json#L156

https://github.com/sublimelsp/LSP/blob/main/sublime-package.json#L555

That should provide autocomplete with LSP-json, when editing LSP-eslint.sublime-settings files.

---

Step 3 - read the "use_node_ipc" setting

> Re use_node_ipc, where do I put it? And should I add some sort of guard, making sure that the command is in accordance with the setting, e.g. if use_node_ipc = True then --node-ipc must be in the command?

Now we need to read "use_node_ipc setting.
The ClientConfiguration is defined in /LSP/plugin/core/types.py

We can search any existing field, and see what places need to covered. For example search "experimental_capabilities" in that file.

 "use_node_ipc" will probably need to be added to TransportConfig, 
and also the ClientConfig.resolve_transport_config method should probably be modified as well.  

---

Step 4 - determine what transport to start based on "use_node_ipc"

in LSP/plugin/core/transports.py create_transport function

```diff
json_rpc_processor = JsonRpcProcessor()
node_ipc_processor = NodeIpcProcessor()


def create_transport(config: TransportConfig, cwd: Optional[str],
                     callback_object: TransportCallbacks) -> Transport[Dict[str, Any]]:
    # ...
+    transport = json_rpc_processor
+    if config.use_node_ipc:
+        transport = node_ipc_processor
-    return ProcessTransport(config.name, process, sock, reader, writer, process.stderr, json_rpc_processor,
+    return ProcessTransport(config.name, process, sock, reader, writer, process.stderr, transport,
                            callback_object)
```

plugin/core/types.py
```diff
@@ -605,24 +607,34 @@ def map_from_remote_to_local(self, uri: str) -> Tuple[str, bool]:
         return _translate_path(uri, self._remote, self._local)
 
 
+NodeIpc = collections.namedtuple('NodeIpc', 'parent_conn,child_conn')
```
alecmev, line 610: How do I type this? I tried `NodeIpc.__annotations__`, but pyright still wouldn't pick it up. And looks like I can't use `typing.NamedTuple`, since it's a 3.5+ exclusive.

plugin/core/transports.py
```diff
@@ -136,8 +171,8 @@ def __del__(self) -> None:
 
     def _read_loop(self) -> None:
         try:
-            while self._reader:
-                payload = self._processor.read_data(self._reader)
+            while True:
```
alecmev, line 174: I tried to Google this and also replicate it in REPL with `io.BytesIO`, but it seems that streams are always truthy?Or am I taking down a Chesterton's fence here?

sublime-package.json
```diff
@@ -75,6 +75,11 @@
               },
               "markdownDescription": "The command to start the language server."
             },
+            "ClientUseNodeIpc": {
+              "type": "boolean",
+              "default": false,
+              "markdownDescription": "Communicate with the language server over Node.js IPC. This lets the server print to stdout without disrupting the LSP communication. It's non-standard, but is used by VSCode. The command must be adjusted accordingly, e.g. `--stdio` must be replaced with `--node-ipc` in case of vscode-eslint. `tcp_port` is ignored if this is enabled."
```
alecmev, line 81: Too verbose?

alecmev:
Thank you for all the feedback again! Please dogfood these changes before merging. ESLint (in the new mode), TypeScript, Pyright, JSON, HTML, YAML work for me so far, but you never know.

rchl:
Have you verified that it still works when not using `use_node_ipc`? I see some weird crashes with without even using it but don't have time to investigate right now.

predragnikolic:
I am also seeing errors,
I've time to look at this now:

<details>

<summary>Errors from ST console</summary>

```
Exception AttributeError: "'ProcessTransport' object has no attribute '_closed'" in <bound method ProcessTransport.__del__ of <LSP.plugin.core.transports.ProcessTransport object at 0x1035bac18>> ignored
Error handling request
  File "/Users/codetribe/Library/Application Support/Sublime Text/Packages/LSP/plugin/core/sessions.py", line 1930, in on_payload
    handler(result, req_id)
  File "/Users/codetribe/Library/Application Support/Sublime Text/Packages/LSP/plugin/core/sessions.py", line 1677, in m_client_registerCapability
    watcher = self._watcher_impl.create(folder.path, [pattern], kind, ignores, self)
  File "/Users/codetribe/Library/Application Support/Sublime Text/Installed Packages/LSP-file-watcher-chokidar.sublime-package/watcher.py", line 93, in create
    return file_watcher.register_watcher(root_path, patterns, events, ignores, handler)
  File "/Users/codetribe/Library/Application Support/Sublime Text/Installed Packages/LSP-file-watcher-chokidar.sublime-package/watcher.py", line 121, in register_watcher
    self._on_watcher_added(root_path, patterns, events, ignores, handler)
  File "/Users/codetribe/Library/Application Support/Sublime Text/Installed Packages/LSP-file-watcher-chokidar.sublime-package/watcher.py", line 134, in _on_watcher_added
    self._start_process()
  File "/Users/codetribe/Library/Application Support/Sublime Text/Installed Packages/LSP-file-watcher-chokidar.sublime-package/watcher.py", line 183, in _start_process
    'lspwatcher', process, None, process.stdout, process.stdin, process.stderr, StringTransportHandler(), self)
TypeError: __init__() takes 7 positional arguments but 9 were given

----

Error handling request
  File "/Users/codetribe/Library/Application Support/Sublime Text/Packages/LSP/plugin/core/sessions.py", line 1930, in on_payload
    handler(result, req_id)
  File "/Users/codetribe/Library/Application Support/Sublime Text/Packages/LSP/plugin/core/sessions.py", line 1677, in m_client_registerCapability
    watcher = self._watcher_impl.create(folder.path, [pattern], kind, ignores, self)
  File "/Users/codetribe/Library/Application Support/Sublime Text/Installed Packages/LSP-file-watcher-chokidar.sublime-package/watcher.py", line 93, in create
    return file_watcher.register_watcher(root_path, patterns, events, ignores, handler)
  File "/Users/codetribe/Library/Application Support/Sublime Text/Installed Packages/LSP-file-watcher-chokidar.sublime-package/watcher.py", line 121, in register_watcher
    self._on_watcher_added(root_path, patterns, events, ignores, handler)
  File "/Users/codetribe/Library/Application Support/Sublime Text/Installed Packages/LSP-file-watcher-chokidar.sublime-package/watcher.py", line 134, in _on_watcher_added
    self._start_process()
  File "/Users/codetribe/Library/Application Support/Sublime Text/Installed Packages/LSP-file-watcher-chokidar.sublime-package/watcher.py", line 183, in _start_process
    'lspwatcher', process, None, process.stdout, process.stdin, process.stderr, StringTransportHandler(), self)
TypeError: __init__() takes 7 positional arguments but 9 were given
```

</details>

Screenshot 2022-08-10 at 09 23 49

rchl:
Then I would try to make changes here so that it's not a breaking change for chokidar watcher. Maybe possible by making the newly added arguments optional. Watcher is using those imports: https://github.com/sublimelsp/LSP-file-watcher-chokidar/blob/1687cbf784dbad6b2fcaa4b3905c414a35cc9900/watcher.py#L7-L11

sublime-package.json
predragnikolic: Yes :)
maybe reduce it to not mention some specificty like VS Code, or LSP-eslint
```
"Communicate with the language server over Node.js IPC. The command might be adjusted accordingly, e.g. `--stdio` must be replaced with `--node-ipc` for some servers."
```

predragnikolic:
If @alecmev  don't mind I would push the change based on Rafal comment? :)

plugin/core/types.py
```diff
@@ -10,8 +10,10 @@
 from wcmatch.glob import BRACE
 from wcmatch.glob import globmatch
 from wcmatch.glob import GLOBSTAR
+import collections
```
predragnikolic, line 13: unused

plugin/core/types.py
predragnikolic: I changed this to a class:

https://github.com/sublimelsp/LSP/pull/2015/files#diff-476d0275847cfb60ebc00c73c76d5c5e6595f5b70dd8c659d4f0fd1a854f49eeR610

alecmev:
Sorry! I didn't realize all these were public APIs.

> If you mind, feel free to revert the last commit.

The opposite, thank you!

plugin/core/transports.py
```diff
@@ -68,31 +86,38 @@ def read_data(self, reader: IO[bytes]) -> Optional[Dict[str, Any]]:
...
+        message, _, self._buf = self._buf.partition(b'\n')
+        return decode_payload(message)
 
 
 class ProcessTransport(Transport[T]):
 
-    def __init__(self, name: str, process: subprocess.Popen, socket: Optional[socket.socket], reader: IO[bytes],
-                 writer: IO[bytes], stderr: Optional[IO[bytes]], processor: AbstractProcessor[T],
+    def __init__(self, name: str, process: subprocess.Popen, socket: Optional[socket.socket], reader: Any,
+                 writer: Any, stderr: Optional[IO[bytes]], processor: AbstractProcessor[T],
```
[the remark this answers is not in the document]
alecmev: This could be solved with generics, I think, but it would again require downstream changes, since it would change `AbstractProcessor`'s signature.

plugin/core/transports.py
```diff
@@ -300,7 +301,7 @@ def start_subprocess() -> subprocess.Popen:
             reader = process.stdout
             writer = process.stdin
     stderr_reader = process.stdout if config.node_ipc else process.stderr
```
predragnikolic, line 303: Can you point me to some docs on why for multiprocessing.connection the process.stdout is used instead of process.stderr?

alecmev: I redirect stdout to stderr here:

https://github.com/sublimelsp/LSP/pull/2015/files#diff-2abf74ab722813d8b12a5da0a054ac7ba7ee95f47278607834d5dc672357b818R265

To have just one pipe to monitor, instead of two. Can't do the opposite redirection, unfortunately.

plugin/core/transports.py
alecmev: @predragnikolic Do you know anything about this? Do `self._reader`/`self._writer` ever become falsey?

predragnikolic: > Do you know anything about this? 

No, I don't know, I just reverted the old code.
I will let someone with more experience answer this question. :)

I can just guess and say that
maybe the ProcessTransport close method should be modified:
```diff
    def close(self) -> None:
        if not self._closed:
            self._send_queue.put_nowait(None)
            if self._socket:
                self._socket.close()
            self._closed = True
+           self._reader.close()
+           self._writer.close()
+.          self._reader = None
+.          self._writer = None
```

alecmev: Don't know about stdio and TCP, but the multiprocessing pipe is garbage-collectable: https://docs.python.org/3/library/multiprocessing.html#multiprocessing.connection.Connection.close

plugin/core/transports.py
```diff
@@ -256,14 +289,24 @@ def start_subprocess() -> subprocess.Popen:
...
                 raise RuntimeError("Failed to connect on port {}".format(config.tcp_port))
-            reader = sock.makefile('rwb')  # type: ignore
-            writer = reader
+            reader = writer = sock.makefile('rwb')
+        elif config.node_ipc:
+            reader = writer = config.node_ipc.parent_connection  # type: ignore
         else:
-            reader = process.stdout  # type: ignore
-            writer = process.stdin  # type: ignore
+            if not process.stdout or not process.stdin:
```
predragnikolic, line 296: this raise Exception can be deleted
there is check down bellow https://github.com/sublimelsp/LSP/pull/2015/files#diff-2abf74ab722813d8b12a5da0a054ac7ba7ee95f47278607834d5dc672357b818R306

rchl: check mypy

rwols:
on Windows, LSP-json stops right away:

```
:: --> LSP-json initialize(1): {'workspaceFolders': [{'name': 'Packages', 'uri': 'file:///C:/Users/Raoul/AppData/Roaming/Sublime%20Text%203/Packages'}], 'clientInfo': {'name': 'Sublime Text LSP', 'version': '1.17.0'}, 'rootPath': 'C:\\Users\\Raoul\\AppData\\Roaming\\Sublime Text 3\\Packages', 'processId': 2224, 'capabilities': {'window': {'showDocument': {'support': True}, 'workDoneProgress': True, 'showMessage': {'messageActionItem': {'additionalPropertiesSupport': True}}}, 'general': {'regularExpressions': {'engine': 'ECMAScript'}, 'markdown': {'version': '3.2.2', 'parser': 'Python-Markdown'}}, 'workspace': {'semanticTokens': {'refreshSupport': True}, 'codeLens': {'refreshSupport': True}, 'workspaceFolders': True, 'didChangeConfiguration': {'dynamicRegistration': True}, 'configuration': True, 'symbol': {'dynamicRegistration': True, 'tagSupport': {'valueSet': [1]}, 'symbolKind': {'valueSet': [4, 5, 21, 15, 3, 1, 17, 18, 12, 7, 11, 16, 2, 25, 19, 9, 13, 10, 24, 20, 22, 23, 6, 14, 8, 26]}}, 'workspaceEdit': {'documentChanges': True, 'failureHandling': 'abort'}, 'applyEdit': True, 'executeCommand': {}}, 'textDocument': {'selectionRange': {'dynamicRegistration': True}, 'semanticTokens': {'tokenTypes': ['keyword', 'macro', 'typeParameter', 'comment', 'class', 'type', 'function', 'enumMember', 'interface', 'struct', 'decorator', 'regexp', 'operator', 'namespace', 'string', 'variable', 'enum', 'event', 'number', 'parameter', 'property', 'method', 'modifier'], 'formats': ['relative'], 'overlappingTokenSupport': False, 'multilineTokenSupport': True, 'requests': {'full': {'delta': True}, 'range': True}, 'tokenModifiers': ['deprecated', 'modification', 'static', 'declaration', 'abstract', 'defaultLibrary', 'definition', 'documentation', 'readonly', 'async'], 'dynamicRegistration': True, 'augmentsSyntaxTokens': True}, 'codeAction': {'resolveSupport': {'properties': ['edit']}, 'dynamicRegistration': True, 'dataSupport': True, 'codeActionLiteralSupport': {'codeActionKind': {'valueSet': ['quickfix', 'refactor', 'refactor.extract', 'refactor.inline', 'refactor.rewrite', 'source.organizeImports']}}, 'disabledSupport': True}, 'hover': {'dynamicRegistration': True, 'contentFormat': ['markdown', 'plaintext']}, 'signatureHelp': {'contextSupport': True, 'dynamicRegistration': True, 'signatureInformation': {'parameterInformation': {'labelOffsetSupport': True}, 'activeParameterSupport': True, 'documentationFormat': ['markdown', 'plaintext']}}, 'publishDiagnostics': {'relatedInformation': True, 'versionSupport': True, 'tagSupport': {'valueSet': [2, 1]}, 'dataSupport': True, 'codeDescriptionSupport': True}, 'declaration': {'dynamicRegistration': True, 'linkSupport': True}, 'documentHighlight': {'dynamicRegistration': True}, 'synchronization': {'willSaveWaitUntil': True, 'dynamicRegistration': True, 'willSave': True, 'didSave': True}, 'rangeFormatting': {'dynamicRegistration': True}, 'references': {'dynamicRegistration': True}, 'colorProvider': {'dynamicRegistration': True}, 'documentSymbol': {'hierarchicalDocumentSymbolSupport': True, 'dynamicRegistration': True, 'tagSupport': {'valueSet': [1]}, 'symbolKind': {'valueSet': [4, 5, 21, 15, 3, 1, 17, 18, 12, 7, 11, 16, 2, 25, 19, 9, 13, 10, 24, 20, 22, 23, 6, 14, 8, 26]}}, 'completion': {'insertTextMode': 2, 'dynamicRegistration': True, 'completionItem': {'deprecatedSupport': True, 'insertTextModeSupport': {'valueSet': [2]}, 'resolveSupport': {'properties': ['detail', 'documentation', 'additionalTextEdits']}, 'tagSupport': {'valueSet': [1]}, 'snippetSupport': True, 'documentationFormat': ['markdown', 'plaintext'], 'labelDetailsSupport': True}, 'completionItemKind': {'valueSet': [14, 19, 18, 7, 9, 1, 15, 11, 3, 10, 8, 22, 17, 24, 4, 6, 13, 16, 23, 20, 2, 21, 12, 5, 25]}}, 'formatting': {'dynamicRegistration': True}, 'typeDefinition': {'dynamicRegistration': True, 'linkSupport': True}, 'definition': {'dynamicRegistration': True, 'linkSupport': True}, 'codeLens': {'dynamicRegistration': True}, 'documentLink': {'tooltipSupport': True, 'dynamicRegistration': True}, 'implementation': {'dynamicRegistration': True, 'linkSupport': True}, 'rename': {'prepareSupport': True, 'dynamicRegistration': True}}}, 'rootUri': 'file:///C:/Users/Raoul/AppData/Roaming/Sublime%20Text%203/Packages', 'initializationOptions': {'handledSchemaProtocols': ['https', 'http', 'file'], 'provideFormatter': True, 'customCapabilities': {'rangeFormatting': {'editLimit': 1000}}}}
LSP-json: child_process.js:135
LSP-json:   p.open(fd);
LSP-json:     ^
LSP-json: 
LSP-json: Error: EBADF: bad file descriptor, uv_pipe_open
LSP-json:     at Object._forkChild (child_process.js:135:5)
LSP-json:     at setupChildProcessIpcChannel (internal/bootstrap/pre_execution.js:355:30)
LSP-json:     at prepareMainThreadExecution (internal/bootstrap/pre_execution.js:52:3)
LSP-json:     at internal/main/run_main_module.js:7:1 {
LSP-json:   errno: -4083,
LSP-json:   code: 'EBADF',
LSP-json:   syscall: 'uv_pipe_open'
LSP-json: }
```

I'll investigate some more.

rwols:
The `pass_fds` parameter of subprocess.Popen is POSIX only, so probably the subprocess doesn't see the child file descriptor at all: https://docs.python.org/3/library/subprocess.html#subprocess.Popen

alecmev:
Thank you for testing!

Sorry, for some reason I was under impression that Windows was POSIX-compliant, but turns out it was [just a subsystem](https://en.wikipedia.org/wiki/Microsoft_POSIX_subsystem), available only in professional editions, and removed a long time ago too.

There is [`lpAttributeList`](https://docs.python.org/3/library/subprocess.html#subprocess.STARTUPINFO.lpAttributeList), which is a Windows equivalent of `pass_fds`, but it is available only in Python 3.7+. I went ahead and just set `close_fds` to `False`, letting the child inherit all handles. This has security implications, of course. Does `plugin_host` isolate the plugins from each other?

When is LSP switching to 3.8?

plugin/core/transports.py
```diff
@@ -220,27 +245,37 @@ def _stderr_loop(self) -> None:
...
-json_rpc_processor = JsonRpcProcessor()
+standard_processor = StandardProcessor()
+node_ipc_processor = NodeIpcProcessor()
 
 
 def create_transport(config: TransportConfig, cwd: Optional[str],
                      callback_object: TransportCallbacks) -> Transport[Dict[str, Any]]:
+    stderr = subprocess.PIPE
+    # https://github.com/python/cpython/blob/17bf6b4671ec02d80ad29b278639d5307baddeb5/Lib/subprocess.py#L706
+    close_fds = True if sys.version_info >= (3, 8, 0) else subprocess._PLATFORM_DEFAULT_CLOSE_FDS  # type: ignore
```
rchl, line 256: `_PLATFORM_DEFAULT_CLOSE_FDS` doesn't sound like a public API, judging from the leading underscore

It looks like just omitting that argument in the subprocess constructor would be the way to go?

rchl (commented):
Note for the future: force-pushing does nothing good. Just complicates things for reviewers (granted, in that case the changes are pretty small so it's no big deal).

PR's are squashed before merging anyway so there is no need to clean up commits before.

plugin/core/transports.py
```diff
@@ -48,18 +50,35 @@ def on_stderr_message(self, message: str) -> None:
...
 
 
-class JsonRpcProcessor(AbstractProcessor[Dict[str, Any]]):
+def encode_payload(data: Dict[str, Any]) -> bytes:
+    return json.dumps(
+        data,
+        ensure_ascii=False,
+        check_circular=False,
+        separators=(',', ':')
+    ).encode('utf-8')
```
rchl, line 66: fits one line now :)

plugin/core/transports.py
```diff
@@ -68,31 +87,38 @@ def read_data(self, reader: IO[bytes]) -> Optional[Dict[str, Any]]:
...
-    def _decode(message: bytes) -> Dict[str, Any]:
-        return json.loads(message.decode('utf-8'))
+        return decode_payload(body)
+
+
+class NodeIpcProcessor(AbstractProcessor[Dict[str, Any]]):
+    _buf = bytearray()
+    _lines = 0
+
+    def write_data(self, connection: multiprocessing.connection._ConnectionBase, data: Dict[str, Any]) -> None:
```
rchl, line 97: Is there really no way to use public type here? Underscore denotes a private property.

plugin/core/transports.py
```diff
@@ -68,31 +87,38 @@ def read_data(self, reader: IO[bytes]) -> Optional[Dict[str, Any]]:
...
-
-    @staticmethod
-    def _decode(message: bytes) -> Dict[str, Any]:
-        return json.loads(message.decode('utf-8'))
+        return decode_payload(body)
+
+
+class NodeIpcProcessor(AbstractProcessor[Dict[str, Any]]):
+    _buf = bytearray()
+    _lines = 0
```
rwols, line 95: Shouldn't these two properties be part of the class instance, instead of the class? If I have LSP-dockerfile, LSP-stylelint, and LSP-typescript running with node IPC, wouldn't these overwrite each other in unpredictable ways? Maybe I'm not understanding something.

plugin/core/transports.py
```diff
@@ -68,31 +87,38 @@ def read_data(self, reader: IO[bytes]) -> Optional[Dict[str, Any]]:
...
+
+    def write_data(self, connection: multiprocessing.connection._ConnectionBase, data: Dict[str, Any]) -> None:
+        body = encode_payload(data) + b"\n"
+        while len(body):
+            n = connection._write(connection.fileno(), body)  # type: ignore
+            body = body[n:]
+
+    def read_data(self, connection: multiprocessing.connection._ConnectionBase) -> Optional[Dict[str, Any]]:
+        while self._lines == 0:
+            chunk = connection._read(connection.fileno(), 65536)  # type: ignore
```
rwols, line 105: 65536 looks like a rather large buffer (65KB), have you tried 4KB? What are the performance implications compared to the standard transport?

plugin/core/transports.py
```diff
@@ -220,27 +245,37 @@ def _stderr_loop(self) -> None:
 
 
 # Can be a singleton since it doesn't hold any state.
-json_rpc_processor = JsonRpcProcessor()
+standard_processor = StandardProcessor()
+node_ipc_processor = NodeIpcProcessor()
```
rwols, line 249: Again, I'm unsure whether this can actually be a singleton.

plugin/core/types.py
```diff
@@ -689,9 +701,10 @@ def from_sublime_settings(cls, name: str, s: sublime.Settings, file: str) -> "Cl
             priority_selector=_read_priority_selector(s),
             schemes=s.get("schemes"),
             command=read_list_setting(s, "command", []),
+            use_node_ipc=bool(s.get("use_node_ipc", False)),
             tcp_port=s.get("tcp_port"),
             auto_complete_selector=s.get("auto_complete_selector"),
-            # Default to True, because an LSP plugin is enabled iff it is enabled as a Sublime package.
```
rwols, line 694: "iff" is short for "if and only if" :) but keep the change, no one really cares.

plugin/core/transports.py
```diff
@@ -48,18 +50,35 @@ def on_stderr_message(self, message: str) -> None:
...
+        exception_log("JSON decode error", ex)
+        return None
+
+
+class StandardProcessor(AbstractProcessor[Dict[str, Any]]):
     def write_data(self, writer: IO[bytes], data: Dict[str, Any]) -> None:
-        body = self._encode(data)
+        body = encode_payload(data)
         writer.writelines(("Content-Length: {}\r\n\r\n".format(len(body)).encode('ascii'), body))
+        writer.flush()
```
rchl, line 81: This is potentially breaking for chokidar watcher. We need to add a flush there. I guess duplicate flush() might be ok to avoid adding too much logic there.

plugin/core/transports.py
rchl, line 105: Also private API...

I'm not saying it's unacceptable but is there really no higher level API exposed for that?

rwols (changes requested):
Very cool overall. With the latest changes using `close_fds` it still doesn't want to start on Windows, but I'll investigate more. I tried setting the `stdout` and `stdin` subprocess handles directly to the `parent` and `child` file descriptors but that doesn't seem to work.

rwols:
> When is LSP switching to 3.8?

See: https://github.com/sublimelsp/LSP/issues/1389

plugin/core/transports.py
rchl: I believe it should work as expected due to how Python works with those.

When instance of the class reads those properties it looks them up first in the instance and then in the class. When writing it always writes to the instance property (creates it if necessary).

The only issue would be if the instance would read a class property (for example `self._buf`) and then mutate that property (`self._buf.append(...)` whatever methods that supports).

So I guess it's still safer to set those on the instance.

plugin/core/transports.py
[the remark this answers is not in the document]
rchl: I have some local changes to clean up the types but I'm yet to find a solution that is fully properly typed. It's complicated.

In any case, that would be something for another PR.

plugin/core/transports.py
rchl: Oh, right. The standard processor wasn't meant to have any instance properties so NodeIpc shouldn't either.

plugin/core/transports.py
rwols: This also bothers me. Perhaps switching to some public mechanism also abstracts away the burden of choosing the buffer size? (Currently set to 65KB)

rwols:
Just to give an update:

- Node ultimately calls this `libuv` piece of code: https://github.com/libuv/libuv/blob/d54c92e3e68f0b8152617d8e97f704dd1e586bd6/src/win/handle-inl.h#L166
- That calls into this win32 OS function: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/get-osfhandle?view=msvc-170. This function returns INVALID_HANDLE_VALUE.
- So libuv really expects a POSIX file descriptor, not a win32 HANDLE.
- Like @alecmev mentioned, we thus really want to pass the opened POSIX file descriptor via https://docs.python.org/3/library/subprocess.html#subprocess.STARTUPINFO.lpAttributeList, but unfortunately that was introduced in Python 3.7.
- The lpAttributeList is passed via this win32 OS function: https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-updateprocthreadattribute (using PROC_THREAD_ATTRIBUTE_HANDLE_LIST)
- Maybe we can call CreateProcessW directly and wrap it as a custom subprocess.Popen? That way we may be able to pass the file descriptors. See: https://stackoverflow.com/a/29737399/990142

rchl:
Again, don't have time to investigate right now, but it appears to break LSP-chokidar, even when not using any of this functionality explicitly. No didChangeWatchedFiles notifications trigger anymore.

plugin/core/transports.py
predragnikolic: _ConnectionBase was used because multiprocessing.Pipe() uses that.
```
class NodeIpcPipe():
    def __init__(self) -> None:
        parent_connection, child_connection = multiprocessing.Pipe()
        self.parent_connection = parent_connection
        # type: _ConnectionBase
        self.child_connection = child_connection
        # type _ConnectionBase
```

I looked a bit at the `multiprocessing.Pipe()`

```
    def Pipe(duplex=True):
        '''
        Returns pair of connection objects at either end of a pipe
        '''
        if duplex:
            s1, s2 = socket.socketpair()
            s1.setblocking(True)
            s2.setblocking(True)
            c1 = Connection(s1.detach())
            c2 = Connection(s2.detach())
        else:
            fd1, fd2 = os.pipe()
            c1 = Connection(fd1, writable=False)
            c2 = Connection(fd2, readable=False)

        return c1, c2
```
and found that maybe we can use the `multiprocessing.connection.Connection` as the type.

rwols:
I gave up on the ctypes approach as it seems like a lot of work :( Maybe this encourages @wbond or @deathaxe to hurry up with the py38 dependencies problems? ;)

predragnikolic:
As much as I would like to get this merged.

I would like to ask,
It seems like python 3.8 will enable to solve this issue, https://github.com/sublimelsp/LSP/pull/2015#issuecomment-1213980226
do you think it is best to wait for 3.8 before processing further with this PR?

rwols:
Yep. Least effort is https://github.com/sublimelsp/LSP/issues/1389.

deathaxe:
> Maybe this encourages @wbond or @deathaxe to hurry up with the py38 dependencies problems? ;)

Just for the record. We are done. 🤣

rwols:
I was working on this in https://github.com/sublimelsp/LSP/tree/node-ipc, but still could not get the "named pipes" to work on Windows. If you have any ideas or insights, it would be appreciated.

Working on this I started to realize that the "duplex pipes" implementation is really just another socket or kind of file descriptor. So I don't see a theoretical performance benefit over plain stdout/stdin. I guess it may help in avoiding a bug in the eslint language server. What other benefits are there with respect to duplex pipes over stdout/stdin ?

alecmev:
> I guess it may help in avoiding a bug in the eslint language server. What other benefits are there with respect to duplex pipes over stdout/stdin ?

It comes down to [this](https://github.com/sublimelsp/LSP/issues/1612#issuecomment-1207483807), IMO:

> > such content should not be included in stdout so in theory that is something that should be fixed in https://github.com/microsoft/vscode-eslint or maybe even eslint itself
>
> This is possible (ESLint could invoke configuration files and plugins in a separate process and pipe the stdout only if asked to, and vscode-eslint could communicate with `eslint/lib/api.js` through a separate process too), but I'm not sure if such changes will ever be accepted, since both of them work fine as-is.

PR #2016, opened 2022-08-11
Handle URLs in the location like URIs

timfjord:
In [sorbet](https://sorbet.org), URLs in hovers are rendered as `location:sorbet@https://srb.help/7003#0,0` and as result, they are not handed with the `LspHoverCommand#_on_navigate` method, because it expects URLs to start with `http://` or `https://`.

I am not sure whether this is an issue with the LSP itself, but this PR handles this situation and opens URLs in the location like URIs with the `webbrowser`

### DIFF
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -336,7 +336,7 @@ class LspHoverCommand(LspTextCommand):
                     sublime.set_timeout_async(functools.partial(session.open_uri_async, uri, r))
         else:
             # NOTE: Remove this check when on py3.8.
-            if not (href.lower().startswith("http://") or href.lower().startswith("https://")):
+            if not href.lower().startswith(("http://", "https://")):
                 href = "http://" + href
             if not webbrowser.open(href):
                 debug("failed to open:", href)
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -325,7 +325,7 @@ class LspHoverCommand(LspTextCommand):
         elif is_location_href(href):
             session_name, uri, row, col_utf16 = unpack_href_location(href)
 
-            if (uri.lower().startswith("http://") or uri.lower().startswith("https://")):
+            if uri.lower().startswith(("http://", "https://")):
                 if not webbrowser.open(uri):
                     debug("failed to open:", uri)
             else:
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -324,11 +324,16 @@ class LspHoverCommand(LspTextCommand):
                 self.handle_code_action_select(config_name, 0)
         elif is_location_href(href):
             session_name, uri, row, col_utf16 = unpack_href_location(href)
-            session = self.session_by_name(session_name)
-            if session:
-                position = {"line": row, "character": col_utf16}  # type: Position
-                r = {"start": position, "end": position}  # type: RangeLsp
-                sublime.set_timeout_async(functools.partial(session.open_uri_async, uri, r))
+
+            if (uri.lower().startswith("http://") or uri.lower().startswith("https://")):
+                if not webbrowser.open(uri):
+                    debug("failed to open:", uri)
+            else:
+                session = self.session_by_name(session_name)
+                if session:
+                    position = {"line": row, "character": col_utf16}  # type: Position
+                    r = {"start": position, "end": position}  # type: RangeLsp
+                    sublime.set_timeout_async(functools.partial(session.open_uri_async, uri, r))
         else:
             # NOTE: Remove this check when on py3.8.
             if not (href.lower().startswith("http://") or href.lower().startswith("https://")):


plugin/hover.py
```diff
@@ -324,11 +324,16 @@ def _on_navigate(self, href: str, point: int) -> None:
                 self.handle_code_action_select(config_name, 0)
         elif is_location_href(href):
             session_name, uri, row, col_utf16 = unpack_href_location(href)
-            session = self.session_by_name(session_name)
-            if session:
-                position = {"line": row, "character": col_utf16}  # type: Position
-                r = {"start": position, "end": position}  # type: RangeLsp
-                sublime.set_timeout_async(functools.partial(session.open_uri_async, uri, r))
+
+            if (uri.lower().startswith("http://") or uri.lower().startswith("https://")):
```
timfjord, line 328: Theoretically, this can be extracted to an utility method: `is_URL`

plugin/hover.py
```diff
@@ -324,11 +324,16 @@ def _on_navigate(self, href: str, point: int) -> None:
...
+                if not webbrowser.open(uri):
+                    debug("failed to open:", uri)
```
timfjord, line 330: This part can be also extracted to a "private" method(`_open_url(uri)`) in this file as it is used in the `else` branch below

timfjord (commented):
Some general suggestions

rchl:
Can you report this in sorbet first? It doesn't seem like the right way to go.

timfjord:
Yep, created it now https://github.com/sorbet/sorbet/issues/6219

Just in case, in VSCode those URLs are resolved correctly, so I assume they handle `http` links even if they look like location

plugin/hover.py
jfcherng, line 328:
```suggestion
            if (uri.lower().startswith(("http://", "https://")):
```

rchl: also no need for parenthesis around

rchl:
> Yep, created it now [sorbet/sorbet#6219](https://github.com/sorbet/sorbet/issues/6219)

So they say that they provide correctly formatted URLs in their payloads. What do their payloads actually look like (not the content rendered to markdown)?

froydnj:
> > Yep, created it now [sorbet/sorbet#6219](https://github.com/sorbet/sorbet/issues/6219)
> 
> So they say that they provide correctly formatted URLs in their payloads. What do their payloads actually look like (not the content rendered to markdown)?

Actual diagnostic payload with some identifying information stripped:

```
{
  "jsonrpc": "2.0",
  "method": "textDocument/publishDiagnostics",
  "params": {
    "uri": "file:///Users/froydnj/...",
    "diagnostics": [
      {
        "range": {
          "start": {
            "line": 23,
            "character": 34
          },
          "end": {
            "line": 23,
            "character": 36
          }
        },
        "severity": 1,
        "code": 7003,
        "codeDescription": {
          "href": "http://srb.help/7003"
        },
        "message": "...",
        "relatedInformation": [
          {
            "location": {
              "uri": "file:///Users/froydnj/...",
              "range": {
                "start": {
                  "line": 22,
                  "character": 6
                },
                "end": {
                  "line": 22,
                  "character": 18
                }
              }
            },
            "message": "Got `...` originating from:"
          }
        ]
      }
    ]
  }
}
```

rchl:
OK, so it looks like a plain link. You should rather investigate how and where it becomes this weird creature.

predragnikolic:
> You should rather investigate how and where it becomes this weird creature.

[froydnj](https://github.com/froydnj) https://github.com/sorbet/sorbet/issues/6219#issuecomment-1211912983 pointed out what part of ST LSP is modifying the url.

https://github.com/sublimelsp/LSP/blob/20a1d74ad166212cd7cdb264af164b10192990de/plugin/core/views.py#L863-L868

rchl:
Possibly it's because I'm looking at this while focusing on real work but I don't understand it then...

In the payload I can see two separate things:

```
      "codeDescription": {
          "href": "http://srb.help/7003"
        },
```

and

```
        "relatedInformation": [
          {
            "location": {
              "uri": "file:///Users/froydnj/...",
              "range": {
                "start": {
                  "line": 22,
                  "character": 6
                },
                "end": {
                  "line": 22,
                  "character": 18
                }
              }
            },
            "message": "Got `...` originating from:"
          }
        ]
```

The code you've linked formats the latter. So how it then becomes a mix of those two things (`location:sorbet@https://srb.help/7003#0,0`)?

timfjord:
Sure, I will take a look

rchl:
> I agree, it should be fixed when we actually generate those links

To me it looks like you've provided a payload that doesn't show the issue as for me, when forcing that payload, it works as expected. The href is `http://srb.help/7003` so it doesn't the `is_location_href` branch.

timfjord:
So basically this part 
```json
{
   "message": "Click for more information on this error.",
   "location": {
      "range": {
         "start": {
            "line": 0,
            "character": 0
         },
         "end": {
            "line": 0,
            "character": 0
         }
      },
      "uri": "https://srb.help/7003"
   }
}
```
renders as `location:sorbet@https://srb.help/7003#0,0`

jwortmann:
Looks like a spec violation from the server to me. The type of [DiagnosticRelatedInformation](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticRelatedInformation).location is [Location](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#location), and Location.uri must be a DocumentUri, not a general URI. So it should point to an actual document (usually `file:` schema) and not to a website.

See also the description for DiagnosticRelatedInformation, which confirms it:
```typescript
/**
 * Represents a related message and source code location for a diagnostic.
 * This should be used to point to code locations that cause or are related to
 * a diagnostics, e.g when duplicating a symbol in a scope.
 */
```

jwortmann:
By the way for the purpose of linking to a website for further information about the diagnostic, there exists the [CodeDescription](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeDescription) property. The server should use that one instead.

timfjord:
@froydnj Could you please take a look at the above?
It looks like the error link should be rather returned in the `codeDescription` property

froydnj:
> @froydnj Could you please take a look at the above? It looks like the error link should be rather returned in the `codeDescription` property

Sorbet already does this (sorbet/sorbet#5714).  If you are on an older version that does not return `codeDescription`, you should upgrade.

OTOH, VSCode (among other editors) was just fine with URLs in `DiagnosticRelatedInformation.location` for a long time; it did not look very nice, but it was functional (see screenshot in above PR).  So perhaps it is worth handling this case specially in Sublime LSP.

timfjord:
Yeah, I also noticed that VSCode handles this case correctly, so I would keep the workaround

rchl:
I don't think we should have a workaround. Especially if Sorbet has already fixed it and we don't know of other servers that would do something like that.

rchl:
Closing per comment above.

PR #2017, opened 2022-08-11
Follow global settings more accurately whether to show snippet completions

jwortmann:
This is part 2 of https://github.com/sublimelsp/LSP/commit/e14bf5de0cacabea9545cf88f27cea7fa8e4e840.

Now both of the builtin ST settings are respected. For example, I personally use
```jsonc
{
    "auto_complete_include_snippets": true,  // (default value)
    "auto_complete_include_snippets_when_typing": false,
}
```

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -426,12 +426,12 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
                     on_navigate=lambda href: self._on_navigate(href, point))
 
     def on_text_command(self, command_name: str, args: Optional[dict]) -> Optional[Tuple[str, dict]]:
-        if command_name == "show_scope_name" and userprefs().semantic_highlighting:
+        if command_name == "auto_complete":
+            self._auto_complete_triggered_manually = True
+        elif command_name == "show_scope_name" and userprefs().semantic_highlighting:
             session = self.session_async("semanticTokensProvider")
             if session:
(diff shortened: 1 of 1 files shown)


plugin/documents.py
```diff
@@ -683,6 +686,7 @@ def _on_query_completions_async(self, resolve_completion_list: ResolveCompletion
         sessions = list(self.sessions_async('completionProvider'))
         if not sessions or not self.view.is_valid():
             resolve_completion_list([], 0)
+            self._auto_complete_triggered_manually = False  # reset state for next completion popup
```
rchl, line 689: It's just a minor nit but for maintenance and clarity reasons I'd prefer that we unconditionally reset it right at the start of the method. Before that we can store it in a local and pass to `_on_all_settled`.

It's a suggestion and you can follow it if you agree that it would be better but feel free to ignore it if you want.

rchl:
> `auto_complete_include_snippets_when_typing`

nice, didn't know that ST had that

rchl:
> > `auto_complete_include_snippets_when_typing`
> 
> nice, didn't know that ST had that

Though I misunderstood its purpose at first. I thought that it's about excluding or not ST snippets while it's about snippet type completions from LSP. So that's maybe not so useful to me as LSP-typescript, for example, will sometimes return snippet-type completions that are specific to the context rather than just generic snippets so in that case I'd still want those when typing...

rwols (approved):

PR #2018, opened 2022-08-12
Inlay hint

### LINKED ISSUE #1746: Look into textDocument/inlayHints
I would like to see inlay hints if possible.

Implement https://github.com/microsoft/vscode-languageserver-node/pull/772

**Additional context**
MetaLS has a custom version of this implemented: https://github.com/scalameta/metals-sublime/pull/36

predragnikolic:
This PR implements the inlay hints spec.

It will show inlay hints in the view.
It exposes a setting to show/hide inlay hints - `"show_inlay_hints": boolean`.

[UPDATED]: The toggle functionality will be left out from this PR:
It exposes a keybinding to toggle that setting:
```jsonc
    {
        "keys": [
            "UNBOUND"
        ],
        "context": [
            {
                "operator": "equal",
                "operand": "inlayHintProvider"
            }
        ]
    }
```
It exposes a command palette command to toggle that setting: `LSP: Toggle Inlay Hints`

Here is an example:

Screenshot 2022-08-12 at 14 46 27

Kapture 2022-08-12 at 14 44 13

This can be tried with this LSP-volar [PR](https://github.com/sublimelsp/LSP-volar/pull/113).

closes #1746


predragnikolic:
related https://github.com/sublimelsp/LSP-typescript/issues/134 , https://github.com/sublimelsp/LSP-rust-analyzer/issues/72

predragnikolic:
I haven't tested:
- if the `label.command` works, do not know a server that supports that
- workspace/inlayHint/refresh request works, do not know a server that supports that

plugin/core/sessions.py
```diff
@@ -349,6 +349,13 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
         "codeLens": {
             "dynamicRegistration": True
         },
+        "inlayHint": {
+            "dynamicRegistration": True,
+            "resolveSupport": {
+                # not sure if I can just put "command" because the command is nested in the label
+                "properties": ["textEdits", "command"]
```
predragnikolic, line 356: todo, I am not sure of what to put for the command,
`"label.command"` or `"command"`

sublime-package.json
```diff
@@ -508,7 +508,12 @@
                 "$ref": "sublime://settings/LSP#/definitions/ClientConfig"
               },
               "markdownDescription": "**DO NOT MODIFY THIS SETTING!** Use `\"clients\"` to override settings instead!",
-            }
+            },
+            "show_inlay_hints": {
+              "type": "boolean",
+              "default": true,
+              "markdownDescription": "Show inlay hints."
```
rwols, line 515: If there's a toggle, as well as a json setting, what does the json setting do here? What if this setting is set to `false` and I run "Toggle Inlay Hints"?

predragnikolic: The json setting is there to preserve the user choice even after a ST restart.

The command is there to make it convenient for the user to toggle that json setting.

predragnikolic: > What if this setting is set to false and I run "Toggle Inlay Hints"?

If you put show_inlay_hints false and save the setting, 
the inlay hint phantoms in the view will go away.

If you run LSP: Toggle Inlay Hints, 
it will update that show_inlay_hints to true(because it was false),
And the inlay hint phantoms in the view will be shown.

predragnikolic: The reason why the toggle command was added, is because it was a suggested feature request 
See https://github.com/sublimelsp/LSP/pull/2018#issuecomment-1213077178

plugin/core/sessions.py
rchl: Spec has an example using `label.location` so it should be `label.command` since it's not a root property of the inline hint object.

sublime-package.json
jwortmann: Maybe then let's make in more generic:
```json
    {
        "command": "lsp_toggle_setting",
        "args": {"setting": "show_inlay_hints"},
        "keys": [
            "UNBOUND"
        ],
        "context": [
            {
                "operator": "equal",
                "operand": "inlayHintProvider"
            }
        ]
    }
```

Then it can be reused for other features. The command would need to check that the given setting exists, and that it is a boolean setting. But it doesn't need to define the `capability` property, because this is already restricted by the keybinding context.

Penguin98kStudio:
Is it possible to add some settings to let the user select the different font options for inlayhint

sublime-package.json
rchl: Triggering this command rewrites the preferences file which then forces all running servers to restart. I really don't want this to work like that, especially when toggling inlay hints since it can take a while to restart and initialize a server.

While this command sounds useful for other purposes, toggling inlay hints should just result in enabling/disabling the inlay hint requests and updating inline hint phantoms, without restarting the server.

predragnikolic:
@Penguin98kStudio ST "font_face" setting will be respected. 
https://github.com/sublimelsp/LSP/pull/2018/commits/d833b4cd170a0d49bd7e36f8ce6a5e4f0c2e654e

plugin/inlay_hint.py
```diff
@@ -0,0 +1,127 @@
...
+    </body>
+    """
+    return html
+
+
+def format_inlay_hint_tooltip(view: sublime.View, tooltip: Optional[Union[str, MarkupContent]]) -> str:
+    if isinstance(tooltip, str):
+        return html.escape(tooltip)
+    elif isinstance(tooltip, dict):  # MarkupContent
+        return minihtml(view, tooltip, allowed_formats=FORMAT_MARKUP_CONTENT)
```
rchl, line 106: ST tooltips don't support HTML so will just show unformatted html code here.

plugin/inlay_hint.py
```diff
@@ -0,0 +1,127 @@
...
+            sv.remove_inlay_hint_phantom(phantom_uuid)
+        apply_text_edits_to_view(text_edits, self.view)
+
+
+def get_inlay_hint_html(view: sublime.View, inlay_hint: InlayHint, session: Session, phantom_uuid: str) -> str:
+    tooltip = format_inlay_hint_tooltip(view, inlay_hint.get("tooltip"))
+    label = format_inlay_hint_label(view, inlay_hint["label"])
+    margin_left = "0.6rem" if inlay_hint.get("paddingLeft", False) else "0"
+    margin_right = "0.6rem" if inlay_hint.get("paddingRight", False) else "0"
+    font = view.settings().get('font_face') or "monospace"
```
rchl, line 56: Is it necessary to explicitly set font? Don't phantoms inherit view/ST font by default?

plugin/inlay_hint.py
```diff
@@ -0,0 +1,127 @@
...
+        tooltip=tooltip,
+        font=font
+    )
+    can_resolve_inlay_hint = session.has_capability('inlayHintProvider.resolveProvider')
+    apply_text_edits_command = sublime.command_url('lsp_inlay_hint_apply_text_edits', {
+        'session_name': session.config.name,
+        'inlay_hint': inlay_hint,
+        'phantom_uuid': phantom_uuid
+    })
+    can_apply_text_edits = inlay_hint.get('textEdits') or (not inlay_hint.get('textEdits') and can_resolve_inlay_hint)
```
rchl, line 89: Simpler:
```suggestion
    can_apply_text_edits = inlay_hint.get('textEdits') or can_resolve_inlay_hint
```

plugin/inlay_hint.py
```diff
@@ -0,0 +1,127 @@
...
+        return minihtml(view, tooltip, allowed_formats=FORMAT_MARKUP_CONTENT)
+    else:
+        return ""
+
+
+def format_inlay_hint_label(view: sublime.View, label: Union[str, List[InlayHintLabelPart]]) -> str:
+    if isinstance(label, str):
+        return html.escape(label)
+    else:
+        return "".join("<div title=\"{tooltip}\">{value}</div>".format(
```
rchl, line 115: In combination with `can_apply_text_edits==True` this will not work because there can be no block elements inside an inline element. I saw that this causes hint content to be hidden.

plugin/inlay_hint.py
```diff
@@ -0,0 +1,127 @@
...
+            tooltip=format_inlay_hint_tooltip(view, label_part.get("tooltip")),
+            value=label_part.get("value")
+        ) for label_part in label)
```
rchl, line 118: Maybe document (comment) here that `command` and `location` are not supported.

plugin/inlay_hint.py
predragnikolic: Nope, On a Mac that I have, not setting the font-family will case the phantoms to have a sans-serif font

plugin/inlay_hint.py
```diff
@@ -0,0 +1,127 @@
...
+        if session and session.has_capability('inlayHintProvider.resolveProvider'):
+            request = Request.resolveInlayHint(inlay_hint, self.view)
+            session.send_request_async(request, lambda response: self.handle(session_name, response, phantom_uuid))
+            return
+        self.handle(session_name, inlay_hint, phantom_uuid)
+
+    def handle(self, session_name: str, inlay_hint: InlayHint, phantom_uuid: str) -> None:
+        self.handle_inlay_hint_text_edits(session_name, inlay_hint, phantom_uuid)
+
+    def handle_inlay_hint_text_edits(self, session_name: str, inlay_hint: InlayHint, phantom_uuid: str) -> None:
```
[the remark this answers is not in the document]
predragnikolic: I brought back handling of commands, so I will keep the handle function for now :)

```
   def handle(self, session_name: str, inlay_hint: InlayHint, phantom_uuid: str,
               label_part: Optional[InlayHintLabelPart] = None) -> None:
        self.handle_inlay_hint_text_edits(session_name, inlay_hint, phantom_uuid)
        self.handle_inlay_hint_command(session_name, label_part)
```

plugin/inlay_hint.py
predragnikolic: Note I brought back command support, will note that location is not supported.

plugin/inlay_hint.py
predragnikolic: Noticed this as well, in [a9e268b](https://github.com/sublimelsp/LSP/pull/2018/commits/a9e268be84a7436f16245ba817656194ea5f509d) I made sure that this does not happen

plugin/session_view.py
```diff
@@ -45,6 +45,7 @@ def __init__(self, listener: AbstractViewListener, session: Session, uri: Docume
         self._listener = ref(listener)
         self.progress = {}  # type: Dict[int, ViewProgressReporter]
         self._code_lenses = CodeLensView(self._view)
+        self._inlay_hints_phantom_set = sublime.PhantomSet(self._view, "lsp_inlay_hints")
```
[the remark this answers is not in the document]
predragnikolic: I tested cloned views plus cloned views in separate ST projects open with inlay hint toggling, but I haven't noticed duplicated phantoms.
If you notice any problems I will move them to session buffer.

plugin/session_view.py
```diff
@@ -379,5 +380,27 @@ def _on_code_lenses_resolved_async(self, mode: str) -> None:
...
+
+        new_phantoms = list(filter(
+            lambda p: getattr(p, 'lsp_uuid') != phantom_uuid,
+            self._inlay_hints_phantom_set.phantoms)
+        )
+        self._inlay_hints_phantom_set.update(new_phantoms)
+
+    def remove_all_inlay_hints(self) -> None:
+        if not self._inlay_hints_phantom_set.phantoms:
+            return
```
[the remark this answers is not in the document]
predragnikolic: I added that because `remove_all_inlay_hints` will be called on each document change, if TemporarySettings.SHOW_INLAY_HINTS is False.

So for me updating phantoms looked more expensive than having this check here.
if but will remove this check.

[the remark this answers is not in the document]
predragnikolic: [113fbb9](https://github.com/sublimelsp/LSP/pull/2018/commits/113fbb9a47b858268c0306df7f0d41bf74f810a9)

plugin/core/sessions.py
```diff
@@ -1622,6 +1643,12 @@ def m_workspace_semanticTokens_refresh(self, params: Any, request_id: Any) -> No
             else:
                 sv.session_buffer.set_semantic_tokens_pending_refresh()
 
+    def m_workspace_inlayHint_refresh(self, params: None, request_id: Any) -> None:
```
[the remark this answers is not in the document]
predragnikolic: After removing the param I got back an error from one LS that shall not be named.

> TypeError: m_workspace_inlayHint_refresh() takes 2 positional arguments but 3 were given

So I will revert that change.

plugin/inlay_hint.py
```diff
@@ -0,0 +1,127 @@
...
+
+class LspToggleInlayHintsCommand(LspTextCommand):
+    capability = 'inlayHintProvider'
+
+    def run(self, _edit: sublime.Edit, _event: Optional[dict] = None) -> None:
+        sessions = self.sessions('inlayHintProvider')
+        TemporarySettings.SHOW_INLAY_HINTS = not TemporarySettings.SHOW_INLAY_HINTS
+        for session in sessions:
+            for sv in session.session_views_async():
+                sv.session_buffer.do_inlay_hints_async(sv.view)
```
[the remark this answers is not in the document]
predragnikolic: agree will remove the Toggle of inlay hints in this PR

sublime-package.json
predragnikolic: I removed the toggle functionality in this PR, will be introduced in a future PR.

plugin/core/protocol.py
```diff
@@ -364,6 +364,37 @@ class SemanticTokenModifiers:
...
+InlayHintParams = TypedDict('InlayHintParams', {
+    'textDocument': TextDocumentIdentifier,
+    'range': RangeLsp,
+}, total=True)
+
+InlayHintLabelPart = TypedDict('InlayHintLabelPart', {
+    'value': str,
+    'tooltip': Union[str, MarkupContent],
+    'location': Location,
+    'command':  Command
```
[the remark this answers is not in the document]
jwortmann: I don't really agree with this. Using typing's `Optional` would mean for the type checker that the value could be `None`, while it would be a spec violation if the server would actually send a `null` JSON value. If desired for documentation, we could mark optional properties with a comment.

And imo the TypedDicts that currently are declared with `total=True` and use `Optional` for optional properties should be converted to `total=False` with `Optional` removed to make the file more consistent.

Using `Optional` here would also be confusing regarding LSP types that have properties explicitly allowing the value `null` with a special meaning, e.g. [WorkspaceFullDocumentDiagnosticReport](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspaceFullDocumentDiagnosticReport).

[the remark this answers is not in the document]
jwortmann: Maybe we can backport `typing.Required` and `typing.NotRequired` from Python 3.11 ([PEP 655](https://peps.python.org/pep-0655/#specification))? That would probably be the best solution for the pyright/mypy type checkers.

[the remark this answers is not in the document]
rchl: As we discussed before, this is a choice between two wrong solutions since python typing doesn't allow us to express this correctly.

Now the question is which one is "less wrong".

- with `total=True` we can access required properties without "null" checking (positive) while with `total=False` we have to null check all (negative).
- with `total=True` directly accessing optional property won't raise a type error (negative)
- `total=True` is problematic for objects that we have to construct as then we need to specify all properties, even optional ones (negative)

I guess with that list I might have convinced myself that `total=True` is worse... :)
Though with `total=False` we are often creating Pyright type errors (or silence them with `type: ignore` which is something I don't like.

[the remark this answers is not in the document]
rchl: With that said, you should choose what you prefer @predragnikolic. I will stay silent on those issues until proper python solution is out.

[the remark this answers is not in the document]
jwortmann: > I guess with that list I might have convinced myself that `total=True` is worse... :) 

:)

> Though with `total=False` we are often creating Pyright type errors (or silence them with `type: ignore`) which is something I don't like.

For that reason I currently use
```
"python.analysis.diagnosticSeverityOverrides": {
    "reportTypedDictNotRequiredAccess": "none",
}
```
in my LSP-pyright settings.

I will try to add a backport for `typing.NotRequired` to see if the Pyright errors go away with it, and open a PR if it works.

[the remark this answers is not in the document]
rchl: I guess should add to the negative that without `Optional`, even if "more correct", when reading the code I have no idea if given property is require or not and it's hard for me to reason about the flow of the logic. I need to lookup actual types in LSP spec for clarification.

plugin/core/sessions.py
```diff
@@ -542,6 +560,9 @@ def set_semantic_tokens_pending_refresh(self, needs_refresh: bool = True) -> Non
     def get_semantic_tokens(self) -> List[Any]:
         ...
 
+    def do_inlay_hints_async(self, view: sublime.View) -> None:
```
rchl, line 563: Existing convention in this interface has those handlers starting with `on_*`.

LSP.sublime-settings
```diff
@@ -134,6 +134,9 @@
   // "phantom" - show a phantom on the top when code actions are available
   "show_code_lens": "annotation",
 
+   // Show inlay hints
```
rchl, line 137: It's quite redundant comment.
Maybe something like this at least?
`// Show inlay hints that interleave with the code.`

plugin/inlay_hint.py
```diff
@@ -0,0 +1,154 @@
...
+class LspInlayHintClickCommand(LspTextCommand):
+    capability = 'inlayHintProvider'
+
+    def run(self, _edit: sublime.Edit, session_name: str, inlay_hint: InlayHint, phantom_uuid: str,
+            event: Optional[dict] = None, label_part: Optional[InlayHintLabelPart] = None) -> None:
+        """
+        Insert textEdits for the given inlay hint.
+        If a InlayHintLabelPart was clicked, label_part will be passed as an argument to the LspInlayHintClickCommand
+        and InlayHintLabelPart.command will be executed.
+        """
```
rchl, line 21: Just a nit but this is not really a public documentation but rather description of internal logic so IMO shouldn't be using a docblock syntax for it, just normal comments.

plugin/session_view.py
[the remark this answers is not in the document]
predragnikolic: toggling was removed so this is not relevant anymore

plugin/core/sessions.py
predragnikolic: the `on_*` convention doesn't sound right,
we use `on_*` convention for `on_diagnostics_async`, but that is for diagnostics that are published from the server and not requested  by the client.

for me the `do_inlay_hints_async` is more appropriate because the client request things from the server.

if you still want me to rename this, just leave a comment. :)

Penguin98kStudio: btw, the do_* convention was also used for the do_semantic_tokens_async function

plugin/inlay_hint.py
predragnikolic: https://github.com/sublimelsp/LSP/pull/2018/commits/d370af964cfb1ec8594ebeec457e3ec328f30028

plugin/core/protocol.py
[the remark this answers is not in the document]
predragnikolic: I hope that in the following 2 months we will migrate to py 3.8.
and include the typing_extension package in ST and properly use the `NotRequired` type.

plugin/core/strip_html.py
```diff
@@ -0,0 +1,29 @@
+from html.parser import HTMLParser
```
predragnikolic, line 1: this file was inspired by https://bobbyhadz.com/blog/python-remove-html-from-string

tests/test_strip_html.py
```diff
@@ -0,0 +1,9 @@
+import unittest
+from LSP.plugin.core.strip_html import strip_html
+
+
+class StripHtmlTest(unittest.TestCase):
+    def test_strip_html(self):
+        result = strip_html("<p>Hello <b>world</b><span id='hello'>!</span></p>")
+        expected = "Hello world!"
```
rchl, line 8: Does it handle <br>'s or just block elements (should create new lines)?

I'm honestly surprised you went this way. It's a very heavy approach and not sure if really reliable. I was thinking that we could just show the original markdown markup which should be pretty readable itself.

plugin/session_view.py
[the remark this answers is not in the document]
rchl: I'll experiment with it a bit when I have more time.

tests/test_strip_html.py
predragnikolic: ++ agree, will just show the original tooltip text without parsing

plugin/core/sessions.py
```diff
@@ -1622,6 +1643,12 @@ def m_workspace_semanticTokens_refresh(self, params: Any, request_id: Any) -> No
             else:
                 sv.session_buffer.set_semantic_tokens_pending_refresh()
 
+    def m_workspace_inlayHint_refresh(self, params: None, request_id: Any) -> None:
+        """handles the workspace/inlayHint/refresh request"""
+        for sv in self.session_views_async():
```
jwortmann, line 1648: Instead of requesting new inlay hints for all of the views, maybe it would be better to use the logic that I used at the semantic tokens refresh. Meaning that only requests for the views that are currently visible ( = selected) are made, and for the other views a variable is set that a request should be made as soon as the view gets activated. See also the description of the refresh requests in the lsp specs, that say clients can delay request for non-visible tabs. The reason I did it this way was because one server (I think it was rust-analyzer) would send a refresh request after each text change. And if someone has for example 10 tabs in the window, but only one of it active, then we send 9 useless requests all the time and do the inlay hints processing and phantom updates even for views that are not visible currently.

tests/test_strip_html.py
predragnikolic: https://github.com/sublimelsp/LSP/pull/2018/commits/3779a1345fc2ec8a511608e9daedd6ae15bc20de

plugin/session_view.py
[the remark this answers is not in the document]
predragnikolic: I managed to produce a case where it would cause to show duplicate phantoms.
So I moved the phantom set to the session buffer.

plugin/core/sessions.py
predragnikolic: I do not think that I would work on optimizing this.

Inlay hints doesn't seem like an expensive request (like semantic token is). 
If this turns to be a problem only than I would optimize and add more code.

Here is one implementation detail of how inlay hints currently function.
If the server implements the workspace/inlayHint/refresh request,
we will request inlay hints for each session view.

So the client will refresh the inlay hints each time the server desires,
as an end result we should not end up with stale inlay hints in one view ever.

Even if the server supports or does not support the workspace/inlayHint/refresh request,
there is this peace of code that will request inlay hints as soon as the view gets active:

https://github.com/sublimelsp/LSP/pull/2018/files#diff-982c110867d9809c978a5597e4da66ce7538b327675a66b8f27214413b7c0ec1R345

This logic is redundant for servers that do support workspace/inlayHint/refresh request,
but I would say that it is necessary for servers for that do not support workspace/inlayHint/refresh. (just in case to update stale inlay hints as soon as the user focuses one view)

But as I said I would not add any code for optimizing this until it turns out to be a perf issue.

plugin/inlay_hint.py
```diff
@@ -0,0 +1,155 @@
...
+        font=font,
+        label=label
+    )
+    return html
+
+
+def format_inlay_hint_tooltip(tooltip: Optional[Union[str, MarkupContent]]) -> str:
+    if isinstance(tooltip, str):
+        # Sublime Text doesn't support HTML for tooltips
+        # so we strip the HTML tags from the tooltip
```
rchl, line 101: Can be removed.

plugin/inlay_hint.py
```diff
@@ -0,0 +1,155 @@
...
+            value += "</a>"
+        # InlayHintLabelPart.location is not supported
+        result += "<div title=\"{tooltip}\">{value}</div>".format(
+            tooltip=format_inlay_hint_tooltip(label_part.get("tooltip")),
+            value=value
+        )
+    return result
+
+
+def inlay_hint_to_phantom(view: sublime.View, inlay_hint: InlayHint, session: Session) -> sublime.Phantom:
```
rchl, line 148: Can you move this function before the `get_inlay_hint_html` so that it's in an easier to read order?

plugin/inlay_hint.py
```diff
@@ -0,0 +1,155 @@
...
+        return result
+
+    for label_part in label:
+        value = ""
+        inlay_hint_click_command = sublime.command_url('lsp_inlay_hint_click', {
+            'session_name': session.config.name,
+            'inlay_hint': inlay_hint,
+            'phantom_uuid': phantom_uuid,
+            'label_part': label_part
+        })
```
rchl, line 133: Can be moved into the `if is_clickable` block.
Same below.

plugin/session_buffer.py
```diff
@@ -112,6 +116,7 @@ def __init__(self, session_view: SessionViewProtocol, buffer_id: int, uri: Docum
         self._last_semantic_region_key = 0
         self._check_did_open(view)
         self._session.register_session_buffer_async(self)
+        self._inlay_hints_phantom_set = sublime.PhantomSet(view, "lsp_inlay_hints")
```
rchl, line 119: Not that it matter in the current code but would be better to move it before `_check_did_open` since it technically can already do something with inlay hints (even if async).

plugin/session_buffer.py
```diff
@@ -181,6 +187,7 @@ def add_session_view(self, sv: SessionViewProtocol) -> None:
 
     def remove_session_view(self, sv: SessionViewProtocol) -> None:
         self._clear_semantic_token_regions(sv.view)
+        self.remove_all_inlay_hints()
```
rchl, line 190: That's not right because phantoms are per-buffer so should not remove them if the first view of many is closed, for example. You can probably check if any views are left and only do that if not.

plugin/session_buffer.py
```diff
@@ -611,6 +619,40 @@ def set_semantic_tokens_pending_refresh(self, needs_refresh: bool = True) -> Non
...
+
+    def _on_inlay_hints_async(self, response: InlayHintResponse) -> None:
+        if response:
+            view = self.some_view()
+            if not view:
+                return
+            phantoms = [inlay_hint_to_phantom(view, inlay_hint, self.session) for inlay_hint in response]
+            sublime.set_timeout(lambda: self.present_inlay_hints_async(phantoms))
+
+    def present_inlay_hints_async(self, phantoms: List[sublime.Phantom]) -> None:
```
rchl, line 643: This is done on the main thread so should not have `_async` suffix.

plugin/session_buffer.py
```diff
@@ -611,6 +619,40 @@ def set_semantic_tokens_pending_refresh(self, needs_refresh: bool = True) -> Non
...
+        }  # type: InlayHintParams
+        self.session.send_request_async(Request.inlayHint(params, view), self._on_inlay_hints_async)
+
+    def _on_inlay_hints_async(self, response: InlayHintResponse) -> None:
+        if response:
+            view = self.some_view()
+            if not view:
+                return
+            phantoms = [inlay_hint_to_phantom(view, inlay_hint, self.session) for inlay_hint in response]
+            sublime.set_timeout(lambda: self.present_inlay_hints_async(phantoms))
```
rchl, line 641: Not entirely sure how a None/null response should be interpreted but I would think that it might be correct to clear phantoms in that case. Maybe some servers instead of sending empty array would send null instead to indicate that no hints are to be shown.

plugin/session_view.py
```diff
@@ -80,6 +80,7 @@ def on_before_remove(self) -> None:
             self.view.erase_regions(self.diagnostics_key(severity, False))
             self.view.erase_regions(self.diagnostics_key(severity, True))
         self.view.erase_regions("lsp_document_link")
+        self.session_buffer.remove_all_inlay_hints()
```
rchl, line 83: Still not right. This can be moved back to previous place but you should only call it there when `self.session_views` becomes empty.

plugin/session_buffer.py
predragnikolic: [56f4f6e](https://github.com/sublimelsp/LSP/pull/2018/commits/56f4f6e004ce8b50909a954ddaaaffb0f7a5aea7) take a look if this is ok

plugin/core/sessions.py
```diff
@@ -542,6 +551,18 @@ def set_semantic_tokens_pending_refresh(self, needs_refresh: bool = True) -> Non
...
+        ...
+
+    def present_inlay_hints_async(self, phantoms: List[sublime.Phantom]) -> None:
+        ...
+
+    def remove_inlay_hint_phantom(self, phantom_uuid: str) -> None:
+        ...
+
+    def remove_all_inlay_hints(self) -> None:
+        ...
```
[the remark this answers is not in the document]
predragnikolic: this one needs to stay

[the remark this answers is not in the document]
predragnikolic: after moving the code to clear inlay hits from on_before_remove, to remove_session_view,
remove_all_inlay_hints can be removed from the interface  [e63422e](https://github.com/sublimelsp/LSP/pull/2018/commits/e63422e8e70de6b8f7cc9b1326d6c9626cfa4bf6)

plugin/session_view.py
predragnikolic: [3823e3b](https://github.com/sublimelsp/LSP/pull/2018/commits/3823e3b7c69b2cf9d7ec29321f38d0fe426d49a7) +
[2523ed8](https://github.com/sublimelsp/LSP/pull/2018/commits/2523ed8743e7184a93ec1e511fffc65475acbca3) ?

rchl (approved):

plugin/inlay_hint.py
```diff
@@ -0,0 +1,153 @@
...
+    p = sublime.Phantom(region, content, sublime.LAYOUT_INLINE)
+    setattr(p, 'lsp_uuid', phantom_uuid)
+    return p
+
+
+def get_inlay_hint_html(view: sublime.View, inlay_hint: InlayHint, session: Session, phantom_uuid: str) -> str:
+    tooltip = format_inlay_hint_tooltip(inlay_hint.get("tooltip"))
+    label = format_inlay_hint_label(inlay_hint, session, phantom_uuid)
+    margin_left = "0.6rem" if inlay_hint.get("paddingLeft", False) else "0"
+    margin_right = "0.6rem" if inlay_hint.get("paddingRight", False) else "0"
```
rchl, line 73: I have now tested this a bit more and I don't think this looks good. The phantoms are spaced too much from the code:
Screenshot 2022-08-19 at 23 34 22
 
The likely reason for this is that ST by itself adds a small margin already so I don't think it makes sense to apply more, even if paddingLeft or paddingRight is set.

rchl: VSCode for comparision. Se how the first hint (`title`) is just next to the parenthesis and has reasonable (not too big) margin on the right.

Screenshot 2022-08-19 at 23 37 30

rchl (approved):

LSP.sublime-settings
```diff
@@ -134,6 +134,9 @@
   // "phantom" - show a phantom on the top when code actions are available
   "show_code_lens": "annotation",
 
+   // Inlay hints are short annotations within the code, which show variable types or parameter names.
+  "show_inlay_hints": true,
```
rwols, line 138: I would prefer this to be set to false by default. The phantoms don’t behave perfectly UX-wise. Furthermore there’s a clear divide in preference for this feature by users; we can see some issues in LSP-rust-analyzer by users asking how to turn this off.

plugin/core/types.py
```diff
@@ -243,6 +243,7 @@ def r(name: str, default: Union[bool, int, str, list, dict]) -> None:
         r("semantic_highlighting", False)
         r("show_code_actions", "annotation")
         r("show_code_lens", "annotation")
+        r("show_inlay_hints", True)
```
rwols, line 246: Set to false by default :)

sublime-package.json
```diff
@@ -508,7 +508,12 @@
                 "$ref": "sublime://settings/LSP#/definitions/ClientConfig"
               },
               "markdownDescription": "**DO NOT MODIFY THIS SETTING!** Use `\"clients\"` to override settings instead!",
-            }
+            },
+            "show_inlay_hints": {
+              "type": "boolean",
+              "default": true,
```
rwols, line 514: Set to false please :)

sublime-package.json
```diff
@@ -508,7 +508,12 @@
...
+              "markdownDescription": "Inlay hints are short annotations within the code, which show variable types or parameter names."
```
rwols, line 515: This now explains what inlay hints are, which is good 👍, but it doesn’t explain what the setting does.

rwols (changes requested):

sublime-package.json
predragnikolic: Feel free to tweak the setting description.
For now I pushed, just to make it more explicit of what the setting does. [804f1c0](https://github.com/sublimelsp/LSP/pull/2018/commits/804f1c0a4c195b8bd0c7798389184c87a0a32050)

Although that setting name for me is pretty self explanatory, but that is subjective :)

rwols (approved):

predragnikolic:
🙌

jwortmann:
I wonder if inlay hints should use an increased debounce time. Currently they are forced to use the FEATURES_TIMEOUT defined as 300 ms for the debounce time after text changed.
https://github.com/sublimelsp/LSP/blob/ee8fc6e16c19fc5478e79d83e281cd6dc18924aa/plugin/session_buffer.py#L275

VSCode uses 1250 ms, see https://github.com/microsoft/vscode/issues/133730

predragnikolic:
@jwortmann yep, I can see that being an issue.

rchl:
The only argument against doing that right now might be the fact that ST UX is kinda broken with those and shorter timeout helps with fixing ST issues faster while typing. For example the phantom doesn't shift to the right when typing right in front of it.

jwortmann:
Another thing I see in the code is that an ongoing request is not canceled when another request is to be made and the response for the previous one hasn't arrived yet. This also means that the result will be drawn unconditionally, despite that the response arrives after more changes in the buffer were made.

For semantic tokens I added a line to cancel a pending request whenever a new request is in preparation:
https://github.com/sublimelsp/LSP/blob/ee8fc6e16c19fc5478e79d83e281cd6dc18924aa/plugin/session_buffer.py#L512-L513

I'm not sure what the expected average/maximum times for inlay hint responses is, but it might be an improvement to also cancel the ongoing request if new changes were made. In fact, from the linked issue apparently VSCode does it this way.

But compared to my implementation of the request canceling for semantic tokens, I think for inlay hints it would be better to move the part shown in the two lines above into `on_text_changed_async` under `if purge:` (and not within the debounced part, which is the case for semantic tokens):
https://github.com/sublimelsp/LSP/blob/ee8fc6e16c19fc5478e79d83e281cd6dc18924aa/plugin/session_buffer.py#L274-L276

Because at that time we already know that a new request will be made very soon and that the pending results are already outdated and drawing them might make the cursor jump arround annoyingly, so that the resonse should better be ignored instead.

PR #2024, opened 2022-08-20
Add overridable methods to LspExecuteCommand

### LINKED ISSUE #1902: Provide an optional response handler in LspExecuteCommand
I would like to extends `LspExecuteCommand` in order to send a `workspace/executeCommand` but `LspExecuteCommand` simply [logs](https://github.com/sublimelsp/LSP/blob/acfd6406ba4680a0e537dc87a72aa5b410a154e7/plugin/execute_command.py#L47) the response. In my case I have to open a file URI that is in the reponse.

`LspExecuteCommand` should provide an optional response handler. When missing simply log the response as the case now. Otherwise delegate the response handling. 

The current alternative is to copy `LspExecuteCommand` logic.

rchl:
A simple addition to make it so that plugins can easily use it with custom response handlers without re-implementing the `run` method.

I've changed the default success handler to not stringify the response as the response can be anything, including a complex object, so it doesn't make sense to assume that it's something that can be presented in the status bar.

Resolves #1902

### DIFF
--- a/plugin/execute_command.py
+++ b/plugin/execute_command.py
@@ -37,16 +37,20 @@ class LspExecuteCommand(LspTextCommand):
             def handle_response(response: Any) -> None:
                 assert command_name
                 if isinstance(response, Error):
-                    sublime.message_dialog("command {} failed. Reason: {}".format(command_name, str(response)))
+                    self.handle_error_async(response, command_name)
                     return
-                msg = "command {} completed".format(command_name)
-                if response:
-                    msg += "with response: {}".format(response)
-                window = self.view.window()
-                if window:
-                    window.status_message(msg)
+                self.handle_success_async(response, command_name)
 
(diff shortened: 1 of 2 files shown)


plugin/execute_command.py
```diff
@@ -37,17 +37,21 @@ def run_async() -> None:
...
-                if response:
-                    msg += "with response: {}".format(response)
-                window = self.view.window()
-                if window:
-                    window.status_message(msg)
+                self.handle_success_async(response, command_name)
 
             session.execute_command(params, progress=True).then(handle_response)
 
+    def handle_success_async(self, result: Any, command_name: str) -> None:
```
rwols, line 46: Can you add a docblock here noting that plugins may override this? Otherwise it’s not clear to me that any plugin would just override this (and accidentally break in the future).

rwols (approved):

PR #2027, opened 2022-08-21
Allow plugins to override name shown in the status field

rchl:
I'm fine with this not going in if there are objections.

On one hand I think it can create a bit of confusion as the name in the status won't match package name. On the other hand, this is already the case for packages overriding `AbstractPlugin.name()` and I'm very much against overriding `AbstractPlugin.name()` as that also affects configuration keys and I believe that configuration keys should match the package name.

The only point of using this would be to create a bit cleaner look in the status field.

Screenshot 2022-08-21 at 23 26 12

vs.

Screenshot 2022-08-21 at 23 26 38

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -683,6 +683,16 @@ class AbstractPlugin(metaclass=ABCMeta):
         raise NotImplementedError()
 
     @classmethod
+    def status_name(cls) -> str:
+        """
+        Overrides the name shown in the status field for a running server. Defaults to name().
+
+        The difference between overriding name() and this is that name() also affects the names of the configuration
+        keys, name shown in the error messages and other places.
+        """
+        return cls.name()
+
+    @classmethod
     def configuration(cls) -> Tuple[sublime.Settings, str]:
         """
         Return the Settings object that defines the "command", "languages", and optionally the "initializationOptions",
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -668,6 +668,7 @@ class ClientConfig:
         self.file_watcher = file_watcher
         self.path_maps = path_maps
         self.status_key = "lsp_{}".format(self.name)
+        self.status_name = self.name
         self.semantic_tokens = semantic_tokens
 
     @classmethod
@@ -793,9 +794,12 @@ class ClientConfig:
                 env[key] = sublime.expand_variables(value, variables)
         return TransportConfig(self.name, command, tcp_port, env, listener_socket)
 
+    def override_status_name(self, name: str) -> None:
+        self.status_name = name
+
     def set_view_status(self, view: sublime.View, message: str) -> None:
         if sublime.load_settings("LSP.sublime-settings").get("show_view_status"):
-            status = "{}: {}".format(self.name, message) if message else self.name
+            status = "{}: {}".format(self.status_name, message) if message else self.status_name
             view.set_status(self.status_key, status)
 
     def erase_view_status(self, view: sublime.View) -> None:
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -244,6 +244,7 @@ class WindowManager(Manager):
             variables = extract_variables(self._window)
             cwd = None  # type: Optional[str]
             if plugin_class is not None:
+                config.override_status_name(plugin_class.status_name())
                 if plugin_class.needs_update_or_installation():
                     config.set_view_status(initiating_view, "installing...")
                     plugin_class.install_or_update()


plugin/core/sessions.py
```diff
@@ -682,6 +682,16 @@ def name(cls) -> str:
         """
         raise NotImplementedError()
 
+    @classmethod
+    def status_name(cls) -> str:
```
predragnikolic, line 686: this is similar to lsp_utils `get_displayed_name`?
https://github.com/sublimelsp/lsp_utils/blob/master/st3/lsp_utils/_client_handler/interface.py#L29

predragnikolic: Based on the `name` method description,
it sounds like the `name` method already does what `status_name` wants to do:
```
    @classmethod
    @abstractmethod
    def name(cls) -> str:
        """
        A human-friendly name. If your plugin is called "LSP-foobar", then this should return "foobar". If you also
        have your settings file called "LSP-foobar.sublime-settings", then you don't even need to re-implement the
        configuration method (see below).
        """
        raise NotImplementedError()
```

rchl: Your comment makes me think that you haven't read the description I gave in the initial comment. :)

As for `get_displayed_name()`, it overrides `name()` under the hood.

rchl: But to clarify, this only affects the name shown in the status field. `name()` affects a lot more (configuration keys, name in LSP logs, error messages and possibly more).

rwols:
Doesn't this only confuse users that want to override their stuff in a .sublime-project? Because now the string in the status bar can be different from the string that you're supposed to write in .sublime-project.

rchl:
> Doesn't this only confuse users that want to override their stuff in a .sublime-project? Because now the string in the status bar can be different from the string that you're supposed to write in .sublime-project.

When overriding `name()` the menu items and commands in the palette are also confusing as those still refer to the package through package's name.

The least confusing thing is not to override any names.

jwortmann:
Is it even possible not to override `name()`? It would throw a NotImplementedError if you don't override it and the docstring also says that it must be overridden.

What I would expect though, is that LSP should add a `LSP-` prefix in the status bar by default. Perhaps it would be useful to have another classmethod `status_prefix() -> bool` which plugins could override to opt-out or opt-in to that.

rchl:
> Is it even possible not to override `name()`? It would throw a NotImplementedError if you don't override it and the docstring also says that it must be overridden.

True, I mispoke but my point was that the `name()` should be matching the name of the package rather than diverging from it.

> What I would expect though, is that LSP should add a `LSP-` prefix in the status bar by default. Perhaps it would be useful to have another classmethod `status_prefix() -> bool` which plugins could override to opt-out or opt-in to that.

So are you suggesting that a package `LSP-foo` should return `foo` from `name()` and LSP should be adding `LSP-` prefix for the status only?

That would still create confusion that @rwols talked about, no?

jwortmann:
> So are you suggesting that a package `LSP-foo` should return `foo` from `name()`

Well, it's not me but the docstring which claims this. And I do follow that in LSP-julia. Unfortunately this leads to the status bar label to be only "julia" and I would prefer it to be "LSP-julia" instead. For language servers which have a unique name, like "Pyright" or "rust-analyzer", the "LSP-" prefix is probably not necessary, but for helper plugins which just use the language name it would be more clear to have the prefix in the status bar. Not sure why it is shown for some other plugins like LSP-css.

> That would still create confusion that @rwols talked about, no?

I must say that I don't use any config overrides in .sublime-project files personally, so I'm not aware of possible implications. But yes, a backward incompatible change here would be quite unfortunate. But I would say the current state is rather confusing too, because some plugins show this prefix and some don't.

rchl:
> Well, it's not me but the docstring which claims this. And I do follow that in LSP-julia. Unfortunately this leads to the status bar label to be only "julia" and I would prefer it to be "LSP-julia" instead. For language servers which have a unique name, like "Pyright" or "rust-analyzer", the "LSP-" prefix is probably not necessary, but for helper plugins which just use the language name it would be more clear to have the prefix in the status bar. Not sure why it is shown for some other plugins like LSP-css.

I was never really a fan of the idea that `name()` should return name without `LSP-` for the reasons stated in my initial comment. That's why I haven't made `lsp_utils` follow that behavior by default (and why LSP-css behaves like it does now, for example).

But I agree that for some servers it makes sense to show the LSP prefix and for some not. `pyright` and `pylsp` can stand on its own without the prefix but `LSP-css` or `LSP-julia` without the prefix would just look like the syntax toggle on the other side of the status bar.

Your suggestion to add `status_prefix()` is not as flexible as this one and wouldn't work to remove the prefix when `name()` included the prefix already.

rchl:
Though I don't think we'll come to any reasonable conclusion here as the state of the matter is already confusing and this potentially will add even more confusion.

PR #2030, opened 2022-08-24
Add option to highlight hover range

jwortmann:
Inspired by VSCode. Not sure if it's needed, and there seems to be a slight aversion to introduce new settings. But I kind of like it (and it's disabled by default).

hover

For the "document_highlight_style" setting I have also renamed the "fill" option to "background" (which I now think is a better name), but kept it fully backwards compatible with "fill".

Feel free to merge or close.

### DIFF
--- a/docs/src/customization.md
+++ b/docs/src/customization.md
@@ -156,7 +156,18 @@ If neither a scope for a custom token type is defined, nor a color scheme rule f
 | `markup.highlight.write.lsp` | Write | Write-access of a symbol, like writing to a variable |
 
 !!! note
-    If `document_highlight_style` is set to "fill" in the LSP settings, the highlighting color can be controlled via the "background" color from a color scheme rule for the listed scopes.
+    If `document_highlight_style` is set to "background" in the LSP settings, the highlighting color can be controlled via the "background" color from a color scheme rule for the listed scopes.
+
+### Hover Highlights
+
+Allows to highlight the word or range for which a hover popup is shown (disabled by default).
+
+| scope |
+| ----- |
+| `markup.highlight.hover.lsp` |
+
+!!! note
+    If `hover_highlight_style` is set to "background" in the LSP settings, the highlighting color can be controlled via the "background" color from a color scheme rule.
 
 ### Diagnostics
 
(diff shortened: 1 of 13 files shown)


docs/src/customization.md
```diff
@@ -156,7 +156,18 @@ If neither a scope for a custom token type is defined, nor a color scheme rule f
...
+### Hover Highlights
+
+Allows to highlight the word or range for which a hover popup is shown (disabled by default).
+
+| scope |
+| ----- |
+| `markup.highlight.hover.lsp` |
+
+!!! note
+    If `hover_highlight_style` is set to "background" in the LSP settings, the highlighting color can be controlled via the "background" color from a color scheme rule.
```
rchl, line 170: I suppose it's expected then that without customizing it it won't look very good (poor contrast)?

jwortmann: Yes, without a special rule "background" is basically unusable (same as for documentHighlights). Should we mention that in the settings description, or maybe point users more agressively to this documentation?

jwortmann: Another alternative would be to provide such a premade rule for the default color schemes in the override files that we already ship. And we could do the same for DocumentHighlights. We just need to agree on some styles which look nice then. I already do this in my own color scheme for example - a background color with alpha value around 0.2 looks nice.

jwortmann: I've added some styles for the default schemes in https://github.com/sublimelsp/LSP/pull/2030/commits/401455408f75a78b4b118ad4cf88ec474c037096.

Feedback appreciated if I should adjust the colors or revert this commit.

Note that I have only addressed `markup.highlight` for the DocumentHighlights, because providing a default rule for the subscopes like `markup.highlight.read` would override the style if a user already has a customization for the more general scope name.

rchl:
I've been using it for a while and it looks good to me. I don't mind the new setting.

I've noticed that it can sometimes happen that the range is not removed after popup is hidden but it seems like an ST issue and it's not a big deal anyway as it can be fixed by hovering something else.

rchl (approved):

jwortmann:
I am able to reproduce the "regions not removed" bug when there are multiple plugins fighting for the popup. But I wasn't able yet to create a minified example in safe mode for the ST issue tracker.

And I think too, that this isn't a critical issue, especially since the setting for the range highlight is chosen to be disabled by default.

Issue #2022, opened 2022-08-19
Ability to open in specified group for go to commands


ninth-dev:
**Is your feature request related to a problem? Please describe.**

I want goto definition to always open in a certain group.

**Describe the solution you'd like**

- Allow a group arg to be supplied to the lsp goto commands. 

**Describe alternatives you've considered**

- hacking the lsp code locally 
- side-by-side, but i want to be able to resize the panel so must use a group.

**Additional context**

<img width="2402" alt="Screen Shot 2022-08-19 at 4 40 45 pm" src="https://user-images.githubusercontent.com/4069123/185558932-5962f5c5-b425-4a25-aba1-64ab50f20df0.png">

As you can see, goto definition should always open on the right hand side.

Desired Behaviour: 

1. goto def: `session.send_request` --> open tab on right ( group 1 )
1. goto def : `method_to_capability` --> open tab on right ( group 1 )

Current Behaviour: 

1. goto def: `session.send_request` --> open tab in active group
1. goto def : `method_to_capability` --> open tab in active group

PR #2031, opened 2022-08-24
Add group argument for LspGotoCommand

ninth-dev:
This PR will allow the user to pass a group for `Goto` commands.

When the group is specified:  

- The result will always open in the group.
- If there's more than one result, it will focus to the group and open the `LocationPicker`.

Relates to: #2022

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -198,7 +198,8 @@
     //     "command": "lsp_symbol_type_definition",
     //     "args": {
     //         "side_by_side": false,
-    //         "force_group": true
+    //         "force_group": true,
+    //         "group": -1
     //     },
     //     "keys": [
     //         "UNBOUND"
@@ -221,7 +222,8 @@
     //     "command": "lsp_symbol_declaration",
     //     "args": {
     //         "side_by_side": false,
-    //         "force_group": true
+    //         "force_group": true,
+    //         "group": -1
     //     },
     //     "keys": [
     //         "UNBOUND"
@@ -244,7 +246,8 @@
     //     "command": "lsp_symbol_implementation",
     //     "args": {
     //         "side_by_side": false,
-    //         "force_group": true
+    //         "force_group": true,
+    //         "group": -1
     //     },
     //     "keys": [
     //         "UNBOUND"
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -174,7 +174,8 @@
     //     "args": {
     //         "side_by_side": false,
     //         "force_group": true,
-    //         "fallback": false
+    //         "fallback": false,
+    //         "group": -1
     //     },
     //     "keys": [
     //         "f12"
--- a/docs/src/features.md
+++ b/docs/src/features.md
@@ -34,6 +34,8 @@ In addition to the basic "Goto Definition", the protocol also provides further r
 Additionally, the LSP's "Goto Definition" command can fall back to the built-in Sublime's "Goto Definition" if the `fallback` argument is set to `true`.
 This way, when there are no results found the built-in "Goto Definition" command will be triggered.
 
+To always open the results in a certain group, you can pass the `group` argument.
+
 ## Find References
 
 [Example GIF 1](https://user-images.githubusercontent.com/6579999/128551752-b37fe407-148c-41cf-b1e4-6fe96ed0f77c.gif)
--- a/docs/src/features.md
+++ b/docs/src/features.md
@@ -34,7 +34,7 @@ In addition to the basic "Goto Definition", the protocol also provides further r
 Additionally, the LSP's "Goto Definition" command can fall back to the built-in Sublime's "Goto Definition" if the `fallback` argument is set to `true`.
 This way, when there are no results found the built-in "Goto Definition" command will be triggered.
 
-To always open the results in a certain group, you can use the `group` argument.
+To attempt to open the results in a certain group, you can use the `group` argument. If the specified `group` does not exist, then it will be ignored.
 
 ## Find References
 
--- a/docs/src/features.md
+++ b/docs/src/features.md
@@ -34,7 +34,7 @@ In addition to the basic "Goto Definition", the protocol also provides further r
 Additionally, the LSP's "Goto Definition" command can fall back to the built-in Sublime's "Goto Definition" if the `fallback` argument is set to `true`.
 This way, when there are no results found the built-in "Goto Definition" command will be triggered.
 
-To always open the results in a certain group, you can pass the `group` argument.
+To always open the results in a certain group, you can use the `group` argument.
 
 ## Find References
 
--- a/plugin/core/open.py
+++ b/plugin/core/open.py
@@ -15,15 +15,16 @@ import subprocess
 opening_files = {}  # type: Dict[str, Tuple[Promise[Optional[sublime.View]], ResolveFunc[Optional[sublime.View]]]]
 
 
-def _return_existing_view(flags: int, is_view_in_active_group: bool, specified_group: int) -> bool:
+def _return_existing_view(flags: int, existing_view_group: int, active_group: int, specified_group: int) -> bool:
     # the specified_group is supplied
     if specified_group > -1:
-        return is_view_in_active_group
+        # use existing_view_group if it is in the specified group
+        return existing_view_group == specified_group
     # open side by side
     if bool(flags & (sublime.ADD_TO_SELECTION | sublime.REPLACE_MRU)):
         return False
-    # view is in active group ( no group is specified and not side by side )
-    if is_view_in_active_group:
+    # existing view is in active group ( no group is specified and not side by side )
+    if existing_view_group == active_group:
         return True
     # Jump to the file if sublime.FORCE_GROUP is not set
     return not bool(flags & sublime.FORCE_GROUP)
@@ -41,7 +42,7 @@ def open_file(
     # window.open_file brings the file to focus if it's already opened, which we don't want (unless it's supposed
     # to open as a separate view).
     view = window.find_open_file(file)
-    if view and _return_existing_view(flags, window.get_view_index(view)[0] == window.active_group(), group):
+    if view and _return_existing_view(flags, window.get_view_index(view)[0], window.active_group(), group):
         return Promise.resolve(view)
 
     view = window.open_file(file, flags, group)
--- a/plugin/core/open.py
+++ b/plugin/core/open.py
@@ -15,7 +15,7 @@ import subprocess
 opening_files = {}  # type: Dict[str, Tuple[Promise[Optional[sublime.View]], ResolveFunc[Optional[sublime.View]]]]
 
 
-def _return_existing_view(flags: int, open_file_group: int, active_group: int, specified_group: int ) -> bool:
+def _return_existing_view(flags: int, open_file_group: int, active_group: int, specified_group: int) -> bool:
     open_side_by_side = bool(flags & (sublime.ADD_TO_SELECTION | sublime.REPLACE_MRU))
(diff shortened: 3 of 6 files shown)


plugin/core/open.py
```diff
@@ -15,6 +15,26 @@
...
+    if specified_group > -1:
+        return False
+    if open_side_by_side:
+        # default is -1... and is ignored
+        # always open in the specific group if pass in
+        return False
+    elif file_in_active_group:
+        return True
+    else:
+        return select_file
```
rchl, line 35: I haven't looked at this logic-wise yet but would suggest reformatting like:
```suggestion
def _return_existing_view(flags: int, open_file_group: int, active_group: int, specified_group: int) -> bool:
    open_side_by_side = bool(flags & (sublime.ADD_TO_SELECTION | sublime.REPLACE_MRU))
    file_in_active_group = open_file_group == active_group
    if specified_group > -1:
        return False
    if open_side_by_side:
        # default is -1... and is ignored. always open in the specific group if pass in
        return False
    if file_in_active_group:
        return True
    # true if DON'T want to force it to open in the active group i.e. just jump to the file
    return not bool(flags & sublime.FORCE_GROUP)
```

- no need for `elif` after return
- no need to break short comments into mulitple lines
- we tend to not add unnecessary blank lines within function blocks

Default.sublime-keymap
```diff
@@ -173,7 +173,8 @@
     //     "args": {
     //         "side_by_side": false,
     //         "force_group": true,
-    //         "fallback": false
+    //         "fallback": false,
+    //         "group": -1
```
rchl, line 177: I suppose that other "goto" bindings should also be updated with this argument?

plugin/core/open.py
```diff
@@ -15,6 +15,26 @@
 opening_files = {}  # type: Dict[str, Tuple[Promise[Optional[sublime.View]], ResolveFunc[Optional[sublime.View]]]]
 
 
+def _return_existing_view(flags: int, open_file_group: int, active_group: int, specified_group: int) -> bool:
+    open_side_by_side = bool(flags & (sublime.ADD_TO_SELECTION | sublime.REPLACE_MRU))
+    file_in_active_group = open_file_group == active_group
```
rchl, line 20: Updating `open_file_group` does absolutely nothing here.

plugin/core/open.py
```diff
@@ -27,11 +47,12 @@ def open_file(
...
+
     if view:
-        opens_in_desired_group = not bool(flags & sublime.FORCE_GROUP) or \
-                                 window.get_view_index(view)[0] == window.active_group()
-        opens_in_side_by_side = bool(flags & (sublime.ADD_TO_SELECTION | sublime.REPLACE_MRU))
-        return_existing_view = opens_in_desired_group and not opens_in_side_by_side
+        return_existing_view = _return_existing_view(
+            flags, window.get_view_index(view)[0], window.active_group(), group
+        )
+
```
rchl, line 55: please don't add unnecessary blank lines

plugin/core/open.py
```diff
@@ -15,6 +15,19 @@
 opening_files = {}  # type: Dict[str, Tuple[Promise[Optional[sublime.View]], ResolveFunc[Optional[sublime.View]]]]
 
 
+def _return_existing_view(flags: int, open_file_group: int, active_group: int, specified_group: int) -> bool:
+    if specified_group > -1:
+        return False
```
jwortmann, line 20: What if the file is already open in the specified group? Then it should return `True` I guess. That is what the former `opens_in_desired_group` variable did.

plugin/core/open.py
```diff
@@ -15,6 +15,19 @@
...
+    open_side_by_side = bool(flags & (sublime.ADD_TO_SELECTION | sublime.REPLACE_MRU))
+    if open_side_by_side:
```
jwortmann, line 22: This could be put directly into the if-condition without initializing a new variable for it.

plugin/core/open.py
```diff
@@ -15,6 +15,19 @@
...
+        return False
+    file_in_active_group = open_file_group == active_group
+    if file_in_active_group:
```
jwortmann, line 25: Same here.

plugin/core/open.py
```diff
@@ -15,6 +15,19 @@
...
+        return True
+    # Jump to the file if sublime.FORCE_GROUP is set
```
jwortmann, line 27: It should be "... is *not* set", right?

jwortmann (commented):
It looks like the `group` argument is ignored in https://github.com/sublimelsp/LSP/blob/ee8fc6e16c19fc5478e79d83e281cd6dc18924aa/plugin/core/sessions.py#L1500.

ninth-dev:
> It looks like the `group` argument is ignored in
> 
> https://github.com/sublimelsp/LSP/blob/ee8fc6e16c19fc5478e79d83e281cd6dc18924aa/plugin/core/sessions.py#L1500
> .

H @jwortmann, could you please elaborate what is required to change here ?

jwortmann:
> Hi @jwortmann, could you please elaborate what is required to change here ?

The "Goto" commands call `open_location_async`, which calls `Session.open_location_async`, which calls `Session.open_uri_async`, which might call `Session._open_uri_with_plugin_async` (for some special URI schemes like `deno:`). All these functions pass the `group` argument, but it is not used in the last function. Therein, the `Window.new_file` API function is called, which doesn't accept a `group` argument, but instead uses the currently active group to create the new tab in. But if the server response is only a single location, the location picker isn't used, so the `self._window.focus_group(group)` isn't invoked in that case.

I think it should work if you do the same at
https://github.com/sublimelsp/LSP/blob/ee8fc6e16c19fc5478e79d83e281cd6dc18924aa/plugin/core/sessions.py#L1515-L1516
right before the call to `new_file`.

ninth-dev:
> > Hi @jwortmann, could you please elaborate what is required to change here ?
> 
> The "Goto" commands call `open_location_async`, which calls `Session.open_location_async`, which calls `Session.open_uri_async`, which might call `Session._open_uri_with_plugin_async` (for some special URI schemes like `deno:`). All these functions pass the `group` argument, but it is not used in the last function. Therein, the `Window.new_file` API function is called, which doesn't accept a `group` argument, but instead uses the currently active group to create the new tab in. But if the server response is only a single location, the location picker isn't used, so the `self._window.focus_group(group)` isn't invoked in that case.
> 
> I think it should work if you do the same at
> 
> https://github.com/sublimelsp/LSP/blob/ee8fc6e16c19fc5478e79d83e281cd6dc18924aa/plugin/core/sessions.py#L1515-L1516
> 
> right before the call to `new_file` (possibly only if `group > -1`).

Thanks for detailed explanation, I've added the change.

rchl:
Is there particular reason why `lsp_symbol_references` doesn't support `group`?

plugin/goto.py
```diff
@@ -34,15 +34,20 @@ def run(
         point: Optional[int] = None,
         side_by_side: bool = False,
         force_group: bool = True,
-        fallback: bool = False
+        fallback: bool = False,
+        group: int = -1
```
rchl, line 38: Should also add this argument to `is_enabled` above (even if it works without it).

docs/src/features.md
```diff
@@ -34,6 +34,8 @@ In addition to the basic "Goto Definition", the protocol also provides further r
 Additionally, the LSP's "Goto Definition" command can fall back to the built-in Sublime's "Goto Definition" if the `fallback` argument is set to `true`.
 This way, when there are no results found the built-in "Goto Definition" command will be triggered.
 
+To always open the results in a certain group, you can use the `group` argument.
```
rchl, line 37: It doesn't appear to "always" open in specified group but rather it opens only in specified group if such group exists. So if there is only one group then basically `group` will be ignored.

At least that's how it seems to work from my testing.

plugin/core/open.py
```diff
@@ -15,6 +15,21 @@
 opening_files = {}  # type: Dict[str, Tuple[Promise[Optional[sublime.View]], ResolveFunc[Optional[sublime.View]]]]
 
 
+def _return_existing_view(flags: int, existing_view_group: int, active_group: int, specified_group: int) -> bool:
```
rchl, line 18: Since we don't need `existing_view_group` and `active_group` to be passed here separately, we can just combine those into a single `is_view_in_active_group`.

plugin/core/open.py
```diff
@@ -28,10 +43,9 @@ def open_file(
...
     view = window.find_open_file(file)
     if view:
-        opens_in_desired_group = not bool(flags & sublime.FORCE_GROUP) or \
-                                 window.get_view_index(view)[0] == window.active_group()
-        opens_in_side_by_side = bool(flags & (sublime.ADD_TO_SELECTION | sublime.REPLACE_MRU))
-        return_existing_view = opens_in_desired_group and not opens_in_side_by_side
+        return_existing_view = _return_existing_view(
+            flags, window.get_view_index(view)[0], window.active_group(), group
+        )
         if return_existing_view:
```
rchl, line 49: Redundant variable. Can be just:

```suggestion
        if _return_existing_view(flags, window.get_view_index(view)[0], window.active_group(), group):
```

or

```suggestion
        if _return_existing_view(flags, window.get_view_index(view)[0] == window.active_group(), group):
```

if you follow my other suggestion.

plugin/core/sessions.py
```diff
@@ -1513,6 +1513,8 @@ def _open_uri_with_plugin_async(
             result = Promise.packaged_task()  # type: PackagedTask[Optional[sublime.View]]
 
             def open_scratch_buffer(title: str, content: str, syntax: str) -> None:
+                if group > -1:
+                    self.window.focus_group(group)
```
rchl, line 1517: I'll admit that I'm not that familiar with group handling in ST. Can you explain why this is needed? I thought that `new_file` takes care of focusing new group (unless maybe `FORCE_GROUP` is specified... not clear about how that works.).

rchl: Never mind. `new_file` doesn't accept `group`, only `open_file`.

ninth-dev:
> Is there particular reason why `lsp_symbol_references` doesn't support `group`?

Hi @rchl, no particular reason. 

This was my first PR to the repo and was trying to keep it small and contained :) 

I can follow up with another PR to add group support for `lsp_symbol_reference`.

plugin/core/open.py
```diff
@@ -15,6 +15,20 @@
 opening_files = {}  # type: Dict[str, Tuple[Promise[Optional[sublime.View]], ResolveFunc[Optional[sublime.View]]]]
 
 
+def _return_existing_view(flags: int, is_view_in_active_group: bool, specified_group: int) -> bool:
+    # the specified_group is supplied
+    if specified_group > -1:
+        return is_view_in_active_group
```
jwortmann, line 21: I wonder if this still isn't a logic error here? I think if a group is given, then it should return whether the view is in the specified group, and not whether it's in the active group, no?

ninth-dev: Thanks for pointing this out again @jwortmann, i got myself confused.

plugin/core/open.py
ninth-dev: Turns out we needed `existing_view_group` and `active_group` to be passed here separately.

See: https://github.com/sublimelsp/LSP/pull/2031#discussion_r958132385

rchl (commented):
I had some pending comments that I forgot to send. Sorry.

ninth-dev:
> I had some pending comments that I forgot to send. Sorry.

No worries! I've removed the redundant comments and ensure the sheet is closed when escape is pressed.

rchl (approved):

rchl:
Thanks.

PR #2036, opened 2022-08-29
Fix unescaped special characters in minihtml

### LINKED ISSUE #2034: Ampersands (&) in inlay hints are broken
**Describe the bug**
Ampersands (&) in inlay hints break the hint.

**To Reproduce**
1. Install the latest (dev) LSP (in my case this is commit bd4e0149c40cc53fa1f0925fecda38920f7cedc4)
2. Install [LSP-rust-analyzer](https://github.com/sublimelsp/LSP-rust-analyzer) via package control or manually.
3. Configure LSP-rust-analyzer to use the latest (system) rust-analyser. LSP-rust-analyzer hasn't quite yet got full support for the new inlay hints system. Use the setting `"command": ["rust-analyzer"]`
4. Enable the rust-analyzer engine globally or just for the project
5. Input this test code:
```rust
fn main(){
    let a = 1337;
    
    let b = "deez nuts";

    println!("Haha funny number: {}, and funny words: {}!", a, b);
}
```

6. Note that b is of type &str, and the type hint is completely busted. Hints which do not contain ampersands are not affected.

image

**Environment:**
- OS: Arch Linux
- Sublime Text version: 4126
- LSP version: main branch, commit bd4e0149c40cc53fa1f0925fecda38920f7cedc4
- Language servers used: rust-analyzer

jwortmann:
Make sure special characters like `&` are properly escaped in minihtml content.

For example with LSP-html and hover over the link in the following file:

```html
<!DOCTYPE html>
<html>
<body>
    <img src="https://avatars.githubusercontent.com/u/684879?s=200&v=4" />
</body>
</html>
```

I assume that this might fix #2034, but due to lack of motivation to install Rust + rust-analyzer manually, I have not tested it.
@minerscale If you want, you could give it a try and see if it fixes your issue :)

### DIFF
--- a/plugin/inlay_hint.py
+++ b/plugin/inlay_hint.py
@@ -101,9 +101,9 @@ def get_inlay_hint_html(view: sublime.View, inlay_hint: InlayHint, session: Sess
 
 def format_inlay_hint_tooltip(tooltip: Optional[Union[str, MarkupContent]]) -> str:
     if isinstance(tooltip, str):
-        return tooltip
+        return html.escape(tooltip)
     if isinstance(tooltip, dict):  # MarkupContent
-        return tooltip.get('value') or ""
+        return html.escape(tooltip.get('value') or "")
     return ""
 
 
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -38,6 +38,7 @@ from .core.views import update_lsp_popup
 from .session_view import HOVER_HIGHLIGHT_KEY
 from urllib.parse import unquote, urlparse
 import functools
+import html
 import re
 import sublime
 import webbrowser
@@ -224,7 +225,7 @@ class LspHoverCommand(LspTextCommand):
             title = link.get("tooltip") or "Follow link"
             if title != "Follow link":
                 link_has_standard_tooltip = False
-            contents.append('<a href="{}">{}</a>'.format(target, title))
+            contents.append('<a href="{}">{}</a>'.format(html.escape(target), html.escape(title)))
         if len(contents) > 1:
             link_has_standard_tooltip = False
         link_range = range_to_region(Range.from_lsp(links[0]["range"]), self.view) if links else None


rchl:
> I assume that this might fix #2034, but due to lack of motivation to install Rust + rust-analyzer manually, I have not tested it.

By looking at the changes I would assume that it's not gonna fix that issue as your change only handles links and title elements and the one in that issue is none of those.

jwortmann:
I can't say for sure without the payload from rust-analyzer, but how do you know that it's not the title element in the linked issue that causes the parsing bug? The label is already properly escaped https://github.com/sublimelsp/LSP/blob/bd4e0149c40cc53fa1f0925fecda38920f7cedc4/plugin/inlay_hint.py#L123
so I guess the tooltip / title is the only possible cause that is left.

rchl:
Because tooltip shows up on hovering the phantom, as tooltips do :)

rchl:
And actually tooltip is not rendered using minihtml so shouldn't be escaped.

jwortmann:
But it's already given in the html content when created the phantom. Here:
https://github.com/sublimelsp/LSP/blob/bd4e0149c40cc53fa1f0925fecda38920f7cedc4/plugin/inlay_hint.py#L90-L95

There is probably an error in the console with "parsing error", so that ST doesn't want to render the phantom at all.

rchl:
Sorry, my bad. The `title` is declared through the `title` attribute in minihtml so it does need to be escaped and it fixes that issue (I've checked).

plugin/inlay_hint.py
```diff
@@ -101,9 +101,9 @@ def get_inlay_hint_html(view: sublime.View, inlay_hint: InlayHint, session: Sess
 
 def format_inlay_hint_tooltip(tooltip: Optional[Union[str, MarkupContent]]) -> str:
     if isinstance(tooltip, str):
-        return tooltip
+        return html.escape(tooltip)
     if isinstance(tooltip, dict):  # MarkupContent
-        return tooltip.get('value') or ""
+        return html.escape(tooltip.get('value') or "")
```
rchl, line 106: I would probably move this escape to the location where this content is inserted into HTML as now this code assumes that it will be used in HTML.

rchl: Well, never mind. I see that this way it's more consistent with its sister function `format_inlay_hint_label`.
If I'd want to be picky I could probably rename those to something like `get_inlay_hint_label_html` and `get_inlay_hint_tooltip_html` but I'll let it go already. :)

rchl (approved):

minerscale:
Huzzah! It works great! Thanks very much everyone y'all great.

PR #2037, opened 2022-08-30
Automatically hide the diagnostics panel on save

### LINKED ISSUE #1904: Close the diagnostics panel automatically on save
When I save a file containing an error, the diagnostics panel opens automatically (in accordance to the `show_diagnostics_panel_on_save` setting). Then I fix the error and save, and the diagnostics panel now contains no errors in it (or no diagnostics equal to or less than the `show_diagnostics_panel_on_save` value). But the panel stays open and I have to type Esc to close it. 

One of these two solutions would be great:

* Make the panel automatically close on save, under the same conditions it automatically opens (`show_diagnostics_panel_on_save` setting, even the current name would still fit, I think)
* Add a new `close_diagnostics_panel_on_save` setting, mirroring the behaviour of the existing one, but for closing the panel.

Typing Esc everytime I fix an error :)

tdaniel22:
This PR extends the "show diagnostics panel on save" feature, and hides the panel if the diagnostics level is higher than the `show_diagnostics_panel_on_save` setting.

I took the easy route and and made this the default behaviour, but maybe a better way to do this would be to implement a new option, e.g. `close_diagnostics_panel_on_save`.

I'm no Python or ST3 plugin expert, so I'm not 100% sure this is the correct way of doing this.

Fixes #1904


rchl:
I don't think we need another option for this (well, unless someone inevitably complains at least)  but I would still tie this functionality to the `show_diagnostics_panel_on_save` option. I wouldn't close automatically unless it's enabled.

tdaniel22:
Right, forgot that detail. Now if the `show_diagnostics_panel_on_save` is 0 (disabled), the panel stays visible.

plugin/session_buffer.py
```diff
@@ -322,10 +322,12 @@ def on_post_save_async(self, view: sublime.View, new_uri: DocumentUri) -> None:
...
                 self.session.send_notification(did_save(view, include_text, self.last_known_uri))
-        if self.should_show_diagnostics_panel:
-            mgr = self.session.manager()
-            if mgr:
+        mgr = self.session.manager()
+        if mgr:
+            if self.should_show_diagnostics_panel:
                 mgr.show_diagnostics_panel_async()
+            elif userprefs().show_diagnostics_panel_on_save > 0:
+                mgr.hide_diagnostics_panel_async()
```
rchl, line 330: Actually I've realized now that this logic is not really correct when there is more than one server running on a file. In that case one server can hide the panel while another server might still have relevant diagnostics.

This logic should rather be around here

https://github.com/sublimelsp/LSP/blob/30a8aac1816c02521d96fc5aa2ffdc6a8e0c66b7/plugin/documents.py#L280-L284

and/or

https://github.com/sublimelsp/LSP/blob/30a8aac1816c02521d96fc5aa2ffdc6a8e0c66b7/plugin/documents.py#L369-L385

plugin/documents.py
```diff
@@ -383,6 +383,17 @@ def on_post_save_async(self) -> None:
...
+        if userprefs().show_diagnostics_panel_on_save > 0:
+            self._hide_empty_diagnostics_panels()
+
+    def _hide_empty_diagnostics_panels(self) -> None:
+        change_count = self.view.change_count()
+        for sb in self.session_buffers_async():
+            if sb.diagnostics_version != change_count or not sb.diagnostics:
+                mgr = sb.session.manager()
+                if mgr:
+                    mgr.hide_diagnostics_panel_async()
```
rchl, line 396: We should only hide the panel if none of the diagnostics (which you can retrieve with `self.diagnostics_async()` match the criteria for opening the panel. So the check have to consider diagnostic's severity levels and not just check if there are no diagnostics.

And, correct me if I'm wrong, but I believe that after saving the document the `sb.diagnostics_version != change_count` will pretty much always be true since the change count just changed and there was not enough time to receive new diagnostics. So there should be no reason for checking it.

tdaniel22: Fixed the logic (hide only if none of the diagnostics is visible), checked against the severity level setting, and used `self.diagnostics_async()` instead of reinventing the wheel.

tdaniel22: > Also this has the same issue as previous solution as SessionBuffer of one server can trigger panel close even if other session still has diagnostics.

Just saw your edit, sorry, are all session buffers sharing a single panel or do they have one each?

plugin/documents.py
```diff
@@ -383,6 +383,19 @@ def on_post_save_async(self) -> None:
             # The URI scheme has changed. This means we need to re-determine whether any language servers should
             # be attached to the view.
             sublime.set_timeout(self._reset)
+        # Hide empty diagnostics panels on save if the show_diagnostics_panel_on_save option is enabled.
+        if userprefs().show_diagnostics_panel_on_save > 0:
+            self._hide_empty_diagnostics_panels()
```
rchl, line 388: It's singular:

```suggestion
            self._hide_empty_diagnostic_panels()
```

plugin/documents.py
```diff
@@ -383,6 +383,19 @@ def on_post_save_async(self) -> None:
...
+
+    def _hide_empty_diagnostics_panels(self) -> None:
+        severity_threshold = userprefs().show_diagnostics_severity_level
+        hide_panel = True
+        for sb, diagnostics in self.diagnostics_async():
+            for diagnostic, _ in diagnostics:
+                if diagnostic_severity(diagnostic) <= severity_threshold:
+                    hide_panel = False
```
rchl, line 396: Must only hide the panel if there is no diagnostics that match the requested severity. Not like now where it decides to hide on first diagnostic that it sees that doesn't match severity.

Also the logic should also close the panel if there are no diagnostics at all.

And lastly, this is comparing with the wrong setting. Should compare severity to `show_diagnostics_panel_on_save`, not `show_diagnostics_severity_level`.

plugin/documents.py
rchl: Panel is per window.

plugin/documents.py
rchl: Sorry, the logic was correct already but the preference name was not.

rchl (approved):
Anyone else wants to review?

predragnikolic:
👀 will give it a look

predragnikolic (approved):

PR #2040, opened 2022-09-03
Add icons and isPreferred support for code actions

jwortmann:
- if there are multiple code actions and one of them uses `"isPreferred": true`, then preselect it in the quick panel
- add icons for the quick panel entries if CodeActionKind is provided :)

I'm not sure whether multiple code actions can be "preferred" for a given range at the same time, but I would interpret the description at https://code.visualstudio.com/updates/v1_32#_auto-fix-and-preferred-code-actions in such a way that it should not happen.

Other possible enhancements for a future PR:
- add a `only_preferred` argument for the `lsp_code_actions` command, which will then only consider code actions with `"isPreferred": true`
- sort code actions by CodeActionKind for the quick panel

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -11,6 +11,7 @@ from .core.settings import userprefs
 from .core.typing import Any, List, Dict, Callable, Optional, Tuple, Union, Sequence
 from .core.views import entire_content_region
 from .core.views import first_selection_region
+from .core.views import format_code_actions_for_quick_panel
 from .core.views import text_document_code_action_params
 from .save_command import LspSaveCommand, SaveTask
 import sublime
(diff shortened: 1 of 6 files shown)


plugin/code_actions.py
```diff
@@ -308,10 +309,14 @@ def combine_commands(self) -> 'List[Tuple[str, str, CodeActionOrCommand]]':
 
     def show_code_actions(self) -> None:
         if len(self.commands) > 0:
-            items = [command[1] for command in self.commands]
             window = self.view.window()
             if window:
-                window.show_quick_panel(items, self.handle_select, placeholder="Code action")
+                items, selected_index = format_code_actions_for_quick_panel([command[2] for command in self.commands])
```
rchl, line 314: Seems like the `self.commands` tuple doesn't need the middle (`title`) element now.

rchl (approved):

plugin/code_actions.py
```diff
@@ -299,19 +300,23 @@ def handle_responses_async(self, responses: CodeActionsByConfigName, run_first:
...
-            items = [command[1] for command in self.commands]
             window = self.view.window()
             if window:
-                window.show_quick_panel(items, self.handle_select, placeholder="Code action")
+                items, selected_index = format_code_actions_for_quick_panel([command[1] for command in self.commands])
+                window.show_quick_panel(
+                    items,
+                    self.handle_select,
+                    selected_index=selected_index,
+                    placeholder="Code action")
```
rchl, line 319: Could potentially create a helper function for showing code actions in quick panel as those 3 places are more or less identical.

PR #2045, opened 2022-09-06
Custom context menu in log panel and "Clear log panel" item

predragnikolic:
A small PR that adds the ability to clear the output panel.

image

### DIFF
--- a/Context.sublime-menu
+++ b/Context.sublime-menu
@@ -49,9 +49,5 @@
                 "caption": "Expand Selection"
             }
         ]
-    },
-    {
-        "command": "lsp_clear_log_panel",
-        "caption": "Clear Log Panel"
     }
 ]
--- a/Context.sublime-menu
+++ b/Context.sublime-menu
@@ -52,6 +52,6 @@
     },
     {
         "command": "lsp_clear_log_panel",
-        "caption": "Clear Logs"
+        "caption": "Clear Log Panel"
     }
 ]
--- a/Context.sublime-menu
+++ b/Context.sublime-menu
@@ -49,5 +49,9 @@
                 "caption": "Expand Selection"
             }
         ]
+    },
+    {
+        "command": "lsp_clear_log_panel",
+        "caption": "Clear Logs"
     }
 ]
--- a/boot.py
+++ b/boot.py
@@ -19,7 +19,7 @@ from .plugin.core.panels import destroy_output_panels
 from .plugin.core.panels import LspClearLogPanelCommand
 from .plugin.core.panels import LspClearPanelCommand
 from .plugin.core.panels import LspUpdatePanelCommand
-from .plugin.core.panels import LspUpdateServerPanelCommand
+from .plugin.core.panels import LspUpdateLogPanelCommand
 from .plugin.core.panels import WindowPanelListener
 from .plugin.core.protocol import Error
 from .plugin.core.protocol import Location
--- a/boot.py
+++ b/boot.py
@@ -16,6 +16,7 @@ from .plugin.core.css import load as load_css
 from .plugin.core.logging import exception_log
 from .plugin.core.open import opening_files
 from .plugin.core.panels import destroy_output_panels
+from .plugin.core.panels import LspClearLogPanelCommand
 from .plugin.core.panels import LspClearPanelCommand
 from .plugin.core.panels import LspUpdatePanelCommand
 from .plugin.core.panels import LspUpdateServerPanelCommand
--- a/boot.py
+++ b/boot.py
@@ -13,11 +13,9 @@ from .plugin.configuration import LspEnableLanguageServerGloballyCommand
 from .plugin.configuration import LspEnableLanguageServerInProjectCommand
 from .plugin.core.collections import DottedDict
 from .plugin.core.css import load as load_css
-from .plugin.core.diagnostics import ensure_diagnostics_panel
 from .plugin.core.logging import exception_log
 from .plugin.core.open import opening_files
 from .plugin.core.panels import destroy_output_panels
-from .plugin.core.panels import ensure_server_panel
 from .plugin.core.panels import LspClearLogPanelCommand
 from .plugin.core.panels import LspClearPanelCommand
 from .plugin.core.panels import LspUpdatePanelCommand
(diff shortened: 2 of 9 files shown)


plugin/core/panels.py
```diff
@@ -191,3 +191,15 @@ def run(self, edit: sublime.Edit, window_id: int) -> None:
...
+class LspClearLogPanelCommand(sublime_plugin.WindowCommand):
+    def run(self) -> None:
+        if is_log_panel_open(self.window):
+            panel = ensure_server_panel(self.window)
+            if not panel:
+                return
+            panel.run_command("lsp_clear_panel")
+
+    def is_visible(self) -> bool:
+        return is_log_panel_open(self.window)
```
rchl, line 205: With it being Window Command the menu item appears in every view as long as log panel is open which doesn't sound right.

Is it possible to make it a TextCommand and restrict to the log panel view?

rchl: Even if TextCommand is not an option (not sure how things work with menu items) then we could still check what is the active view here.

plugin/core/panels.py
```diff
@@ -191,3 +191,20 @@ def run(self, edit: sublime.Edit, window_id: int) -> None:
...
+            return
+        panel = ensure_server_panel(window)
+        if panel:
+            panel.run_command("lsp_clear_panel")
+
+    def is_visible(self) -> bool:
+        window = self.view.window()
+        if not window:
+            return False
+        panel = ensure_server_panel(window)
```
rchl, line 209: Maybe not quite right for `is_visible` to create the panel?

To be fair I don't like that the panel is created with some delay. I would be fine with creating it as soon as possible, which this change basically achieves, but would want it more explicit. So you have two options:
 - change this code to not create the panel from `is_visible`
 - ensure that creation happens as soon as possible when LSP package is initializing

(Hopefully others share my opinion that it would be fine to create the panel early. Don't want to create back-and-forth situation here. :))

plugin/core/panels.py
```diff
@@ -191,3 +191,20 @@ def run(self, edit: sublime.Edit, window_id: int) -> None:
...
+        return panel.id() == self.view.id() if panel else False
```
rchl, line 210:
```suggestion
        return bool(panel and panel.id() == self.view.id())
```

Context.sublime-menu
```diff
@@ -49,5 +49,9 @@
                 "caption": "Expand Selection"
             }
         ]
+    },
+    {
+        "command": "lsp_clear_log_panel",
+        "caption": "Clear Logs"
```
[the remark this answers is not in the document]
predragnikolic: I agree, it makes more sense to name the panel "LSP Log Panel". Done with [012017e](https://github.com/sublimelsp/LSP/pull/2045/commits/012017eef5bff81a9ae137223b456ed10b827ba9)

plugin/core/panels.py
predragnikolic: Ok, created all panels on start.
c70552a

But If I open a new ST window, I will still be in a situation like before this commit c70552a

boot.py
```diff
@@ -106,6 +111,10 @@ def plugin_loaded() -> None:
     client_configs.update_configs()
     for window in sublime.windows():
         windows.lookup(window)
+        ensure_server_panel(window)
+        ensure_references_panel(window)
+        ensure_diagnostics_panel(window)
+        ensure_rename_panel(window)
```
rchl, line 117: Think this should happen in `WindowManager` constructor. Should then also handle more cases.

rchl: Actually I'm thinking that references and rename panels shouldn't be created initially.

plugin/core/panels.py
```diff
@@ -144,11 +144,11 @@ def run(self, edit: sublime.Edit, characters: Optional[str] = "") -> None:
 
 
 def ensure_server_panel(window: sublime.Window) -> Optional[sublime.View]:
```
rchl, line 146: should also rename this one and `maybe_update_server_panel`

boot.py
rchl: The diagnostics panel is already created from there (`WindowManager._update_panel_main_thread`).

plugin/core/panels.py
```diff
@@ -69,8 +69,8 @@ def on_window_command(self, window: sublime.Window, command_name: str, args: Dic
             sublime.set_timeout(lambda: self.maybe_update_server_panel(window))
 
     def maybe_update_server_panel(self, window: sublime.Window) -> None:
```
rchl, line 71: `maybe_update_log_panel`

rchl:
Please also search for `Language Servers` and `Language servers` in the project and replace all relevant references.

rchl:
Also https://github.com/sublimelsp/LSP/blob/30a8aac1816c02521d96fc5aa2ffdc6a8e0c66b7/sublime-package.json#L502-L502

rchl (approved):

plugin/core/panels.py
```diff
@@ -183,11 +183,28 @@ def run(self, edit: sublime.Edit, window_id: int) -> None:
...
+        panel = ensure_log_panel(window)
+        if panel:
+            panel.run_command("lsp_clear_panel")
+
+    def is_visible(self) -> bool:
+        window = self.view.window()
+        if not window:
+            return False
+        panel = ensure_log_panel(window)
+        return bool(panel and panel.id() == self.view.id())
```
jwortmann, line 210:
```suggestion
        return self.view == window.find_output_panel(PanelName.Log)
```

Would be simpler and doesn't create the output panel from `is_visible`.

predragnikolic: `ensure_log_panel` will first try to `find_output_panel` before it creates a panel.

```
def ensure_panel(window: sublime.Window, name: str, result_file_regex: str, result_line_regex: str,
                 syntax: str) -> Optional[sublime.View]:
    return window.find_output_panel(name) or create_panel(window, name, result_file_regex, result_line_regex, syntax)
```

rchl: The point is that it will create the panel if it's not open and created yet.
But it's only theoretical issue though as now the panel is created early and always anyway.

rchl (approved):

PR #2046, opened 2022-09-06
Log panel should never be empty

### LINKED ISSUE #1508: The logging panel should never be empty
When opening the logging panel for the first time (without messing with the settings), it is empty. You need to set `log_server: [panel]` in the settings first to have something show up there.

I suggest to automatically enable that setting temporarily when opening the log panel, because it is less confusing for users (and even maintainers) for debugging.

I would also wonder whether `log_stderr` should be enabled by default, but only do something when panel logging is enabled. It depends on whether language servers put a lot of garbage in there, I suppose. If that would be a problem, maybe setting this per language server in `default_clients` would be reasonable.

predragnikolic:
- [make "log_server" default to ["panel"]](https://github.com/sublimelsp/LSP/commit/611da895475299fab71450ad1aabb738f6e82b3a)

- [always enable logs when toggling the log panel](https://github.com/sublimelsp/LSP/commit/043b1ebb99b22c0c68d9419f1420e3a0f11dcbbf)

- [remove the log_debug](https://github.com/sublimelsp/LSP/commit/566a4a6df41060f1d4cc4938dc6e3ee80502f052)

closes #1508

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -172,9 +172,6 @@
   // The maximum number of characters (approximately) before a scrollbar appears.
   "popup_max_characters_height": 1000,
 
-  // Show verbose debug messages in the sublime console.
-  "log_debug": false,
-
   // Log communication from and to language servers.
   // Set to an array of values:
   // - "panel" - log to the LSP Language Servers output panel
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -185,7 +185,7 @@
   // This output panel can be toggled from the command palette with the
   // command "LSP: Toggle Log Panel".
   "log_server": [
-    // "panel",
+    "panel"
     // "remote",
   ],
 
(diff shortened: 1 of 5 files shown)


rchl:
> [always enable logs when toggling the log panel](https://github.com/sublimelsp/LSP/commit/043b1ebb99b22c0c68d9419f1420e3a0f11dcbbf)

Seems quite opinionated? Maybe wouldn't mind if it toggled back to whatever value is in settings on closing the panel.

> [remove the log_debug](https://github.com/sublimelsp/LSP/commit/566a4a6df41060f1d4cc4938dc6e3ee80502f052)

I'm not convinced. I will have to be quick to open the panel (which is not created until some server is started) to see all potential debug messages?

rchl:
Maybe you should handle things one by one with individual PRs as it might be hard to go through with it otherwise.

Would like to hear other's opinions but maybe the server logging could always be disabled by default (remove/deprecate `log_server`) and only possible to enable through the panel's context menu.

predragnikolic:
Yes, I will split this PR into multiple smaller changes :)
we will definitely resolve https://github.com/sublimelsp/LSP/issues/1508, it is just a matter of time 

Also sorry for not taking discussion in this PR, 
I'll try to focus on one PR at a time. :) 
so for now I will close this PR and revisit this topic, when we reduce the number of open PRs.

PR #2047, opened 2022-09-08
Add context menu entry in log panel for toggling lines limit

### LINKED ISSUE #1810: Logging panel is too short
I am trying to debug an issue with `rust-analyzer`. According to [the RA docs](https://rust-analyzer.github.io/manual.html#configuration) I can get some useful information out of it by increasing logging. When I do that, by the time I can open the LSP log panel, the relevant messages are gone.

**To Reproduce**
1. Go to Preferences > Package Settings > LSP > Settings.
2. Add:

    ```json
    {
        "clients":
        {

            "rust-analyzer":
            {
                "command": ["~/.cargo/bin/rustup", "run", "nightly", "rust-analyzer"],
                "enabled": true,
                "env": {
                    "RA_LOG": "rust_analyzer=info",
                }
            },
        },
        "log_max_size": 0,
    }
    ```
    (Note that the `command` will probably have to change depending on how you installed RA.)

3. Restart the language server, or open a Rust project.
4. Open the LSP log panel.
5. Watch as the relevant messages disappear out of the top of the panel.

I am not just talking about scrolling here, scrolling the panel contents up will still hit this limit.

This information is not available via the `remote` log setting.

**Expected behavior**
I would expect to be able to set this somewhere.

**Screenshots**

Here's a GIF! I did not want to capture my actual code, but shortly after the GIF starts, I am restarting the RA language server.

Peek 2021-07-26 13-08

**Environment:**
- OS: Ubuntu 21.04
- Sublime Text version: 4113
- LSP version: 1.7.0
- Language servers used: `rust-analyzer`

rchl:
Toggles the limit per window rather than globally. I've assumed it's better that way as it will "limit the damage" if someone leaves it disabled for a long period of time.

Resolves #1810

### DIFF
--- a/Context LSP Log Panel.sublime-menu
+++ b/Context LSP Log Panel.sublime-menu
@@ -1,6 +1,6 @@
 [
     {
-        "caption": "Clear Log Panel",
+        "caption": "Clear",
         "command": "lsp_clear_log_panel",
     },
     {
--- a/Context LSP Log Panel.sublime-menu
+++ b/Context LSP Log Panel.sublime-menu
@@ -1,7 +1,11 @@
 [
     {
+        "caption": "Clear Log Panel",
         "command": "lsp_clear_log_panel",
-        "caption": "Clear Log Panel"
+    },
+    {
+        "caption": "Limit to 500 Lines",
+        "command": "lsp_toggle_log_panel_lines_limit",
     },
     {
         "caption": "-"
--- a/Context LSP Log Panel.sublime-menu
+++ b/Context LSP Log Panel.sublime-menu
@@ -6,6 +6,7 @@
     {
         "caption": "Limit to 500 Lines",
         "command": "lsp_toggle_log_panel_lines_limit",
+        "checkbox": true
     },
     {
         "caption": "-"
--- a/boot.py
+++ b/boot.py
@@ -18,6 +18,7 @@ from .plugin.core.open import opening_files
 from .plugin.core.panels import destroy_output_panels
 from .plugin.core.panels import LspClearLogPanelCommand
 from .plugin.core.panels import LspClearPanelCommand
+from .plugin.core.panels import LspToggleLogPanelLinesLimitCommand
 from .plugin.core.panels import LspUpdatePanelCommand
 from .plugin.core.panels import LspUpdateLogPanelCommand
 from .plugin.core.panels import WindowPanelListener
--- a/plugin/core/panels.py
+++ b/plugin/core/panels.py
@@ -114,14 +114,39 @@ def create_panel(window: sublime.Window, name: str, result_file_regex: str, resu
     return panel
 
 
+def get_panel(window: Optional[sublime.Window], panel_name: str) -> Optional[sublime.View]:
+    if not window:
+        return None
+    return window.find_output_panel(panel_name)
+
+
 def is_panel_open(window: sublime.Window, panel_name: str) -> bool:
     return window.is_valid() and window.active_panel() == "output.{}".format(panel_name)
 
 
 def ensure_panel(window: sublime.Window, name: str, result_file_regex: str, result_line_regex: str,
                  syntax: str, context_menu: Optional[str] = None) -> Optional[sublime.View]:
-    return window.find_output_panel(name) or \
+    return get_panel(window, name) or \
         create_panel(window, name, result_file_regex, result_line_regex, syntax, context_menu)
+
+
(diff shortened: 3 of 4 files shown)


jwortmann:
So if you disable the limit and forget about it, the then it just keeps growing until you're out of memory?

I guess a better solution in that case would be to write the LSP logs into a log file to disk. Or open an input field to ask the user for a different line limit.

rchl:
The use case is to be able to inspect logs even when server prints a ton of stuff. We can think of alternate solutions but it has to stay simple both from the code complexity and user UI perspective so I think any UI inputs or external files are out of scope.

We could for example have a checkbox that rises the limit to 10000 lines or something. Just not sure how we would communicate in concise way through a menu item.

Context LSP Log Panel.sublime-menu
```diff
@@ -1,7 +1,11 @@
 [
     {
+        "caption": "Clear Log Panel",
         "command": "lsp_clear_log_panel",
-        "caption": "Clear Log Panel"
+    },
+    {
+        "caption": "Limit to 500 Lines",
+        "command": "lsp_toggle_log_panel_lines_limit",
```
predragnikolic, line 8:
```suggestion
        "command": "lsp_toggle_log_panel_lines_limit",
        "checkbox": true
```

```

is_checked() → [bool](https://docs.python.org/3/library/functions.html#bool)

    Return whether a checkbox should be shown next to the menu item. Command arguments are passed as keyword arguments. The .sublime-menu file must have the "checkbox" key set to true for this to be used.
```

rchl: Why it works without it then? :)

predragnikolic: I doesn't work for me on Linux :)

Context LSP Log Panel.sublime-menu
```diff
@@ -1,7 +1,12 @@
 [
     {
+        "caption": "Clear Log Panel",
         "command": "lsp_clear_log_panel",
-        "caption": "Clear Log Panel"
+    },
+    {
+        "caption": "Limit to 500 Lines",
```
predragnikolic, line 7:
```suggestion
        "caption": "Limit To 500 Lines",
```

predragnikolic: My suggested change looks weird, but it is consistent.

predragnikolic: I was thinking of suggesting `"Limit Logs To 500 Lines"`, but the word `Logs` is maybe unnecessary.

predragnikolic (approved):
To me this looks good.

Context LSP Log Panel.sublime-menu
rchl: Here is excerpt from some random site (https://capitalizemytitle.com/):

While the above words are generally capitalized in titles regardless of style, there are some words that are generally not capitalized when using title case. Again, these will depend on the specific style you choose (see [Title Capitalization Rules by Style](https://capitalizemytitle.com/#capitalizationrules) section). These include short words and conjunctions:

- Articles (a, an, the)
- Coordinating Conjunctions (and, but, for)
- Short (fewer than 4 letters)
- [Prepositions](https://capitalizemytitle.com/what-is-a-preposition/) (at, by, to, etc.)

rchl: If you look through ST menu items you'll see that this is often followed but maybe not 100% consistently.

rchl: > I was thinking of suggesting `"Limit Logs To 500 Lines"`, but the word `Logs` is maybe unnecessary.

It's a context menu for the log panel so it would be redundant.

(And to be fair we could also change the "Clear Log Panel" to just "Clear")

predragnikolic: Ok, resolved, there is a convention for that. It does look weird when it is uppercase.
 
+1 for "Clear"

jwortmann:
> We could for example have a checkbox that rises the limit to 10000 lines or something. Just not sure how would we communicate that in a concise way through a menu item.

Then just do it that way, without communicating this limit to the user. 10000 lines should be sufficient and basically equivalent to "no limit" from the user perspective.

This PR right now is a memory leak, disguised as a feature.

rchl:
for the record, rust-analyzer added about 900 lines on init for the `rust-analyzer` project itself.

PR #2049, opened 2022-09-11
Parse position or selection from link fragment

### LINKED ISSUE #2009: Support URI navigation
I would like to be able to click a file URI in hovers and be taken to the correct file and range. This works in vscode, but not in Sublime LSP.

I would like Sublime to parse URI links in the format and on click navigates to the correct place. The Line, Char definition in the fragment is 1-based.
```
file://path/to/file#L34,1-L35,3
file://path/to/file#34,1-35,3
file://path/to/file#L34,1-35,3
file://path/to/file#L34
file://path/to/file#L34,1
```

none

**Additional context**
An example hover: 
image
How vscode navigates and select on click on URI in hover:
image

jwortmann:
Should handle the following formats which are supported by VSCode:
https://github.com/microsoft/vscode/blob/835ace5796cec0ed19a7eec119b26b57220b0f1a/src/vs/platform/opener/common/opener.ts#L153-L160

It applies to Markdown links in the hover content and also to `DocumentLink.target` from the "Document Link" response.

Can be tested for example with LSP-html and the following file:
```html
<!DOCTYPE html>
<html>
<body>
    <h1 id="anchor">Heading</h1>

    <a href="#anchor">link</a>
</body>
</html>
```

Closes #2009.

### DIFF
--- a/plugin/core/open.py
+++ b/plugin/core/open.py
@@ -132,7 +132,7 @@ def center_selection(v: sublime.View, r: RangeLsp) -> sublime.View:
 
 def open_in_browser(uri: str) -> None:
     # NOTE: Remove this check when on py3.8.
-    if not (uri.lower().startswith("http://") or uri.lower().startswith("https://")):
+    if not uri.lower().startswith(("http://", "https://")):
         uri = "https://" + uri
     if not webbrowser.open(uri):
         sublime.status_message("failed to open: " + uri)
--- a/plugin/core/open.py
+++ b/plugin/core/open.py
@@ -2,9 +2,9 @@ from .logging import exception_log
 from .promise import Promise
 from .promise import ResolveFunc
 from .protocol import DocumentUri
-from .protocol import UINT_MAX
 from .protocol import Range
 from .protocol import RangeLsp
+from .protocol import UINT_MAX
 from .typing import Dict, Tuple, Optional
 from .url import parse_uri
 from .views import range_to_region
@@ -26,32 +26,23 @@ def open_file_uri(
 
     def parse_fragment(fragment: str) -> RangeLsp:
         match = FRAGMENT_PATTERN.match(fragment)
+        selection = {'start': {'line': 0, 'character': 0}, 'end': {'line': 0, 'character': 0}}  # type: RangeLsp
         if match:
-            # assume that line and column numbers in the fragment are 1-based
-            start_line, start_column, end_line, end_column = [max(1, int(g)) if g else None for g in match.groups()]
+            # Line and column numbers in the fragment are assumed to be 1-based and need to be converted to 0-based
+            # numbers for the LSP Position structure.
+            start_line, start_column, end_line, end_column = [max(0, int(g) - 1) if g else None for g in match.groups()]
             if start_line is not None:
-                if end_line is not None:
-                    if start_column is not None and end_column is not None:
-                        return {
-                            "start": {"line": start_line - 1, "character": start_column - 1},
-                            "end": {"line": end_line - 1, "character": end_column - 1}
-                        }
-                    else:
-                        return {
-                            "start": {"line": start_line - 1, "character": 0},
-                            "end": {"line": end_line - 1, "character": UINT_MAX}
-                        }
-                elif start_column is not None:
-                    return {
-                        "start": {"line": start_line - 1, "character": start_column - 1},
-                        "end": {"line": start_line - 1, "character": start_column - 1}
-                    }
-                else:
-                    return {
-                        "start": {"line": start_line - 1, "character": 0},
-                        "end": {"line": start_line - 1, "character": 0}
-                    }
-        return {"start": {"line": 0, "character": 0}, "end": {"line": 0, "character": 0}}
+                selection['start']['line'] = start_line
+                selection['end']['line'] = start_line
+            if start_column is not None:
+                selection['start']['character'] = start_column
+                selection['end']['character'] = start_column
+            if end_line is not None:
+                selection['end']['line'] = end_line
+                selection['end']['character'] = UINT_MAX
+            if end_column is not None:
+                selection['end']['character'] = end_column
+        return selection
 
     decoded_uri = unquote(uri)  # decode percent-encoded characters
     parsed = urlparse(decoded_uri)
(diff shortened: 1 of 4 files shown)


plugin/core/open.py
```diff
@@ -2,17 +2,71 @@
...
+
+        def handle_continuation(view: Optional[sublime.View]) -> Promise[Optional[sublime.View]]:
+            if view:
+                center_selection(view, r)
+                return Promise.resolve(view)
+            return Promise.resolve(None)
+
+        return open_file(window, decoded_uri, flags, group).then(handle_continuation)
+    else:
+        return open_file(window, decoded_uri, flags, group)
```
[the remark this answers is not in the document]
jwortmann: Thanks, added the second alternative and renamed to `_select_and_center`.

rchl (approved):

plugin/core/open.py
```diff
@@ -2,17 +2,69 @@
...
+        if match:
+            # assume that line and column numbers in the fragment are 1-based
+            start_line, start_column, end_line, end_column = [max(1, int(g)) if g else None for g in match.groups()]
+            if start_line is not None:
+                if end_line is not None:
+                    if start_column is not None and end_column is not None:
+                        return {
+                            "start": {"line": start_line - 1, "character": start_column - 1},
+                            "end": {"line": end_line - 1, "character": end_column - 1}
+                        }
```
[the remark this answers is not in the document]
jwortmann: `Range` has a method `to_lsp()` which would produce the correct type for `center_selection`. But I'm not really convinced that it would look better that way. We would basically create two new Point objects and a Range object, just to immediately convert it into a RangeLsp dict again. Seems a bit inefficient and needless to me.

But in general, if I were asked to redesign the type definitions, I would rename `RangeLsp` to `Range`, remove the current `Range` class altogether, and where necessary use `sublime.Region` together with correspondingly updated functions `region_to_range` and `range_to_region` which will then operate on LSP's "Range" dicts. We usually deal with `sublime.Region`s anyway when when we interact with the View. And currently we always have to use that additional and seemingly unnecessary step to create a `Range` object together with `Range.from_lsp(...)` or `to_lsp()` when converting from or to LSP's data structures.

plugin/core/open.py
```diff
@@ -78,6 +130,14 @@ def center_selection(v: sublime.View, r: RangeLsp) -> sublime.View:
     return v
 
 
+def open_in_browser(uri: str) -> None:
+    # NOTE: Remove this check when on py3.8.
+    if not (uri.lower().startswith("http://") or uri.lower().startswith("https://")):
+        uri = "http://" + uri
```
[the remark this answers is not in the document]
jwortmann: Yes, seems more useful.

plugin/core/open.py
```diff
@@ -2,17 +2,69 @@
 from .promise import Promise
 from .promise import ResolveFunc
 from .protocol import DocumentUri
+from .protocol import UINT_MAX
```
predragnikolic, line 5: sort imports 

sorry Rafal, for stealing your comment :D

plugin/core/open.py
```diff
@@ -2,17 +2,69 @@
...
+                    return {
+                        "start": {"line": start_line - 1, "character": start_column - 1},
+                        "end": {"line": start_line - 1, "character": start_column - 1}
+                    }
+                else:
+                    return {
+                        "start": {"line": start_line - 1, "character": 0},
+                        "end": {"line": start_line - 1, "character": 0}
+                    }
+        return {"start": {"line": 0, "character": 0}, "end": {"line": 0, "character": 0}}
```
predragnikolic, line 54: to reduce the indent, how about something like this: 
```py
    def parse_fragment(fragment: str) -> RangeLsp:
        match = FRAGMENT_PATTERN.match(fragment)
        result = {"start": {"line": 0, "character": 0}, "end": {"line": 0, "character": 0}}  # type: RangeLsp
        if match:
            # assume that line and column numbers in the fragment are 1-based
            start_line, start_column, end_line, end_column = [max(1, int(g)) if g else None for g in match.groups()]
            if start_line:
                result['start']['line'] = start_line - 1
            if start_column:
                result['start']['character'] = start_column - 1
                result['end']['character'] = start_column - 1
            if end_line:
                result['end']['line'] = end_line - 1
            if end_column:
                result['end']['character'] = end_column - 1
        return result
```

jwortmann: Added with a few adjustments. It requires 1-2 more if-checks than before and possibly several overrides of the result dict, but that shouldn't really matter and it appears visually more pleasant this way. Should be equivalent to the replaced code now.

rchl: Doesn't the lines still have to be compared against `None`? I'm looking at this on mobile so maybe I'm missing something but line 0 would evaluate to false.

rchl (approved):

plugin/core/open.py
jwortmann: > Doesn't the lines still have to be compared against None? I'm looking at this on mobile so maybe I'm missing something but line 0 would evaluate to false.

Initially I thought that as well, but 0 is already the default for all of the values, so it wouldn't make any difference to overwrite it with 0 again.

I think there are only two differences to the `is not None` check, that could theoretically occur:
* `end_line` is 0, but `start_line` is not -> I would assume that for valid ranges, "end" should come after "start", so this should not hopefully not happen (and if it happens, with the current code we would ignore the assumingly faulty "end" to fix it)
* the range spans several lines, and at the same time the `end_column` is set to 0. In other words, it should include the newline character. Yeah, I guess this might be a valid usecase, so I better adjust the if-check for that.

plugin/core/open.py
```diff
@@ -4,15 +4,58 @@
...
+                selection['end']['line'] = start_line
+            if start_column:
+                selection['start']['character'] = start_column
+                selection['end']['character'] = start_column
+            if end_line:
+                selection['end']['line'] = end_line
+                selection['end']['character'] = UINT_MAX
+            if end_column is not None:
+                selection['end']['character'] = end_column
+        return selection
```
rchl, line 45: Does it really make sense to default to 0,0 position when fragment doesn't match the pattern? I would think that we should just ignore it then.

rchl: Or is it on purpose to create more consistency (always scroll to the top) when opening files?

jwortmann: I guess this question is for when the file is already open? Actually I have not considered this. Probably it makes sense indeed to ignore the fragment if it doesn't match the pattern, i.e. keep the current scroll position.

jwortmann: https://github.com/sublimelsp/LSP/pull/2049/commits/3e9362742615a3dcbfcba6c81f71e4dbb364d1ff

rchl (approved):

PR #2050, opened 2022-09-13
Add json suggestions for disabled capabilities

predragnikolic:
I never know what to put in the `disabled_capabilities`,
always have to look up the spec.

With this PR we will get JSON schema validation for `disabled_capabilities`.

### DIFF
--- a/sublime-package.json
+++ b/sublime-package.json
@@ -128,8 +128,25 @@
                   "default": true
                 },
                 "signatureHelpProvider": {
-                  "type": "boolean",
-                  "default": true
+                  "oneOf": [
+                    {
+                      "type": "object",
+                      "properties": {
+                        "triggerCharacters": {
+                          "type": "boolean",
+                          "default": true
+                        },
+                        "retriggerCharacters": {
+                          "type": "boolean",
+                          "default": true
+                        }
+                      }
+                    },
+                    {
+                      "type": "boolean",
+                      "default": true
+                    }
+                  ]
                 },
                 "declarationProvider": {
                   "type": "boolean",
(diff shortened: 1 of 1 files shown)


sublime-package.json
```diff
@@ -103,7 +103,177 @@
...
+              "markdownDescription": "A dictionary that \"masks\" (disables) the capabilities registered by this language server. For instance, to disable completions, use `{\"completionProvider\": true}`. To disable hover info, use `{\"hoverProvider\": true}`. Multiple providers can be disabled. To disable a subset of a provider, use a nested dictionary. For intance, to disable only the trigger characters of a language server, but not the completion functionality itself, use `{\"completionProvider\": {\"triggerCharacters\": true}}`.",
+              "additionalProperties": false,
```
rwols, line 107: Maybe allow additional properties for future protocol updates?

sublime-package.json
```diff
@@ -103,7 +103,177 @@
...
+                "documentSymbolProvider": {
+                  "type": "boolean",
+                  "default": true
+                },
+                "codeActionProvider": {
+                  "type": "boolean",
+                  "default": true
+                },
+                "codeLensProvider": {
+                  "type": "boolean",
```
rchl, line 184: can you sort :)

sublime-package.json
```diff
@@ -105,6 +105,22 @@
...
+                  "default": true
+                },
+                "colorProvider": {
+                  "type": "boolean",
+                  "default": true
+                },
```
rchl, line 123: I suppose all (or most, if not all) capabilities should also allow an object form, even if we don't specify the exact properties in the schema.

rchl:
I've added a `$ref` to avoid repeating the same block.
Have not investigating whether I can merge default with non-trivial cases - thought it's not worth the effort.

rchl (approved):

PR #2051, opened 2022-09-13
Change log_server default value to  ["panel"]

predragnikolic:
Currently if the user want to enable logging, 
the following settings have to be set:
```
"log_debug": true,
"log_server": ["panel"],
```

----

With this PR the user just have to enable:
```
"log_debug": true,
```
So we skip one additional thing that needs to be enabled.

----

EDIT:

closes https://github.com/sublimelsp/LSP/issues/1508

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -193,7 +193,7 @@
   // This output panel can be toggled from the command palette with the
   // command "LSP: Toggle Log Panel".
   "log_server": [
-    // "panel",
+    "panel",
     // "remote",
   ],
 
--- a/docs/src/troubleshooting.md
+++ b/docs/src/troubleshooting.md
@@ -5,7 +5,6 @@ To get more visibility into the inner-workings of the LSP client and the server 
 | Option                  | Description                                                          |
 | ----------------------- | -------------------------------------------------------------------- |
 | `log_debug: true`       | Show verbose debug messages in the Sublime Text console.             |
-| `log_server: ["panel"]` | Log communication from and to language servers in the output panel.  |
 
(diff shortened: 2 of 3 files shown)


rwols:
I thought log_debug was only for the Console?

rchl:
> I thought log_debug was only for the Console?

I suppose @predragnikolic meant the case when one is following the steps at https://lsp.sublimetext.io/troubleshooting/#self-help-instructions

rchl (commented):
I'm fine with that.

We are not writing to the panel unless it's open so there will be no performance impact from that. We'll use a bit more memory to store last 500 log lines and and do some list operations to keep it at 500 items but that should be negligible.

rchl:
Could also set `"default": ["panel"]` in schema.

rwols:
Well, I’m wondering whether Sublime Text eagerly renders hidden syntax-highlighted panels, or whether it lazily renders hidden syntax-highlighted panels. If it eagerly renders these panels, then, assuming that the majority of users never enable this panel, this would incur some runtime overhead. In this case, I would be against enabling this setting. If instead ST renders hidden panels lazily, then I’m all for this.

rchl:
In my comment above I've mentioned that the panel is only mutated by LSP when it's actually opened though.

rwols:
Ah yeah, I missed that. Also I remember that now. Then no complaints here 🙂

predragnikolic:
https://github.com/sublimelsp/LSP/pull/2051#pullrequestreview-1105991246

Yeah this comment sums up the behavior introduced in this PR,
could have written it in the PR description :)


> Could also set "default": ["panel"] in schema.

That was already done.

rchl (approved):

PR #2052, opened 2022-09-13
Make `Preferences: LSP Key Bindings` open the default keybinding file

predragnikolic:
Currently, my user keybindings are stored in  `Default.sublime-keymap`.

Note, the `platform` is not in the file name (`Default (Linux).sublime-keymap`, `Default (Windows).sublime-keymap`, `Default (OSX).sublime-keymap`) 

Currently,
Each time I run  `Preferences: LSP Key Bindings`, 
ST will open an empty `Default (Linux).sublime-keymap`, because the incorrect "user_file" was specified (I store my keybindings in `Default.sublime-keymap`).

image

With this PR,
ST will open the correct user defined keymap file.

image

### DIFF
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -12,7 +12,6 @@
         "command": "edit_settings",
         "args": {
             "base_file": "${packages}/LSP/Default.sublime-keymap",
-            "user_file": "${packages}/User/Default ($platform).sublime-keymap",
             "default": "[\n\t$0\n]\n",
         }
     },
--- a/Main.sublime-menu
+++ b/Main.sublime-menu
@@ -159,7 +159,6 @@
                 "command": "edit_settings",
                 "args": {
                   "base_file": "${packages}/LSP/Default.sublime-keymap",
-                  "user_file": "${packages}/User/Default (${platform}).sublime-keymap",
                   "default": "[\n\t$0\n]\n",
                 }
               },


rchl:
I imagine most people store their key bindings in the files that include the platform as that's the default thing that opens when you use the `Preferences: Key Bindings`.
So this would make things better for you while probably making it incorrect (or at least worse) for majority.

predragnikolic:
>  So this would make things better for you while probably making it incorrect (or at least worse) for majority.

No.

I run a test,
If people have `Default (Linux).sublime-keymap` (and not `Default.sublime-keymap`), 
even with this change, ST will open `Default (Linux).sublime-keymap`.

rchl:
OK, so I have both files so this change would affect me and potentially others also.

And the default `Preferences: Key Bindings` command doesn't behave that way either as it opens the file with platform in the name.

jwortmann:
> And the default `Preferences: Key Bindings` command doesn't behave that way either as it opens the file with platform in the name.

But the reason for that is because the file with the default key bindings already includes the platform name. And if you don't have a similarly named file in your user package, but a `Default.sublime-keymap` file instead, then it will actually open that one instead of trying to create a new file.

If you have both of the files present, then it seems ambiguous which file should be opened by LSP. Actually it is more consistent to open the `Default.sublime-keymap` rather than the platform-specific variant, because file which is shipped in LSP is also named this way and doesn't have the platform specific suffix (iirc this was different some time ago but changed to the current filename).

So I would be in favor of this change, unless you can find an actual example where it would worsen the current behavior.

This is how the command definition in ST's `Default.sublime-commands` looks like:
```json
	{
		"caption": "Preferences: Key Bindings",
		"command": "edit_settings", "args":
		{
			"base_file": "${packages}/Default/Default ($platform).sublime-keymap",
			"default": "[\n\t$0\n]\n"
		}
	}
```

jwortmann:
This change should then also be applied to the menu file

https://github.com/sublimelsp/LSP/blob/5926089d17c4092bd22f8ef323a4eacadf02dfb0/Main.sublime-menu#L157-L165

rchl:
> And if you don't have a similarly named file in your user package, but a `Default.sublime-keymap` file instead, then it will actually open that one instead of trying to create a new file.

Maybe so but `Default.sublime-keymap` doesn't exist by default so most users will just see the platform-specific variant of the user bindings and will put their keybindings there. Now, until this PR, LSP was opening the same file so it was consistent for most users (those that don't go out of their way to manually create `Default.sublime-keymap`) while with this change those people will suddenly see an empty keybidnings file open.

predragnikolic:
I wasn't right, 
this is a not backwards compatible change, it introduces a change in behavior.

Here is the comparison of the previous behavior and the behavior in this PR:

**Before:**
1. User has `Default (Linux).sublime-keymap` file.
Preferences: LSP Key Bindings will open `Default (Linux).sublime-keymap`.

2. User has `Default.sublime-keymap` file.
Preferences: LSP Key Bindings will open a `Default (Linux).sublime-keymap`. (which I find incorrect in my case)

3. User had both `Default.sublime-keymap` and `Default (Linux).sublime-keymap` files.
Preferences: LSP Key Bindings will open `Default (Linux).sublime-keymap`.

**After PR:**
1. User has `Default (Linux).sublime-keymap` file.
Preferences: LSP Key Bindings will still open `Default (Linux).sublime-keymap`.

2. User has `Default.sublime-keymap` file.
Preferences: LSP Key Bindings will open `Default.sublime-keymap`. (THIS CHANGED)

3. User had both `Default.sublime-keymap` and `Default (Linux).sublime-keymap` files.
Preferences: LSP Key Bindings will open `Default.sublime-keymap`. (THIS CHANGED)

Case 2 and 3 have changed behaviors. 

----

I wanted to fix the `case 2`, 
`case 3` is a (negative) consequence of fixing `case 2`.

predragnikolic:
Ok I'm not for this PR because it introduces a change in behavior, more that I expected.

jwortmann:
Assuming that the description in https://github.com/sublimelsp/LSP/pull/2052#issuecomment-1245869654 is correct (I haven't tested myself), then I would say
- 1\. is a negative and unacceptable change from this PR
- 2\. is better with this PR
- 3\. is also better with this PR in my opinion, because it is more consistent to open the user file with the same name (but it is probably a rare usecase and not very important)

I looked into the Default/settings.py file and the following part for the `edit_settings` command seems to be the reason why it works in one direction (open non-platform file if platform-specific file doesn't exist), but not in the other direction:
```python
        if user_file is None:
            user_package_path = os.path.join(sublime.packages_path(), 'User')
            user_file = os.path.join(user_package_path, file_name)

            # If the user path does not exist, and it is a supported
            # platform-variant file path, then try and non-platform-variant
            # file path.
            if not os.path.exists(os.path.join(user_package_path, file_name)):
                for suffix in {'.sublime-keymap', '.sublime-mousemap', '.sublime-menu'}:
                    platform_suffix = ' (%s)%s' % (platform_name, suffix)
                    if not file_name.endswith(platform_suffix):
                        continue
                    non_platform_file_name = file_name[:-len(platform_suffix)] + suffix
                    non_platform_path = os.path.join(user_package_path, non_platform_file_name)
                    if os.path.exists(non_platform_path):
                        user_file = non_platform_path
                        break
```

So if we would revert https://github.com/sublimelsp/LSP/commit/5e7e659608975ccfa4a4be05db058069c12eb127 and remove the `user_file` argument, then case 2. could be fixed.

PR #2054, opened 2022-09-14
Only update content of diagnostics panel when visible

jwortmann:
This makes the console command `sublime.log_commands(True)` usable again if you type in a view which has diagnostics.

### DIFF
--- a/docs/src/keyboard_shortcuts.md
+++ b/docs/src/keyboard_shortcuts.md
@@ -23,8 +23,8 @@ Refer to the [Customization section](customization.md#keyboard-shortcuts-key-bin
 | Goto Type Definition | unbound | `lsp_symbol_type_definition`
 | Hover Popup | unbound | `lsp_hover`
 | Insert/Replace Completions | <kbd>alt</kbd> <kbd>enter</kbd> | `lsp_commit_completion_with_opposite_insert_mode`
-| Next Diagnostic | <kbd>F4</kbd> | -
-| Previous Diagnostic | <kbd>shift</kbd> <kbd>F4</kbd> | -
+| Next Diagnostic | unbound | `lsp_next_diagnostic`
+| Previous Diagnostic | unbound | `lsp_prev_diagnostic`
 | Rename | unbound | `lsp_symbol_rename`
 | Restart Server | unbound | `lsp_restart_server`
 | Run Code Action | unbound | `lsp_code_actions`
--- a/docs/src/keyboard_shortcuts.md
+++ b/docs/src/keyboard_shortcuts.md
@@ -5,9 +5,6 @@ Refer to the [Customization section](customization.md#keyboard-shortcuts-key-bin
 !!! Mac
     If you using macOS, replace <kbd>ctrl</kbd> with <kbd>command</kbd>.
 
-!!! Note
-    If <kbd>F4</kbd> / <kbd>shift</kbd> <kbd>F4</kbd> stop working, double-click a diagnostic in the Diagnostics Panel. It should work again afterwards.
-
 | Feature | Shortcut | Command |
 | ------- | -------- | ------- |
 | Auto Complete | <kbd>ctrl</kbd> <kbd>space</kbd> (also on macOS) | `auto_complete`
--- a/plugin/core/registry.py
+++ b/plugin/core/registry.py
@@ -1,9 +1,15 @@
 from .configurations import ConfigManager
+from .protocol import Diagnostic
+from .protocol import Point
 from .sessions import AbstractViewListener
 from .sessions import Session
 from .settings import client_configs
-from .typing import Optional, Any, Generator, Iterable
+from .typing import Optional, Any, Generator, Iterable, List
+from .views import MissingUriError
+from .views import point_to_offset
+from .views import uri_from_view
 from .windows import WindowRegistry
+import operator
 import sublime
 import sublime_plugin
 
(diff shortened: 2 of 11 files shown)


jwortmann:
Ah, I didn't know that LSP (ab)uses the next/previous built result commands to jump between diagnostics.

I guess then the intention of this PR is impossible to implement...

rchl:
> So shouldn't LSP better use it's own commands `lsp_next_diagnostic` & `lsp_prev_diagnostic` for that? Then it could be implemented without needing the diagnostics panel.

I'm very much for a custom handling navigation for diagnostics.
We had a (pretty complex) code for that before that we could try to bring back. Or rewrite it from scratch...

jwortmann:
I think this is what's going on:
The `next_result` & `prev_result` commands are quite overloaded in a way that they always navigate between the results of the particular panel which had the last focus:
* Build Results panel (ST builtin)
* Find Results panel (ST builtin)
* Output: References (LSP panel)
* Output: Diagnostics (LSP panel)
* maybe more from other packages...?

Even after you close the panel, the commands will still do their job, with the positions of the last focused panel.
If multiple panels are available, you can change the navigation behavior by opening another panel, and clicking into that panel to focus it (see https://github.com/sublimelsp/LSP/issues/1721#issuecomment-950272718, but in fact a single click is sufficient, double clicking is not necessary).

With this PR, opening the diagnostics panel would become necessary, before the commands an key bindings to navigate the diagnostics begin to work, so I don't know if that would be acceptable. And another (perhaps more important) problem is, that if you close the panel and edit the file with diagnostics, then the diagnostics which are still listed in the panel won't get updated and if you then use the key bindings to navigate results, it will likely jump to outdated positions. So I guess the diagnostics panel would have to be cleared to disable the result navigation when the panel is not visible.

But there are other issues for the current key bindings too (https://github.com/sublimelsp/LSP/issues/1721#issuecomment-951836215). So I'm unsure if using these commands is really appropriate for the diagnostics, or whether it shouldn't indeed be rewritten with custom commands, which would navigate only between diagnostics in the current file.

By the way, I think the linked issue #1721 is not fixed at all, but I would say that the "problem" described therein is actually intended behavior for the `next_result` / `prev_result` commands.

(Another aspect is that the first 3 panels from the list above are triggered manually, while the diagnostics are received from the language server in the background. So maybe it is kind of expected that navigation should only work if the panel is visible?)

@rwols what is your point of view about this?

---

Tests are still failing though for some reason. But for me even on the `main` branch popups are not shown reliably when navigating between diagnostics with <kbd>F4</kbd> / <kbd>Shift</kbd>+<kbd>F4</kbd> (and it seems to require focusing the diagnostics panel as well, before those keybindings begin to work (?)).

rchl:
I'm fine with F4 not working when the panel is not open.

Eventually we need to create custom commands for navigating diagnostics but for now this is just a broken feature with no guarantee of working.

rchl (approved):
Great stuff. 👍

jwortmann:
I found out why the popups sometimes don't show up (also on `main` branch). ST seems to immediately hide the popup for being out of bounds of the currently visible viewport when using the scrolling animation. So I added a small delay (200ms) and the popups seem to work reliably now.

It would also work without needing the delay if we use `animate=False` for `show_at_center`, but I like the animation :)

plugin/core/registry.py
```diff
@@ -161,3 +168,49 @@ def run_async() -> None:
...
+    if not diagnostics:
+        return
+    # Sort diagnostics by location
+    diagnostics.sort(key=lambda d: operator.itemgetter('line', 'character')(d['range']['start']), reverse=not forward)
+    if point is None:
+        region = first_selection_region(view)
+        point = region.b if region is not None else 0
+    # Find next/previous diagnostic or wrap around and jump to the first/last one, if there are no more diagnostics in
+    # this view after/before the cursor
+    op_func = operator.gt if forward else operator.lt
```
predragnikolic, line 193: This is subjective. (if others do not have problems with following this code, skip this comment)

This code is a little bit harder to follow
do you see a way to simplify this:

```
op_func = operator.gt if forward else operator.lt
for diagnostic in diagnostics:
        diag_pos = point_to_offset(Point.from_lsp(diagnostic['range']['start']), view)
        if op_func(diag_pos, point):
            break
```

docs/src/keyboard_shortcuts.md
```diff
@@ -23,8 +23,8 @@ Refer to the [Customization section](customization.md#keyboard-shortcuts-key-bin
 | Goto Type Definition | unbound | `lsp_symbol_type_definition`
 | Hover Popup | unbound | `lsp_hover`
 | Insert/Replace Completions | <kbd>alt</kbd> <kbd>enter</kbd> | `lsp_commit_completion_with_opposite_insert_mode`
-| Next Diagnostic | <kbd>F4</kbd> | -
-| Previous Diagnostic | <kbd>shift</kbd> <kbd>F4</kbd> | -
+| Next Diagnostic | unbound | `lsp_next_diagnostic`
+| Previous Diagnostic | unbound | `lsp_prev_diagnostic`
```
predragnikolic, line 27: this note could be updated
LSP/docs/src/keyboard_shortcuts.md

```
!!! Note
    If <kbd>F4</kbd> / <kbd>shift</kbd> <kbd>F4</kbd> stop working, double-click a diagnostic in the Diagnostics Panel. It should work again afterwards.
```

plugin/core/registry.py
predragnikolic: Ok, It was a little bit hard for me because this is the first time seeing a for/else syntax and the operator module :) 
but I am ok with the code after staring at it for a little bit longer. 

And it makes sense why you did it in this way.

docs/src/keyboard_shortcuts.md
jwortmann: I removed this note now.

plugin/core/registry.py
jwortmann: It's just a different way of writing

```python
for diagnostic in diagnostics:
    diag_pos = point_to_offset(Point.from_lsp(diagnostic['range']['start']), view)
    if forward:
        if diag_pos > point:
            break
    else:
        if diag_pos < point:
            break
```

but with fewer lines and indentation, and it allows to move one "if" to outside of the loop.

The for/else syntax is actually one of my favorite aspects of Python and I use it whereever I can :)

rchl:
Looks like this broke `show_diagnostics_count_in_view_status` but don't worry about it, I'm doing some work around that so will fix.

PR #2056, opened 2022-09-14
WIP: Investigate code lens refresh issues

rchl:
I've noticed that this weird thing is happening with code lens annotations on moving the cursor:

https://user-images.githubusercontent.com/153197/190275044-6db9ace3-c2c1-4097-a741-eefe14cb14ba.mov

- there is no relevant communication with the server on moving the cursor
- we are just redrawing the code lenses that we already have cached
- we are calling `view.add_regions` without erasing the regions first. I guess the assumption is that ST handles duplicate region keys itself and it will result in a no-op but it seems that this issue is more apparent when not erasing explicitly
- even with erasing and using more unique region key (to avoid conflicts between sessions), the issue still reproduces so it seems to be an ST bug.

I will create an ST issue when I have more time.

### DIFF
--- a/plugin/code_lens.py
+++ b/plugin/code_lens.py
@@ -86,11 +86,11 @@ class CodeLensView:
         return self._init
 
     def _clear_annotations(self) -> None:
-        for index, _ in enumerate(self._flat_iteration()):
-            self.view.erase_regions(self._region_key(index))
+        for index, lens in enumerate(self._flat_iteration()):
+            self.view.erase_regions(self._region_key(lens.session_name, index))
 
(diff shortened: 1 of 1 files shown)


rchl:
Created https://github.com/sublimehq/sublime_text/issues/5566

rchl:
The core issue is fixed in 4144. I think part of this still makes sense to do to avoid code lenses from multiple sessions conflicting with each other. Although that's pretty much a corner case I imagine.

rchl:
I think it makes sense to add all those changes anyway.

rchl:
Should be fine :)

PR #2062, opened 2022-09-20
refactor(types): Mark properties with NotRequired

rchl:
Make use of auto-generated  types from https://github.com/sublimelsp/lsp-python-types that utilize [`NotRequired`](https://peps.python.org/pep-0655/) for properties that might not be present.

https://github.com/sublimelsp/lsp-python-types generates two separate files - `lsp_types.py` and `lsp_types_sublime_text_33.py`. Former uses new class syntax that is documentation-annotations-friendly but is not compatible with Python 3.3. Latter doesn't use the new class syntax. We have to use the latter which means that documentation for properties won't be picked up by pyright but otherwise there are no downsides.

Even though ST's 3.3 host doesn't support `NotRequired` (or even `typing` for that matter), that doesn't matter because for 3.3 the existing LSP code has a dummy type aliases that do nothing anyway. The important part is that type checkers like `mypy` and `pyright` type-check using Python version that supports those types.

Some notes about changes:
 - Our `WorkspaceFolder` class conflicted with same-named LSP type. Moved our class to `workspace.py`. Since it's a public export, plugins that imported it from `LSP.plugin` will be unaffected. The few that imported from `LSP.core.protocol` will be updated and I took care of that.
 - Mypy is not compatible with those generated types and errors with `Cannot resolve name "X" (possible cyclic definition)`. This is seemingly [already supported](https://github.com/python/mypy/issues/731#issuecomment-1213482527) on `master` (as an opt-in) but not released yet. For now mypy is disabled due to that. I'm not sure but I don't think we can ignore type errors for specific file as if mypy can't process those types then I don't think it would be able to type check anyway.
 - Fixed various type issues with `Promise`.
 - Enabled type checking using `pyright` in `tox.ini` and CI and fixed ~~almost~~ all errors that it reported. ~~The remaining ones are the stupid cases like `view = None  # type: sublime.View` that I think we shouldn't be doing as that looks and feels wrong. Haven't investigated how we could fix those yet. Potentially we can just ignore those.~~
 - I'm sure that there are places in the existing code that could be updated to use more of the generated types but that's something that can be done later, while making changes in relevant code.


plugin/core/views.py
```diff
@@ -41,6 +42,7 @@
 import tempfile
 
 MarkdownLangMap = Dict[str, Tuple[Tuple[str, ...], Tuple[str, ...]]]
+QuickPanelKind = Tuple[int, str, str]
```
jwortmann, line 45: [`sublime.Kind`s](https://www.sublimetext.com/docs/api_reference.html#sublime.Kind) are not only used for the quick panel (also for completion popup and ListInputHandler).

Maybe call it "SublimeKind" instead?

rchl: I wonder why `sublime.Kind` is documented but doesn't seem to actually exist...

rchl: Never mind, it exists as a type, not as a value. And only in 3.8 host.

rwols (approved):

.github/workflows/main.yml
```diff
@@ -38,7 +38,8 @@ jobs:
           python-version: '3.8'
       - run: sudo apt update
       - run: sudo apt install --no-install-recommends -y x11-xserver-utils
-      - run: pip3 install mypy==0.910 flake8==3.9.2 yapf==0.31.0 --user
+      - run: pip3 install mypy==0.971 flake8==5.0.4 pyright==1.1.271 yapf==0.31.0 --user
       - run: echo "$HOME/.local/bin" >> $GITHUB_PATH
-      - run: mypy -p plugin
+      # - run: mypy -p plugin
```
predragnikolic, line 43: I would remove it, or leave a comment above the line saying why it was disabled.

tests/test_file_watcher.py
```diff
@@ -102,7 +102,7 @@ def tearDown(self) -> Generator:
         self.assertEqual(len(TestFileWatcher._active_watchers), 0)
         # Restore original project data.
         window = sublime.active_window()
-        window.set_project_data(None)
+        window.set_project_data({})
```
predragnikolic, line 105: is this intentional?
what is the difference between None and {}?

predragnikolic (approved):

tests/test_file_watcher.py
rchl: Fix for type issue. According to types it doesn't accept `None`.
I don't think there are differences in practice. The tests still pass at least.

.github/workflows/main.yml
rchl: I would expect that we could re-enable it soon so adding a comment for now

plugin/core/typing.py
```diff
@@ -21,9 +22,11 @@
     from typing import Set
     from typing import Tuple
     from typing import Type
-    from typing import TypedDict
     from typing import TypeVar
     from typing import Union
+    from typing_extensions import NotRequired
+    from typing_extensions import Required
+    from typing_extensions import TypedDict
```
rwols, line 29: typing_extensions is a third-party package, so this won't work on python 3.8.

rchl: In 3.8 we could keep using real imports for type checking and dummy ones for the plugin host, like we do now.
Then:
 - this won't really matter
 - if we set python version to 3.11 for pyright then we can import those from `typing`

One thing I'm not sure about (although it doesn't matter for ST) is how both pyright and mypy don't have problems with this `typing_extensions` import. I guess that they either come with own copies or use system ones?

rchl: Addressing in https://github.com/sublimelsp/LSP/pull/2067

Issue #1846, opened 2021-09-12
Split up Code Actions into "Refactors" and "Source Actions"


rwols:
**Is your feature request related to a problem? Please describe.**

Currently we bin all code actions into "Code Actions", but servers like

- jdtls
- OmniSharp

have a clear distinction between "Refactorings" and "Source Actions".

**Describe the solution you'd like**

 In case of "Refactorings", these kinds should not always be directly visible to the user. Source Actions should be visible, like they are right now with annotations.

I think Refactorings should go into the "Edit" top menu to be selected with your mouse or in the command palette.

rchl:
> Source Actions should be visible, like they are right now with annotations.

So you want to show them in the same way as normal code actions or in some other way? We already have a `Source action...` in the right click context menu. Showing those as annotations would not be correct IMO as those don't bind to specific line in the code.

rwols:
Sorry for the late reply. I'm talking about refactorings.

<img width="1080" alt="Schermafbeelding 2022-02-27 om 11 26 25" src="https://user-images.githubusercontent.com/2431823/155878748-6c770f5f-3092-4287-95e1-54149f8910cd.png">

You can see below I get mostly refactoring kinds.

```
:: --> OmniSharp textDocument/codeAction(5): {'textDocument': {'uri': 'file:///Users/raoulwols/dev/csharp-project/Program.cs'}, 'context': {'diagnostics': []}, 'range': {'end': {'line': 14, 'character': 22}, 'start': {'line': 14, 'character': 22}}}
[...]

PR #2064, opened 2022-09-22
Ensure "Source Actions..." request includes the "source" kind

rchl:
Makes sure that the code action request that we make when triggering the "Source actions..." command explicitly includes `"only": ["source"]` kind.

Related to https://github.com/sublimelsp/LSP/issues/1846#issuecomment-1255150008

(More readable diff with `Hide whitespace changes` enabled)

@jwortmann

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -96,8 +96,8 @@ class CodeActionsManager:
             view,
             region,
             session_buffer_diagnostics,
-            False,
-            actions_handler,
+            only_with_diagnostics=False,
+            actions_handler=actions_handler,
             on_save_actions=None,
             only_kinds=only_kinds,
             manual=manual)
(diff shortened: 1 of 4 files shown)


predragnikolic (approved):

jwortmann:
This looks good to me, but beside that I would also suggest the following changes (not necessarily in this PR):
* remove the "Source Action..." entry from the context menu (because source actions are not bound to a particular point or selection, but apply to the entire file)
* Add "LSP: Source Actions..." to the "Edit" main menu, and either add another argument `full_range: bool = False` (or so) for the code actions command, which would be activated for the source actions menu entry, or alternatively derive this from whether `only_kinds` includes `"source"`. So that the request for source actions would use the range of the entire file and also include all of the diagnostics in the file (this could be useful for the `source.fixAll` kind)

rchl:
Created https://github.com/sublimelsp/LSP/issues/2066 as a follow up.

PR #2065, opened 2022-09-23
Request color presentations when clicking on a color box

jwortmann:
This runs the `textDocument/colorPresentation` request when clicking on a color box, and shows the results in a quick panel. I think it is better to apply the change into the file only after selecting one of the quick panel entries, rather than immediately like it's implemented in the VSCode color picker when cycling through the different color formats. Because the conversion from RGB to HSL/HWB and back again might alter the original color (e.g. `#dddddd` -> `hsl(0, 0%, 87%)` -> `hwb(0 87% 13%)` -> `rgb(222, 222, 222)` -> `#dedede`).

There is no special server capability for this request by the way, so we can't decide beforehand whether or not it is supported by the server and is useful to make the color box clickable. (But maybe it's expected that servers that have the colorProvider capability, which is used for `textDocument/documentColor`, should also support colorPresentation.)

And there can be optional "additionalTextEdits" in the response, which are ignored in this PR. I don't really know what should be the use of those; maybe it's to replace all occurencies of the particular color. But that seems rather dangerous or unexpected for the user to me, so I guess it's better to leave this out for now.

Note that you can observe two color boxes for a very short amount of time after selecting a quick panel entry, but this seems unrelated to this pull request. It's probably some kind of bug in the code how the color box phantoms are updated, or in the lsp_apply_document_edit command (or in how these two situations work together).

And in case a minihtml color picker gets implemented at a later point, this functionality could be adjusted with minimal changes to be invoked from a link in the color picker.

### DIFF
--- a/plugin/color.py
+++ b/plugin/color.py
@@ -3,7 +3,7 @@ from .core.protocol import ColorPresentation
 from .core.protocol import ColorPresentationParams
 from .core.protocol import Request
 from .core.registry import LspTextCommand
-from .core.typing import List, Optional
+from .core.typing import List
 from .core.views import range_to_region
 import sublime
 
@@ -12,12 +12,15 @@ class LspColorPresentationCommand(LspTextCommand):
 
     capability = 'colorProvider'
 
-    def run(self, edit: sublime.Edit, params: ColorPresentationParams, event: Optional[dict] = None) -> None:
+    def run(self, edit: sublime.Edit, params: ColorPresentationParams) -> None:
         session = self.best_session(self.capability)
         if session:
             self._version = self.view.change_count()
             self._range = params['range']
             session.send_request_async(Request.colorPresentation(params, self.view), self._handle_response_async)
+
+    def want_event(self) -> bool:
+        return False
 
     def _handle_response_async(self, response: List[ColorPresentation]) -> None:
         if not response:
--- a/plugin/color.py
+++ b/plugin/color.py
@@ -1,10 +1,12 @@
 from .core.edit import parse_text_edit
+from .core.protocol import ColorInformation
 from .core.protocol import ColorPresentation
 from .core.protocol import ColorPresentationParams
 from .core.protocol import Request
 from .core.registry import LspTextCommand
 from .core.typing import List
 from .core.views import range_to_region
+from .core.views import text_document_identifier
 import sublime
 
 
@@ -12,11 +14,16 @@ class LspColorPresentationCommand(LspTextCommand):
 
     capability = 'colorProvider'
 
-    def run(self, edit: sublime.Edit, params: ColorPresentationParams) -> None:
+    def run(self, edit: sublime.Edit, color_information: ColorInformation) -> None:
         session = self.best_session(self.capability)
         if session:
             self._version = self.view.change_count()
-            self._range = params['range']
+            self._range = color_information['range']
+            params = {
+                'textDocument': text_document_identifier(self.view),
+                'color': color_information['color'],
+                'range': self._range
+            }  # type: ColorPresentationParams
             session.send_request_async(Request.colorPresentation(params, self.view), self._handle_response_async)
 
     def want_event(self) -> bool:
--- a/plugin/color.py
+++ b/plugin/color.py
@@ -41,10 +41,11 @@ class LspColorPresentationCommand(LspTextCommand):
             elif item['label'] == old_text:
                 continue
             self._filtered_response.append(item)
-        window.show_quick_panel(
-            [sublime.QuickPanelItem(item['label']) for item in self._filtered_response],
-            self._on_select,
-            placeholder="Change color format")
+        if self._filtered_response:
+            window.show_quick_panel(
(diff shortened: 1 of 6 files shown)


plugin/session_buffer.py
```diff
@@ -353,9 +354,8 @@ def _do_color_boxes_async(self, view: sublime.View, version: int) -> None:
                 self._if_view_unchanged(self._on_color_boxes_async, version)
             )
 
-    def _on_color_boxes_async(self, view: sublime.View, response: Any) -> None:
-        color_infos = response if response else []
-        self.color_phantoms.update([lsp_color_to_phantom(view, color_info) for color_info in color_infos])
+    def _on_color_boxes_async(self, view: sublime.View, response: List[ColorInformation]) -> None:
+        self.color_phantoms.update([lsp_color_to_phantom(view, color_info) for color_info in response])
```
rchl, line 358: See https://github.com/sublimelsp/LSP/pull/2053#discussion_r972370161

plugin/color.py
```diff
@@ -0,0 +1,54 @@
...
+            return
+        if self._version != self.view.change_count():
+            return
+        old_text = self.view.substr(range_to_region(self._range, self.view))
+        self._filtered_response = []  # type: List[ColorPresentation]
+        for item in response:
+            # Filter out items that would apply no change
+            text_edit = item.get('textEdit')
+            if text_edit:
+                if text_edit['range'] == self._range and text_edit['newText'] == old_text:
```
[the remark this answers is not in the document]
jwortmann: Hm, what makes you think that this would be the case?

According to https://stackoverflow.com/a/40921229 this is the correct way to compare equality of keys and values in Python dicts.

And I can confirm that it works with LSP-css, i.e. the server does use `textEdit`s and the unnecessary item is correctly filtered out.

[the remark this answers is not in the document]
rchl: Well, Python is smarter than I thought. Never mind.

plugin/session_buffer.py
jwortmann: Okay, good to know. I added an explicit comment about that, so that the next one who touches this code won't try to do the same again :D

rchl (approved):

plugin/core/views.py
```diff
@@ -781,28 +783,49 @@ def run(self, view_id: int, command: str, args: Optional[Dict[str, Any]] = None)
...
+        return "#{:02x}{:02x}{:02x}{:02x}".format(red, green, blue, int(alpha_dec * 255))
+    return "#{:02x}{:02x}{:02x}".format(red, green, blue)
 
 
-def lsp_color_to_phantom(view: sublime.View, color_info: Dict[str, Any]) -> sublime.Phantom:
+def lsp_color_to_html(view: sublime.View, color_info: ColorInformation) -> str:
+    command = sublime.command_url('lsp_color_presentation', {
+        'params': {
+            'textDocument': text_document_identifier(view),
+            'color': color_info['color'],
```
predragnikolic, line 819: Have you though about passing the hole color_information dict,
instead of passing color and range dicts?

plugin/core/views.py
```diff
@@ -781,28 +783,49 @@ def run(self, view_id: int, command: str, args: Optional[Dict[str, Any]] = None)
...
 COLOR_BOX_HTML = """
-<style>html {{padding: 0; background-color: transparent}}</style>
+<style>
+    html {{
+        padding: 0;
+        background-color: transparent;
+    }}
+    a {{
+        display: inline-block;
+        height: 0.8rem;
```
predragnikolic, line 793: What do you think about 0.9rem for the height and width?

plugin/core/views.py
jwortmann: Changed in https://github.com/sublimelsp/LSP/pull/2065/commits/78151f2792c4778385b554a5423feefb8a894855.

Not sure if you were thinking about doing it exactly this way, but it is actually a good idea to postpone the `text_document_identifier` function call to later when the color box is clicked.

I still need to use a new variable and extract the "color" and "range" values from ColorInformation one-by-one, because iirc otherwise pyright would complain about invalid assignment of a new unknown "textDocument" key for the params needed for the request.

plugin/core/views.py
jwortmann: Personally I'd prefer to keep it at 0.8rem. It's probably only 1 pixel difference, but with the font size and screen resolution I use there seems only to be 1 pixel between adjacent color boxes on consecutive lines when using 0.9rem.

But I would be okay with increasing them to 0.9rem if bigger color boxes are preferred by the majority.

rchl: I've tried both and I don't really have strong preference either way.

predragnikolic:
on Linux there is an underline
image

this should fix that
```diff
COLOR_BOX_HTML = """
<style>
        display: inline-block;
        // ....
        background-color: {color};
+        text-decoration: none;
    }}
```

predragnikolic (approved):
Apart from https://github.com/sublimelsp/LSP/pull/2065#issuecomment-1257257846

Everything else looks good.

PR #2069, opened 2022-09-26
ignore background files for toggling diagnostics panel on save

rchl:
The issue was that files in the background have received new diagnostics and since the "last save change count" has still matched, that has triggered the panel to hide or open.

To fix, reset the "last save change count" after showing or hiding the panel so that it's only relevant again after next save.

It was reproducible by:
 1. having an error in one file and saving it
 2. hiding the panel
 3. switching to another file with no errors and saving it

The file saved in 1 would receive updated diagnostics and the logic would pop the panel open again.

(Of course those steps depend on what servers are used and whether saves in certain files trigger diagnostics in other files)

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -287,7 +287,9 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
             self._do_code_actions()
         self._update_diagnostic_in_status_bar_async()
         if self.view.change_count() == self._change_count_on_last_save:
-            self._toggle_diagnostics_panel_if_needed_async()
+            window = self.view.window()
+            if window and window.active_view() == self.view:
+                self._toggle_diagnostics_panel_if_needed_async()
 
     def _update_diagnostic_in_status_bar_async(self) -> None:
         if userprefs().show_diagnostics_in_view_status:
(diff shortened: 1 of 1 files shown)


rchl:
Argh... This re-introduces the previous problem where the panel might close and not re-open on introducing new error and saving quickly...

rchl:
Only handling toggling of the panel for the active view when diagnostics are updated should make it much better.

There can still be cases where:
 - the panel is open
 - a first error is introduced and file quickly saved
 - the panel gets hidden and shown again quickly

The next step to improve those cases could be to have a short delay (debouncer) before reacting to diagnostic changes but that can be for another PR.

plugin/documents.py
```diff
@@ -287,7 +287,9 @@ def on_diagnostics_updated_async(self) -> None:
             self._do_code_actions()
         self._update_diagnostic_in_status_bar_async()
         if self.view.change_count() == self._change_count_on_last_save:
-            self._toggle_diagnostics_panel_if_needed_async()
+            window = self.view.window()
+            if window and window.active_view() == self.view:
+                self._toggle_diagnostics_panel_if_needed_async()
```
predragnikolic, line 292: Maybe rewrite it a bit to keep the indendtaion level smaller:
```py
        window = self.view.window()
        is_active_view = bool(window and window.active_view() == self.view)
        if is_active_view and  self.view.change_count() == self._change_count_on_last_save:
            self._toggle_diagnostics_panel_if_needed_async()
```

predragnikolic (approved):

PR #2071, opened 2022-09-27
Allow empty string for workspace symbol request

### LINKED ISSUE #1008: Workspace Symbol command should accept an empty query string, since 3.15

predragnikolic:
```
/**
 * The parameters of a Workspace Symbol Request.
 */
interface WorkspaceSymbolParams extends WorkDoneProgressParams,
    PartialResultParams {
    /**
     * A query string to filter symbols by. Clients may send an empty
     * string here to request all symbols.
     */
    query: string;
}
```

closes #1008

### DIFF
--- a/Main.sublime-menu
+++ b/Main.sublime-menu
@@ -45,6 +45,10 @@
         "command": "lsp_document_symbols"
       },
       {
+        "caption": "LSP: Goto Symbol In Project…",
+        "command": "lsp_workspace_symbols"
+      },
+      {
         "caption": "LSP: Goto Definition…",
         "command": "lsp_symbol_definition"
       },
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -5889,8 +5889,8 @@ class Request:
         return Request("textDocument/willSaveWaitUntil", params, view)
 
     @classmethod
-    def documentSymbols(cls, params: Mapping[str, Any], view: sublime.View) -> 'Request':
-        return Request("textDocument/documentSymbol", params, view)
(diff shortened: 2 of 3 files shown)


predragnikolic:
this is the second attempt, 
the first was https://github.com/sublimelsp/LSP/pull/1018

and was mainly closed due to https://github.com/sublimelsp/LSP/pull/1018#issuecomment-624237916

----

> It might be possible that servers that support < 3.15 will assume the query-string is non-empty (and might crash if provided with an empty query string). But I might be splitting hairs here.

There was also this concern that a language server would could potentially crash if the client sends an empty string...
but we are still adhering to the spec, so that means that the server would have to fix.

rchl:
I wonder why LSP menus for the main menu don't include `lsp_workspace_symbols`...
Would it be a good idea to add it here?

I'm gonna test this out but I remember doing it myself some time and for some reason not following up with it. Not sure why...

rchl (approved):

Issue #1694, opened 2021-05-28
Function parameters highlight the wrong text in the tooltip


guniu:
**Describe the bug**
I don't know how to call this, but when typing out function name and parenthesis you get the tooltip with function's parameters and maybe description. The parameters highlight in orange color, but if the parameter name is a substring of the function name (case-sensitive) it will highlight that substring instead (look screenshot).

**Expected behavior**
Parameters inside the parentheses should be highlighted.

**Screenshots**
Notice the params are highlighted in the function name instead of inside the brackets:
![image](https://user-images.githubusercontent.com/19828894/119947966-a17a1980-bf98-11eb-8b20-991e6b44e3ea.png)

How it should be:
![image](https://user-images.githubusercontent.com/19828894/119948112-cd959a80-bf98-11eb-811d-dd179ba494e5.png)

**Environment (please complete the following information):**
- OS: Windows 10
- Sublime Text version: 4107
- LSP version: 1.3.1
- Language servers used: pylsp

**Additional context**
This bug was present right after updating ST3 to ST4 (with old LSP and pyls).
I had some error regarding LSP/pyls after updating ST, so I uninstalled LSP and python-language-server, installed new python-lsp-server and LSP (I had to change in the LSP settings 'pyls' to 'pylsp') but after all that the bug is still there.

rwols:
labelOffsetSupport was added in 3.14 of the spec. We're at 3.17 now. labelOffsetSupport is a much more clear way to communicate to a language client (us) on what to emphasize. pylsp still uses the unreliable old way of using substrings. I'll add word boundary checks for this, but pylsp should really use labelOffsetSupport. It might be the only language server left that's still using this bad way of communicating labels.

PR #2072, opened 2022-09-27
Improve highlighting of signature parameters when labels are of type string

predragnikolic:
fixes https://github.com/sublimelsp/LSP/issues/1694

### DIFF
--- a/plugin/core/signature_help.py
+++ b/plugin/core/signature_help.py
@@ -9,6 +9,7 @@ from .views import MarkdownLangMap
 from .views import minihtml
 import functools
 import html
+import re
 import sublime
 
 
@@ -115,7 +116,8 @@ class SigHelp:
                     # route relies on the client being smart enough to figure where the parameter is inside of
                     # the signature label. The above case where the label is a tuple of (start, end) positions is much
                     # more robust.
-                    start = label[prev:].find(rawlabel)
+                    label_match = re.search(r"\b{}\b".format(rawlabel), label[prev:])
+                    start = label_match.start() if label_match else -1
                     if start == -1:
                         debug("no match found for {}".format(rawlabel))
                         continue
--- a/plugin/core/signature_help.py
+++ b/plugin/core/signature_help.py
@@ -116,10 +116,8 @@ class SigHelp:
                     # route relies on the client being smart enough to figure where the parameter is inside of
                     # the signature label. The above case where the label is a tuple of (start, end) positions is much
                     # more robust.
-                    label_match = re.search(r"\W{}\W".format(re.escape(rawlabel)), label[prev:])
-                    start = -1
-                    if label_match:
-                        start = label_match.start() + 1  # skip the first matched \W by adding + 1
+                    label_match = re.search(r"\b{}\b".format(re.escape(rawlabel)), label[prev:])
+                    start = label_match.start() if label_match else label[prev:].find(rawlabel)
                     if start == -1:
                         debug("no match found for {}".format(rawlabel))
                         continue
--- a/plugin/core/signature_help.py
+++ b/plugin/core/signature_help.py
@@ -116,7 +116,7 @@ class SigHelp:
                     # route relies on the client being smart enough to figure where the parameter is inside of
                     # the signature label. The above case where the label is a tuple of (start, end) positions is much
                     # more robust.
-                    label_match = re.search(r"\W{}\W".format(rawlabel), label[prev:])
+                    label_match = re.search(r"\W{}\W".format(re.escape(rawlabel)), label[prev:])
                     start = -1
                     if label_match:
                         start = label_match.start() + 1  # skip the first matched \W by adding + 1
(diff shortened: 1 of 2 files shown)


predragnikolic:
Here is a code snippet that can be tried with LSP-pylsp

```py
def ExampleFunctionName(Ex, Fun, i, name):
    pass


ExampleFunctionName() 
```

plugin/core/signature_help.py
```diff
@@ -115,7 +116,8 @@ def _render_label(self, view: sublime.View, signature: SignatureInformation) ->
                     # route relies on the client being smart enough to figure where the parameter is inside of
                     # the signature label. The above case where the label is a tuple of (start, end) positions is much
                     # more robust.
-                    start = label[prev:].find(rawlabel)
+                    label_match = re.search(r"\b{}\b".format(rawlabel), label[prev:])
+                    start = label_match.start() if label_match else -1
```
rwols, line 120: Does it work with

```
def f(str: str, int: int) -> str:
    return ""
```

rwols: What about `f(x: str, str: int)`

jwortmann: > What about `f(x: str, str: int)`

Then I assume it will do the same as before, i.e. mark the first occurence of "str" it can find. But I guess there is no way to fix that without adding complex and language-dependent code about what should be considered as a parameter and what is a type, etc.

But if you want to find an example where this PR would actually be a regression compared to the status quo, you can take this one:

```julia
f(🐰, 🐺) = 🐰 + 🐺
```

Before (at least the underline is visible for the highlighted parameter):
before

After:
after

This happens because this PR assumes that the parameter names must always begin and end with what Python's regex considers as a "word" character (otherwise the word boundary `\b` wouldn't match).

But it's probably a very rare example (but not a complete joke example, actually it's adopted with some adjustments from https://www.youtube.com/watch?v=QwVO0Xh2Hbg&t=4765s), which seems to be acceptable in exchange for the improvement from this PR. I also tried with greek letters (which would be much more common than emojis in parameter identifiers), and in that case it works fine because they are considered as word characters.

predragnikolic: Offtopic :D 
This is the first time seeing emojis being used in code like this XD `f(🐰, 🐺) = 🐰 + 🐺`

Julia ls takes forever(maybe around 30s ~ 1min) to initialize on linux on my machine, is that normal?

predragnikolic:
I am unsure on how to fix this `sre_constants.error: unexpected end of regular expression` build error.

plugin/core/signature_help.py
```diff
@@ -115,7 +116,10 @@ def _render_label(self, view: sublime.View, signature: SignatureInformation) ->
                     # route relies on the client being smart enough to figure where the parameter is inside of
                     # the signature label. The above case where the label is a tuple of (start, end) positions is much
                     # more robust.
-                    start = label[prev:].find(rawlabel)
+                    label_match = re.search(r"\W{}\W".format(re.escape(rawlabel)), label[prev:])
+                    start = -1
+                    if label_match:
+                        start = label_match.start() + 1  # skip the first matched \W by adding + 1
```
rchl, line 122: `main`:

Screenshot 2022-09-28 at 12 25 47

this branch:

Screenshot 2022-09-28 at 12 26 07

predragnikolic: To avoid having issues like this,

Prefer searching by word boundary, 
if a match is not found,
fallback to the previous logic for finding the parameter.

jwortmann: Another alternative that I think should always work would be to use a negative lookbehind and negative lookahead for word characters in the regex: `r"(?<!\w){}(?!\w)"`

But the idea to fallback to the previous logic should be fine too.

plugin/core/signature_help.py
jwortmann: >This is the first time seeing emojis being used in code like this XD `f(🐰, 🐺) = 🐰 + 🐺`

Yeah, I think nobody would actually use emojis for things like this in real code ;)
It's more that Julia is often used for numerical calculations from a mathematical or physics background, and so the language allows a wide range of unicode characters in identifiers to make possible e.g. the use of greek letters for variables / physical constants or mathematical symbols for special operators.
 
> Julia ls takes forever(maybe around 30s ~ 1min) to initialize on linux on my machine, is that normal?

Hm, 1min seems pretty long, but yes it takes a while for the server to initialize (I can measure ~ 20sec here). That is because the server will be compiled each time when it starts, and also you can observe that using any feature (e.g. hover) will also lag a few seconds when it is used the very first time (again, the particular code parts for that feature will be compiled just-in-time directly before usage). But after that it should be fast. Unfortunately, afaik there is no good way to fully compile Julia code (i.e. the language server) ahead of time yet. But I think it still starts a bit faster than on VSCode :)

rchl (approved):

plugin/core/signature_help.py
predragnikolic: IMO not having a fallback logic is better.
So I went with your suggestion by using the regex.

PR #2074, opened 2022-09-28
Add code lens image to the docs

### LINKED ISSUE #2073: Add screenshots for Code Lenses in documentation
Code Lenses are missing screenshots in the docs. It would be nice if they included a little animated gif like the other features do.

https://lsp.sublimetext.io/features/#code-lenses

predragnikolic:
closes #2073


rchl:
Maybe more interesting with something like:

Screenshot 2022-09-28 at 14 53 59

?

Also, your screenshot doesn't use default value for `show_code_lens` and I think it's better to show the default (or both).

rchl:
With default value of `show_code_lens`:

Screenshot 2022-09-28 at 14 55 51

rchl:
Kinda feel the default value should be `phantom` now :)

predragnikolic:
Kind feel like we should remove the "annotation" and only leave "phantom" :D (because LSP uses annotations for code actions, code lens and maybe diagnostics in the future) but I would discuss that, before doing anything.

Yeah, I'll steal your screenshot (with phantoms) :)

predragnikolic:
Last commit uses annotations value for code lens. 
I can revert the last commit if you like the phantoms image more.

rchl:
We could show both, mentioning also the relevant option that controls that.

LDAP:
> Kind feel like we should remove the "annotation" and only leave "phantom" :D (because LSP uses annotations for code actions, code lens and maybe diagnostics in the future) but I would discuss that, before doing anything.

Wouldn't remove it entirely. Sometimes it is quite annoying when the code shifts because new code lenses are available or the server just started.

jwortmann:
> Kinda feel like the default value should be `phantom` now :)

I have no strong opinion about this particular setting, but I would also feel like the following settings could better use a different default value (everything is very subjective of course):

- `"show_diagnostics_panel_on_save": 2` -> `0`
  Does anyone really use the default value for this setting? Wouldn't it be very annoying that the diagnostics panel pops up every time you save if a language server thinks that there are errors or warnings... perhaps it's just me and I save much too often :D
- `"show_diagnostics_in_view_status": true` -> `false`
   Certainly it's an interesting and useful feature, but for me it feels like that this should be something to opt-in. Because status bar space is quite limited and constantly updating messages there can be a bit distractive
- `"diagnostics_gutter_marker": "dot"` -> `"sign"`
  The different "sign" icons for the severities are clearly superior to the boring "dot" :D (in my opinion of course).
  Also one could consider to not show any gutter icon for the "hint" severity at all, because currently both "info" and "hint" are blue with the same icons and underlines, so there is no way to visibly distinguish them (but I don't know in general what should be the difference between "info" and "hint" severities from the LSP specs)
- `"link_highlight_style": "underline"` -> `"none"`
  Would just be a subtle visual preference from my side - no decisive reason here
- `"show_references_in_quick_panel": false` -> `true`
  The built-in "Goto Reference..." also uses the quick panel if there are multiple results. And it can easily and quickly be navigated with the keyboard.

rchl (approved):

LDAP (commented):
Is docs/src/images/code-lens.png used anywhere?

LDAP (approved):

genericptr:
I'm still a little confused about this implementation. What happens when you click the buttons? In VSCode they would bring up a panel which shows context around the code (thus the idea of a lens) but here they appear to be like code actions.

PR #2079, opened 2022-10-05
Session.open_uri_async: Focus view when open file: URI

### LINKED ISSUE #2078: `session.open_uri_async` should always focus the view
`session.open_uri_async` does only focus the view if a `range` is passed. `open_file` has the same issue. I think there is also some inconsistency in the implementation of `open_uri_async`: If the uri does not start with `file:` then the view is focused as I would expect.

_Originally posted by @LDAP in https://github.com/sublimelsp/repository/issues/71#issuecomment-1266020892_

LDAP:
Closes #2078 

`Session.open_uri_async` focuses a view if there is a preexisting session buffer:
https://github.com/sublimelsp/LSP/blob/969a9c8d8f581f4815643b3297501cb55b49a1d9/plugin/core/sessions.py#L1487-L1490
or if the URI was opened using a plugin, since in this case the view was newly created:
https://github.com/sublimelsp/LSP/blob/969a9c8d8f581f4815643b3297501cb55b49a1d9/plugin/core/sessions.py#L1531-L1534
However, when the URI starts with `file:` then the view is only focused if the file was previously not open.


plugin/core/sessions.py
```diff
@@ -1508,6 +1508,11 @@ def _open_file_uri_async(
         def handle_continuation(view: Optional[sublime.View]) -> None:
             if view and r:
                 center_selection(view, r)
+            elif view:
+                window = view.window()
+                if window:
+                    window.focus_view(view)
```
rchl, line 1514: There is one place in LSP code that exercises this branch and that's `apply_workspace_edit_async`. And in that case it makes more sense not to focus all views that are edited IMO.

I feel like there is some refactoring that is needed here to make this all easier to follow and know what to expect. Possibly there should be an explicit argument in all "open file" functions (and there is quite a few) to tell them what to do.

rchl: I guess one obstacle to making this more predictable is the fact that there is no way to open a new file and **not** give it focus...

LDAP: But that also means that there is currently no way to always focus the view with Session.open_uri_async, which is a quite common use case in my opinion.

LDAP: Maybe apply_workspace_edit_async shouldn't call Session.open_uri_async at all since that is not what I would expect from the method name. I may attempt a refactoring in that regard once I got some spare time.

rchl: > Maybe apply_workspace_edit_async shouldn't call Session.open_uri_async at all since that is not what I would expect from the method name.

We have to open the file to apply the edit.

PR #2081, opened 2022-10-05
Filter out non-quickfix actions in view

rchl:
Changes:
 - refactor by removing the `CodeActionsCollector` which was basically a variant of `Promise.all` implementation.
 - add session name as an annotation to the Code Actions Quick Panel.
 - filter out non-quickfix code actions for code actions requested for current cursor position

I've made this code a week or so ago and left it in a WIP state but I don't remember exactly what was left to do. There might be missing things or bugs so I will run with it for a few days to try to catch any possible issues.


plugin/code_actions.py
```diff
@@ -74,7 +74,8 @@ def response_filter(session: Session, actions: List[CodeActionOrCommand]) -> Lis
             if only_kinds:
                 return [a for a in actions if not is_command(a) and is_kinds_include_kind(only_kinds, a.get('kind'))]
             return [
-                a for a in actions if is_command(a) or is_kinds_include_kind([CodeActionKind.QuickFix], a.get('kind'))
+                a for a in actions
+                if is_command(a) or not a.get('kind') or is_kinds_include_kind([CodeActionKind.QuickFix], a.get('kind'))
```
rchl, line 78: The `is_command(a)` is technically not needed.

Ideally the `is_command` type guard would tell pyright/mypy that `a` in later conditions is not of `Command` type but they are not that smart so it's still an union. But at least pyright doesn't complain about `.get('kind')` (`mypy` does but we don't care right now).

plugin/code_actions.py
```diff
@@ -7,227 +7,190 @@
...
 
+def is_kinds_include_kind(kinds: List[CodeActionKind], kind: Optional[CodeActionKind]) -> bool:
+    """
+    The "kinds" list includes "kind" if any of the "kinds" matches "kind" exactly or the "kind" prefix matches one
+    of the "kinds".
+    """
+    if not kind:
+        return False
+    kind_parts = kind.split('.')
+    return any(part for part in kind_parts if part in kinds)
```
jwortmann, line 191: This like doesn't look right.

Some examples where it would fail:

* if `"only_kinds": ["refactor"]` and returned kind "foobar.refactor" would falsely match
* if `"only_kinds": ["source.organizeImports"]` and returned kind is also "source.organizeImports", then it wouldn't match


Also compare with my implementation at https://github.com/sublimelsp/LSP/pull/2080/files

plugin/code_actions.py
```diff
@@ -293,65 +253,60 @@ def run(
...
+            self._show_code_actions(actions)
 
-    def handle_select(self, index: int) -> None:
-        if index > -1:
+    def _show_code_actions(self, actions: List[Tuple[ConfigName, CodeActionOrCommand]]) -> None:
+        window = self.view.window()
+        if not window:
+            return
+        items, selected_index = format_code_actions_for_quick_panel(actions)
+        window.show_quick_panel(
```
jwortmann, line 291: `show_quick_panel` is smart enough to not do anything if `items` is empty, but it still seems wasteful to do that API call in this case. Also there needs to be some kind of feedback if no code actions are available, otherwise it's bad UX if user selects the menu entry and nothing happens (I like that the strange popup is gone now, though). A status bar message seems appropriate to me:

```python
if items:
    window.show_quick_panel(...)
else:
    window.status_message("No actions available")
```

tests/test_code_actions.py
```diff
@@ -232,25 +232,35 @@ def _setup_document_with_missing_semicolon(self) -> Generator:
...
 
+    def test_kind_matching(self) -> None:
+        self.assertTrue(kinds_include_kind(['a'], 'a.b'))
+        self.assertFalse(kinds_include_kind(['a'], 'b.a'))
+        self.assertFalse(kinds_include_kind(['a.b'], 'b'))
+        self.assertTrue(kinds_include_kind(['a.b'], 'a.b'))
+        self.assertFalse(kinds_include_kind(['a.b'], 'a'))
+        self.assertFalse(kinds_include_kind(['aa'], 'a'))
+        self.assertFalse(kinds_include_kind(['aa.b'], 'a'))
+        self.assertFalse(kinds_include_kind(['aa.b'], 'b'))
```
rwols, line 262: This reminds me a lot of the `sublime.score_selector` function. Is it equivalent?

rchl: It does indeed seem to work in the same way if we transform first argument to space-separate string. I guess using it directly might be confusing but having a wrapper function with more relevant name should be fine...

rchl: Actually maybe not since it's "backwards".

`scope_selector` matches single selectors against many scopes while we need to match many selectors against a single scope.

We could call it separately for each selector (item in `kinds`) but that seem a bit wasteful.

rchl:
It works on marksman.
Can you post code action response where it doesn't work?

jwortmann:
> Can you post code action response where it doesn't work?

```
:: <<< julia 30: [{'command': {'command': 'AddDocstringTemplate', 'arguments': ['file:///D%3A/code/Julia/lsptest/main.jl', 5], 'title': 'Add docstring template for this method'}, 'title': 'Add docstring template for this method'}, {'kind': 'refactor', 'command': {'command': 'ExpandFunction', 'arguments': ['file:///D%3A/code/Julia/lsptest/main.jl', 5], 'title': 'Expand function definition.'}, 'title': 'Expand function definition.'}, {'command': {'command': 'AddLicenseIdentifier', 'arguments': ['file:///D%3A/code/Julia/lsptest/main.jl', 5], 'title': 'Add SPDX license identifier.'}, 'title': 'Add SPDX license identifier.'}]
```

rchl:
And which ones of those 3 are shown in the view?

Two of those are `Command`s so don't have a kind and we don't filter those out.
One of them is a `refactor` kind so should be filtered out. Will check if it is...

jwortmann:
It seems that items are not filtered out if `is_command` returns `True`.
```python
            return [
                a for a in actions
                if is_command(a) or not a.get('kind') or kinds_include_kind([CodeActionKind.QuickFix], a.get('kind'))
            ]
```

But a CodeAction can also have the `command` property, so this seems to be the bug.

jwortmann:
* Filtering the code actions shown in the view seems to work now, but when I manually select "Code Action..." from the context menu, the other kinds are also filtered out now. So there seems to be no way anymore to get the "refactor" code actions. I would expect manually running "Code Action..." from the context menu to show all actions which are returned by the server.

* Compared to my PR I'm still missing the entries for code actions under the "Edit" main menu. We could also think about adding a "Refactor..." entry with `"only_kinds": ["refactor"]`, either to the context menu like in VSCode or to the main menu (or both).

* If the setting to show code actions in the hover popup is activated (the default), then code actions for this are also filtered. Not sure if this is intended or the best solution. Maybe it should also show all returned actions in the popup, because in the popup they are not so distracting compared to permanently be shown in the view. Or alternatively, if they should only be quickfix actions, then the label could be changed from "[server name] Code Action: ..." to "[server name] Quick Fix: ..." (maybe also drop the "[server name]"?)

plugin/code_actions.py
```diff
@@ -7,227 +7,198 @@
...
+    if not kind:
+        return False
+    for kinds_item in kinds:
+        if kinds_item == kind:
+            return True
+        kind_dots = re.finditer(r'\.', kind)
+        for dot in kind_dots:
+            kind_prefix = kind[0:dot.start()]
+            if kind_prefix == kinds_item:
+                return True
```
jwortmann, line 198: This works, but it looks overly complicated compared to my 2-line solution, which doesn't require using the regex library just to search for `.` in the string, and doesn't need a for-loop.

Any reason for doing it this way?

rchl:
> * So there seems to be no way anymore to get the "refactor" code actions.

Good point.
I agree that explicitly triggered `lsp_code_actions` should bring up all code actions.

> * Compared to my PR I'm still missing the entries for code actions under the "Edit" main menu. We could also think about adding a "Refactor..." entry with `"only_kinds": ["refactor"]`, either to the context menu like in VSCode or to the main menu (or both).

I don't know about the Edit menu. I feel like that should be added/discussed separately.
But adding context menu entry for Refactor sounds good to me.

> * If the setting to show code actions in the hover popup is activated (the default), then code actions for this are also filtered. Not sure if this is intended or the best solution. Maybe it should also show all returned actions in the popup, because in the popup they are not so distracting compared to permanently be shown in the view. Or alternatively, if they should only be quickfix actions, then the label could be changed from "[server name] Code Action: ..." to "[server name] Quick Fix: ..." (maybe also drop the "[server name]"?)

You are asking some hard questions :)

I think you are bringing up some good ones, but I think here we should focus on only one aspect of that - whether to include non-quick fixes in the popup or not. I don't want the scope to expand to semi-related improvements.

plugin/code_actions.py
rchl: Your solution is simpler and more readable. Thanks.

plugin/code_actions.py
```diff
@@ -7,227 +7,195 @@
...
+    The "kinds" include "kind" if "kind" matches one of the "kinds" exactly or one of the "kinds" is a prefix
+    of the whole "kind" (where prefix must be followed by a dot).
+    """
+    if not kind:
+        return False
+    for kinds_item in kinds:
+        if kinds_item == kind:
+            return True
+        kinds_item_len = len(kinds_item)
+        if kind.startswith(kinds_item) and kind[kinds_item_len] == '.':
```
jwortmann, line 194:
```suggestion
        if len(kind) > kinds_item_len and kind.startswith(kinds_item) and kind[kinds_item_len] == '.':
```

You need this check too, otherwise `kind[kinds_item_len]` could throw an IndexError

plugin/documents.py
```diff
@@ -590,13 +590,17 @@ def _on_sighelp_navigate(self, href: str) -> None:
...
-    def _on_code_actions(self, responses: CodeActionsByConfigName) -> None:
-        action_count = sum(map(len, responses.values()))
+        actions_manager \
+            .request_for_region_async(self.view, covering, diagnostics_by_config, manual=False) \
+            .then(self._on_code_actions)
+
+    def _on_code_actions(self, responses: List[CodeActionsByConfigName]) -> None:
+        # flatten list
+        action_lists = [actions for _, actions in responses if len(actions)]
+        all_actions = [action for actions in action_lists for action in actions]
```
predragnikolic, line 602: This line burned my brain :⁠-⁠D

rchl:
> I agree that explicitly triggered `lsp_code_actions` should bring up all code actions.

I've tried to also verify this logic with VSCode and got mixed results.

For marksman it doesn't report "Create Toc" source action when triggering code actions manually.
Though it could be because marksman doesn't report supported code action kinds in `codeActionProvider.codeActionKinds`. I didn't go through the trouble of verifying that by re-compiling the server with it added...

rchl:
I wish I could add tests for this filtering but we don't store results of `actions_manager.request_for_region_async` anywhere so it's not really possible right now.

rchl:
Some changes in hover popup:
 - Change "Code Action" to "Quick Fix"
 - Move ugly session name within square braces to the end of the line and make the text less prominent

Screenshot 2022-10-08 at 22 20 15

I don't think we'd want to remove the session name completely.

jwortmann:
Looks good!

I would probably even collect and combine the code actions from different servers into a single link, so that there is only one row with "Quick Fix: choose (x available)" in the popup. The server / config name is already shown in the annotation of the quick panel entries. If there is only one code action in total (for all sessions), then the full title to run it directly and dimmed config name after the title seems fine.

rchl:
> I would probably even collect and combine the code actions from different servers into a single link, so that there is only one row with "Quick Fix: choose (x available)" in the popup.

That could be a bit controversial so I would not do it here.

It can be that one server has single and very relevant quick fix and the other one has many not so relevant (something like in the screenshot). Then the access to the relevant one will be more difficult.

rchl:
Seems like there are no outstanding comments so merging.

Various potential improvements mentioned in comments can be considered separately.

PR #2086, opened 2022-10-12
Update diagnostics gutter icons and change default to "sign"

jwortmann:
This is a suggestion to ...

1. Change the default diagnostic icon style to "sign", to let users who doesn't bother with looking through and adjusting the setting values to enjoy the superior icon style by default :)
2. Replace the old icons with GitHub's [Octicons](https://github.com/primer/octicons), which imo look better and seem a little bit more sharp at the smallest icon size (they use a larger stroke width and I made them a little bit (1px) larger than the old icons).
Bonus: they use a much simpler license (MIT)
3. Never show the gutter icon for the "hint" severity (also for "dot" and other styles), which according to https://github.com/microsoft/language-server-protocol/issues/325#issuecomment-344570720 and other comments in that thread shouldn't really be rendered as diagnostic in the editor, but for example to fade out unused code. VSCode doesn't even draw the underline or show them in the problems panel, but here I kept the dotted underline due to the lack of a good ST API for foreground color dimming. At the moment, there is no visible difference between "info" and "hint" severity in Sublime LSP.

---

Comparison (on Full HD display with font size 10.5, and on the bottom with UI scale 2.0)

Before:

before
before2

After:

after
after2


rchl:
The new icons do overlap the line number a bit here. Possibly because of being 1px larger

Before:

Screenshot 2022-10-12 at 21 34 02

After:

Screenshot 2022-10-12 at 21 32 27

icons/convert.bat
```diff
@@ -1,12 +1,12 @@
...
-inkscape --export-type=png --export-area=-1:-1:18:18 -w 16 -h 16 --export-filename=info.png info.svg
-inkscape --export-type=png --export-area=-1:-1:18:18 -w 32 -h 32 --export-filename=info@2x.png info.svg
-inkscape --export-type=png --export-area=-1:-1:18:18 -w 48 -h 48 --export-filename=info@3x.png info.svg
-inkscape --export-type=png --export-area=-1:-1:18:18 -w 16 -h 16 --export-filename=lightbulb.png lightbulb.svg
-inkscape --export-type=png --export-area=-1:-1:18:18 -w 32 -h 32 --export-filename=lightbulb@2x.png lightbulb.svg
-inkscape --export-type=png --export-area=-1:-1:18:18 -w 48 -h 48 --export-filename=lightbulb@3x.png lightbulb.svg
-inkscape --export-type=png --export-area=-1:-1:18:18 -w 16 -h 16 --export-filename=warning.png warning.svg
-inkscape --export-type=png --export-area=-1:-1:18:18 -w 32 -h 32 --export-filename=warning@2x.png warning.svg
-inkscape --export-type=png --export-area=-1:-1:18:18 -w 48 -h 48 --export-filename=warning@3x.png warning.svg
+inkscape --export-type=png --export-area=-1.23:-1.23:18.46:18.46 -w 16 -h 16 --export-filename=error.png error.svg
```
rchl, line 1: I think WIndows users working on LSP are in minority but I guess those commands can easily be copied and executed in the terminal.

Though if you have a setup that allows running shell scripts then consider that instead. :)

rchl (approved):
Looks good to me but I would like to ask @rwols @predragnikolic if they have something against changes default value of `diagnostics_gutter_marker`.

predragnikolic (approved):
:+1:

predragnikolic:
Yeah, looks better.

Not related to this PR, 
I wish that the `"dot", "circle", "bookmark"` options were removed entirely :D (maybe leave dot or circle)

PR #2088, opened 2022-10-13
fix issues with restarting servers

rchl:
There were issues with restarting servers when there were more than one views using the same session and the corresponding server was restarted. Only the active view would start the session again and on other views the session would not start.

Fixed by using `WindowManager.restart_sessions_async` which re-registers all view listeners rather than the (internal) `WindowManager._end_sessions_async` and manually registering just the active view listener.

Also removed unnecessary `WindowManager._end_sessions_async` call after disabling server globally as ending the sessions is already handled on disabling the config (saving settings).

### DIFF
--- a/plugin/configuration.py
+++ b/plugin/configuration.py
@@ -46,8 +46,6 @@ class LspDisableLanguageServerGloballyCommand(sublime_plugin.WindowCommand):
         if index > -1:
             config_name = self._items[index]
             client_configs.disable(config_name)
-            wm = windows.lookup(self.window)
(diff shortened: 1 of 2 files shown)


predrag-codetribe:
> There were issues with restarting servers when there were more than one views using the same session and the corresponding server was restarted. Only the active view would start the session again and on other views the session would not start.

Can confirm that this does fix this issue.

predrag-codetribe (approved):

rchl:
I don't know who you are but I'll take it ;)

PR #2092, opened 2022-10-15
Add `$line` and `$character` expansion variables

merhard:
Some language server commands can only take the `line` and/or `character` integers individually instead of the entire `position` object. Allowing those variables to also be expanded adds to the flexibility of individual LSP plugins.

This was tested locally using https://github.com/sublimelsp/LSP-elixir and https://github.com/elixir-lsp/elixir-ls (specifically the `manipulatePipes` command: https://github.com/elixir-lsp/elixir-ls/blob/a4bd85e/apps/language_server/lib/language_server/providers/execute_command/manipulate_pipes.ex#L21)

### DIFF
--- a/docs/src/commands.md
+++ b/docs/src/commands.md
@@ -50,5 +50,7 @@ You can include special variables in the `command_args` array that will be autom
 | `"$selection_begin"` or `"${selection_begin}"` | int | Character offset of the begin of the (topmost) selection |
 | `"$selection_end"` or `"${selection_end}"` | int | Character offset of the end of the (topmost) selection |
 | `"$position"` or `"${position}"` | object | JSON object `{ 'line': int, 'character': int }` of the (topmost) cursor position, see [Position](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#position) |
+| `"$line"` or `"${line}"` | int | Zero-based line number of the (topmost) cursor position, see [Position](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#position) |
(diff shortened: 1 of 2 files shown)


docs/src/commands.md
```diff
@@ -50,5 +50,7 @@ You can include special variables in the `command_args` array that will be autom
 | `"$selection_begin"` or `"${selection_begin}"` | int | Character offset of the begin of the (topmost) selection |
 | `"$selection_end"` or `"${selection_end}"` | int | Character offset of the end of the (topmost) selection |
 | `"$position"` or `"${position}"` | object | JSON object `{ 'line': int, 'character': int }` of the (topmost) cursor position, see [Position](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#position) |
+| `"$line"` or `"${line}"` | int | Zero-based line number of the (topmost) cursor position, see [Position](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#position) |
+| `"$character"` or `"${character}"` | int | Zero-based character offset of the (topmost) cursor position, see [Position](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#position) |
```
rchl, line 54:
```suggestion
| `"$character"` or `"${character}"` | int | Zero-based character offset relative to the current line of the (topmost) cursor position, see [Position](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#position) |
```

merhard: Thanks for the suggestion! 💯 I added it to the PR.

predragnikolic (approved):
👍

rchl (approved):

PR #2094, opened 2022-10-21
Focus symbol closest to selection on showing document symbols

### LINKED ISSUE #2093: `lsp_document_symbols` should initially focus on current symbol under caret
Sublime's default implementation of Goto Symbol pops up with current symbol having focus in the dropdown list, this way I can use up/down arrows to go to prev/next symbols (since Sublime doesn't have goto next/prev symbol features built-in, this is imo a superior way to mimic that feature).

LSP is capable of finding all the symbols for a given language but the problem is, `lsp_document_symbols` commands focuses the first symbol in the file, so I can't easily use the arrow keys to browse the next/previous symbols.

So it would be better if LSP's replacement Goto Symbol popup focused the symbol of the current scope / the symbol under the caret.

rchl:
Match ST behavior and focus symbol closest to selection on triggering Go to Symbol command.

Taking end point of first selection and finding symbol that intersects it or is just before it.

I think ST might have slightly different behavior with multiple selections (takes into account last while we take first) but otherwise it should match ST behavior.

Fixes #2093

### DIFF
--- a/plugin/symbols.py
+++ b/plugin/symbols.py
@@ -94,7 +94,6 @@ class LspDocumentSymbolsCommand(LspTextCommand):
         super().__init__(view)
         self.old_regions = []  # type: List[sublime.Region]
         self.regions = []  # type: List[Tuple[sublime.Region, Optional[sublime.Region], str]]
-        self.is_first_selection = False
 
     def run(self, edit: sublime.Edit, event: Optional[Dict[str, Any]] = None) -> None:
         self.view.settings().set(SUPPRESS_INPUT_SETTING_KEY, True)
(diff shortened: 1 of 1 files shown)


plugin/symbols.py
```diff
@@ -109,13 +108,22 @@ def handle_response(self, response: Union[List[DocumentSymbol], List[SymbolInfor
         self.view.settings().erase(SUPPRESS_INPUT_SETTING_KEY)
         window = self.view.window()
         if window and isinstance(response, list) and len(response) > 0:
+            panel_items = self.process_symbols(response)
```
rchl, line 111: I hate code that creates side effects (self.regions) and also returns something but decided not to refactor anything here.

plugin/symbols.py
```diff
@@ -146,9 +154,6 @@ def on_symbol_selected(self, index: int) -> None:
         self.regions.clear()
 
     def on_highlighted(self, index: int) -> None:
-        if self.is_first_selection:
-            self.is_first_selection = False
-            return
```
rchl, line 151: We had code that prevent scrolling to selected symbol on opening the quick panel but with this behavior change I think it's no longer needed. Otherwise the behavior would be inconsistent as the outline would not show initially but only after changing selection.

predragnikolic (approved):
Tested, works.

PR #2098, opened 2022-10-23
Improvements for DocumentLink in hover popup

jwortmann:
I think it would be better for the popup on hover over a DocumentLink, to already indicate that it will be opened in a browser if that is the case. Otherwise it could be a bit surprising for users, because most of the time these are links to a location in another file.

-> Adjust label to "Open in Browser" if the target is not a file uri.

And in case there are multiple sessions at the same time, which all provide DocumentLinks, then combine them into a single link and open a quick panel with the target URIs to choose from when clicked, instead of just putting multiple links into the hover popup.

### DIFF
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -247,7 +247,7 @@ class LspHoverCommand(LspTextCommand):
             target = link.get("target")
             label = "Follow Link" if link.get("target", "file:").startswith("file:") else "Open in Browser"
             title = link.get("tooltip")
-            tooltip = 'title="{}"'.format(html.escape(title)) if title else ""
+            tooltip = ' title="{}"'.format(html.escape(title)) if title else ""
             region = range_to_region(link["range"], self.view)
             return '<a href="{}"{}>{}</a>'.format(html.escape(target), tooltip, label) if target else label, region
         else:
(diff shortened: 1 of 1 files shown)


plugin/hover.py
```diff
@@ -248,6 +233,26 @@ def symbol_actions_content(self, listener: AbstractViewListener, point: int) ->
...
+            for link in self._document_links[1:]:
+                combined_region = combined_region.cover(range_to_region(link["range"], self.view))
+            if all(link.get("target") for link in self._document_links):
+                return '<a href="quick-panel:DocumentLink">Follow Link…</a>', combined_region
+            else:
+                return "Follow Link…", combined_region
+        elif len(self._document_links) == 1:
+            link = self._document_links[0]
+            target = link.get("target")
+            label = "Follow Link" if link.get("target", "file:").startswith("file:") else "Open in Browser"
```
rchl, line 248: I would probably go with `Open in Browser` if starts with `http` and otherwise `Follow Link` because there could be cases where link would use a custom protocol like `virtual:` or something and it would be a link to a local content.

rchl: Hmm, but I see that `_on_navigate` handler would then have to be somehow changes to follow that logic...

rchl (approved):

PR #2100, opened 2022-10-26
Add elixir to language ids

predragnikolic:
relates to https://github.com/sublimelsp/LSP-tailwindcss/issues/42

### DIFF
--- a/language-ids.sublime-settings
+++ b/language-ids.sublime-settings
@@ -91,6 +91,7 @@
     "source.yaml.yarn": "yaml", // https://github.com/SublimeText/AFileIcon
     "text.advanced_csv": "csv", // https://github.com/SublimeText/AFileIcon
     "text.html.basic": "html",
+    "text.html.elixir": "html",  // https://github.com/elixir-editors/elixir-tmbundle
     "text.html.markdown.academicmarkdown": "markdown", // https://github.com/mangecoeur/AcademicMarkdown
     "text.html.markdown.license": "markdown", // https://github.com/SublimeText/AFileIcon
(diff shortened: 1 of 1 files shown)


language-ids.sublime-settings
```diff
@@ -91,6 +91,7 @@
     "source.yaml.yarn": "yaml", // https://github.com/SublimeText/AFileIcon
     "text.advanced_csv": "csv", // https://github.com/SublimeText/AFileIcon
     "text.html.basic": "html",
+    "text.html.elixir": "html",  // https://github.com/elixir-editors/elixir-tmbundle
```
rchl, line 94: Kinda feels weird to map it to `html`. For tailwindcss I guess it makes sense but not sure it would make sense if there would exist elixir-specific server. Feels like this is something that should be possible to define per-server maybe.

rchl (approved):

language-ids.sublime-settings
rwols: There's also `source.elixir`, I think this change doesn't hurt https://github.com/sublimelsp/LSP-elixir/blob/master/LSP-elixir.sublime-settings

predragnikolic: The elixir package has multiple syntaxes. 
https://github.com/elixir-editors/elixir-tmbundle/tree/master/Syntaxes

the ["text.html.elixir"](https://github.com/elixir-editors/elixir-tmbundle/blob/master/Syntaxes/HTML%20(EEx).tmLanguage) scope is meant for `*.html.heex` files.

https://user-images.githubusercontent.com/9070844/178098566-d4b34773-2a42-42bc-82cb-bc89af583d56.png

PR #2106, opened 2022-10-30
Fix diagnostics underline color priority

### LINKED ISSUE #2103: Wrong underline color in case of multiple diagnostics with different severities
**Describe the bug**
Diagnostics underline color doesn't use the highest severity if there are multiple diagnostics with different severity

**To Reproduce**
1. Install LSP-pyright (for example)
2. Create and open a Python file
3. Type `im`
4. Observe that there are two diagnostics, one "error" and one "warning", but the squiggly underline is orange

**Screenshots**

severity

**Additional context**
I just wanted to log this issue here that I incidentally noticed, and I don't remember it being discussed before. But unfortunately I think it is impossible to solve (*edit*: possible workaround see below). The reason lies in ST's behavior for the `add_regions` draw order, which is different for icons and for inline decorations (underline). See here:

https://github.com/sublimelsp/LSP/blob/deeb0cbafce3deba4a37af47f9629b52179f8075/plugin/session_view.py#L107-L115

and

https://github.com/sublimelsp/LSP/blob/deeb0cbafce3deba4a37af47f9629b52179f8075/plugin/session_view.py#L130

If we reverse the range, then the underline would have the correct color (highest severity), but the icon would be the one with the lowest severity. So I think it's probably better to have the correct icon and potentially wrong underline color.

---

*Edit*: Perhaps a workaround could be to create two region types for each diagnostic severity, one for the icon and one for the underline, and initialize the icon regions in reversed order. But not sure whether the additional code needed for that is worth it.

jwortmann:
Fixes #2103

### DIFF
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -80,8 +80,10 @@ class SessionView:
             self.session.unregister_session_view_async(self)
         self.session.config.erase_view_status(self.view)
         for severity in reversed(range(1, len(DIAGNOSTIC_SEVERITY) + 1)):
-            self.view.erase_regions(self.diagnostics_key(severity, False))
-            self.view.erase_regions(self.diagnostics_key(severity, True))
+            self.view.erase_regions("{}_icon".format(self.diagnostics_key(severity, False)))
+            self.view.erase_regions("{}_underline".format(self.diagnostics_key(severity, False)))
+            self.view.erase_regions("{}_icon".format(self.diagnostics_key(severity, True)))
+            self.view.erase_regions("{}_underline".format(self.diagnostics_key(severity, True)))
         self.view.erase_regions("lsp_document_link")
         self.session_buffer.remove_session_view(self)
 
@@ -134,7 +136,10 @@ class SessionView:
         self.view.add_regions("lsp_document_link", r)
         for severity in range(1, 5):
             for mode in line_modes:
-                self.view.add_regions("lsp{}d{}{}".format(self.session.config.name, mode, severity), r)
+                self.view.add_regions("lsp{}d{}{}_icon".format(self.session.config.name, mode, severity), r)
+        for severity in range(4, 0, -1):
+            for mode in line_modes:
+                self.view.add_regions("lsp{}d{}{}_underline".format(self.session.config.name, mode, severity), r)
         if document_highlight_style in ("underline", "stippled"):
             for kind in document_highlight_kinds:
                 for mode in line_modes:
@@ -277,6 +282,7 @@ class SessionView:
             listener.on_diagnostics_updated_async()
 
     def _draw_diagnostics(self, severity: int, max_severity_level: int, flags: int, multiline: bool) -> None:
+        ICON_FLAGS = sublime.HIDE_ON_MINIMAP | sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE
         key = self.diagnostics_key(severity, multiline)
         key_tags = {tag: '{}_tags_{}'.format(key, tag) for tag in DIAGNOSTIC_TAG_VALUES}
         for key_tag in key_tags.values():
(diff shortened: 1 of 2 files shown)


plugin/session_view.py
```diff
@@ -80,8 +80,10 @@ def on_before_remove(self) -> None:
             self.session.unregister_session_view_async(self)
         self.session.config.erase_view_status(self.view)
         for severity in reversed(range(1, len(DIAGNOSTIC_SEVERITY) + 1)):
-            self.view.erase_regions(self.diagnostics_key(severity, False))
-            self.view.erase_regions(self.diagnostics_key(severity, True))
+            self.view.erase_regions("{}_icon".format(self.diagnostics_key(severity, False)))
+            self.view.erase_regions("{}_underline".format(self.diagnostics_key(severity, False)))
+            self.view.erase_regions("{}_icon".format(self.diagnostics_key(severity, True)))
+            self.view.erase_regions("{}_underline".format(self.diagnostics_key(severity, True)))
```
predragnikolic, line 86: Hmm,
I kind of liked the previous code more (mostly because it was less code)
In the prev code, things were done twice, with this change things needs to be handled x4. 

Is there a way to fix it differently with out introducing new regions for `_icon` and `_underline`?

jwortmann: I can't really think of another way how it could be solved in practice. The problem is basically that ST behaves inconsistent in regard to which of the gutter icons and which of the inline decorations it will render (on top) when there are multiple regions overlapping. If ST would make this behavior consistent, or give the plugin more control to for example explicitly set the priority for regions, then this workaround wouldn't be required.

In theory another approach could be to compare the different diagnostics for a file, and only draw the single one with the highest severity, if there are multiple diagnostics with different severities and the same range. But I imagine that this would be quite computationally expensive. And the ranges could also be partially overlapping, then you would have to calculate the overlapping range and potentially generate new ranges by subtracting them, etc., so I think this would be infeasable (not even considering diagnostics from multiple sessions...).

If you have another idea, suggestions are welcome, of course.

predragnikolic: I was also thinking about sorting the diagnostics by severity and only draw the first with the highest severity,
but you have a good point with:
> And the ranges could also be partially overlapping, then you would have to calculate the overlapping range and potentially generate new ranges by subtracting them, etc.,

rchl (approved):

predragnikolic:
Is it expected to have a hint icon in the gutter(coming from LSP-cspell), one char  yellow underline(coming from diagnostics-ls)?
The good thing is that the error underline(coming from LSP-pyright) is drown on top of the other yellow underline(coming from LSP-pyright).

image
On the first line, I will see the hint in the gutter, one char wide line for yellow diagnostics (coming from diagnostics-ls for)

I have LSP-cspell & LSP-pyright & diagnostics-ls (flake8) running.
<details>
<summary>diagnostics-ls client configuration</summary>

```jsonc
    "diagnostic-ls": {
			"enabled": true,
			"command": [
				"diagnostic-languageserver",
				"--stdio",
				"--log-level",
				"2"
			],
			"selector": "source.python",
			"initializationOptions": {
				"linters": {
					"cspell": {
						"command": "cspell",
						"debounce": 100,
						"args": [
							"%filepath"
						],
						"sourceName": "cspell",
						"formatLines": 1,
						"formatPattern": [
							// the error message looks like this:
							// /home/predrag/.config/sublime-text/Packages/Sone/index.py:1:1 - Unknown word (funer)
							".*?:(\\d+):(\\d+)\\s*-\\s*(.*)",
							{
								"line": 1,
								"column": 2,
								"message": 3,
							}
						],
					},
					"flake8": {
						"command": "flake8",
						"debounce": 100,
						"args": [
							"--format",
							"%(row)d,%(col)d,%(code).1s,%(text)s [%(code)s]",
							"-"
						],
						"sourceName": "flake8",
						"formatLines": 1,
						"formatPattern": [
							"(\\d+),(\\d+),()([A-Z]),(.*)",
							{
								"line": 1,
								"column": 2,
								"security": 3, // Match a zero-length group so everything can be classified as a Warning rather than Error
								"message": 5,
							}
						],
						"securities": {
							"": "warning",
						},
					},
				},
				"formatters": {},
				"filetypes": {
					// "python": "cspell",
					// "python": "cspell",
					"python": "flake8",
				},
			},
		},
	}
```
</details>

jwortmann:
I haven't tested it myself, but I assume the icon for the "info" severity is shown in this case because you have more than one session and the region keys are initialized separately per session at the time when the corresponding SessionView is created. So all of the icons for session A take priority over all icons for session B. It might even be random if you restart the servers, and probably depends on some internal `for` loop over the sessions.

Ideally it would show the error icon on line 1, which has all of "error", "warning" and "info" diagnostics.
If we want to fix it, the function to initialize the region keys would have to be moved out of the SessionView class to some other place with a loop over all sessions for each of the severities.

But all of that should not be affected by this pull request. The initialize order for the icons is still the same here, I only changed the region key string to have a "_icon" suffix, and the `flags` parameter when drawing the icons. So you should probably see the same icons if you checkout the "main" branch.

predragnikolic (approved):
...in short I think I'm wrong.

So I do not want to stop this PR :)

PR #2107, opened 2022-10-30
Fix diagnostics_additional_delay_auto_complete_ms not working after split view

### LINKED ISSUE #2102: diagnostics_additional_delay_auto_complete_ms doesn't work after split view
**To Reproduce**
1. Install LSP-pyright (for example)
2. Set `"diagnostics_additional_delay_auto_complete_ms": 5000` is LSP settings
3. Create and open a Python file
4. Type `im`
5. Confirm that there is an auto-completion popup and that the diagnostic(s) ("im" is not defined) appear after a delay of 5 seconds - so far so good
6. Remove the text and use *File > Split View*
7. Select the left (original) view
8. Type `im`
9. Observe that the diagnostic(s) immediately appears 

**Expected behavior**
Diagnostics appear only after the delay specified in "diagnostics_additional_delay_auto_complete_ms"

- OS: Windows 11
- Sublime Text version: 4126
- LSP version: main branch

**Additional context**
Looks like the bug is here:
https://github.com/sublimelsp/LSP/blob/deeb0cbafce3deba4a37af47f9629b52179f8075/plugin/session_buffer.py#L439-L443

By using `SessionBuffer.some_view()`, a random view is checked for visibility of the autocompletion popup. But instead it should use the *active* view, i.e. the view where I'm currently typing.

jwortmann:
Fixes #2102

### DIFF
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -322,9 +322,15 @@ class SessionBuffer:
                 self.session.send_notification(did_save(view, include_text, self.last_known_uri))
 
     def some_view(self) -> Optional[sublime.View]:
+        if not self.session_views:
+            return None
+        # Prefer active view if possible
+        active_view = self.session.window.active_view()
+        for sv in self.session_views:
+            if sv.view == active_view:
+                return active_view
         for sv in self.session_views:
             return sv.view
-        return None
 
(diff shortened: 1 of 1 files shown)


rchl (approved):

rchl:
Noticed that also semantic highlighting is a bit buggy with split view (on initial split) but that's unrelated issue.

PR #2108, opened 2022-10-31
Don't trigger code action requests for background views

### LINKED ISSUE #2101: Don't trigger code action requests for background documents
Editing a file can trigger diagnostics for other open files in which case we'll do a, potentially expensive, code action requests in all of them.

Try to delay code action requests until view is visible.

rchl:
Servers like typescript trigger `textDocument/publishDiagnostics` for all open files on making any changes. Our `DocumentSyncListener` triggered [code action request](https://github.com/sublimelsp/LSP/blob/439b01efe862497ee67dedb7d7d9662521c12537/plugin/documents.py#L285-L286) in that case.

Changed so that code action request is only made if the view is visible.

Since the code action response can be affected by changed diagnostics, make sure that the code action request is then also made when view is activated later.

We should have been triggering code action request on view activation anyway as otherwise we have relied on `textDocument/publishDiagnostics` triggering it which is not guaranteed.

Also made sure that `on_load_async` triggers `on_activated_async` which triggers code actions/code lenses/semantic tokens because when file is initially focused the view is not registered yet so we have missed all the logic currently in `on_activated_async`.

Fixes #2101

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -68,7 +68,7 @@ from .types import DocumentSelector
 from .types import method_to_capability
 from .types import SettingsRegistration
 from .types import sublime_pattern_to_glob
-from .typing import Callable, cast, Dict, Any, Optional, List, Tuple, Generator, Iterable, Type, Protocol, Mapping, TypeVar, Union  # noqa: E501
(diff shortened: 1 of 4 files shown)


rchl:
Will take no feedback as "no issues" :)

PR #2109, opened 2022-10-31
Nicer presentation for "find references/definition" quick panel

rchl:
Before:

Screenshot 2022-10-31 at 22 29 01

After:

Screenshot 2022-10-31 at 22 28 11

### DIFF
--- a/plugin/locationpicker.py
+++ b/plugin/locationpicker.py
@@ -5,6 +5,7 @@ from .core.sessions import Session
 from .core.typing import Union, List, Optional, Tuple
 from .core.views import get_uri_and_position_from_location
 from .core.views import location_to_human_readable
+from .core.views import SYMBOL_KINDS, SymbolKind
 from .core.views import to_encoded_filename
 from urllib.request import url2pathname
 import functools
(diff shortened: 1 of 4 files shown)


predragnikolic (approved):

plugin/locationpicker.py
```diff
@@ -77,8 +78,14 @@ def __init__(
         manager = session.manager()
         base_dir = manager.get_project_path(view.file_name() or "") if manager else None
         self._window.focus_group(group)
+        config_name = session.config.name
+        kind = SYMBOL_KINDS[SymbolKind.File]
```
jwortmann, line 82: I have some doubts about using the "File" kind icon here, because...

1. It's not really useful to show the icon if there is only one possible kind for it and all items always have this kind - it doesn't provide any new information in this case
2. The "f" letter could easily be mistaken to mean "function", since functions are a common, but not necessarily the only kind of symbol where you would use "Goto Definition" or "Find references" for
3. It's not even correct in general, because the returned locations from the response just have some DocumentUri, which isn't necessarily a "file:" URI. It could also be a "res:" (technically being inside a (.sublime-package) file on disk, but not in the sense one would really expect for the "File" kind) or "buffer:". And `location_to_human_readable` would also handle e.g. web URIs.

So I'd suggest to either don't use "kind" for the items at all, or differentiate the kind based on the location's URI scheme (meaning to add a few new kinds for the other possible schemes).

rchl: I guess the proper thing to do (and what native "Go to symbol" does) would be to show the kind of the symbol itself but we likely don't have that info readily available from LSP or ST itself...

jwortmann: I made another PR #2110 to remove the kind and add the placeholder text from the builtin "Goto Definition..." and "Goto Reference..." quick panels.

PR #2110, opened 2022-11-02
Placeholder text

jwortmann:
Don't get confused from the title ;-)

This is just a pull request with small improvements to the branch of the other pull request.

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -174,6 +174,17 @@ CODE_ACTION_KINDS = {
     CodeActionKind.Source: KIND_SOURCE
 }  # type: Dict[CodeActionKind, SublimeKind]
 
+# Symbol scope to kind mapping, based on https://github.com/sublimetext-io/docs.sublimetext.io/issues/30
+SYMBOL_SCOPE_KINDS = {
+    "keyword | storage.modifier | storage.type | keyword.declaration | variable.language | constant.language": sublime.KIND_KEYWORD,  # noqa: E501
+    "entity.name.type | entity.name.class | entity.name.enum | entity.name.trait | entity.name.struct | entity.name.impl | entity.name.interface | entity.name.union | support.type | support.class": sublime.KIND_TYPE,  # noqa: E501
+    "entity.name.function | entity.name.method | entity.name.macro | meta.method entity.name.function | support.function | meta.function-call variable.function | meta.function-call support.function | support.method | meta.method-call variable.function": sublime.KIND_FUNCTION,  # noqa: E501
+    "entity.name.module | entity.name.namespace | support.module | support.namespace": sublime.KIND_NAMESPACE,
+    "entity.name.definition | entity.name.label | entity.name.section": sublime.KIND_NAVIGATION,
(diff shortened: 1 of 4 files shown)


plugin/core/views.py
```diff
@@ -327,6 +327,10 @@ def position(view: sublime.View, offset: int) -> Position:
     return offset_to_point(view, offset).to_lsp()
 
 
+def word(view: sublime.View, offset: int) -> str:
+    return view.substr(view.expand_by_class(offset, sublime.CLASS_WORD_START | sublime.CLASS_WORD_END, "[]{}()<>:."))
```
jwortmann, line 331: This is the same what Default/symbol.py uses at a first attempt to get the word at the current cursor position.

I'm not sure why exactly this is preferred over `view.word()`, but I thought it would be better to keep it consistent with the builtin implementation. It would for example include string quotation marks, while `view.word()` would strap them off.

rchl: Is it actually consistent with builtin implementation? Check the placeholder text in the second case, when the cursor is a the beginning of the class property:

https://user-images.githubusercontent.com/153197/199826271-2e387b01-a6f3-4158-a4b6-15ecc868d1df.mov

Built-in implementation can't actually show references for this symbol but it correctly shows "Unable to find capability" in the status field.

rchl:
It would be nice to show the kind icon because without it the list looks kinda incomplete or less pleasing. To me at least.

But if we can't find a way to grab the symbol then I guess we have to go with that.

plugin/core/views.py
rchl: I guess looking at default implementation explains it:

```py
def reference_at_point(view, pt):
    symbol = view.substr(view.expand_by_class(pt, sublime.CLASS_WORD_START | sublime.CLASS_WORD_END, "[]{}()<>:."))
    locations = lookup_references(view.window(), symbol)

    if len(locations) == 0:
        symbol = view.substr(view.word(pt))
        locations = lookup_references(view.window(), symbol)

    return symbol, locations
```

jwortmann:
Please try again with the latest changes. It uses a heuristic based on the scope name at the original position now. So the icon is the same for all entries in the quick panel, but that should be fine because the definitions or references are supposed to be for the exact same symbol.

rchl:
The scope-based kind is a bit random.
For example when cursor is at the end of a word (reference) then ST scopes are relevant since those describe character on the right. So that means that depending on where the cursor is, scope might or might not show up (or possibly even show two different kinds depending on the position).

But I don't think I really mind that to be honest...

rchl:
Though built-in "goto reference" does seem to handle that case correctly regardless...

For example with LSP-typescript in `test.js`:

```ts
function foo() {}

foo()
foo()
```

and cursor at `foo|()`.

rchl:
> Though built-in "goto reference" does seem to handle that case correctly regardless...

ST uses native API to figure out `kind` and we can't really see what it's doing.

What do you think about taking position, expanding to "word" and getting scope from the beginning of that range?

jwortmann:
> What do you think about taking position, expanding to "word" and getting scope from the beginning of that range?

Seems like a good idea and probably the best we can do here.

PR #2113, opened 2022-11-03
Ignore diagnostics for files in folder_exclude_patterns

jwortmann:
Files within folders excluded by `"folder_exclude_patterns"` are not shown in the sidebar and can't be opened via "Goto Anything...", so it's probably better to ignore diagnostics for those files.

The line which was removed in the diff did basically nothing, because fnmatch would never match just a folder name or path segment from the "folder_exclude_patterns" if not adapted to glob pattern conventions.

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -378,6 +378,8 @@ class WindowManager(Manager):
                 for pattern in folder.get('folder_exclude_patterns', []):
                     if pattern.startswith('//'):
                         patterns.append(sublime_pattern_to_glob(pattern, True, folder['path']))
+                    elif pattern.startswith('/'):
+                        patterns.append(sublime_pattern_to_glob(pattern, True))
                     else:
                         patterns.append(sublime_pattern_to_glob('//' + pattern, True, folder['path']))
                         patterns.append(sublime_pattern_to_glob('//**/' + pattern, True, folder['path']))
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -24,6 +24,7 @@ from .settings import userprefs
 from .transports import create_transport
 from .types import ClientConfig
 from .types import matches_pattern
+from .types import sublime_pattern_to_glob
 from .typing import Optional, Any, Dict, Deque, List, Generator, Tuple, Union
 from .url import parse_uri
 from .views import extract_variables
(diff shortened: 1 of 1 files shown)


plugin/core/windows.py
```diff
@@ -370,7 +371,17 @@ def should_present_diagnostics(self, uri: DocumentUri) -> Optional[str]:
...
+        patterns = [sublime_pattern_to_glob(pattern, True) for pattern in settings.get("folder_exclude_patterns") or []]
+        project_data = self.window.project_data()
+        if project_data:
+            for folder in project_data.get('folders', []):
+                for pattern in folder.get('folder_exclude_patterns', []):
+                    if pattern.startswith('//'):
+                        patterns.append(sublime_pattern_to_glob(pattern, True, folder['path']))
+                    else:
+                        patterns.append(sublime_pattern_to_glob('//' + pattern, True, folder['path']))
+                        patterns.append(sublime_pattern_to_glob('//**/' + pattern, True, folder['path']))
```
jwortmann, line 383: I see that this probably doesn't work if the pattern starts with just a single slash. I will take another look later.

rchl (approved):
`file_exclude_patterns` and `binary_file_patterns` could use the same treatment but you it doesn't need to be done here if you don't feel like it.

jwortmann:
There are a lot of other bugs in the `should_present_diagnostics` function.

For example it doesn't work if diagnostics arrive at the time when a HtmlSheet has currently focus, because `window.active_view()` returns `None` in that case:
https://github.com/sublimelsp/LSP/blob/fe9fbd1f031cd0993d87eeac691e5c0993a1a9f6/plugin/core/windows.py#L365-L367

Diagnostics for files not in the workspace folders are ignored, but not if the scheme isn't "file:"
https://github.com/sublimelsp/LSP/blob/fe9fbd1f031cd0993d87eeac691e5c0993a1a9f6/plugin/core/windows.py#L361-L364
So if I use "View Package File" from the command palette to open a Python file from the Default package, it still shows me all the diagnostics from pyright, even though it's not part of the workspace and I can't edit this file.

"binary_file_patterns" can probably be ignored, because there are no lines or columns in binary files, so language servers can't send any diagnostics for them.

The function name is misleading, it returns a truish value if diagnostics should be ignored, and a falsey value if not. So it should better be named "should_not_present_diagnostics" or maybe "should_ignore_diagnostics".

Yeah, "file_exclude_patterns" could need the same treatment like in the change here, but then "file_include_patterns" (and "folder_include_patterns") should ideally also be taken into account.

This function should be rewritten from scratch... but not in this PR.

PR #2114, opened 2022-11-04
Remove unnecessary is_range_equal function

jwortmann:
Python has this actually built-in for dict/TypedDict.


rchl (approved):
I suppose it's a remnant from the times when we had a Range class.

PR #2115, opened 2022-11-09
Make Goto Diagnostic overlays consistent

jwortmann:
At the moment, "Goto Diagnostic..." opens a slightly different overlay compared to invoking "Goto Diagnostic in Project..." and selecting a file. Doing the latter will show the filename in the command palette input and when pressing <kbd>Backspace</kbd> it will go back to the list of files, which is not the case for "Goto Diagnostic...".

This PR would make both of the commands to show a consistent overlay by adding the file preselected to the input field for "Goto Diagnostic...".

Feel free to discuss or reject if this implementation feels too dubious/hackish...

### DIFF
--- a/plugin/goto_diagnostic.py
+++ b/plugin/goto_diagnostic.py
@@ -89,11 +89,11 @@ class PreselectedListInputHandler(sublime_plugin.ListInputHandler, metaclass=ABC
     Similar to ListInputHandler, but allows to preselect a value like some of the input overlays in Sublime Merge.
     Inspired by https://github.com/sublimehq/sublime_text/issues/5507.
 
-    Subclasses of PreselectedListInputHandler must implement the _list_items() method, i.e. just prepend the regular
-    list_items() with an underscore.
+    Subclasses of PreselectedListInputHandler must not implement the `list_items` method, but instead `_list_items`,
+    i.e. just prepend an underscore to the regular `list_items`.
 
     When an instance of PreselectedListInputHandler is created, it must be given the window as an argument.
-    An optional second argument initial_value can be provided to preselect a value.
+    An optional second argument `initial_value` can be provided to preselect a value.
     """
 
     _window = None  # type: Optional[sublime.Window]
--- a/plugin/goto_diagnostic.py
+++ b/plugin/goto_diagnostic.py
@@ -4,12 +4,14 @@ from .core.registry import windows
 from .core.sessions import Session
 from .core.settings import userprefs
 from .core.types import ClientConfig
-from .core.typing import Dict, Iterable, Iterator, List, Optional, Tuple, Union
+from .core.typing import Any, Dict, Iterable, Iterator, List, Optional, Tuple, Union
 from .core.url import parse_uri, unparse_uri
 from .core.views import DIAGNOSTIC_KINDS
 from .core.views import MissingUriError, uri_from_view, get_uri_and_position_from_location, to_encoded_filename
 from .core.views import format_diagnostic_for_html
 from .core.views import diagnostic_severity, format_diagnostic_source_and_code, format_severity
+from abc import ABCMeta
+from abc import abstractmethod
 from collections import Counter, OrderedDict
 from pathlib import Path
 import functools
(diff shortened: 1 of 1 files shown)


plugin/goto_diagnostic.py
```diff
@@ -67,6 +69,7 @@ def input(self, args: dict) -> Optional[sublime_plugin.CommandInputHandler]:
         if uri == "$view_uri":
             try:
                 uri = uri_from_view(view)
+                return DiagnosticUriInputHandler(self.window, view, uri)
```
rchl, line 72: Can we move this below the try/except block? I guess we don't expect it to throw so we shouldn't handle that gracefully.

plugin/goto_diagnostic.py
```diff
@@ -77,18 +80,61 @@ def input_description(self) -> str:
...
+    def list_items(self) -> ListItemsReturn:
+        if self._preselect and self._initial_value is not None and self._window is not None:
+            self._preselect = False
+            sublime.set_timeout(functools.partial(self._window.run_command, 'select'))
+            return [self._initial_value], 0  # pyright: ignore[reportGeneralTypeIssues]
+        else:
+            return self._list_items()
+
+    @abstractmethod
+    def _list_items(self) -> ListItemsReturn:
```
rchl, line 121: Rather than having two "list_items", I think I would prefer calling it `get_list_items`. And whether there should be a leading underscore or not, I'm not sure so I'm fine with either.

plugin/goto_diagnostic.py
```diff
@@ -77,18 +80,61 @@ def input_description(self) -> str:
...
+        super().__init__()
+        if initial_value is not None:
+            self._window = window
+            self._initial_value = initial_value
+            self._preselect = True
+
+    def list_items(self) -> ListItemsReturn:
+        if self._preselect and self._initial_value is not None and self._window is not None:
+            self._preselect = False
+            sublime.set_timeout(functools.partial(self._window.run_command, 'select'))
```
rchl, line 115: Could/should we drop `_preselect` completely and just reset the `self._initial_value` to None? We'd just add a method here that selects and resets it:

```py
    def _select_and_reset_initial_value(self) -> None:
        if self._window:
            self._window.run_command('select')
            self._initial_value = None
```

plugin/goto_diagnostic.py
```diff
@@ -77,18 +80,61 @@ def input_description(self) -> str:
...
+
+    Subclasses of PreselectedListInputHandler must not implement the `list_items` method, but instead `_list_items`,
+    i.e. just prepend an underscore to the regular `list_items`.
+
+    When an instance of PreselectedListInputHandler is created, it must be given the window as an argument.
+    An optional second argument `initial_value` can be provided to preselect a value.
+    """
+
+    _window = None  # type: Optional[sublime.Window]
+    _initial_value = None  # type: Optional[Union[str, sublime.ListInputItem]]
```
rchl, line 100: Why would we declare those on the class object rather than just assigning in the constructor?

plugin/goto_diagnostic.py
jwortmann: I originally had the `sublime.set_timeout` logic within `__init__` and moved it into `list_items` later. I think that was the reason why I would need another variable `_preselect`, but now it's not needed anymore and your suggestion works fine too. 👍

jwortmann:
I think I've addressed all comments.

jwortmann:
I've noticed that it doesn't look very nice if you open just a single file (without workspace folder): instead of the filename it will show "None" in the command palette input. But this is unrelated to this PR, it also happens on the "main" branch if you run "Goto Diagnostic in Project..." in single file mode. It is just more obvious with this PR.

I think the bug is caused by the following:

This is the function which defines the label:
https://github.com/sublimelsp/LSP/blob/0267a1a2a6d184a9df1135892792e936462f5832/plugin/goto_diagnostic.py#L135-L136

and it calls `_project_path`, which returns `str(project_path( ... )) or path` for `file:` URIs.

https://github.com/sublimelsp/LSP/blob/0267a1a2a6d184a9df1135892792e936462f5832/plugin/goto_diagnostic.py#L159-L163

But the `project_path` function returns `None` if there is no window folder: https://github.com/sublimelsp/LSP/blob/0267a1a2a6d184a9df1135892792e936462f5832/plugin/goto_diagnostic.py#L263-L269

So we end up with `str(None) or path`, which will display as literal "None", i.e. the `or path` which is probably meant as a fallback, doesn't work at all.

I guess I can add a fix to this PR.

rchl (approved):

PR #2119, opened 2022-11-18
Don't show a source if diagnostic doesn't have a source

Sainan:
Looks better & it's the same as what VS Code does.

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -853,7 +853,6 @@ def format_diagnostic_for_panel(diagnostic: Diagnostic) -> Tuple[str, Optional[i
     """
     formatted, code, href = diagnostic_source_and_code(diagnostic)
     lines = diagnostic["message"].splitlines() or [""]
-    # \u200B is the zero-width space
     result = " {:>4}:{:<4}{:<8}{}".format(
         diagnostic["range"]["start"]["line"] + 1,
         diagnostic["range"]["start"]["character"] + 1,
@@ -861,6 +860,7 @@ def format_diagnostic_for_panel(diagnostic: Diagnostic) -> Tuple[str, Optional[i
         lines[0]
     )
     if formatted != "":
+        # \u200B is the zero-width space
         result += " \u200B{}".format(formatted)
     offset = len(result) if href else None
     for line in itertools.islice(lines, 1, None):
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -855,7 +855,7 @@ def format_diagnostic_for_panel(diagnostic: Diagnostic) -> Tuple[str, Optional[i
         format_severity(diagnostic_severity(diagnostic)),
         lines[0]
     )
-    if formatted != "":
+    if formatted != "" or code != "":
         # \u200B is the zero-width space
         result += " \u200B{}".format(formatted)
     offset = len(result) if href else None
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -975,9 +975,9 @@ def format_diagnostic_for_html(
     else:
         code = None
     source = diagnostic.get("source")
-    if source is not None or code:
+    if source or code:
         formatted.extend(" ")
-    if source is not None:
+    if source:
         formatted.extend(_with_color(source, "color(var(--foreground) alpha(0.6))"))
     if code:
         formatted.extend(code)
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -868,25 +868,23 @@ def format_diagnostic_source_and_code(diagnostic: Diagnostic) -> str:
     formatted, code, href = diagnostic_source_and_code(diagnostic)
     if href is None or code is None:
         return formatted
-    return formatted + code
+    return formatted + "(" + code + ")"
 
 
 def diagnostic_source_and_code(diagnostic: Diagnostic) -> Tuple[str, Optional[str], Optional[str]]:
-    formatted = []
-    source = diagnostic.get("source")
-    if source is not None:
-        formatted = [source]
+    formatted = diagnostic.get("source")
+    if formatted is None:
+        formatted = ""
     href = None
     code = diagnostic.get("code")
     if code is not None:
         code = str(code)
-        formatted.append(":")
         code_description = diagnostic.get("codeDescription")
         if code_description:
             href = code_description["href"]
         else:
-            formatted.append(code)
-    return "".join(formatted), code, href
+            formatted += "(" + code + ")"
+    return formatted, code, href
 
 
 def location_to_human_readable(
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -872,9 +872,7 @@ def format_diagnostic_source_and_code(diagnostic: Diagnostic) -> str:
 
 
 def diagnostic_source_and_code(diagnostic: Diagnostic) -> Tuple[str, Optional[str], Optional[str]]:
-    formatted = diagnostic.get("source")
-    if formatted is None:
-        formatted = ""
+    formatted = diagnostic.get("source", "")
     href = None
     code = diagnostic.get("code")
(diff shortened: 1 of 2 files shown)


plugin/core/views.py
```diff
@@ -854,13 +854,14 @@ def format_diagnostic_for_panel(diagnostic: Diagnostic) -> Tuple[str, Optional[i
     formatted, code, href = diagnostic_source_and_code(diagnostic)
     lines = diagnostic["message"].splitlines() or [""]
     # \u200B is the zero-width space
```
rwols, line 856: Please move this comment as well.

rchl:
Needs some more to address various places that show that info.

For example the diagnostics quick panel shows some stray `:` in the list and somewhat weirdly spaced `:` in the annotation below the list:

Screenshot 2022-11-20 at 15 01 48

Hover info also missing space in that case:

Screenshot 2022-11-20 at 15 07 53

For reference, diagnostics panel looks like this without `source` but with `code`. 

Screenshot 2022-11-20 at 15 02 11

plugin/core/views.py
```diff
@@ -837,8 +837,8 @@ def diagnostic_severity(diagnostic: Diagnostic) -> DiagnosticSeverity:
     return diagnostic.get("severity", DiagnosticSeverity.Error)
 
 
-def diagnostic_source(diagnostic: Diagnostic) -> str:
-    return diagnostic.get("source", "unknown-source")
+def diagnostic_source(diagnostic: Diagnostic) -> Optional[str]:
+    return diagnostic.get("source")
```
rchl, line 841: Not really that useful to keep this function now that it doesn't have any logic besides getting dict property.

rchl:
I would probably go with a much simpler solution that uses server name (`configuration.name`) when source is missing. As you can see from my previous comment, this is riddled with edge cases because `source` is entangled with other diagnostic properties.

Sainan:
I'd like to get this behaviour right because there is no need to show a source and getting this behaviour to be closer to VS Code would be better for overall "LSP compliance" I suppose.

Gonna dump some comparisons between LSP 1.20 and VS Code here to highlight the differences.

no "source" or "code"
image
image


"code", no "source"
image
image

"source" and "code"
image
image

I'll try my best to patch this into the LSP plugin.

rchl:
> I'll try my best to patch this into the LSP plugin.

Then maybe LSP should follow VSCode here and use parenthesis for `code` since it perhaps wouldn't be so clear otherwise when there is `code` but not `source`.

Sainan:
Using parenthesis now. Only caveat is that the link in the diagnostics panel includes the parenthesis:
image
image

rchl:
Check the preview at the bottom here:

Screenshot 2022-11-20 at 19 41 37

There should be some contrast between the message and the source/code part since now it looks like part of the sentence.

rchl:
> Using parenthesis now. Only caveat is that the link in the diagnostics panel includes the parenthesis:

You should be able to adjust the syntax file to address that.

Sainan:
> There should be some contrast between the message and the source/code part since now it looks like part of the sentence.

but the source/code part does appear to be slightly darker? So I didn't introduce this issue.

Sainan:
> You should be able to adjust the syntax file to address that.

I don't know what this means. For context, before I patched this plugin, I had no experience with Python or the Sublime API, and I'm still just verifying my changes by opening the .sublime-package in 7zip, replacing the files, and restarting ST.

rchl:
> > You should be able to adjust the syntax file to address that.
> 
> I don't know what this means.

Actually it's a phantom so it's even easier to change. I'll make a comment in relevant place.

plugin/core/views.py
```diff
@@ -973,16 +969,22 @@ def format_diagnostic_for_html(
...
         code = None
-    source = diagnostic_source(diagnostic)
-    formatted.extend((" ", _with_color(source, "color(var(--foreground) alpha(0.6))")))
+    source = diagnostic.get("source")
+    if source is not None or code:
+        formatted.extend(" ")
+    if source is not None:
+        formatted.extend(_with_color(source, "color(var(--foreground) alpha(0.6))"))
     if code:
-        formatted.extend((_with_scope_color(view, ":", "punctuation.separator.lsp"), code))
```
rchl, line 985: `_with_scope_color` can be removed now.

plugin/core/views.py
```diff
@@ -973,16 +969,22 @@ def format_diagnostic_for_html(
...
+        formatted.extend(code)
```
rchl, line 987: Use `append` when not passing a list. `extend` creates new array item for each character.

plugin/core/views.py
```diff
@@ -973,16 +969,22 @@ def format_diagnostic_for_html(
...
+            code.append(_with_color(diagnostic["code"], "color(var(--foreground) alpha(0.6))"))
+        code.append(_with_color(")", "color(var(--foreground) alpha(0.6))"))
     else:
         code = None
-    source = diagnostic_source(diagnostic)
-    formatted.extend((" ", _with_color(source, "color(var(--foreground) alpha(0.6))")))
+    source = diagnostic.get("source")
+    if source is not None or code:
+        formatted.extend(" ")
+    if source is not None:
```
rchl, line 984:
```suggestion
    if source or code:
        formatted.extend(" ")
    if source:
```

Don't see the point of being so explicit here as for empty strings we would also not add anything.

plugin/core/views.py
```diff
@@ -871,22 +868,21 @@ def format_diagnostic_source_and_code(diagnostic: Diagnostic) -> str:
...
     if code is not None:
         code = str(code)
-        formatted.append(":")
         code_description = diagnostic.get("codeDescription")
         if code_description:
             href = code_description["href"]
         else:
-            formatted.append(code)
-    return "".join(formatted), code, href
+            formatted += "(" + code + ")"
```
rchl, line 884: We prefer formatting function:
```suggestion
            formatted += "({})".format(code)
```

rchl:
> Screenshot 2022-11-20 at 19 41 37

Ignoring the part in red, the annotations on the right of diagnostics don't look very readable with parenthesis. We can't use HTML in that area so we can't adjust contrast here. I wonder whether presenting in this place specifically should have space before `(`...

plugin/core/windows.py
```diff
@@ -485,7 +485,7 @@ def _update_panel_main_thread(self, characters: str, prephantoms: List[Tuple[int
         for row, col, code, href in prephantoms:
             point = panel.text_point(row, col)
             region = sublime.Region(point, point)
-            phantoms.append(sublime.Phantom(region, make_link(href, code), sublime.LAYOUT_INLINE))
+            phantoms.append(sublime.Phantom(region, make_link(href, "(" + code + ")"), sublime.LAYOUT_INLINE))
```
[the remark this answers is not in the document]
Sainan: That's what this would look like

image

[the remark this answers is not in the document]
rchl: Yes, looks good to me.

plugin/core/views.py
Sainan: `code` is a list now.

rchl: `code` is but two other places pass strings.

Sainan:
> Ignoring the part in red, the annotations on the right of diagnostics don't look very readable with parenthesis. We can't use HTML in that area so we can't adjust contrast here. I wonder whether presenting in this place specifically (or even in every place) should have space before `(`...

Before & after my changes
image
image

rchl:
> Ignoring the part in red, the annotations on the right of diagnostics don't look very readable with parenthesis. We can't use HTML in that area so we can't adjust contrast here. I wonder whether presenting in this place specifically (or even in every place) should have space before `(`...

You can ignore that part for now. We'll maybe tweak it later.

rchl (approved):

PR #2121, opened 2022-11-22
Add language id for Django templates

jnns:
I wondered why LSP wasn't working for my [Django](https://djangoproject.com) HTML templates. After I followed the steps outlined in https://github.com/sublimelsp/LSP-tailwindcss/issues/47 it worked. So here's a pull request so other people don't have to customize their configs to make it work.

### DIFF
--- a/language-ids.sublime-settings
+++ b/language-ids.sublime-settings
@@ -90,6 +90,7 @@
     "source.yaml.sublime.syntax": "yaml", // https://github.com/SublimeText/PackageDev
     "source.yaml.yarn": "yaml", // https://github.com/SublimeText/AFileIcon
     "text.advanced_csv": "csv", // https://github.com/SublimeText/AFileIcon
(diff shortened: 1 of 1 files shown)


language-ids.sublime-settings
```diff
@@ -97,6 +97,7 @@
     "text.html.markdown.rmarkdown": "r", // https://github.com/REditorSupport/sublime-ide-r
     "text.html.ngx": "html", // https://github.com/princemaple/ngx-html-syntax
     "text.jinja": "html", // https://github.com/Sublime-Instincts/BetterJinja
+    "text.django": "html", // https://github.com/willstott101/django-sublime-syntax
```
rchl, line 100: Please ensure the key is sorted.

jnns: Thanks, you're totally right.

Issue #2066, opened 2022-09-23
Suggestions for `Source Action...` menu handling


rchl:
From https://github.com/sublimelsp/LSP/pull/2064#issuecomment-1256289124:
>  -  remove the "Source Action..." entry from the context menu (because source actions are not bound to a particular point or selection, but apply to the entire file)
> -   Add "LSP: Source Actions..." to the "Edit" main menu, and either add another argument full_range: bool = False (or so) for the code actions command, which would be activated for the source actions menu entry, or alternatively derive this from whether only_kinds includes "source". So that the request for source actions would use the range of the entire file and also include all of the diagnostics in the file (this could be useful for the source.fixAll kind)

rchl:
@jwortmann 

Your two points make sense on the surface but what makes me a bit unsure is seeing how VSCode doesn't follow those.
It has `Source actions...` in the context menu and sends current selection range for the request. So I'm not sure we should be removing it from the context menu. If we did, then should we also remove "Format file" since it's a file-level action? And also, if we look outside of LSP, there are other actions in the view context menu that are file level rather than selection-level.

jwortmann:
> VSCode [...] sends current selection range for the request.

Maybe the `range` is ignored anyways by the servers when using `"only": ["source"]` in the request params. But from my point of view it would make more sense to use the whole content range and also include all of the diagnostics in the file. If in doubt, maybe we should ask what's expected for this request in https://github.com/microsoft/language-server-protocol/issues?

> If we did, then should we also remove "Format file" since it's a file-level action? And also, if we look outside of LSP, there are other actions in the view context menu that are file level rather than selection-level.

Personally I would be okay with removing "Format file" from the context menu. But probably it's not only about the file-level action, but should also depend on the expected usage frequence whether an entry should be included in the context menu. I guess most users run "Source Action..." only very rarely, while "Format file" is maybe used more often (and probably more often than "Format Selection", because it's simpler and not all servers are also documentRangeFormattingProvider).

---

I would probably also remove "Expand Selection" from the context menu, add a
```python
def is_visible(self, event: Optional[dict] = None, point: Optional[int] = None) -> bool:
        return self.is_enabled(event, point)
```
method to all of the commands in order to hide them when not activated, and then remove the "LSP" submenu.

predragnikolic:
> but from my point of view it would make more sense to use the **whole** content range and also include all of the diagnostics in the file

[...]

Issue #1983, opened 2022-07-03
Add back id in context menu or remove submenu


jwortmann:
LSP-* packages may want to extend the context menu with additional server-specific commands.

Example from VS Code:

![example1](https://user-images.githubusercontent.com/6579999/177052390-1e53761c-4409-4a62-b630-82f109fd6a91.png)

The "LSP" submenu was introduced in 8491948ca70acd1f691975c2d1a5d622a699d580 and the "id" for the LSP menu entries was removed in this commit. Extending a submenu from other packages is only possible if the submenu label has an "id". It would work, but then it will be at the bottom of the context menu, because all entries with "id" are appended to the bottom, after the native entries from the Default package (see https://docs.sublimetext.io/guide/customization/menus.html#menu-merging).

![example2](https://user-images.githubusercontent.com/6579999/177052418-53e6df5e-7f11-45db-95c8-b4e8b789f840.png)

(Specific entries from LSP-* helper packages could be hidden for not applicable files via the `is_visible` method of the command)

**Alternative**

We could also remove the "LSP" submenu again, and add back all menu items to the top-level context menu (without "id"). Then they will still be at the top, and LSP-* packages could just add their own items in a similar way. Due to ST's ordering by package name when merging resource files, they would immediately follow the LSP standard items then. To save space, we could also hide all disabled standard items via `is_visible()`, rather than just disable/grey out them via `is_enabled()`.

**Alternative 2**

LSP-* packages could also add their own submenu under the "LSP" submenu. But this is super ugly and also I think there is no way to hide this, even when all submenu items are hidden via `is_visible()`. So in this case you would end up with a non-functional submenu without any items even for unrelated files:

![example3](https://user-images.githubusercontent.com/6579999/177052428-6908b23f-390c-48b1-85c5-55be179e05f4.png)

---

Note that currently I don't have any concrete implementation, I'm just experimenting a bit. So this request isn't super urgent, but it would be useful to find a better solution for the future. Any opinions?

PR #2123, opened 2022-11-24
Combine file and range formatting entries in context menu

jwortmann:
This is a first step towards cleaning up the context menu (#1983) by combining "Format File" and "Format Selection" into a single item with dynamic label. It also makes the question obsolete whether "Format File" should be included in the context menu or if the main menu entry is enough (https://github.com/sublimelsp/LSP/issues/2066#issuecomment-1256614309).

### DIFF
--- a/Context.sublime-menu
+++ b/Context.sublime-menu
@@ -42,12 +42,7 @@
                 "caption": "Source Action…"
             },
             {
-                "command": "lsp_format_document",
-                "caption": "Format File"
-            },
-            {
-                "command": "lsp_format_document_range",
-                "caption": "Format Selection"
+                "command": "lsp_format"
             },
             {
                 "command": "lsp_expand_selection",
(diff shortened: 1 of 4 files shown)


rwols (approved):

rchl (approved):

rchl:
BTW. Should same treatment be applied to the main menu item (in `Edit` menu)?

jwortmann:
I didn't adjust the main menu, because in contrast to the context menu, space limitation is not a significant factor in the main menu. In theory, with the current main menu you could still choose to format the whole file, even if there is a selection in the view (though I'm not sure if anybody would actually want to do this).

Personally I don't have any objections to change it in the main menu too, and make it consistent with the context menu. Then the command palette entries should probably be adjusted as well?
https://github.com/sublimelsp/LSP/blob/658fcfecf465f1b54b702c2ddddc8ce09919756a/Default.sublime-commands#L56-L63

---

Btw I just saw that the command palette has "Goto Declaration" and "Goto Implementation", but not "Goto Definition" or "Find References".

rchl:
I don't have super strong opinion about it but I would lean towards consistency everywhere.

> Btw I just saw that the command palette has "Goto Declaration" and "Goto Implementation", but not "Goto Definition" or "Find References".

I've noticed it before and it seemed wrong to me.

PR #2124, opened 2022-11-25
Hide inactive items in context menu

jwortmann:
This is another step towards cleaning up the context menu, by hiding unavailable items there (but not in the main menu).

For the "Expand Selection" command I added a `fallback` argument (default to `True` to preserve current behavior) similar to the "Goto" commands.

I'd personally prefer to eventually remove the submenu and put the items at the top level, maybe with an "LSP:" prefix, so that they can be accessed more quickly. In that case it would probably be better to change the "Expand Selection" `fallback` default to `False`.

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -262,7 +262,7 @@ class LspCodeActionsCommand(LspTextCommand):
         point: Optional[int] = None,
         only_kinds: Optional[List[CodeActionKind]] = None
     ) -> bool:
-        if event is not None and 'x' in event:
+        if self.applies_to_context_menu(event):
             return self.is_enabled(event, point)
         return True
 
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -256,6 +256,16 @@ class LspCodeActionsCommand(LspTextCommand):
 
     capability = 'codeActionProvider'
 
+    def is_visible(
+        self,
+        event: Optional[dict] = None,
+        point: Optional[int] = None,
+        only_kinds: Optional[List[CodeActionKind]] = None
+    ) -> bool:
+        if event is not None and 'x' in event:
+            return self.is_enabled(event, point)
+        return True
+
     def run(
         self,
         edit: sublime.Edit,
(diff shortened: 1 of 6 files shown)


rwols (approved):

plugin/code_actions.py
```diff
@@ -256,6 +256,16 @@ class LspCodeActionsCommand(LspTextCommand):
 
     capability = 'codeActionProvider'
 
+    def is_visible(
+        self,
+        event: Optional[dict] = None,
+        point: Optional[int] = None,
+        only_kinds: Optional[List[CodeActionKind]] = None
+    ) -> bool:
+        if event is not None and 'x' in event:
```
rchl, line 265: The meaning of the "x in event" check is maybe not immediately clear. Perhaps it would be useful to create a function for it and make it clearer with appropriate name.

Effectively this seems to mean "is_in_context_menu_and_triggered_by_mouse"?
Although I wonder what happens on Windows, for example, when triggering context menu with a keyboard (if even possible). I would expect that we still get "x" in that case...

jwortmann: I can add a function. Like this?
```
def is_context_menu(event: Optional[dict]) -> bool:
    return event is not None and 'x' in event
```

> for example, when triggering context menu with a keyboard

Good point. I've just checked, and `x` is indeed contained in the `event` if triggered via keyboard (x and y apparently are the coordinates where the context menu opens, i.e. either at the mouse cursor, or at the caret position if triggered via keyboard). But actually this is exactly what we want (we want to distinguish whether the `is_visible` check is called for the context menu or for the main menu / command palette).

rchl: `is_triggered_from_context_menu` maybe? Otherwise sounds good.

Sublime API documentation also states:

> The X mouse position when a user clicks on a menu, or in a minihtml document.

But I'm feel like `is_visible` would probably not be relevant in the case of minihtml...

jwortmann: I'm struggling a bit to find a good name. "is_triggered_from_context_menu" sounds misleading to me, as if it were an action triggered by the user. I've named it "applies_to_context_menu" menu now, but if you have another better idea I'm open to change it.

> Sublime API documentation also states:

Right, in the minihtml case our function name is misleading, but it doesn't seem very relevant here indeed.

rchl: Yeah, it's a tricky one.

rchl (approved):

PR #2127, opened 2022-12-03
Fix prepareRename support

rchl:
The issue with the old code was that it has shown the input overlay asking for new name before sending the `prepareRename` request and didn't use the range (or placeholder) from the response.

Added big comment about how `lsp_symbol_rename` works since it behaves very differently depending on if server supports prepareRename request.

Related to https://github.com/sublimelsp/LSP-typescript/issues/178

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -837,7 +837,6 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
     def purge_changes_async(self) -> None:
         for sv in self.session_views_async():
             sv.purge_changes_async()
-            break
 
     def trigger_on_pre_save_async(self) -> None:
         for sv in self.session_views_async():
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -837,6 +837,7 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
     def purge_changes_async(self) -> None:
         for sv in self.session_views_async():
             sv.purge_changes_async()
+            break
 
     def trigger_on_pre_save_async(self) -> None:
         for sv in self.session_views_async():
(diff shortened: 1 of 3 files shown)


plugin/documents.py
```diff
@@ -837,6 +837,7 @@ def has_capability_async(self, session: Session, capability_path: str) -> bool:
     def purge_changes_async(self) -> None:
         for sv in self.session_views_async():
             sv.purge_changes_async()
+            break
```
rchl, line 840: Added break since technically we only need to purge once per SessionBuffer, regardless how many SessionViews there are.
That said, it was correctly before too because subsequent purges would do nothing.

plugin/rename.py
```diff
@@ -1,61 +1,58 @@
...
-        new_name: str = "",
-        placeholder: str = "",
-        position: Optional[int] = None,
-        event: Optional[dict] = None,
-        point: Optional[int] = None
-    ) -> bool:
-        if self.best_session("renameProvider.prepareProvider"):
-            # The language server will tell us if the selection is on a valid token.
-            return True
-        return super().is_enabled(event, point)
```
rchl, line 57: This was redundant and it's enough to use base implementation.
When this code was introduced it looked a bit differently and it made sense then.

plugin/rename.py
```diff
@@ -93,29 +94,26 @@ def run(
         event: Optional[dict] = None,
         point: Optional[int] = None
     ) -> None:
-        if position is None:
-            tmp_pos = get_position(self.view, event, point)
-            if tmp_pos is None:
+        listener = self.get_listener()
+        if listener:
+            listener.purge_changes_async()
```
rchl, line 101: I've added "purge" because I've noticed an issue with lsp-typescript where it requests client to apply edits and immediately triggers rename before client has triggered "didChange".

https://github.com/typescript-language-server/typescript-language-server/blob/0403b36a6ae181fb25f8236f47029afeb2a739fb/src/lsp-server.ts#L938-L949

plugin/documents.py
rchl: my bad, that was wrong to change as this needs to happen for all sessions in the file

PR #2132, opened 2022-12-08
Fix plugin overwrite `on_workspace_configuration`

LDAP:
Currently, plugins can not set the configuration object in `on_workspace_configuration` if it is `None` or it has to be replaced completely.

This did only work in those cases where `configuration` was a reference to an object that could be altered.

See 
https://github.com/sublimelsp/LSP-jdtls/commit/85f2392c9af3cdbf5048bdac8fad81cb29a1c95a
for an example where the current implementation does not work.

This PR changes `on_workspace_configuration` so that it must return the new configuration item. The default implementation now forwards the "pre-resolved" configuration.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -866,14 +866,16 @@ class AbstractPlugin(metaclass=ABCMeta):
         """
         pass
 
-    def on_workspace_configuration(self, params: Dict, configuration: Any) -> None:
+    def on_workspace_configuration(self, params: Dict, configuration: Any) -> Any:
         """
         Override to augment configuration returned for the workspace/configuration request.
 
         :param      params:         A ConfigurationItem for which configuration is requested.
-        :param      configuration:  The resolved configuration for given params.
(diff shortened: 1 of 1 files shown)


rchl:
It's a breaking change for LSP-eslint: https://github.com/sublimelsp/LSP-eslint/blob/fd903029a075497d6c8eeecbe980a6c4748e6ac7/plugin.py#L70-L70

Will you also create a PR that adapts it to use the changed API?

(Normally I would prefer a non-breaking change or new API but I think that we can pretty confidently say that nothing else is using this API so it should be fine.)

LDAP:
Yes, I'll create a PR there too.

LDAP:
I opened a PR as far as I can see we can publish the LSP-eslint change without breaking anything. The LSP must be released before LSP-jdtls.

(I tried to think of a non-breaking solution but could not come up with something better)

rchl (approved):

rchl:
Released new LSP version with this fix.

PR #2134, opened 2022-12-11
Pass force_group to LocationPicker

rchl:
Addressing issue mentioned in https://github.com/sublimelsp/LSP/pull/1994#issuecomment-1270279134

### DIFF
--- a/plugin/goto.py
+++ b/plugin/goto.py
@@ -89,7 +89,8 @@ class LspGotoCommand(LspTextCommand):
                 placeholder = self.placeholder_text + " " + self.view.substr(self.view.word(position))
                 kind = get_symbol_kind_from_scope(self.view.scope_name(position))
                 sublime.set_timeout(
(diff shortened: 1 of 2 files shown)


jwortmann:
Judging from the diff here, the argument is still not passed from the "Goto ..." commands. So I guess at the current state it wouldn't solve the linked issue, and still just use the default value `True` ...

rchl:
I admit I haven't bothered to test the original issue but I've assumed that this is fixing the case when somewhere has defined a custom command that passes "force_group: false". Is that not the case?

rchl:
I see now. Forgot to change the place that you've actually indicated is the issue.

That said, still haven't tested as pyright doesn't trigger this code path and I didn't bother to find out how to trigger it.

rchl:
Looks better now @jwortmann ?

jwortmann:
Yes, should be fine to merge now.

It was obvious to me that you had just forgotten to adjust the other place in the code. Maybe I should have expressed that more clearly ;-)

PR #2139, opened 2022-12-13
fix code lenses not updating after Undo

### LINKED ISSUE #2136: Code Lenses fail to update their position (and content) after Undo
**Describe the bug**

Code lenses are not getting re-requested and updated in some circumstances like for for example being scrolled out of the view and Undo being performed that restores removed line above the code lenses.

**To Reproduce**

Notice how code lenses are mispositioned after reverting removed line:

https://user-images.githubusercontent.com/153197/207286020-e3ad46e7-0442-4432-b522-063bd3a8c341.mov

- Sublime Text version: 4143
- LSP version: 1.20.0
- Language servers used: LSP-volar

**Additional information**

Checking the LSP logs I can see that on doing Undo we are sending `didChange` notification but not making the `textDocument/codeLens` request to update code lenses.

rchl:
Updating of code lenses should IMO not be tied to selection update as document can get modified programmatically in which case selection might not change.

There is also another weird case where I've noticed this issue during "Undo" and that one has deeper roots... When undoing removal of a line, `on_text_changed_async` is quickly called twice and on the first call selection is `different` from the stored one but on the second call, it's not. That results in `_when_selection_remains_stable_async` not being triggered (first time because region no longer matches in `debounced` and second time because `different` is `False` so we return early from `on_text_changed_async`). This is an issue that should still be fixed but it's fairly minor and probably doesn't even effect `_do_highlights_async` much.

That said, switching to `Debouncer` class avoids the above issue affecting code lenses.

(I guess we might be a bit concerned about using locks in more places but I think and hope that it's not really that expensive. Especially as `Debouncer` is only holding the lock to update a single property.)

Fixes #2136

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -6,7 +6,6 @@ from .typing import Any, Optional, List, Dict, Generator, Callable, Iterable, Un
 from .typing import cast
 from .url import filename_to_uri
 from .url import parse_uri
-from threading import RLock
 from wcmatch.glob import BRACE
 from wcmatch.glob import globmatch
 from wcmatch.glob import GLOBSTAR
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -138,7 +138,7 @@ class DebouncerNonThreadSafe:
     Debouncer for delaying execution of a function until specified timeout time.
 
     When calling `debouce()` multiple times, with time span between calls being shorter than the specified `timeout_ms`,
-    the specified callback function will only be called once, after a period of `timeout_ms` since last call.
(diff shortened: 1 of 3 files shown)


rchl:
Note on `Debouncer` and locks. I guess we could create a non-thread-safe variant of it that would not use locks but then it would be our responsibility to ensure that `.debounce()` is only called on the thread as the function it runs.

rchl:
Changed Debouncer to be not thread-safe as we don't currently seem to have a need for thread safe version.

rchl:
No complains I guess. :)

PR #2141, opened 2022-12-15
Add "Refactor" entry to the "Edit" main menu

### LINKED ISSUE #1846: Split up Code Actions into "Refactors" and "Source Actions"

jwortmann:
Here's another experiment of mine. It can dynamically show up to 5 "Refactor" code actions in the "Edit" main menu, without the need to open the quick panel overlay.

I haven't tested with actions from multiple sessions at the same time, but I think it should work™.

Closes #1846 (?)

refactor

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -34,6 +34,7 @@ class CodeActionsManager:
 
     def __init__(self) -> None:
         self._response_cache = None  # type: Optional[Tuple[str, Promise[List[CodeActionsByConfigName]]]]
+        self.refactor_actions_cache = []  # type: List[Tuple[str, CodeAction]]
 
     def request_for_region_async(
         self,
@@ -49,6 +50,7 @@ class CodeActionsManager:
         """
         listener = windows.listener_for_view(view)
         if not listener:
+            self.refactor_actions_cache.clear()
             return Promise.resolve([])
         location_cache_key = None
         use_cache = not manual
@@ -71,6 +73,9 @@ class CodeActionsManager:
             return Request.codeAction(params, view)
 
         def response_filter(session: Session, actions: List[CodeActionOrCommand]) -> List[CodeActionOrCommand]:
+            self.refactor_actions_cache.extend(
+                [(session.config.name, cast(CodeAction, a)) for a in actions
+                    if not is_command(a) and kinds_include_kind([CodeActionKind.Refactor], a.get('kind'))])
             # Filter out non "quickfix" code actions unless "only_kinds" is provided.
             if only_kinds:
                 return [a for a in actions if not is_command(a) and kinds_include_kind(only_kinds, a.get('kind'))]
@@ -82,6 +87,7 @@ class CodeActionsManager:
                 if is_command(a) or not a.get('kind') or kinds_include_kind([CodeActionKind.QuickFix], a.get('kind'))
             ]
 
+        self.refactor_actions_cache.clear()
         task = self._collect_code_actions_async(listener, request_factory, response_filter)
         if location_cache_key:
             self._response_cache = (location_cache_key, task)
@@ -329,6 +335,46 @@ class LspCodeActionsCommand(LspTextCommand):
                     .then(lambda response: self._handle_response_async(config_name, response))
 
         sublime.set_timeout_async(run_async)
+
+    def _handle_response_async(self, session_name: str, response: Any) -> None:
+        if isinstance(response, Error):
+            sublime.error_message("{}: {}".format(session_name, str(response)))
+
+
+class LspRefactorCommand(LspTextCommand):
+
+    capability = 'codeActionProvider'
+
+    def is_visible(self, id: int, event: Optional[dict] = None, point: Optional[int] = None) -> bool:
+        return len(actions_manager.refactor_actions_cache) > id
+
+    def description(self, id: int, event: Optional[dict] = None, point: Optional[int] = None) -> Optional[str]:
+        if self.is_visible(id):
+            return actions_manager.refactor_actions_cache[id][1]['title']
+
+    def run(self, edit: sublime.Edit, id: int, event: Optional[dict] = None, point: Optional[int] = None) -> None:
+        sublime.set_timeout_async(partial(self.run_async, id))
+
+    def _is_cache_valid(self) -> bool:
+        v = self.view
+        if actions_manager._response_cache is None:
+            return False
+        region = first_selection_region(v)
+        if region is None:
+            return False
+        listener = windows.listener_for_view(v)
+        if not listener:
+            return False
+        _, covering = listener.diagnostics_intersecting_async(region)
+        return actions_manager._response_cache[0] == "{}#{}:{}".format(v.buffer_id(), v.change_count(), covering)
+
+    def run_async(self, id: int) -> None:
+        if self._is_cache_valid():
+            config_name, action = actions_manager.refactor_actions_cache[id]
+            session = self.session_by_name(config_name)
+            if session:
+                session.run_code_action_async(action, progress=True) \
+                    .then(lambda response: self._handle_response_async(config_name, response))
 
     def _handle_response_async(self, session_name: str, response: Any) -> None:
         if isinstance(response, Error):
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -80,9 +80,13 @@ class CodeActionsManager:
             # Filter out non "quickfix" code actions unless "only_kinds" is provided.
             if only_kinds:
                 if manual and only_kinds == [CodeActionKind.Refactor]:
+                    refactor_actions = [
+                        cast(CodeAction, a) for a in actions
+                        if not is_command(a) and kinds_include_kind([CodeActionKind.Refactor], a.get('kind'))
+                    ]
                     self.refactor_actions_cache.extend(
-                        [(session.config.name, cast(CodeAction, a)) for a in actions
-                            if not is_command(a) and kinds_include_kind([CodeActionKind.Refactor], a.get('kind'))])
+                        [(session.config.name, a) for a in refactor_actions] if len(refactor_actions) <= 10 else
+                        [(session.config.name, a) for a in refactor_actions if not a.get('disabled')])
                 return [
                     a for a in actions
                     if not is_command(a) and kinds_include_kind(only_kinds, a.get('kind')) and not a.get('disabled')
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -364,9 +364,7 @@ class LspRefactorCommand(LspTextCommand):
         return len(actions_manager.refactor_actions_cache) > id and self._is_cache_valid()
 
     def description(self, id: int, event: Optional[dict] = None, point: Optional[int] = None) -> Optional[str]:
-        if id == -1:
-            return None
-        if len(actions_manager.refactor_actions_cache) > id:
+        if id > -1 and len(actions_manager.refactor_actions_cache) > id:
             return actions_manager.refactor_actions_cache[id][1]['title']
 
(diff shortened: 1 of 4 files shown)


rchl:
It seems like in the case when there are no refactors we can end up with a menu item that doesn't open a submenu. Are we able to handle that so that the sub menu doesn't show up or eventually shows submenu with one disabled item?

Screenshot 2022-12-16 at 09 34 36

jwortmann:
> Are we able to handle that so that the menu item doesn't show up

I tried to add a command to the "LSP: Refactor" item, but it seems that the `is_enabled` and `is_visible` methods are not called if the item has `"children"`. So I think it would have to be implemented in ST core to change the appearance depending on whether submenu items are visible. But note that the `is_visible` call for the submenu items is delayed until you hover with the mouse cursor over "LSP: Refactor". This strategy would have to be changed in order to implement it in ST.

Ideally "LSP: Refactor" should be visible, but disabled (grayed out), like e.g. the menu in Firefox works:

menu

> or eventually shows submenu with one disabled item?

That would be possible, but would it really be better? What would you use for the label?

jwortmann:
I moved the "Rename..." item into the submenu. That will guarantee that it is never empty, because even if there is no renameProvider available, this item will always be present (grayed out in that case).

rchl:
I have an impression I saw a bug where the items were shown but were disabled. I can't reproduce it again though... Will run with those changes for a while.

But do you have any intentions to do the same with the context menu? Feels weird to have different behavior depending on from which menu the command is triggered.

plugin/code_actions.py
```diff
@@ -71,6 +73,9 @@ def request_factory(session: Session) -> Optional[Request]:
             return Request.codeAction(params, view)
 
         def response_filter(session: Session, actions: List[CodeActionOrCommand]) -> List[CodeActionOrCommand]:
+            self.refactor_actions_cache.extend(
+                [(session.config.name, cast(CodeAction, a)) for a in actions
+                    if not is_command(a) and kinds_include_kind([CodeActionKind.Refactor], a.get('kind'))])
```
rchl, line 78: It's not fully correct to cache and get refactors from the standard region request because standard request doesn't include "only: [refactor]" and servers can and do return more refactors when "refactor" kind is explicitly specified. `typescript-language-server` does that for example.

Triggering "refactor..." from the context menu includes it but request that happens on moving the cursor doesn't.

plugin/code_actions.py
```diff
@@ -333,3 +339,43 @@ def run_async() -> None:
...
+
+    def run_async(self, id: int) -> None:
+        if self._is_cache_valid():
+            config_name, action = actions_manager.refactor_actions_cache[id]
+            session = self.session_by_name(config_name)
+            if session:
+                session.run_code_action_async(action, progress=True) \
+                    .then(lambda response: self._handle_response_async(config_name, response))
+
+    def _handle_response_async(self, session_name: str, response: Any) -> None:
```
rchl, line 379: A method order like:
- run
- run_async
- _handle_response_async
- _is_cache_valid

would be better (or with two last reversed)  since that would better match the calling order.

plugin/code_actions.py
jwortmann: Hm, that's unfortunate. I guess we don't want to send another request with `"only": ["refactor"]` everytime when the cursor/selection changes. So if other servers also don't send the "refactor" actions in the response to the regular request after selection change, then this PR is probably useless for those servers. On the other hand, the implementation here is a strict addition to the current behavior, because currently there isn't any such item in the "Edit" main menu at all.

It's probably worth to check more servers, if they include the "refactor" actions or not.

jwortmann:
> I have an impression I saw a bug where the items were shown but were disabled. I can't reproduce it again though...

I saw that too one time. But from looking at the code, I don't see how this could happen.

> But do you have any intentions to do the same with the context menu?

I'm not sure. From the UX perspective I prefer the submenu from this PR with the cached items. Because you immediately see what's available and can run the refactor just with a single click. The alternative with the quick panel requires another mouse click or <kbd>Enter</kbd> press, and it can happen that the response will return no refactor actions at all for the location.

But if it doesn't work with most servers (see other comment), then of course it would be bad to change the current entry in the context menu. Therefore I think it would be better to keep it for now.

I think the filtering in the quick panel is not very important, because probably there are seldom more than 2 or 3 different "refactor" actions. And the session name in the quick panel is maybe a small nice-to-have addition, but also without real impact for the user, and I guess most of the time there is only a single server running anyway.

Another idea for the submenu would be to also show the "disabled" actions, but grayed out (via `is_enabled -> False`). It seems like the CodeActionsManager would need a bit of refactoring for that, because right now the disabled actions are filtered out immediately. Then we can reintroduce `"disabledSupport": True` ;-)

rchl:
I maybe wouldn't mind switching completely from quick panel to menus for refactors but then we should have it handled that way in all places (main menu, context menu, and command palette) and (more importantly) it should use the results from the request that includes `"only": ["refactor"]`. And that second reason sounds hard to achieve because ideally the request would be made on opening the menu and ST API is not async so it's not really possible. Due to that, I'm not sure we can really follow that path.

But maybe others have different opinions and should override my judgment here...

plugin/code_actions.py
rchl: > It's probably worth to check more servers, if they include the "refactor" actions or not.

To be fair, I'm not sure if I came up with this logic for `typescript-language-server` or if that's something that I've read about or some elsewhere... Maybe it's not exactly expected to return different refactor results based on `only`...

rchl:
Also, if we would switch to menus for refactors then maybe one could make argument that we should switch to menus for all code actions...

If only ST menus would have async API and richer functionality. VSCode has this:

Screenshot 2022-12-17 at 22 44 02

jwortmann:
Could you try again with the TypeScript server and the latest changes? Now it should send a request with `"only_kinds": ["refactor"]` when you open the "Edit" main menu.

rchl:
There are 4 enabled refactors but the menu shows only 1. I guess because it doesn't prioritize enabled ones.

Screenshot 2022-12-18 at 21 15 12

jwortmann:
So are there more than 5 refactor actions in total? I guess we can just increase the number of possible menu items then. I just arbitrarily chose 5, because I thought that servers won't provide more than 2 or 3 refactors.

Yes, enabled actions are not prioritized and they are listed in the order as defined in the server response.

rchl:
Yes, there is more than 5.

This is what VSCode shows for the same place in the code. Note that it doesn't even show disabled refactors by default and I wonder whether it's really worth adding support for those. It's probably not that useful to show disabled. Especially since we are not showing the "reason" for disabling anyway.

Screenshot 2022-12-18 at 21 22 18

rchl:
There is actually 15 refactors returned so we are still missing one of the enabled ones :)

Screenshot 2022-12-18 at 21 26 07

jwortmann:
> It's probably not that useful to show disabled. Especially since we are not showing the "reason" for disabling anyway.

I will think about showing actions which are disabled, maybe it's indeed better to never show them.
Does VSCode show the "reason" for disabled actions on mouse hover (or in another way)?

rchl:
No, they don't show the reason either.

jwortmann:
I'm still undecided about the disabled actions, but I've removed them from the menu for now.

There are still the following questions:
- add "Source Action" to the "Edit" main menu with a similar submenu? (#2066)
- should the context menu be adjusted to match this submenu instead of opening the command palette?

But I think those are not blocking for this PR and can be discussed / implemented separately.
For the "Code Action..." item in the context menu I would keep using the command palette though, because this allows to show the icon for the action kind, which would not be possible if it were a submenu.

rchl:
I would think that handling it the same way for the "refactor" in the context menu would be in scope of this PR but I'm fine if you want to handle it separately.

plugin/core/sessions.py
```diff
@@ -336,7 +336,7 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
                 }
             },
             "dataSupport": True,
-            "isPreferredSupport": True,
+            "disabledSupport": True,
```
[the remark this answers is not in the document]
jwortmann: Sorry, yes thanks

rchl (approved):

rchl:
Noticed that results from `Refactor` in the context menu can differ from results from the main menu because the former expands the range to include all overlapping diagnostics but the latter doesn't. The context menu behavior also has a bug that it will consider diagnostics for that purpose even if those are from different session.

I think that extending the range and including diagnostics in the range is the right thing to do but you can challenge me on that :)

jwortmann:
> I think that extending the range and including diagnostics in the request is the right thing to do but you can challenge me on that :)

I considered about whether or not to extend the range to diagnostics as well, but I dropped them on purpose because I would say they are not really connected to refactors, but only to "quickfixes".

Maybe you can wait 1 or 2 days more before you make a new release. I think it could be better and less confusing for users to finish the main and context menu cleanup first.

I could make another PR soon, but feedback would be very welcome, because these UI things are very subjective.
What I still have in mind would be submenus in the Edit main menu with
```
| ------------------ | ----------------------------------|
| LSP: Source Action | <dynamic server specific entries> |
| ------------------ | ----------------------------------|
| LSP: Format        | Format File                       |
|                    | Format Selection                  |
| ------------------ | ----------------------------------|
```
and for the context menu perhaps something like (left side is at the top level, i.e. no more single "LSP" submenu)
```
| --------------------- | ----------------------------------|
| LSP: Goto...          | Goto Definition...                |
|                       | Goto Declaration...               |
|                       | Goto Declaration...               |
|                       | Goto Type Definition...           |
|                       | Goto Implementation...            |
|                       | Find References                   |
| --------------------- | ----------------------------------|
| LSP: Code Action...   |
| --------------------- | ----------------------------------|
| LSP: Refactor         | Rename...                         |
|                       | <dynamic server specific entries> |
| --------------------- | ----------------------------------|
| LSP: Source Action    | <dynamic server specific entries> |
| --------------------- | ----------------------------------|
| LSP: <Format File / Format Selection> |
| --------------------- | ----------------------------------|
| LSP: Expand Selection |
| --------------------- | ----------------------------------|
```
(or maybe drop "Source Action" and "Expand Selection" from the context menu?)

rchl:
I don't have to release right away but it would be good to finally get some good fixes out there.

The follow up changes that you are describing here would be nice but I would think we should have a longer period of testing before releasing those so if we wait 1, 2 days for those, we might as well have to wait a week or more to properly test those.

So the alternative is not wait and release without this change to avoid inconsistent behavior.

As far as your change suggestons:
 - I wouldn't drop "Source action", I think. It can has useful items in there and not having it would make discoverability pretty hard.
 - Expand selection, probably OK to drop. Using this functionality through a context menu sounds like a pain anyway since you usually have to press it multiple times and going back to the context menu each time is not fun. It's really meant to be bound to a key.
 - How would handle case when there are no items to show for "LSP: Source Action" or "LSP: Goto..." since we can't hide parent items?

jwortmann:
I made another PR to add "Source Action" to the main menu. I think this is something useful and consistent with the "Refactor" menu.  At least this is where I would look for things like "organize imports" (the servers which I personally use don't provide any source actions though).

> * Expand selection, probably OK to drop. Using this functionality through a context menu sounds like a pain anyway since you usually have to press it multiple times and going back to the context menu each time is not fun. It's really meant to be bound to a key.

I agree.

> * How would we handle the case when there are no items to show for "LSP: Source Action" or "LSP: Goto..." since we can't hide parent items?

I came to the same conclusion, and this is indeed very unsatisfying. Therefore I didn't touch the context menu for now.

PR #2145, opened 2022-12-20
Auto-restart on server crashing, up to 5 times

LDAP:
This PR aims to do two things:
- [x] BREAKING: Provide the exit code to the LSP-* helper plugins. That allows for special actions on certain exit codes. (E.g. jdtls is known to crash with code -13 if the workspace cache is corrupted.)
- [x] Auto restart servers a few times before asking the user what to do

Open questions:
- Where should the state about crashed servers be stored? (WindowManager, ClientConfig, ...)
- How often do we attempt to restart the server? When is the counter reset?
  (e.g. Increase the counter if the last crash was <30s apart, else reset the counter. If the counter reaches 5, ask the user and reset the counter?)

### DIFF
--- a/plugin/core/configurations.py
+++ b/plugin/core/configurations.py
@@ -85,8 +85,6 @@ class WindowConfigManager(object):
         Signal that a session has crashed.
 
         Returns True if the session should be restarted automatically.
-        Returns False if the session should not be restarted automatically,
-        in this case the counter is reset.
         """
         now = datetime.now()
         if config_name not in self._crashes:
--- a/plugin/core/configurations.py
+++ b/plugin/core/configurations.py
@@ -93,19 +93,8 @@ class WindowConfigManager(object):
         self._crashes[config_name].append(now)
         timeout = now - RETRY_COUNT_TIMEDELTA
         crash_count = len([crash for crash in self._crashes[config_name] if crash > timeout])
-        msg = "".join((
-            "session for config {} crashed ",
-            "({} / {} times in the last {} seconds), ",
-            "exit code {}, exception: {}",
-        )).format(
-            config_name,
-            crash_count,
-            RETRY_MAX_COUNT,
-            RETRY_COUNT_TIMEDELTA.total_seconds(),
-            exit_code,
-            exception
-        )
-        printf(msg)
+        printf("{} crashed ({} / {} times in the last {} seconds), exit code {}, exception: {}".format(
+            config_name, crash_count, RETRY_MAX_COUNT, RETRY_COUNT_TIMEDELTA.total_seconds(), exit_code, exception))
         return crash_count < RETRY_MAX_COUNT
 
     def _disable_for_session(self, config_name: str) -> None:
--- a/plugin/core/configurations.py
+++ b/plugin/core/configurations.py
@@ -3,10 +3,10 @@ from .logging import exception_log
 from .types import ClientConfig
 from .typing import Generator, List, Optional, Set, Dict, Deque
 from .workspace import enable_in_project, disable_in_project
+from collections import deque
+from datetime import datetime, timedelta
 import sublime
 import urllib.parse
-from datetime import datetime, timedelta
-from collections import deque
 
 
 RETRY_MAX_COUNT = 5
--- a/plugin/core/configurations.py
+++ b/plugin/core/configurations.py
@@ -100,11 +100,7 @@ class WindowConfigManager(object):
             "({} / {} times in the last {} seconds)"
         )).format(config_name, crash_count, RETRY_MAX_COUNT, RETRY_COUNT_TIMEDELTA.total_seconds())
         debug(msg)
-        if crash_count >= RETRY_MAX_COUNT:
-            # CRASH_MAX_COUNT crashes in _CRASH_COUNT_TIMEDELTA
-            del self._crashes[config_name]
-            return False
-        return True
+        return crash_count < RETRY_MAX_COUNT
 
     def _disable_for_session(self, config_name: str) -> None:
         self._disabled_for_session.add(config_name)
--- a/plugin/core/configurations.py
+++ b/plugin/core/configurations.py
@@ -1,10 +1,16 @@
 from .logging import debug
 from .logging import exception_log
 from .types import ClientConfig
-from .typing import Generator, List, Optional, Set, Dict
+from .typing import Generator, List, Optional, Set, Dict, Deque
 from .workspace import enable_in_project, disable_in_project
 import sublime
 import urllib.parse
+from datetime import datetime, timedelta
+from collections import deque
+
+
+RETRY_MAX_COUNT = 5
+RETRY_COUNT_TIMEDELTA = timedelta(minutes=3)
 
 
 class WindowConfigManager(object):
@@ -12,6 +18,7 @@ class WindowConfigManager(object):
         self._window = window
         self._global_configs = global_configs
         self._disabled_for_session = set()  # type: Set[str]
+        self._crashes = {}  # type: Dict[str, Deque[datetime]]
         self.all = {}  # type: Dict[str, ClientConfig]
         self.update()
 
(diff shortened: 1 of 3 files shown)


rchl:
> * Where should the state about crashed servers be stored? (WindowManager, ClientConfig, ...)

We remember state when the plugin crashes in WindowsConfigManager so that's the right place to use.

> How often do we attempt to restart the server? When is the counter reset?

I believe VSCode triggers it on 5th crash in the last 3 minutes.

LDAP:
I could not find any plugin that uses the `on_session_end_async` API, which was changed in this PR.

rchl:
> I could not find any plugin that uses the `on_session_end_async` API, which was changed in this PR.

Yes, me neither.

plugin/core/sessions.py
```diff
@@ -943,12 +943,16 @@ def on_session_buffer_changed_async(self, session_buffer: SessionBufferProtocol)
         """
         pass
 
-    def on_session_end_async(self) -> None:
+    def on_session_end_async(self, exit_code: Optional[int], exception: Optional[Exception]) -> None:
```
rchl, line 946: Why optional `exit_code`? What does `None` symbolize?

rchl: I've read the docstring now but I don't think I agree with that. When session hasn't crashed then the exit code of 0 should be perfectly valid.

LDAP: The API is called from two places. Here: https://github.com/sublimelsp/LSP/blob/fd1a1ce115ddcebb13c5fcd9961d870ac3dc02b4/plugin/core/sessions.py#L1856 the server is not yet stopped.

It is explained in the docstring:
> If the session hasn't crashed, a shutdown message will be send immediately
        after this method returns. In this case exit_code and exception are None.
        If the session has crashed, the exit_code or the exception is not None.

plugin/core/configurations.py
```diff
@@ -73,6 +80,26 @@ def disable_config(self, config_name: str, only_for_session: bool = False) -> No
...
+            self._crashes[config_name] = deque([now], maxlen=RETRY_MAX_COUNT)
+        else:
+            self._crashes[config_name].append(now)
+        timeout = now - RETRY_COUNT_TIMEDELTA
+        crash_count = len([crash for crash in self._crashes[config_name] if crash > timeout])
+        msg = "".join((
+            "session for config {} crashed. ",
+            "({} / {} times in the last {} seconds)"
+        )).format(config_name, crash_count, RETRY_MAX_COUNT, RETRY_COUNT_TIMEDELTA.total_seconds())
+        debug(msg)
```
rchl, line 100: I would want exception included in the debug log.

And also a question of whether it should be `debug` or a normal log... Most people don't run with debug logging and having server crash without any feedback (even if only in the console) can result in people not understanding what happened and would likely also make support harder. We won't be able to just ask: "do you see any crash messages in the console" if someone asks why server resets sometimes?

plugin/core/sessions.py
rchl: But I would still report `0`. It's an explicit shutdown of the server not caused by a crash so IMO it's OK to specify 0. Is there more specific reason why you think `None` is better?

plugin/core/configurations.py
LDAP: Agree! Pushed an update

plugin/core/sessions.py
LDAP: `None` would allow detecting if the server quit on his own or crashed with exit-code 0 for some reason (`exit_code == 0`) or if it was shutdown by the LSP plugin (`exit_code == None`).

LDAP: I just realized that there is a better reason to differentiate: In case of None the session is still usable and requests to the server can be made. In case of not None this isn't the case.

plugin/core/configurations.py
```diff
@@ -73,6 +81,34 @@ def disable_config(self, config_name: str, only_for_session: bool = False) -> No
...
+            "exit code {}, exception: {}",
+        )).format(
+            config_name,
+            crash_count,
+            RETRY_MAX_COUNT,
+            RETRY_COUNT_TIMEDELTA.total_seconds(),
+            exit_code,
+            exception
+        )
+        printf(msg)
```
rchl, line 109: So many lines... Can we just do this?
Note that the message is slightly altered. I think that "session for config" is a bit too low-level. Just printing the name of the config should be enough.

```suggestion
        printf("{} crashed ({} / {} times in the last {} seconds), exit code {}, exception: {}".format(
            config_name, crash_count, RETRY_MAX_COUNT, RETRY_COUNT_TIMEDELTA.total_seconds(), exit_code, exception))
```

plugin/core/sessions.py
```diff
@@ -943,12 +943,16 @@ def on_session_buffer_changed_async(self, session_buffer: SessionBufferProtocol)
...
         """
         Notifies about the session ending (also if the session has crashed). Provides an opportunity to clean up
         any stored state or delete references to the session or plugin instance that would otherwise prevent the
-        instance from being garbage-collected. If the plugin hasn't crashed, a shutdown message will be send immediately
-        after this method returns.
+        instance from being garbage-collected.
+
+        If the session hasn't crashed, a shutdown message will be send immediately
+        after this method returns. In this case exit_code and exception are None.
+        If the session has crashed, the exit_code or the exception is not None.
```
rchl, line 954: I suppose it could be more precise. In case of a crash `exit_code` will always be set and `exception` is optional.

```suggestion
        If the session has crashed, the exit_code and an optional exception are provided.
```

plugin/core/configurations.py
LDAP: Sure.

plugin/core/windows.py
```diff
@@ -393,15 +393,18 @@ def on_post_exit_async(self, session: Session, exit_code: int, exception: Option
...
+            if not restart:
+                msg = "".join((
+                    "{0} exited with status code {1}. ",
+                    "Do you want to restart it? If you choose Cancel, it will be disabled for this window for the ",
+                    "duration of the current session. ",
+                    "Re-enable by running \"LSP: Enable Language Server In Project\" from the Command Palette."
+                )).format(config.name, exit_code)
+                if exception:
+                    msg += "\n\n--- Error: ---\n{}".format(str(exception))
+                restart = sublime.ok_cancel_dialog(msg, "Restart {}".format(config.name))
```
rchl, line 406: After trying it out I've realized that I would want this text to mention that it crash 5 times during last 3 minutes and also make it a bit more user-friendly. My suggestion would be something like this but we can discuss further:

```suggestion
                msg = "".join((
                    "The {0} server crashed 5 times in the last 3 minutes.\n\n",
                    "You can try to Restart it or you can choose Cancel to disable it for this window for the ",
                    "duration of the current session. ",
                    "Re-enable by running \"LSP: Enable Language Server In Project\" from the Command Palette."
                )).format(config.name)
                if exception:
                    msg += "\n\n--- Error: ---\n{}".format(str(exception))
                restart = sublime.ok_cancel_dialog(msg, "Restart")
```

(Ideally the 5 and 3 variables would be imported of course)

LDAP: Agree!

plugin/core/windows.py
```diff
@@ -393,15 +393,19 @@ def on_post_exit_async(self, session: Session, exit_code: int, exception: Option
...
-                "Re-enable by running \"LSP: Enable Language Server In Project\" from the Command Palette."
-            )).format(config.name, exit_code)
-            if exception:
-                msg += "\n\n--- Error: ---\n{}".format(str(exception))
-            if sublime.ok_cancel_dialog(msg, "Restart {}".format(config.name)):
+            restart = self._config_manager.record_crash(config.name, exit_code, exception)
+            if not restart:
+                msg = "".join((
+                    "The {0} server has crashed {1} times in the last {2} seconds.\n",
+                    "Exit code: {3}\n\n",
```
rchl, line 400: I've removed this on purpose because:
 - it's quite low-level
 - it will be logged int he console anyway

LDAP: I had some issues in the past where users just posted the message or a screenshot. In these cases, the exit code and exception were quite useful.

If we remove the exit code, should we remove the exception too?

rchl: In my experience exit code in most cases doesn't provide actionable information as most users don't know what each codes mean for specific servers. The exception can be very useful for diagnosing the issue on the other hand. And for those that really want to look deep into it, they can find exit code in the console.

LDAP: Right. Updated the message :)

rchl (approved):

PR #2146, opened 2022-12-20
fix completion documentation being parsed as markdown twice

rchl:
Based on investigation in https://github.com/sublimelsp/LSP-pyright/pull/207.

I believe this is correct since `minihtml_content` is built from parts that should have already gone through mdpopups markdown parsing but I might have missed something.

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -68,13 +68,13 @@ class LspResolveDocsCommand(LspTextCommand):
             if not self.view.is_valid():
                 return
             if self.view.is_popup_visible():
-                update_lsp_popup(self.view, minihtml_content, md=True)
+                update_lsp_popup(self.view, minihtml_content, md=False)
             else:
                 show_lsp_popup(
                     self.view,
                     minihtml_content,
                     flags=sublime.COOPERATE_WITH_AUTO_COMPLETE,
-                    md=True,
+                    md=False,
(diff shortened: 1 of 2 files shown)


predragnikolic (approved):
```py
class LspResolveDocsCommand(LspTextCommand):
    def _handle_resolve_response_async():
    	#...
        documentation = self._format_documentation(item.get("documentation") or "", language_map) # <- _format_documentation
        if documentation:
            minihtml_content += documentation

        def run_main() -> None:
                update_lsp_popup(self.view, minihtml_content, md=True) # <-
            else:
                show_lsp_popup(
                    ...
                    md=True, # <-

    def _format_documentation():
        return minihtml()


def minihtml():
   	# ...
    if is_plain_text:
        return "<p>{}</p>".format(text2html(result)) if result else ''
    else:
        return mdpopups.md2html(view, mdpopups.format_frontmatter(frontmatter) + result)
```

just a personal note:

We do not need to enable `md=True` for `update_lsp_popup` and `show_lsp_popup`
because `minihtml_content` markdown is already transformed to html with 
`_format_documentation` which calls `minihtml` which calls `mdpopups.md2html`.

predragnikolic:
Tested :heavy_check_mark:

rchl:
Yep, I'm also confident that this is correct.

PR #2147, opened 2022-12-21
when going to definition scroll to start of the region, not end

rchl:
Noticed that on "Goto definition", when the region of the definition spans many lines, ST scrolls to the end of the region rather than the start.

Imagine triggering "Goto definition" on a symbol that jumps to the constructor of that symbol and that constructor spans many lines. Focusing on the end of the constructor implementation is rather confusing and also it's not where the selection/cursor is in this case.

### DIFF
--- a/plugin/core/open.py
+++ b/plugin/core/open.py
@@ -117,10 +117,10 @@ def center_selection(v: sublime.View, r: Range) -> sublime.View:
     if window:
         window.focus_view(v)
     if int(sublime.version()) >= 4124:
-        v.show_at_center(selection, animate=False)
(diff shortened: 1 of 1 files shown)


LDAP:
Can you give an example? I could not reproduce this with pyright.

LDAP (approved):
Looks good.

predragnikolic (approved):
I didn't test this with a LS, but I did test the behvior of show_at_center, with this peace of code:
```py
view.show_at_center(sublime.Region(0, view.size()))
```
And yes, by default `show_at_center` will scroll to the end.

PR #2148, opened 2022-12-21
improve performance of completion & signature request on typing

rchl:
Before triggering completion or signature request, we are forcing a [`purge_changes_async()`](https://github.com/sublimelsp/LSP/blob/0ee0a7f6658805f61187a8a591b4980e99a0e5d5/plugin/session_buffer.py#L281-L281) because we have to. This immediately triggered various non-essential requests like semantic highlighting, inline hints, document link, color boxes which can dramatically reduce responsiveness due to those requests jumping in front of the queue. Instead trigger those in the next "tick" so that the completion/signature requests are triggered first.

Before those changes (with semantic highlighting enabled and being the main reason for worse responsiveness):

https://user-images.githubusercontent.com/153197/208928929-d35cef67-4223-4947-b00b-92e7f7fb2eda.mov

With those changes:

https://user-images.githubusercontent.com/153197/208928964-67127fc2-fa39-4581-a79f-02cc2f1fbef8.mov

### DIFF
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -279,29 +279,33 @@ class SessionBuffer:
     on_reload_async = on_revert_async
 
     def purge_changes_async(self, view: sublime.View) -> None:
-        if self.pending_changes is not None:
-            sync_kind = self.text_sync_kind()
-            if sync_kind == TextDocumentSyncKind.None_:
-                return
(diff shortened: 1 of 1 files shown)


rchl:
Disable showing whitespace changes if the diff is hard to read.

predragnikolic (approved):

PR #2149, opened 2022-12-22
Add "Source Action" entry to the "Edit" main menu

### LINKED ISSUE #2066: Suggestions for `Source Action...` menu handling

jwortmann:
Same as #2141, but for "Source Action".

I also moved "Format File" and "Format Selection" into a submenu.

edit-menu

Note that no diagnostics are sent when requesting the "source" and "refactor" actions, but I would expect those actions to not depend on diagnostics, so I think this should be okay.

It would be good if someone could test it with a server that also provides "source" actions.

---

And I've experimented a little bit with the context menu, but I'm not happy with how it turns out if the same submenus for "Refactor" and "Source Action" are applied to the context menu. Because those menu items for the submenus cannot be hidden if they contain no items. That means that they will always be present in the context menu, even if no LSP is running (for example for basic `.txt` files).

The following example demonstrates this, (here also with the "Goto..." commands grouped into a submenu):
menu

Furthermore, in a quick test it didn't work reliably for me from the context menu, even though (or maybe because) I think I properly took into account that the region for the code actions request should be derived from the mouse cursor / menu position, and not from the selection when triggered from the context menu. (I think this is also a bug currently existing in the `lsp_code_action` command, which doesn't work from the context menu if the selection/caret is at another position)

Due to these reasons the context menu is unchanged for now.

Closes #2066

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -353,6 +353,7 @@ class LspCodeActionsCommand(LspTextCommand):
 
 
 class LspMenuActionCommand(LspTextCommand, metaclass=ABCMeta):
+    """Handles a particular kind of code actions with the purpose to list them as items in a submenu."""
 
     capability = 'codeActionProvider'
 
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -84,18 +84,15 @@ class CodeActionsManager:
         def response_filter(session: Session, actions: List[CodeActionOrCommand]) -> List[CodeActionOrCommand]:
             # Filter out non "quickfix" code actions unless "only_kinds" is provided.
             if only_kinds:
+                code_actions = [cast(CodeAction, a) for a in actions if not is_command(a) and not a.get('disabled')]
                 if manual and only_kinds == MENU_ACTIONS_KINDS:
-                    code_actions = [cast(CodeAction, a) for a in actions if not is_command(a) and not a.get('disabled')]
                     for action in code_actions:
                         kind = action.get('kind')
                         if kinds_include_kind([CodeActionKind.Refactor], kind):
                             self.refactor_actions_cache.append((session.config.name, action))
                         elif kinds_include_kind([CodeActionKind.Source], kind):
                             self.source_actions_cache.append((session.config.name, action))
-                return [
-                    a for a in actions
-                    if not is_command(a) and kinds_include_kind(only_kinds, a.get('kind')) and not a.get('disabled')
-                ]
+                return [action for action in code_actions if kinds_include_kind(only_kinds, action.get('kind'))]
             if manual:
                 return [a for a in actions if not a.get('disabled')]
             # On implicit (selection change) request, only return commands and quick fix kinds.
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -17,12 +17,15 @@ from .core.views import first_selection_region
 from .core.views import format_code_actions_for_quick_panel
 from .core.views import text_document_code_action_params
 from .save_command import LspSaveCommand, SaveTask
+from abc import ABCMeta
+from abc import abstractmethod
 from functools import partial
 import sublime
 
 ConfigName = str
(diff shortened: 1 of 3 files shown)


jwortmann:
By the way, the ST core issue for the "hide empty submenu" problem is at https://github.com/sublimehq/sublime_text/issues/1859.

plugin/code_actions.py
```diff
@@ -63,9 +67,10 @@ def request_for_region_async(
                     return task
                 else:
                     self._response_cache = None
-        elif only_kinds == [CodeActionKind.Refactor]:
-            self.refactor_actions_key = "{}#{}:{}".format(view.buffer_id(), view.change_count(), region)
+        elif only_kinds == MENU_ACTIONS_KINDS:
```
rchl, line 70: Python will not handle it correctly when both lists have the same items but in different order but I guess you are assuming that that's OK because we expect to only care about our specific request that we make?

rchl (approved):

rchl:
> Furthermore, in a quick test it didn't work reliably for me from the context menu, even though (or maybe because) I think I properly took into account that the region for the code actions request should be derived from the mouse cursor / menu position, and not from the selection when triggers from the context menu. (I think this is also a bug currently existing in the `lsp_code_action` command, which doesn't work from the context menu if the selection/caret is at another position)

As far as that, I've also been thinking about this before and on one hand I agree that ideally the action should be triggered in the place that was right clicked rather than the current cursor position but on the other hand, this would be confusing because in ST the cursor doesn't move to the location that was right clicked unlike in many other programs. For example in VSCode it moves. In browsers it doesn't move in plain text field but it does in Discord/Slack input field or Gmail compose field for example.

Even in ST the built-in "Goto definition" seems to use the event position while the `Line History...` that opens Sublime Merge doesn't...

But still I'd lean towards using event position.

plugin/code_actions.py
jwortmann: If I understand it correctly, you mean for example that `[1, 2] != [2, 1]`, but that should be irrelevant here, because we always refer to the `MENU_ACTIONS_KINDS` variable now, both when passing `only_kinds` to the function for the request, and then on this line when comparing it.

jwortmann:
> because in ST the cursor doesn't move to the location that was right clicked unlike in many other programs. For example in VSCode it moves.

Interesting, I never particularly noticed that, but you are right. If we were a bit reckless, we could do the same:
```python
import sublime_plugin


class ContextMenuMoveCaretCommand(sublime_plugin.TextCommand):

    def is_enabled(self, event):
        return False

    def is_visible(self, event):
        if isinstance(event, dict) and 'x' in event and 'y' in event and self.view.settings().get('lsp_active'):
            pt = self.view.window_to_text((event['x'], event['y']))
            self.view.run_command('lsp_selection_set', {'regions': [(pt, pt)]})
        return False

    def want_event(self):
        return True

    def run(self, edit, event):
        pass
```

*Context.sublime-menu:*
```json
[
    { "command": "context_menu_move_caret" },
]
```
But not that I would reccomend it.

> But still I'd lean towards using event position.

I think I do too. It might be an uncomfortable change for a small number of users who got used to click anywhere else in the view to open code actions for the current selection. But on the other hand, the event position is already used for the "Goto..." commands and "Find References" (not for "Format Selection" and "Expand Selection", but for those it wouldn't make sense to not use the selection).

PR #2151, opened 2022-12-26
Implement call hierarchy request

### LINKED ISSUE #1429: Implement textDocument/callHierarchy

jwortmann:
There is no Sublime API to visualize and interact with a tree structure, like for example the folders and files in the sidebar. So I made my own implementation for that, using a HtmlSheet. Let me know what you think!

It can be tested e.g. with LSP-pyright:

[call_hierarchy.webm](https://user-images.githubusercontent.com/6579999/209542234-2d833a5e-ff26-4c54-9503-426ff0724d8c.webm)

~~For now only with "incoming calls" (which is quite similar to "Find References"), but just due to lack of a good solution how a toggle or so can be represented in the UI. VSCode has a small button to change between incoming/outgoing calls.~~

The "TreeView" is a quite generic implementation (inspired by the VSCode API). So it should be straightforward to reuse it for the "type hierarchy" request, but I haven't tested that because I don't know which language server supports type hierarchy requests.

Closes #1429.


rchl:
Nice!
For now just one initial feedback after quick testing that the initial expand should be done automatically as this is the point of opening it in the first place.

predrag-codetribe:
This is really nice!

Here is just one issue I have noticed:
My OS, doesn't recognise `&#9205;` or `&#9207;` (12.2.1 On MacOS 12.2.1)
Screenshot 2022-12-27 at 10 02 14
Screenshot 2022-12-27 at 10 13 19

it shows a question mark.

----


One other thing,
Currently when I click a node, 
it will take me to the Location and close the Call Hierarchy. 
I will just see one file.

I would rather open a side by side view.
That way I could still see the Call Hierarchy and click other Locations, 
to have a better look how they are called.
```diff
    def get_tree_item(self, element: CallHierarchyItem) -> TreeItem:
        command_url = sublime.command_url('lsp_open_location', {
            'location': {
                'targetUri': element['uri'],
                'targetRange': element['range'],
                'targetSelectionRange': element['selectionRange']
            },
            'session_name': self.session_name,
+            'flags': sublime.ADD_TO_SELECTION | sublime.CLEAR_TO_RIGHT
        })
```

https://user-images.githubusercontent.com/115091796/209645092-2ed3809c-53d2-469b-81dc-aed27fd755e9.mp4

WDYT?

jwortmann:
The first item is expanded now, and I added a toggle button to switch between incoming / outgoing calls.

Not sure what's going on with the CI. The lint errors seem to be completely unrelated to this PR. But I don't get those errors on the main branch...
If you merge my other PR, I can rebase on main and see if I can fix them.

jwortmann:
> My OS, doesn't recognise `&#9205;` or `&#9207;` (12.2.1 On MacOS 12.2.1)

Oh, I see. I think it uses the code font for these characters, i.e. the font you configured in *Preferences.sublime-settings*. And it seems those characters are not supported by this font. I guess we can try to set the "system" font for them, and then see if the arrows are rendered correctly.

> One other thing, Currently when I click a node, it will take me to the Location and close the Call Hierarchy. I will just see one file.
> 
> I would rather open a side by side view. That way I could still see the Call Hierarchy and click other Locations, to have a better look how they are called.

Yeah I had this in mind already, but haven't tried it yet. I wonder whether there is a way to open the new tab to the left of the "Call Hierarchy" sheet. So that the call hierarchy sheet will always be the rightmost tab.
But it doesn't look bad in your recording.

predrag-codetribe:
I also tried system, but it didn't work.
Other apps also do not render that symbol.

Screenshot 2022-12-27 at 11 19 07

I found the following two chars to work on my system:  `▶` and `▼`

predrag-codetribe:
Also, it would be nice to distinguish items, that do not have children. 
If a node do not have children, do not render ▶ or ▼, instead, use a different symbol, or no symbol at all?

... That doesn't have to be part of this PR.

plugin/core/tree_view.py
```diff
@@ -0,0 +1,269 @@
...
+        self.collapsible_state = TreeItemCollapsibleState.COLLAPSED
+        self.id = str(uuid.uuid4())
+
+    def html(self, sheet_name: str, indent_level: int) -> str:
+        indent_html = '<span style="padding-left: {}rem;">&nbsp;</span>'.format(indent_level)
+        if self.collapsible_state == TreeItemCollapsibleState.COLLAPSED:
+            disclosure_button_html = '<a class="disclosure-button" href="{}">&#9654;</a>'.format(
+                sublime.command_url('lsp_expand_tree_item', {'name': sheet_name, 'id': self.id}))
+        elif self.collapsible_state == TreeItemCollapsibleState.EXPANDED:
+            disclosure_button_html = '<a class="disclosure-button" href="{}">&#9660;</a>'.format(
```
rchl, line 51: Sorry for nitpcking but we could likely just paste an actual  utf8 characters instead of using entities. Entities are less readable for the developer and I don't think there is really a benefit in using entities (as a web dev I should know but frankly I don't really ;)).

jwortmann: The reason I tend to avoid uncommon unicode characters if there is an ASCII equivalent, is that often they mess up the horizontal alignment in the source code. For example if you have `"rulers": [120]` set for Python, then the ruler is not reliable anymore for such lines (and often I don't know which unicode characters count as 1 or as 2 character-width).

emoji

But here I can change it, no prob

plugin/core/registry.py
```diff
@@ -220,3 +275,31 @@ class LspPrevDiagnosticCommand(LspTextCommand):
 
     def run(self, edit: sublime.Edit, point: Optional[int] = None) -> None:
         navigate_diagnostics(self.view, point, forward=False)
+
+
+class LspExpandTreeItemCommand(LspWindowCommand):
```
rchl, line 280: Those commands and call-hierarchy specific but the name doesn't suggest it.
Also those have quite a bit of duplicated logic.

I'm wondering about an alternative ways to do it:
 - Hav a common base for sheet commands so that we can reuse the code to lookup the sheet at least
 - or... Have a single command for handling (call hierarchy) sheet commands that accepts an argument to decide what action within the sheet should be done.

jwortmann: They are not call-hierarchy specific. I intentionally made the "TreeViewSheet" implementation as generic as possible. So that it can be reused for type-hierarchy, for example. If you want to implement the type-hierarchy request, you can probably just copy the *call_hierarchy.py* file, change a handful of lines (mainly `get_children` and `get_tree_item` methods) and it should work.

> * or... Have a single command for handling (call hierarchy) sheet commands that accepts an argument to decide what action within the sheet should be done.

Not possible/useful for the reason explained above.

rchl: I'm talking about the `LspExpandTreeItemCommand` / `LspCollapseTreeItemCommand` commands.

jwortmann: I mean you can of course combine the "expand" and "collapse" commands into a single command, and I actually wanted to do this initially, but then you need to pass another variable that you have to handle within the command.
And you can't combine it with the `LspCallHierarchyToggleCommand`, because that one is indeed call-hierarchy specific.

rchl: Though `LspExpandTreeItemCommand` and `LspCollapseTreeItemCommand` also check for call hierarchy capability so can't be used for other things.

plugin/core/tree_view.py
jwortmann: I replaced the arrow symbols, but left the entity in place of the ⇄ symbol for now. Because that one is exactly such a case with the implication from the GIF (on my OS and system font).

plugin/core/registry.py
jwortmann: > also check for call hierarchy capability

Fixed in https://github.com/sublimelsp/LSP/pull/2151/commits/ec84c88fece861dc5fa713efb37d7f226907c48e

rchl:
General note: The "Incoming" / "Outgoing" naming from the spec is IMO not very easy to understand. I think we could adopt the VSCode naming which uses "Callers of" / "Calls from".

rchl:
(BTW. Adding support for Call Hierarchy in https://github.com/typescript-language-server/typescript-language-server/pull/649 also)

plugin/call_hierarchy.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def get_tree_item(self, element: CallHierarchyItem) -> TreeItem:
+        command_url = sublime.command_url('lsp_open_location', {
+            'location': {
+                'targetUri': element['uri'],
+                'targetRange': element['range'],
+                'targetSelectionRange': element['selectionRange']
+            },
+            'session_name': self.session_name,
+            'flags': sublime.SEMI_TRANSIENT | sublime.ADD_TO_SELECTION
```
predrag-codetribe, line 73:
```suggestion
            'flags': sublime.ADD_TO_SELECTION | sublime.CLEAR_TO_RIGHT
```

I would add  `sublime.CLEAR_TO_RIGHT`, to avoid too many open views.
Screenshot 2022-12-28 at 11 37 37

jwortmann: I see the exact behavior from your screenshot with `sublime.ADD_TO_SELECTION | sublime.CLEAR_TO_RIGHT`.
But I see now that even without CLEAR_TO_RIGHT this still happens if the links are to different files.
It's probably a bug which has been fixed on the latest dev built. So maybe it's better to remove the `"flags"` argument alltogether, until there is another ST stable build with the bugfix?

jwortmann: I use another approach now:
If you hold <kbd>Ctrl</kbd> or <kbd>Shift</kbd> when clicking on the link, then it will be added to the selection, and otherwise it will open as a single regular tab which takes the focus. This means that you have to know that you have to use the modifier key if you want the call hierarchy tab to be still visible after clicking on a item label, but this is consistent with how the built-in definitions/references popup from ST works.

We can reconsider using the ADD_TO_SELECTION flag by default when a new stable build of ST with the mentioned bugfix lands.

boot.py
```diff
@@ -190,27 +195,39 @@ def on_post_window_command(self, window: sublime.Window, command_name: str, args
                 sublime.set_timeout(panel_manager.update_log_panel)
 
 
-class LspOpenLocationCommand(sublime_plugin.TextCommand):
+class LspOpenLocationCommand(LspWindowCommand):
```
rchl, line 198: I found two places that ran this command from the view:
 - https://github.com/sublimelsp/LSP-volar/blob/32f0c42ecf2d10036a913518938a91d704c809f1/plugin.py#L69-L69
 - https://github.com/sublimelsp/LSP-typescript/blob/e6ae40de6c3322fbe064698ff49f958ab6b3e4f9/commands.py#L35-L35

boot.py
```diff
@@ -190,27 +195,39 @@ def on_post_window_command(self, window: sublime.Window, command_name: str, args
...
+
     def _run_async(
-        self, location: Union[Location, LocationLink], session_name: Optional[str], flags: int = 0, group: int = -1
+        self, location: Union[Location, LocationLink], session_name: Optional[str], flags: int, group: int
     ) -> None:
-        manager = windows.lookup(self.view.window())
-        if manager:
-            manager.open_location_async(location, session_name, self.view, flags, group) \
+        if session_name:
+            self.session_name = session_name
```
rchl, line 227: Doesn't seem right to set this as a class property. I believe that would mean that the command becomes specific to that session from that point and won't be available for other sessions anymore because `is_enabled` will only get sessions that match `self.session_name`.

Can't this just use `self.session_by_name` instead?

plugin/call_hierarchy.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def __init__(
+        self,
+        window: sublime.Window,
+        session_name: str,
+        direction: CallHierarchyDirection,
+        root_elements: List[CallHierarchyItem]
+    ) -> None:
+        self.window = window
+        self.session_name = session_name
```
rchl, line 40: It might not be the only place that has this issue but I feel that it might not be correct to use `session_name` for matching as there might be multiple sessions with same config name in a single window. Maybe let's instead create a weak reference to session (`weakref.ref(session)`) and use that?

plugin/call_hierarchy.py
```diff
@@ -0,0 +1,157 @@
...
+
+    capability = 'callHierarchyProvider'
+
+    def is_visible(self, event: Optional[dict] = None, point: Optional[int] = None) -> bool:
+        if self.applies_to_context_menu(event):
+            return self.is_enabled(event, point)
+        return True
+
+    def run(self, edit: sublime.Edit, event: Optional[dict] = None, point: Optional[int] = None) -> None:
+        self._window = self.view.window()
```
rchl, line 103: Would generally prefer validating with window manager at every point so that we stop processing everything as soon as it goes away.

plugin/call_hierarchy.py
```diff
@@ -0,0 +1,157 @@
...
+            new_direction = CallHierarchyDirection.OutgoingCalls
+            tooltip = 'Show Outgoing Calls'
+        elif direction == CallHierarchyDirection.OutgoingCalls:
+            current_label = 'Calls from…'
+            new_direction = CallHierarchyDirection.IncomingCalls
+            tooltip = 'Show Incoming Calls'
+        else:
+            return
+        header = 'Call Hierarchy: {} <a href="{}" title="{}">&#8644;</a>'.format(
+            current_label, make_toggle_command(session_name, new_direction, root_elements), tooltip)
```
rchl, line 145: Would be nice to reuse `make_command_link` here instead of creating new function. I believe it would be enough to extend it with an optional `title` argument?

plugin/core/tree_view.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def __init__(self, element: T, tree_item: TreeItem, indent_level: int = 0) -> None:  # pyright: ignore
+        self.element = element
+        self.tree_item = tree_item
+        self.indent_level = indent_level
+        self.children = []  # type: List[str]  # IDs of child nodes
+        self.is_resolved = False
+
+
+class TreeDataProvider(metaclass=ABCMeta):
```
rchl, line 105: I believe it would be a bit more correct from types standpoint (I've noticed some mypy errors in subclass) to use `class TreeDataProvider(Generic[T]):` but then we couldn't take advantage of `@abstractmethod`. Or at least I don't see a way to use both.

rchl (commented):
Since we are OK with this approach, it would be nice if you could also add some bits about it to documentation (at least the new command name) and maybe add to main menu if relevant to you.

plugin/call_hierarchy.py
rchl: I can confirm that `sublime.CLEAR_TO_RIGHT` works nice in the latest dev build.

boot.py
jwortmann: Oh. Good catch, that would be an ugly bug. Should be fixed in https://github.com/sublimelsp/LSP/pull/2151/commits/30a93550ee1637cd3b7197518aeebe8f07adf460.

plugin/call_hierarchy.py
jwortmann: I use a weakref for the CallHierarchyDataProvider now https://github.com/sublimelsp/LSP/pull/2151/commits/fbfc6cc67c6ca6afa17d2090b4866e03b427a6b0. But I think the same is not possible for the "toggle direction" command, because you can only pass basic JSON-compatible values in the [command args](https://www.sublimetext.com/docs/api_reference.html#sublime.CommandArgs).

rchl: Might be worth investigating adding a unique ID to session and lookup through that. But not in this PR of course.

predragnikolic (approved):

rwols (approved):

boot.py
```diff
@@ -190,27 +195,37 @@ def on_post_window_command(self, window: sublime.Window, command_name: str, args
...
+    def want_event(self) -> bool:
+        return True
+
     def _run_async(
-        self, location: Union[Location, LocationLink], session_name: Optional[str], flags: int = 0, group: int = -1
+        self, location: Union[Location, LocationLink], session_name: Optional[str], flags: int, group: int
     ) -> None:
-        manager = windows.lookup(self.view.window())
-        if manager:
-            manager.open_location_async(location, session_name, self.view, flags, group) \
```
rchl, line 213: I would now remove the `manager.open_location_async` implementation since it doesn't appear to be used anymore (even within third-party packages).

boot.py
```diff
@@ -22,10 +24,13 @@
 from .plugin.core.protocol import Error
 from .plugin.core.protocol import Location
 from .plugin.core.protocol import LocationLink
+from .plugin.core.registry import LspCollapseTreeItemCommand
+from .plugin.core.registry import LspExpandTreeItemCommand
 from .plugin.core.registry import LspNextDiagnosticCommand
 from .plugin.core.registry import LspPrevDiagnosticCommand
 from .plugin.core.registry import LspRecheckSessionsCommand
 from .plugin.core.registry import LspRestartServerCommand
+from .plugin.core.registry import LspWindowCommand
```
jwortmann, line 33: I wonder if we should move LspOpenLocationCommand somewhere else (core/registry.py?), and just import it in this file, so that this import line can be removed. Currently this `lsp_window` command would be exposed to users, but it has no real use by itself.

jwortmann: Since there were no comments, and I didn't see a reason to have this single command being defined in `boot.py`, I moved it into `core/registry.py`. If there are any objections, I could revert the commit.

plugin/call_hierarchy.py
```diff
@@ -0,0 +1,168 @@
...
+        if not session:
+            return
+        data_provider = CallHierarchyDataProvider(weaksession, CallHierarchyDirection.IncomingCalls, response)
+        header = 'Call Hierarchy: Callers of… {}'.format(
+            make_command_link('lsp_call_hierarchy_toggle', "⇄", {
+                'session_name': session.config.name,
+                'direction': CallHierarchyDirection.OutgoingCalls,
+                'root_elements': response
+            }, tooltip="Show outgoing calls"))
+        new_tree_view_sheet(self._window, "Call Hierarchy", data_provider, header, flags=sublime.ADD_TO_SELECTION)
```
rchl, line 138: A question popped into my head...

Would it make sense to actually open the call hierarchy view in its own, primary tab instead of attaching it to the active tab and then ending up with three tabs after clicking something? It takes too much space to be usable with three tabs opened at once IMO.

jwortmann: Certainly possible, but since the call hierarchy doesn't use much horizontal space (unless you have excessively long function names...), I thought it would be nice to have it as a side-by-side view, so that you don't lose the context. And you can just click on the call hierarchy tab in the tab bar if you want to exit the side-by-side mode and only focus this tab.

I have made quite good experiences with a custom "Show Documentation" command from LSP-julia, which I also open in the same way as a side-by-side HtmlSheet. But of course it's very subjective and I haven't really used the call hierarchy functionality in practice yet.

We could also consider to add a `side_by_side` command argument, like the "Goto..." commands already have. It probably wouldn't be very useful for the context menu, because the menu can't easily be configured by users. But it might be useful in case someone wants to add a custom key binding for it.

rchl: > I thought it would be nice to have it as a side-by-side view, so that you don't lose the context.

Just gonna throw another idea out there but it might hacky or hard to do

Open call hierarchy in a new new tab and then add the previously open file to selection (I guess that would technically open it again in the same location). Then allow the user to just click call hierarchy items to replace the view on the right.

> I have made quite good experiences with a custom "Show Documentation" command from LSP-julia

I haven't tried it but I imagine documentation sheet doesn't need to open further sheets on the right which is the main thing here?

> We could also consider to add a `side_by_side` command argument, like the "Goto..." commands already have.

I suppose the default behavior could be controlled by an LSP option (as much as we hate adding new options).

predragnikolic: Also to throw another idea. (not for now) 

Instead of a sheet for the call hierarchy use a regular View.
Benefits, it allows to have a cursor.
- We could than move in the View with the cursor, 
and add a keybinding for example. `space` or `enter` to toggle the nodes. That way we don't need to use the mouse.
- We can lock the view to not be editable.

rchl: I agree that the sheet is not very Sublime-like since it's not keyboard accessible. Plain view should allow more or less equivalent functionality.

But I'm not suggesting to change it now and also it wouldn't really affect this specific issue I've mentioned in this thread.

jwortmann: I don't really think that a regular view would be better:

* It probably wouldn't look as nice with the mouse cursor always visible and everthing forced as monospace font (including the header) - the HtmlSheet allows more styling
* It would be unexpected if you bind the left and right arrow keys to collapse / expand a tree item like in the sidebar, because there is still a cursor visible which you couldn't move to the left/right then.
* If the disclosure buttons (triangles) should be clickable with the mouse, then they must be phantoms, which probably would make things very ugly from the implementation side, because you need to keep track of all the positions when you update the icons, while new rows for child items are being added or removed. Because phantom's contents can't be updated, so you would need to recreate them each time when an icon should change.
* You can't have the kinds rendered like an icon, unless they are phantoms.
* You can't click the function name to open the location, unless it is a phantom too. But then basically everything in the view is a phantom again... so you could use the HtmlSheet in the first place instead :-)
* Can't have a tooltip when hovering over the function names if they are plain text.
* If you want to style the colors, then you need to implement a syntax definition for this view.

So basically it amounts to only keyboard navigation (regular view) vs. only mouse navigation (HtmlSheet), and I would say for such trees mouse navigation feels more natural.

The only other advantage I see, is that you could place a regular view as an output panel at the bottom. But I think that's also not really beneficial for the call hierarchy tree view, because it usually needs more vertical space than horizontal space.

If ST some day adds a sidebar API to render custom trees, then yes, we should switch to that, but right now the HtmlSheet seems to be the simplest and best we can do from my point of view.

---

> Just gonna throw another idea out there but it might hacky or hard to do
> 
> Open call hierarchy in a new new tab and then add the previously open file to selection (I guess that would technically open it again in the same location). Then allow the user to just click call hierarchy items to replace the view on the right.

I just tried, and `window.open_file("...", flags=sublime.ADD_TO_SELECTION|sublime.SEMI_TRANSIENT)` for the original view would open it a second time to the right. But in order to replace it with new views, it needs to be semi-transient, and as already mentioned the SEMI_TRANSIENT flag doesn't work reliably in the API functions currently. Also I haven't/can't check what happens if it has unsaved changes and you want to replace it with another view when clicking on the call hierarchy items. It would be annyoing if ST asks to save the file in that case.

plugin/call_hierarchy.py
jwortmann: Is this comment still relevant? Because the CallHierarchyDataProvider checks each time in `get_children` with the window manager for the session with the correct name, and if the session dies you can't interact with the call hierarchy sheet anymore. So as I see it, the only place where we could add another check would be in `_handle_response_async` here, but the time between initial request and response is probably only a few milliseconds.

Otherwise, could you tell what exactly should I change?

plugin/call_hierarchy.py
```diff
@@ -0,0 +1,168 @@
...
+        session = weaksession()
+        if not session:
+            return
+        data_provider = CallHierarchyDataProvider(weaksession, CallHierarchyDirection.IncomingCalls, response)
+        header = 'Call Hierarchy: Callers of… {}'.format(
+            make_command_link('lsp_call_hierarchy_toggle', "⇄", {
+                'session_name': session.config.name,
+                'direction': CallHierarchyDirection.OutgoingCalls,
+                'root_elements': response
+            }, tooltip="Show outgoing calls"))
```
rchl, line 137: Is it intended that clicking this opens a new sheet? I think that's not how it worked when I've tested it initially.

rchl: Actually that only happens after restarting ST and interacting with pre-existing sheet.

plugin/core/registry.py
```diff
@@ -12,12 +17,56 @@
...
+                    if isinstance(sheet, sublime.HtmlSheet) and sheet.id() == sheet_id:
+                        if sheet not in selected_sheets:
+                            selected_sheets.append(sheet)
+                            window.select_sheets(selected_sheets)
+                        break
+            else:
+                window.focus_sheet(tree_view_sheet)
+            return tree_view_sheet
+    tree_view_sheet = TreeViewSheet(
+        sublime_api.window_new_html_sheet(window.window_id, name, "", flags, group),
```
rchl, line 61: Why are we using internal ST API here? Doesn't the public `window.new_html_sheet` do the same?

plugin/call_hierarchy.py
jwortmann: The `new_tree_view_sheet` function, which is called when you click on this, should ensure that there is always only one sheet with a particular name (e.g. "Call Hierarchy", "Type Hierarchy", etc.) per window, and the sheet is stored in the `WindowManager`. But yeah, if you restart ST, then this data isn't preserved and a new `WindowManager` is initialized with empty data. I didn't run into this because I use `"hot_exit": "disabled"`.

Also interaction with the tree view doesn't work anymore if you drag the tab out of the window, for the same reason. But I think this is a broader issue in general, and various other LSP features suffer from the same problem. This is mainly a fundamental problem of the idea to associate "workspace folders" or "sessions" to a particular window, and it probably can't be solved completely.

rchl: I don't mind this issue for now.
Ideally the sheet would not survive the restart IMO but I suppose there is no easy way to achieve that right now.

plugin/core/registry.py
jwortmann: It's a bit difficult to explain, but the alternative would be something like
```python
_throwaway_sheet = window.new_html_sheet(name, "", flags, group)
tree_view_sheet = TreeViewSheet(_throwaway_sheet.id(), name, data_provider, header)
return tree_view_sheet
```
and because I made `TreeViewSheet` a subclass of `HtmlSheet` (so that it can use its `set_name` and `set_contents` methods, and so that things like `isinstance(HtmlSheet, ...)` works), it will also call its `__init__` method and initialize another `sublime.HtmlSheet`. So we basically have two HtmlSheet objects with the same id. I'm not sure if that would do any harm, because the only thing its (respectively the `Sheet`'s) `__init__` method does is to assign the `sheet_id` property, but that still felt wrong to me I could imagine that in theory it might be a potential cause of bugs if ST decides at some point to do more stuff in the `Sheet`'s `__init__` method (which probaly won't happen, I know, but still....).

So I basically just wanted to avoid that two `HtmlSheet` objects are created with the same id.

rchl: Right. There would be no question if I'd notice that `window.new_html_sheet` calls `make_sheet` before returning.

plugin/call_hierarchy.py
rchl: I can merge it and we can refine it later or we can wait until next ST version and make use of SEMI_TRANSIENT then.

I'm not too happy with how it works now (by default) where if the clicked entry is in the same file it will scroll to it but if it's in another then it will open new file. I would like this to be more consistent.

What would you prefer? Get this in or wait?

jwortmann: > What would you prefer? Get this in or wait?

Based on the previous release frequency, the next stable release might be several months or even longer away from now. We probably don't want to wait *that* long.

I was thinking if it would be better to make the behavior when holding <kbd>Ctrl</kbd> / <kbd>Cmd</kbd> while clicking on an item, to be the default, even if no modifier key is used. Because currently there is no indication that you are able to use the modifier keys when clicking the item, so I could imagine that most people don't find out about this ability at all. But on the other hand that would be inconsistent with how ST's built-in features work, for example the links from the "Definitions popup" (this one shows a hint for the modifier key in the tooltip, but in our case I already used the tooltip for the file location of the link). And in general I'm in favor of consistency and to follow ST default behavior.

Another option would be to add an `if int(sublime.version()) > 4144` check (I think this bug is `https://github.com/sublimehq/sublime_text/issues/5729` and was fixed in that version) and have two different code paths, at least until the next stable ST release. But I'm not sure if we want to add this additional complexity and maintenance effort.

rchl: > I was thinking if it would be better to make the behavior when holding Ctrl / Cmd while clicking on an item, to be the default

I think that would be better than the current behavior.

> Because currently there is no indication that you are able to use the modifier keys when clicking the item,

Since we have a lot of flexibility with minihtml, we can easily add a text that would mention that in the output.

> have two different code paths, at least until the next stable ST release

I would be fine with that if it's just temporary and wouldn't require a ton of extra code.

jwortmann: I added a commit to always open the links in side-by-side mode now. I think I've tried this already before https://github.com/sublimelsp/LSP/pull/2151#discussion_r1058283725, but the trick was to also use the `sublime.SEMI_TRANSIENT` flag to make it work.

Also it will put the HtmlSheet on the left side now, and already open the location of the root item to the right.

* Is it better than before?
* Does it work on the latest dev build?

If not, I could revert the last commit.

rchl: It works really nice now besides the fact that we are hitting https://github.com/sublimehq/sublime_text/issues/2411 for the initially open file. Switching to another file and back makes it work for the initial file.


https://user-images.githubusercontent.com/153197/213924885-5702dee5-730d-4519-8f73-57bf64ad13dc.mov

plugin/core/protocol.py
```diff
@@ -5912,6 +5912,18 @@ def semanticTokensFullDelta(cls, params: Mapping[str, Any], view: sublime.View)
...
+    def prepareCallHierarchy(cls, params: CallHierarchyPrepareParams, view: sublime.View) -> 'Request':
+        return Request("textDocument/prepareCallHierarchy", params, view)
+
+    @classmethod
+    def incomingCalls(cls, params: CallHierarchyIncomingCallsParams) -> 'Request':
+        return Request("callHierarchy/incomingCalls", params, None)
+
+    @classmethod
+    def outgoingCalls(cls, params: CallHierarchyOutgoingCallsParams) -> 'Request':
+        return Request("callHierarchy/outgoingCalls", params, None)
```
rchl, line 5925: It would probably be good to enable `progress` for those since this is user-initiated action and could in theory take a second to complete.

plugin/call_hierarchy.py
rchl: I've pushed a work around for that. I think it's good to have it as it should help non-call-hierarchy cases also.

rchl: I'll put this change in a separate PR (https://github.com/sublimelsp/LSP/pull/2174) so that we have it on record and change log. Otherwise this looks good to me.

plugin/core/open.py
```diff
@@ -86,8 +87,13 @@ def open_file(
     if view and _return_existing_view(flags, window.get_view_index(view)[0], window.active_group(), group):
         return Promise.resolve(view)
 
+    was_already_open = view is not None
     view = window.open_file(file, flags, group)
     if not view.is_loading():
+        if was_already_open and (flags & sublime.SEMI_TRANSIENT):
+            # work-around bug https://github.com/sublimehq/sublime_text/issues/2411 where transient view might not get
+            # its view listeners initialized.
+            sublime_plugin.check_view_event_listeners(view)  # type: ignore
```
predragnikolic, line 96: I tested this.
This needs to exist, if not the server will not start in the side by side view. 

@rchl I don't know how you managed to find check_view_event_listeners, but it works :)

plugin/call_hierarchy.py
predragnikolic: > Is it better than before?
Yes, definitely!  
Pretty happy with how this works

plugin/core/open.py
rchl: deathaxe found this out in the core issue

rchl (approved):

PR #2159, opened 2023-01-01
Allow missing window/workDoneProgress/create request from the server

rwols:
This commit adds a workaround for servers that report server-initiated progress but don't request window/workDoneProgress/create before reporting progress.

I haven't actually tested whether this works with jdlts regarding its progress reporting behavior in https://github.com/sublimelsp/LSP-jdtls/issues/37, could you check @LDAP ?

### DIFF
--- a/plugin/active_request.py
+++ b/plugin/active_request.py
@@ -7,6 +7,7 @@ from .core.typing import Any, Optional, Dict
 from weakref import ref
 import sublime
 
+
 class ActiveRequest:
     """
     Holds state per request.
--- a/plugin/active_request.py
+++ b/plugin/active_request.py
@@ -1,0 +1,72 @@
+from .core.sessions import SessionViewProtocol
+from .core.progress import ProgressReporter
+from .core.progress import ViewProgressReporter
+from .core.progress import WindowProgressReporter
+from .core.protocol import Request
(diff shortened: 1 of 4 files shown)


rchl:
It's nothing new in this PR but we shouldn't be relying on destructor for updating UI as destructor is not guaranteed to run immediately and that can result in progress lingering for longer than necessary. For example this was easily visible before (and still is for some features) when restarting the server. So regions would keep showing for a second or so after restarting a server on my system.

rwols:
I understand the concern but I don't quite know how to address it.

plugin/active_request.py
```diff
@@ -0,0 +1,73 @@
+from .core.sessions import SessionViewProtocol
```
rchl, line 1: Shouldn't this file live in `core/` since it only deals with core types?

LDAP (approved):
PR fixes the issues with JDTLS.

plugin/active_request.py
rchl: @rwols you want me to address that?

rwols: No let me do it :D I'll get to it this evening, I hope.

rchl (approved):

PR #2160, opened 2023-01-01
Implement a view size limit for LSP protocol to be enabled on views

TerminalFi:
Implements a new setting `lsp_view_size_limit`

This setting can be used to control whether or not LSP is enabled on a view depending on it's returned `view.size()`


rchl:
I'm not against it but at the minimal there should be some feedback to the user that the view is ignored.

Ideally also a way for the user to override that decision per-file temporarily but since ST is lacking some form of notification popups, I don't see how this could be done in a way that is easily discoverable.

TerminalFi:
I'm going to rethink this approach and see if there is a way to make it work. 

Sublime notification api would be nice.

Issue #2158, opened 2023-01-01
Backslash in hover popup is not shown


jfcherng:
**Describe the bug**

As the title.

Original response payload:

```
:: <<< LSP-intelephense 22: {'range': {'start': {'line': 4, 'character': 6}, 'end': {'line': 4, 'character': 9}}, 'contents': {'kind': 'markdown', 'value': '__A\\B\\C\\Foo__\n\n```php\n<?php\nclass Foo { }\n```'}}
```

**To Reproduce**

1. Have the following codes with intelephense.

```php
<?php

namespace A\B\C;

class Foo {}
```

2. Hover on `Foo`

**Expected behavior**

The backslashes show as-is.

**Screenshots**

![image](https://user-images.githubusercontent.com/6594915/210172427-8df90198-2cad-45f7-a6bd-5b9c44c8d158.png)

**Environment (please complete the following information):**
- OS: Win10 x64 22H2
- Sublime Text version: 4147
- LSP version: [4070-1.21.0](https://github.com/sublimelsp/LSP/releases/tag/4070-1.21.0)
- Language servers used: intelephense

rchl:
Minimal reproduce:

```py
mdpopups.show_popup(view, mdpopups.md2html(view, mdpopups.format_frontmatter({'markdown_extensions': [{'pymdownx.escapeall': {}}]}) + '__A\\B\\C\\Foo__'))
```

I'll have to throw it at @facelessuser  :)

facelessuser:
Please reference the documentation for the plugin: https://facelessuser.github.io/pymdown-extensions/extensions/escapeall/.

If you have any questions or confusion about its behavior after reading, please feel free to ask. I feel like the documentation is pretty clear, but I'm happy to update if there is anything unclear.

rchl:
Right, makes sense.

The [PR that added `pymdownx.escapeall`](https://github.com/sublimelsp/LSP/pull/1188) doesn't reference any issues but I believe it was meant to fix https://github.com/sublimelsp/LSP-css/pull/16 and the  `LSP-css` case seems to be working now without `pymdownx.escapeall` (probably fixed in the language server itself) so I would lean towards just remove that extension.

rchl:
Though the `hardbreak` option of `pymdownx.escapeall` was later enabled in https://github.com/sublimelsp/LSP/pull/1591 to fix different case. And that case still seems to need this workaround.

[...]

PR #2163, opened 2023-01-04
Don't use "escapeall" extension when formatting with mdpopups

rchl:
The motivation is provided in #2158 but basically [both](https://github.com/sublimelsp/LSP-css/pull/16) [issues](https://github.com/sublimelsp/LSP/pull/1591) that we had that needed this extension don't seem to manifest itself anymore and there is a new #2151 issue that it causes.

Fixes #2151

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -721,12 +721,6 @@ def minihtml(
             "markdown_extensions": [
                 "markdown.extensions.admonition",
                 {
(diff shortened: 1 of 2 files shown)


rchl:
Hmm, we might still want the `'hardbreak': True` functionality from that extension but without any other logic. This is not currently possible. That's because in github markdown a single `\n` creates new line but not in flavor used by mdpopups.

PR #2166, opened 2023-01-08
Make Document Symbols behavior more consistent with built-in Goto Symbol

jwortmann:
Some suggestions for the "LSP: Goto Symbol..." feature to follow ST's "Goto Symbol..." more closely (the current implementation is closer to what VSCode does, though, but I don't think that matters much).

* If the server response is of type SymbolInformation (deprecated), then adjust the outline color for highlighted items to use the color from the symbol item kind (which is defined by the theme, but there is a strong recommendation in the [theme docs](https://www.sublimetext.com/docs/themes.html#quick-panel) under `kind_container`, and the built-in themes follow those recommendation), rather than the scope color from the color scheme for the symbol. By using the theme's kind color the outline matches the icon color in the quick panel.
* If the server response is of type DocumentSymbol, i.e. it has a `selectionRange`, select that range when a quick panel item is selected and even when highlighted, similar to ST's built-in "Goto Symbol". I had previously questioned the current behavior of the empty selection at https://github.com/sublimelsp/LSP/pull/1031#discussion_r429376293, but I think to select the given "selectionRange" should be fine if we differentiate based on response type. The full `range` isn't highlighted anymore for this response type now, and according to the LSP specs it seems to be intended to highlight items in a permanent outline view (which doesn't exist in ST):
  > This information is typically used to determine if the clients cursor is inside the symbol to reveal in the symbol in the UI.
* Minor code cleanup: remove unused `format_symbol_kind` function, replace `if len(some_list) > 0:` by `if some_list:`

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -196,34 +196,26 @@ SUBLIME_KIND_SCOPES = {
     sublime.KIND_VARIABLE: "entity.name.constant | constant.other | support.constant | variable.other | variable.parameter | variable.other.member | variable.other.readwrite.member"  # noqa: E501
 }  # type: Dict[SublimeKind, str]
 
-SYMBOL_KIND_SCOPES = {
-    SymbolKind.File: "string",
-    SymbolKind.Module: "entity.name.namespace",
-    SymbolKind.Namespace: "entity.name.namespace",
-    SymbolKind.Package: "entity.name.namespace",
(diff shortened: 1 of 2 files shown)


rchl:
I like how it works but will check out the code as soon as possible.

rchl (approved):

PR #2172, opened 2023-01-21
trigger "on_server_response_async" also for "initialize" response

rchl:
We didn't trigger `AbstractPlugin.on_server_response_async` for the `initialize` response because plugin instance is created a bit later, from the initialize handler. Have exception for that case and trigger it manually from the handler.

I'm contemplating using it in LSP-rome to show server version that is provided in initialize response.

Also making `Repsonse` type Generic and more accurate. Not really doing much to improve anything. Just makes it easier to annotate in `on_server_response_async`.

### DIFF
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -5987,6 +5987,7 @@ class Error(Exception):
 
 T = TypeVar('T', bound=Union[None, bool, int, Uint, float, str, Mapping[str, Any], Iterable[Any]])
 
+
 class Response(Generic[T]):
 
     __slots__ = ('request_id', 'result')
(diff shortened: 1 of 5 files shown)


plugin/core/protocol.py
```diff
@@ -5985,11 +5985,14 @@ def from_exception(cls, ex: Exception) -> 'Error':
         return Error(ErrorCodes.InternalError, str(ex))
 
 
-class Response:
+T = TypeVar('T', bound=Union[None, bool, int, Uint, float, str, Mapping[str, Any], Iterable[Any]])
+
+
+class Response(Generic[T]):
```
predragnikolic, line 5991: this is nice

predragnikolic (approved):

predragnikolic:
Tested :heavy_check_mark:

rchl:
Type issue is a regression in pyright 1.1.290 (https://github.com/microsoft/pyright/issues/3997).

I have no idea why it doesn't reproduce in LSP-pyright and only on command line though...

I will revert to the older version for the time being.

PR #2173, opened 2023-01-21
feat(API): allow setting additional text in permanent server status

rchl:
This allows a plugin to call `Session.set_config_status_async` which would add extra text in parenthesis in the server's permanent status. For example can add server version info like:

Screenshot 2023-01-21 at 23 00 06

It could also be used to show whether workspace-local or global server is used or anything else.


predragnikolic (approved):
Tested :heavy_check_mark:

Issue #1979, opened 2022-06-28
side_by_side feature sometimes works


dmilith:
**Describe the bug**
Given that keybindings have a command using the "side_by_side" arguments like this:

```js
{
    "command": "lsp_symbol_declaration",
    "keys": ["super+shift+e"], "context": [
          {"key": "selector", "operator": "equal", "operand": "source"}
      ],
    "args": {"side_by_side": true}
  }
```

is causing LSP features NOT to work in opened split window. It affects all LSP servers, it's a general issue.

**To Reproduce**
    Start ST
    Install any LSP package, make sure it works for your project
    Hit your shortcut action with "side_by_side" set to true
    Now on the split window (usually on the right) all LSP functionality is no longer enabled.

**Expected behavior**
All LSP functionality should work normally when side_by_side is set to true.

**Screenshots**

**Environment (please complete the following information):**
- OS: macOS 12.4
- Sublime Text version: 4134
- LSP version: 1.16.3
- Language servers used: LSP-bash, LSP-shellcheck, LSP-Rust-analyzer, … the same for all I've tried

**Additional context**
none

rchl:
I believe that it only reproduces if the reference is in the same file you are already in. At least I was only able to reproduce in that case.

rchl:
It's this core issue: https://github.com/sublimehq/sublime_text/issues/2411

rchl:
Seems like we can workaround this issue. Created #2174

PR #2174, opened 2023-01-22
workaround for View Listeners not being attached for new transient view

rchl:
Workaround for core issue https://github.com/sublimehq/sublime_text/issues/2411

Fixes https://github.com/sublimelsp/LSP/issues/1979

### DIFF
--- a/plugin/core/open.py
+++ b/plugin/core/open.py
@@ -91,7 +91,7 @@ def open_file(
     view = window.open_file(file, flags, group)
     if not view.is_loading():
         if was_already_open and (flags & sublime.SEMI_TRANSIENT):
-            # work-around bug https://github.com/sublimehq/sublime_text/issues/2411 where transient view might not get
+            # workaround bug https://github.com/sublimehq/sublime_text/issues/2411 where transient view might not get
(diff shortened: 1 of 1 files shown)


plugin/core/open.py
```diff
@@ -86,8 +87,13 @@ def open_file(
     if view and _return_existing_view(flags, window.get_view_index(view)[0], window.active_group(), group):
         return Promise.resolve(view)
 
+    was_already_open = view is not None
     view = window.open_file(file, flags, group)
     if not view.is_loading():
+        if was_already_open and (flags & sublime.SEMI_TRANSIENT):
+            # work-around bug https://github.com/sublimehq/sublime_text/issues/2411 where transient view might not get
```
predragnikolic, line 94:
```suggestion
            # workaround bug https://github.com/sublimehq/sublime_text/issues/2411 where transient view might not get
```

To be consistent with how we wrote workaround in other places in code... so one day we can hopefully just remove those...

predragnikolic (approved):

predragnikolic:
tested :heavy_check_mark:

PR #2175, opened 2023-01-22
Insert inlay hint text edits on double instead of single click

predragnikolic:
This PR will fix the issue where inlay hints could be accidentally inserted in the view.

It introduces a `lsp_click` command, that accepts the flowing arugments:
- `command` - the command to execute
- `args` - the args of the command that will be executed
- `count` - the number of times the user needs to click it

The default time to wait for the click is 500ms - Yes I know, other systems might tweak this value, but I still think that 500ms is ok.


plugin/click.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class LspClick(sublime_plugin.TextCommand):
+    def __init__(self, view: sublime.View) -> None:
+        super().__init__(view)
+        self.click_count = 0
+
+    def run(self, edit: sublime.Edit, count: int, command: str, args: dict) -> None:
+        self.click_count += 1
+        if (self.click_count == count):
```
rchl, line 10: we don't need parenthesis in python

predragnikolic:
I experimented with other approaches, like sublime-mousemaps. 

That would be the best approach, it it worked. 
I can create a mousemap, and specify the `count` to 2, 
but then I would lose the default behavior when clicking a word 2 times. 

I discovered that mouse-maps also do not support context, but even if they did, I don't know how I would use it to my advantage...

Other options I have considered, 
I wanted to open an issue at the ST repo to ask for a new API. For the `event` to also introduce a `count` property, next to `x` and `y` values.

I also tried [this](https://github.com/sublimelsp/LSP/tree/run-inlay-hint-command)
but I was not happy with that implementation/behavior.

plugin/click.py
```diff
@@ -0,0 +1,19 @@
+import sublime
+import sublime_plugin
+
+
+class LspClick(sublime_plugin.TextCommand):
+    def __init__(self, view: sublime.View) -> None:
+        super().__init__(view)
+        self.click_count = 0
+
+    def run(self, edit: sublime.Edit, count: int, command: str, args: dict) -> None:
```
rchl, line 8: Does it really make sense to have the flexibility of specifying `count`? For single clicks we don't need to use it and more than double click seems kinda not that useful? If we don't need that then we could call it `LspDoubleClickCommand` or even `LspOnDoubleClickCommand` since otherwise it sounds like it performs a double click itself.

plugin/click.py
```diff
@@ -0,0 +1,19 @@
+import sublime
```
rchl, line 1: Maybe this code could live in `tooling.py`? Not sure such a tiny command requires its own file.

rchl:
I forgot which server has clickable inlay hints. Can you remind me?

Feels like this would create a bit of confusing UX because the point will show that the things is clickable but nothing will happen on a single click and there is no indication anyway that it's possible to double click it. Maybe the tooltip (if there is one) should include something like "double click to insert".

predragnikolic:
>  "double click to insert".

Good point, 

I took a look,
ST LSP implement showing the tooltip on the client,
and it looks like it is the server responsibility to provide the toolitp content.

```
export interface InlayHint {
        /**
	 * The tooltip text when you hover over this item.
	 *
	 * Depending on the client capability `inlayHint.resolveSupport` clients
	 * might resolve this property late using the resolve request.
	 */
	tooltip?: string | [MarkupContent](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#markupContentInnerDefinition);
```

That said, I would not put any string on the client like "double click to insert",
instead I would suggest opening a feat request to the server to do that.

rchl:
> That said, I would not put any string on the client like "double click to insert",
> instead I would suggest opening a feat request to the server to do that.

It's not the responsibly of the server to say that because server can't dictate how the UI should work. Whether it will insert text on single, double click or even using keyboard.

What I was suggesting was to either set our text or append it to the existing text (in parenthesis for example).
That said, I don't think it's a great suggestion since once user is familiar with how those work, it will just be an annoyance. But I really think that it can confusing if the mouse pointer shows "hand" but nothing happens on single clicking.

plugin/tooling.py
```diff
@@ -535,3 +535,18 @@ def on_transport_close(self, exit_code: int, exception: Optional[Exception]) ->
         self._transport = None
         output = str(exception) if exception else '\n'.join(self._stderr_lines).rstrip()
         sublime.set_timeout(lambda: self._on_close(self._resolved_command, output, exit_code))
+
+
+class LspDoubleClick(sublime_plugin.TextCommand):
```
rchl, line 540: Still think that `LspOnDoubleClickCommand` would be a more accurate name.

Also the `Command` suffix would be good to have as a convention.

predragnikolic:
You have a point,
Will push a commit when I grab the time

plugin/tooling.py
```diff
@@ -535,3 +535,18 @@ def on_transport_close(self, exit_code: int, exception: Optional[Exception]) ->
...
+
+    def run(self, edit: sublime.Edit, command: str, args: dict) -> None:
+        self.click_count += 1
+        if self.click_count == 2:
+            self.view.run_command(command, args)
+            self.click_count = 0
+            return
+
+        def reset() -> None:
+            self.click_count = 0
```
rchl, line 551: Why not just a class method (possibly with underscore prefix)? It's changing instance property so I don't think it's warranted to have inline function here.

plugin/inlay_hint.py
```diff
@@ -126,7 +127,10 @@ def format_inlay_hint_label(inlay_hint: InlayHint, session: Session, phantom_uui
                     'phantom_uuid': phantom_uuid
                 }
             })
-            result += '<a href="{command}">'.format(command=inlay_hint_click_command)
+            result += '<a href="{command}" title="Double Click {tooltip}">'.format(
```
rchl, line 130: Actually we are adding tooltip on the outer element already if you look at the code below. So this inner one won't show (or maybe it will override the other one)? I think you should consolidate that so that we do everything in the outer tooltip.

Also, do you think "Double Click {original tooltip}" is universal enough solution? You don't know what the original tooltip will say so I feel this can be a weird Frankenstein like "Double Click Parameter Name" or something weird otherwise.

What about "{original tooltip} [double click to insert]", for example?

predragnikolic: >  I think you should consolidate that so that we do everything in the outer tooltip.

I didn't see this comment, but I did the opposite. I think you should consolidate that so that we do everything in the outer tooltip.
I removed the outer tooltip and moved it inside. So currently the tooltip will not override anything.

I will continue experimenting tomorrow.
If you have any suggestions, feel free to push directly on this branch.

rchl: There are still two tooltips. One on the inner `a` and one on the outer `span`.
Also, is "Double click" clear enough? Maybe "Double-click to execute"?

jwortmann: Isn't the word "execute" a bit confusing? The `InlayHint.textEdits` to insert the text from the inlay hint into the document, not to execute arbritrary commands or edits.

```typescript
	/**
	 * Optional text edits that are performed when accepting this inlay hint.
	 *
	 * *Note* that edits are expected to change the document so that the inlay
	 * hint (or its nearest variant) is now part of the document and the inlay
	 * hint itself is now obsolete.
	 *
	 * Depending on the client capability `inlayHint.resolveSupport` clients
	 * might resolve this property late using the resolve request.
	 */
	textEdits?: TextEdit[];
```

VSCode has "Double click to insert".

plugin/inlay_hint.py
```diff
@@ -112,40 +110,51 @@ def format_inlay_hint_tooltip(tooltip: Optional[Union[str, MarkupContent]]) -> s
...
+                }
             })
             value += '<a href="{command}">'.format(command=inlay_hint_click_command)
         value += html.escape(label_part['value'])
         if is_clickable:
             value += "</a>"
         # InlayHintLabelPart.location is not supported
         result += "<span title=\"{tooltip}\">{value}</span>".format(
-            tooltip=format_inlay_hint_tooltip(label_part.get("tooltip")),
+            tooltip=tooltip + '(Double-click to execute)' if is_clickable else tooltip,
```
jwortmann, line 157: There is currently no space between the original tooltip and the addition. So it will look like `Original tooltip(Double-click to execute)`.

I'm also not confident that this addition is really needed. Instead we could add a sentence about it in the LSP docs, that you can insert inlay hints with double click (if supported by the server).

Or alternatively, use "Double click to insert" only if there is no original tooltip from the server.

plugin/tooling.py
```diff
@@ -535,3 +535,19 @@ def on_transport_close(self, exit_code: int, exception: Optional[Exception]) ->
...
+
+    def run(self, edit: sublime.Edit, command: str, args: Dict[Any, Any]) -> None:
+        self.click_count += 1
+        if self.click_count == 2:
+            self.view.run_command(command, args)
+            self.click_count = 0
+            return
+        sublime.set_timeout(self.reset, 500)
+
+    @classmethod
```
rchl, line 551: I didn't mean to use `@classmethod` if that's due to my comment. It should be a normal method of the instance since `run` changes instance property so this should too.

plugin/tooling.py
```diff
@@ -535,3 +535,19 @@ def on_transport_close(self, exit_code: int, exception: Optional[Exception]) ->
         self._transport = None
         output = str(exception) if exception else '\n'.join(self._stderr_lines).rstrip()
         sublime.set_timeout(lambda: self._on_close(self._resolved_command, output, exit_code))
+
+
+class LspDoubleClickCommand(sublime_plugin.TextCommand):
```
rchl, line 540: Are you on purpose not using `LspOnDoubleClickCommand` or you just haven't noticed that my suggestion has subtle difference (`*On*`)?

plugin/tooling.py
jwortmann: I would also make `click_count` to be an instance variable, because a TextCommand will be created for each view, and if it is a class variable, all instances will share this variable now.

So in theory, you can click on an inlay hint in a view, and within half a second click on another hint in a different view, and it will still be counted as a double-click.

To be very precise, this would also still happen if you use an instance variable for this, but when clicking on different inlay hints in the same view. If this should be avoided, this command could check whether the given args are the same between `click_count == 0` and `click_count == 1` (i.e. temporarily store the `args` when the first click happens).

rchl: > I would also make `click_count` to be an instance variable, because a TextCommand will be created for each view, and if it is a class variable, all instances will share this variable now.

In practice it should actually work fine because the class property will only be used for the initial value and if something sets self.click_count then it will also create new property on the instance that will shadow the class property.

That is as long as this `@classmethod` is removed.

plugin/inlay_hint.py
predragnikolic: [6e48d6a](https://github.com/sublimelsp/LSP/pull/2175/commits/6e48d6afab925497b72b2590d913d5bd648baf58)

plugin/tooling.py
predragnikolic: > To be very precise, this would also still happen if you use an instance variable for this, but when clicking on different inlay hints in the same view. If this should be avoided, this command could check whether the given args are the same between click_count == 0 and click_count == 1 (i.e. temporarily store the args when the first click happens).

Pushed some code that will prevent this.

predragnikolic: > I didn't mean to use @classmethod if that's due to my comment. It should be a normal method of the instance since run changes instance property so this should too.

Yes, that is how I (miss)understood it.

plugin/inlay_hint.py
rchl: > Or alternatively, use "Double click to insert" only if there is no original tooltip from the server.

I'm not sure about this distinction because server is not gonna say how it can be inserted so it can still be a case that is confusing.

But I guess I'll give up here as it's not that important.

rchl (approved):

rchl:
I've actually now realized that the inlay hint can include a command and that is not necessarily gonna insert anything. We should probably differentiate those cases and only include the text when there is a text edit present.

predragnikolic:
> inlay hint can include a command and that is not necessarily gonna insert anything

#### Regarding tooltip text for commands.

If we look at `inlayHints.label: InlayHintLabelPart[]`), we will see that a [InlayHintLabelPart](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#inlayHintLabelPart) may contain a `command`. 

This PR [will show the following text](https://github.com/sublimelsp/LSP/pull/2175/files#diff-a41cda9f185e4d7a0ec73e044c52b0cc825fe80d84ec8c11c4dd09d7a85dac06R157) in the title tooltip when iterating over InlayHintLabelParts:
- `InlayHintLabelPart.tooltip` if exist
- else "Double-click to execute" if `InlayHintLabelPart.command` exist
- else it will not display anything

#### Regarding tooltip text for text edits.

This PR [will show the following text](https://github.com/sublimelsp/LSP/pull/2175/files#diff-a41cda9f185e4d7a0ec73e044c52b0cc825fe80d84ec8c11c4dd09d7a85dac06R130) in the title tooltip if an inlay hint contains `textEdits`:
- `InlayHint.tooltip` if exist
- else "Double-click to insert" if `InlayHint.textEdits` exist
- else it will not display anything

>  We should probably differentiate those cases and only include the text when there is a text edit present.

I would say that we are handling those cases,
but maybe I'm missing something that you have concluded.

plugin/inlay_hint.py
```diff
@@ -112,40 +110,51 @@ def format_inlay_hint_tooltip(tooltip: Optional[Union[str, MarkupContent]]) -> s
...
+                'args': {
+                    'session_name': session.config.name,
+                    'inlay_hint': cast(dict, inlay_hint),
+                    'phantom_uuid': phantom_uuid
+                }
             })
             result += '<a href="{command}">'.format(command=inlay_hint_click_command)
-        result += html.escape(label)
+        result += '<span title="{tooltip}">{value}</span>'.format(
+            tooltip=tooltip or 'Double-click to insert' if is_clickable else "",
```
jwortmann, line 130: I'm not 100% sure, but I think this is not exactly right, because `or` has higher precedence than `if ... else` - https://docs.python.org/3/reference/expressions.html#operator-precedence

So this is currently the same as `(tooltip or 'Double-click to insert') if is_clickable else ""`, but we actually want `tooltip or ('Double-click to insert' if is_clickable else "")`

Because what we want (I think):
* if `is_clickable`, and server provides a tooltip, we show the tooltip
* if `is_clickable`, but server doesn't provide a tooltip, we show "Double-click to insert"
* if not clickable, but server provides a tooltip, we show the tooltip (perhaps a possible case for this would be something like label is `: bool` and tooltip is `Boolean value`)
* if not clickable, and server doesn't provide a tooltip, we show no tooltip at all

(but I have not tested with any language server, so please check whether what I wrote makes sense)

There is another similar place below, for the label parts.

You could also consider to not set the `title` attribute at all, if tooltip is `""` (empty). That would probably make the code a bit more complex, for example by building the result HTML string step by step. But feel free to ignore if `<span title="">` is fine.

plugin/inlay_hint.py
jwortmann: My idea was that if the server provides a tooltip, then the available space is already taken, because "Server's tooltip text (Double-click to insert)" looks a bit ugly and we don't know what the server uses as the tooltip text (maybe it could even result in more confusing confusing combinations with the server's tooltip text). And with the mouse cursor being the hand/finger icon, users should still easily find out that you can double-click the inlay hint (if supported by the server), so this tooltip hint probably doesn't provide a lot of value.

plugin/inlay_hint.py
predragnikolic: > So this is currently the same as (tooltip or 'Double-click to insert') if is_clickable else "", but we actually want tooltip or ('Double-click to insert' if is_clickable else "")

You are right.
will push a fix

plugin/inlay_hint.py
rchl: Honestly I don't agree with the statement that "with the mouse cursor being the hand/finger icon, users should still easily find out that you can double-click the inlay hint". At least when talking about myself, I'm not accustomed to having to double click when cursor uses a hand icon. I can not think of any other app/place where that would be the case. Certainly not in browsers with which most people are familiar with.

I have just tried this change with mocked, clickable payload just a while ago and even confused me since I've clicked and I didn't see any reaction so I've started looking through the document to check if anything has changed. For people that didn't see this PR or are not that familiar what inline hints are supposed to do on clicking it will be even more confusing.

Maybe I can suggest something else - no extra text on the tooltip but instead show the message in the status field if the lsp_on_double_click timeouts on single click?

jwortmann: Hm, okay I could indeed not find any example of another application where you need to double click when the hand icon is shown. ST's minihtml doesn't support the CSS `cursor` property, though.

I searched a bit on GitHub how inlay hint tooltips are used, and VSCode seems to render a horizontal separator between the tooltip and the double click message:

tooltip

We cannot do that, because ST's tooltips don't support HTML markup, but it seems that you can force a linebreak by including `\n` in the "title" string:

phantom

So maybe we could use that, i.e. always show the "Double click to insert" message, but if the server also provides a tooltip, then use `server_tooltip + "\nDouble click to insert"`. Not sure if that would be a better alternative than just a space.

Also servers can use MarkupContent for the tooltip, for example see https://github.com/sjsone/vscode-neos-fusion-lsp/blob/80ea71b2d23c05abbcf47c51711888669b915be4/server/src/languageFeatures/InlayHintLanguageFeature.ts#L43-L51 where it's used for a fenced code block. That would be rendered as literal characters in ST...

rchl: I would be fine with a newline but also I'm having second thoughts about using title for that. It's quite shouty in ST being all yellow and weird looking. It might seem a little too much for servers that set text edits. And user just needs to be educated about it and not necessarily shown on every occasion where the inlay hint is hovered. In that sense I think my suggestion about showing status text on timeout might make sense.

jwortmann: > being all yellow and weird looking

It's controlled by the theme, and has a white background in the Default, Default Dark and Adaptive themes. For the beautiful tooltips from my second screenshots, try out my "Brackets" theme :)

But I'm fine with trying out the status message on timeout. The implementation should be made generic in that case, i.e. I would pass another optional argument `timeout_message: str = ""` for the `lsp_on_double_click` command, so that it is only shown if this message is defined.

predragnikolic: For now I just pushed a commit to always append the instruction text if a inlay hint contains a command or text edit.
That way the user will know if the inlay hint will insert text, or if a command will be executed as part of double clicking phantoms.

I also tough to completely remove the instruction text from the `title` and just put that info in the release notes and tell users that they have to double click phantoms. (note only one server supports text edits for inlay hints, and not many people use it, don't know any server that supports commands... so I would not put to much effort and thoughts into this).

plugin/inlay_hint.py
```diff
@@ -112,40 +110,54 @@ def format_inlay_hint_tooltip(tooltip: Optional[Union[str, MarkupContent]]) -> s
...
+                    'session_name': session.config.name,
+                    'inlay_hint': cast(dict, inlay_hint),
+                    'phantom_uuid': phantom_uuid
+                }
             })
             result += '<a href="{command}">'.format(command=inlay_hint_click_command)
-        result += html.escape(label)
+        instruction_text = '\nDouble-click to insert' if has_text_edits else ""
+        result += '<span title="{tooltip}">{value}</span>'.format(
+            tooltip=tooltip + instruction_text,
```
rchl, line 132: Won't we have an empty line at the beginning in case the `tooltip` is not provided?

rchl: It seems like ST might be stripping leading newline internally but I have still added an explicit strip just in case.

rchl (approved):

PR #2179, opened 2023-01-27
disable diagnostics panel

glaucocustodio:
https://discord.com/channels/280102180189634562/645268178397560865/1068560891806752798

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -38,7 +38,7 @@
   // warning: 2
   // info: 3
   // hint: 4
-  "show_diagnostics_panel_on_save": 2,
+  "show_diagnostics_panel_on_save": 0,
 
   // Show errors and warnings count in the status bar
   "show_diagnostics_count_in_view_status": false,
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -255,7 +255,7 @@ class Settings:
         r("show_diagnostics_count_in_view_status", False)
         r("show_diagnostics_in_view_status", True)
         r("show_multiline_diagnostics_highlights", True)
-        r("show_diagnostics_panel_on_save", 2)
+        r("show_diagnostics_panel_on_save", 0)
         r("show_diagnostics_severity_level", 2)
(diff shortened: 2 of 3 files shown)


rchl:
There are also those places that have a `default` of 2:

https://github.com/sublimelsp/LSP/blob/fbbfa5a41afc42d3d4e171aa8408e434f80d1a6e/sublime-package.json#L423-L423

https://github.com/sublimelsp/LSP/blob/bf5a1f21fc0175cbf4a5231ee72e177013b156e6/plugin/core/types.py#L258-L258

BTW. I've suggest this change so obviously I'm fine with it. :)
I believe others have also suggested that before (@jwortmann ?).

jwortmann:
> I believe others have also suggested that before (@jwortmann ?).

Yes, I always thought that this default value is annoying :D
Another related setting, where I would change the default value is `"diagnostics_panel_include_severity_level"` (I would use 3 instead of 4, because iirc\* diagnostics with the "hint" severity are not supposed to be shown in the UI - we also draw the stippled underline in the view though)

\* I think that I've read this somewhere in an issue from the LSP specs repository, but I don't know the exact place right now.

glaucocustodio:
just did the changes

jwortmann:
Please note that changing `"diagnostics_panel_include_severity_level"` was just a suggestion from my side and perhaps needs a separate discussion first.

I already removed the gutter icon for the "hint" severity in a previous PR #2086 (this also contains a link to the related discussion/clarification about the idea behind the "hint" severity rendering).

In VSCode, diagnostics with this severity are not shown in the "Problems" panel, by the way.

glaucocustodio:
ok, just reverted diagnostics_panel_include_severity_level changes

rchl:
Yeah, probably best to have it as a separate PR if we'd want that (to be honest I wouldn't mind but would be good to hear opinions of other maintainers).

rchl (approved):

predragnikolic (approved):

PR #2180, opened 2023-01-27
Implement type hierarchy request

jwortmann:
The type hierarchy specs/interface is very similar to call hierarchy, so `type_hierarchy.py` is basically a copy/paste of `call_hierarchy.py`. I'm unsure whether there should be added another abstraction layer, with `LspHierarchyDataProvider(TreeDataProvider)`, `LspHierarchyCommand`, and `LspHierarchyToggleCommand`, from which call hierarchy & type hierarchy can derive, but then you need to pass a lot of strings and the protocol types in `__init__`. And there is a bit of difference between the call and type hierarchy responses, in that type hierarchy always returns `TypeHierarchyItem[]` directly, but call hierarchy uses wrapper classes `CallHierarchyOutgoingCall[]` / `CallHierarchyIncomingCall[]` with `from`/`to` fields. That could make a generic implementation a bit awkward.
(I assume another abstraction layer would be the preferred choice, but I just wanted to wait for opinions about this first)

I also moved the generic utility functions for filepaths from `goto_diagnostic.py` into its own file `core/paths.py`.

A server which supports type hierarchy is clangd.

### DIFF
--- a/plugin/core/paths.py
+++ b/plugin/core/paths.py
@@ -1,0 +1,69 @@
+from .protocol import DocumentUri
+from .sessions import Session
+from .typing import Iterable, Optional, Tuple
+from .views import parse_uri
+from pathlib import Path
+
+
+def simple_path(session: Optional[Session], uri: DocumentUri) -> str:
+    scheme, path = parse_uri(uri)
+    if not session or scheme != 'file':
+        return path
+    simple_path = simple_project_path([Path(folder.path) for folder in session.get_workspace_folders()], Path(path))
+    return str(simple_path) if simple_path else path
+
+
+def project_path(project_folders: Iterable[Path], file_path: Path) -> Optional[Path]:
+    """
+    The project path of `/path/to/project/file` in the project `/path/to/project` is `file`.
+    """
+    folder_path = split_project_path(project_folders, file_path)
+    if folder_path is None:
+        return None
+    _, file = folder_path
+    return file
+
+
+def simple_project_path(project_folders: Iterable[Path], file_path: Path) -> Optional[Path]:
+    """
+    The simple project path of `/path/to/project/file` in the project `/path/to/project` is `project/file`.
+    """
+    folder_path = split_project_path(project_folders, file_path)
+    if folder_path is None:
+        return None
+    folder, file = folder_path
+    return folder.name / file
+
+
+def resolve_simple_project_path(project_folders: Iterable[Path], file_path: Path) -> Optional[Path]:
+    """
+    The inverse of `simple_project_path()`.
+    """
+    parts = file_path.parts
+    folder_name = parts[0]
+    for folder in project_folders:
+        if folder.name == folder_name:
+            return folder / Path(*parts[1:])
+    return None
+
+
+def project_base_dir(project_folders: Iterable[Path], file_path: Path) -> Optional[Path]:
+    """
+    The project base dir of `/path/to/project/file` in the project `/path/to/project` is `/path/to`.
+    """
+    folder_path = split_project_path(project_folders, file_path)
+    if folder_path is None:
+        return None
+    folder, _ = folder_path
+    return folder.parent
+
+
+def split_project_path(project_folders: Iterable[Path], file_path: Path) -> Optional[Tuple[Path, Path]]:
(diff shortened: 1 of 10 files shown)


rchl:
> A server which supports type hierarchy is clangd.

Does the feature needs to be explicitly enabled?

When I'm trying it I'm getting:

```
:: --> clangd textDocument/prepareTypeHierarchy(8): {'textDocument': {'uri': 'file:///usr/local/workspace/sublime-packages/LSP/plugin/test.cpp'}, 'position': {'line': 6, 'character': 9}, 'workDoneToken': '$ublime-8'}
:: <~~ clangd 8: {'message': 'method not found', 'code': -32601}
```

Maybe I'm triggering it in a wrong place?

jwortmann:
It works for me with LSP-clangd. Do you perhaps have an older server version installed? Note that LSP-clangd by default uses any server version that is found in the PATH (according to the description in the settings).

rchl:
Mac ships with clangd 14 but I've also retested using latest one installed from homebrew and had same results. Maybe I'll retest using some full project later since that was just a single c++ file within an unrelated project.

jwortmann:
So what is your clangd server version? Mine is `15.0.1`.

Works even in single file mode on my side:


[type_hierarchy.webm](https://user-images.githubusercontent.com/6579999/215344447-0176cdfa-ce13-4398-967e-745a18f8ce8c.webm)

---

Another question would be whether we should show subtypes or supertypes initially. Currently it's supertypes, and I could imagine that this is more useful, but it would probably look more natural for such a tree to show the subtypes.

rchl:
It works now. Didn't realize that LSP-clangd handles `command` on its own and it's not enough to override it.

> Another question would be whether we should show subtypes or supertypes initially. Currently it's supertypes, and I could imagine that this is more useful, but it would probably look more natural for such a tree to show the subtypes.

Don't have opinion on that. Seems I won't use that feature anyway in my languages of choice so we can re-consider if someone who has stronger opinion complains.

plugin/core/paths.py
```diff
@@ -0,0 +1,69 @@
+from .protocol import DocumentUri
+from .sessions import Session
+from .typing import Iterable, Optional, Tuple
+from .views import parse_uri
+from pathlib import Path
+
+
+def simple_path(session: Optional[Session], uri: DocumentUri) -> str:
```
rchl, line 8: Not quite sure "simple" is very meaningful but it's not your fault.

plugin/type_hierarchy.py
```diff
@@ -0,0 +1,166 @@
...
+            return Promise.resolve(self.root_elements)
+        session = self.weaksession()
+        if not session:
+            return Promise.resolve([])
+        if self.direction == TypeHierarchyDirection.Supertypes:
+            params = cast(TypeHierarchySupertypesParams, {'item': element})
+            return session.send_request_task(Request.supertypes(params)).then(self._ensure_list)
+        elif self.direction == TypeHierarchyDirection.Subtypes:
+            params = cast(TypeHierarchySubtypesParams, {'item': element})
+            return session.send_request_task(Request.subtypes(params)).then(self._ensure_list)
```
rchl, line 55: Less code:

```suggestion
        if self.direction == TypeHierarchyDirection.Supertypes:
            return session.send_request_task(Request.supertypes({'item': element})).then(self._ensure_list)
        if self.direction == TypeHierarchyDirection.Subtypes:
            return session.send_request_task(Request.subtypes({'item': element})).then(self._ensure_list)
```

Note that I also prefer "no elif after return"

rchl (approved):

jwortmann:
@rchl How did you manage to trigger the `textDocument/prepareTypeHierarchy` request in you first attempt https://github.com/sublimelsp/LSP/pull/2180#issuecomment-1407502182?

If the server doesn't support it, both the context menu and command palette items should be hidden, and it also shouldn't be possible to invoke it via key binding. I wonder whether we have a bug somewhere? (But I don't see any wrong behavior in that regard on my side)

rchl:
It seems like version 14.0.0 of clangd shipped with macOS is lying about supporting it while it doesn't really.

PR #2182, opened 2023-01-28
fix stale state or lack of updates on changing branches

rchl:
When file has been "reloaded" by ST due to branch change or just modified externally then the requests that normally happen on file changes didn't run. That includes code lenses, symbol highlights, color boxes, semantic tokens, document link and inlay hints. That result in lack of those features until file was modified or stale state like with those code lenses for example:

Screenshot 2023-01-28 at 22 44 56

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -866,18 +866,6 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
                 sv.on_reload_async()
         self._on_view_updated_async()
 
-    def _on_view_updated_async(self) -> None:
-        different, current_region = self._update_stored_region_async()
-        self._code_lenses_debouncer_async.debounce(
-            self._do_code_lenses_async, timeout_ms=self.code_lenses_debounce_time)
-        if not different:
-            return
-        self._clear_highlight_regions()
-        if userprefs().document_highlight_style:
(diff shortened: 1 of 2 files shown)


plugin/documents.py
```diff
@@ -908,6 +901,18 @@ def _register_async(self) -> None:
                     debug("also registering", listener)
                     listener.on_load_async()
 
+    def _on_view_updated_async(self) -> None:
+        self._code_lenses_debouncer_async.debounce(
+            self._do_code_lenses_async, timeout_ms=self.code_lenses_debounce_time)
+        different, current_region = self._update_stored_region_async()
+        if not different:
+            return
```
rchl, line 909: As has been mentioned in some other issue, it's not really correct to return here if region didn't change as the file might have still changed (especially on file "reload"). The impact of this is very small currently though (pretty much only symbol highlights could misbehave).

predragnikolic (approved):

rchl:
> IMO, It would be best to replace the `_update_stored_region_async` (because it doesn't function correctly)
> with a different way to check if the document have changed. See [sublimelsp/LSP-typescript@8e0d468](https://github.com/sublimelsp/LSP-typescript/commit/8e0d46890fd5b78824703125acbfa51ff074ce41)

Currently only symbol highlight and signature help relies on this feature so it's not much of an issue. Other features don't go through that code path.

PR #2184, opened 2023-01-31
Refactor call & type hierarchy to derive from a common command

jwortmann:
Reduce code duplication by adding an abstraction layer for the call hierarchy & type hierarchy commands.

~~It was not my preferred choice to put everything in `core/registry.py`, but after failed attempts with cyclic import errors I gave up trying to put it in `core/tree_view.py` (or elsewhere).~~

*Edit:* Moved to its own file `hierarchy.py`.

There is also one small fix to prevent an exception message in the console, in case of error responses (clangd apparently doesn't support `callHierarchy/outgoingCalls`), by using `isinstance(response, list)`, but in general no functionality change in this PR.


plugin/hierarchy.py
```diff
@@ -0,0 +1,211 @@
...
+) -> HierarchyDataProvider:
+    if sheet_name == "Call Hierarchy":
+        request1 = Request.incomingCalls
+        request2 = Request.outgoingCalls
+        handler1 = lambda response: [incoming_call['from'] for incoming_call in response] if isinstance(response, list) else []  # noqa: E501,E731
+        handler2 = lambda response: [outgoing_call['to'] for outgoing_call in response] if isinstance(response, list) else []  # noqa: E501,E731
+    elif sheet_name == "Type Hierarchy":
+        request1 = Request.supertypes
+        request2 = Request.subtypes
+        handler1 = handler2 = lambda response: response if isinstance(response, list) else []  # noqa: E731
```
rchl, line 96: Isn't it an instance of our `Error` in case of clangd? It would be more explicit to check that if that's the case.

jwortmann: Yes, but it can also be `None` if it's no error. So this `isinstance(..., list)` catches all cases at once. That's basically why I used it here, so that it fits into a single line of code (and if I see it correctly, because the `send_request_task` doesn't have an anology to the `send_request_async`'s `on_error` callback).

jwortmann: Should I create separate, named functions for the response handlers, instead of lambdas, to make it more explicit? (flake8 already complained :D)

rchl: Yes, separate named functions would be better IMO.

Also those request1, request2, etc. names don't speak to me at all so I would perhaps suggest refactoring this way: 5c7fa1818a62be77c3e860db0d50fd04773fc670 (not sure if I got everything right but something like that).

jwortmann: Yeah, makes more sense to do it that way. When you move things around and focus on initial ideas, you sometimes don't see the obvious.

rchl (approved):

PR #2185, opened 2023-01-31
hierarchy: indicate that there are no child items by hiding expander

rchl:
I think that hiding expander when there are no child items is more natural thing to do and makes it clear that there are no items rather than giving impression that it's still loading or broken.


jwortmann (changes requested):
Iirc this idea was already brought up in the initial PR, and I think I've also experimented with it a bit (that's probably the reason why I had already implemented the `else` case in `TreeItem.html`).

A disadvantage would be that the expander icon suddenly "vanishes" when you click on it, which might look a bit weird/surprising too. But I don't have a strong opinion here, so from my side it would be okay to merge it.

Another idea could be to "prefetch" the next level of child items, i.e. then it would be possible to already hide the icon from the beginning, when an item is initially show in "collapsed" state. And when you expand an item, the children are already stored in the "cache", and only their expander icons would appear when the responses for the next level arrive (it could use the default "collapsed" item initally, until they are updated). But maybe that's a bit excessive and too complicated for this feature. And there would probably be a lot of unnecessary requests. Maybe I'll experiment with that idea and try out if it works.

jwortmann (approved):

PR #2187, opened 2023-02-06
Fix hidden code action menu entries sometimes being visible

jwortmann:
This happend when the main menu is opened and no view or input field has currently focus (for example when you focused the side bar or an output panel). Because there exists no instance of a corresponding active TextCommand in that case, and therefore also no `is_visible` method to check for the visibility state - see https://github.com/sublimehq/sublime_text/issues/5857.

The solution is to use a WindowCommand instead, because there is always an instance of that for the active window.

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -6,6 +6,7 @@ from .core.protocol import Diagnostic
 from .core.protocol import Error
 from .core.protocol import Request
 from .core.registry import LspTextCommand
+from .core.registry import LspWindowCommand
 from .core.registry import windows
 from .core.sessions import AbstractViewListener
 from .core.sessions import Session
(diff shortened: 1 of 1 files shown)


rchl:
I imagine ST could allow making menu item default to "not visible" which would fix those issues without any hacks. But this might never happen, of course.

rchl (approved):

jwortmann:
> On Mac it's possible to have no windows open so then it still reproduces. But I guess it's still better with this change.

Ah, okay then it would probably be possible to solve that by using an `ApplicationCommand` instead. But we have no `LspApplicationCommand` currently, and we would need that for (or reimplement) `self.session_by_name`, which is used in `run_async`. Not sure if it's worth it to implement that only for this menu UI detail, so I guess the WindowCommand is probably good enough for now.

PR #2188, opened 2023-02-06
fix log panel not scrolling on initial open

rchl:
There was annoying issue with LSP Log Panel that meant that the content didn't scroll when opened initially.

I've experimented a bit with a simple plugin and noticed this this seems like a bug that triggers when:
 - panel view's `scroll_past_end` setting is set to `False`
 - and panel is shown with more than a screenfull of content inserted at the same time

The work around is to just trigger `view.show(view.sel(), animate=False)` after inserting text to scroll the view to the end (or wherever selection was set too if it's not the initial opening).

### DIFF
--- a/boot.py
+++ b/boot.py
@@ -191,4 +191,4 @@ class Listener(sublime_plugin.EventListener):
             if panel_manager.is_panel_open(PanelName.Diagnostics):
                 sublime.set_timeout_async(wm.update_diagnostics_panel_async)
             elif panel_manager.is_panel_open(PanelName.Log):
-                sublime.set_timeout(panel_manager.update_log_panel)
+                sublime.set_timeout(lambda: panel_manager.update_log_panel(scroll_to_selection=True))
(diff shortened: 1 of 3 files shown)


rchl:
Used this plugin to test, reproduce and workaround the issue:

```py
import sublime_plugin

PANEL_NAME = 'test-panel'
OUTPUT_PANEL_SETTINGS = {
    'scroll_past_end': False,
}


class CreateTestPanelCommand(sublime_plugin.WindowCommand):
    def run(self) -> None:
        panel = self.window.find_output_panel(PANEL_NAME)
        if not panel:
            panel = self.window.create_output_panel(PANEL_NAME)
            settings = panel.settings()
            for key, value in OUTPUT_PANEL_SETTINGS.items():
                settings.set(key, value)
        self.window.run_command('show_panel', {'panel': 'output.{}'.format(PANEL_NAME)})
        panel.set_read_only(False)
        panel.run_command('insert', {'characters': 'abc\n' * 100})
        # panel.show(panel.sel())
        panel.set_read_only(True)


class DestroyTestPanelCommand(sublime_plugin.WindowCommand):
    def run(self) -> None:
        self.window.destroy_output_panel(PANEL_NAME)
```

rchl:
Will get it in as I've been battle testing it for a while.

PR #2192, opened 2023-02-09
refactor: move format_completion into completion.py

rchl:
I didn't want to muddle previous PR with the move so doing it separately. I think there is too much stuff in `views.py` and this code really belongs in `completion.py` anyway.

There are no changes besides the move.


predragnikolic:
Do imports in tests need to be updated?

predragnikolic (approved):

PR #2193, opened 2023-02-09
improve heuristic for picking what to show as completion trigger

### LINKED ISSUE #2189: Improve heuristic related to showing completion labels
**Describe the bug**

`typescript-language-server` will soon introduce `filterText` property which will make ST presentation of completions worse. See https://github.com/typescript-language-server/typescript-language-server/pull/678#pullrequestreview-1284240548

The change will be spec compliant and might improve things for other editors but it's not gonna work great for us so we need to improve our heuristic.

This issue also already manifests with Volar, for example. See:

Screenshot 2023-02-07 at 13 57 11

rchl:
See the issue for more info.

I'm not saying that this is the best or even correct solution but it's to kick-start the discussion.

Fixes #2189

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -56,10 +56,11 @@ def format_completion(
             view_id, index, session_name)
         href = 'subl:lsp_run_text_command_helper {}'.format(args)
         details.append(make_link(href, 'More'))
-    if lsp_label_detail and (lsp_label + lsp_label_detail).startswith(lsp_filter_text):
+    ignore_filter_text = ' ' in lsp_filter_text
+    if lsp_label_detail and (ignore_filter_text or (lsp_label + lsp_label_detail).startswith(lsp_filter_text)):
         trigger = lsp_label + lsp_label_detail
         annotation = lsp_label_description or lsp_detail
-    elif lsp_label.startswith(lsp_filter_text) or ' ' in lsp_filter_text:
+    elif ignore_filter_text or lsp_label.startswith(lsp_filter_text):
         trigger = lsp_label
         annotation = lsp_detail
         if lsp_label_detail:
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -59,7 +59,7 @@ def format_completion(
     if lsp_label_detail and (lsp_label + lsp_label_detail).startswith(lsp_filter_text):
         trigger = lsp_label + lsp_label_detail
         annotation = lsp_label_description or lsp_detail
-    elif lsp_label.startswith(lsp_filter_text):
+    elif lsp_label.startswith(lsp_filter_text) or ' ' in lsp_filter_text:
         trigger = lsp_label
         annotation = lsp_detail
         if lsp_label_detail:
--- a/stubs/sublime.pyi
+++ b/stubs/sublime.pyi
@@ -297,10 +297,13 @@ class Syntax:
 
 
 class CompletionItem:
-    flags = ...  # type: int
-    details = ...  # type: str
+    trigger = ...  # type: str
     annotation = ...  # type: str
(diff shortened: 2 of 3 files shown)


predragnikolic:
```diff
+   trigger = lsp_label
+   annotation = lsp_detail
    if lsp_label_detail:
        trigger = lsp_label + lsp_label_detail
        annotation = lsp_label_description or lsp_detail
-    elif lsp_label.startswith(lsp_filter_text):
-        trigger = lsp_label
-        annotation = lsp_detail
-        if lsp_label_detail:
-            details.append(html.escape(lsp_label + lsp_label_detail))
-        if lsp_label_description:
-            details.append(html.escape(lsp_label_description))
-    else:
-        trigger = lsp_filter_text
-        annotation = lsp_detail
-        details.append(html.escape(lsp_label + lsp_label_detail))
-        if lsp_label_description:
-            details.append(html.escape(lsp_label_description))
```

The spec is clear for the label(trigger) part, 
we should show the `CompletionItem.label` or the `CompletionItem.label` + `CompletionItem.labelDetails` if `labelDetails` is present in the autocomplte popup.

Id doesn't feel right to show `filterText` as the `label` in the autocomplete popup(because that value is not meant that, and I am not sure why it this line  `trigger = lsp_filter_text` was added... there was probably a reason but I forgot it).

My ideal case would be to remove the elif and else branches completely if we can.

rchl:
The spec is written with assumption that the client can show one thing (`label`) and filter on another thing (`filterText`). ST can only filter on what it is visible so that's why we need those heuristics. Those heuristics were introduced in #2010 so you can check it for more info.

rchl:
Added also some tests so that it's easier to reason about any changes.

rchl:
@predragnikolic With your suggested change there is one silly test that fails:

```py
    def test_filter_text_4(self) -> None:
        # See the `test_filter_text_is_not_a_prefix_of_label` test above and
        # also https://github.com/sublimelsp/LSP/issues/771
        # This is probably a silly server behavior that we probably shouldn't need to support?
        self._verify_completion(
            {
                "filterText": "e",
            },
            trigger='e',
        )
```

But honestly it is quite weird server behavior and it might even have been addressed in the server already (maybe).

No other tests that I've added seem to fail which is quite surprising to me and means that either we are missing some relevant tests or it could actually work.

predragnikolic:
Currently reading this https://github.com/sublimelsp/LSP/pull/990  🙂

I will study all the changes and hopefully I will have a conclusion tomorrow.

predragnikolic:
filterText was used in the era when lsp_label_details didn't exist, 
to better format things with in the autocomplete popup.

currently, when we have dedicated fields in the spec, label detail and description, the workaround with the filterText is no longer necessary. It made sense then, but not now.

The more I read it the more I believe that we will be able to delete the elif, else branches 🙂

Also if you have time, do a quick look of what other servers return for the filterText, some return just a char, which IMO would be pretty bad to show in the AC popup (but that case didn't happen with servers that I use, and nobody reported such issues)... 
But the reason why we have this PR in the first place is because we have a bug because we are using filterText as a label. 

If you(or somebody) still have some case why filterText should be used, please say. Otherwise I am 100% for removing it.

jwortmann:
I think this change doesn't make sense, for the reasons explained in https://github.com/sublimelsp/LSP/issues/2189#issuecomment-1426199595. We should use `filterText` for the trigger (unless no information is lost if using the `label` instead). And we have to live with the downside that it doesn't look very nice in some special occasions, because ST just doesn't have the ability to match against something different to what is shown in the trigger.

rchl:
Closing based on discussions here and in the issue.

I will till get the tests in separately.

PR #2194, opened 2023-02-11
Add support for CompletionList.itemDefaults

### LINKED ISSUE #2043: Support CompletionList.itemDefaults
See:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionList

predragnikolic:
closes #2043 

This PR add support for item defaults, for the following fields:
- data (it was easy to add this)
- insertTextFormat (it was easy to add this)
- editRange (it needed [some additional logic](https://github.com/sublimelsp/LSP/compare/feat/completionList-itemDefaults?expand=1#diff-1e31f37d39e534dca2dca7fd81d29d7e5a039a4b55a5765fdb4292c7a87760aaR124-R141))

The additional logic transforms `editRange` + `textEditText` (if present) to a TextEdit. So all of the logic of how we dealed with TextEdits, stayed the same.

This can be tested with this [PR](https://github.com/sublimelsp/LSP-tailwindcss/pull/54) of LSP-tailwindcss. That PR is not merged yet, but I do plan to merge it soon.


plugin/core/protocol.py
```diff
@@ -6133,3 +6133,5 @@ def to_lsp(self) -> 'Position':
 
 # Temporary for backward compatibility with LSP packages.
 RangeLsp = Range
+CompletionItemDefaults = __ItemDefaults_Type_15
+CompletionEditRange = __EditRange_Type_15
```
predragnikolic, line 6137: this is not great, but I need those types

I avoided importing `__ItemDefaults_Type_15` directly, instead I declared `CompletionItemDefaults`, so if the generated type changes I just need to update this place... but It would be good to not deal with this at all.

Any suggestions? 

I will try to find a `Pick` typescript alternative in mypy.

predragnikolic:
For the spec, see: 
CompletionClientCapabilities.completionList.itemDefaults  https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionClientCapabilities

CompletionList.itemDefaults https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionList

CompletionItem.textEditText https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem

plugin/core/protocol.py
rchl: It would need to be fixed in the types generator but how exactly I'm not sure. First thing that comes to mind is that it's name (or suffix) should be based on **where it's used** (in which parent object) since then the name shouldn't change based on other types being added earlier.

plugin/completion.py
```diff
@@ -35,10 +37,15 @@
...
+CompletionsStore = TypedDict('CompletionsStore', {
+    'items': List[CompletionItem],
+    'itemDefaults': CompletionItemDefaults
+})
 
 
 def format_completion(
-    item: CompletionItem, index: int, can_resolve_completion_items: bool, session_name: str, view_id: int
+    item: CompletionItem, index: int, can_resolve_completion_items: bool,
+    session_name: str, item_defaults: CompletionItemDefaults, view_id: int
```
rchl, line 48: if it doesn't fit one line then I'd rather go for one argument per line

plugin/completion.py
```diff
@@ -105,6 +112,44 @@ def get_text_edit_range(text_edit: Union[TextEdit, InsertReplaceEdit]) -> Range:
...
+        elif is_range(edit_range):
+            default_text_edit = {
+                'newText': new_text,
+                'range': edit_range
+            }
+    if default_text_edit:
+        item['textEdit'] = item.get('textEdit', default_text_edit)
+    default_insert_text_format = item_defaults.get('insertTextFormat')
+    if default_insert_text_format:
+        item['insertTextFormat'] = item.get("insertTextFormat", default_insert_text_format)
```
rchl, line 146: Minor but would prefer this to avoid unnecessary assignments.

```suggestion
    if default_text_edit and 'textEdit' not in item:
        item['textEdit'] = default_text_edit
    default_insert_text_format = item_defaults.get('insertTextFormat')
    if default_insert_text_format and 'insertTextFormat' not in item:
        item['insertTextFormat'] = default_insert_text_format
    default_data = item_defaults.get('data')
    if default_data and 'data' not in item:
        item['data'] = default_data
```

plugin/completion.py
```diff
@@ -35,10 +37,15 @@
 SessionName = str
 CompletionResponse = Union[List[CompletionItem], CompletionList, None]
 ResolvedCompletions = Tuple[Union[CompletionResponse, Error], 'weakref.ref[Session]']
+CompletionsStore = TypedDict('CompletionsStore', {
```
predragnikolic, line 40: I don't like this name,
if you have something else to suggest, I am for changing it.

It is only used in this file, so I am not that bothered by it.

plugin/core/protocol.py
predragnikolic: I tired doing that before, 
but run into some issue like the generated underscore types could have clashing names.
But I will try again.

plugin/core/sessions.py
```diff
@@ -265,12 +265,15 @@ def get_initialize_params(variables: Dict[str, str], workspace_folders: List[Wor
...
-                "labelDetailsSupport": True
+                "labelDetailsSupport": True,
             },
             "completionItemKind": {
                 "valueSet": completion_kinds
             },
-            "insertTextMode": InsertTextMode.AdjustIndentation
+            "insertTextMode": InsertTextMode.AdjustIndentation,
+            "completionList": {
+                "itemDefaults": ["editRange", "insertTextFormat", "data"]
```
rchl, line 275: We don't support `commitCharacters` per individual completion but if we say that we do then that will potentially result in smaller payload right?

plugin/core/protocol.py
predragnikolic: Let me just know if that is a blocker for this PR, or something that could be done later?

tests/test_completion.py
```diff
@@ -797,3 +799,106 @@ def test_additional_edits_if_session_does_not_have_the_resolve_capability(self)
             completion_items=[completion_item],
             insert_text='',
             expected_text='import ghjk;\nghjk')
+
+
+class ItemDefaultTests(CompletionsTestsBase):
```
rchl, line 804: Inherit from `unittest.TestCase` since we don't need to set up LSP on a file for those tests.

plugin/completion.py
```diff
@@ -172,16 +218,21 @@ def _resolve_completions_async(self, responses: List[ResolvedCompletions]) -> No
...
+            LspSelectCompletionCommand.completions[session.config.name] = {
+                'items': response_items,
+                'itemDefaults': item_defaults
+            }
             can_resolve_completion_items = session.has_capability('completionProvider.resolveProvider')
             config_name = session.config.name
             items.extend(
-                format_completion(response_item, index, can_resolve_completion_items, config_name, self._view.id())
+                format_completion(response_item, index, can_resolve_completion_items,
+                                  config_name, item_defaults, self._view.id())
```
rchl, line 235: prefer this formatting:
```suggestion
                format_completion(
                    response_item, index, can_resolve_completion_items, config_name, item_defaults, self._view.id())
```

plugin/completion.py
```diff
@@ -35,10 +37,15 @@
 SessionName = str
 CompletionResponse = Union[List[CompletionItem], CompletionList, None]
 ResolvedCompletions = Tuple[Union[CompletionResponse, Error], 'weakref.ref[Session]']
+CompletionsStore = TypedDict('CompletionsStore', {
+    'items': List[CompletionItem],
+    'itemDefaults': CompletionItemDefaults
+})
```
rchl, line 43: Can you use a tuple here? Would simplify code a bit:

```suggestion
CompletionsStore = Tuple[List[CompletionItem], CompletionItemDefaults]
```

plugin/completion.py
```diff
@@ -278,10 +331,12 @@ def run(self, edit: sublime.Edit, event: Optional[dict] = None) -> None:
...
 class LspSelectCompletionCommand(LspTextCommand):
 
-    completions = {}  # type: Dict[SessionName, List[CompletionItem]]
+    completions = {}  # type: Dict[SessionName, CompletionsStore]
 
     def run(self, edit: sublime.Edit, index: int, session_name: str) -> None:
-        item = LspSelectCompletionCommand.completions[session_name][index]
+        item = LspSelectCompletionCommand.completions[session_name]['items'][index]
+        item_defaults = LspSelectCompletionCommand.completions[session_name].get('itemDefaults', {})
+        item = completion_with_defaults(item, item_defaults)
```
rchl, line 339: If you change `CompletionsStore` to a tuple then this can be written in a cleaner way:

```suggestion
        items, item_defaults = LspSelectCompletionCommand.completions[session_name]
        item = completion_with_defaults(items[index], item_defaults)
```

plugin/core/sessions.py
predragnikolic: The LSP client for ST didn't announce `commitCharactersSupport` in it's capabilities, 
so I would expect the server to not send the `commitCharacters`.

rchl: Right. I did not recall that there is such capability. That's good news because it means I can drop those from typescript-language-server if the capability is missing.

plugin/completion.py
rchl: It's fine I think :)

plugin/core/protocol.py
rchl: Shouldn't be a blocker since it won't change on us by itself.

predragnikolic: I will work on it now.

plugin/completion.py
```diff
@@ -105,6 +113,47 @@ def get_text_edit_range(text_edit: Union[TextEdit, InsertReplaceEdit]) -> Range:
...
+        item['textEdit'] = default_text_edit
+    default_insert_text_format = item_defaults.get('insertTextFormat')
+    if default_insert_text_format and 'insertTextFormat' not in item:
+        item['insertTextFormat'] = default_insert_text_format
+    default_data = item_defaults.get('data')
+    if default_data and 'data' not in item:
+        item['data'] = default_data
+    default_data = item_defaults.get('data')
+    if default_data:
+        item['data'] = item.get('data', default_data)
```
rchl, line 153: Something went wrong here with suggestions.

```suggestion
    if default_data and 'data' not in item:
        item['data'] = default_data
```

plugin/completion.py
```diff
@@ -105,6 +113,47 @@ def get_text_edit_range(text_edit: Union[TextEdit, InsertReplaceEdit]) -> Range:
     return text_edit['range']
 
 
+def is_range(val: Optional[Any]) -> TypeGuard[Range]:
+    return isinstance(val, dict) and 'start' in val and 'end' in val
+
+
+def is_edit_range(val: Optional[Any]) -> TypeGuard[CompletionEditRange]:
+    return isinstance(val, dict) and 'insert' in val and 'replace' in val
```
rchl, line 121: `Any` doesn't need `Optional`

```suggestion
def is_range(val: Any) -> TypeGuard[Range]:
    return isinstance(val, dict) and 'start' in val and 'end' in val


def is_edit_range(val: Any) -> TypeGuard[CompletionEditRange]:
    return isinstance(val, dict) and 'insert' in val and 'replace' in val
```

rchl (approved):
Looks good to me but lets wait for one extra eyes.

plugin/core/protocol.py
predragnikolic: 8ca4d8e

rchl (approved):

plugin/completion.py
```diff
@@ -105,6 +113,44 @@ def get_text_edit_range(text_edit: Union[TextEdit, InsertReplaceEdit]) -> Range:
...
+    return isinstance(val, dict) and 'start' in val and 'end' in val
+
+
+def is_edit_range(val: Any) -> TypeGuard[CompletionEditRange]:
+    return isinstance(val, dict) and 'insert' in val and 'replace' in val
+
+
+def completion_with_defaults(item: CompletionItem, item_defaults: CompletionItemDefaults) -> CompletionItem:
+    """ Currently supports defaults for: ["editRange", "insertTextFormat", "data"] """
+    default_text_edit = None  # type: Optional[Union[TextEdit, InsertReplaceEdit]]
```
jwortmann, line 126:
```suggestion
    """ Currently supports defaults for: ["editRange", "insertTextFormat", "data"] """
    if not item_defaults:
        return item
    default_text_edit = None  # type: Optional[Union[TextEdit, InsertReplaceEdit]]
```

All of the following dict accesses and if-checks are unnecessary if the server doesn't provide the CompletionList.itemDefaults, which is probably the case for most language servers currently.

jwortmann (approved):

PR #2196, opened 2023-02-12
add some tests for completion formatting for the UI

rchl:


tests/test_completion.py
```diff
@@ -903,3 +903,140 @@ def test_conversion_of_edit_range_to_text_edit_when_it_includes_insert_replace_f
...
+                    "detail": "()"
+                },
+            },
+            trigger='banner?()',
+        )
+
+    def test_filter_text_and_label_details_2(self) -> None:
+        self._verify_completion(
+            {
+                "filterText": "banner?()",
```
predragnikolic, line 1006: Just by looking at this test, 
I do not know if `filterText` will be the trigger or `label+labelDetails`

I would change `filterText` to "ban" just so it is more obvious that `label+labelDetails` will be the trigger...
```suggestion
                "filterText": "ban",
```

predragnikolic (approved):

tests/test_completion.py
rchl: I think the point of this test is to test the `(lsp_label + lsp_label_detail).startswith(lsp_filter_text)` logic so the filter text should be either like now or have extra characters at the end.

rchl: Does that clarify or am I wrong here?

rchl: actually will remove that test since it appears to be basically the same as the one above it, just less clear on what is expected.

predragnikolic (approved):

PR #2204, opened 2023-02-17
tell if any selection changed in addition to just first region

rchl:
Detect whenever any selection region has changed in addition to detecting whether just the first one has changed.

This is an extra functionality that is current not utilized but will be useful for #1702.

Fixes #2182

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -425,16 +425,9 @@ def text_document_identifier(view_or_uri: Union[DocumentUri, sublime.View]) -> T
     return {"uri": uri}
 
 
-def get_first_selection_region(selection: sublime.Selection) -> Optional[sublime.Region]:
-    return selection[0] if len(selection) else None
-
-
 def first_selection_region(view: sublime.View) -> Optional[sublime.Region]:
-    """@deprecated Use :meth:`get_first_selection_region` instead."""
-    try:
-        return view.sel()[0]
-    except IndexError:
-        return None
+    selection = view.sel()
+    return selection[0] if len(selection) else None
 
 
 def has_single_nonempty_selection(view: sublime.View) -> bool:
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -426,8 +426,10 @@ def text_document_identifier(view_or_uri: Union[DocumentUri, sublime.View]) -> T
 
 
 def first_selection_region(view: sublime.View) -> Optional[sublime.Region]:
-    selection = view.sel()
-    return selection[0] if len(selection) else None
+    try:
+        return view.sel()[0]
+    except IndexError:
+        return None
 
 
 def has_single_nonempty_selection(view: sublime.View) -> bool:
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -432,6 +432,14 @@ def get_first_selection_region(selelection: sublime.Selection) -> Optional[subli
         return None
 
 
+def first_selection_region(view: sublime.View) -> Optional[sublime.Region]:
+    """@deprecated Use :meth:`get_first_selection_region` instead."""
+    try:
+        return view.sel()[0]
+    except IndexError:
+        return None
+
+
 def has_single_nonempty_selection(view: sublime.View) -> bool:
     selections = view.sel()
     return len(selections) == 1 and not selections[0].empty()
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -425,9 +425,9 @@ def text_document_identifier(view_or_uri: Union[DocumentUri, sublime.View]) -> T
     return {"uri": uri}
 
 
-def first_selection_region(view: sublime.View) -> Optional[sublime.Region]:
+def get_first_selection_region(selelection: sublime.Selection) -> Optional[sublime.Region]:
     try:
-        return view.sel()[0]
+        return selelection[0]
     except IndexError:
         return None
 
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -425,11 +425,8 @@ def text_document_identifier(view_or_uri: Union[DocumentUri, sublime.View]) -> T
     return {"uri": uri}
 
 
-def get_first_selection_region(selelection: sublime.Selection) -> Optional[sublime.Region]:
-    try:
-        return selelection[0]
-    except IndexError:
-        return None
+def get_first_selection_region(selection: sublime.Selection) -> Optional[sublime.Region]:
+    return selection[0] if len(selection) else None
 
 
 def first_selection_region(view: sublime.View) -> Optional[sublime.Region]:
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -299,12 +299,13 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
                             return
         self.view.erase_status(self.ACTIVE_DIAGNOSTIC)
 
+    def session_buffers_async(self, capability: Optional[str] = None) -> Generator[SessionBuffer, None, None]:
+        for sv in self.session_views_async():
+            if capability is None or sv.has_capability_async(capability):
+                yield sv.session_buffer
+
     def session_views_async(self) -> Generator[SessionView, None, None]:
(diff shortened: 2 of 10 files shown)


plugin/documents.py
```diff
@@ -842,31 +846,37 @@ def _register_async(self) -> None:
...
             return
         self._clear_highlight_regions()
         if userprefs().document_highlight_style:
             self._when_selection_remains_stable_async(
-                self._do_highlights_async, current_region, after_ms=self.highlights_debounce_time)
+                self._do_highlights_async, changed_region, after_ms=self.highlights_debounce_time)
         self.do_signature_help_async(manual=False)
 
-    def _update_stored_region_async(self) -> Tuple[bool, sublime.Region]:
+    def _update_stored_selection_async(self) -> Tuple[Optional[sublime.Region], bool]:
```
jwortmann, line 858: Each call of this function now makes up to three API calls of `View.sel()` (one in `first_selection_region` and you do `list(self.view.sel())` in the code below twice).
It shouldn't be hard to reduce this to a single API call.

jwortmann:
I haven't looked exactly at all changes in this PR, but I neither get any more DocumentHighlights nor code actions on this branch. So something doesn't seem right at the moment.

plugin/documents.py
rchl: Good catch! Did not realize (or forgot) that empty regions evaluate to false.

plugin/core/views.py
```diff
@@ -425,7 +425,15 @@ def text_document_identifier(view_or_uri: Union[DocumentUri, sublime.View]) -> T
...
 
+def get_first_selection_region(selelection: sublime.Selection) -> Optional[sublime.Region]:
+    try:
+        return selelection[0]
+    except IndexError:
+        return None
+
+
 def first_selection_region(view: sublime.View) -> Optional[sublime.Region]:
+    """@deprecated Use :meth:`get_first_selection_region` instead."""
```
jwortmann, line 436: Hm... what's wrong with this function to deprecate it?

If I had written this function, I probably would not have used a `try ... except` for it, but instead
```python
selection = view.sel()
return selection[0] if selection else None
```
... to save a few lines, but that's just a matter of preference.

plugin/core/views.py
```diff
@@ -425,7 +425,15 @@ def text_document_identifier(view_or_uri: Union[DocumentUri, sublime.View]) -> T
     return {"uri": uri}
 
 
+def get_first_selection_region(selelection: sublime.Selection) -> Optional[sublime.Region]:
```
jwortmann, line 428: Is this function really necessary? Now you need to put `view.sel()` into the argument at all the places where it is called.
And I think you can rewrite the relevant change from this PR in a simpler way without the need to use this function
```python
selection = view.sel()
if selection:
    current_first_region = selection[0]
    ...
```

plugin/documents.py
```diff
@@ -842,31 +849,37 @@ def _register_async(self) -> None:
...
-                return True, current_region
-        return False, sublime.Region(-1, -1)
+        current_first_region = None
+        any_selection_changed = False
+        selection = self.view.sel()
+        first_region = get_first_selection_region(selection)
+        if first_region is not None:
+            if not self._stored_selection or self._stored_selection[0] != first_region:
+                current_first_region = first_region
+            current_selection = list(selection)
```
jwortmann, line 878: selection is already a list, so this operation seems needless

plugin/core/views.py
rchl: I think there is a bug in 3.3 host that makes it evaluate to true when not empty. That is why try catch is used

rchl: As far as deprecation. Well, decide to switch to it everywhere instead of having two similar ones but then noticed that is used in a couple of helper packages

plugin/documents.py
rchl: It's not a list. It's a custom object that acts as a list but not quite

plugin/core/views.py
rchl: And if we want to avoid calling it multiple times then i needed to change it like that. Though I'm not convinced it was necessary to options that in the first place

plugin/core/views.py
jwortmann: > I think there is a bug in 3.3 host that makes it evaluate to true when not empty.

The result of `View.sel()` is a list, so I would expect it to evaluate to true in all cases except if there is no cursor in the view (only possible via plugins). Or do you have more information about this "bug"?

plugin/documents.py
jwortmann: You're right, I think I've always used it like a list, and so I falsely assumed the return type was just `List[sublime.Region]`.

plugin/core/views.py
jwortmann: Looking at the API docs, I think
```python
return selection[0] if len(selection) else None
```
should do the trick. But not necessary to change it here.

plugin/core/views.py
rchl: Note that `len` does a call to a native API so it would be in practice less efficient. Does it matter? Probably not.

```py
    def __len__(self):
        return sublime_api.view_selection_size(self.view_id)

    def __getitem__(self, index):
        r = sublime_api.view_selection_get(self.view_id, index)
        if r.a == -1:
            raise IndexError()
        return r
```

rchl: Well, I figured I should test instead of guessing.

```py
    def test_len():
        return sel[0] if len(sel) else None

    repeat = 100000
    times = timeit.repeat(test_len, repeat=3, number=repeat)
    return 'Best of {}: {} usec.'.format(repeat, min(times))

	# 0.26699970833396947 usec.
```

```py
    def test_trycatch():
        try:
            return sel[0]
        except Exception:
            return None

    repeat = 100000
    times = timeit.repeat(test_trycatch, repeat=3, number=repeat)
    return 'Best of {}: {} usec.'.format(repeat, min(times))

	# 0.4004749166670081 usec.
```

So no, `len` variant is faster.

rchl: I rewrote `_update_stored_selection_async` in a different way.

jfcherng: Weird. I get different benchmark results.

```py
import sublime
import sublime_plugin
import timeit


class ExampleCommand(sublime_plugin.TextCommand):
    def run(self, edit: sublime.Edit) -> None:
        sel = self.view.sel()
        repeat = 100000

        def test_len_1():
            return sel[0] if len(sel) else None

        def test_len_2():
            try:
                return sel[0]
            except Exception:
                return None

        def test_len_3():
            return next(iter(sel), None)

        times = timeit.repeat(test_len_1, repeat=3, number=repeat)
        print("(test_len_1) Best of {}: {} usec.".format(repeat, min(times)))
        times = timeit.repeat(test_len_2, repeat=3, number=repeat)
        print("(test_len_2) Best of {}: {} usec.".format(repeat, min(times)))
        times = timeit.repeat(test_len_3, repeat=3, number=repeat)
        print("(test_len_3) Best of {}: {} usec.".format(repeat, min(times)))
```

results in 

```
(test_len_1) Best of 100000: 0.39621629999999186 usec.
(test_len_2) Best of 100000: 0.2581017000000543 usec.
(test_len_3) Best of 100000: 0.2883709000000181 usec.
```

rchl: I can see the same here when using actual `view.sel` rather than manually created `sublime.Selection(1).

```
* view.sel()
(test_len_1) Best of 100000: 0.26706908333332535 usec.
(test_len_2) Best of 100000: 0.4003557916666409 usec.
(test_len_3) Best of 100000: 0.4005454583333119 usec.

* sublime.Selection(1)
(test_len_1) Best of 100000: 0.5351135833333274 usec.
(test_len_2) Best of 100000: 0.26820062500001995 usec.
(test_len_3) Best of 100000: 0.2768708750000428 usec.
```

rchl: Which seems to boil down to `__len__` implementation having to do an API call and that appears to have more overhead when view ID is not valid (does not exist).

rchl: Any more comments on this?

PR #2206, opened 2023-02-24
better grouping of default settings

rchl:
Did a bit of grouping of options to make default settings a bit more organized.

Also removed default configs for rust-analyzer, clangd and gopls since we have dedicated packages for those.

Improvement suggestions welcome.

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -5,10 +5,20 @@
   // Show permanent language server status in the status bar.
   "show_view_status": true,
 
+  // --- Document Formatting ------------------------------------------------------------
+
   // Run the server's formatProvider (if supported) on a file before saving.
   // This option is also supported in syntax-specific settings and/or in the
   // "settings" section of project files.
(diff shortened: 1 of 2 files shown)


predragnikolic (approved):

LSP.sublime-settings
```diff
@@ -305,12 +316,6 @@
       ],
       "selector": "source.reason | source.ocaml",
     },
-    "gopls": {
-      "command": [
-        "gopls"
-      ],
-      "selector": "source.go",
-    },
```
rwols, line 313: I used to announce this removal in a previous release message. Make sure to announce this in the current release message. Now that I’m writing this, I’m not sure if it’s a good idea to remove this without an announcement message.

rwols: Because it may break current user’s workflow.

rchl: Added changelog listing the breaking changes.

It will be updated before release to list all other changes.

PR #2208, opened 2023-02-25
Call hierarchy: highlight call location for incoming calls

jwortmann:
Based on a suggestion from https://github.com/sublimelsp/LSP-typescript/issues/186#issuecomment-1442843626.

For the incoming calls ("callers of...") in the call hierarchy, if you click on an item, it will currently open the first line of the function definition. But with this change, it would highlight the (first) location of the actual call within the function instead.

I've also tested what VSCode does, and it seems that it randomly highlights either the call location, or the first line of the function. And it randomly changes even for the same item... probably they have a bug somewhere.
<details>
<summary>Recording from VSCode</summary>

[vscode-call-hierarchy.webm](https://user-images.githubusercontent.com/6579999/221362887-45291e80-62ef-427e-aef1-7c388492dcf5.webm)

</details>

Anyhow, I think it is more useful doing it this way (up for discussion).

The implementation is unfortunately very ugly, but I don't see a different way other than storing the location into the `CallHierarchyItem` itself, because a CallHierarchyItem is not necessarily unique, and the `get_tree_item` method to render an item has no other input. If someone sees a better way, any improvements are of course welcome.


plugin/hierarchy.py
```diff
@@ -28,7 +33,21 @@
...
+    'kind': SymbolKind,
+    'tags': NotRequired[List[SymbolTag]],
+    'detail': NotRequired[str],
+    'uri': DocumentUri,
+    'range': Range,
+    'selectionRange': Range,
+    'data': NotRequired[LSPAny],
+    # A custom field to store the first call location, given in CallHierarchyIncomingCall.fromRanges.
+    '_$ublime_call_range': NotRequired[Range]
+})
```
rchl, line 47: Would it be a big lift to instead make this a wrapper element that would contain the original `CallHierarchyItem` as an `item` property and whatever extra data we need (call range in this case) as a separate property? Then we wouldn't need to duplicate `CallHierarchyItem` properties here and risk things getting out of sync with the original `CallHierarchyItem`.

rchl:
Sorry but I still didn't quite like it due to loosening of the types and mixing wrapper item and normal items.

I've pushed an alternative fix that normalizes hierarchy items into a single `HierarchyData` structure (perhaps name is not ideal?). Let me know what you think.

plugin/core/protocol.py
```diff
@@ -5840,8 +5840,10 @@ class TokenFormat(Enum):
 
 # AUTOGENERATED TYPES END
 
+R = TypeVar('R')
 
-class Request:
+
+class Request(Generic[R]):
```
rchl, line 5846: This is not strictly necessary but I was meaning to do this for a while. It allows us to type the responses.

jwortmann:
This is not how it works. You lose information in the `to_hierarchy_item` function (e.g. `data`, alter the `selectionRange`). You need to preserve the `CallHierarchyItem`'s / `TypeHierarchyItem`'s as they are reported from the server, in order to make the new requests. Make sure the `element` in line 68 has the correct type. (`CallHierarchyItem` or `TypeHierarchyItem`, respectively)

rchl:
Ok. I will look a bit more tomorrow. I thought that nothing else from the response was accessed.

And of course we can revert my commits if needed

rchl:
Pushed alternative version that always uses a wrapper element.

jwortmann:
To be honest, I don't really like some of the new names. `element` should keep its name, because otherwise the overridden method signatures don't match the base class anymore. I'm surprised pyright doesn't emit an error for this. The reason why it was called `element` is, because that's the name from the VSCode API.

And I would still use HirarchyItemWrapper as the name instead of HierarchyData.

Besides that, I think it should work now and the other changes seem fine to me.

plugin/hierarchy.py
```diff
@@ -30,15 +31,20 @@
 
 HierarchyItem = Union[CallHierarchyItem, TypeHierarchyItem]
 
+HirarchyItemWrapper = TypedDict('HirarchyItemWrapper', {
```
jwortmann, line 34: typo Hi**e**rarchy

jwortmann: It was my own typo from above btw :)

jwortmann:
I can't approve my own pull request, but it looks good to me now.

rchl (approved):

PR #2209, opened 2023-02-25
protect again invalid `auto_complete_triggers` values

rchl:
could crash with invalid user values: https://github.com/TerminalFi/LSP-copilot/issues/90


rwols (approved):

rchl:
This fix is not released yet so it would not help with your issue at this point.

The invalid setting could in theory be set by some random package dynamically for open file in which case you wouldn't see it in preferences.

If you stumble upon this issue then you can run `view.settings().get('auto_complete_triggers')` in the ST console and you should see the faulty entry. Maybe it would be clear when seeing it which package has set it.

PR #2211, opened 2023-02-26
notify opened files immediately if there are pending diagnostics

### LINKED ISSUE #2200: Does not cache diagnostics when switching between files
**Describe the bug**
Sublime LSP does not cache diagnostics that are reported from the server for a folder/file.

**To Reproduce**
1. Language Server sends diagnostics on saving a file and on adding a workspace folder via initialization or config update.
2. The Open/Close capability is not reported, nor implemented
3. When switching between `fileA` to `fileB` and then back to `fileA` the diagnostics that are previously displayed in `fileA` are lost.

**Expected behavior**
Reported diagnostics are only cleared, when the server clears them. According to the spec, diagnostics are owned by the server and must be explicitly cleared by it.

From the spec: `Diagnostics are “owned” by the server so it is the server’s responsibility to clear them if necessary.`

**Screenshots**

**Additional context**
Implementing Open/Close mitigates that issue, but actually on File open, most likely pull diagnostics should be called on the LS if it has that capability.

rchl:
Diagnostics are stored in `Session` and there might exist diagnostics for not-yet-opened files so notify Session Buffers immediately about those when creating new buffers.

This fixes somewhat obscure bug for servers that don't implement `didOpen` but also makes diagnostics show more quickly after opening files for other servers (in case diagnostics were already reported for those files).

Fixes #2200

### DIFF
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -53,6 +53,7 @@ class SessionView:
         if session_buffer is None:
             session_buffer = SessionBuffer(self, buffer_id, uri)
             self._session_buffers[key] = session_buffer
+            self._session.register_session_buffer_async(session_buffer)
         else:
             session_buffer.add_session_view(self)
         self._session_buffer = session_buffer
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -53,10 +53,11 @@ class SessionView:
         if session_buffer is None:
             session_buffer = SessionBuffer(self, buffer_id, uri)
             self._session_buffers[key] = session_buffer
+            self._session_buffer = session_buffer
             self._session.register_session_buffer_async(session_buffer)
         else:
+            self._session_buffer = session_buffer
             session_buffer.add_session_view(self)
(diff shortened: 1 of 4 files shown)


plugin/session_view.py
```diff
@@ -53,6 +53,7 @@ def __init__(self, listener: AbstractViewListener, session: Session, uri: Docume
         if session_buffer is None:
             session_buffer = SessionBuffer(self, buffer_id, uri)
             self._session_buffers[key] = session_buffer
+            self._session.register_session_buffer_async(session_buffer)
```
rchl, line 56: Previously `SessionBuffer` registered itself in `Session` with:

```py
self._session.register_session_buffer_async(self)
```

This caused crash as `Session.register_session_buffer_async` was called before `self._session_buffers[key] = session_buffer` and would crash depending what we would want to do with `SessionBuffer`.

I think it makes more sense for `SessionView` to register `SessionBuffer` with `Session` as it is also responsible for creating SB.

PR #2212, opened 2023-02-27
Improve label detail support in completions

ryuukk:
Before: 

image

After:

image


plugin/completion.py
```diff
@@ -69,7 +69,7 @@ def format_completion(
         annotation = lsp_label_description or lsp_detail
     elif lsp_label.startswith(lsp_filter_text):
         trigger = lsp_label
-        annotation = lsp_detail
+        annotation = lsp_label_description or lsp_detail
         if lsp_label_detail:
             details.append(html.escape(lsp_label + lsp_label_detail))
         if lsp_label_description:
```
rchl, line 75: If `lsp_label_description` is used here then I would probably not add it to `details` since it would just be a duplicated data then.

rchl (commented):
Will need to be manually tested with various servers to ensure that it doesn't result in worse behavior in some cases.

I will do my share of testing, hopefully others will do theirs. :)

plugin/completion.py
```diff
@@ -69,7 +69,7 @@ def format_completion(
         annotation = lsp_label_description or lsp_detail
     elif lsp_label.startswith(lsp_filter_text):
         trigger = lsp_label
-        annotation = lsp_detail
+        annotation = lsp_label_description or lsp_detail
```
rchl, line 72: One datapoint from Pyright.

Before:

Screenshot 2023-03-04 at 22 44 01

After:

Screenshot 2023-03-04 at 22 39 32

I would say that it makes things better in this case too.

plugin/completion.py
rchl: To not loose any information, we should put `lsp_detail` in `details` if `lsp_label_description` is used and vice-versa.

rchl:
@jwortmann can you have a look since you were involved with related changes?

plugin/completion.py
```diff
@@ -76,11 +76,13 @@ def format_completion(
         annotation = lsp_label_description or lsp_detail
     elif lsp_label.startswith(lsp_filter_text):
         trigger = lsp_label
-        annotation = lsp_detail
+        if lsp_label_description:
+            annotation = lsp_label_description
+            details.append(html.escape(lsp_detail))
```
jwortmann, line 81: I think this line should be put under a `if lsp_detail:` condition, because otherwise and if `lsp_detail` is not set by the server, there will be an empty field like `| |` in the details pane, because the items (empty string in this case) get joined together like that (`details=" | ".join(details)`).

jwortmann:
So if we use `lsp_label_description` for the annotation field here, we should probably do the same also in the `else` block below, right? That would be more consistent.

I think it depends on the server whether it is better or worse than using the `lsp_detail`. For pyright I would even find the "Auto-import" annotation more important, because those completions automatically add another import line at the top of the file, and this way I could avoid them more easily ;) (but I use `"python.analysis.autoImportCompletions": false` setting anyway).
So in general this will be a matter of preference too. But from my side it should be okay to merge it (after the two other points are resolved).

rchl:
> For pyright I would even find the "Auto-import" annotation more important, because those completions automatically add another import line at the top of the file, and this way I could avoid them more easily ;)

For this specific case I think it's much more preferred to see the name of the file/package in the completion list rather than just "auto-import" because, if you know the server a little bit, you know that seeing the name equals auto-import and so you still know that the auto-import will happen and at the same time you have much more information up-front.

plugin/completion.py
rchl: I've tried to reproduce this case with a test but I can't quite do it.

It's possible that `lsp_detail` is not set by the server but if are gonna just empty string to `details` then `" | ".join(details)` will still result in empty string.

Now this would be problematic with there would be another item added to `details` but here is where this seems like it can't happen. It could only happen if we'd also add another item with:

```
        if lsp_label_detail:
            details.append(html.escape(lsp_label + lsp_label_detail))
```

below but it's not possible to create values that would trigger this case rather than the previous block guarded by  the `if lsp_label_detail and (lsp_label + lsp_label_detail).startswith(lsp_filter_text):` condition.

rchl: I will still add an if check to not unnecessarily add empty item to `details` and potentially guard future code from triggering this issue.

rchl:
> So if we prefer `lsp_label_description` (if set) for the annotation field here, we should probably do the same also in the `else` block below, right? That would be more consistent.

Applied same logic to the else branch and also consolidated those two blocks. It's probably not easy to review.

tests/test_completion.py
```diff
@@ -708,7 +708,7 @@ def check(
         )
         check(
             resolve_support=False,
-            expected_regex=r"^f\(X&amp; x\) \| does things$",
+            expected_regex=r"^f\(X&amp; x\)$",
             label="f",
             label_details={"detail": "(X& x)", "description": "does things"}
         )
```
rchl, line 714: Those test are poor-mans version of the new ones so it would possibly be OK to remove those and convert to the new ones, if those test anything extra.

I say "poor-mans" version because those ignore "annotation" for example.

jwortmann (approved):

jwortmann:
> For this specific case I think it's much more preferred to see the name of the file/package in the completion list rather than just "auto-import" because, if you know the server a little bit, you know that seeing the name equals auto-import and so you still know that the auto-import will happen and at the same time you have much more information up-front.

Yeah okay, I kind of agree, but this is only a valid assumption if this LSP server is the only source of completions which use the `annotation` field.

rchl:
> Yeah okay, I kind of agree, but this is only a valid assumption if this LSP server is the only source of completions which use the `annotation` field.

Yes, but that's intentional since we can't implement things the way spec intended due to ST limitations so we have to use heuristics or optimize for specific cases.

PR #2213, opened 2023-03-03
don't initialize TextChange listeners for widgets

rchl:
while debugging something I've realized that the `TextChangeListener` would be initialized when triggering Command Palette because `is_applicable` returned True.

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -63,7 +63,8 @@ SUBLIME_WORD_MASK = 515
 
 def is_regular_view(v: sublime.View) -> bool:
     # Not from the quick panel (CTRL+P), and not a special view like a console, output panel or find-in-files panels.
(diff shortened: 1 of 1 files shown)


rwols (approved):

jwortmann:
Wouldn't this already been covered by the `v.element() is None` condition?

If not, that would appear to be a bug in the ST API then?

https://www.sublimetext.com/docs/api_reference.html#sublime.View.element

docs

jfcherng:
`.element()` is only for ST >= 4050?

rchl:
I can report it.

rchl:
It's probably a timing issue as checking from EventListener.on_activate does return a value.

rchl:
https://github.com/sublimehq/sublime_text/issues/5887

PR #2222, opened 2023-03-19
Fix some features might not work with dynamical registration

jwortmann:
While working on another PR #2221, I noticed that some features in the SessionBuffer class only check for the server capabilities which are stored in the Session, but not specific to the SessionBuffer. This means that it could miss a server capability, if it was dynamically registered with a `documentSelector` (see [TextDocumentRegistrationOptions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentRegistrationOptions)).

The relevant code for that is at https://github.com/sublimelsp/LSP/blob/4ed8b7a75b091311bff041eb63aea2744c35b23e/plugin/core/sessions.py#L1779-L1785

I have no example of this where it happens in practice (I don't know any server which uses dynamic registration with documentSelector). But it should be relevant for pull diagnostics (the other PR), because they need to include the optional `identifier` which can be provided in the server capabilities, in the diagnostics requests. So if a server dynamically registers this capabiliy twice, and with different `documentSelector` and `identifier`, we must make sure to use the correct one when sending the requests.

Please check if I didn't missunderstand something.

### DIFF
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -366,7 +366,7 @@ class SessionBuffer:
     # --- textDocument/documentColor -----------------------------------------------------------------------------------
 
     def _do_color_boxes_async(self, view: sublime.View, version: int) -> None:
-        if self.session.has_capability("colorProvider"):
+        if self.has_capability("colorProvider"):
             self.session.send_request_async(
                 Request.documentColor(document_color_params(view), view),
                 self._if_view_unchanged(self._on_color_boxes_async, version)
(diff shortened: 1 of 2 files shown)


rchl:
Maybe also this can use `self.get_capability_async`:

https://github.com/sublimelsp/LSP/blob/6d96bd7e38c0c587dfbb6550fce92af668669121/plugin/session_view.py#L359-L359

rchl (approved):

rchl:
Seems like a correct thing to do to me.

PR #2228, opened 2023-03-31
html-escape diagnostic-related strings

### LINKED ISSUE #2223: Invalid HTML in diagnostics["code"] breaks hover popups.
**Describe the bug**
If the `code` field of the diagnostic contains invalid html, the hover popup is broken

**To Reproduce**
1. Use haskell-language-server
2. hover over a line with `redundant <$>` hint (for example, `inputForConfig config = (<> config) <$> standardIOConfig >>= inputForConfig`
3. The hover popup is cut in half, and ST's console complaints about parsing errors

**Expected behavior**
Hover is properly formatted

- OS: Win10
- Sublime Text version: 4143
- LSP version: current master
- Language servers used: [HLS] with suggested config

**Root cause**
In the `format_diagnostics_for_html` method we append the contents of `diagnostics["code"]` as-is, without escaping (https://github.com/sublimelsp/LSP/blob/main/plugin/core/views.py#L1006).

As far as I understand, the specs of LSP say nothing about that field, it can be an arbitrary text. I suggest
the following diff:
```diff
diff --git a/plugin/core/views.py b/plugin/core/views.py
index cfdce3e..7af8239 100644
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -1003,7 +1003,7 @@ def format_diagnostic_for_html(
         if code_description:
             code.append(make_link(code_description["href"], diagnostic.get("code")))
         else:
-            code.append(_with_color(diagnostic["code"], "color(var(--foreground) alpha(0.6))"))
+            code.append(_with_color(text2html(diagnostic["code"]), "color(var(--foreground) alpha(0.6))"))
         code.append(_with_color(")", "color(var(--foreground) alpha(0.6))"))
     else:
         code = None
```
I do not have a good idea on how it could impact other LSP servers, comments and suggestions are welcome.

I'll submit a PR sometime later this week.

Cheers!

rchl:
- Removed unused `view` argument from `format_diagnostic_for_html`
- Escape diagnostic properties when including them in HTML

Fixes #2223

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -62,13 +62,12 @@ def format_completion(
         args = '{{"view_id":{},"command":"lsp_resolve_docs","args":{{"index":{},"session_name":"{}"}}}}'.format(
             view_id, index, session_name)
         href = 'subl:lsp_run_text_command_helper {}'.format(args)
-        details.append("<a href='{}'>More</a>".format(href))
+        details.append('<a href="{}">More</a>'.format(href))
     if lsp_label_detail and (lsp_label + lsp_label_detail).startswith(lsp_filter_text):
         if lsp_label_detail[0].isalnum() and lsp_label.startswith(lsp_filter_text):
             # labelDetails.detail is likely a type annotation
             # Don't append it to the trigger: https://github.com/sublimelsp/LSP/issues/2169
             trigger = lsp_label
-
             details.append(html.escape(lsp_label_detail))
         else:
             # labelDetails.detail is likely a function signature
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -63,7 +63,7 @@ def format_completion(
         args = '{{"view_id":{},"command":"lsp_resolve_docs","args":{{"index":{},"session_name":"{}"}}}}'.format(
             view_id, index, session_name)
         href = 'subl:lsp_run_text_command_helper {}'.format(args)
-        details.append(make_link(href, 'More'))
+        details.append("<a href='{}'>More</a>".format(href))
     if lsp_label_detail and (lsp_label + lsp_label_detail).startswith(lsp_filter_text):
         if lsp_label_detail[0].isalnum() and lsp_label.startswith(lsp_filter_text):
             # labelDetails.detail is likely a type annotation
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -62,7 +62,7 @@ def format_completion(
         args = '{{"view_id":{},"command":"lsp_resolve_docs","args":{{"index":{},"session_name":"{}"}}}}'.format(
             view_id, index, session_name)
         href = 'subl:lsp_run_text_command_helper {}'.format(args)
-        details.append('<a href="{}">More</a>'.format(href))
+        details.append("<a href='{}'>More</a>".format(href))
     if lsp_label_detail and (lsp_label + lsp_label_detail).startswith(lsp_filter_text):
         if lsp_label_detail[0].isalnum() and lsp_label.startswith(lsp_filter_text):
             # labelDetails.detail is likely a type annotation
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -21,7 +21,6 @@ from .core.settings import userprefs
 from .core.typing import Callable, List, Dict, Optional, Generator, Tuple, Union, cast, Any, TypeGuard
 from .core.views import COMPLETION_KINDS
 from .core.views import FORMAT_STRING, FORMAT_MARKUP_CONTENT
-from .core.views import make_link
 from .core.views import MarkdownLangMap
 from .core.views import minihtml
 from .core.views import range_to_region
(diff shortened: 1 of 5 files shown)


plugin/core/views.py
```diff
@@ -780,13 +780,12 @@ def text2html(content: str) -> str:
...
 
 def make_link(href: str, text: Any, class_name: Optional[str] = None, tooltip: Optional[str] = None) -> str:
-    if isinstance(text, str):
-        text = text.replace(' ', '&nbsp;')
     link = "<a href='{}'".format(href)
     if class_name:
         link += " class='{}'".format(class_name)
     if tooltip:
         link += " title='{}'".format(html.escape(tooltip))
+    text = text2html(str(text)).replace(' ', '&nbsp;')
```
predrag-codetribe, line 788: make_link  is called inside the `format_completion` 
I would say that calling text2html inside here will affect the completion performance

I will test with
https://github.com/sublimelsp/LSP-performance-test

plugin/core/views.py
```diff
@@ -976,44 +975,36 @@ def _format_diagnostic_related_info(
     location = info["location"]
     return '<a href="{}">{}</a>: {}'.format(
         location_to_href(config, location),
-        location_to_human_readable(config, base_dir, location),
-        info["message"]
+        text2html(location_to_human_readable(config, base_dir, location)),
```
predrag-codetribe, line 978: is there a need to call text2html for this place, the location_to_human_readable will return a path?

plugin/core/views.py
```diff
@@ -976,44 +975,36 @@ def _format_diagnostic_related_info(
...
+        text2html(info["message"])
     )
 
 
-def _with_color(text: Any, hexcolor: str) -> str:
-    return '<span style="color: {};">{}</span>'.format(hexcolor, text)
+def _with_color(text: str, hexcolor: str) -> str:
+    return '<span style="color: {};">{}</span>'.format(hexcolor, text2html(text))
```
predrag-codetribe, line 984: I would not place the text2html inside this function, 
To avoid unnecessary work, for example
```py
formatted.append(_with_color("(", "color(var(--foreground) alpha(0.6))")) // we would unnecessary call text2html
formatted.append(_with_color(str(code), "color(var(--foreground) alpha(0.6))")) // only here calling text2html would be appreciated
formatted.append(_with_color(")", "color(var(--foreground) alpha(0.6))")) // we would unnecessary call text2html
```

Instead I would suggest:
```py
formatted.append(_with_color("(", "color(var(--foreground) alpha(0.6))"))
formatted.append(_with_color(text2html(code), "color(var(--foreground) alpha(0.6))"))
formatted.append(_with_color(")", "color(var(--foreground) alpha(0.6))")) 
```

plugin/core/views.py
predrag-codetribe: this branch
Best of 3: 0.2930799584137276 usec. All times: [0.2931855833157897, 0.29389379161875695, 0.2930799584137276]
Best of 3: 0.2897154582897201 usec. All times: [0.2897154582897201, 0.29021529166493565, 0.2925562916789204]

main branch
Best of 3: 0.20980629161931574 usec. All times: [0.21008566673845053, 0.20980891666840762, 0.20980629161931574]
Best of 3: 0.20813845819793642 usec. All times: [0.20813845819793642, 0.20843349990900606, 0.20843862497713417]

rchl: Yes, it will affect performance because it's an extra logic. But it's a logic that has to run if we want things to work correctly in all circumstances.

That said, I will look into it a bit more to see if we can do "fast paths" in some cases.
But I don't want to decide outside of such functions whether to escape or not because that would make the logic error-prone in regards to future changes.

Applies to all 3 of your comments.

predragnikolic: All of my comments have one thing in common, to call the text2html less times.

> That said, I will look into it a bit more to see if we can do "fast paths" in some cases.

+1

predragnikolic (approved):

plugin/core/views.py
predragnikolic: Pushed this code 964e5a8
It makes the format completion function faster

predragnikolic:
Does anyone have any comments?
If not, I would like to merge this?

plugin/completion.py
```diff
@@ -63,12 +62,13 @@ def format_completion(
...
             view_id, index, session_name)
         href = 'subl:lsp_run_text_command_helper {}'.format(args)
-        details.append(make_link(href, 'More'))
+        details.append("<a href='{}'>More</a>".format(href))
     if lsp_label_detail and (lsp_label + lsp_label_detail).startswith(lsp_filter_text):
         if lsp_label_detail[0].isalnum() and lsp_label.startswith(lsp_filter_text):
             # labelDetails.detail is likely a type annotation
             # Don't append it to the trigger: https://github.com/sublimelsp/LSP/issues/2169
             trigger = lsp_label
+
```
rchl, line 71: This blank line is to fix linting? We don't generally add blank lines in normal flow.

PR #2232, opened 2023-04-03
Allow style overrides for inlay_hints.css

### LINKED ISSUE #2229: Settings for editing inlay hints
I wish to style the inlay hints that LSP provides, however, LSP only allows for very limited styling, basically background color, font type, font color, but nothing else. There are these ugly round borders on the little hint that cannot be removed without editing inlay_hint.py directly to change the enforced CSS.

The ability to either change all CSS values of the inlay hint though variables, or the ability to outright overwrite the css that's currently there from a settings file.

Editing LSP with PackageResourceViewer by hand each time there is an update.

**Additional context**
Here is my cute little CSS improvement for my own enjoyment, no more distracting button-like border :<)
image

jwortmann:
Closes #2229.

This allows user modifications of the inlay hint styles more easily, by overriding only the `inlay_hints.css` file.
When you create an override for the file, ST must be restarted, because the CSS is loaded only when the LSP package is loaded (on ST start).

I hope ST devs one day will fix the color scheme bug that I mentioned in the linked issue, but I think for now this is a feasible workaround.

### DIFF
--- a/docs/src/customization.md
+++ b/docs/src/customization.md
@@ -13,6 +13,13 @@ html {
 to `Packages/User/mdpopups.css`.
 See the [mdpopups documentation](http://facelessuser.github.io/sublime-markdown-popups/) for more details.
 
+## Inlay Hints
+
+The style for inlay hints is defined in a [inlay_hints.css](https://github.com/sublimelsp/LSP/blob/main/inlay_hints.css) file in the root directory of the LSP package.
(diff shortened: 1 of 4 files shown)


rwols (approved):

predragnikolic:
Can you also add that to the the docs? (preferably to the customization section
https://lsp.sublimetext.io/customization/)

PR #2234, opened 2023-04-06
adding outline as an option to drawing document_highlight

TerminalFi:
image

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -305,6 +305,8 @@ class Settings:
     def highlight_style_region_flags(self, style_str: str) -> Tuple[int, int]:
         if style_str in ("background", "fill"):  # Backwards-compatible with "fill"
             return sublime.DRAW_NO_OUTLINE, sublime.DRAW_NO_OUTLINE
+        elif style_str == "outline":
+            return sublime.DRAW_NO_FILL, sublime.DRAW_OUTLINED | sublime.DRAW_NO_FILL
         elif style_str == "stippled":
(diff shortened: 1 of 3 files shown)


plugin/core/types.py
```diff
@@ -305,6 +305,8 @@ def r(name: str, default: Union[bool, int, str, list, dict]) -> None:
     def highlight_style_region_flags(self, style_str: str) -> Tuple[int, int]:
         if style_str in ("background", "fill"):  # Backwards-compatible with "fill"
             return sublime.DRAW_NO_OUTLINE, sublime.DRAW_NO_OUTLINE
+        elif style_str == "outline":
+            return sublime.DRAW_NO_FILL, sublime.DRAW_OUTLINED | sublime.DRAW_NO_FILL
```
jwortmann, line 309: deprecated

I think we should follow that suggestion. Also `32 | 32` is just `32`, so you can omit the first part `sublime.DRAW_OUTLINED | ` of the second return argument.

Should probably also update the settings file with the new value
https://github.com/sublimelsp/LSP/blob/43d00544f988f4a415f8fc9569436ff4eb7764d3/LSP.sublime-settings#L169

jwortmann:
I've noticed that the optional hover region highlights also use this same function. So could you add the "outline" option for the "hover_highlight_style" setting description and the JSON schema as well?

jwortmann (approved):

predragnikolic (approved):

PR #2237, opened 2023-04-17
Fix handling of folder_exclude_patterns in projects

### LINKED ISSUE #2171: Missing diagnostics due to "folder_exclude_patterns" with multiple project folders
**Describe the bug**
A recent change has caused error messages to no longer get highlighted in my project. After some fiddling (with help on Discord) I figured out that I had a `"folder_exclude_patterns"` set in my Project settings. This was for a subfolder that was added as a separate project folder:

```json
{
    "folders":
    [
        {
            "path": "my-project",
            "folder_exclude_patterns": ["frontend"],
        },
        {
            "path": "my-project/frontend",
            "name": "my-project/frontend",
        },
    ],
    // ...
}
```
As a result, errors reported by LSP-typescript and LSP-eslint within the `my-project/frontend` directory were being ignored and not rendered in my editor.

Note: I added the redundant frontend project folder a while ago because LSP-eslint needs the .eslintrc file to be found inside a project folder root. I then added the `"folder_exclude_patterns"` so I don't get duplicate entries in the ctrl+p file search.

**Expected behavior**
Errors should be reported for any folder listed in the project folders setting.

- OS: Linus Mint 21.1
 - LSP: 1.21.0
 - Sublime Text: 4143
 - eslint, typescript (also pyright and a few others but not in this sub project)

rchl:
ProjectFolders now holds a mapping of `folder_exclude_patterns` matching each of the workspace `folders`. This way we can do more targeted check to know whether path is not ignored in at least one of the folders.

I think this logic might be a bit hard to wrap the head around and not sure if the design of that fix is good.

Fixes #2171

### DIFF
--- a/plugin/core/workspace.py
+++ b/plugin/core/workspace.py
@@ -69,7 +69,6 @@ class ProjectFolders(object):
             return
         for i, folder in enumerate(project_data.get('folders', [])):
             exclude_patterns = []
-            self._folders_exclude_patterns.append(exclude_patterns)
             # Use canoncial path from `window.folders` rather than potentially relative path from project data.
             path = folders[i]
             for pattern in folder.get('folder_exclude_patterns', []):
@@ -80,6 +79,7 @@ class ProjectFolders(object):
                 else:
                     exclude_patterns.append(sublime_pattern_to_glob('//' + pattern, True, path))
                     exclude_patterns.append(sublime_pattern_to_glob('//**/' + pattern, True, path))
+            self._folders_exclude_patterns.append(exclude_patterns)
 
     def update(self) -> bool:
         new_folders = self._window.folders()
--- a/plugin/core/workspace.py
+++ b/plugin/core/workspace.py
@@ -106,7 +106,7 @@ class ProjectFolders(object):
                 continue
             exclude_patterns = self._folders_exclude_patterns[i]
             is_excluded = matches_pattern(file_path, exclude_patterns)
-            if is_excluded:
+            if not is_excluded:
                 break
         return is_excluded
 
--- a/plugin/core/workspace.py
+++ b/plugin/core/workspace.py
@@ -1,5 +1,7 @@
 from .protocol import WorkspaceFolder as LspWorkspaceFolder
 from .types import diff
+from .types import matches_pattern
+from .types import sublime_pattern_to_glob
 from .typing import Any, List, Union
 from .url import filename_to_uri
 import sublime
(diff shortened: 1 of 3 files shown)


plugin/core/workspace.py
```diff
@@ -56,9 +58,32 @@ class ProjectFolders(object):
...
+    def _update_exclude_patterns(self, folders: List[str]) -> None:
+        self._folders_exclude_patterns = []
+        project_data = self._window.project_data()
+        if not isinstance(project_data, dict):
+            return
+        for i, folder in enumerate(project_data.get('folders', [])):
+            exclude_patterns = []
+            self._folders_exclude_patterns.append(exclude_patterns)
+            # Use canoncial path from `window.folders` rather than potentially relative path from project data.
+            path = folders[i]
```
jwortmann, line 74: I haven't tested it, but what is this `folders` variable? Should it be `self.folders`?

Also could you move the `self._folders_exclude_patterns.append(exclude_patterns)` line to the end; I'm always unsure for which data structures Python treats variables by reference if you modify them later, and reordering those lines would make the logic clearer.

jwortmann:
Could you explain in a nutshell how does this fix the linked issue? From a quick look, it seems to me that it does the same as before, just with additional caching instead of calling the API functions each time. Because in the `includes_excluded_path` function, the loop still `break`s when a pattern matches and then returns `True`. The bug report is about the situation when a pattern matches for a particular folder, but in the project data this same folder is listed another time again later, so that we actually would need to ignore this match. (I might be wrong because I haven't tested)

plugin/core/workspace.py
rchl: `folders` is `self.folders` but it's passed as an argument to this function rather than accessed directly as I feel that improves the clarity in regards to the timing/order of when `self.folders` is updated relative to these patterns being update. In other words, if this code would access `self.folders` directly then I feel it would be easier to make a mistake of running this code before `self.folders` is updated.

rchl:
> Could you explain in a nutshell how does this fix the linked issue? From a quick look, it seems to me that it does the same as before, just with additional caching instead of calling the API functions each time. Because in the `includes_excluded_path` function, the loop still `break`s when a pattern matches and then returns `True`. The bug report is about the situation when a pattern matches for a particular folder, but in the project data this same folder (or subfolder) is listed another time again later, so that we actually would need to ignore this match. (I might be wrong because I haven't tested)

There would ideally be tests for this and those should make it clearer what the intentions are... I think it would be easy to make low-level unit tests but I would prefer higher level end to end tests but not sure if it's possible to mock `window.project_data()`.

That said, you are right that this logic was incorrect. I did accidentally reverse the condition when doing last minute refactorings. Should be better now.

But in a nutshell:
 - if the file is not contained within project folders (because there are no folders or just because it's not part of existing folders) then we don't ignore diagnostics
 - if there are project folders and the file is within one or more of those folders then it must not be excluded in at least one of those folders for the diagnostics to not be ignored.

(I wish I would find a better way to describe it as it doesn't sound very clear even to me, when I'm writhing it ;)).

jwortmann (approved):
Okay, I think the logic is correct now.

At a later point, we could try to convert this function into an `is_visible_in_sidebar`, by also honoring `file_exclude_patterns`, `file_include_patterns` and `folder_include_patterns` from the project settings. But all this path handling and conditional logic is quite error prone and no fun to implement, so I will probably skip on that... ;)

PR #2242, opened 2023-04-26
Add "Quick Add Next" command

jwortmann:
Here is another experiment of mine. This adds replacements for ST's "Quick Add Next" (<kbd>Ctrl</kbd>+<kbd>D</kbd>) and "Quick Skip Next" (<kbd>Ctrl</kbd>+<kbd>K</kbd>, <kbd>Ctrl</kbd>+<kbd>D</kbd>) commands. This was inspired by a recent issue https://github.com/sublimehq/sublime_text/issues/5945, especially the question for a potential workaround https://github.com/sublimehq/sublime_text/issues/5945#issuecomment-1522431972. I think LSP would be in a good position to provide this feature with help of the "Document Highlights", which are in most cases used to highlight tokens which are identical to the one under the caret.

Besides the `$` bug of the built-in command (see issue link), there is another advantage from LSP; because a language server has knowledge about the scope of a variable, it can highlight only the relevant occurrences, even if the same identifier name is used again elsewhere. For example if you want to quickly rename a local variable (without using the rename dialog), you can use this command to quickly select only the relevant occurrences of this variable.

Example with LSP-typescript:
```js
function foo() {
    let $getNodeByKey = 1;

    $getNodeByKey++;

    console.log($getNodeByKey);
}



function bar() {
    let $getNodeByKey = 1;

    $getNodeByKey++;

    console.log($getNodeByKey);
}
```

[quick_add_next.webm](https://user-images.githubusercontent.com/6579999/234655080-a3147c69-269b-4122-807a-bdcc5192bca0.webm)

I added a context to the key binding, so it falls back to the built-in command if the caret is not within a highlighted region (e.g. on a keyword or something else). At the moment the context & key binding isn't active if there are multiple carets in the view, and the first one is on a highlighted region, but the last one is not. The reason for this is, that our "Document Highlights" take only the *first* selection into account, which is conflicting with this feature.

PR is draft for now, let me know what you think, if this command is useful or not.

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -213,19 +213,19 @@
     //     "command": "lsp_expand_selection",
     //     "context": [{"key": "lsp.session_with_capability", "operand": "selectionRangeProvider"}]
     // },
-    // Quick Find All (an alternative for ST's "Quick Find All", using only the related highlights)
+    // Quick Find All (an alternative to ST's "Quick Find All", using only the related highlights)
     // {
     //     "keys": ["alt+f3"],  // macOS: "ctrl+super+g"
     //     "command": "lsp_find_all_under",
     //     "context": [{"key": "lsp.in_highlight_region"}]
     // },
-    // Quick Add Next (an alternative for ST's "Quick Add Next", using only the related highlights)
+    // Quick Add Next (an alternative to ST's "Quick Add Next", using only the related highlights)
     // {
     //     "keys": ["primary+d"],
     //     "command": "lsp_find_under_expand",
     //     "context": [{"key": "lsp.in_highlight_region"}]
     // },
-    // Quick Skip Next (an alternative for ST's "Quick Skip Next", using only the related highlights)
+    // Quick Skip Next (an alternative to ST's "Quick Skip Next", using only the related highlights)
     // {
     //     "keys": ["primary+k", "primary+d"],
     //     "command": "lsp_find_under_expand_skip",
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -214,11 +214,11 @@
     //     "context": [{"key": "lsp.session_with_capability", "operand": "selectionRangeProvider"}]
     // },
     // Quick Find All (an alternative for ST's "Quick Find All", using only the related highlights)
-    {
-        "keys": ["alt+f3"],  // macOS: "ctrl+super+g"
-        "command": "lsp_find_all_under",
-        "context": [{"key": "lsp.in_highlight_region"}]
-    },
+    // {
+    //     "keys": ["alt+f3"],  // macOS: "ctrl+super+g"
+    //     "command": "lsp_find_all_under",
+    //     "context": [{"key": "lsp.in_highlight_region"}]
+    // },
     // Quick Add Next (an alternative for ST's "Quick Add Next", using only the related highlights)
     // {
     //     "keys": ["primary+d"],
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -213,13 +213,19 @@
     //     "command": "lsp_expand_selection",
     //     "context": [{"key": "lsp.session_with_capability", "operand": "selectionRangeProvider"}]
     // },
-    // Quick Add Next (a replacement for ST's "Quick Add Next")
+    // Quick Find All (an alternative for ST's "Quick Find All", using only the related highlights)
+    {
+        "keys": ["alt+f3"],  // macOS: "ctrl+super+g"
+        "command": "lsp_find_all_under",
+        "context": [{"key": "lsp.in_highlight_region"}]
+    },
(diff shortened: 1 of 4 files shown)


rchl:
My first thoughts are a bit conflicting because this is overriding ST's feature but rather than just expanding it to work better, it changes quite significantly the original logic (since it works in semantic way).

Also it seems to have equivalent use case as "rename symbol" so I wonder if it's worth having just for the sake of providing slightly different interface.

But assuming that it's useful to some people or in specific cases, I wonder if it should really be bound to the same keys as the original feature. I know it's just a suggestion but I don't know if it's a good suggestion...

That said, I don't want to discourage you from working on it. I'm just a bit conflicted internally and maybe I don't quite see the appeal that is there.

rchl:
Also... I wonder if it would be enough to provide a command to select all instance of the current symbol rather than supporting "add" and "skip".

Slapbox:
I'm not in a position to judge the right way to fix #5945, but it's killing me right now.

jwortmann:
> I'm not in a position to judge the right way to fix #5945, but it's killing me right now.

@Slapbox Well, this is not gonna fix the core issue, but would just be a possible workaround for users of the LSP package. The reported issue is clearly a bug in Sublime, but nobody knows whether or when it will be fixed. If you need a simple temporary solution for now, open a JavaScript and/or TypeScript file, run *Preferences: Settings - Syntax Specific* from the command palette, and add
```jsonc
{
    "word_separators": "./\\()\"'-:,.;<>~!@#$%^&*|+=[]{}`~?",
}
```
to revert the removal of the dollar from the word separators in the default setting for those languages (e.g. https://github.com/sublimehq/Packages/blob/master/JavaScript/JavaScript.sublime-settings).
This way the leading `$` will not be selected with <kbd>Ctrl</kbd>+<kbd>D</kbd> or when you double click with the mouse, but at least the subsequent <kbd>Ctrl</kbd>+<kbd>D</kbd> actions should work.

---

> I wonder if it should really be bound to the same keys as the original feature

Well, the key bindings are commented out by default, so it's up to the user to assign keys anyway. So this PR would not result in a surprising behavior change by default, but just give the ability of using the semantic highlighting regions from the server if so desired. I think that I would personally probably activate the bindings, and in that case I would certainly use the default keys for it, rather than assign and learn entirely new keys.

I agree that it is indeed a behavior change when activating the key bindings, but maybe it's an advantage after you get used to it. I wonder whether ST would maybe even have implemented this feature in a similar way, if they had the specific scope boundary information of identifiers, that language servers can have.

But the "select all highlights" is not a bad idea too. There is also a built-in "Quick Find All" command for that (<kbd>Alt</kbd>+<kbd>F3</kbd> on Windows), which does this in the "stupid" text-based way over the whole file. Regarding to LSP's "rename symbol", I must say that I personally use that extremely rarely, and usually select the instances one by one with <kbd>Ctrl</kbd>+<kbd>D</kbd>. Maybe I think I'm faster with that, or I don't trust the LSP server to not do something stupid. At least for simple renaming of variables within a function or class and not over multiple files.

If anyone else reads this and has an opinion whether the commands would be useful, any comments are welcome. But we can also leave it at the "experiment" status and close the PR if there is no demand for it.

Slapbox:
Many thanks for the tip about the workaround @jwortmann - that should hold us over!

rchl:
I've started testing it today and pretty quickly noticed an issue with code like below in typescript where I've expected to be able to select both FOO's without creating selection prior but it didn't allow me because it's technically not the same symbol.

```ts
export const FOO = 'FOO';
```

rchl:
Also trying to workaround it by creating selection manually first doesn't really work, depending if you make selection from the left or right of the symbol.

jwortmann:
> noticed an issue with code like below in typescript where I've expected to be able to select both FOO's without creating selection prior but it didn't allow me because it's technically not the same symbol

I mean it should be pretty clear what it does, but if you expect both to be the same symbol, then I would file that as an issue in typescript-language-server, no? I'm not too familiar with TypeScript, but if it were e.g. Python code, I would *not* expect the command to select both, because the left one looks like a (constant) variable, and the right one looks like a string to me. So in Python, if I would want to rename the variable, then I don't want the string to change as well.

jwortmann:
This example is by the way the biggest drawback of the built-in command from ST, i.e. that it selects words from strings or comments too. But it depends on the situation whether that is wanted, so maybe that would be an argument to indeed use different key bindings for the new commands.

I was not really aware of the behaviour change of the built-in command depending on whether you manually created a selection (of the whole word) first, or whether the selection was created by running the command with a normal caret. That means that the built-in command internally saves state between subsequent runs and always has knowledge in which way it was executed in the very first run. I'm not sure whether it would be useful to do something similar for the new commands from this PR as well, but I can think about it.

jwortmann:
I think I'll close this PR for now, because there seems to be little demand for it, and the behavior is perhaps too unexpected in comparison with the built-in functionality.

PR #2247, opened 2023-05-08
add "show_multiline_document_highlights" setting

### LINKED ISSUE #2246: ability to disable document_highlight_style on multiline string blocks
Unless Im mistaken, it seems to me that `document_highlight_style` is causing multiline string to have a full border, which is very distracting. 

image

I have searched and couldn't find a way to disable highlighting for multiline strings. Could be possible to add a preference to avoid this?

titoBouzout:
https://github.com/sublimelsp/LSP/issues/2246


LSP.sublime-settings
```diff
@@ -172,6 +172,9 @@
   // customize the color.
   "document_highlight_style": "underline",
 
+  // Enable highlighting on multiline blocks
+  "document_highlight_multiline": true,
```
rchl, line 176: There is `show_multiline_diagnostics_highlights` already so it would be good to keep the naming consistent with it - `show_multiline_document_highlights`.

LSP.sublime-settings
```diff
@@ -172,6 +172,9 @@
   // customize the color.
   "document_highlight_style": "underline",
 
+  // Enable highlighting on multiline blocks
```
rchl, line 175:
```suggestion
  // Show document highlights spanning multiple lines in the view (as outlines).
  // See also: "document_highlight_style".
```

rchl (changes requested):

rchl (approved):

PR #2250, opened 2023-05-09
Add "diagnostics_ignored_messages" setting

titoBouzout:
This pull request adds a setting to ignore specific diagnostic messages.

I have a problem with TypeScript defaulting to `React` (thing I'm not even using) which is very difficult to change. LSP-typescript has a helpful option to ignore messages by code, that is useful for some very picky codes, but that would imply to silent every error of the same kind which I don't want. What I want is to ignore a very specific message.  This already proved to be useful by silencing errors that I'm not interested in accentuating errors of the same kind on which I am interested in.


predrag-codetribe:
Have you tried using the `// @ts-expect-error` comment?
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments

I am not in favour of this PR,
IMO it makes more sense to add a ignore_diagnostics_code per client configuration setting,
than a global `diagnostics_ignored_messages`

Also you have described a solution for a problem, but not the cause of the problem itself.
Which makes me think that the problem might be fixed on a different way...
There are probably reasons why TS reports such errors, 
silencing them in LSP will not silence them in the ts compiler.

Can you explain the exact problem you are facing in typescript?

predrag-codetribe:
Try adding the following to your tsconfig.
```
{
  "compilerOptions": {
    "jsx": "preserve",
    "jsxImportSource": "solid-js",
    "noEmit": true,
  }
}
```

predrag-codetribe:
> I'm not even interested to make TS happy or fix this error anymore.

But resolving the TS should be the correct way to go.

That said, I am not if favour of adding new settings unless necessary.
I do not see this as necessary.
But I am interested and happy to help in resolving the typescript issue.

titoBouzout:
@rchl 

> It's limited - only filters based on message and only based on full message. Also it's global while it seems like something that should be server-specific.

Yes, the idea is to ignore a specific message. Do you have any hint how to make this server specific? I can move the preference to be server specific, but I don't know where the relevant bits of that code exists. Linking to an example may be enough.

> It's also not efficient since it has to compare message of all diagnostics and there might be hundreds in some cases.
EDIT: OK, the fact that it compares whole message makes it limited but at least it should be pretty efficient.

> (Also not sure why it would be exposed to the user as a dict instead of a list but that's not relevant given the above.)

It is a dict, so it's as efficient as possible without any pre-processing.  Not sure how to be more efficient than that without using something like a database.

> Also note that we don't take adding new settings very lightly as it's an additional maintenance burden. So there has to be really good reasons for adding one.

That's fine, per your considerations.  On the previous message, I think I gave good reasons. The fact that there are error codes, you can silence, makes me think that silencing a very specific error would be welcome, and not being able to do so could be considered a bit hypocritical(given that silencing error codes is a thing).

@predrag-codetribe Thanks for your interest, I have tried that and many other combinations and I still see that diagnostic message everywhere. Please understand this issue is about adding a new preference and not fixing my error. As I said before, I tried everything and I am not interested any more into fixing it. It is not my job to make TS happy, as I don't even use it.

predrag-codetribe:
My point is that this specific issue can be solved without a PR on the LSP side.
If you create a minimal repository, I could take a look.

titoBouzout:
I appreciate a lot your motivation to help me deal with that particular thing, but is not worth our time. I don't use typescript, nor I want to add packages or files that nothing had to do with my project just to make typescript happy.  I understand the surge to make typescript happy when you actually use typescript, for me is just a linter for my js files.

rchl:
If there exists a solution to this that is specific to the given server (typescript in this case), and I believe there is in this case, then there is even less (could even say "no") motivation for this change.

titoBouzout:
I mean, the solution given by the server is to get rid of an complete `error type`. Like, let's say `a variable is not defined`. So if you get rid of that error code, then you will never again get diagnostic messages for any kind of undefined vars. That's not the solution to this problem. I want to know when vars are not defined. I just don't want to know when `React` is not defined.

rchl:
I meant what @predragnikolic said -- it should be possible for this specific error to go away by configuring server through tsconfig.

If you want to help us help you then provide a repo that reproduces and we might spend our time helping you.

Alternatively the LSP-typescript (or rather the underlying LSP server) could provide more filtering options since it already provides a way to ignore by error code. But be aware that you'll have to deal with me again there ;)

titoBouzout:
Lol, its cool, 

> I meant what @predragnikolic said -- it should be possible for this specific error to go away by configuring server through tsconfig.

I see, I tried both configurations that @predrag-codetribe kindly provided and these won't work, I tried many other things too, that's why I invested time into this pull request. 

If you're up for a challenge, just create a file named `test.js`  with content `export function test() {return <div />}`. Then try to get rid of the error on the `div` that says `'React' is not defined.`. Just remember, I don't use typescript, and I'm not interested into converting my project to typescript to make it happy. For me is a linter via LSP **not** a compiler. This should be ideally fixed by a possible `tsconfig.json` file, if includes any other kind of change, then that's already too much.

> Alternatively the LSP-typescript (or rather the underlying LSP server) could provide more filtering options since it already provides a way to ignore by error code. But be aware that you'll have to deal with me again there ;)

I think it would be a cool idea to allow people to do it by server but implemented in the core, so every server won't have to code the same thing.

rchl:
Not sure what do you mean by "I don't use typescript". This error comes from typescript so you are using it (regardless if only for linting) and the way to configure it is through `tsconfig.json`.

It seems like you want to on purpose ignore configuration provided by the server and that's IMO not a compelling reason for introducing a new option here. Also many servers already provide ways to ignore their diagnostics through some means. For example eslint, pyright through comments in the code. Typescript in fact also does through `@ts-ignore`/`@ts-nocheck` comments (even if those might not help in that particular case).

titoBouzout:
> It seems like you want to on purpose ignore configuration provided by the server and that's IMO not a compelling reason for introducing a new option here. 

Mmh, the server configuration is not being ignored. There's no configuration on that server to get rid of "React is not defined" while being alerted at the same time of "SomethingElse is not defined". Unless you are referring to something else that I am not aware.

rchl:
Well, we've asked about repro. Can't say more without it. Just creating a file with `export function test() {return <div />}` doesn't reproduce. Might depend on installed dependencies or the existing `tsconfig.json`, if you have one.

titoBouzout:
If you wait for the LSP server to kick, it should say React is not defined because that's the default.

Here is a repro: 
[tsconfig.zip](https://github.com/sublimelsp/LSP/files/11441805/tsconfig.zip)

rchl:
I honestly don't see any error in test.js.

Can you run `LSP: Troubleshoot server` on this file? There might be other settings affecting what you see.

rchl:
No issues:

Screenshot 2023-05-10 at 14 48 56

predrag-codetribe:
Initally I haven't been able to reproduce it.

But I managed to preproduce it with the following LSP-typescript.sublime-settings file.
```
{
  "settings": {
    "diagnostics": {
      "ignoredCodes": [
        7044,
        2307,
        7045,
        2531,
        80002,
        7047,
        2790,
        17001,
        7016
      ]
    },
    "implicitProjectConfiguration": {
      "checkJs": true,
      "experimentalDecorators": false,
      "module": "ESNext",
      "strictFunctionTypes": false,
      "strictNullChecks": false,
      "target": "ES2020"
    },
    "javascript": {
      "format": {
        "insertSpaceAfterCommaDelimiter": true,
        "insertSpaceAfterConstructor": false,
        "insertSpaceAfterFunctionKeywordForAnonymousFunctions": true,
        "insertSpaceAfterKeywordsInControlFlowStatements": true,
        "insertSpaceAfterOpeningAndBeforeClosingEmptyBraces": false,
        "insertSpaceAfterOpeningAndBeforeClosingJsxExpressionBraces": false,
        "insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces": true,
        "insertSpaceAfterOpeningAndBeforeClosingNonemptyBrackets": false,
        "insertSpaceAfterOpeningAndBeforeClosingNonemptyParenthesis": false,
        "insertSpaceAfterOpeningAndBeforeClosingTemplateStringBraces": false,
        "insertSpaceAfterSemicolonInForStatements": true,
        "insertSpaceAfterTypeAssertion": false,
        "insertSpaceBeforeAndAfterBinaryOperators": true,
        "insertSpaceBeforeFunctionParenthesis": false,
        "insertSpaceBeforeTypeAnnotation": false,
        "placeOpenBraceOnNewLineForControlBlocks": false,
        "placeOpenBraceOnNewLineForFunctions": false,
        "semicolons": "ignore",
        "trimTrailingWhitespace": true
      },
      "inlayHints": {
        "includeInlayEnumMemberValueHints": false,
        "includeInlayFunctionLikeReturnTypeHints": false,
        "includeInlayFunctionParameterTypeHints": false,
        "includeInlayParameterNameHints": "none",
        "includeInlayParameterNameHintsWhenArgumentMatchesName": false,
        "includeInlayPropertyDeclarationTypeHints": false,
        "includeInlayVariableTypeHints": false,
        "includeInlayVariableTypeHintsWhenTypeMatchesName": false
      }
    },
    "statusText": "$version, $source",
    "typescript": {
      "format": {
        "insertSpaceAfterCommaDelimiter": true,
        "insertSpaceAfterConstructor": false,
        "insertSpaceAfterFunctionKeywordForAnonymousFunctions": true,
        "insertSpaceAfterKeywordsInControlFlowStatements": true,
        "insertSpaceAfterOpeningAndBeforeClosingEmptyBraces": false,
        "insertSpaceAfterOpeningAndBeforeClosingJsxExpressionBraces": false,
        "insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces": true,
        "insertSpaceAfterOpeningAndBeforeClosingNonemptyBrackets": false,
        "insertSpaceAfterOpeningAndBeforeClosingNonemptyParenthesis": false,
        "insertSpaceAfterOpeningAndBeforeClosingTemplateStringBraces": false,
        "insertSpaceAfterSemicolonInForStatements": true,
        "insertSpaceAfterTypeAssertion": false,
        "insertSpaceBeforeAndAfterBinaryOperators": true,
        "insertSpaceBeforeFunctionParenthesis": false,
        "insertSpaceBeforeTypeAnnotation": false,
        "placeOpenBraceOnNewLineForControlBlocks": false,
        "placeOpenBraceOnNewLineForFunctions": false,
        "semicolons": "ignore",
        "trimTrailingWhitespace": true
      },
      "inlayHints": {
        "includeInlayEnumMemberValueHints": false,
        "includeInlayFunctionLikeReturnTypeHints": false,
        "includeInlayFunctionParameterTypeHints": false,
        "includeInlayParameterNameHints": "none",
        "includeInlayParameterNameHintsWhenArgumentMatchesName": false,
        "includeInlayPropertyDeclarationTypeHints": false,
        "includeInlayVariableTypeHints": false,
        "includeInlayVariableTypeHintsWhenTypeMatchesName": false
      }
    }
  },
  "initializationOptions": {
    "preferences": {
      "importModuleSpecifierEnding": "js"
    }
  }
}
```

Now I only need to find the cause :)

rchl:
I don't remember exact logic but I thought that `implicitProjectConfiguration` should only apply when there is no tsconfig.

In any case, I think it works when specifying `"checkJs": true,` explicitly in tsconfig.

rchl:
I can reproduce same in VSCode. Setting the "js/ts.implicitProjectConfig.checkJs": true` option reproduces but setting `checkJs` in `tsconfig` doesn't.

Maybe there is some good reason for that but I guess the next step would be to search the typescript repo for similar issue or report it there.

predrag-codetribe:
Yes, specifying the `"checkJs": true,` fixed the issue. @titoBouzout 
```diff
{
  // https://github.com/sublimelsp/LSP/pull/2250#issuecomment-1541973212
  "compilerOptions": {
    "strict": true,
+    "checkJs": true,
    "target": "ESNext",
    "module": "ESNext",
    "moduleResolution": "node",
    "allowSyntheticDefaultImports": true,
    "esModuleInterop": true,
    "jsx": "preserve",
    "jsxImportSource": "solid-js",
    "types": ["vite/client"],
    "noEmit": true,
    "isolatedModules": true
  }
}
```

--- 

But I have noticed something off in your LSP-typescript.sublime settings file.
@titoBouzout, did you maybe manually edited the **left** LSP-typescript.sublime-settings file? instead of overriding the values on in the right file?

I noticed in the logs, this:
```
 - settings

```json
  //...
  "implicitProjectConfiguration": {
    "checkJs": true, /// <- this here should be false, unless you mistakenly edited the left LSP-typescript.sublime-settings file.
```

The default value for `checkJs` should be false, 
but in the troubleshoot.txt file I saw it being true, although you have not set it to true in you LSP-typescript.sublime-settings override file.

Screenshot 2023-05-10 at 16 18 12

Please check if your left LSP-typescript.sublime-settings file looks like [this](https://github.com/sublimelsp/LSP-typescript/blob/master/LSP-typescript.sublime-settings)

predragnikolic:
True, my bad, haven't seen it.

titoBouzout:
Well, mystery solved, that fix it for some _weird reason_, much appreciated thanks for that!  :) This change introduces a number of oddities on my project, so I will have to study that.

Kind unfortunate that the example to add this setting happens to be TypeScript. People are addicted to fix TS things, even when sometimes is unfixable.  I can imagine a variety of reasons when this setting could help:
1. unable to fix upstream, upstream code is unmaintained or author unreachable, not wanting to fix upstream, types in upstream are bogus
2. unable to upgrade something because breaks xyz
3. not wanting to introduce ts stuff in a js project
4. not willing to add comments to ignore a message in a million of lines/files
5. keeping the list of ignored messages private
6. It could be updated to be a regular expression or whatever
7. alternatives are nice
8. because reasons

I personally will keep this code around and see how it goes~ I will leave to your consideration the pull request

rchl:
Sorry but there is no convincing for adding it.

We are not in favor of adding new options that are potentially useful to only a fraction of users.

PR #2254, opened 2023-05-13
update note about custom color scheme rule used for diagnostics

### LINKED ISSUE #2253: Customizing style of diagnostics with tags no longer works in dev builds
**Describe the bug**

This approach:

https://github.com/sublimelsp/LSP/blob/b49da9e81a8e6096509a24e6b98a4d6897f41c05/plugin/session_view.py#L298-L299

no longer works in current ST dev builds due to this "issue":  https://github.com/sublimehq/sublime_text/issues/5942

I don't remember if there is a reason that we are not adding regions if custom color scheme rule is not defined...
Is it just a minor optimization and it would be fine to always do it? @jwortmann ?

rchl:
Fixes #2253

### DIFF
--- a/docs/src/customization.md
+++ b/docs/src/customization.md
@@ -37,7 +37,7 @@ If you want to bind some action to a mouse, open `Preferences / Browser Packages
 | Linux    | `/User/Default (Linux).sublime-mousemap` |
 | Mac      | `/User/Default (OSX).sublime-mousemap` |
 
-Here is an example mouse binding that triggers LSP's "go to symbol definition" command on pressing the <kbd>ctrl</kbd>+<kbd>left click</kbd>:
+Here is an example of mouse binding that triggers LSP's "go to symbol definition" command on pressing the <kbd>ctrl</kbd>+<kbd>left click</kbd>:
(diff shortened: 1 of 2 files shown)


docs/src/customization.md
```diff
@@ -37,7 +37,7 @@ If you want to bind some action to a mouse, open `Preferences / Browser Packages
 | Linux    | `/User/Default (Linux).sublime-mousemap` |
 | Mac      | `/User/Default (OSX).sublime-mousemap` |
 
-Here is an example mouse binding that triggers LSP's "go to symbol definition" command on pressing the <kbd>ctrl</kbd>+<kbd>left click</kbd>:
+Here is an example of mouse binding that triggers LSP's "go to symbol definition" command on pressing the <kbd>ctrl</kbd>+<kbd>left click</kbd>:
```
jwortmann, line 40: English is not my first language, but "example of mouse binding" sounds really weird to me. Shouldn't it be "example of a mouse binding", or alternatively leave it as it was before?

jwortmann (approved):

PR #2256, opened 2023-05-14
Fix package storage path in a docstring description

jwortmann:
The `%AppData%/Sublime Text/Roaming` path from the docstring of the `get_storage_path` function doesn't exist.

Could someone confirm whether the paths on macOS and Linux are correct?

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -311,7 +311,7 @@ def get_storage_path() -> str:
 
     - on macOS: ~/Library/Application Support/Sublime Text
     - on Windows: %LocalAppData%/Sublime Text
-    - on Linux: $XDG_CONFIG_DIR/sublime-text
+    - on Linux: ~/.cache/sublime-text
     """
     return os.path.abspath(os.path.join(sublime.cache_path(), "..", "Package Storage"))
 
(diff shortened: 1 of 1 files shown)


predragnikolic:
The path on Linux for package Storage is  `~/.cache/sublime-text/Package\ Storage`

plugin/core/views.py
```diff
@@ -310,7 +310,7 @@ def get_storage_path() -> str:
     Its path is '$DATA/Package Storage', where $DATA means:
 
     - on macOS: ~/Library/Application Support/Sublime Text
-    - on Windows: %AppData%/Sublime Text/Roaming
+    - on Windows: %LocalAppData%/Sublime Text
     - on Linux: $XDG_CONFIG_DIR/sublime-text
```
predragnikolic, line 314:
```suggestion
    - on Linux: ~/.cache/sublime-text
```

rchl: When googling for `XDG_CONFIG_DIR` it doesn't seem to exist. There is `XDG_CONFIG_DIRS` and there is XDG_CONFIG_HOME.

Is `XDG_CONFIG_HOME` defined in your distro? I would imagine that `XDG_CONFIG_HOME` is some standard thing that would work in more cases.

predragnikolic: It is not defined on my distro (Ubuntu 22.10)
```
λ  ~  echo $XDG_CONFIG_DIR

λ  ~  echo $XDG_CONFIG_HOME

```

rchl: Actually in my virtual Debian the `XDG_CONFIG_DIRS` seems to be defined but points to `/etc/xdg`.
The `XDG_CACHE_HOME` or `XDG_CONFIG_HOME` are not defined...

Let's just use your suggestion :)

PR #2257, opened 2023-05-14
use class for diagnostic info instead of hardcoding color

### LINKED ISSUE #2255: How can I change the color of this?
I didn't find anything in your documentation to change this part in white, or maybe I am blind?
It's annoying because it's unreadable, I would like to keep the background yellow for the warnings.

unreadable

Here is my mdpopups.css file:
```css
html.dark {
    background-color: #252525;
    color: #9f9f9f;
}

.dark .mdpopups .highlight {
    background-color: #0f0f0f;
}

html.light {
    background-color: #e5e5e5;
    color: #666666;
}

.light .mdpopups .highlight {
    background-color: #fff;
}

html {
    --mdpopups-font-mono: "Operator Mono Book";
}

.errors a,
.warnings a {
    display: block;
    color: #010192;
    margin-top: 0.5rem;
}

.lsp_popup .errors {
    border-width: 0;
    background-color: #fd3434;
    color: #0f0f0f;
    margin-bottom: 0.5rem;
    padding: 0.5rem;
}

.lsp_popup .warnings {
    border-width: 0;
    background-color: #feff60;
    color: #0f0f0f;
    margin-bottom: 0.5rem;
    padding-top: 0.5rem;
}

.lsp_popup .info {
    border-width: 0;
    background-color: #00bfff;
    color: #0f0f0f;
    margin-bottom: 0.5rem;
    padding-top: 0.5rem;
}

.lsp_popup .hints {
    border-width: 0;
    background-color: #f8f8f8;
    color: #0f0f0f;
    margin-bottom: 0.5rem;
    padding-top: 0.5rem;
}


``` 

Add an option to be able to modify this color if possible.

Thank you very much and again for your hard work 🙏.

rchl:
Use `.color-muted` class instead of hardcoding color when showing diagnostic info.

Changed opacity of `.color-muted` from `0.5` to `0.6`. Should still be fine and is more consistent with other places.

Fixes #2255

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -1008,7 +1008,8 @@ def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, bas
     if related_infos:
         info = "<br>".join(_format_diagnostic_related_info(config, info, base_dir) for info in related_infos)
         html += _html_element("pre", info, class_name="related_info", escape=False)
-    return _html_element("pre", html, class_name=DIAGNOSTIC_SEVERITY[diagnostic_severity(diagnostic) - 1][1])
+    severity_class = DIAGNOSTIC_SEVERITY[diagnostic_severity(diagnostic) - 1][1]
+    return _html_element("pre", html, class_name=severity_class, escape=False)
 
 
 def format_code_actions_for_quick_panel(
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -993,17 +993,14 @@ def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, bas
     code = diagnostic.get("code")
     source = diagnostic.get("source")
     if source or code is not None:
-        html += " "
+        meta_info = " "
         if source:
-            html += _html_element("span", source, class_name="color-muted")
+            meta_info += text2html(source)
         if code is not None:
-            html += _html_element("span", "(", class_name="color-muted")
             code_description = diagnostic.get("codeDescription")
-            if code_description:
-                html += make_link(code_description["href"], str(code))
-            else:
-                html += _html_element("span", str(code), class_name="color-muted")
-            html += _html_element("span", ")", class_name="color-muted")
+            meta_info += "({})".format(
+                make_link(code_description["href"], str(code)) if code_description else text2html(str(code)))
+        html += _html_element("span", meta_info, class_name="color-muted", escape=False)
     related_infos = diagnostic.get("relatedInformation")
     if related_infos:
         info = "<br>".join(_format_diagnostic_related_info(config, info, base_dir) for info in related_infos)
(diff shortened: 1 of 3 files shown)


rchl:
Also we had this existing issue where the "related info" did not have a line break before it in the diagnostic panel.

Before:

Screenshot 2023-05-14 at 22 28 34

After:

Screenshot 2023-05-14 at 22 27 27

The HTML in the quick panel preview seems to be more limited so applying styles from `popup.css` wouldn't help there. So added extra `<br>` instead. Hover popup still shows things correctly with extra BR:

Screenshot 2023-05-14 at 22 29 39

predragnikolic (approved):
OK, with this PR, someone will be able to override the styles to their liking


Tested :heavy_check_mark: 

before: 
Screenshot from 2023-05-16 00-00-46

after (I've overridden some CSS classes)
Screenshot from 2023-05-15 23-58-30

jwortmann:
The screenshot don't look right to me, why is there a linebreak before the link (and also before the closing `)` in the second screenshot)? Have you maybe applied a margin/padding to the `.color-muted` class in your custom rules, which could cause this?

I think instead of having the `<span class="color-muted">` tags around each separate token, it would be better to use only a single wrapper tag of this around the whole `Code(CodeDescription)` expression. This would prevent such oddities as seen above.

rchl:
I think the intentions of @predragnikolic were not to achieve something nice looking but just test that overrides work.
We can't prevent people from messing up stuff if they want to.

We could try to have one color-muted wrapper but there is an optional link inside that element so not sure how will it work.

rchl:
It seems to look the same when the link is also within the `.color-muted` class.

Before

Screenshot 2023-05-16 at 08 57 58

After

Screenshot 2023-05-16 at 09 06 36

predragnikolic:
> The screenshot don't look right to me, why is there a linebreak before the link (and also before the closing ) in the second screenshot)?

There was some custom CSS that caused such behavior :)

I still didn't try to make it pretty, but I just tested that it is possible to target the class name.

Screenshot from 2023-05-16 14-06-10


> I think the intentions of @predragnikolic were not to achieve something nice looking but just test that overrides work.

Yes

> We can't prevent people from messing up stuff if they want to.

And yes

plugin/core/views.py
```diff
@@ -980,39 +980,33 @@ def _format_diagnostic_related_info(
...
-            formatted.append(_with_color(str(code), "color(var(--foreground) alpha(0.6))"))
-        formatted.append(_with_color(")", "color(var(--foreground) alpha(0.6))"))
+        meta_info = " "
+        if source:
+            meta_info += text2html(source)
+        if code is not None:
+            code_description = diagnostic.get("codeDescription")
+            meta_info += "({})".format(
+                make_link(code_description["href"], str(code)) if code_description else text2html(str(code)))
+        html += _html_element("span", meta_info, class_name="color-muted", escape=False)
```
predragnikolic, line 1003: Maybe we should keep "color-muted", 
but add a `diagnostic-code-description` class name that people who want to override it can with that class?

```
class_name="diagnostic-code-description color-muted"
```

predragnikolic:
Looks good to me,
@jwortmann, wdyt?

PR #2259, opened 2023-05-16
Use regular font style in sighelp popup if already highlighted by color scheme

jwortmann:
The signature help popup uses bold & underlined font style for the active parameter, which looks a bit ugly in my opinion.
With this PR, it will keep the regular font style if the color scheme already defines different colors for active and non-active parameters. This is the case for example [in my own color scheme](https://github.com/jwortmann/brackets-color-scheme/blob/54c4e4a429b07f824af3c11098ee612caa596f8f/Brackets%20Dark.sublime-color-scheme#L2154-L2163), (though I will probably adjust the colors a bit if the PR gets merged).

sighelp

Also the number of `View.style_for_scope` API calls is reduced by storing the results into variables and avoiding redundant calls.

### DIFF
--- a/plugin/core/signature_help.py
+++ b/plugin/core/signature_help.py
@@ -68,7 +68,11 @@ class SigHelp:
         formatted = []  # type: List[str]
         if self.has_multiple_signatures():
             formatted.append(self._render_intro())
-        formatted.extend(self._render_label(view, signature))
+        self._function_color = view.style_for_scope("entity.name.function.sighelp.lsp")["foreground"]
+        self._active_parameter_color = view.style_for_scope("variable.parameter.sighelp.active.lsp")["foreground"]
+        self._inactive_parameter_color = view.style_for_scope("variable.parameter.sighelp.lsp")["foreground"]
+        self._emphasize_active_parameter = self._active_parameter_color == self._inactive_parameter_color
+        formatted.extend(self._render_label(signature))
         formatted.extend(self._render_docs(view, signature))
         return "".join(formatted)
 
@@ -96,7 +100,7 @@ class SigHelp:
             len(self._signatures),
         )
 
-    def _render_label(self, view: sublime.View, signature: SignatureInformation) -> List[str]:
+    def _render_label(self, signature: SignatureInformation) -> List[str]:
         formatted = []  # type: List[str]
         # Note that this <div> class and the extra <pre> are copied from mdpopups' HTML output. When mdpopups changes
         # its output style, we must update this literal string accordingly.
@@ -124,13 +128,13 @@ class SigHelp:
                     start += prev
                     end = start + len(rawlabel)
                 if prev < start:
-                    formatted.append(_function(view, label[prev:start]))
-                formatted.append(_parameter(view, label[start:end], i == active_parameter_index))
+                    formatted.append(self._function(label[prev:start]))
+                formatted.append(self._parameter(label[start:end], i == active_parameter_index))
                 prev = end
             if end < len(label):
-                formatted.append(_function(view, label[end:]))
+                formatted.append(self._function(label[end:]))
         else:
-            formatted.append(_function(view, label))
+            formatted.append(self._function(label))
         formatted.append("</pre></div>")
         return formatted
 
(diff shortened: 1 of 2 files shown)


plugin/core/signature_help.py
```diff
@@ -68,7 +68,11 @@ def render(self, view: sublime.View) -> str:
         formatted = []  # type: List[str]
         if self.has_multiple_signatures():
             formatted.append(self._render_intro())
-        formatted.extend(self._render_label(view, signature))
+        self._function_color = view.style_for_scope("entity.name.function.sighelp.lsp")["foreground"]
+        self._active_parameter_color = view.style_for_scope("variable.parameter.sighelp.active.lsp")["foreground"]
+        self._inactive_parameter_color = view.style_for_scope("variable.parameter.sighelp.lsp")["foreground"]
+        self._emphasize_active_parameter = self._active_parameter_color == self._inactive_parameter_color
```
rchl, line 74: The optimization better be worth it because I guess it will make it more difficult to experiment with different styles. :)

rchl (commented):
Is it expected that there is no change with Mariana?

Before

Screenshot 2023-05-16 at 09 16 21


After

Screenshot 2023-05-16 at 09 17 06

plugin/core/signature_help.py
rchl: Never mind, this just caches it for the life of the signature popup which disappears on switch views anyway.

jwortmann:
> Is it expected that there is no change with Mariana?

Yes, that is expected. An alternative to the bold & underline font style by default, would be to use a different scope for the active parameter color. But that would probably also complicate things, because you can't really be sure that it creates a sufficient contrast to the color for non-active parameters.

rchl (approved):

frou:
Has this change taken away the ability for someone to use custom sighelp colours **and** have bolding/underlining behaviour?

predragnikolic:
Yes

frou:
Thank you. To me, It's not all about contrast. The underline in particular made it very easy to pick out the active parameter at a glance.

> Then it would be the same default as it is now, but users could re-add the bold/underline via mdpopups.css if they want.

Maybe I'm misunderstanding the relative importance of the HTML/CSS aspect, but wouldn't it be more correct to delegate the bold/underline decision to the Colour Scheme?

i.e. a rule could be:

```json
{"scope": "variable.parameter.sighelp.active.lsp", "foreground": "var(green)", "font_style": "bold underline"}
```

and that desire for `bold`/`underline` would be reflected in the dict returned by `view.style_for_scope("variable.parameter.sighelp.active.lsp")`, just as the `foreground` colour already is.

jwortmann:
> wouldn't it be more correct to delegate the bold/underline decision to the Colour Scheme?

Actually not a bad idea. I think it would work, but we need to apply a little "trick" which is already used for the diagnostics `unnecessary` tag, to preserve the bold & underline when there is no color scheme rule for the sighelp-specific scopes (otherwise, the active parameters couldn't be distinguished at all anymore in the built-in and probably all third-party color schemes, because they don't have set bold & underline font style for those scopes by default). I think I can make a PR with that.

jwortmann:
The "trick" that I had in mind was actually used for the semantic tokens, and not for diagnostic tags.
https://github.com/sublimelsp/LSP/blob/a6edba3d254c25307f4818b34c84e5cf89e147d8/plugin/session_buffer.py#L679
I tried to implement using the font style from the color scheme for the active parameter, but there seems to be a bug in the `style_for_scope` API method, which makes parts of the results unreliable. In particular it doesn't seem to work correctly for the `variable.parameter.sighelp.active.lsp` scope when I tested it with my own color scheme. I filed a report on the ST issue tracker at https://github.com/sublimehq/sublime_text/issues/5983.

Unfortunately this probably means that we're stuck here in regard to implementing such a solution, unless someone has another idea how to retain the bold & underline if there is no color scheme rule for the scope. Alternatively we could use the approach of the CSS class for now.

frou:
Thanks for trying it. Let's see if that issue possibly gets a quick fix.

PR #2274, opened 2023-05-17
Fix tagged diagnostics flickering on document changes

### LINKED ISSUE #2272: erase_regions causes "markup.unnecessary.lsp" to flicker
**Describe the bug**

**To Reproduce**
1. Run Customize Color scheme from the command palette:
2. Paste this
```
{
    "rules": [
        {
            "scope": "markup.unnecessary.lsp",
            "foreground_adjust": "l(- 40%)",
            "background": "#00000001"
        },
    ]
}
```
4. Open a `hello.vue` file, that looks like this
```
<script setup lang="ts">
let xas = 312

asdaaasssdsasdasd // type here and see the flicker 

</script>
```
6. Type where the `asdaaasssdsasdasd` string is
7. See flicker

**Expected behavior**
There should be no flicker.

**Screenshots**

output

This cannot be reproduced in all vue/ts files, 
only in certain fails like the one I provided.

----

#### Solution

I solved the problem by modifying this function,
but don't know if you guys have a better suggestion.
2023-05-17-212104_855x298_scrot

output

rchl:
Avoids erasing regions if we're gonna re-add regions with same key right away.

Fixes #2272

### DIFF
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -19,8 +19,17 @@ from weakref import WeakValueDictionary
 import functools
 import sublime
 
-DIAGNOSTIC_TAG_VALUES = [v for (k, v) in DiagnosticTag.__dict__.items() if not k.startswith('_')]
+DIAGNOSTIC_TAG_VALUES = [v for (k, v) in DiagnosticTag.__dict__.items() if not k.startswith('_')]  # type: List[int]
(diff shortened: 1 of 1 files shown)


predragnikolic (approved):
This fixes it.
Thanks!

Issue #2244, opened 2023-05-03
Add support to pass includeDeclaration as True for textDocument/references


EmasXP:
When searching for references, I would like to include the declarations in the result.

My thinking is that this could be done by adding a `include_declaration` argument to `lsp_symbol_references`. The default value should in that case be `False`.

What do you think?

jwortmann:
Sounds like a good idea to me. It seems that the `False` value has always been hardcoded for this, probably because it's inline with Sublime's built-in "Goto References..." behavior. The new command argument would mean that you need to add your own context menu entry and/or key binding if you want to change this behavior.

An alternative would be to expose this as a new global setting, which would be more user friendly. But new settings mean more maintenance work in the future, and I think that the impact of changing this option would be rather small. So I would say your suggestion is better.

EmasXP:
I can do a PR for this if you want? Disclosure: that would be my first ever PR. I am happy to do it though :slightly_smiling_face:

predragnikolic:
@EmasXP feel free to open a PR :)

PR #2275, opened 2023-05-19
Adding argument "include_declaration" to "lsp_symbol_references".

EmasXP:
Closes https://github.com/sublimelsp/LSP/issues/2244

### DIFF
--- a/plugin/references.py
+++ b/plugin/references.py
@@ -55,7 +55,8 @@ class LspSymbolReferencesCommand(LspTextCommand):
         side_by_side: bool = False,
         force_group: bool = True,
         fallback: bool = False,
-        group: int = -1
+        group: int = -1,
+        include_declaration: bool = False
     ) -> None:
         session = self.best_session(self.capability)
         file_path = self.view.file_name()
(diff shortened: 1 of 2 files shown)


plugin/references.py
```diff
@@ -55,7 +55,8 @@ def run(
         side_by_side: bool = False,
         force_group: bool = True,
         fallback: bool = False,
-        group: int = -1
+        group: int = -1,
+        include_declaration: bool = False
```
rchl, line 59: Needs to be added in `is_visible` and `is_enabled` above also.

EmasXP: Thank you. I have updated the PR.

What is the practice here, shall I resolve the conversation, or shall a reviewer do it?

rchl: Think best to leave for the issue maker to verify and resolve.

predragnikolic:
Should probably add the include_declaration argument in this file ->

https://github.com/sublimelsp/LSP/blob/9ec8001622131df129ed3833e975074be7288e8f/Default.sublime-keymap#L66

(I am on mobile, please do a find in project and see if you need to add the argument to any other place.

EmasXP:
I added `include_declaration` as `false` in Default.sublime-keymap. 

The command is referenced here too:

* Context.sublime-menu
* Default.sublime-commands
* Main.sublime-menu

I decided to not add the argument there since since no other argument is specified there either. Do you want me to add the argument anyway?

predragnikolic:
No need to add it there.

predragnikolic (approved):

jwortmann (approved):

PR #2276, opened 2023-05-20
fix crash on checking excluded folders with missing project data

rchl:
Ensure that the `_folders_exclude_patterns` list has same number of elements as `folders` so that accessing by index doesn't throw when `window.project_data()` is missing.

Fixes https://github.com/TerminalFi/LSP-copilot/issues/109

### DIFF
--- a/plugin/core/workspace.py
+++ b/plugin/core/workspace.py
@@ -63,7 +63,7 @@ class ProjectFolders(object):
         self._update_exclude_patterns(self.folders)
 
     def _update_exclude_patterns(self, folders: List[str]) -> None:
-        self._folders_exclude_patterns = []
+        self._folders_exclude_patterns = [[]] * len(folders)
         project_data = self._window.project_data()
         if not isinstance(project_data, dict):
             return
@@ -79,7 +79,7 @@ class ProjectFolders(object):
                 else:
                     exclude_patterns.append(sublime_pattern_to_glob('//' + pattern, True, path))
                     exclude_patterns.append(sublime_pattern_to_glob('//**/' + pattern, True, path))
-            self._folders_exclude_patterns.append(exclude_patterns)
+            self._folders_exclude_patterns[i] = exclude_patterns
 
     def update(self) -> bool:
         new_folders = self._window.folders()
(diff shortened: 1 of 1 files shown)


jwortmann (approved):

plugin/core/workspace.py
```diff
@@ -63,7 +63,7 @@ def __init__(self, window: sublime.Window) -> None:
         self._update_exclude_patterns(self.folders)
 
     def _update_exclude_patterns(self, folders: List[str]) -> None:
-        self._folders_exclude_patterns = []
+        self._folders_exclude_patterns = [[]] * len(folders)
         project_data = self._window.project_data()
```
predragnikolic, line 67: I would definitely put a comment on that line, or link to the issue,
it was not obvious why it needs to be multiplied by the length of the folders

jwortmann: `[x] * n` is a concise syntax in Python to create a list with n entries which are all x, i.e. here it will be for example `[[], [], []]` with 3 folders.
But yeah, I was also surprised that this bug can happen, and I never saw it with "normal" usage. The `window.project_data()` even returns a dict with the (sidebar) folders, if there is no project at all (iirc ST automatically creates a "hidden" project in that case), or with a single .sublime-workspace file. It might be possible with multiple workspace files or if you edit them manually.
But with this initialization line we are certain to be on the safe side, so it looks fine to me.

jfcherng: Note that those lists created via `[[]] * len(folders)` refer to exactly the same thing. Not sure whether this is what you want but just provide a potential issue.

```
>>> a = [[]] * 4
>>> a[0].append('foo')
>>> a
[['foo'], ['foo'], ['foo'], ['foo']]
```

jwortmann: It should be fine, because each entry gets replaced completely, not appended to:
```python
self._folders_exclude_patterns[i] = exclude_patterns
```

For example
```
>>> a = [[]] * 4
>>> a[1] = ['foo']
>>> a
[[], ['foo'], [], []]
```

predragnikolic (approved):

rwols (approved):

PR #2277, opened 2023-05-21
fix "Error rewriting command" warning triggered on startup

rchl:
Fixes `Error rewriting command lsp_recheck_sessions. Encountered infinite loop` warning on startup.

The warning is a bit misleading. It triggers not due to an infinite loop but due to triggering a window command directly from `plugin_loaded`. Basically means that triggering this command on startup had no effect but that's fine because at that point there is no sessions yet so it wouldn't really do anything anyway.

This PR removes the `LspRecheckSessionsCommand` command in favor of a more explicit notification triggered from the `WindowConfigManager` to the `WindowManager` to let it know that sessions should be restarted (the approach with using `LspRecheckSessionsCommand` was just a hack around the fact that `WindowConfigManager` did not have access to `WindowManager`).

Also note that I've changed `WindowConfigManager` to not trigger sessions restart on `__init__` since it's unnecessary since there are no sessions yet. That's why most of the `update()` code was moved to `_reload_configs()` and called on init.

### DIFF
--- a/boot.py
+++ b/boot.py
@@ -25,7 +25,6 @@ from .plugin.core.registry import LspExpandTreeItemCommand
 from .plugin.core.registry import LspNextDiagnosticCommand
 from .plugin.core.registry import LspOpenLocationCommand
 from .plugin.core.registry import LspPrevDiagnosticCommand
-from .plugin.core.registry import LspRecheckSessionsCommand
 from .plugin.core.registry import LspRestartServerCommand
 from .plugin.core.registry import windows
 from .plugin.core.sessions import AbstractPlugin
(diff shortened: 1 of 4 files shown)


jwortmann (approved):
I couldn't find a way how to reproduce the "Error rewriting command", but the refactor from this PR looks useful to me.

rchl:
Thanks for checking.

The warning is new addition in latest dev builds but the issue was there even before the warning was introduced. The issue being that triggering a command directly from `plugin_loaded` didn't really do anything.

Also related issue in ST about this warning being misleading: https://github.com/sublimehq/sublime_text/issues/5978

PR #2279, opened 2023-05-27
Take font style of sighelp active parameter from color scheme

jwortmann:
Implements https://github.com/sublimelsp/LSP/pull/2259#issuecomment-1554831618.

It turned out we don't need the bugged "source_line" to decide whether there is a special color scheme rule for the active parameter. Instead, we can just compare the foreground colors.


rchl (approved):

frou:
Thanks for fixing this!

(I just got around to updating to LSP 1.25 since Package Control 3.x is broken on macOS).

PR #2281, opened 2023-06-01
Remove experimental hover range support from core

rchl:
~~The implementation on the rust-analyzer side has changed so updating our code to follow.~~

~~ref: https://github.com/rust-lang/rust-analyzer/blob/7e19d99d4f0631b141d4d49a49cd5c504726cf9c/docs/dev/lsp-extensions.md#hover-range~~

Moving experimental functionality from here to respective packages:
 - https://github.com/sublimelsp/LSP-rust-analyzer/pull/98
 - https://github.com/scalameta/metals-sublime/pull/108

### DIFF
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -6148,8 +6148,7 @@ CodeLensExtended = TypedDict('CodeLensExtended', {
 
 ExperimentalTextDocumentRangeParams = TypedDict('ExperimentalTextDocumentRangeParams', {
     'textDocument': TextDocumentIdentifier,
-    'position': Position,
-    'range': Range,
+    'position': Range,
 })
 
 CompletionItemDefaults = __CompletionList_itemDefaults_Type_1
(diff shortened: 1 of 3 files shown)


rchl:
Just realized that the original code was to support Metals (#1900). So we might need to support both...

jwortmann:
So, it looks like the code on the rust-analyzer side was always like this. Then I wonder, wouldn't it be better to move this code into LSP-rust-analyzer and LSP-Metals, respectively (`AbstractPlugin.on_pre_send_request_async`)? Otherwise, if we follow the strategy to put everything into LSP core, there is the risk that it will eventually lead to conflicts if different servers use the same capability name but expect different parameters. And the corresponding issue on Microsoft's LSP repo is already open since 5 and a half years, so I think we cannot expect this to be included into the official specs soon... (and I see no PR for it).

rchl:
Good idea. Created PR for metals: https://github.com/scalameta/metals-sublime/pull/108

rchl:
Created respective PRs for https://github.com/sublimelsp/LSP-rust-analyzer/pull/98 and https://github.com/scalameta/metals-sublime/pull/108 and removing this experimental functionality from here.

Issue #2270, opened 2023-05-16
Run code actions on save only once 


predragnikolic:
**Describe the bug**
When pressing save a file 
and I have code actions on save enabled

```
	"lsp_code_actions_on_save": {
		"source.fixAll.eslint": true
	},
```

With the following zipped repro, this happens:

![output](https://github.com/sublimelsp/LSP/assets/22029477/386d7e91-5b77-415b-a28e-dc7110b99c11)

**To Reproduce**
1. Unzip
[eslint-caos.zip](https://github.com/sublimelsp/LSP/files/11492021/eslint-caos.zip)
2. run `npm i` in root

**Expected behavior**
TBD.
I would suggest running the code action on save only once,
instead of retrying, (also that is what VS Code does)

![output](https://github.com/sublimelsp/LSP/assets/22029477/00622e79-c31c-4be5-b367-a0963478c5ba)

**Current Behavior**
LSP tries to run code action on save multiple times, which makes undo sometimes hard to undo.
The same issue happens with `organizeImports`.

While a fix for this wont fully fix `organizeImports`, (because that needs to be fixed on the server),
users would at least be able to undo things, (because CAOS will only be run once)

Related:
https://github.com/sublimelsp/LSP/issues/2231
https://github.com/sublimelsp/LSP-typescript/issues/202

> NOTE:
> I know that the code example looks like it should never happen in practice, 
but it happened to me on 2 projects which had eslintrc files, with rules that in some conditions conflict each other. (I created a mini repro in the zip)

PR #2283, opened 2023-06-03
Only do a single pass on running code actions on save

### LINKED ISSUE #2231: organizeImports and fixAll in a loop!
Hi. Got a weird issue that happens:

Code gets reorganised to this via the `"source.organizeImports": true` option (note the extra spacing before each import):

```
import {
    useAcceptContracts,
    useAddressUpdate,
    useReserveAppointment,
    useSetVisitAttributes,
    useSetVisitContactDetails,
    useSubmitCancellationReasons,
    useSubmitOrder,
    useUpdateVisitReasons
} from '../../utilities/hooks';
```

I fix all autofixable eslint issues (the issue is remove the extra spacing before each import):

```
import {
  useAcceptContracts,
  useAddressUpdate,
  useReserveAppointment,
  useSetVisitAttributes,
  useSetVisitContactDetails,
  useSubmitCancellationReasons,
  useSubmitOrder,
  useUpdateVisitReasons
} from '../../utilities/hooks';
```

Save it and it reverts to the 1st code where it has double spaces! Saving via the Sublime menu does not trigger the LSP commands so is a way out of the loop.

Not sure it could be anything other than LSP as saving via cmd + s works when this is false `"source.organizeImports": false`.

**Expected behavior**
The `source.organizeImports` should just add one space otherwise that command and the `source.fixAll` command seem to be stuck in a loop!

- OS:macOS 13.2.1
- Sublime Text version: 4143
- LSP version: v1.23.0
- Language servers used: eslint (for js and jsx files)

rchl:
Previously we have been re-running Code Actions on Save if the document got modified by code action. This have in some cases resulted in a loop that would only terminate on timeout. Changed so that each server only has one chance to run code actions on save.

Fixes #2231
Fixes https://github.com/sublimelsp/LSP-typescript/issues/202

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -253,20 +253,12 @@ class CodeActionOnSaveTask(SaveTask):
     def _handle_response_async(self, responses: List[CodeActionsByConfigName]) -> None:
         if self._cancelled:
             return
-        document_version = self._task_runner.view.change_count()
         tasks = []  # type: List[Promise]
         for config_name, code_actions in responses:
             session = self._task_runner.session_by_name(config_name, 'codeActionProvider')
             if session:
                 tasks.extend([session.run_code_action_async(action, progress=False) for action in code_actions])
-        Promise.all(tasks).then(lambda _: self._on_code_actions_completed(document_version))
-
-    def _on_code_actions_completed(self, previous_document_version: int) -> None:
-        if previous_document_version != self._task_runner.view.change_count():
-            # Give on_text_changed_async a chance to trigger.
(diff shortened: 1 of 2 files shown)


predrag-codetribe (approved):
Tested ✅ 

I checked 
- `"source.fixAll.eslint"` - with the code from this PR, this will no longer happen https://github.com/sublimelsp/LSP/issues/2270
- and `"source.organizeImports"` (with LSP-typescript), I am able to undo as expected, (previously it was sometimes impossible to undo)

PR #2286, opened 2023-06-11
refactor: get rid of `SessionBuffer.data_per_severity`

rchl:
Just a small refactor to get rid of publicly exposed `SessionBuffer.data_per_severity` and pass that data to `SessionView` instead.

Also add `_` prefix to private properties of `SessionBuffer` to make it clear that those are private.


rchl:
A noop change assuming that I've made no mistakes so merging.

PR #2292, opened 2023-06-26
docs: add ols server for Odin

ryuukk:


rchl:
The instructions in the Odin repo seem incomplete:
 - the command is Windows-specific
 - the `only_show_lsp_completions` option doesn't exist in LSP
 - should specify which syntax package is required since LSP won't do anything without proper syntax
 - maybe should also suggest specific command to use to enable the server since it's by default disabled.

rchl (approved):

PR #2293, opened 2023-06-29
Replace dots in the trigger with a different character that is visually the same

ryuukk:
This helps cases with Sublime Text strips some completion items

Check the following issue for more details: https://github.com/sublimehq/sublime_text/issues/6033

Question:

1. trigger wont insert anything?
2. if it inserts something, where would this trick be used instead?

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -97,7 +97,7 @@ def format_completion(
         details.append("<a href='{}'>{}</a>".format(command_url, oposite_insert_mode))
     # https://github.com/sublimehq/sublime_text/issues/6033
     # replaces . with a different one that looks the same
-    trigger = trigger.replace(".", "ꓸ")
+    trigger = trigger.replace(".", "․")
     completion = sublime.CompletionItem(
         trigger,
         annotation,
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -95,6 +95,9 @@ def format_completion(
         oposite_insert_mode = 'Replace' if insert_mode == 'insert' else 'Insert'
         command_url = "subl:lsp_commit_completion_with_opposite_insert_mode"
         details.append("<a href='{}'>{}</a>".format(command_url, oposite_insert_mode))
+    # https://github.com/sublimehq/sublime_text/issues/6033
+    # replaces . with a different one that looks the same
+    trigger = trigger.replace(".", "ꓸ")
     completion = sublime.CompletionItem(
         trigger,
         annotation,


jwortmann:
>     1. trigger wont insert anything?

Correct, as far as I can tell. It will use the original CompletionItem from the language server again, for the inserted text, so we are free to modify the trigger. But it will only have an effect on filtering.

---

Some observations:

1. When I try this in a simplified example, it seems like the normal `.` in the trigger cannot be the single cause of this. The following example seems to work fine for me:
```python
import sublime
import sublime_plugin
from typing import List, Optional


class CompletionsListener(sublime_plugin.ViewEventListener):

    def on_query_completions(self, prefix: str, locations: List[int]) -> Optional[sublime.CompletionList]:
        completion_list = sublime.CompletionList()
        sublime.set_timeout_async(lambda: self._on_query_completions_async(completion_list))
        return completion_list

    def _on_query_completions_async(self, completion_list: sublime.CompletionList) -> None:
        flags = sublime.DYNAMIC_COMPLETIONS | sublime.INHIBIT_REORDER
        trigger = "erro(fmt: string, args: ..any, location := caller_location)"  # unmodified
        # trigger = "erro(fmt: string, args: ꓸꓸany, location := caller_location)"  # U+A4F8
        # trigger = "erro(fmt: string, args: ․․any, location := caller_location)"  # U+2024
        annotation = "rt.erro: proc(fmt: string, args: ..any, location := caller_location)"
        completion = "erro"
        completions = [sublime.CompletionItem(trigger, annotation, completion)]
        sublime.set_timeout(lambda: completion_list.set_completions(completions, flags))
```

unmodified

So as a first step we should identify the root cause or the combination of factors which makes the completion not work in your example (linked issue). Just to exchange some random symbols might not work in all situations or even possibly make it worse for other examples.

2. You chose a quite uncommon replacement symbol in this PR (Lisu Letter Tone Mya Ti) https://www.compart.com/en/unicode/U+A4F8
I doubt that many fonts have support for this symbol, and probably use some fallback instead. It doesn't look very good here:

pr

Maybe this one would be a little bit better: https://www.compart.com/en/unicode/U+2024

predragnikolic:
Closing this, as this change is unnecessary.

If you need this behavior, you can create a LSP package for this language and override the language server response:
https://github.com/sublimelsp/LSP/pull/1992/files#diff-8fe432afb96c719bc734bcb2a1045d852979d09ea395956f1b979861432c2b05R878

PR #2296, opened 2023-07-05
Show server crashed dialog on unexpected output in server's stdout

### LINKED ISSUE #2295: LSP server initialization does not respond
This may be a bug upstream, but I haven't been able to reproduce it in VSCode, so I'm reporting here in case this appears to be an obvious problem with LSP-volar.

Reproduce steps:

1. Unzip [volar-starter-with-nodemodules.zip](https://github.com/sublimelsp/LSP-volar/files/11952697/volar-starter-with-nodemodules.zip) and open it in Sublime.
2. Open App.vue
3. Open LSP logs, you should see the initialization request sent, but no response received.

It's worth noting that if you delete node_modules and reinstall it via `pnpm i`, LSP will work fine.

rchl:
On `int(headers.get("Content-Length"))` crashing in `JsonRpcProcessor.read_data`, we've assumed that it's an expected case happening on server exit. As far as I've observed, the typical data we get on process exit is a newline, but it can also be any other random output that doesn't conform to the HTTP header format.

In either of those cases we were just silently closing the transport, treating it as expected session shutdown.

Changed so that we only silently exit transport if the stdout was a single newline. In other cases propagate the output to the UI so that a crash dialog is shown.

I've also changed the `JSON decode error` case to also propagate the error to the UI instead of silently ignoring payload. I think this makes more sense but we'd have to see how it works out in practice.

Resolves #2295

### DIFF
--- a/plugin/core/transports.py
+++ b/plugin/core/transports.py
@@ -66,13 +66,16 @@ class JsonRpcProcessor(AbstractProcessor[Dict[str, Any]]):
         try:
             body = reader.read(int(headers.get("Content-Length")))
         except TypeError:
-            # Expected error on process stopping. Stop the read loop.
-            raise StopLoopError()
+            if str(headers) == '\n':
+                # Expected on process stopping. Gracefully stop the transport.
+                raise StopLoopError()
+            else:
+                # Propagate server's output to the UI.
+                raise Exception("Unexpected payload in server's stdout:\n\n{}".format(headers))
         try:
             return self._decode(body)
         except Exception as ex:
-            exception_log("JSON decode error", ex)
-            return None
+            raise Exception("JSON decode error: {}".format(ex))
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


rwols (approved):

predrag-codetribe:
Questions,
Will the server continue to run when a "Unexpected payload in server's stdout" occurs?

(I think that it will continue to run, but I just want to confirm, as I didn't yet tried to test this)

predragnikolic:
I am wondering if we could print unexpected output to the LSP console,(so we will be aware that the server is returning something unexpected)
and keep the server running.

That would fix https://github.com/sublimelsp/LSP/issues/1612

rchl:
Where do you draw the line? Such behavior is essentially not spec-compliant. I don't think we should be fixing or ignoring server issues like that.

Also I would think that that specific eslint issue should be fixable on the server level.

PR #2299, opened 2023-07-14
Add support for multi-range formatting

jwortmann:
This should add support for the new `textDocument/rangesFormatting` request from the LSP 3.18 specs, see https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocument_rangeFormatting.

(I didn't add a new command for it, but instead made it a special case of the `lsp_format_document_range` command, because I think it would be unnecessary to expose this difference to users.)

Disclaimer: untested due to lack of server support at the moment


plugin/core/views.py
```diff
@@ -592,6 +592,14 @@ def text_document_range_formatting(view: sublime.View, region: sublime.Region) -
     }, view, progress=True)
 
 
+def text_document_ranges_formatting(view: sublime.View) -> Request:
+    return Request("textDocument/rangesFormatting", {
+        "textDocument": text_document_identifier(view),
+        "options": formatting_options(view.settings()),
+        "ranges": [region_to_range(view, region) for region in view.sel() if not region.empty()]
```
rchl, line 599: Spec doesn't seem to specify that range can't be empty. I would think that it could potentially be useful to place multiple cursors in multiple places and expect the code around those places to be formatted.

jwortmann: I would understand the specs in the way that servers must respond with TextEdits which are only allowed to modify the document within the given range(s).

> The document range formatting request is sent from the client to the server to format a given range in a document.

So if the range is empty, there is nothing useful a server could do, other than sending an empty/null response. Do you know of any server which behaves different?

In fact, LSP already prevents empty selections for range formatting at the moment:
https://github.com/sublimelsp/LSP/blob/e608f878e7e9dd34aabe4ff0462540fadcd88fcc/plugin/formatting.py#L117-L119

rchl: Based on me just trying LSP-json and using Format Selection, it seems like you are right. I haven't tried with empty selection of course since that is disabled, but I've tried with very small selection to see if server expands formatting beyond the selection but it seemed to only format within the selection.

Maybe worth trying some more servers before agreeing here...

rchl (approved):

Issue #2214, opened 2023-03-04
Do not start the server if i provide a TCP port, Sublime is a Client, not a Server starter


ryuukk:
When i provide the "tcp_port", sublime text will not connect to the server, instead it'll start the server

**To Reproduce**

[app.zip](https://github.com/sublimelsp/LSP/files/10889479/app.zip)

Sample server, it only start a TCP server on port ``6969``

Use this config:

```
		"mine": {
			"enabled": true,
			"tcp_port": 6969,
			//"command": ["c:\\dev\\kdom\\projects\\lsp\\app.exe"],
			"selector": "source.d",
			"settings": {
			},
		}
```

if i comment the "command" i get this error btw:

![image](https://user-images.githubusercontent.com/44361234/222925312-82113874-79bc-4a47-a2bb-1a8453881a19.png)

**Expected behavior**

Sublime Text should initiate a TCP connection to my already open server

It should not start the server

**Environment (please complete the following information):**
- OS: Windows 10 x64
- Sublime Text version: 4143
- LSP version: 1.22.0
- Language servers used: mine, provided above

**Additional context**

Use cases:

- Compiler server that manages its own language server
- Easier to debug as a whole since i manage server lifetime completely
- language server (compiler server) that lives in my server and clients, i can close sublime windows freely

rwols:
I agree this would be nice to have.

weckr:
+1 for me, i use this with godot, and the editor (and so lsp) is always running when i code by the nature of the work, dont really want sublime to mess with it. That would even be tolerable but for some reason the plugin is spinning one core to 100 percent continuously. Ive tried to give it random dummy shell command like ls or something so it would let me use the lsp, and the lsp with the tcp port filled in actually works like this, but again it spins the core making it really annoying on laptop in summer...
I might look into this myself, since its python. It would be what, just checking if tcp-port setting is present and then ignoring the missing command setting and execution?

rwols:
@weckr the place to change this is probably in https://github.com/sublimelsp/LSP/blob/31945a6b0a573935907536efacbaa482771be472/plugin/core/transports.py#L233

weckr:
thanks, the nc way gets buggy and slow for some reason (although at this point it just may be godots lsp), so in the end I managed to hack it in somehow.

predragnikolic:
fixed with https://github.com/sublimelsp/LSP/pull/2300

PR #2300, opened 2023-07-16
support client config with `tcp_port` but without `command`

weckr:
https://github.com/sublimelsp/LSP/issues/2214

I have no idea if this is too hacky to be pulled, or what it breaks. It does work for my usecase tho. Would be cool if it doesn't find a server, it would just wait until one shows up instead of disabling the lsp until manually enabled again. Maybe i'll do it if I find it annoying enough.

If there would be small edits needed to be pulled I can totally do them, if there are fundamental issues I probably wont bother.

### DIFF
--- a/plugin/core/transports.py
+++ b/plugin/core/transports.py
@@ -94,8 +94,8 @@ class JsonRpcProcessor(AbstractProcessor[Dict[str, Any]]):
 
 class ProcessTransport(Transport[T]):
 
-    def __init__(self, name: str, process: Optional[subprocess.Popen], socket: Optional[socket.socket], reader: IO[bytes],
-                 writer: IO[bytes], stderr: Optional[IO[bytes]], processor: AbstractProcessor[T],
+    def __init__(self, name: str, process: Optional[subprocess.Popen], socket: Optional[socket.socket],
+                 reader: IO[bytes], writer: IO[bytes], stderr: Optional[IO[bytes]], processor: AbstractProcessor[T],
                  callback_object: TransportCallbacks[T]) -> None:
         self._closed = False
         self._process = process
@@ -137,7 +137,7 @@ class ProcessTransport(Transport[T]):
         self.close()
         self._join_thread(self._writer_thread)
         self._join_thread(self._reader_thread)
-        
+
         if self._stderr_thread is not None:
             self._join_thread(self._stderr_thread)
 
@@ -168,7 +168,7 @@ class ProcessTransport(Transport[T]):
 
     def _end(self, exception: Optional[Exception]) -> None:
         exit_code = 0
-        if not exception:
+        if not exception and self._process:
             try:
                 # Allow the process to stop itself.
                 exit_code = self._process.wait(1)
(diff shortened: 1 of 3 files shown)


plugin/core/transports.py
```diff
@@ -95,23 +95,25 @@ def _decode(message: bytes) -> Dict[str, Any]:
 class ProcessTransport(Transport[T]):
 
     def __init__(self, name: str, process: subprocess.Popen, socket: Optional[socket.socket], reader: IO[bytes],
-                 writer: IO[bytes], stderr: Optional[IO[bytes]], processor: AbstractProcessor[T],
+                 writer: IO[bytes], processor: AbstractProcessor[T],
```
rchl, line 98: We shouldn't change this signature because this class used by [external package](https://github.com/sublimelsp/LSP-file-watcher-chokidar) and so it would be a breaking change.

We can just pass `None` since it was already optional.

plugin/core/transports.py
```diff
@@ -258,7 +260,8 @@ def start_subprocess() -> subprocess.Popen:
             start_subprocess
         )
     else:
-        process = start_subprocess()
+        if config.command:
+            process = start_subprocess()
```
rchl, line 264: We should have some error handling and raise an error for the case when we expect `command` but it's not provided. I suppose we should only allow that when `tcp_port` is not set.

rchl:
We should mention this new behavior in documentation (in the docs folder). Currently it states that we start the server - https://lsp.sublimetext.io/client_configuration/#tcp-localhost-subprocess-acts-as-a-tcp-server

plugin/core/transports.py
```diff
@@ -95,23 +95,25 @@ def _decode(message: bytes) -> Dict[str, Any]:
...
         self._callback_object = weakref.ref(callback_object)
         self._send_queue = Queue(0)  # type: Queue[Union[T, None]]
         self._reader_thread.start()
         self._writer_thread.start()
-        self._stderr_thread.start()
+
+        if process is not None:
+            self._stderr = process.stderr
+            self._stderr_thread = threading.Thread(target=self._stderr_loop, name='{}-stderr'.format(name))
+            self._stderr_thread.start()
```
rchl, line 116: There is `__del__` method below that will try to join this thread even when not created. Should be some if-check there.

plugin/core/transports.py
rchl, line 98: `process` can be `None` now so that should be reflected in types

weckr:
Thanks for detailed feedback, tried to fix it all, the way I pass the stderr argument to the constructor looks filthy but no idea what is correct form to do this so I just kept it simple and ugly.

weckr:
the linter stuff should be fixed, no idea what the macos error is about

rchl (approved):

weckr:
thanks for putting up with my mess, do you use any specific stuff for the formatting? black? or do am i missing something in the pylsp/pyright that i have in sublime.

rchl:
I'm just following the conventions that we have in this code base and the changes were all manual.

PR #2304, opened 2023-07-26
Add support for folding range request

jwortmann:
ST doesn't provide an API to show fold buttons in the gutter for custom (LSP) folding ranges, but I think we can still make it available via the menu or a key binding. It could be useful if there is no built-in indentation based folding, and no ST syntax based folding available. And maybe one day it will be possible to provide the folding regions to the gutter via the API.

Example with LSP-TexLab (to be honest, the folding ranges from this server are a bit strange, but it could be the server's design decision):

[fold.webm](https://github.com/sublimelsp/LSP/assets/6579999/1f2879aa-55a5-48bc-9124-a61522da4608)

<details><summary>LaTeX code</summary>

```latex
\documentclass{article}
\usepackage[utf8]{inputenc}
\usepackage[T1]{fontenc}
\usepackage{amsmath}


\begin{document}

\section{First Section}

Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod
tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse
cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non
proident, sunt in culpa qui officia deserunt mollit anim id est laborum.

\section{Second Section}

Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod
tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse
cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non
proident, sunt in culpa qui officia deserunt mollit anim id est laborum.


\end{document}
```
</details>

In practice it is probably more useful to define a key binding, instead of using the menu.

The advantage is that this feature causes almost no overhead, because the request is only sent on demand when you open the "Edit" menu or when invoked via key binding.

Initially I tried to make it available via the context menu, but it turned out to not really work, because for the context menu ST will call the command's `description`/`is_visible` methods for submenu entries at the same time as for top-level menu entries. (The trick for the "Edit" menu is, that ST only calls those methods for submenu entries at the time when the submenu gets opened. So we already have the server response when that happens, and can show/hide the entry and an appropriate description if applicable - similar to the code actions in the menu.)


jwortmann:
> Looks like CI is broken on Mac 🤷

This seems to be a general issue caused by incompatibility to outdated OpenSSL/libcrypto dependencies on macOS, and will persist until a Package Control version 4.0 is released: https://github.com/wbond/package_control/issues/1612#issuecomment-1642451382

We should disable CI on macOS in the meanwhile.

---

For this PR, I noticed a minor bug which will show a greyed out "Lsp Fold" entry in the "Edit" menu, when the sidebar has focus. This is the same problem as previously described in #2187 and could be fixed by using a WindowCommand instead (but I think doing that would have other disadvantages, so I'm not sure yet whether it's feasible here).

plugin/folding_range.py
```diff
@@ -0,0 +1,138 @@
...
+                pt = selection[0].b
+            session = self.best_session(self.capability)
+            if session:
+                params = {'textDocument': text_document_identifier(self.view)}  # type: FoldingRangeParams
+                session.send_request_async(
+                    Request.foldingRange(params, self.view),
+                    partial(self._handle_response_manual_async, pt)
+                )
+        elif self.folding_region is not None:
+            self.view.fold(self.folding_region)
```
rchl, line 122: I've been testing a bit and I think we could potentially improve behavior in some cases (but not necessarily in this PR).

What I find a bit inconsistent is when I fold a region and then press the fold shortcut again.

Let's take JSON as an example:

1.

Screenshot 2023-07-30 at 15 57 32

2.

Screenshot 2023-07-30 at 15 57 50

At this point pressing the shortcut doesn't do anything. I would expect it to either unfold the folded region or fold  the "outer" region.

(Of course there is no structure to regions so it would be our job to figure out which regions overlap and what would make sense if we'd want to fold "outer" region in this case.)

jwortmann: > At this point pressing the shortcut doesn't do anything. I would expect it to either unfold the folded region or fold the "outer" region.

Good point, and I think both alternatives wouldn't be too difficult to implement. When the region is already folded, the search in the for-loop for the relevant region will still match, because the caret is on one side of the region. I just need to add another check in the for-loop if the relevant region is already folded, and then either ignore it (fold "outer" region) or unfold it.

Which of the two option would you prefer / think is better?

rchl: The ST behavior is to fold the "outer" region if current is already folded so I guess it would make sense to match that.

Another argument for that would be that the action is called "Fold" and not "Toggle fold" so in theory it should never unfold (even when there is non-empty selection).

plugin/folding_range.py
```diff
@@ -0,0 +1,138 @@
...
+                # self.is_visible and self.run
+                self.folding_region = region
+                return {
+                    FoldingRangeKind.Comment: "LSP: Fold this comment",
+                    FoldingRangeKind.Imports: "LSP: Fold imports",
+                    FoldingRangeKind.Region: "LSP: Fold this region",
+                    'array': "LSP: Fold this array",  # used by LSP-json
+                    'object': "LSP: Fold this object",  # used by LSP-json
+                }.get(folding_range.get('kind', ''), "LSP: Fold")
+        return "LSP <debug>"  # is_visible will return False
```
rchl, line 95: Noticed interesting behavior with:

```jsonc
{
    "foo": [1,
        2,
        3
    ]
}
```

Folding regions in this case are: `[{'startLine': 15, 'endLine': 16, 'kind': 'array'}, {'startLine': 1, 'endLine': 3, 'kind': 'array'}, {'startLine': 0, 'endLine': 4, 'kind': 'object'}]`

(not sure what the first region is about but we can ignore it)

With cursor anywhere within line 1, we fold the root object.
In VSCode it applies the second fold region.

While I understand that per spec it's supposed to fold from the end of the `startLine`, maybe the cursor intersection region should actually be checked against the start of the line. It seems like a smart behavior.

More usuful example would be:

```py
def fn():
    if foo:
        'xxx'
```

where it could fold the body of `if foo` when on that line but currently it folds the the whole body of the function.

jwortmann: > With cursor anywhere within line 1, we fold the root object.
> In VSCode it applies the second fold region.

I think a part of this problem is that the JSON language server doesn't report the optional `startCharacter` & `endCharacter` values. If you put the caret within `"foo"`, then the builtin `fold` command (Edit > Code Folding > Fold) also folds the root object. But if you put it between the `[` and `1`, the builtin command folds the array, as expected. If the server would use the `startCharacter`, then the `lsp_fold` behavior would be the same.

Maybe it can be improved with some heuristics if `startCharacter` is missing, but I'm unsure if this would do the right thing in all cases. For example with
```json
{
    "foo": [1, ["a",
            "b",
            "c"
        ],
        2,
        3
    ]
}
```
I think there is no way to decide which of the arrays should be folded. It would probably only makes sense to fold the outermost array in that case.

For the Python example, the builtin behavior is also to fold the function body when the cursor is within the `if foo` line. And it does nothing if the caret is on the first line. This is only the indentation based folding, but it would probably still be trivial to implement to fold the "inner" block, if the caret is on the last line before a new indentation level, if this was the desired behavior. So I'm not sure whether it would really be better to implement this in `lsp_fold`, when it differs from the builtin functionality. But I'm open for discussion.

LSP-TexLab has support for `startCharacter` & `endCharacter`, but doesn't really seem to use it advantageously.

rchl: I have a feeling that the behavior that VSCode shows makes more sense. At least for languages like python (and to JSON with some corner-case exceptions too) where it feels natural to be able to fold a method or a class while having the cursor on the `def ...` or `class ...` line.

Also, that behavior kinda matches ST if you compare it with the folding markers. ST provides folding markers next to the `def ...` or `class ...` and folds bodies without including that line itself. I feel like comparing the LSP functionality to those markers makes more sense than comparing it to the plain `Fold` command since both the syntax-based folding and the LSP folding is supposed to be "smarter".

jwortmann: Both ways have advantages and disadvantages.

I added a `strict` parameter to make this configurable; `strict = true` means ST built-in behavior (caret must be contained within the folding region) and `strict = false` means to fold the next block/range even if the caret is anywhere on the starting line (starting with innermost/rightmost range if there are multiple ranges on the same line). Using `true` as the default value, because I think an intention for the LSP commands would be to use them as a direct replacement in key bindings for the built-in ST command, so they should probably follow ST behavior as close as possible/useful. And `strict = true` allows more precise behavior when there are multiple folding ranges on a single line, dependent on the exact caret position.

> Also, that behavior kinda matches ST if you compare it with the folding markers. ST provides folding markers next to the `def ...` or `class ...` and folds bodies without including that line itself.

I think folding buttons must be considered independent of the behavior what happens when you fold via key binding or from the menu. The fold buttons in the gutter are on the line of the `def ...` because this line is where the folded region begins; in particular the linebreak (`\n`) is the first character of the folded region, and this is the place where ST places the (typically) yellow fold marker `(...)`. You can also confirm by placing the caret at the end of such a line; in that case ST built-in `fold` command indeed folds the following block. So you cannot say that the one way or the other would match the fold button placement any more or less.

When/if ST starts to allow plugins to controll the fold buttons in the gutter, we would of course also use the line from the starting position of the folding range.

jwortmann:
I've also added a `lsp_fold_all` command for parity with built-in `fold_all`. VS Code has a command "Fold All Block Comments" which could be useful, but I'm a bit hesitant to add it to the command palette, because it has no effect if the server doesn't report the optional `kind` property for the folding ranges. So it might be confusing if it doesn't work, even when there are block comments present in the file. And sometimes the built-in `fold_all` is a bit smarter, for example in JSON it only folds from the level of the current caret position (I haven't figured out the exact logic yet, and it seems to be syntax dependent).

jwortmann:
Another idea I had would be to add a new user setting like
```jsonc
{
  // Controls the initial state of folding regions when a document gets opened.
  // Note that only a few language servers have support for this.
  "initial_folding_collapsed_state": [ // TODO find a better name
    // "comment",
    // "imports",
    // "region",
  ],
}
```

Then we could run the request in `on_load_async` and fold the corresponding regions if so configured.

jwortmann:
> Another idea I had would be to add a new user setting like

Added in https://github.com/sublimelsp/LSP/pull/2304/commits/b5a9d4f587108578440c4bef9d825634942d0436, since there were no objections in the meanwhile ;)
And confirmed to work in a quick test with LSP-json and `"initially_folded": ["array"]` setting (any suggestions for a possibly better setting name are welcome).

plugin/folding_range.py
```diff
@@ -0,0 +1,205 @@
...
+        reverse=True
+    )
+
+
+class LspFoldCommand(LspTextCommand):
+    """A command to fold at the current caret position or at a given point.
+
+    Command arguments:
+
+    - `manual`: Should be `true` when invoked from the command palette or via key binding, and `false` for menu items.
```
rchl, line 47: The name of this one feels a little misleading.

Do I understand correctly that you think `manual=False` is appropriate when used within menu item because then `is_visible` and `description` are invoked without triggering the command explicitly? If so then maybe that makes sense but when clicking the menu item it's definitively a "manual" action but it's still invoked with `manual=False`. Would it be clearer if it was just called "from_menu_item"?

Default.sublime-keymap
```diff
@@ -213,6 +213,13 @@
     //     "command": "lsp_expand_selection",
     //     "context": [{"key": "lsp.session_with_capability", "operand": "selectionRangeProvider"}]
     // },
+    // Fold around caret position
+    // {
```
rchl, line 217: Can we add a line about what `strict` does?

Default.sublime-commands
```diff
@@ -162,4 +162,9 @@
         "caption": "LSP: Toggle Inlay Hints",
         "command": "lsp_toggle_inlay_hints",
     },
+    // {
+    //     "caption": "LSP: Fold All Block Comments",
```
rchl, line 166: I wonder if it's always true that it's only gonna apply to block comments. I don't think spec mentions "block" comments specifically.

jwortmann: True, I just copied the label from VS Code here. But I had a better idea, instead of "block comments" we can use "comment blocks", this should include both block comments and consecutive line comments, so it should be accurate (assuming that no server gets the idea to add inline folding for single-line comments, but that would be stupid and I think we can also rule that out due to VS Code having no support for inline folding at the moment).

plugin/folding_range.py
jwortmann: Yeah, I didn't think much about it and just used "manual" because it was already used in some other commands. I renamed it to "prefetch" and updated the docstring. In practice users should only care about the "strict" argument if they want to add a custom key binding or command palette entry.

Default.sublime-commands
```diff
@@ -162,4 +162,9 @@
         "caption": "LSP: Toggle Inlay Hints",
         "command": "lsp_toggle_inlay_hints",
     },
+    // {
+    //     "caption": "LSP: Fold All Comment Blocks",
+    //     "command": "lsp_fold_all",
+    //     "args": {"kind": "comment"}
+    // },
```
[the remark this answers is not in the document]
jwortmann: I'm a bit undecided about it, but my intention here was to keep it in the file as an example, so it can be used as a reference in case someone asks how to add such an entry into the command palette, or maybe we decide to activate in the future. I think it would potentially be more appropriate if an LSP-* helper package would add this command, e.g. as "LSP-javascript: Fold All Comment Blocks", because they know whether the server supports the `kind` or not. In that case it would be best to create a derived command like
```python
from LSP.plugin.folding_range import LspFoldAllCommand

class LspJavascripFoldAllCommand(LspFoldAllCommand):
    session_name = 'LSP-javascript'
```
and then use that command, so that it is only active for the particular server.

plugin/folding_range.py
```diff
@@ -0,0 +1,207 @@
...
+                    not strict and sublime.Region(self.view.line(region.a).a, region.b).contains(pt)) and \
+                    not self.view.is_folded(region):
+                # Store the relevant folding region, so that we don't need to do the same computation again in
+                # self.is_visible and self.run
+                self.folding_region = region
+                kind = folding_range.get('kind')
+                if kind == FoldingRangeKind.Imports:
+                    return "LSP: Fold Imports"
+                elif kind:
+                    return "LSP: Fold this {}".format(kind.title())  # pyright: ignore  # FoldingRangeKind -> StrEnum
```
jwortmann, line 130: @rchl Could you add `'FoldingRangeKind': 'StrEnum',` in https://github.com/sublimelsp/lsp-python-types/blob/3b91862c0ec61c4e248e3c60cb74a836798d0a53/generate.py#L12-L15

rchl: done

jwortmann: Thanks!

PR #2305, opened 2023-07-29
Fix CI failing on macOS

jwortmann:
This disables unit tests on Mac for now.

See https://github.com/sublimelsp/LSP/pull/2304#issuecomment-1656873774 and the Package Control issue linked in that comment.


rchl:
https://github.com/wbond/package_control/issues/1612 gave a better solution that I've applied in #2307

PR #2311, opened 2023-08-17
Feat/format on paste

### LINKED ISSUE #1094: Format-on-paste
It would be nice to format the just-inserted text when pasting code into the buffer with `lsp_format_range` (enabled with a boolean, of course).

predragnikolic:
closes #1094 

This PR introduces a new setting `format_on_paste` which defaults to `false`.

If enabled, 
it will [override](https://github.com/sublimelsp/LSP/compare/feat/format-on-paste?expand=1#diff-982c110867d9809c978a5597e4da66ce7538b327675a66b8f27214413b7c0ec1R488) the paste command  and execute the  `lsp_paste_and_format` command.

The `lsp_paste_and_format` command will do the following:
1, Insert the text form clipboard(also trying to mimic the original `paste` command behavior with multiple cursors)
2. Send a didChange notification
3. Send a textDocument/rangeFormatting or textDocument/rangesFormatting request (depending of the server support).

----

This feature doesn't work reliably,
because I cannot seem to figure out how to reliably notify the server of the new text changes, 
before sending the format range request ->
https://github.com/sublimelsp/LSP/compare/feat/format-on-paste?expand=1#diff-979aae35b78d493b209370b60e0d852a90d0214e31bafc18481faae1e4180337R24-R32


predragnikolic:
... and it seems that I somehow destroyed my Ubuntu OS by uninstalling python 3.11. and setting python 3.8 as default... I can't no longer boot my laptop (that said I can't use my laptop and make any changes to this PR, until I reinstall Linux, or buy a Mac) 😅

plugin/documents.py
```diff
@@ -482,6 +482,10 @@ def on_hover(self, point: int, hover_zone: int) -> None:
                     on_navigate=lambda href: self._on_navigate(href, point))
 
     def on_text_command(self, command_name: str, args: Optional[dict]) -> Optional[Tuple[str, dict]]:
+        if command_name == 'paste' and userprefs().format_on_paste:
+            session = self.session_async("documentRangeFormattingProvider")
+            if session:
+                return ("lsp_paste_and_format", {})
         if command_name == "auto_complete":
```
jwortmann, line 489: This can be `elif` instead of `if` now (or alternatively reorder the conditions and use `elif` in the new code).

plugin/paste_and_format.py
```diff
@@ -0,0 +1,33 @@
+from .core.registry import LspTextCommand
+from .core.typing import Optional
+from .core.views import first_selection_region
+import sublime
+
+
+class LspPasteAndFormatCommand(LspTextCommand):
```
jwortmann, line 7: Do I understand it correctly that this command is needed only because `paste` doesn't select the pasted text, but `self.view.replace` does so (if it's on a selection), and the non-empty selection(s) is/are needed for `lsp_format_document_range`?

If that is the case, wouldn't it be somewhat "cleaner" to not intercept the `paste` command, but instead to put the formatting logic into `on_post_text_command`? The selections could be derived from the caret positions after pasting and with the clipboard text. I wonder whether this would be simpler or more complicated. Then it wouldn't be necessary to reinvent/reimplement the `paste` command, and in general I'd be a bit hesitant to replace such an essential command like this one, because it might have some unintended side effects (for example what would happen if another package tries to do the same?)

plugin/documents.py
predragnikolic: I no longer use the on_text_command, instead I went ahead and implemented your second [comment](https://github.com/sublimelsp/LSP/pull/2311#discussion_r1297531796) to use on_post_text_command

plugin/paste_and_format.py
predragnikolic: > If that is the case, wouldn't it be somewhat "cleaner" to not intercept the paste command, but instead to put the formatting logic into on_post_text_command? 

Interesting idea. 

> The selections could be derived from the caret positions after pasting and with the clipboard text. I wonder whether this would be simpler or more complicated. 

The code for finding the inserted text [looks like this](https://github.com/sublimelsp/LSP/pull/2311/files#diff-982c110867d9809c978a5597e4da66ce7538b327675a66b8f27214413b7c0ec1R518-R527).

> .... it wouldn't be necessary to reinvent/reimplement the paste command, and in general I'd be a bit hesitant to replace such an essential command like this one

You are right. 
I removed the code that tried to re-implement the paste command.

The worst that can happen now is if we do not find the text, and add it to selection,
in which case the code will not be formatted. 

In my testing this worked well.
If you find an example where this doesn't function, please let me know.

I haven't tested multiple cursors formatting on paste,
because I don't know a server that supports the `textDocument/rangesFormatting` capability.

plugin/documents.py
```diff
@@ -507,6 +512,21 @@ def on_query_completions(self, prefix: str, locations: List[int]) -> Optional[su
...
+        self.purge_changes_async()
+        clipboard_text = sublime.get_clipboard()
+        sel = self.view.sel()
+        number_of_cursors = len(sel)
+        split_clipboard_text = clipboard_text.split('\n')
+        for index, region in enumerate(sel):
+            look_text = clipboard_text
+            if len(split_clipboard_text) == number_of_cursors:
+                look_text = split_clipboard_text[index]
+            found_region = self.view.find(look_text, region.end(), sublime.REVERSE)
```
predragnikolic, line 525: currently LSP supports ST 4070
sublime.REVERSE was introduced in build 4149
if this PR gets merged the last supported version needs to be 4149.

predragnikolic:
> This feature doesn't work reliably,

Well, it works reliably after this commit [26f2a13](https://github.com/sublimelsp/LSP/pull/2311/commits/26f2a139bff58a9104264773c3f05f63756d91ee).

I don't know why it didn't work reliably... or why it works reliably now

plugin/documents.py
```diff
@@ -511,6 +512,21 @@ def on_query_completions(self, prefix: str, locations: List[int]) -> Optional[su
...
+            if found_region:
+                sel.add(found_region)
+        self.view.run_command('lsp_format_document_range')
```
rchl, line 528: This will leave the pasted content selected, I believe. Normal paste doesn't do that.

plugin/documents.py
```diff
@@ -511,6 +512,21 @@ def on_query_completions(self, prefix: str, locations: List[int]) -> Optional[su
...
+        self.purge_changes_async()
+        clipboard_text = sublime.get_clipboard()
+        sel = self.view.sel()
+        number_of_cursors = len(sel)
+        split_clipboard_text = clipboard_text.split('\n')
+        for index, region in enumerate(sel):
+            look_text = clipboard_text
+            if len(split_clipboard_text) == number_of_cursors:
+                look_text = split_clipboard_text[index]
+            found_region = self.view.find(look_text, region.end(), sublime.REVERSE)
```
rchl, line 525: In

```ts
type MyCustomType = {
    foo: string;
    bar?: string;
}
```

If I copy and paste in the same place `bar?` then it will crash (and split selection into two).

I wonder if this approach can be made robust... It seems quite fragile.

rchl: And if the whole content is copied and pasted then it won't find the region and do no formatting.

plugin/documents.py
```diff
@@ -491,6 +491,11 @@ def on_text_command(self, command_name: str, args: Optional[dict]) -> Optional[T
         return None
 
     def on_post_text_command(self, command_name: str, args: Optional[Dict[str, Any]]) -> None:
+        if command_name == 'paste' and userprefs().format_on_paste:
+            # if we don't delay execution self.format_on_paste,
+            # on_text_changed_async will not pick up the change
+            # thus a call to self.purge_changes_async will not see new text changes.
```
rchl, line 497: Might be enough to say:

```suggestion
            # ensure on_text_changed_async is triggered before self.purge_changes_async()
```

plugin/documents.py
predragnikolic: [25b7806](https://github.com/sublimelsp/LSP/pull/2311/commits/25b78067a53dd54b09f65a6fcd0e67498d521263) should fix it

plugin/documents.py
predragnikolic: [7e1b2d8](https://github.com/sublimelsp/LSP/pull/2311/commits/7e1b2d8205dfccce45f11ca0b45906b8ea1149da)

plugin/documents.py
```diff
@@ -491,6 +491,9 @@ def on_text_command(self, command_name: str, args: Optional[dict]) -> Optional[T
         return None
 
     def on_post_text_command(self, command_name: str, args: Optional[Dict[str, Any]]) -> None:
+        if command_name == 'paste' and userprefs().format_on_paste:
```
jwortmann, line 494: There exists also a command `paste_and_indent` (<kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>V</kbd>).
Should it run the formatting too, when that command was invoked?

By the way, the next `if` can be `elif` now ;-)

plugin/documents.py
```diff
@@ -507,6 +510,27 @@ def on_query_completions(self, prefix: str, locations: List[int]) -> Optional[su
...
+        clipboard_text = sublime.get_clipboard()
+        sel = self.view.sel()
+        number_of_cursors = len(sel)
+        split_clipboard_text = clipboard_text.split('\n')
+        restore_regions = []
+        # add regions to selection, in order for lsp_format_document_range to format those regions
+        for index, region in enumerate(sel):
+            restore_regions.append(region)
+            look_text = clipboard_text
+            if len(split_clipboard_text) == number_of_cursors:
```
jwortmann, line 524:
```suggestion
        split_clipboard_text = clipboard_text.split('\n')
        multi_cursor_paste = len(sel) == len(split_clipboard_text)
        restore_regions = []
        # add regions to selection, in order for lsp_format_document_range to format those regions
        for index, region in enumerate(sel):
            restore_regions.append(region)
            look_text = clipboard_text
            if multi_cursor_paste:
```

plugin/documents.py
predragnikolic: >  Should it run the formatting too, when that command was invoked?

I would not add it, but don't have any strong opinions about it.
If you think that formatting should happen, I will add it.

predragnikolic:
Overall the behavior seems not satisfying as regular paste. 
I see flickering.

I also enabled format_on_paste in vs code
and was surprised to see the "same" flickering there. (I had a feeling that they did something smarter to avoid that flicker)

[Screencast from 2023-08-18 18-15-57.webm](https://github.com/sublimelsp/LSP/assets/22029477/36bad8c4-f66d-4e3c-81b4-b43b178bf53b)

I was thinking of how to avoid that flicker,
Currently the text first get inserted(first edit), than formatted(second edit) which are done fast, but not fast enough to not see the flicker.

I don't know if it would been better if the text first got formatted than inserted(one edit), that way we would avoid the flicker, but there will be a delay before a text is inserted... but there is no LSP request that would allow formatting text on the fly, whiteout first being inserted in the document.

jwortmann:
> I don't know if it would been better if the text first got formatted than inserted(one edit), that way we would avoid the flicker, but there will be a delay before a text is inserted... but there is no LSP request that would allow formatting text on the fly, whiteout first being inserted in the document.

Maybe it would be possible to create a "virtual" document with another (`buffer://`) URI, including the pasted text, let the server format that one and then derive from the formatted document what should be pasted into the actual document. But I imagine that the delay between key press and text insertion in that case would be even more annoying than the flickering right now. So I think I would not do this.

I can not really see what exactly happens in your example video, but another thing I noticed is, that when the pasted text gets inserted at a wrong indentation level (with too much indentation), then I would expect it to fix the indentation after formatting. But this is not possible via range formatting if the indentation is not contained within the selection (pasted text). So would it be better to extend the selection to the leading and trailing whitespace around the pasted text?

predragnikolic:
The server that was used in the video above was the json language server.

> So would it be better to extend the selection to the leading and trailing whitespace around the pasted text?

No. Even if I extend the selection, it will still not work correctly with the json langauge server. 

The reason why the text doesn't get formatted correctly is because the json language server thinks that it needs to be formatted like that. In other words, the server would need to adjust the logic for the formatting range.

----

If I paste at the wrong indent level with the typescript language server, the server will format it as you expect:
 
[Screencast from 2023-08-18 19-52-19.webm](https://github.com/sublimelsp/LSP/assets/22029477/1a59a37e-5440-47ea-91c2-920c865fbb94)

predragnikolic:
The reason why I marked it as draft is because I found an issue.

If I copy the whole code block bellow:
```jsonc
	{
			"name": "JS Template String Converter",
			"details": "https://github.com/predragnikolic/JS-Template-String-Converter",
			"labels": ["javascript", "typescript", "jsx", "vue", "svelte"],
			"releases": [
				{
					"sublime_text": ">=4070",
					"tags": true
				}
			]
		},
```
and paste in a file that has the following [content](https://github.com/predragnikolic/package_control_channel/blob/33ee555ecef34986cc0b1bca8bbdb57e8b613f21/repository/j.json)

```jsonc
...
			"releases": [
				{
					"sublime_text": "*",
					"branch": "master"
				}
			]
		},
// paste here at the wrong indent level,
		{
			"name": "Jaggeryjs",
			"details": "https://github.com/dakshika/jaggeryjs-sublime-text",
			"labels": ["language syntax", "snippets", "auto-complete"],
			"previous_names": ["Jaggery Snippets"],
			"releases": [
				{
					"sublime_text": "*",
					"branch": "master"
				}
			]
		},
		{
...
```

The following will be inseted:
```diff
			"releases": [
				{
					"sublime_text": "*",
					"branch": "master"
				}
			]
		},
+	{n
+	etails": "https://gi
+	abels": ["javascri
+ r
+	
+		
+	
+	}
+ ]
,
		{
			"name": "Jaggeryjs",
			"details": "https://github.
```

the reason for this is because the range request is still send before the didChange request...
```
:: [11:09:17.071] --> LSP-json textDocument/rangeFormatting (11): {'range': {'start': {'line': 81, 'character': 0}, 'end': {'line': 91, 'character': 4}}, 'workDoneToken': '$ublime-work-done-progress-11', 'textDocument': {'uri': 'file:///home/predragnikolic/Documents/sandbox/package_control_channel/repository/j.json'}, 'options': {'tabSize': 4, 'trimFinalNewlines': False, 'trimTrailingWhitespace': True, 'insertSpaces': False, 'insertFinalNewline': False}}
:: [11:09:17.073] <<< LSP-json (11) (duration: 2ms): [{'range': {'start': {'line': 81, 'character': 0}, 'end': {'line': 82, 'character': 2}}, 'newText': ''}, {'range': {'start': {'line': 82, 'character': 3}, 'end': {'line': 83, 'character': 3}}, 'newText': '\n\t'}, {'range': {'start': {'line': 83, 'character': 22}, 'end': {'line': 84, 'character': 3}}, 'newText': '\n\t'}, {'range': {'start': {'line': 84, 'character': 22}, 'end': {'line': 84, 'character': 22}}, 'newText': '\n\t\t'}, {'range': {'start': {'line': 84, 'character': 27}, 'end': {'line': 84, 'character': 27}}, 'newText': '\n\t'}, {'range': {'start': {'line': 84, 'character': 29}, 'end': {'line': 85, 'character': 3}}, 'newText': '\n\t'}, {'range': {'start': {'line': 85, 'character': 55}, 'end': {'line': 86, 'character': 3}}, 'newText': '\n\t'}, {'range': {'start': {'line': 86, 'character': 16}, 'end': {'line': 87, 'character': 4}}, 'newText': '\n\t\t'}, {'range': {'start': {'line': 87, 'character': 5}, 'end': {'line': 88, 'character': 5}}, 'newText': '\n\t\t\t'}, {'range': {'start': {'line': 88, 'character': 30}, 'end': {'line': 89, 'character': 5}}, 'newText': '\n\t\t\t'}, {'range': {'start': {'line': 89, 'character': 17}, 'end': {'line': 90, 'character': 4}}, 'newText': '\n\t\t'}, {'range': {'start': {'line': 90, 'character': 5}, 'end': {'line': 91, 'character': 3}}, 'newText': '\n\t'}]
:: [11:09:17.383]  -> LSP-json textDocument/didChange: {'textDocument': {'uri': 'file:///home/predragnikolic/Documents/sandbox/package_control_channel/repository/j.json', 'version': 203}, 'contentChanges': [{'rangeLength': 0, 'range': {'start': {'line': 81, 'character': 0}, 'end': {'line': 81, 'character': 0}}, 'text': '\t{\n\t\t\t"name": "JS Template String Converter",\n\t\t\t"details": "https://github.com/predragnikolic/JS-Template-String-Converter",\n\t\t\t"labels": ["javascript", "typescript", "jsx", "vue", "svelte"],\n\t\t\t"releases": [\n\t\t\t\t{\n\t\t\t\t\t"sublime_text": ">=4070",\n\t\t\t\t\t"tags": true\n\t\t\t\t}\n\t\t\t]\n\t\t},'}, {'rangeLength': 4, 'range': {'start': {'line': 90, 'character': 4}, 'end': {'line': 91, 'character': 3}}, 'text': '\n\t'}, {'rangeLength': 5, 'range': {'start': {'line': 89, 'character': 5}, 'end': {'line': 90, 'character': 4}}, 'text': '\n\t\t'}, {'rangeLength': 6, 'range': {'start': {'line': 88, 'character': 17}, 'end': {'line': 89, 'character': 5}}, 'text': '\n\t\t\t'}, {'rangeLength': 31, 'range': {'start': {'line': 87, 'character': 5}, 'end': {'line': 88, 'character': 5}}, 'text': '\n\t\t\t'}, {'rangeLength': 5, 'range': {'start': {'line': 86, 'character': 5}, 'end': {'line': 87, 'character': 4}}, 'text': '\n\t\t'}, {'rangeLength': 4, 'range': {'start': {'line': 85, 'character': 16}, 'end': {'line': 86, 'character': 3}}, 'text': '\n\t'}, {'rangeLength': 41, 'range': {'start': {'line': 84, 'character': 29}, 'end': {'line': 85, 'character': 3}}, 'text': '\n\t'}, {'rangeLength': 0, 'range': {'start': {'line': 84, 'character': 27}, 'end': {'line': 84, 'character': 27}}, 'text': '\n\t'}, {'rangeLength': 0, 'range': {'start': {'line': 84, 'character': 22}, 'end': {'line': 84, 'character': 22}}, 'text': '\n\t\t'}, {'rangeLength': 61, 'range': {'start': {'line': 83, 'character': 22}, 'end': {'line': 84, 'character': 3}}, 'text': '\n\t'}, {'rangeLength': 43, 'range': {'start': {'line': 82, 'character': 3}, 'end': {'line': 83, 'character': 3}}, 'text': '\n\t'}, {'rangeLength': 5, 'range': {'start': {'line': 81, 'character': 0}, 'end': {'line': 82, 'character': 2}}, 'text': ''}]}
```

if I increase the delay:
```
- sublime.set_timeout_async(self._format_on_paste, 1)
+ sublime.set_timeout_async(self._format_on_paste, 10)
```
it will didChange will be send before the range request...
I do not want to increase the delay. It seems as a  workaround and probably will not work for slower laptops.

I will try to see if I can make sure that the range request is send only after the didChange notification is send.

plugin/documents.py
```diff
@@ -917,6 +923,27 @@ def _update_stored_selection_async(self) -> Tuple[Optional[sublime.Region], bool
...
+        restore_regions = []
+        # add regions to selection, in order for lsp_format_document_range to format those regions
+        for index, region in enumerate(sel):
+            restore_regions.append(region)
+            look_text = clipboard_text
+            if multi_cursor_paste:
+                look_text = split_clipboard_text[index]
+            found_region = self.view.find(look_text, region.end(), sublime.REVERSE | sublime.LITERAL)
+            if found_region:
+                sel.add(found_region)
```
rchl, line 941: Shouldn't be appending to the list while enumerating it.

plugin/documents.py
```diff
@@ -917,6 +923,27 @@ def _update_stored_selection_async(self) -> Tuple[Optional[sublime.Region], bool
         self._stored_selection = selection
         return changed_first_region, True
 
+    def _format_on_paste(self) -> None:
+        self.purge_changes_async()
+        clipboard_text = sublime.get_clipboard()
+        sel = self.view.sel()
+        split_clipboard_text = clipboard_text.split('\n')
+        multi_cursor_paste = len(split_clipboard_text) == len(sel) > 1
+        restore_regions = []
```
rchl, line 932: naming: maybe call "original_selection"
Also I suppose it can be just `original_selection = list(sel)`

plugin/documents.py
```diff
@@ -917,6 +923,27 @@ def _update_stored_selection_async(self) -> Tuple[Optional[sublime.Region], bool
...
+        for index, region in enumerate(sel):
+            restore_regions.append(region)
+            look_text = clipboard_text
+            if multi_cursor_paste:
+                look_text = split_clipboard_text[index]
+            found_region = self.view.find(look_text, region.end(), sublime.REVERSE | sublime.LITERAL)
+            if found_region:
+                sel.add(found_region)
+        self.view.run_command('lsp_format_document_range')
+        # restore selection how it was
```
rchl, line 943: If the "restore_regions" would be renamed to "original_selection" then I think that the code would be self-documenting then.

plugin/documents.py
```diff
@@ -158,6 +158,7 @@ def on_change() -> None:
         self._registration = SettingsRegistration(view.settings(), on_change=on_change)
         self._completions_task = None  # type: Optional[QueryCompletionsTask]
         self._stored_selection = []  # type: List[sublime.Region]
+        self.paste_was_run = False
```
rchl, line 161: Private properties should start with an underscore.
Also maybe call it `_did_paste`

plugin/documents.py
```diff
@@ -917,6 +923,27 @@ def _update_stored_selection_async(self) -> Tuple[Optional[sublime.Region], bool
...
+            if multi_cursor_paste:
+                look_text = split_clipboard_text[index]
+            found_region = self.view.find(look_text, region.end(), sublime.REVERSE | sublime.LITERAL)
+            if found_region:
+                sel.add(found_region)
+        self.view.run_command('lsp_format_document_range')
+        # restore selection how it was
+        sel.clear()
+        sel.add_all(iter(restore_regions))
+
```
rchl, line 946: We can probably avoid all this if the selection only contains empty regions.

plugin/documents.py
```diff
@@ -917,6 +923,27 @@ def _update_stored_selection_async(self) -> Tuple[Optional[sublime.Region], bool
         self._stored_selection = selection
         return changed_first_region, True
 
+    def _format_on_paste(self) -> None:
+        self.purge_changes_async()
+        clipboard_text = sublime.get_clipboard()
+        sel = self.view.sel()
+        split_clipboard_text = clipboard_text.split('\n')
+        multi_cursor_paste = len(split_clipboard_text) == len(sel) > 1
```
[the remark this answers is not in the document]
jwortmann: I wrote this line :-)
I thought that it would be more readable... this line is two comparisons with two numbers each, like you would write in maths. It does the same as `len(split_clipboard_text) == len(sel) and len(sel) > 1`, but with `len(sel)` only evaluated a single time.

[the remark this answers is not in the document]
rchl: I see. So it's a Python `chained comparison` feature ([ref](https://stackoverflow.com/questions/57400620/python-comparison-operator-precedence)). Sorry but I don't like it. It's not clear from the code if it's intentional or it's just a bug in the code and all that just for optimization that doesn't even matter in this case :)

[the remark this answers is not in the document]
jwortmann: I agree that the optimization doesn't really matter here, but it's much more readable than the long form, no?
What would it make more clear that it's not a bug? We could rename the variable to `has_multiple_selections_and_the_same_number_of_insertions`...
Do we need some kind of contribution guidelines with Python features which are allowed or are not allowed to be used? Otherwise we must assume that every line can be a bug....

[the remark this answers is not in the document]
rchl: I suppose it's just a "me" issue and you can ignore this suggestion but I just really don't like this Python feature.

Probably because it looks just like a series of comparisons but behaves differently than any other language I know and use.

`len(sel) > 1 and len(split_clipboard_text) == len(sel)` would look so much cleaner to **me**.

jwortmann:
> The reason why the text doesn't get formatted correctly is because the json language server thinks that it needs to be formatted like that. In other words, the server would need to adjust the logic for the formatting range.
> 
> If I paste at the wrong indent level with the typescript language server, the server will format it as you expect:

This is interesting, and it surprises me a bit. I have tested and can confirm your observations. It disproves my assumption that the changes from the server response are restricted to the given range (https://github.com/sublimelsp/LSP/pull/2299#discussion_r1282733696). But maybe it's just an inaccuracy of the TypeScript server, or the LSP specs are not intended to be taken literally in this point - besides that, the server's behavior is probably useful from the user perspective.

However, for the JSON server I found an example where inclusion of the whitespace in the formatting range *does* affect the result, namely when the opening bracket/brace is on the same line as the formatting range (even when it's not included in the range):

[format.webm](https://github.com/sublimelsp/LSP/assets/6579999/ea30b70b-ab66-4291-9451-9dc7d36ed8e1)

It's not in the video, but the same behavior can be observed in a similar case when another key-value pair is preceding on the same line; formatting only works if whitespace is included in the formatting range:
```json
{
    "bar": 0,    "foo": 1,
    "ads": 3211
}
```

So I think it would be advantageous to include leading and trailing whitespace. I haven't tested with any other servers, but I could imagine that there are other cases where the formatting on paste also needs the leading/surrounding whitespace in the formatting range to work reliably.

plugin/documents.py
```diff
@@ -917,6 +923,27 @@ def _update_stored_selection_async(self) -> Tuple[Optional[sublime.Region], bool
...
+            look_text = clipboard_text
+            if multi_cursor_paste:
+                look_text = split_clipboard_text[index]
+            found_region = self.view.find(look_text, region.end(), sublime.REVERSE | sublime.LITERAL)
+            if found_region:
+                sel.add(found_region)
+        self.view.run_command('lsp_format_document_range')
+        # restore selection how it was
+        sel.clear()
+        sel.add_all(iter(restore_regions))
```
jwortmann, line 945: Maybe something like this (continuing from my last comment):

```suggestion
    def _format_on_paste(self) -> None:
        self.purge_changes_async()
        clipboard_text = sublime.get_clipboard()
        sel = self.view.sel()
        original_selection = list(sel)
        split_clipboard_text = clipboard_text.split('\n')
        multi_cursor_paste = len(split_clipboard_text) == len(sel) > 1
        pasted_text = clipboard_text
        formatting_regions = []  # type: List[sublime.Region]
        # add regions to selection, in order for lsp_format_document_range to format those regions
        for index, region in enumerate(sel):
            if multi_cursor_paste:
                pasted_text = split_clipboard_text[index]
            pasted_region = self.view.find(pasted_text, region.end(), sublime.REVERSE | sublime.LITERAL)
            if pasted_region:
                # Expand to preceding whitespace on the same line
                a = self.view.find_by_class(pasted_region.a, False,
                        sublime.CLASS_WORD_END | sublime.CLASS_PUNCTUATION_END | sublime.CLASS_LINE_START)
                formatting_region = sublime.Region(a, pasted_region.b)
                formatting_regions.append(formatting_region)
        sel.add_all(iter(formatting_regions))
        self.view.run_command('lsp_format_document_range')
        sel.clear()
        sel.add_all(iter(original_selection))
```

I couldn't figure out how to expand to the trailing whitespace on the same line, because `sublime.CLASS_LINE_END` doesn't seem to work when I tried it.

predragnikolic:
I haven't forgotten about the PRs.
Thanks for the feedback!

I will address them as soon as I find the time and will 🙂

predragnikolic:
I will continue this in https://github.com/sublimelsp/LSP/pull/2397 because I didn't want to resolve merge conflicts.

PR #2328, opened 2023-09-22
Handle multiple formatters

### LINKED ISSUE #2150: Handle multiple formatters
If multiple sessions with formatting capability are active then we should maybe allow the user to select which one to use.

But it could get annoying to select the formatter each time so maybe preferred choice should be stored somewhere then.
And then there should maybe exist an option to change the preferred choice.

VSCode for reference:
Screenshot 2022-12-22 at 20 18 54

jwortmann:
Closes #2150.

### DIFF
--- a/plugin/formatting.py
+++ b/plugin/formatting.py
@@ -128,8 +128,8 @@ class LspFormatDocumentCommand(LspTextCommand):
                 session = self.session_by_name(formatter, self.capability)
                 if session:
                     session.send_request_task(text_document_formatting(self.view)).then(self.on_result)
-            else:
-                self.select_formatter(base_scope, session_names)
+                    return
+            self.select_formatter(base_scope, session_names)
         else:
             format_document(self).then(self.on_result)
 
--- a/plugin/formatting.py
+++ b/plugin/formatting.py
@@ -1,3 +1,4 @@
+from .core.collections import DottedDict
 from .core.edit import parse_text_edit
 from .core.promise import Promise
 from .core.protocol import Error
@@ -109,7 +110,7 @@ class LspFormatDocumentCommand(LspTextCommand):
 
     def run(self, edit: sublime.Edit, event: Optional[dict] = None, select: bool = False) -> None:
         session_names = [session.config.name for session in self.sessions(self.capability)]
-        base_scope = self.view.scope_name(0).split()[0]
+        base_scope = self.view.syntax().scope
         if select:
             self.select_formatter(base_scope, session_names)
         elif len(session_names) > 1:
@@ -119,7 +120,7 @@ class LspFormatDocumentCommand(LspTextCommand):
             formatter = None
             project_data = window.project_data()
             if isinstance(project_data, dict):
-                formatter = project_data.get('settings', {}).get('LSP', {}).get('formatters', {}).get(base_scope)
+                formatter = DottedDict(project_data).get('settings.LSP.formatters.{}'.format(base_scope))
             else:
                 window_manager = windows.lookup(window)
                 if window_manager:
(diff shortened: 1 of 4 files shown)


sublime-package.json
```diff
@@ -780,6 +780,14 @@
                 "LSP": {
                   "type": "object",
                   "markdownDescription": "The dictionary of your configured language servers or overrides for existing configurations. The keys of this dictionary are free-form. They give a humany-friendly name to the server configuration. They are shown in the bottom-left corner in the status bar once attached to a view (unless you have `\"show_view_status\"` set to `false`).",
+                  "properties": {
+                    "formatters": {
```
jwortmann, line 784: I guess we can assume that there will be no client config with this name.

plugin/formatting.py
```diff
@@ -100,16 +102,70 @@ class LspFormatDocumentCommand(LspTextCommand):
...
+    def is_enabled(self, event: Optional[dict] = None, select: bool = False) -> bool:
+        if select:
+            return len(list(self.sessions(self.capability))) > 1
         return super().is_enabled() or bool(self.best_session(LspFormatDocumentRangeCommand.capability))
 
-    def run(self, edit: sublime.Edit, event: Optional[dict] = None) -> None:
-        format_document(self).then(self.on_result)
+    def run(self, edit: sublime.Edit, event: Optional[dict] = None, select: bool = False) -> None:
+        session_names = [session.config.name for session in self.sessions(self.capability)]
+        base_scope = self.view.scope_name(0).split()[0]
```
rchl, line 112: In other parts of the code we are using `view.syntax().scope`

plugin/formatting.py
```diff
@@ -100,16 +102,70 @@ class LspFormatDocumentCommand(LspTextCommand):
...
+        if select:
+            self.select_formatter(base_scope, session_names)
+        elif len(session_names) > 1:
+            window = self.view.window()
+            if not window:
+                return
+            formatter = None
+            project_data = window.project_data()
+            if isinstance(project_data, dict):
+                formatter = project_data.get('settings', {}).get('LSP', {}).get('formatters', {}).get(base_scope)
```
rchl, line 122: We could instead do:

```suggestion
                formatter = DottedDict(project_data).get('settings.LSP.formatters.{}'.format(base_scope))
```

plugin/formatting.py
```diff
@@ -100,16 +102,70 @@ class LspFormatDocumentCommand(LspTextCommand):
...
+            else:
+                window_manager = windows.lookup(window)
+                if window_manager:
```
rchl, line 125: Can we do this in place of `window = self.view.window()` already? In theory we should always go through WindowManager to verify that LSP is still operational.

plugin/formatting.py
```diff
@@ -100,16 +102,70 @@ class LspFormatDocumentCommand(LspTextCommand):
...
 
+    def select_formatter(self, base_scope: str, session_names: List[str]) -> None:
+        window = self.view.window()
+        if not window:
+            return
+        window.show_quick_panel(
+            session_names, partial(self.on_select_formatter, base_scope, session_names), placeholder="Select Formatter")
+
+    def on_select_formatter(self, base_scope: str, session_names: List[str], index: int) -> None:
+        if index > -1:
```
rchl, line 148: Would be IMO nicer to return early in this case to minimize indentation.

plugin/formatting.py
```diff
@@ -100,16 +102,70 @@ class LspFormatDocumentCommand(LspTextCommand):
...
+                if window_manager:
+                    project_data = window.project_data()
+                    if isinstance(project_data, dict):
+                        project_settings = project_data.setdefault('settings', dict())
+                        project_lsp_settings = project_settings.setdefault('LSP', dict())
+                        project_formatter_settings = project_lsp_settings.setdefault('formatters', dict())
+                        project_formatter_settings[base_scope] = session_name
+                        # Prevent restart of all sessions after project file save
+                        window_manager.formatter_updated_in_project = True
+                        window.set_project_data(project_data)
```
rchl, line 162: DottedDict should also help to make this shorter here?

sublime-package.json
rchl: Feels weird to mix client configurations and other settings but I suppose this could be considered a bad initial design...

rchl (commented):
I think the `FormattingTask` task (the `lsp_format_on_save` setting) should also consider the stored formatter preference.

If nothing is stored yet and it's ambiguous which formatter to use then maybe use first like now and post a status message saying "No formatter selecting, using default"

plugin/formatting.py
jwortmann: We still need the window for `window.project_data()` a few lines above. So I don't think it would simplify the code here. And this `if window_manager` is just to satisfy pyright type checking, otherwise it shouldn't really be needed because the command wouldn't run anyway (`is_enabled` would return false) if there is no window_manager / LSP active. So I think it's most useful to defer this `lookup` line into the else branch and only do that function call if necessary. Or did you mean something else?

plugin/formatting.py
jwortmann: Can it? You need to still store the `project_data` as a normal Python dict for `set_project_data`, and DottedDict doesn't have any method to export to dict, because it would be ambiguous.

I just copied this code from https://github.com/sublimelsp/LSP/blob/fbd9652438e0b3b71362493ad6d9c6c2b9dce55f/plugin/core/workspace.py#L141-L146

plugin/formatting.py
rchl: We can get window from `wm.window`.

And yes, you are right that in that particular case it shouldn't be possible for window to be missing but it wouldn't hurt to always assume that it can instead of picking on a case by case basis whether we should be extra safe.

plugin/formatting.py
rchl: `DottedDict.get()` without argument will return the internal dict. There is also a `DottedDict.copy()` that will return a copy.

plugin/formatting.py
jwortmann: Moved to a separate function. I think it should be okay now.

plugin/formatting.py
jwortmann: I see. But it still won't work because we can't set the base scopes as entire keys in the DottedDict, because they contain a dot in the scope name. If we would try that, they would be splitted like
```json
{
  "formatters": {
    "source": {
      "js": "LSP-typescript"
    }      
  }
}
```
when saving in the project file.

sublime-package.json
jwortmann: Alternatively we could use a different top-level key under "settings", for example "lsp_formatters". But I'd prefer to keep all LSP settings together under "LSP". And even if there were a config named "formatters" everything should still work fine, because the settings from that config and the formatter scope keys stored under the same name wouldn't influence each other.

rchl (approved):

predragnikolic:
Can you add a section in the docs about the command and setting?

jwortmann:
> Can you add a section in the docs about the command and setting?

Well, the command is the same as before (`lsp_format_document`), now it just has a new optional `select` argument, following the design of the built-in `build` command (Tools > Build / Build With...). This is not really intended to be used for key bindings or such, it's just something that automatically shows up when you trigger document formatting at the first time with multiple language servers running, and then after you selected your preferenced formatter you can forget about it (more or less at least, as it's "only" saved per project or for the running session if there is no project). Similarly, the setting in the project file is also not really intended to be edited manually.

So I don't think there is anything which needs to be added to the docs. Or is there anything special you had in mind?

rchl:
I think it's fine. The new "setting" is not really meant to be manually controlled anyway.

PR #2339, opened 2023-10-15
Handle custom URI schemes in hover text links

rwols:
Before this change, custom URI schemes would be delegated to the browser.

With this change, an attempt is made to look for a session that can handle the URI.

Moreover, if there is a fragment present in the URI, then it's assumed that that fragment encodes the (row, col) to jump to. This logic is somewhat dubious, but it's the only reasonable way (as far as I can see) to handle this.

Anoher approach could be to introduce a new on_open_uri2_async callback for plugins where they can also provide a (row, col) to jump to. But, then we would also require an opt-in switch for plugins where they would advertise that they can handle the "version 2" of on_open_uri.

(I've attached a GIF of goto-definition now working even for hover text links with Java, but I don't think the mouse cursor is visible)

jdtls-goto-def

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1622,13 +1622,13 @@ class Session(TransportCallbacks):
         return self._maybe_resolve_code_action(code_action) \
             .then(lambda code_action: self._apply_code_action_async(code_action, view))
 
-    def open_uri_async(
+    def try_open_uri_async(
         self,
         uri: DocumentUri,
         r: Optional[Range] = None,
         flags: int = 0,
         group: int = -1
-    ) -> Promise[Optional[sublime.View]]:
+    ) -> Optional[Promise[Optional[sublime.View]]]:
         if uri.startswith("file:"):
             return self._open_file_uri_async(uri, r, flags, group)
         # Try to find a pre-existing session-buffer
@@ -1642,7 +1642,19 @@ class Session(TransportCallbacks):
         # There is no pre-existing session-buffer, so we have to go through AbstractPlugin.on_open_uri_async.
         if self._plugin:
             return self._open_uri_with_plugin_async(self._plugin, uri, r, flags, group)
-        return Promise.resolve(None)
+        return None
+
+    def open_uri_async(
+        self,
+        uri: DocumentUri,
+        r: Optional[Range] = None,
+        flags: int = 0,
+        group: int = -1
+    ) -> Promise[Optional[sublime.View]]:
+        promise = self.try_open_uri_async(uri, r, flags, group)
+        if promise is None:
+            raise RuntimeError("unexpected URI scheme")
+        return promise
 
     def _open_file_uri_async(
         self,
@@ -1668,7 +1680,7 @@ class Session(TransportCallbacks):
         r: Optional[Range],
         flags: int,
         group: int,
-    ) -> Promise[Optional[sublime.View]]:
+    ) -> Optional[Promise[Optional[sublime.View]]]:
         # I cannot type-hint an unpacked tuple
         pair = Promise.packaged_task()  # type: PackagedTask[Tuple[str, str, str]]
         # It'd be nice to have automatic tuple unpacking continuations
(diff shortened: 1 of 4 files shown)


plugin/hover.py
```diff
@@ -366,3 +370,15 @@ def run_async() -> None:
...
+        row, col_utf16 = row_col_from_uri_fragment(href)
+        if isinstance(row, int):
+            position = {"line": row, "character": col_utf16 or 0}  # type: Position
+            r = {"start": position, "end": position}  # type: Optional[Range]
+        else:
+            r = None
+        for session in self.sessions():
+            promise = session.open_uri_async(href, r)
+            if not promise.resolved or isinstance(promise.value, sublime.View):
+                return
```
rwols, line 384: I'm a bit unsure about the correctness of this promise state check. I need to check whether a session will handle the URI. I check that by looking whether the promise is resolved immediately with a None result.

rwols:
Related: https://github.com/sublimelsp/LSP-jdtls/pull/43

plugin/core/views.py
```diff
@@ -997,6 +998,34 @@ def is_location_href(href: str) -> bool:
     return href.startswith("location:")
 
 
+def starts_with_custom_uri_scheme(href: str) -> bool:
+    return urlparse(href).scheme.lower() not in ("", "http", "https")
+
+
+def row_col_from_uri_fragment(href: str) -> Tuple[Optional[int], Optional[int]]:
```
jwortmann, line 1005: I've only skimmed over the code for now, but iirc I had already implemented the row/col parsing in https://github.com/sublimelsp/LSP/blob/c06d811dc0bd9f1176f349ac6f82ae84dba94a67/plugin/core/open.py#L28-L47
which should cover all combinations that are supported by VS Code. Could that be reused after moving this local function to somewhere else to make it available from the outside?

rwols: Nice, I missed that functionality. I’ll reuse that.

plugin/hover.py
rchl: It seems not that easy to handle that due to limitations of the current Promise implementation. If it supported `catch` then we could chain all and throw when current session has handled it.

But something like this might work (untested).

```py
        promise = Promise.resolve(False)
        for session in self.sessions():
            promise = promise.then(partial(self.try_open_next, session, href, r))

    def try_open_next(self, session: Session, href: str, r: Optional[Range], was_handled: bool) -> Promise[bool]:
        if was_handled:
            return Promise.resolve(True)
        return session.open_uri_async(href, r).then(lambda view: bool(view))
```

rwols: I solved it by introducing another method `try_open_uri_async` that returns `None` when it can't handle the uri scheme.

plugin/core/views.py
```diff
@@ -984,6 +985,10 @@ def is_location_href(href: str) -> bool:
     return href.startswith("location:")
 
 
+def starts_with_custom_uri_scheme(href: str) -> bool:
```
jwortmann, line 988: With this function name, shouldn't it exclude at least `file`, `res` and `subl` schemes too?

plugin/core/sessions.py
```diff
@@ -1642,7 +1642,19 @@ def open_uri_async(
...
+    def open_uri_async(
+        self,
+        uri: DocumentUri,
+        r: Optional[Range] = None,
+        flags: int = 0,
+        group: int = -1
+    ) -> Promise[Optional[sublime.View]]:
+        promise = self.try_open_uri_async(uri, r, flags, group)
+        if promise is None:
+            raise RuntimeError("unexpected URI scheme")
```
rchl, line 1656: There is code like https://github.com/sublimelsp/LSP/blob/07ef2a05e55034a222e6f2f4da13b463a51ac417/plugin/locationpicker.py#L31-L35 and https://github.com/sublimelsp/LSP/blob/757908ad385d22d31c3dd69c6a898fcb50758571/plugin/core/registry.py#L243-L250 that did show sublime message on unhandled schema and I guess it will no longer trigger and we'll get an exception instead.

I must say that I feel a bit weird about this returning either `None` or `Promise[None]`. Would be more consistent with always being a Promise like before...

plugin/core/views.py
rwols: I inlined the function

plugin/core/views.py
```diff
@@ -1,3 +1,4 @@
+from urllib.parse import urlparse
```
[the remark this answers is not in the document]
rwols: Removed :)

plugin/core/sessions.py
rwols: This is a good point, I'll have to study this again.

rwols: I restored the original behavior.

rchl (approved):

PR #2341, opened 2023-10-17
Add support for remote images in hover popups

### LINKED ISSUE #1363: Hover popup does not render images from the internet
Description

The window that appears when you hover over when looking for documentation about a class/object/function is not rendering images correctly on latest ST4:

The rendered images should be as follows:

Steps to reproduce

First step: hover over any class/object/function (that has images in its documentation) to view the documentation
Second step: Scroll to the part of the documentation that has images in it
Images don't show
Expected behavior

Images should show on the documentation window

Actual behavior

Images show as missing (missing icon)

Environment

Build: 4087
Operating system and version: MacOS 10.15.7

jwortmann:
Fixes #1363.

I have tested it with a small modification to LSP-pyright:
```diff
--- Installed Packages/LSP-pyright/plugin.py
+++ Packages/LSP-pyright/plugin.py
@@ -94,6 +94,7 @@
             contents = hover.get("contents")
             if isinstance(contents, dict) and contents.get("kind") == "markdown":
                 contents["value"] = self.patch_markdown_content(contents["value"])
+                contents["value"] += "\n\n![](https://flutter.github.io/assets-for-api-docs/assets/material/Colors.pinkAccent.png)"
             return
         if method == "completionItem/resolve" and isinstance(response.result, dict):
             completion = cast(CompletionItem, response.result)
```

### DIFF
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -37,6 +37,7 @@ from .core.views import update_lsp_popup
 from .session_view import HOVER_HIGHLIGHT_KEY
 from functools import partial
 import html
+import mdpopups
 import sublime
 
 
@@ -99,6 +100,7 @@ class LspHoverCommand(LspTextCommand):
     def __init__(self, view: sublime.View) -> None:
         super().__init__(view)
         self._base_dir = None   # type: Optional[str]
+        self._image_resolver = None
 
     def run(
         self,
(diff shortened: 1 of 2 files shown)


rwols (approved):

plugin/hover.py
```diff
@@ -311,6 +313,13 @@ def _show_hover(self, listener: AbstractViewListener, point: int, only_diagnosti
                     location=point,
                     on_navigate=lambda href: self._on_navigate(href, point),
                     on_hide=lambda: self.view.erase_regions(HOVER_HIGHLIGHT_KEY))
+            self._image_resolver = mdpopups.resolve_images(
+                contents, mdpopups.worker_thread_resolver, partial(self._on_images_resolved, contents))
+
+    def _on_images_resolved(self, original_contents: str, contents: str) -> None:
+        self._image_resolver = None
+        if contents != original_contents and self.view.is_popup_visible():
+            update_lsp_popup(self.view, contents)
```
rchl, line 322: It looks like this could trigger a situation like:
 - the popup is shown
 - images start resolving
 - the original popup is closed and a different one is opened
 - the original resolve completes and replaces the new popup with original contents.

Should we have some safety measures to ignore resolved results that are "outdated"?

jwortmann: Here is a simple way to test it with Pyright:
```python
def show_cat():
    """ Cute cat photo on hover.
    
    ![](https://upload.wikimedia.org/wikipedia/commons/8/8e/Hauskatze_langhaar.jpg)
    """
    pass
```

I guess it depends on your internet connection speed, but on my end the delay is negligible, in the way that I doubt you can open another popup in the meanwhile. The image size is ~1MB.

PR #2343, opened 2023-10-18
fix: Code Actions work with capability with selector register

### LINKED ISSUE #2342: Lsp-Omnisharp Code Actions does not work
**Describe the bug**
as title say

**To Reproduce**
1. Install Lsp and Lsp-Omnisharp
2. Enable Omnisharp Lsp Server
3. using xxx; Code Actions
4. Does not work

**Expected behavior**
change document to add `using xxx`

- OS: Arch
- Sublime Text version: 4152
- LSP version: 1.25.0
- Language servers used: Lsp-Omnisharp

**Additional context**

`self.has_capability("codeActionProvider.resolveProvider")` False
image

CodeActionProvider register with selector. So `self.has_capability` check failed

image

suiyuex:
fix #2342

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1338,6 +1338,9 @@ class Session(TransportCallbacks):
         """
         yield from self._session_buffers
 
+    def get_session_buffer_by_id(self, buffer_id: int) -> SessionBufferProtocol:
+        return next(filter(lambda buffer: buffer.is_self(buffer_id), self._session_buffers), None)
+
     def get_session_buffer_for_uri_async(self, uri: DocumentUri) -> Optional[SessionBufferProtocol]:
         scheme, path = parse_uri(uri)
         if scheme == "file":
@@ -1619,7 +1622,8 @@ class Session(TransportCallbacks):
         # A code action can have an edit and/or command. Note that it can have *both*. In case both are present, we
         # must apply the edits before running the command.
         code_action = cast(CodeAction, code_action)
-        return self._maybe_resolve_code_action(code_action) \
+        session_buffer = self.get_session_buffer_by_id(view.buffer_id())
+        return self._maybe_resolve_code_action(code_action, session_buffer) \
             .then(lambda code_action: self._apply_code_action_async(code_action, view))
 
     def open_uri_async(
@@ -1704,9 +1708,9 @@ class Session(TransportCallbacks):
         if self._plugin:
             self._plugin.on_session_buffer_changed_async(session_buffer)
 
-    def _maybe_resolve_code_action(self, code_action: CodeAction) -> Promise[Union[CodeAction, Error]]:
-        if "edit" not in code_action and self.has_capability("codeActionProvider.resolveProvider"):
-            # TODO: Should we accept a SessionBuffer? What if this capability is registered with a documentSelector?
+    def _maybe_resolve_code_action(self, code_action: CodeAction, session_buffer: SessionBufferProtocol) -> Promise[Union[CodeAction, Error]]:
+        has_capability =  self.has_capability("codeActionProvider.resolveProvider") or session_buffer.has_capability("codeActionProvider.resolveProvider")
+        if "edit" not in code_action and  has_capability:
             # We must first resolve the command and edit properties, because they can potentially be absent.
             request = Request("codeAction/resolve", code_action)
             return self.send_request_task(request)
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -185,6 +185,9 @@ class SessionBuffer:
                 return view
         return next(iter(self.session_views)).view
 
+    def is_self(self, id: int) -> bool:
+        return self._id == id;
+
     @property
     def language_id(self) -> str:
         """


plugin/core/sessions.py
```diff
@@ -1338,6 +1338,9 @@ def session_buffers_async(self) -> Generator[SessionBufferProtocol, None, None]:
         """
         yield from self._session_buffers
 
+    def get_session_buffer_by_id(self, buffer_id: int) -> SessionBufferProtocol:
```
jwortmann, line 1341: This should be typed `Optional[SessionBufferProtocol]` if it can return `None` for some unknown id.

plugin/core/sessions.py
```diff
@@ -1704,9 +1708,9 @@ def notify_plugin_on_session_buffer_change(self, session_buffer: SessionBufferPr
         if self._plugin:
             self._plugin.on_session_buffer_changed_async(session_buffer)
 
-    def _maybe_resolve_code_action(self, code_action: CodeAction) -> Promise[Union[CodeAction, Error]]:
-        if "edit" not in code_action and self.has_capability("codeActionProvider.resolveProvider"):
-            # TODO: Should we accept a SessionBuffer? What if this capability is registered with a documentSelector?
+    def _maybe_resolve_code_action(self, code_action: CodeAction, session_buffer: SessionBufferProtocol) -> Promise[Union[CodeAction, Error]]:
+        has_capability =  self.has_capability("codeActionProvider.resolveProvider") or session_buffer.has_capability("codeActionProvider.resolveProvider")
```
jwortmann, line 1712: `SessionBuffer.has_capability` falls back to the capabilities of the session, so I think using only the second part here should be sufficient. The next line has two spaces after `and` and you could move this condition directly to the next line instead of defining a variable for it. But I would probably pass the view and defer getting the SessionBuffer to do it only when necessary, like this:
```python
if "edit" not in code_action:
    session_buffer = self.get_session_buffer_by_id(view.buffer_id())
    if session_buffer and session_buffer.has_capability("..."):
        request = ...
```

jwortmann (changes requested):

plugin/core/sessions.py
suiyuex: This line cause code actions unavailable when file doesn't have any change.

image

plugin/core/sessions.py
```diff
@@ -1704,12 +1710,19 @@ def notify_plugin_on_session_buffer_change(self, session_buffer: SessionBufferPr
...
+
+            if not has_capability and view:
+                session_buffer = self.get_session_buffer_by_id(view.buffer_id())
+                if session_buffer:
+                     has_capability = session_buffer.has_capability("codeActionProvider.resolveProvider")
+
+            if has_capability:
+                # We must first resolve the command and edit properties, because they can potentially be absent.
+                request = Request("codeAction/resolve", code_action)
+                return self.send_request_task(request)
```
rchl, line 1725: Use `session_view_for_view_async` to get `SessionView` and call `has_capability_async` on it. No need to do any other API changes.

suiyuex: 👌🏿

plugin/core/sessions.py
```diff
@@ -605,6 +605,9 @@ def get_language_id(self) -> Optional[str]:
     def get_view_in_group(self, group: int) -> sublime.View:
         ...
 
+    def is_self(self, buffer_id: int) -> bool:
+        ...
```
rchl, line 609: Remove unnecessary changes and fix linter issues.

rchl (changes requested):

plugin/core/sessions.py
suiyuex: 👌🏿

plugin/core/sessions.py
```diff
@@ -1704,12 +1704,17 @@ def notify_plugin_on_session_buffer_change(self, session_buffer: SessionBufferPr
         if self._plugin:
             self._plugin.on_session_buffer_changed_async(session_buffer)
 
-    def _maybe_resolve_code_action(self, code_action: CodeAction) -> Promise[Union[CodeAction, Error]]:
-        if "edit" not in code_action and self.has_capability("codeActionProvider.resolveProvider"):
-            # TODO: Should we accept a SessionBuffer? What if this capability is registered with a documentSelector?
-            # We must first resolve the command and edit properties, because they can potentially be absent.
-            request = Request("codeAction/resolve", code_action)
-            return self.send_request_task(request)
+    def _maybe_resolve_code_action(self, code_action: CodeAction, view: Optional[sublime.View]) -> Promise[Union[CodeAction, Error]]:
```
rwols, line 1707:
```suggestion
    def _maybe_resolve_code_action(
        self,
        code_action: CodeAction,
        view: Optional[sublime.View]
    ) -> Promise[Union[CodeAction, Error]]:
```

rchl: Or even this since we tend to use more compact pattern when all arguments fit on a line.

```suggestion
    def _maybe_resolve_code_action(
        self, code_action: CodeAction, view: Optional[sublime.View]
    ) -> Promise[Union[CodeAction, Error]]:
```

rchl (approved):

PR #2349, opened 2023-10-31
Show diagnostics popup when hovering over gutter icons

jwortmann:
This will show a popup with all diagnostics from the line (except hint severity), sorted by severity, when you hover over a gutter icon. It is compatible with the code actions popup if applicable for the current caret position and using the "bulb" setting.

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -179,8 +179,8 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         self._stored_selection = []
         self._sighelp = None  # type: Optional[SigHelp]
         self._lightbulb_line = None  # type: Optional[int]
-        self._diagnostics_by_config = []  # type: List[Tuple[SessionBufferProtocol, List[Diagnostic]]]
-        self._actions_by_config = []  # type: List[CodeActionsByConfigName]
+        self._diagnostics_for_selection = []  # type: List[Tuple[SessionBufferProtocol, List[Diagnostic]]]
+        self._code_actions_for_selection = []  # type: List[CodeActionsByConfigName]
         self._registered = False
 
     def _cleanup(self) -> None:
@@ -480,7 +480,7 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
     def _on_hover_gutter_async(self, point: int) -> None:
         content = ''
         if self._lightbulb_line == self.view.rowcol(point)[0]:
-            content += code_actions_content(self._actions_by_config)
+            content += code_actions_content(self._code_actions_for_selection)
         if userprefs().show_diagnostics_severity_level:
             diagnostics_with_config = []  # type: List[Tuple[ClientConfig, Diagnostic]]
             diagnostics_by_session_buffer = []  # type: List[Tuple[SessionBufferProtocol, List[Diagnostic]]]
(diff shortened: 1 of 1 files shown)


plugin/documents.py
```diff
@@ -471,17 +474,33 @@ def on_hover(self, point: int, hover_zone: int) -> None:
...
-                    flags=sublime.HIDE_ON_MOUSE_MOVE_AWAY,
-                    location=point,
-                    on_navigate=lambda href: self._on_navigate(href, point))
+            sublime.set_timeout_async(partial(self._on_hover_gutter_async, point))
+
+    def _on_hover_gutter_async(self, point: int) -> None:
+        content = ''
+        if self._lightbulb_line == self.view.rowcol(point)[0]:
+            content += code_actions_content(self._actions_by_config)
+        if userprefs().diagnostics_gutter_marker and userprefs().show_diagnostics_severity_level:
```
predrag-codetribe, line 483: Is this the expected behavior?

1 
`"diagnostics_gutter_marker": "circle",`
Screenshot 2023-10-31 at 22 59 16

If there are 2 diagnostics, 
and `diagnostics_gutter_marker` is set to `"circle"`
**display** the diagnostics on gutter hover.

2
`"diagnostics_gutter_marker": "",`
Screenshot 2023-10-31 at 22 58 53

If there are 2 diagnostics, 
and `diagnostics_gutter_marker` is set to `""`
**don't display** the diagnostics on gutter hover.

jwortmann: In case 2 it doesn't show the diagnostics in addition to the code actions, because in general they can be unrelated to each other. But I'm not sure, maybe would be better to also see the diagnostics, what do you think? In that case it should probably only filter for the diagnostics which are overlapping the current caret position or selection.

jwortmann: If I did it correctly, then the popup in your example 2 should show only the first diagnostic with small "d" now (where the caret is).

plugin/documents.py
```diff
@@ -176,6 +179,7 @@ def _setup(self) -> None:
         self._stored_selection = []
         self._sighelp = None  # type: Optional[SigHelp]
         self._lightbulb_line = None  # type: Optional[int]
+        self._diagnostics_by_config = []  # type: List[Tuple[SessionBufferProtocol, List[Diagnostic]]]
```
rchl, line 182: Maybe call it `diagnostics_for_selection` since in this context it looks like diagnostics in the whole file.

But then I suppose `_actions_by_config` could also be renamed.

plugin/documents.py
predrag-codetribe: > If I did it correctly, then the popup in your example 2 should show only the first diagnostic with small "d" now (where the caret is).

That is correct.
I will see the diagnotics where the cursor is if I hover the gutter.

Screenshot 2023-11-04 at 22 37 35

But I am not sure I think that behavior is the best choice, 
because I can already hover with the cursor over the diagnostics and see the diagnostic.
So it is kind of redudant.

>  it should probably only filter for the diagnostics which are overlapping the current caret position or selection.

For me it would be more intuitive 
if hovering the gutter should show all diagnostics that **just exist on the line**. (not based on cursor or selection)

So my expected behavior would be the following:
if `"diagnostics_gutter_marker": ""`
If there are 2 diagnostics,
display the 2 diagnostics on gutter hover.

jwortmann: > But I am not sure I think that behavior is the best choice,
> because I can already hover with the cursor over the diagnostics and see the diagnostic.
> So it is kind of redudant.

In this case with the diagnostic icons disabled it will show only the diagnostic(s) which is/are sent as part of the code action request ([CodeActionContext](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeActionContext)). So there is a good chance that the code actions are related to the diagnostics at the cursor position, especially because we only show Quick Fix actions here. And the code actions request is sent for a particular `range`, so the actions should also be only applicable for the current cursor position.

You could consider the diagnostic from this screenshot to be part of the lightbulb hover, and the other diagnostics on the line or on different lines are not shown because the diagnostics gutter icons are disabled. I think it makes sense.

And I would not say that it is redundant, it's rather just a different way to see the diagnostic message. For example I have the code actions in the regular hover popup disabled, so I will never see them if I hover over the error underline.

rchl (approved):

PR #2350, opened 2023-10-31
change default value of show_references_in_quick_panel to true

rchl:
The panel feels archaic and less user-friendly due to issue with navigating results.
The default should IMO be as close to what ST offers as possible.

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -160,7 +160,7 @@
   // --- Other --------------------------------------------------------------------------
 
   // Show symbol references in Sublime's quick panel instead of the bottom panel.
-  "show_references_in_quick_panel": false,
+  "show_references_in_quick_panel": true,
 
   // Show code lens:
   // "annotation" - show an annotation on the right when code actions are available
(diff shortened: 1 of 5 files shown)


predrag-codetribe:
I am pretty much like panels :)

Even with that being said, 
I would give `show_references_in_quick_panel: true` a try,
as I think it is just a matter of getting use to it.

If someone is not getting use to it, they can just set `show_references_in_quick_panel: false`, so it is not a big deal.

The quick panel has one bonus, you can filter with text in it.
There are still some improvements, like moving the repeating LSP-* label on the right to some other place(ideally displaying just one label at some place)
Screenshot 2023-10-31 at 22 42 58

But that is just something I've noticed, 
not something that needs to be addressed.

predrag-codetribe (approved):

jwortmann:
I think the bottom panel follows the "Find in Files" panel design, so it is also close to what ST offers in some way. But I do prefer the quick panel which allows to navigate quickly with the arrow keys.

The update message skips 3 minor versions though...

rchl:
> I think the bottom panel follows the "Find in Files" panel design, so it is also close to what ST offers in some way. But I do prefer the quick panel which allows to navigate quickly with the arrow keys.

But ST has also "Find references" functionality and that one uses quick panel. That's what I'm aligning with.

jwortmann (approved):

PR #2351, opened 2023-11-01
Add menu item to toggle code lenses

rchl:
Code Lens is one of those features that is IMO too annoying to have always enabled but can be useful to use once in a while to catch un-referenced symbols and such so I think it deserves its own toggle like that.

Threw it in very quickly based on Inlay Hints and Hover Popups toggles. Might have missed something.

### DIFF
--- a/Main.sublime-menu
+++ b/Main.sublime-menu
@@ -157,6 +157,11 @@
         "caption": "-"
       },
       {
+        "caption": "LSP: Show Code Lens",
+        "command": "lsp_toggle_code_lens",
+        "checkbox": true
+      },
+      {
         "caption": "LSP: Show Inlay Hints",
         "command": "lsp_toggle_inlay_hints"
       },
--- a/Main.sublime-menu
+++ b/Main.sublime-menu
@@ -157,13 +157,14 @@
         "caption": "-"
       },
       {
-        "caption": "LSP: Show Code Lens",
-        "command": "lsp_toggle_code_lens",
+        "caption": "LSP: Show Code Lenses",
+        "command": "lsp_toggle_code_lenses",
         "checkbox": true
       },
       {
         "caption": "LSP: Show Inlay Hints",
-        "command": "lsp_toggle_inlay_hints"
+        "command": "lsp_toggle_inlay_hints",
+        "checkbox": true
       },
       {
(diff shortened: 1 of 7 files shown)


Main.sublime-menu
```diff
@@ -156,6 +156,11 @@
         "id": "lsp",
         "caption": "-"
       },
+      {
+        "caption": "LSP: Show Code Lens",
```
jwortmann, line 160: I would use "Lenses" here (plural, like the other items), and probably for the command name too.

Maybe we should add the checkbox for the inlay hints toggle as well?

plugin/code_lens.py
```diff
@@ -1,16 +1,48 @@
...
 from .core.views import make_command_link
 from .core.views import range_to_region
 from html import escape as html_escape
+from functools import partial
 import itertools
 import sublime
 
 
+class LspToggleCodeLensCommand(LspWindowCommand):
+    capability = 'inlayHintProvider'
```
jwortmann, line 18: Wrong capability

predrag-codetribe:
based on the PR title,
I would you explicitly just added in the menu.

Currently for inlay hints there is a command in the command palette, 
do you think that it should be added there, for consistency?

Also by looking at the name  "LSP: Show Code Lens" 
it is not obvous that this command can be toggled? "LSP: Toggle Code Lens"  makes more sense to me

But I get why "LSP: Toggle Code Lens"  was choosen, because inlay hints toggle also uses "LSP: Show Inlay Hint"

rchl:
> Also by looking at the name "LSP: Show Code Lens"
> it is not obvous that this command can be toggled? "LSP: Toggle Code Lens" makes more sense to me

consistency

Screenshot 2023-11-02 at 23 43 50

As far as whether it's obvious or not that it can be toggled... this is about naming conventions for checkboxes. If it's clear that it's a checkbox then it would be even enough to just say "LSP: Code Lenses" since it would be clear from the context that it's a toggle. On Mac it's not that obvious though when checkbox is disabled so maybe that's why "Show..." is better. 

And of course we should try to be consistent with ST but it seems like ST uses checkboxes very sparingly. It prefers to instead change the item title based on the state (showing "Show..." or "Hide...").

PR #2352, opened 2023-11-02
Add "ruby-lsp" section to documentation

themilkman:
Hi there,

there is another notable LSP implementation for Ruby, called `ruby-lsp`. I added it to the docs.
Config inspired by their docs: https://github.com/Shopify/ruby-lsp/blob/main/EDITORS.md#sublime-text-lsp
Please feel free to request changes or do them on your own if I missed something. Thanks!

Cheers!

### DIFF
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -489,7 +489,7 @@ There are multiple options:
 
 ### Sorbet
 
-1. Install the sorbet and sorbet-runtime gem (see [github:sorbet/sorbet](https://github.com/sorbet/sorbet)):
+1. Install the `sorbet` and `sorbet-runtime` gem (see [github:sorbet/sorbet](https://github.com/sorbet/sorbet)):
 
     ```sh
     gem install sorbet
@@ -518,7 +518,7 @@ There are multiple options:
 
 ### Ruby LSP
 
-1. Install the Ruby LSP gem (see [github:Shopify/ruby-lsp](https://github.com/Shopify/ruby-lsp)):
+1. Install the `ruby-lsp` gem (see [github:Shopify/ruby-lsp](https://github.com/Shopify/ruby-lsp)):
 
     ```sh
     gem install ruby-lsp
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -535,7 +535,7 @@ There are multiple options:
                 "selector": "source.ruby | text.html.ruby",
                 "initializationOptions": {
                     "enabledFeatures": {
-                        "diagnostics": false
+                        "diagnostics": true
                     },
                     "experimentalFeaturesEnabled": true
                 }
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -512,6 +512,32 @@ There are multiple options:
     }
     ```
 
+### ruby-lsp
+
+1. Install the ruby-lsp gem (see [github:Shopify/ruby-lsp](https://github.com/Shopify/ruby-lsp)):
+
+        gem install ruby-lsp
+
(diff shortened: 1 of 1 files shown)


rchl (approved):

docs/src/language_servers.md
```diff
@@ -512,6 +516,34 @@ There are multiple options:
...
+    {
+        "clients": {
+            "ruby-lsp": {
+                "enabled": true,
+                "command": ["ruby-lsp"],
+                "selector": "source.ruby | text.html.ruby",
+                "initializationOptions": {
+                    "enabledFeatures": {
+                        "diagnostics": false
+                    },
```
rchl, line 539: Their page says:

> Ruby LSP only supports pull diagnostics

So that's probably why it was disabled but we do support pull diagnostics now so this should work.

themilkman: Took the config from: https://github.com/Shopify/ruby-lsp/blob/main/EDITORS.md#sublime-text-lsp
Generally it works in ST, I tested it.

rchl: I know but it looks outdated since we support pull diagnostics now.

It might work but likely misses diagnostics which are pretty important part of LSP

docs/src/language_servers.md
```diff
@@ -491,12 +491,16 @@ There are multiple options:
 
 1. Install the sorbet and sorbet-runtime gem (see [github:sorbet/sorbet](https://github.com/sorbet/sorbet)):
 
-        gem install sorbet
-        gem install sorbet-runtime
+    ```sh
```
themilkman, line 494: I've oriented myself on the other two LSP-clients, maybe we should add the `sh` to their fences, too?

docs/src/language_servers.md
```diff
@@ -512,11 +516,13 @@ There are multiple options:
     }
     ```
 
-### ruby-lsp
+### Ruby LSP
 
-1. Install the ruby-lsp gem (see [github:Shopify/ruby-lsp](https://github.com/Shopify/ruby-lsp)):
+1. Install the Ruby LSP gem (see [github:Shopify/ruby-lsp](https://github.com/Shopify/ruby-lsp)):
```
themilkman, line 521: Wouldn't be `ruby-lsp` better? It is the actual name of the gem. Also for the heading I'd personally at least leave the "-" in the name.

rchl: Yeah, true, it should be `ruby-lsp` here.

But not in the heading, I would say. Look at the convention used on this page: https://lsp.sublimetext.io/language_servers/ . We tend to refer to servers using natural names and not their package/gem names (although there are exceptions of course).

docs/src/language_servers.md
rchl: Not sure I get what you mean. I thought I've added it to all Ruby clients.

docs/src/language_servers.md
themilkman: Alright, I changed it for sorbet & ruby-lsp.

docs/src/language_servers.md
themilkman: So we should enable "diagnostics", do I get that right? If so, it mabybe should be done for solargraph, too (personally, I disabled it there because it is to distractive for me).

rchl: I don't know the reason why diagnostics were disabled in Solargraph configuration. Was there a technical reason or just a personal preference? Unless we have clear reasoning, I would probably leave it as is.

rchl: I would think that diagnostics should ideally be enabled by default and if someone doesn't like them there is always an option of disabling. Probably better than missing out completely on a potentially useful feature.

themilkman: Changed it for ruby-lsp.

rchl (approved):

rwols (approved):

PR #2353, opened 2023-11-02
Upgrade pyright CI

jwortmann:
After installing Python 3.12, running `tox` in this repo is currently broken on my Windows PC. An upgrade of pyright seems to fix it.

> Ignore new type errors

It's unclear to me how to fix those, but if anyone knows it feel free to write or push directly:

promise

sessions

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1489,7 +1489,7 @@ class Session(TransportCallbacks):
             self._workspace_folders = self._workspace_folders[:1]
         self.state = ClientStates.READY
         if self._plugin_class is not None:
-            self._plugin = self._plugin_class(weakref.ref(self))
+            self._plugin = self._plugin_class(weakref.ref(self))  # type: ignore  # TODO fix me
             # We've missed calling the "on_server_response_async" API as plugin was not created yet.
             # Handle it now and use fake request ID since it shouldn't matter.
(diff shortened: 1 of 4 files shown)


rchl:
> It's unclear to me how to fix those, but if anyone knows it feel free to write or push directly:

Pyright's author helpful as always: https://github.com/microsoft/pyright/discussions/6301

plugin/core/sessions.py
```diff
@@ -1489,7 +1489,7 @@ def _handle_initialize_success(self, result: InitializeResult) -> None:
             self._workspace_folders = self._workspace_folders[:1]
         self.state = ClientStates.READY
         if self._plugin_class is not None:
-            self._plugin = self._plugin_class(weakref.ref(self))  # type: ignore  # TODO fix me
+            self._plugin = self._plugin_class(weakref.ref(cast(Session, self)))
```
rchl, line 1492: This type issue is also a bug:

https://github.com/microsoft/pyright/discussions/6334
https://github.com/microsoft/pyright/issues/6335

jwortmann: So we should rather wait for the release of an unbugged Pyright version, I guess?

rchl: yeah

PR #2361, opened 2023-11-10
try to match at least 2 components of base scope with the map

rchl:
This is a candidate fix for https://github.com/sublimelsp/LSP-typescript/issues/225.

If the base scope doesn't match exactly against our map then take out last components until maybe it does. Do it as long as there are two components or more.

This changed behavior can make more cases work without having to add explicit mappings but it can also "break" things. For example I had to add explicit `text.html.vue` mapping or otherwise it would fall-back to `text.html` mapping and receive a `html` language ID.

Let me know what you think. I think that this approach is potentially more universal than the previous one. And if it causes some scope to match incorrectly then we can always add more specific mapping to fix it.


language-ids.sublime-settings
```diff
@@ -21,103 +21,38 @@
...
     "text.django": "html", // https://github.com/willstott101/django-sublime-syntax
-    "text.html.basic": "html",
-    "text.html.elixir": "html",  // https://github.com/elixir-editors/elixir-tmbundle
-    "text.html.markdown.academicmarkdown": "markdown", // https://github.com/mangecoeur/AcademicMarkdown
-    "text.html.markdown.license": "markdown", // https://github.com/SublimeText/AFileIcon
+    "text.html": "html",
+    "text.html.markdown": "markdown",
     "text.html.markdown.rmarkdown": "r", // https://github.com/REditorSupport/sublime-ide-r
-    "text.html.ngx": "html", // https://github.com/princemaple/ngx-html-syntax
+    "text.html.vue": "vue",
```
jwortmann, line 53: Looking through the language IDs from the specs, two other scopes where it breaks are `text.html.handlebars` -> `handlebars` and `text.xml.xsl` -> `xsl`.

Could you also add `"source.ruby": "ruby"` here, then I could remove the [settings file](https://github.com/jwortmann/brackets-theme/blob/e72123f57a478c2cde9c382b6eb0bbe2ed338cf2/settings/language-ids.sublime-settings) from my theme.

plugin/core/types.py
```diff
@@ -36,7 +36,17 @@
...
+    result = ""
+    # Try to find exact match or less specific match consisting of at least 2 components.
+    scope_parts = base_scope.split('.')
+    while len(scope_parts) >= 2:
+        result = base_scope_map.get('.'.join(scope_parts))
+        if result:
+            break
+        scope_parts.pop()
+    if not result:
+        # If no match use last compoent of the scope as the language ID.
```
istudyatuni, line 48:
```suggestion
        # If no match use last component of the scope as the language ID.
```

predragnikolic:
one test is failing:
```
FAIL: test_basescope2languageid (test_types.TestDocumentSelector)
  File "/home/predragnikolic/.config/sublime-text/Packages/LSP/tests/test_types.py", line 166, in test_basescope2languageid
    self.assertEqual(basescope2languageid("source.js.react"), "javascriptreact")
AssertionError: 'javascript' != 'javascriptreact'
- javascript
+ javascriptreact
?           +++++
```

language-ids.sublime-settings
```diff
@@ -21,103 +21,40 @@
...
-    "source.groovy.gradle": "groovy", // https://github.com/SublimeText/AFileIcon
-    "source.groovy.jenkins": "groovy", // https://github.com/SublimeText/AFileIcon
+    "source.groovy": "groovy",
     "source.js": "javascript",
-    "source.js.eslint": "javascript", // https://github.com/SublimeText/AFileIcon
-    "source.js.gruntfile": "javascript", // https://github.com/SublimeText/AFileIcon
-    "source.js.gulpfile": "javascript", // https://github.com/SublimeText/AFileIcon
-    "source.js.postcss": "javascript", // https://github.com/SublimeText/AFileIcon
-    "source.js.puglint": "javascript", // https://github.com/SublimeText/AFileIcon
-    "source.js.react": "javascriptreact", // https://github.com/Thom1729/Sublime-JS-Custom
```
predragnikolic, line 37: I guess we will have to keep this one

language-ids.sublime-settings
predragnikolic: Off topic, 
I didn't know there was a brackets theme for ST <3

Brackets was my first editor I enjoyed using. 
It had a really unique look, and the theme really looks like Brackets(only the tabs should be hidden ;) )

tests/test_types.py
```diff
@@ -145,3 +146,106 @@ def test_pattern_negated_character_range(self) -> None:
...
+        self.assertEqual(basescope2languageid("source.js.react"), "javascriptreact")
+        self.assertEqual(basescope2languageid("source.js.stylelint"), "javascript")
+        self.assertEqual(basescope2languageid("sourcet.js.unittest"), "javascript")
+        self.assertEqual(basescope2languageid("source.js.webpack"), "javascript")
+        self.assertEqual(basescope2languageid("source.json-tmlanguage"), "jsonc")
+        self.assertEqual(basescope2languageid("source.json.babel"), "json")
+        self.assertEqual(basescope2languageid("source.json.bower"), "json")
+        self.assertEqual(basescope2languageid("source.json.composer"), "json")
+        self.assertEqual(basescope2languageid("source.json.eslint"), "json")
+        self.assertEqual(basescope2languageid("source.json.npm"), "json")
```
rchl, line 175: Having **all** those tests might be missing the point a bit. If we want to test that the function does proper matching then we should only need to test "unique" inputs from the point of the function. And should then likely mock the map so that it's clear from the test point of view what are all the inputs.

Unless we are testing **the mappings** and not the function itself which looks more like what we are doing here. But that doesn't do much for unittesting the code itself and it won't really help to catch the cases we don't account for now.

In any case, we can keep it but we can rewrite those with a lot less duplicated code.

jwortmann:
I'm wondering if we shouldn't simply either...

1. Add all the standard scopes for the language IDs from the table in the specs to this list, for example `"source.python": "python"`. To be safe for any syntax alias from themes or file icon packages which might be introduced in the future.

or

2. Fall back to use the second scope segment as language ID instead of the last segment. Then we would only need to add the exceptions in this list, for example `text.tex.latex` -> `latex`, but none of the "regular" scopes.

There are a few examples where probably no language server exists at the moment, but which might become problematic in the future. For example `source.ini.editorconfig` is used by both the INI and EditorConfig packages, but language ID would probably be `ini`. Or `text.git.commit` -> `git-commit`. And some (outdated or not) packages on Package Control might use some scope deviations which are probably missed at the moment, e.g. [`source.python.3`](https://github.com/jwortmann/brackets-color-scheme/blob/30a6291fb79811be60660cae77e2952a580a2ea2/Brackets%20Dark.sublime-color-scheme#L1582-L1585).

rchl:
> 2\. Fall back to use the second scope segment as language ID instead of the last segment. Then we would only need to add the exceptions in this list, for example `text.tex.latex` -> `latex`, but none of the "regular" scopes.

Think I like that.

tests/test_types.py
predragnikolic: I just wrote the test to test the mapping to  guarantee that we don't miss some things.

But a agree that we can remove it, before we merge.

rwols (approved):

PR #2362, opened 2023-11-12
improve buffer URI format

rchl:
Per discussion in https://github.com/sublimelsp/LSP/pull/2360, change the URI format for unsaved files so that the part after the `scheme` is parsed as a path and not as an `authority` (`netloc` in python).

Shouldn't really affect anything (neither fix the Pyright issue) but seems like a more correct thing to do.

(Also workarounds the issue with `urlparse` failing to parse the URI correctly when path starts with numbers - add `View-` prefix).

### DIFF
--- a/plugin/core/url.py
+++ b/plugin/core/url.py
@@ -26,7 +26,7 @@ def filename_to_uri(file_name: str) -> str:
 def view_to_uri(view: sublime.View) -> str:
     file_name = view.file_name()
     if not file_name:
-        return "buffer://{}".format(view.buffer_id())
+        return "buffer:View-{}".format(view.buffer_id())
     return filename_to_uri(file_name)
 
 
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -70,4 +70,4 @@ class MultiplatformTests(unittest.TestCase):
         view.file_name = unittest.mock.MagicMock(return_value=None)
         view.buffer_id = unittest.mock.MagicMock(return_value=42)
         uri = view_to_uri(view)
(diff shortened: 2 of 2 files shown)


jwortmann:
> Might be better to change the format to `buffer:view-<id>` or something like that rather than adding a workaround for urllib.

This prefix is also just a workaround for the urllib bug, it's just a different one.

Actually it is the buffer ID, not the view ID. I can push my test changes from yesterday so that we could compare; I think I'd slightly prefer the other (explicit) workaround, and it might improve a few other places in the code too. But I'm not sure yet.

plugin/core/url.py
```diff
@@ -26,7 +26,7 @@ def filename_to_uri(file_name: str) -> str:
 def view_to_uri(view: sublime.View) -> str:
     file_name = view.file_name()
     if not file_name:
-        return "buffer://{}".format(view.buffer_id())
+        return "buffer:View-{}".format(view.buffer_id())
```
jfcherng, line 29: Not `Buffer-` prefix? 🤪

jwortmann:
Two reasons why I probably prefer the other approach:

- The underlying bug is that urllib can't parse the `buffer:123` format correctly, and not that the format would be invalid somehow. So I think the fix/workaround should be for the parsing, and not to modify the URI format.
- We should keep in mind what might be a potential format that could natively be used by ST in the future (even if it might never be added): https://github.com/sublimehq/sublime_text/issues/3742
  I assume that `buffer:123` should be more likely than `buffer:Buffer-123`.

PR #2363, opened 2023-11-12
Improve buffer URI format alternative

jwortmann:
Alternative to #2362

### DIFF
--- a/plugin/core/url.py
+++ b/plugin/core/url.py
@@ -60,6 +60,9 @@ def parse_uri(uri: str) -> Tuple[str, str]:
             else:
                 return parsed.scheme, path
         return parsed.scheme, path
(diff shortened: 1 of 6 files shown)


plugin/core/url.py
```diff
@@ -60,6 +60,9 @@ def parse_uri(uri: str) -> Tuple[str, str]:
             else:
                 return parsed.scheme, path
         return parsed.scheme, path
+    elif parsed.scheme == '' and ':' in parsed.path:
```
jwortmann, line 63: I'm not sure though whether this condition is really correct. It seems that colons are allowed in the actual path too.

jwortmann: I think it should be fixed now.

rchl (approved):

PR #2364, opened 2023-11-14
docs: Enable "diagnostics" in solargraph example

themilkman:
Hi there,

as a follow up to https://github.com/sublimelsp/LSP/pull/2352 I tested enabled diagnostics for the solargraph server. It works fine so far so I do not see any reason not to "enable" it per default in the docs.

As always, please feel free to request changes or do them on your own if I missed something. Thanks!

Cheers!

### DIFF
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -480,7 +480,7 @@ There are multiple options:
                 "command": ["solargraph", "stdio"],
                 "selector": "source.ruby | text.html.ruby",
                 "initializationOptions": {
-                    "diagnostics": false
+                    "diagnostics": true
                 }
             }
         }


jwortmann (approved):
For someone who never used Ruby, do you know what are the files with the `text.html.ruby` scope? Are those HTML files which are handled by the server, or Ruby files? I don't even know where this scope comes from, it doesn't seem to be in the syntaxes from the default packages.

I'm asking because there was a recent change for the [language identifier](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentItem) which is derived from the scope name. Previously it would have been `ruby` for this scope, but with the recent changes it will be `html`. So I wonder if this is the right ID, or do we need to add it into the `language-ids.sublime-settings`? Some servers might not care about the ID, but we should probably fix it if it is the wrong one.

rchl:
> Now that you mention it, the scope have been renamed in [sublimehq/Packages@b2553fc](https://github.com/sublimehq/Packages/commit/b2553fcfd99ff95ebfda975358c37ae6e43d6505) from ruby → rails

Then this configuration should be updated to reflect that.

PR #2366, opened 2023-11-18
Various refactorings

jwortmann:
These are a bunch of minor refactorings that I wanted to do. Most of it just moves some code to other places where it makes more sense, but there are no functional changes. See the individual commits for details.

Everything touched by this PR should just be internal functions and constants, so I believe it should be safe to do. At some time the public API for LSP-* packages in `LSP/plugin/__init__.py` should probably be revised and extended, because various packages already import utility functions directly from `LSP/plugin/core/views.py`.

To be released together with:
 - [ ] https://github.com/sublimelsp/LSP-volar/pull/212
 - [ ] https://github.com/sublimelsp/LSP-jdtls/pull/44

### DIFF
--- a/plugin/hover.py
+++ b/plugin/hover.py
@@ -369,10 +369,10 @@ class LspHoverCommand(LspTextCommand):
                 position = {"line": row, "character": col_utf16}  # type: Position
                 r = {"start": position, "end": position}  # type: Range
                 sublime.set_timeout_async(partial(session.open_uri_async, uri, r))
-        elif parse_uri(href)[0].lower() not in ("", "http", "https"):
-            sublime.set_timeout_async(partial(self.try_open_custom_uri_async, href))
-        else:
+        elif parse_uri(href)[0].lower() in ("", "http", "https"):
             open_in_browser(href)
+        else:
+            sublime.set_timeout_async(partial(self.try_open_custom_uri_async, href))
 
     def handle_code_action_select(self, config_name: str, actions: List[CodeActionOrCommand], index: int) -> None:
         if index == -1:
--- a/Main.sublime-menu
+++ b/Main.sublime-menu
@@ -25,50 +25,50 @@
       },
       {
         "command": "lsp_source_action",
-        "args": {"id": -1}
+        "args": {"index": -1}
       },
       {
         "caption": "LSP: Source Action",
         "children": [
           {
             "command": "lsp_source_action",
-            "args": {"id": 0}
+            "args": {"index": 0}
           },
           {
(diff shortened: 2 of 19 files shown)


rchl:
Those broke for me:

https://github.com/sublimelsp/LSP-jdtls/blob/6a0041c7e8ccffe2f77cdc924e3e1a362a3028c8/modules/test_extension_commands.py#L9-L11

https://github.com/sublimelsp/LSP-volar/blob/70ccf1c86258f54932be07babead4c91308e3e53/volar_virtual_file_hierarchy.py#L7-L7

jwortmann:
Oh wow, I didn't know that you used the TreeView somewhere else. That is cool.

The reason I initially put `new_tree_view_sheet` into `core/registry.py` was only to avoid cyclic imports which I didn't know how to fix differently at that time. But I think it would be cleaner code now with the `TYPE_CHECKING` approach that I used in this PR, and then we could just update the import in LSP-volar?

And I could do a PR to update the imports in LSP-jdtls with the next LSP release. I guess that would be easier than to leave behind some aliases to those constants in `views.py` for backwards compatibility.

rchl:
Yeah, new paths are nicer. Only that we should make sure to update those and do a synced release.

plugin/hover.py
```diff
@@ -369,10 +369,10 @@ def on_select(targets: List[str], idx: int) -> None:
                 position = {"line": row, "character": col_utf16}  # type: Position
                 r = {"start": position, "end": position}  # type: Range
                 sublime.set_timeout_async(partial(session.open_uri_async, uri, r))
-        elif urlparse(href).scheme.lower() not in ("", "http", "https"):
-            sublime.set_timeout_async(partial(self.try_open_custom_uri_async, href))
-        else:
+        elif urlparse(href).scheme.lower() in ("", "http", "https"):
```
predragnikolic: Is there a reason to open the browser when `""`

predragnikolic: this message is not that important, as you didn't change this line of code...
I wanted just to recall why it was an empty string..

jwortmann: I think it is for something like `www.example.com/index.html`. Here the scheme is `""`, but it is likely meant to be a website URL. For local files I guess it can be assumed that they are always absolute paths and have a `file` scheme.

rchl:
Created https://github.com/sublimelsp/LSP-volar/pull/212

rchl (approved):

rchl:
Looks like this needs to be rebased to be be able to use "reabse and merge" strategy.

PR #2370, opened 2023-12-04
add NO_UNDO flags to all regions

### LINKED ISSUE #2359: Investigate using `NO_UNDO` flag for regions
Recently I became aware of the `NO_UNDO` flag for `add_regions` calls. Apparently by default, the regions added with `add_regions` insert themselves to the undo stack. This can result in higher memory usage and bigger session files so in general is less performant.

We should investigate the use of the `NO_UNDO` flag if we don't really need it. Not adding it does result in different behavior when undoing region removal but in theory that should be fine with LSP as the regions should get updated/re-added on document changes anyway. The only thing is that there might be more flicker in that case.

This shows the difference between undoing removed region with and without the flag:

https://github.com/sublimelsp/LSP/assets/153197/b3d9ee26-6b9f-4b82-86db-92b5bcffc4f1

rchl:
Added `NO_UNDO` flag to all (?) `add_regions` calls. I have not really battle tested it yet so intending to spend some time doing that and checking whether there are any downsides.

Resolves #2359


plugin/core/views.py
```diff
@@ -53,9 +53,10 @@
 
 MarkdownLangMap = Dict[str, Tuple[Tuple[str, ...], Tuple[str, ...]]]
 
-DOCUMENT_LINK_FLAGS = sublime.HIDE_ON_MINIMAP | sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE | sublime.DRAW_SOLID_UNDERLINE  # noqa: E501
+DOCUMENT_LINK_FLAGS = sublime.HIDE_ON_MINIMAP | sublime.DRAW_NO_FILL | sublime.DRAW_NO_OUTLINE | sublime.DRAW_SOLID_UNDERLINE | sublime.NO_UNDO  # noqa: E501
+SEMANTIC_TOKEN_FLAGS = sublime.DRAW_NO_OUTLINE | sublime.NO_UNDO
```
jwortmann, line 57: Perhaps these two would fit into `constants.py`?

plugin/core/types.py
```diff
@@ -320,31 +320,33 @@ def r(name: str, default: Union[bool, int, str, list, dict]) -> None:
...
+        if style_str == "box":
+            return default
+        if style_str == "underline":
+            return default | sublime.DRAW_NO_OUTLINE | sublime.DRAW_SOLID_UNDERLINE
+        if style_str == "stippled":
+            return default | sublime.DRAW_NO_OUTLINE | sublime.DRAW_STIPPLED_UNDERLINE
+        if style_str == "squiggly":
+            return default | sublime.DRAW_NO_OUTLINE | sublime.DRAW_SQUIGGLY_UNDERLINE
+        # default style (includes NO_UNDO)
+        return None
```
rchl, line 349: Don't have time to investigate now but I've messed something here (perhaps).

Getting filled regions now

Screenshot 2023-12-06 at 11 47 55

predragnikolic:
I will start testing this

jwortmann:
I wonder is there any reason why NO_UNDO was not added when drawing the regions in
https://github.com/sublimelsp/LSP/blob/34c49dd9ad3c2f786db527393dcdd8638b1ecd74/plugin/session_view.py#L127

I doubt that there is a measurable performance difference from doing that, because those regions are drawn only a single time for each view, but it probably also doesn't hurt to add the flag. So I would add it just for consistency. And then I would probably also add the HIDDEN flag, which in this case doesn't have any effect because the regions are empty anyway, but it feels more correct to have it and it might make the code a bit more comprehensible.

For example
```python
REGIONS_INITIALIZE_FLAGS = sublime.HIDDEN | sublime.NO_UNDO 
```
(haven't tested though)

rchl:
Seeing decent memory savings on this branch.

This is tested on the typescript-language-server repo with 10 different files opened showing semantic tokens and code lens annotations.

The physical session file doesn't seem to be affected much. Which I guess makes sense because those regions don't survive restart so those are only stored in memory.

## Before

Screenshot 2023-12-10 at 21 11 24

## After

Screenshot 2023-12-10 at 21 10 36

predragnikolic:
I haven't noticed any issues with this.

predragnikolic (approved):

PR #2374, opened 2023-12-12
Fix some mypy type issues in stubs

rchl:
- Check stubs using mypy. Noticed some type issues when checking `lsp_utils`.
- Ignore `reportIncompatibleMethodOverride` pyright warning which is reported in the new pyright version because it has introduced a new, stricter `typeCheckingMode` "standard" that is enabled by default.

### DIFF
--- a/stubs/sublime.pyi
+++ b/stubs/sublime.pyi
@@ -704,7 +704,7 @@ class Window:
         """
         ...
 
-    def hwnd(self):
+    def hwnd(self) -> int:
         """
         A platform specific window handle. Windows only.
         """
@@ -944,14 +944,14 @@ class Window:
         """
         ...
 
-    def layout(self):
+    def layout(self) -> Dict[str, Any]:
         """
         Get the group layout of the window.
         """
         ...
 
     @deprecated("Use layout() instead")
-    def get_layout(self):
+    def get_layout(self) -> Dict[str, Any]:
         ...
 
     def set_layout(self, layout: Dict[str, Any]) -> None:
@@ -1145,7 +1145,7 @@ class Window:
         """
         ...
 
-    def template_settings(self):
+    def template_settings(self) -> Settings:
         """
         Per-window settings that are persisted in the session, and duplicated into new windows.
         """
@@ -1349,7 +1349,7 @@ class TextChange:
     str: str
     """A string of the new contents of the region specified by `a` and `b`."""
 
-    def __init__(self, pa: HistoricPosition, pb: HistoricPosition, len_utf16: int, len_utf8: int, s: str) -> None:
+    def __init__(self, pa: HistoricPosition, pb: HistoricPosition, len_utf16: int, len_utf8: int, s: str) -> None:  # type: ignore[valid-type] # noqa
         ...
 
 
@@ -1388,7 +1388,7 @@ class Selection(Reversible):
         """
         ...
 
-    def __eq__(self, rhs: Selection | None) -> bool:
+    def __eq__(self, rhs: object) -> bool:
         """
         Whether the selections are identical.
         """
@@ -1929,7 +1929,7 @@ class View:
         """
         ...
 
-    def style(self):
+    def style(self) -> Dict[str, str]:
         """
         The global style settings for the view. All colors are normalized to the six character hex form with a leading
         hash, e.g. `#ff0000`.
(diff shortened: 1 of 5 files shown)


stubs/sublime.pyi
```diff
@@ -1388,7 +1388,7 @@ class Selection(Reversible):
         """
         ...
 
-    def __eq__(self, rhs: Selection | None) -> bool:
+    def __eq__(self, rhs: object) -> bool:
```
rchl, line 1391:
```
stubs/sublime.pyi:1391: error: Argument 1 of "__eq__" is incompatible with supertype "object"; supertype defines the argument type as "object"  [override]
stubs/sublime.pyi:1391: note: This violates the Liskov substitution principle
stubs/sublime.pyi:1391: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
stubs/sublime.pyi:1391: note: It is recommended for "__eq__" to work with arbitrary objects, for example:
stubs/sublime.pyi:1391: note:     def __eq__(self, other: object) -> bool:
stubs/sublime.pyi:1391: note:         if not isinstance(other, Selection):
stubs/sublime.pyi:1391: note:             return NotImplemented
stubs/sublime.pyi:1391: note:         return <logic to compare two Selection instances>
```

Makes sense to me as comparison is not guaranteed to be done against the same type.

jwortmann: Actually I put this on purpose to `Selection | None`.

The reason for that is because it might crash (i.e. raise a runtime exception) if you compare to an arbitrary object. The implementation in sulime.py is this:
```python
    def __eq__(self, rhs):
        return rhs is not None and list(self) == list(rhs)
``` 
So the argument is at least restricted to objects that are iterable. As you pasted above, Pyright indeed suggests a proper implementation which would make it work with any objects. But we don't have any influence on the implementation, so I would say here it's better to get a type checker warning with the current type stubs, instead of a potential runtime error in case of improper usage. What do you think?

Also the stub files are excluded from type checking in CI, so I guess errors/warnings for those files are not very important and we could ignore them.

rchl: This is a mypy error. For `__lt__` it doesn't raise an issue.

But I don't fully understand what you mean by saying that the following implementation could crash:

```py
    def __eq__(self, rhs):
        return rhs is not None and list(self) == list(rhs)
```

if `rhs` is not a `Selection`. (Could be my ignorance regarding how `__eq__` works)

jwortmann:
```
Python 3.3>>> view.sel() == 1
  File "__main__", line 1, in <module>
  File "C:\Program Files\Sublime Text\Lib\python33\sublime.py", line 1094, in __eq__
    return rhs is not None and list(self) == list(rhs)
TypeError: 'int' object is not iterable
```
Similar for other classes which are not iterable.

If you run it in the 3.8 API it doesn't raise a runtime error because of the `isinstance` check:
```
>>> view.sel() == 1
False
```

stubs/sublime.pyi
```diff
@@ -2614,7 +2614,7 @@ class Phantom:
     ) -> None:
         ...
 
-    def __eq__(self, rhs: 'Phantom') -> bool:
+    def __eq__(self, rhs: 'Phantom') -> bool:  # type: ignore[override]
```
jwortmann, line 2617:
```suggestion
    def __eq__(self, rhs: Phantom) -> bool:  # type: ignore[override]
```

I missed those quoted annotations earlier. It's not necessary to use quotes for the types in stub files, even for forward references.

stubs/sublime.pyi
```diff
@@ -2726,7 +2726,7 @@ class CompletionItem:
         ) -> None:
         ...
 
-    def __eq__(self, rhs: 'CompletionItem') -> bool:
+    def __eq__(self, rhs: 'CompletionItem') -> bool:  # type: ignore[override]
```
jwortmann, line 2729:
```suggestion
    def __eq__(self, rhs: CompletionItem) -> bool:  # type: ignore[override]
```

tox.ini
```diff
@@ -20,11 +20,11 @@ per-file-ignores =
 
 [testenv]
 deps =
-    ; mypy==0.971
+    mypy==1.7.1
     flake8==5.0.4
-    pyright==1.1.335
+    pyright==1.1.339
```
jwortmann, line 25: Ideally GitHub CI should be updated to use the same version(s).

stubs/sublime.pyi
```diff
@@ -2471,7 +2471,7 @@ class View:
 
     def export_to_html(
         self,
-        regions=None,
+        regions: Optional[Region | List[Region]] = None,
```
jwortmann, line 2474:
```suggestion
        regions: Region | List[Region]] | None = ...,
```

Just for consistency. I used `Optional` only at two places inside of tuples, because `Foo | None` for a single tuple element is a bit difficult to read.

PR #2376, opened 2023-12-16
Add only_files argument for lsp_save_all command

### LINKED ISSUE #2371: Provide a way to save all modified files after applying workspace edits

When applying refactorings (like renames) it's often the case that multiple files are modified. It's a chore to then have to find and save all those modified files.

We could provide a dialog after more than one file was modified asking the user whether all modified files should be saved. Note that some people don't like dialogs (#1922).

Or maybe even show a "tree view" in a sheet, showing all modified files and allowing the user to navigate to them and save all at once.

- Saving all edited files one by one.
- Using "save all" command but that fails when there is some unsaved buffer in the window (for example I like to have some to keep notes in them)

jwortmann:
Closes #2371 (?)

Using this argument, you can add a key binding (or menu entry or command palette entry) to save all files, without being bothered by save dialogs for new buffers.

Also did a small adjustment to the caption of the example key binding, to make it clear that `lsp_save_all` only saves views for which a language server is running.

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -17,10 +17,11 @@
             {"key": "auto_complete_visible"}
         ]
     },
-    // Save all open files with lsp_save
+    // Save all open files that have a language server attached with lsp_save
     // {
     //     "keys": ["UNBOUND"],
-    //     "command": "lsp_save_all"
+    //     "command": "lsp_save_all",
+    //     "args": {"only_files": false}
     // },
     // Run Code Action
     // {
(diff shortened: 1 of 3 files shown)


rchl (approved):
Looks good but I'm not sure about resolving https://github.com/sublimelsp/LSP/issues/2371 (as you are also, I guess).

1. It requires separate, manual shortcut press
2. It saves all unsaved files rather than just the ones that were affected by rename.

That said, it might do well 99% of the time.

jwortmann:
Right, I made this PR because you mentioned this as an alternative solution to the linked issue, and I thought it would be an easy implementation and a nice little improvement for the "Save all" behavior in general.

I don't really think that adding a new setting to automatically save after workspace edits would be justified. But just to throw in an idea for the case of automatically saving; then it should probably first check all the open files which already had unsaved modifications before, and exclude those files from saving.

---

A bit unrelated, but while looking through the implementations, I saw that the `lsp_save` command should maybe accept the arguments of the built in `save` command (`async`, `encoding`, `quiet`) and just pass them through when calling built-in `save`. Then `async` should be specified in the key binding arguments, instead of being hardcoded in the code. Similar to the built-in key binding works (Windows example):
```json
{ "keys": ["ctrl+s"], "command": "save", "args": { "async": true } },
```

PR #2377, opened 2023-12-18
docs: Move Steep in Ruby Section & minor cleanups

themilkman:
As always, please feel free to request changes or do them on your own if I missed something.

Thanks & cheers!

### DIFF
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -544,26 +544,20 @@ There are multiple options:
 
 ### Steep
 
-1. Add the `steep` gem (see [github:soutaro/steep](https://github.com/soutaro/steep)) into your Gemfile and install it
+1. Install the `steep` gem (see [github:soutaro/steep](https://github.com/soutaro/steep)):
 
     ```sh
-    bundle install
+    gem install steep
     ```
 
-2. Binstub steep executable
-
-    ```sh
-    steep binstub
-    ```
-
-3. Open `Preferences > Package Settings > LSP > Settings` and add the `"steep"` client configuration to the `"clients"`:
+2. Open `Preferences > Package Settings > LSP > Settings` and add the `"steep"` client configuration to the `"clients"`:
 
     ```jsonc
     {
         "clients": {
             "steep": {
                 "enabled": true,
-                "command": ["bin/steep", "langserver"],
+                "command": ["steep", "langserver"],
                 "selector": "source.ruby | text.html.ruby",
             }
         }
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -562,6 +562,7 @@ There are multiple options:
     {
         "clients": {
             "steep": {
+                "enabled": true,
                 "command": ["bin/steep", "langserver"],
                 "selector": "source.ruby | text.html.ruby",
             }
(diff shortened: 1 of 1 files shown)


docs/src/language_servers.md
```diff
@@ -542,6 +542,33 @@ There are multiple options:
...
+    {
+        "clients": {
+            "steep": {
+                "command": ["bin/steep", "langserver"],
+                "selector": "source.ruby | text.html.ruby",
+            }
+        }
+    }
+    ```
+
```
predragnikolic, line 571: Did you forgot to include the fourth step?
```suggestion

4. Activate server for the currect project - open Command Palette `LSP: Enable Language Server in Project > steep`
```

predragnikolic: Ok, I saw this commit now [75a2468](https://github.com/sublimelsp/LSP/pull/2377/commits/75a246881a64051aac3332f2cac62613efff8b70)

predragnikolic: It is true that the instruction for others do not have this step,
but they do have:
```
"enabled": true,
```

docs/src/language_servers.md
```diff
@@ -542,6 +542,33 @@ There are multiple options:
...
+    steep binstub
+    ```
+
+3. Open `Preferences > Package Settings > LSP > Settings` and add the `"steep"` client configuration to the `"clients"`:
+
+    ```jsonc
+    {
+        "clients": {
+            "steep": {
+                "command": ["bin/steep", "langserver"],
```
predragnikolic, line 565:
```suggestion
                "enabled": true,
                "command": ["bin/steep", "langserver"],
```

Specify `"enabled": true` if we dont want to have a paragraph like this:
```
4. Activate server for the currect project - open Command Palette `LSP: Enable Language Server in Project > steep`
```

docs/src/language_servers.md
```diff
@@ -542,6 +542,34 @@ There are multiple options:
...
+    ```
+
+3. Open `Preferences > Package Settings > LSP > Settings` and add the `"steep"` client configuration to the `"clients"`:
+
+    ```jsonc
+    {
+        "clients": {
+            "steep": {
+                "enabled": true,
+                "command": ["bin/steep", "langserver"],
```
rchl, line 566: Is this supposed to be a relative path? Because it is here but since this is a global server configuration, it's not likely to work.

docs/src/language_servers.md
```diff
@@ -542,6 +542,34 @@ There are multiple options:
     }
     ```
 
+### Steep
+
+1. Add the `steep` gem (see [github:soutaro/steep](https://github.com/soutaro/steep)) into your Gemfile and install it
+
+    ```sh
+    bundle install
```
rchl, line 550: I don't know the ruby ecosystem but why are those instructions say to add to "gemfile" while instructions for other servers say to use `gem install`? Is it because this approach is local to the project while `gem install` is for installing globally? If so then shouldn't we recommend to install globally in this case also?

rchl: I would say that project-specific setup should be combined with a project-specific configuration. Unless there is something specific about this server that makes it more suitable for project-specific use, I'd suggest following other setups for global server.

rchl (approved):

Issue #2118, opened 2022-11-18
"list index out of range" when no command is defined


Sainan:
**Describe the bug**
"command" is documented as "can be empty if starting manually, then TCP transport must be configured". However, doing so raises "list index out of range". This is possibly a regression of #1564.

**To Reproduce**
Use the following user configuration:
```JSON
{
    "clients": {
        "pluto-language-server": {
            "enabled": true,
            "selector": "source.lua",
            "tcp_port": 5007,
            "command": []
        }
    }
}
```

**Expected behavior**
It shouldn't've attempted to start the process with the `[]` configuration.

**Screenshots**
![image](https://user-images.githubusercontent.com/63328889/202662990-d2183a49-0212-4544-b050-ffd7cfd6994a.png)

**Environment:**
- OS: Windows 10
- Sublime Text version: 4143
- LSP version: 1.20.0
- Language servers used: [pluto-language-server](https://github.com/PlutoLang/pluto-language-server)

predragnikolic:
The "list index out of range" error should be fixed with https://github.com/sublimelsp/LSP/pull/2300

Feel free to reopen, if you stumble upon any issue.

PR #2378, opened 2023-12-18
fix empty command behavior on windows

AlexeyBond:
Plugin fails with "list index out of range" error when running on windows and a server is configured with empy `command` and non-empty `tcp_port`.

Looks like regression of #2118, but it also doesn't look like it was really fixed.

### DIFF
--- a/plugin/core/transports.py
+++ b/plugin/core/transports.py
@@ -254,8 +254,13 @@ def create_transport(config: TransportConfig, cwd: Optional[str],
 
     if config.listener_socket:
         assert isinstance(config.tcp_port, int) and config.tcp_port > 0
-        process, sock, reader, writer = _await_tcp_connection(
-            config.name, config.tcp_port, config.listener_socket, start_subprocess)
+        if config.command:
+            process, sock, reader, writer = _start_client_process_and_await_connection(
+                config.listener_socket,
+                start_subprocess
+            )
+        else:
+            sock, reader, writer = _await_client_connection(config.listener_socket)
     else:
         if config.command:
             process = start_subprocess()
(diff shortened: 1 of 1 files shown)


predragnikolic:
Hello @AlexeyBond , 

Can you paste the `clients` configuration that you have in LSP.sublime-settings that can reproduce the issue?

AlexeyBond:
> Can you paste the `clients` configuration that you have in LSP.sublime-settings that can reproduce the issue?

`LSP.sublime-settings` contains the following configuration:

```JavaScript
// Settings in here override those in "LSP/LSP.sublime-settings"
{
	"clients": {
		"test-lsp": {
			"selector": "source.lisp",
			"tcp_port": 9257,
			"enabled": true
		},
	},
}
```

I'm running an example server from tower-lsp on port 9257.

plugin/core/transports.py
```diff
@@ -245,11 +245,11 @@ def create_transport(config: TransportConfig, cwd: Optional[str],
     else:
         stdout = subprocess.PIPE
         stdin = subprocess.PIPE
-    startupinfo = _fixup_startup_args(config.command)
     sock = None  # type: Optional[socket.socket]
     process = None  # type: Optional[subprocess.Popen]
 
     def start_subprocess() -> subprocess.Popen:
+        startupinfo = _fixup_startup_args(config.command)
         return _start_subprocess(config.command, stdin, stdout, subprocess.PIPE, startupinfo, config.env, cwd)
```
predragnikolic, line 253: the bug is caused by the `_fixup_startup_args` function 
```py
def _fixup_startup_args(args: List[str]) -> Any:
    startupinfo = None
    if sublime.platform() == "windows":
        # ... this line bellow causes the Error because, args is None
        executable_arg = args[0]
```

`start_subprocess` will be called in 2 cases:
```diff
-   startupinfo = _fixup_startup_args(config.command)


    def start_subprocess() -> subprocess.Popen:
+       startupinfo = _fixup_startup_args(config.command)
        return _start_subprocess(config.command, stdin, stdout, subprocess.PIPE, startupinfo, config.env, cwd)

    if config.listener_socket:
        assert isinstance(config.tcp_port, int) and config.tcp_port > 0
        process, sock, reader, writer = _await_tcp_connection(
            config.name, config.tcp_port, config.listener_socket, start_subprocess) # <- Case 1
    else:
        if config.command:
            process = start_subprocess() # <- Case 2 
```

In this PR the `_fixup_startup_args` will not be called in Case 2 anymore,
so the bug in that case will be fixed.

But in Case 1, when `start_subprocess` is called, the `_fixup_startup_args` will also be called, and the command will still be None. So we still have a bug there.

IMO, a better option would be to fix the `_fixup_startup_args` function:
```diff
- def _fixup_startup_args(args: List[str]) -> Any:
+ def _fixup_startup_args(args: Optional[List[str]]) -> Any:
+     if args is None:
+         return None
    startupinfo = None
    if sublime.platform() == "windows":
        startupinfo = subprocess.STARTUPINFO()  # type: ignore
        startupinfo.dwFlags |= subprocess.SW_HIDE | subprocess.STARTF_USESHOWWINDOW  # type: ignore
        executable_arg = args[0]
        _, ext = os.path.splitext(executable_arg)
        if len(ext) < 1:
            path_to_executable = shutil.which(executable_arg)
            # what extensions should we append so CreateProcess can find it?
            # node has .cmd
            # dart has .bat
            # python has .exe wrappers - not needed
            for extension in ['.cmd', '.bat']:
                if path_to_executable and path_to_executable.lower().endswith(extension):
                    args[0] = executable_arg + extension
                    break
    return startupinfo
```

AlexeyBond: Looks like something else is wrong with case 1: it doesn't make sense to start a subprocess with empty command. And `_start_subprocess` will fail when first argument is empty:

```Python
def _start_subprocess(
    args: List[str],
# ...
) -> subprocess.Popen:
    debug("starting {} in {}".format(args, cwd if cwd else os.getcwd()))
    process = subprocess.Popen( # <- Will throw if args is empty
        args=args,
        # ...
    )
    _subprocesses.add(process)
    return process
```

So, even with check added in `_fixup_startup_args`, it will still fail, but in `_start_subprocess`.

rchl (approved):

PR #2383, opened 2023-12-22
Added solidity language server setup guide

kozlovzxc:
### Description

Just added a quick guide how to make LSP package to work with solidity, I've also tried to make [solc](https://docs.soliditylang.org/en/v0.8.23/installing-solidity.html#macos-packages) lsp to work, but [hardhat-vscode](https://github.com/NomicFoundation/hardhat-vscode/tree/development/server) feels much more suitable and easier to set up.

### Demo

image

image

### DIFF
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -629,14 +629,15 @@ Follow installation instructions on [LSP-metals](https://github.com/scalameta/me
 
 ## Solidity
 
-1. Install [github:NomicFoundation/hardhat-vscode](https://github.com/NomicFoundation/hardhat-vscode/tree/development/server) (follow instructions in the repo):
-
-2. Open `Preferences > Package Settings > LSP > Settings` and add the `"solidity"` client configuration to the `"clients"`:
+1. Install the [Ethereum](https://packagecontrol.io/packages/Ethereum) package from Package Control for syntax highlighting.
+2. Install [github:NomicFoundation/hardhat-vscode](https://github.com/NomicFoundation/hardhat-vscode/tree/development/server).
+3. Open `Preferences > Package Settings > LSP > Settings` and add the `"solidity"` client configuration to the `"clients"`:
 
     ```jsonc
     {
         "clients": {
             "solidity": {
+                "enabled": true,
                 "command": ["nomicfoundation-solidity-language-server", "--stdio"],
                 "selector": "source.solidity"
             }
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -627,7 +627,7 @@ Follow installation instructions on [LSP-metals](https://github.com/scalameta/me
     }
     ```
 
-### Solidity
+## Solidity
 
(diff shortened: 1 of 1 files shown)


docs/src/language_servers.md
```diff
@@ -627,6 +627,23 @@ Follow installation instructions on [LSP-metals](https://github.com/scalameta/me
...
+
+1. Install [github:NomicFoundation/hardhat-vscode](https://github.com/NomicFoundation/hardhat-vscode/tree/development/server) (follow instructions in the repo):
+
+2. Open `Preferences > Package Settings > LSP > Settings` and add the `"solidity"` client configuration to the `"clients"`:
+
+    ```jsonc
+    {
+        "clients": {
+            "solidity": {
+                "command": ["/PATH/TO/node", "PATH/TO/nomicfoundation-solidity-language-server", "--stdio"],
```
rchl, line 640: Haven't checked but should be enough with something like:

```suggestion
                "command": ["nomicfoundation-solidity-language-server", "--stdio"],
```

docs/src/language_servers.md
```diff
@@ -627,6 +627,23 @@ Follow installation instructions on [LSP-metals](https://github.com/scalameta/me
     }
     ```
 
+### Solidity
+
+1. Install [github:NomicFoundation/hardhat-vscode](https://github.com/NomicFoundation/hardhat-vscode/tree/development/server) (follow instructions in the repo):
+
+2. Open `Preferences > Package Settings > LSP > Settings` and add the `"solidity"` client configuration to the `"clients"`:
```
rchl, line 634: I'm guessing the user also has install some syntax for "solidity"?

docs/src/language_servers.md
```diff
@@ -627,6 +627,23 @@ Follow installation instructions on [LSP-metals](https://github.com/scalameta/me
     }
     ```
 
+### Solidity
```
rchl, line 630: This is under the "Shell" programming language section. Use 2 hashes if it's a separate programming language.

docs/src/language_servers.md
kozlovzxc: I've started with just `nomicfoundation-solidity-language-server`, but it didn't work for me, it wasn't able to locate `node` for some reason 🤔 Will take another look if there were any misconfig on my side

docs/src/language_servers.md
kozlovzxc: Oh, my bad, will fix 👍

docs/src/language_servers.md
jwortmann: The `source.solidity` scope used for `"selector"` is not built-in from Sublime Text, so you would need to have the according syntax package from Package Control installed to make the language server start. The other config examples on this page link to the syntax packages.

docs/src/language_servers.md
rchl: If node is not in the path then I would consider it a separate issue that is explained at https://lsp.sublimetext.io/troubleshooting/#updating-the-path-used-by-lsp-servers

predragnikolic (approved):
Thanks! 
and sorry for the wait.

PR #2384, opened 2023-12-28
refactor task running on save

rchl:
Refactor the code that runs tasks on saving to handle cancellation properly when running `lsp_save` command twice in a row, in a sync fashion.

Previously the queued `_run_next_task_async` callback from the first invocation still triggered after the first invocation was canceled, triggering an exception.

Now the tasks running logic is encapsulated into its own class, making cancelling easier and more robust.


plugin/save_command.py
```diff
@@ -60,6 +61,47 @@ def _purge_changes_async(self) -> None:
...
+        self._text_command = text_command
+        self._tasks = tasks
+        self._on_tasks_completed = on_complete
+        self._pending_tasks = []  # type: List[SaveTask]
+        self._canceled = False
+
+    def run(self) -> None:
+        for Task in self._tasks:
+            if Task.is_applicable(self._text_command.view):
+                self._pending_tasks.append(Task(self._text_command, self._on_task_completed_async))
```
jwortmann, line 77: It wasn't changed in this PR, but the uppercase `Task` is a bit inconsistent with the lines 113-115 in this file, where it is lowercase. I don't know what the convention is (or if a convention exists for it), but I would say it should either be lowercase here or uppercase in lines 113-115.

PR #2386, opened 2024-01-03
Update instructions for OCaml and ReasonML

smorimoto:
I plan to document instructions for Sublime on the OCaml-LSP side, but this change basically allows us to use OCaml-LSP as is.

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -344,10 +344,9 @@
     },
     "ocaml": {
       "command": [
-        "ocaml-language-server",
-        "--stdio"
+        "ocamllsp"
       ],
-      "selector": "source.reason | source.ocaml",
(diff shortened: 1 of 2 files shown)


LSP.sublime-settings
```diff
@@ -344,10 +344,9 @@
     },
     "ocaml": {
       "command": [
-        "ocaml-language-server",
-        "--stdio"
```
rwols, line 348: Reminder: this change would require some notice in the release notes.

rchl:
> I plan to document instructions for Sublime on the OCaml-LSP side, but this change basically allows us to use OCaml-LSP as is.

We should first see what ends up in their documentation so that we know what should we add here. Maybe then will not want too specific instructions and just general ones on how to start the server in which case we should document the rest here.

PR #2389, opened 2024-01-04
feat: support setting cursor position in text edits

rchl:
Add opt-in handling for placeholders in text edits which define the position of the cursor after text edits are applied. This is required (well, not required but it improves experience) for custom rust-analyzer `moveItem` command that I'm adding support for in https://github.com/sublimelsp/LSP-rust-analyzer/pull/111#pullrequestreview-1805125212

Based on rust-analyzer extension code: https://github.com/rust-lang/rust-analyzer/blob/f7823f31069c6ec5be24d0497847bf1bb8a4c683/editors/code/src/snippets.ts#L41-L87

Also check rust-analyzer description of their custom snippet text edit: https://github.com/rust-lang/rust-analyzer/blob/master/docs/dev/lsp-extensions.md#snippet-textedit

### DIFF
--- a/plugin/edit.py
+++ b/plugin/edit.py
@@ -1,8 +1,9 @@
 from .core.edit import TextEditTuple
 from .core.logging import debug
-from .core.typing import List, Optional, Any, Generator, Iterable
+from .core.typing import List, Optional, Any, Generator, Iterable, Tuple
 from contextlib import contextmanager
 import operator
+import re
 import sublime
 import sublime_plugin
 
@@ -21,8 +22,11 @@ def temporary_setting(settings: sublime.Settings, key: str, val: Any) -> Generat
 
 
 class LspApplyDocumentEditCommand(sublime_plugin.TextCommand):
+    re_snippet = re.compile(r'\$(0|\{0:([^}]*)\})')
 
-    def run(self, edit: sublime.Edit, changes: Optional[List[TextEditTuple]] = None) -> None:
+    def run(
+        self, edit: sublime.Edit, changes: Optional[List[TextEditTuple]] = None, process_snippets: bool = False
+    ) -> None:
         # Apply the changes in reverse, so that we don't invalidate the range
         # of any change that we haven't applied yet.
         if not changes:
@@ -30,10 +34,27 @@ class LspApplyDocumentEditCommand(sublime_plugin.TextCommand):
         with temporary_setting(self.view.settings(), "translate_tabs_to_spaces", False):
             view_version = self.view.change_count()
             last_row, _ = self.view.rowcol_utf16(self.view.size())
+            snippet_region_count = 0
             for start, end, replacement, version in reversed(_sort_by_application_order(changes)):
                 if version is not None and version != view_version:
                     debug('ignoring edit due to non-matching document version')
                     continue
+                snippet_region = None  # type: Optional[Tuple[Tuple[int, int], Tuple[int, int]]]
+                if process_snippets and replacement:
+                    parsed = self.parse_snippet(replacement)
+                    if parsed:
+                        replacement, (placeholder_start, placeholder_length) = parsed
+                        # There might be newlines before the placeholder. Find the actual line and character offset
+                        # of the placeholder.
+                        prefix = replacement[0:placeholder_start]
+                        last_newline_start = prefix.rfind('\n')
+                        start_line = start[0] + prefix.count('\n')
+                        if last_newline_start == -1:
+                            start_column = start[1] + placeholder_start
+                        else:
+                            start_column = len(prefix) - last_newline_start - 1
+                        end_column = start_column + placeholder_length
+                        snippet_region = ((start_line, start_column), (start_line, end_column))
                 region = sublime.Region(
                     self.view.text_point_utf16(*start, clamp_column=True),
                     self.view.text_point_utf16(*end, clamp_column=True)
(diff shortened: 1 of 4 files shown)


plugin/edit.py
```diff
@@ -55,6 +86,15 @@ def apply_change(self, region: sublime.Region, replacement: str, edit: sublime.E
             else:
                 self.view.erase(edit, region)
 
+    def parse_snippet(self, replacement: str) -> Optional[Tuple[str, Tuple[int, int]]]:
```
jwortmann, line 89: I haven't taken a closer look to understand every part of this PR, but I wonder is there a particular reason for all this manual snippet handling? Also can it handle snippets with multiple tabstops correctly? If I understand the code above correcty, it manually just adds a selection for each tabstop. But this is not how snippets work, if there are multiple tabstops it should set a single cursor and then with <kbd>tab</kbd> you can jump to the next one.

I think instead of `view.insert(...)` the logic to apply text edits should better user the built-in commands instead (at least for snippets, to handle them correctly). Like in
https://github.com/sublimelsp/LSP/blob/36871c2a027d60e4fe77aa11ba4d26a8905bc511/plugin/completion.py#L354-L357

rchl: It's obviously tailored for rust-analyzer and taken from it so this implementation assumes that there is only a single placeholder (specifically `$0` or `${0:...}`). So no multiple tab stops. But yes, those are not so much snippets but more a functionality to set cursor(s). The rust-analyzer-initiated [LSP protocol feature request](https://github.com/microsoft/language-server-protocol/issues/724) even calls it "Allow CodeActions to specify cursor position".

I remember pretty well that we've tried `insert_snippet` (or @rwols did) when implementing the original code and it didn't work at all since it contains a lot of extra magic that for example auto-figures indentation. We need to do a raw insert or replace that won't do any of that.

And we need to set the cursor after applying the edit. Otherwise the selection will shift randomly.

jwortmann: I see. I guess it could be tried like this for now, but I just read through that thread and I saw that there was already an example in https://github.com/microsoft/language-server-protocol/issues/724#issuecomment-483206386 where they consider to use it as a "regular" snippet with multiple tab stops. So dependent on how this will end up in the specs, it's probably only a matter of time until a server will use the snippet in this way. For the autocompletion the indentation problem is solved by client announcing [insertTextMode](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#insertTextMode) capability, where this client only supports `adjustIndentation`. I guess something like this would be needed as well for the snippet text edits then. Or alternatively they should introduce a new "simplified snippet" structure that only supports a single tab marker (or guarantee this in some other way in the specs).

---

I wonder don't we need to set `{ "snippetTextEdit": boolean }` experimental client capability for this to work, which is mentioned in the docs from rust-analyzer?

rchl: > I wonder don't we need to set `{ "snippetTextEdit": boolean }` experimental client capability for this to work, which is mentioned in the docs from rust-analyzer?

Looks like they don't explicitly check for the capability for this "move item" functionality (https://github.com/rust-lang/rust-analyzer/blob/f8eac19b3354722a6fa0177968af54a58bb5b9e1/crates/ide/src/move_item.rs#L139-L165). They do check it many other places but in that case I wouldn't go and enable it without first checking that we correctly handle it in all cases (which we probably don't).

plugin/formatting.py
```diff
@@ -50,9 +50,11 @@ def format_document(text_command: LspTextCommand, formatter: Optional[str] = Non
     return Promise.resolve(None)
 
 
-def apply_text_edits_to_view(response: Optional[List[TextEdit]], view: sublime.View) -> None:
+def apply_text_edits_to_view(
+    response: Optional[List[TextEdit]], view: sublime.View, *, process_placeholders: bool = False
+) -> None:
     edits = list(parse_text_edit(change) for change in response) if response else []
-    view.run_command('lsp_apply_document_edit', {'changes': edits})
+    view.run_command('lsp_apply_document_edit', {'changes': edits, 'process_placeholders': process_placeholders})
```
jwortmann, line 57: It's a bit confusing that the `process_placeholders` is also added to this function, even though it is not used anywhere in the code. I see that you used it in the rust-analyzer PR, but `apply_text_edits_to_view` is not part of the public API...

I think it would be better if LspApplyDocumentEditCommand would simply take a list of `TextEdit | AnnotatedTextEdit` and optional `version` argument, instead of a list of custom `TextEditTuple`. With the current design there is no way to properly use LspApplyDocumentEditCommand from the outside when you have no access to `parse_text_edit`. I wonder why this part, which is only a single line (line 56 here), is not simply part of the command anyway to reduce code duplication. Here are the relevant usages of the command, all with the same additional code required before:
```
C:\Users\jwortmann\AppData\Roaming\Sublime Text\Packages\LSP\plugin\color.py:
   59              color_pres = self._filtered_response[index]
   60              text_edit = color_pres.get('textEdit') or {'range': self._range, 'newText': color_pres['label']}
   61:             self.view.run_command('lsp_apply_document_edit', {'changes': [parse_text_edit(text_edit, self._version)]})
   62  

C:\Users\jwortmann\AppData\Roaming\Sublime Text\Packages\LSP\plugin\completion.py:
  373          if additional_edits:
  374              edits = [parse_text_edit(additional_edit) for additional_edit in additional_edits]
  375:             self.view.run_command("lsp_apply_document_edit", {'changes': edits})
  376          command = item.get("command")
  377          if command:

C:\Users\jwortmann\AppData\Roaming\Sublime Text\Packages\LSP\plugin\formatting.py:
   53  def apply_text_edits_to_view(response: Optional[List[TextEdit]], view: sublime.View) -> None:
   54      edits = list(parse_text_edit(change) for change in response) if response else []
   55:     view.run_command('lsp_apply_document_edit', {'changes': edits})
```

plugin/edit.py
jwortmann: Okay. I also realized that the proper place would probably be in the "experimental_capabilities" client config for rust-analyzer anyway.

plugin/formatting.py
rchl: I'm not sure why is that. I imagine potential reasons could be to optimize the payload size or avoid passing unserializable data but neither of those is relevant here IMO.

As for public vs. non-public, we both know that there are many "non-public" APIs that are used from plugins. People will use whatever they need since we don't prevent them :)

jwortmann (approved):
Should be safe to add, since the process_placeholders argument is disabled by default. But we should keep an eye on how it will end up in the LSP specs, for the case if any server will start to use multiple tab stops.

predragnikolic:
Could you add a test for this?

rchl:
I will do some further refactoring separately but for now I think this is fine. If there are any remaining comments (about added tests) then I can address them in the next PR.

PR #2392, opened 2024-01-10
fix missing check for multiline highlighting

titoBouzout:
When `show_multiline_document_highlights` is `false` it shouldn't be highlighting regions that spawn multiple lines

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -820,6 +820,8 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
                 if not regions:
                     continue
                 kind, multiline = tup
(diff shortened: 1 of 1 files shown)


plugin/documents.py
```diff
@@ -820,6 +820,8 @@ def render_highlights_on_main_thread() -> None:
                 if not regions:
                     continue
                 kind, multiline = tup
+                if multiline and not userprefs().show_multiline_document_highlights:
+                    continue
```
rchl, line 824: I don't remember exactly how all of this works but shouldn't the existing same check above prevent those from being added to `kind2regions` in the first place?

titoBouzout: You are right, yes, thanks, it seems I had an unrelated multiline preference set and wasn't working

PR #2393, opened 2024-01-13
refactor: make LspApplyDocumentEditCommand take plain TextEdits

rchl:
Refactor per discussion in https://github.com/sublimelsp/LSP/pull/2389#discussion_r1447905532

- Make `lsp_apply_document_edit` take `[TextEdit]` rather than parsed custom structure.
- Remove private and deprecated `LSP.plugin.core.edit.apply_workspace_edit` function (use `session.apply_workspace_edit_async` instead)
- Remove private `LSP.plugin.formatting.apply_text_edits_to_view` function (use `LSP.plugin.apply_text_edits` instead)
- Move `LSP.plugin.core.edit.parse_text_edit` to `LSP.plugin.edit._parse_text_edit` (and make "private")
- Change return value of private `LSP.plugin.core.edit.parse_workspace_edit` function (returns unparsed edits)
- Expose `apply_text_edits` as a public API

In one test I had to change the test for overflow handling from passing `sys.maxsize + 1` offset to `UINT_MAX + 1` offset because ST API won't allow `sys.maxsize + 1`. I guess this could be one minor reason why parsing before triggering `lsp_apply_document_edit` was there but since LSP spec limits to `UINT_MAX` anyway, it should be fine.

Note that changes to those private APIs would break a couple of LSP packages so I'll be creating PRs for those and linking here.

Also note that I've exposed `apply_text_edits` as a public API since I am very much in favor of exposing an API function rather than LSP packages calling the ST command directly since the latter is not type-checked.

Needs to be released simultaneously with:
- https://github.com/sublimelsp/LSP-rust-analyzer/pull/113
- https://github.com/sublimelsp/LSP-cspell/pull/12
- https://github.com/sublimelsp/LSP-json/pull/176
- https://github.com/sublimelsp/LSP-jdtls/pull/47

### DIFF
--- a/plugin/core/edit.py
+++ b/plugin/core/edit.py
@@ -1,21 +1,17 @@
 from .logging import debug
-from .open import open_file
-from .promise import Promise
 from .protocol import Position
 from .protocol import TextEdit
 from .protocol import UINT_MAX
 from .protocol import WorkspaceEdit
 from .typing import List, Dict, Optional, Tuple
-from functools import partial
 import sublime
 
 
-# tuple of start, end, newText, version
-TextEditTuple = Tuple[Tuple[int, int], Tuple[int, int], str, Optional[int]]
+WorkspaceChanges = Dict[str, Tuple[List[TextEdit], Optional[int]]]
 
 
-def parse_workspace_edit(workspace_edit: WorkspaceEdit) -> Dict[str, List[TextEditTuple]]:
-    changes = {}  # type: Dict[str, List[TextEditTuple]]
+def parse_workspace_edit(workspace_edit: WorkspaceEdit) -> WorkspaceChanges:
+    changes = {}  # type: WorkspaceChanges
     document_changes = workspace_edit.get('documentChanges')
     if isinstance(document_changes, list):
         for document_change in document_changes:
@@ -26,13 +22,13 @@ def parse_workspace_edit(workspace_edit: WorkspaceEdit) -> Dict[str, List[TextEd
             text_document = document_change["textDocument"]
             uri = text_document['uri']
             version = text_document.get('version')
-            text_edit = list(parse_text_edit(change, version) for change in document_change.get('edits'))
-            changes.setdefault(uri, []).extend(text_edit)
+            edits = document_change.get('edits')
+            changes.setdefault(uri, ([], version))[0].extend(edits)
     else:
         raw_changes = workspace_edit.get('changes')
         if isinstance(raw_changes, dict):
-            for uri, uri_changes in raw_changes.items():
-                changes[uri] = list(parse_text_edit(change) for change in uri_changes)
+            for uri, edits in raw_changes.items():
+                changes[uri] = (edits, None)
     return changes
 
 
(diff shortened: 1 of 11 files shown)


plugin/core/edit.py
```diff
@@ -26,38 +22,32 @@ def parse_workspace_edit(workspace_edit: WorkspaceEdit) -> Dict[str, List[TextEd
...
+    required_view_version: Optional[int] = None
+) -> None:
+    view.run_command(
+        'lsp_apply_document_edit',
+        {
+            'changes': edits or [],
+            'process_placeholders': process_placeholders,
+            'required_view_version': required_view_version,
+        }
     )
```
jwortmann, line 53: I would do it like this:
```suggestion
    if edits:
        view.run_command(
            'lsp_apply_document_edit',
            {
                'changes': edits,
                'process_placeholders': process_placeholders,
                'required_view_version': required_view_version,
            }
        )
```

rchl: addressed with early exit to avoid extra indent

predragnikolic:
When I saw the PR, I immediately wrote a list of plugins affected by this change:

<details>

```
Remove private LSP.plugin.formatting.apply_text_edits_to_view function (use LSP.plugin.apply_text_edits instead):

The following plugins should be updated:
~/.config/sublime-text/Packages/LSP-cspell/plugin.py:
    5: from LSP.plugin.formatting import apply_text_edits_to_view

~/.config/sublime-text/Packages/LSP-rust-analyzer/plugin_commands.py:
   10: from LSP.plugin.formatting import apply_text_edits_to_view

Move LSP.plugin.core.edit.parse_text_edit to LSP.plugin.edit._parse_text_edit (and make "private")

LSP tests and LSP-json needs to be updated:

~/.config/sublime-text/Packages/LSP-json/commands.py:
    3: from LSP.plugin.core.edit import parse_text_edit
   39:         edits = [parse_text_edit(change) for change in response]
   40          self.view.run_command('lsp_apply_document_edit', {'changes': edits}


Change return value of private LSP.plugin.core.edit.parse_workspace_edit function (returns unparsed edits)
   ~/.config/sublime-text/Packages/LSP-jdtls/modules/workspace_execute_command_handler.py:
    5: from LSP.plugin.core.edit import apply_workspace_edit, parse_workspace_edit
   45  def _apply_workspace_edit(session: Session, done: Callable[[], None], *arguments):
   46:     changes = parse_workspace_edit(arguments[0])

~/.config/sublime-text/Packages/LSP-jdtls/modules/test_extension_commands.py:
    6: from LSP.plugin.core.edit import WorkspaceEdit, parse_workspace_edit
   56:             parsed_worspace_edit = parse_workspace_edit(workspace_edit)
```

</details>

... I continued to read the PR description only to find out that this work had already been done :)

rchl:
I guess it's good that you've double checked. I might be missing some repos although it seems like we aligned in this case.

plugin/core/edit.py
jwortmann: You can remove the ` or []` now

PR #2395, opened 2024-01-19
include lsp_utils settings in LSP

rchl:
Doing as suggested in https://github.com/sublimelsp/LSP/pull/2388 and moving `lsp_utils` settings into `LSP`. A temporary step until we migrate to a new plugin API that I've started looking into.


predragnikolic (approved):
Tested, works!

PR #2397, opened 2024-01-22
Add format on paste

predragnikolic:
I created a new PR for the "format on paste" feat - I didn't want to resolve the merge conflict with https://github.com/sublimelsp/LSP/pull/2311

TODO address these two comments:
- https://github.com/sublimelsp/LSP/pull/2311#discussion_r1303323097 (done - [07f23a2](https://github.com/sublimelsp/LSP/pull/2397/commits/07f23a2af10eb3453dd5bf91db721158b163ad4d))
- https://github.com/sublimelsp/LSP/pull/2311#discussion_r1299431409 (skipping this)


plugin/documents.py
```diff
@@ -959,6 +965,32 @@ def _update_stored_selection_async(self) -> Tuple[Optional[sublime.Region], bool
...
+                # Including whitespace may help servers format a range better
+                # More info at https://github.com/sublimelsp/LSP/pull/2311#issuecomment-1688593038
+                a = self.view.find_by_class(pasted_region.a, False,
+                                            sublime.CLASS_WORD_END | sublime.CLASS_PUNCTUATION_END)
+                formatting_region = sublime.Region(a, pasted_region.b)
+                regions_to_format.append(formatting_region)
+        sel.add_all(regions_to_format)
+        self.view.run_command('lsp_format_document_range')
+        sel.clear()
+        sel.add_all(original_selection)
```
predragnikolic, line 992: Hello Rafal, the last time you left this:

> We can probably avoid all this if the selection only contains empty regions.

I wanted to address this, but didn't know exactly how to achieve what you wanted.
could you maybe add a code snippet?

plugin/documents.py
```diff
@@ -959,6 +965,32 @@ def _update_stored_selection_async(self) -> Tuple[Optional[sublime.Region], bool
...
+        # add regions to selection, in order for lsp_format_document_range to format those regions
+        for index, region in enumerate(sel):
+            if multi_cursor_paste:
+                pasted_text = split_clipboard_text[index]
+            pasted_region = self.view.find(pasted_text, region.end(), sublime.REVERSE | sublime.LITERAL)
+            if pasted_region:
+                # Including whitespace may help servers format a range better
+                # More info at https://github.com/sublimelsp/LSP/pull/2311#issuecomment-1688593038
+                a = self.view.find_by_class(pasted_region.a, False,
+                                            sublime.CLASS_WORD_END | sublime.CLASS_PUNCTUATION_END)
```
predragnikolic, line 986: @jwortmann your original code snippet contained these flags `sublime.CLASS_WORD_END | sublime.CLASS_PUNCTUATION_END | sublime.CLASS_LINE_START`. As you can see above `sublime.CLASS_LINE_START` was left out. 

If `sublime.CLASS_LINE_START` was specified, then LSP-json would not format a region as expected.

predragnikolic:
All other code review comments left in #2311 were addressed in this PR.

plugin/documents.py
rchl: I don't know what I meant there. Maybe I've misunderstood the code there...

LSP.sublime-settings
```diff
@@ -7,6 +7,9 @@
 
   // --- Document Formatting ------------------------------------------------------------
 
+  // Run the server's documentRangeFormattingProvider (if supported) on the pasted text.
+  "format_on_paste": false,
```
rchl, line 11: Worth to make it a global setting (`lsp_format_on_paste`) maybe? Then it's at least possible to set it per syntax and per project (like `lsp_format_on_save`) which might be useful given how different servers can handle range formatting better or worse.

plugin/documents.py
```diff
@@ -959,6 +965,32 @@ def _update_stored_selection_async(self) -> Tuple[Optional[sublime.Region], bool
         self._stored_selection = selection
         return changed_first_region, True
 
+    def _format_on_paste_async(self) -> None:
+        self.purge_changes_async()
```
rchl, line 969: This could be moved to just before `run_command`

plugin/documents.py
```diff
@@ -936,6 +939,9 @@ def _on_view_updated_async(self) -> None:
             self._when_selection_remains_stable_async(
                 self._do_highlights_async, first_region, after_ms=self.highlights_debounce_time)
         self.do_signature_help_async(manual=False)
+        if self._did_paste:
+            self._did_paste = False
+            self._format_on_paste_async()
```
rchl, line 944: Think this should be before `do_signature_help_async` so that it can still trigger after paste.

plugin/documents.py
```diff
@@ -528,7 +529,9 @@ def on_text_command(self, command_name: str, args: Optional[dict]) -> Optional[T
         return None
 
     def on_post_text_command(self, command_name: str, args: Optional[Dict[str, Any]]) -> None:
-        if command_name in ("next_field", "prev_field") and args is None:
+        if command_name == 'paste' and userprefs().format_on_paste:
+            self._did_paste = True
```
rchl, line 533: It would kinda make sense to me that this functionality can be triggered on `paste_and_indent` (only or also) since it should work like an enhenced version of ST's native functionality.

Though not sure how this could be done.

I was thinking that there could be an LSP command that the user could assign custom keybinding to but that would have to work rather differently than what you did.

Alternatively the `format_on_save` could maybe be set to names of the commands that it should override so that one can set `paste_and_indent` only...

predragnikolic: https://github.com/sublimelsp/LSP/pull/2311#discussion_r1298568537

@jwortmann has the same comment.
 > > There exists also a command paste_and_indent (Ctrl+Shift+V).
Should it run the formatting too, when that command was invoked?
> 
>I would not add it, but don't have any strong opinions about it.
If you think that formatting should happen, I will add it.


Well you are the second person who asked the same question.
So I think I will add it.

> It would kinda make sense to me that this functionality can be triggered on paste_and_indent (only or also) since it should work like an enhenced version of ST's native functionality.

Are you suggesting the following:
If a user has `lsp_format_on_paste: false`,
and a user runs  `paste_and_indent`,
LSP should format the pasted range?

```diff
        if command_name == 'paste' and lsp_format_on_paste_enabled:
            self._did_paste = True
+        elif command_name == 'paste_and_indent':
+            self._did_paste = True
```
-----

Here was my proposal.
If a user has `lsp_format_on_paste: true`,
then regardless if the user run `paste` or `paste_and_indent`,
LSP will format the pasted range.

WDYT?


```diff
- if command_name == 'paste' and lsp_format_on_paste_enabled:
+ if command_name in ['paste', 'paste_and_indent'] and lsp_format_on_paste_enabled:
```

rchl: The latter is what I meant but note that once `paste_and_ident` reformats things on its own, your code that tries to find pasted contents will likely fail...

plugin/documents.py
```diff
@@ -528,7 +529,12 @@ def on_text_command(self, command_name: str, args: Optional[dict]) -> Optional[T
         return None
 
     def on_post_text_command(self, command_name: str, args: Optional[Dict[str, Any]]) -> None:
-        if command_name in ("next_field", "prev_field") and args is None:
+        format_on_paste = self.view.settings().get('lsp_format_on_paste', None)
+        format_on_paste_enabled = format_on_paste if isinstance(format_on_paste, bool) \
+            else userprefs().lsp_format_on_paste
```
rchl, line 534: Should be fine with just:

```
self.view.settings().get('lsp_format_on_paste', userprefs().lsp_format_on_paste)
```

sublime-package.json
```diff
@@ -444,6 +449,11 @@
               "default": false,
               "markdownDescription": "Show errors and warnings count in the status bar."
             },
+            "lsp_format_on_paste": {
+              "type": "boolean",
+              "default": false,
+              "markdownDescription": "Run the server's documentRangeFormattingProvider (if supported) on the pasted text."
+            },
```
rchl, line 456: reference a definition like `lsp_format_on_save` does below.

plugin/documents.py
predragnikolic: [c9d1b85](https://github.com/sublimelsp/LSP/pull/2397/commits/c9d1b852b55e41a166c73251da2e9a8be09842f2)

in this commit, I add code to intercept `paste_and_indent` and instead run the `paste` command.
> ... `paste_and_indent` reformats things on its own

as you noted,
I didn't want to add code to improve finding the reformatted text.
This approach didn't require a lot of code 
and while the intention is not that obvious obvious, I added a comment to explain why it was added.

plugin/documents.py
```diff
@@ -520,12 +520,18 @@ def _on_hover_gutter_async(self, point: int) -> None:
...
             self._auto_complete_triggered_manually = True
         elif command_name == "show_scope_name" and userprefs().semantic_highlighting:
             session = self.session_async("semanticTokensProvider")
             if session:
                 return ("lsp_show_scope_name", {})
+        elif command_name == 'paste_and_indent' and format_on_paste:
+            # if `lsp_format_on_paste` is enabled and paste_and_indent is run,
+            # it is easier to find the region to format when `paste` is invoked,
+            # so we intercept the `paste_and_indent` and replace it with the `paste` command.
+            return ('paste', {})
```
jwortmann, line 534: I'm not sure right now, but isn't this ViewEventListener created for all views, regardless of whether LSP is active or not? So I think this needs something like
```python
session = self.session_async(...)
if session:
    ...
```
otherwise it would always modifiy this command even when used for unrelated files. This check is also needed for the case if a server doesn't support range formatting. Also I'm unsure whether this is the best approach to do it, but I haven't thought about alternatives yet (but maybe it's fine).

rchl: I would really prefer that we let ST format things and then do formatting on top of that but I won't demand that as I know it would be rather hard to handle it in a clean way.

predragnikolic: I will look into this, thanks

rchl: I'll just say right away that if the implementation would have to get too complex or hacky then I'd prefer that you don't waste your time.

predragnikolic: > oh, I suppose you were referring to the actual issue that @jwortmann raised.

Yes, I was referring to the message from @jwortmann. I didn't saw your message at the time I was writing the comment.

predragnikolic: [f654bc8](https://github.com/sublimelsp/LSP/pull/2397/commits/f654bc813a0aa7b222efe461934759e3ca794a94)

plugin/documents.py
```diff
@@ -519,16 +520,27 @@ def _on_hover_gutter_async(self, point: int) -> None:
...
             self._auto_complete_triggered_manually = True
         elif command_name == "show_scope_name" and userprefs().semantic_highlighting:
             session = self.session_async("semanticTokensProvider")
             if session:
                 return ("lsp_show_scope_name", {})
+        elif command_name == 'paste_and_indent' and format_on_paste:
+            # if `lsp_format_on_paste` is enabled and paste_and_indent is run,
+            # it is easier to find the region to format when `paste` is invoked,
+            # so we intercept the `paste_and_indent` and replace it with the `paste` command.
+            if self.session_async("documentRangeFormattingProvider"):
```
rchl, line 534: Minor but would prefer it like that so that all related logic is kept together and we also do less work.

```suggestion
        elif command_name == 'paste_and_indent'
            # if `lsp_format_on_paste` is enabled and paste_and_indent is run,
            # it is easier to find the region to format when `paste` is invoked,
            # so we intercept the `paste_and_indent` and replace it with the `paste` command.
            format_on_paste = self.view.settings().get('lsp_format_on_paste', userprefs().lsp_format_on_paste):
            if format_on_paste and self.session_async("documentRangeFormattingProvider"):
```

Same below.

rchl: BTW. Most of this text is redundant since it's self-explanatory through the code itself. Basically only some variant of "it is easier to find the region to format when `paste` is invoked" is something that would be worth keeping.

plugin/documents.py
```diff
@@ -959,6 +974,32 @@ def _update_stored_selection_async(self) -> Tuple[Optional[sublime.Region], bool
...
+                # More info at https://github.com/sublimelsp/LSP/pull/2311#issuecomment-1688593038
+                a = self.view.find_by_class(pasted_region.a, False,
+                                            sublime.CLASS_WORD_END | sublime.CLASS_PUNCTUATION_END)
+                formatting_region = sublime.Region(a, pasted_region.b)
+                regions_to_format.append(formatting_region)
+        sel.add_all(regions_to_format)
+        self.purge_changes_async()
+        self.view.run_command('lsp_format_document_range')
+        sel.clear()
+        sel.add_all(original_selection)
```
rchl, line 1001: Hmm, I was pretty sure that ST always runs commands on the main thread... If that would be the case then you would restore the original selection before the request would actually be sent. Can you verify whether this works as expected?

plugin/documents.py
predragnikolic: there is a syntax error, but I will push fix for it

plugin/documents.py
predragnikolic: I put a print statements:
```py
        self.purge_changes_async()
        self.view.run_command('lsp_format_document_range')
        print('restore')
        sel.clear()
        sel.add_all(original_selection)
# ...

class LspFormatDocumentRangeCommand(LspTextCommand):
        print('command invoked')
        if has_single_nonempty_selection(self.vi
```

first is the command invoked then the selection is restored:
```
command invoked
restore
```

It looks ok to me.

plugin/documents.py
```diff
@@ -519,16 +520,27 @@ def _on_hover_gutter_async(self, point: int) -> None:
                 on_navigate=lambda href: self._on_navigate(href, point))
 
     def on_text_command(self, command_name: str, args: Optional[dict]) -> Optional[Tuple[str, dict]]:
+        format_on_paste = self.view.settings().get('lsp_format_on_paste', userprefs().lsp_format_on_paste)
```
rchl, line 523: remove now

plugin/documents.py
```diff
@@ -519,16 +520,27 @@ def _on_hover_gutter_async(self, point: int) -> None:
...
+            # it is easier to find the region to format when `paste` is invoked,
+            # so we intercept the `paste_and_indent` and replace it with the `paste` command.
+            format_on_paste = self.view.settings().get('lsp_format_on_paste', userprefs().lsp_format_on_paste)
+            if format_on_paste and self.session_async("documentRangeFormattingProvider"):
+                return ('paste', {})
         return None
 
     def on_post_text_command(self, command_name: str, args: Optional[Dict[str, Any]]) -> None:
-        if command_name in ("next_field", "prev_field") and args is None:
+        format_on_paste = self.view.settings().get('lsp_format_on_paste', userprefs().lsp_format_on_paste)
```
rchl, line 539: move inside `if`

plugin/documents.py
rchl: Didn't realize before but ST appears to be blocking the calling (async) thread while the command runs on the main thread. So this acts like a sync code more or less.

I always thought that it schedules a task on the main thread.

rchl (approved):

predragnikolic:
There was some minor visual annoyance, where I could sometimes see the selected regions, before they get formatted, I addressed that with [0cfcc71](https://github.com/sublimelsp/LSP/pull/2397/commits/0cfcc71b41e9bd4e898a0ff8e9b864d9a97399f8)

LSP.sublime-settings
```diff
@@ -7,6 +7,9 @@
 
   // --- Document Formatting ------------------------------------------------------------
 
+  // Run the server's documentRangeFormattingProvider (if supported) on the pasted text.
```
jwortmann, line 10: I see that this comment was formulated similar to the next comment. But I think it would me more user friendly to reword it without using the capability names from the specs (documentRangeFormattingProvider). "If supported, format the range of the pasted text"

And for the next setting: "Format a file before saving"

rchl (approved):

PR #2400, opened 2024-01-25
docs: add asm-lsp

WillLillis:
Hi! Would it be possible to add [asm-lsp](https://github.com/bergercookie/asm-lsp) to the docs?

I wasn't super sure what to list for the selectors in the configuration. Checking the selector on my test files (see `samples/` in the project's repo) yielded `text.plain`. I tried finding a more comprehensive lists of scopes to use a more assembly-specific scope, but the best I could find was [this](https://stackoverflow.com/questions/30153110/get-all-scope-names-on-sublime-text-3) stack overflow post. The post listed `source.disasm`, but I struggled to find a file that used that. I tested the configuration as-is and it works, but I'm happy to change anything if I've made a mistake :)

### DIFF
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -16,7 +16,7 @@ If there are no setup steps for a language server on this page, but a [language 
 
 Follow installation instructions on [LSP-angular](https://github.com/sublimelsp/LSP-angular).
 
-## ASM LSP
+## Assembly
 
 1. Install `asm-lsp` via Cargo (see [github:bergercookie/asm-lsp](https://github.com/bergercookie/asm-lsp)):
 
@@ -32,7 +32,7 @@ Follow installation instructions on [LSP-angular](https://github.com/sublimelsp/
             "asm-lsp": {
                 "enabled": true,
                 "command": ["asm-lsp"],
-                "selector": "text.plain | source.disasm"
+                "selector": "source.asm | source.assembly"
             }
         }
     }
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -24,7 +24,9 @@ Follow installation instructions on [LSP-angular](https://github.com/sublimelsp/
     cargo install asm-lsp
     ```
 
-2. Open `Preferences > Package Settings > LSP > Settings` and add the `"asm-lsp"` client configuration to the `"clients"`:
+2. Install the [x86 and x86_64 Assembly](https://packagecontrol.io/packages/x86%20and%20x86_64%20Assembly) package from Package Control.
+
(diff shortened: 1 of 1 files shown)


jwortmann:
A quick search on the Package Control website lists [NASM x​86 Assembly](https://packagecontrol.io/packages/NASM%20x86%20Assembly) and [x​86 and x​86_64 Assembly](https://packagecontrol.io/packages/x86%20and%20x86_64%20Assembly) as two popupar syntax highlighting packages for assembly languages, and they use the scope names [`source.assembly`](https://github.com/Nessphoro/sublimeassembly/blob/38d3fd64e5b981ac585f5c9e361647c09e208935/Assembly%20x86.tmLanguage#L185) and [`source.asm`](https://github.com/13xforever/x86-assembly-textmate-bundle/blob/a8066379d1eaa3574c82e69fce2072372389057f/x86_64%20Assembly.tmbundle/Syntaxes/Nasm%20Assembly.sublime-syntax#L4). I do think that `text.plain` should be removed, so I would probably use `"selector": "source.asm | source.assembly"`, assuming that files highlighted by the mentioned syntax packages are applicable for this LSP server.

The h2 heading is supposed to be the language name, so I would probably put `## Assembly` there instead of `## ASM LSP`.

rchl:
BTW. based on [this search](https://github.com/search?q=repo%3Abergercookie%2Fasm-lsp%20language_id&type=code) I would guess that the server doesn't care about the language ID.

WillLillis:
> I would probably use "selector": "source.asm | source.assembly"

After installing the two packages you linked the selector for the file showed up as `source.asm.x86_64`. The highlights were also applied to the files. Thanks!

> The h2 heading is supposed to be the language name, so I would probably put ## Assembly there instead of ## ASM LSP

Makes sense! Sorry about that

WillLillis:
> BTW. based on [this search](https://github.com/search?q=repo%3Abergercookie%2Fasm-lsp%20language_id&type=code) I would guess that the server doesn't care about the language ID.

That would be correct. I've contributed a couple features to the server and we don't have any strict language requirements. :)

docs/src/language_servers.md
```diff
@@ -16,6 +16,28 @@ If there are no setup steps for a language server on this page, but a [language
...
+    {
+        "clients": {
+            "asm-lsp": {
+                "enabled": true,
+                "command": ["asm-lsp"],
+                "selector": "source.asm | source.assembly"
+            }
+        }
+    }
+    ```
```
rchl, line 39: You should also specify which syntax should be installed since the server won't start without it (since the scope won't match).

rchl (approved):

PR #2401, opened 2024-01-25
fix KeyError exception triggered on closing a window

### LINKED ISSUE #2399: `KeyError` in LSP plugin when trying to open Preferences
My ST4 setup: version: 4169 osx arm64 channel: stable

LSP v1.27.0

I see this trace in the console whenever I try to open Settings/Preferences (it opens successfully, but this error worries me):

```
Unable to open /Users/samir.bajaj/Library/Application Support/Sublime Text/Packages/Default/Preferences.sublime-settings
  File "/Users/samir.bajaj/Library/Application Support/Sublime Text/Installed Packages/LSP.sublime-package/plugin/documents.py", line 968, in clear_async
  File "/Users/samir.bajaj/Library/Application Support/Sublime Text/Installed Packages/LSP.sublime-package/plugin/session_view.py", line 104, in on_before_remove
  File "/Users/samir.bajaj/Library/Application Support/Sublime Text/Installed Packages/LSP.sublime-package/plugin/session_buffer.py", line 203, in remove_session_view
  File "./python3.3/_weakrefset.py", line 109, in remove
KeyError: <weakref at 0x104a93ef8; to 'SessionView' at 0x104a5b7b8>
```

I originally posted this on the Sublime Forum, but got no responses.

Thank you for your help.

rchl:
`wm.destroy` is called in two cases:
- from `EventListener.on_pre_close_window`, before the window is closed
- from `plugin_unloaded` hook when LSP package is unloaded

The changed line is triggered in the former case.

As the [comment in `WindowManager.destroy()` says](https://github.com/sublimelsp/LSP/blob/f10f45e4323fbe39e2f138dd9f7ce47544c110d6/plugin/core/windows.py#L402-L406), we should (have to!) call it from the main thread when the plugin is unloaded. But afaics, we don't need to call it from the main thread when closing a window so change the code to not do that.

Calling it from the main thread on window closing triggers a `KeyError` exception due to `DocumentListener.on_close` being triggered first and `DocumentListener.on_session_shutdown_async` later. The latter removes session view from the main thread first and the former tries to remove it again from scheduled execution on the async thread.

Fixes #2399

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -545,7 +545,7 @@ class WindowRegistry:
     def discard(self, window: sublime.Window) -> None:
         wm = self._windows.pop(window.id(), None)
         if wm:
-            wm.destroy()
+            sublime.set_timeout_async(wm.destroy)
 
 
 class RequestTimeTracker:


predragnikolic:
Not that related to the PR,
but I have a question,

If you have multiple ST windows, and invoke  WindowRegistry.disable, that will nuke all WindowManagers from all ST windows?
If you have multiple ST windows, and invoke WindowRegistry.discard, that will discard only the WindowManager for one ST window?

rchl:
Yeah, as you said. The `discard` function takes a `window` argument while `disable` doesn't.

samirbajaj:
Thank you @rchl for fixing this issue -- much appreciated!

Reviewers -- please approve and merge. Thank you.

predragnikolic:
I was **only** able to reproduce it in safe mode.

I cannot reproduce the exception with this PR.

predragnikolic (approved):

PR #2406, opened 2024-01-26
docs: rewritten "common problems" section

rchl:
Refactored "Common problems" section:

- Combined `Popup error Language server <your_server_language_name> has crashed` and `LSP cannot find my language server` into a single `Error dialog saying Failed to start...` section. That's because both of those issues trigger the same error dialog, only with some potentially different errors at the bottom of the dialog.
- Removed some very obscure `the language servers may have dependencies that should also be in your PATH` suggestion specific to haskell server and instead linking to "Self help" section that should provide more generic understanding of whatever the issue might be.
- Removed "Known issues" header since the `Completions not shown after certain keywords` section can easily be considered "Common problems" instead. Note that this is also pretty obscure issue and not sure it really needs to be here. Haven't seen any issues like that for a long time.

### DIFF
--- a/docs/src/troubleshooting.md
+++ b/docs/src/troubleshooting.md
@@ -62,31 +62,21 @@ Another solution could be (at least on Linux) to update the server `PATH` using 
 
 ## Common problems
 
-### 1. LSP doesn't start my language server
+### Error dialog saying `Failed to start...`
+
+If you are getting an error that the server binary can't be found (`No such file or directory...`) but it does start when running it from the terminal, then the issue is likely due to Sublime Text's internal environment not picking up the same `PATH` environment variable as you've configured in your shell. See ["Updating the PATH used by LSP servers"](troubleshooting.md#updating-the-path-used-by-lsp-servers) on how to make Sublime Text aware of the location of your language server.
+
+Otherwise refer to the ["Self-help instructions"](troubleshooting.md#self-help-instructions) section to try to understand the issue better.
+
+### LSP doesn't start my language server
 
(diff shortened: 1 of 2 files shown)


docs/src/troubleshooting.md
```diff
@@ -62,31 +62,21 @@ Another solution could be (at least on Linux) to update the server `PATH` using
...
-* Make sure that the language server is not disabled globally either in its own settings or in `Preferences: LSP Settings`, or in the project settings (`Project: Edit Project` from the Command Palette).
-
-### 2. LSP cannot find my language server (`No such file or directory: 'xyz'`)
-
-If you are getting an error that the server binary can't be found but it does start when running it from the terminal, then the issue is likely due to Sublime Text's internal environment not picking up the same `PATH` environment variable as you've configured in your shell.
+If you are getting an error that the server binary can't be found (`No such file or directory...`) but it does start when running it from the terminal, then the issue is likely due to Sublime Text's internal environment not picking up the same `PATH` environment variable as you've configured in your shell. See ["Updating the PATH used by LSP servers"](troubleshooting.md#updating-the-path-used-by-lsp-servers) on how to make Sublime Text aware of the location of your language server.
```
predragnikolic, line 67:
```suggestion
If you are getting an error that the server binary can't be found (`No such file or directory...`) but it does start when running Sublime Text from the terminal, then the issue is likely due to Sublime Text's internal environment not picking up the same `PATH` environment variable as you've configured in your shell. See ["Updating the PATH used by LSP servers"](troubleshooting.md#updating-the-path-used-by-lsp-servers) on how to make Sublime Text aware of the location of your language server.
```

predragnikolic (approved):

Issue #2084, opened 2022-10-11
Ability to show `Find References` in the output panel on demand


rchl:
**Is your feature request related to a problem? Please describe.**
There is an option (`show_references_in_quick_panel`) to choose whether `Find References` should use the Output Panel or the Quick Panel for showing references but while I usually want to use the Quick Panel, occasionally I would want to have it output to the Output Panel so that I can jump between all references without having to re-trigger "Find References" all the time.

**Describe the solution you'd like**
- Extend `lsp_symbol_references` with a parameter to choose which output format to use.
- (Optionally) Expose additional command in the Command Palette/Menus. In that case, figure out the name and how it should be named depending on what is the state of the `show_references_in_quick_panel` option.

**Describe alternatives you've considered**
Toggling the `show_references_in_quick_panel` when needing to use the Output Panel occasionally.

rchl:
I've also considered changing the existing behavior of `LSP: Find References` command in the Command Palette depending on whether shift is pressed but unfortunately when triggered commands from there the `event` object is not passed, even if command defines `want_event(self): True`

rchl:
Actually I've realized that my use case would be covered if the quick panel would select by default to file and location closest to the current file/position. Then it wouldn't always scroll me to the first file in the list and I could jump from current to the next one easily.

rchl:
The improvement done in https://github.com/sublimelsp/LSP/pull/2337 should satisfy this use case and I feel doing exactly what I've suggested would complicate things too much.

PR #2409, opened 2024-01-28
add commands for opening "find references" in bottom or quick panel

rchl:
Include an additional variant of the `Find references` command in Command Palette and menus that allows to show references in the way that is opposite to the default behavior (which depends on `show_references_in_quick_panel`).

So when `show_references_in_quick_panel` is enabled, we'll show `Find References (in bottom panel)` in addition to the default `Find References` and if `show_references_in_quick_panel` is disabled we'll show `Find References (in quick panel)` so that both variants are always accessible.

The reason for that is that while showing references in the quick panel is typically the preferred way, one might want to occasionally show those in the bottom panel to be able to keep the list as a reference when wanting to update all results manually.

### DIFF
--- a/plugin/__init__.py
+++ b/plugin/__init__.py
@@ -12,6 +12,7 @@ from .core.protocol import Response
 from .core.registry import LspTextCommand
 from .core.registry import LspWindowCommand
 from .core.sessions import AbstractPlugin
+from .core.sessions import AbstractPluginV2
 from .core.sessions import register_plugin
 from .core.sessions import Session
 from .core.sessions import SessionBufferProtocol
@@ -30,6 +31,7 @@ from .core.workspace import WorkspaceFolder
 __all__ = [
     '__version__',
     'AbstractPlugin',
+    'AbstractPluginV2',
     'apply_text_edits',
     'ClientConfig',
     'css',
--- a/plugin/__init__.py
+++ b/plugin/__init__.py
@@ -12,7 +12,6 @@ from .core.protocol import Response
 from .core.registry import LspTextCommand
 from .core.registry import LspWindowCommand
 from .core.sessions import AbstractPlugin
-from .core.sessions import AbstractPluginV2
 from .core.sessions import register_plugin
 from .core.sessions import Session
 from .core.sessions import SessionBufferProtocol
@@ -31,7 +30,6 @@ from .core.workspace import WorkspaceFolder
 __all__ = [
     '__version__',
     'AbstractPlugin',
-    'AbstractPluginV2',
     'apply_text_edits',
     'ClientConfig',
     'css',
--- a/plugin/references.py
+++ b/plugin/references.py
@@ -99,6 +99,7 @@ class LspSymbolReferencesCommand(LspTextCommand):
                     fallback,
                     group,
                     show_in,
+                    event,
                     word_range.begin()
                 )
             )
@@ -114,11 +115,12 @@ class LspSymbolReferencesCommand(LspTextCommand):
         fallback: bool,
         group: int,
         show_in: Optional[ShowInArgument],
+        event: Optional[dict],
         position: int,
         response: Optional[List[Location]]
     ) -> None:
         sublime.set_timeout(lambda: self._handle_response(
-            word, session, side_by_side, force_group, fallback, group, show_in, position, response))
+            word, session, side_by_side, force_group, fallback, group, show_in, event, position, response))
 
     def _handle_response(
         self,
@@ -129,17 +131,21 @@ class LspSymbolReferencesCommand(LspTextCommand):
         fallback: bool,
         group: int,
         show_in: Optional[ShowInArgument],
+        event: Optional[dict],
         position: int,
         response: Optional[List[Location]]
     ) -> None:
-        if response:
-            if show_in == 'quick_panel' or (show_in is None and userprefs().show_references_in_quick_panel):
-                self._show_references_in_quick_panel(
-                    word, session, response, side_by_side, force_group, group, position)
-            else:
-                self._show_references_in_output_panel(word, session, response)
-        else:
+        if not response:
             self._handle_no_results(fallback, side_by_side)
+            return
+        modifier_keys = (event or {}).get('modifier_keys', {})
+        show_in_quick_panel = show_in == 'quick_panel' or show_in is None and userprefs().show_references_in_quick_panel
+        if modifier_keys.get('primary') and show_in_quick_panel and side_by_side is False:
+            side_by_side = True
+        if show_in_quick_panel:
+            self._show_references_in_quick_panel(word, session, response, side_by_side, force_group, group, position)
+        else:
+            self._show_references_in_output_panel(word, session, response)
 
     def _handle_no_results(self, fallback: bool = False, side_by_side: bool = False) -> None:
         window = self.view.window()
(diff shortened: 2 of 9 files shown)


jwortmann:
For the context menu, would it be better to change to the alternative display method if a modifier key is pressed when you select the "Find References" item? I think the additional entry is okay for the command palette, but I think it's not the best UI to have two of the items with basically the same functionality in the context menu. Modifier keys for the context menu should work fine (you can confirm by adding `print(event)` in `run()`): 
```
{'modifier_keys': {}, 'x': 654.5, 'y': 518.5, 'text_point': 2176}
{'modifier_keys': {'primary': True, 'ctrl': True}, 'x': 652.5, 'y': 518.5, 'text_point': 2176}
```

rchl:
That would make the feature pretty much impossible to discover though...

jwortmann:
I gues most users are satisfied with having the global "how_references_in_quick_panel": true/false setting. The only request for that in the issue tracker I can see is from you: https://github.com/sublimelsp/LSP/issues/2084
I don't want to say that it isn't a valid request, but we could mention it in a sentence in the docs and if someone searches or asks for it, it shouldn't be too hard to find out about it. There are also no additional context menu items for the other command arguments; for example we have the key binding for side-by-side (using the default keys from ST) even commented out, and we have no explicit context menu item "Find References (side by side)"
https://github.com/sublimelsp/LSP/blob/8023408dfd7917d23f89c9bb3f7356a4d2d1d895/Default.sublime-keymap#L70C1-L76C10

jwortmann:
Actually we could even use the modifier key too for the side-by-side mode. This would only be relevant if it is displayed in the quick panel. I.e. activate side-by-side if <kbd>Shift</kbd> is pressed (like in the default key binding) and change panel mode (bottom/quick panel) if <kbd>Primary</kbd> is pressed.

plugin/references.py
```diff
@@ -44,10 +48,18 @@ def is_visible(
...
+        show_in: Optional[ShowInArgument] = None,
     ) -> bool:
+        # We include "in bottom panel" and "in quick panel" variants of `LSP: Find References` in the Command Palette
+        # but we only show the one that is not the same as the default one (per `show_references_in_quick_panel`
+        # setting).
+        if show_in == 'bottom_panel' and not userprefs().show_references_in_quick_panel or \
+                show_in == 'quick_panel' and userprefs().show_references_in_quick_panel:
+            return False
         if self.applies_to_context_menu(event):
-            return self.is_enabled(event, point, side_by_side, fallback)
```
rchl, line 50: BTW. Here we had a bug - passed "fallback" to "force_group".

rchl:
> activate side-by-side if Shift is pressed (like in the default key binding)

Actually, the `primary` key is used to trigger side-by-side. Which does give us a lot of wiggle room.

That means that ideally for the side-by-side triggered from the context menu we'd also need to use primary (ctrl on win, cmd on mac) but then what would we use for "default-inverted" action? Shift? Doesn't feel quite right.
- "alternative" action:

rchl:
I've added support for holding `primary` key while triggering context menu item but haven't added support for triggering "non-default" behavior. I guess the only option is to go with `shift` here?

jwortmann:
Yes, I confused the <kbd>Shift</kbd> and <kbd>Primary</kbd> keys above. I think it is better now than having additional menu items.

A few more suggestions for the labels and names:
- Is there maybe a better name for the "show_in" argument? I'd prefer finding one or two words to describe the argument (usually as a noun) if possible, instead of looking like an incomplete sentence. For example
  - "quick_panel": true/false (this would be bad if another mode should be added in the future though)
  - "display_mode"
  - "presentation"
  - or maybe you have another idea with a better name
- I would probably use "output_panel" instead of "bottom_panel" for the option and for the command palette label. Because it may not be restricted to the bottom forever, if ST adds more flexibility for the panels in a future release (https://github.com/sublimehq/sublime_text/issues/1451)
- I think the word "in" should be removed from the command palette labels, currently it reads a bit as the command would do the search inside the panel. Or alternatively add "show in"
  - "LSP: Find References (Quick Panel)"
  - "LSP: Find References (Output Panel)"

rchl:
The reason I've used "bottom panel" is to match the description of the setting:

```
  // Show symbol references in Sublime's quick panel instead of the bottom panel.
  "show_references_in_quick_panel": true,
```

but I'm fine with changing both.

rchl:
> * "quick_panel": true/false

Don't like this since it's not clear what "false" means.
I don't know if I have good suggestions... `output` or `target` come to mind

predragnikolic:
I tried pressing `shift+enter` in the context menu, but it does nothing for me.
Only when I press just `enter`, then it works (end of the gif when the quick panel opens) 

output

I did a little search in my keybindings,  and found out that I had `shift+enter` in it, so I commented that out, but it still doesn't work.
```
    // {
    //     "command": "lsp_commit_completion_with_opposite_insert_mode",
    //     "keys": [
    //         "shift+enter"
    //     ],
    //     "context": [
    //         {
    //             "key": "auto_complete_visible",
    //             "operator": "equal",
    //             "operand": true
    //         },
    //         {
    //             "key": "lsp.session_with_capability",
    //             "operator": "equal",
    //             "operand": "completionProvider"
    //         }
    //     ]
    // },
```

rchl:
Do other menu items work? If it doesn't work with those that don't handle shift specifically (so basically most of them) then I would assume it's a Linux/Window Manager thing.

plugin/references.py
```diff
@@ -113,17 +130,24 @@ def _handle_response(
...
-            else:
-                self._show_references_in_output_panel(word, session, response)
-        else:
+        if not response:
             self._handle_no_results(fallback, side_by_side)
+            return
+        modifier_keys = (event or {}).get('modifier_keys', {})
+        show_in_quick_panel = show_in == 'quick_panel' or show_in is None and userprefs().show_references_in_quick_panel
+        if modifier_keys.get('shift'):
+            show_in_quick_panel = not show_in_quick_panel
```
rchl, line 144: I've realized that I should only be handling shift when "show_in" is not specified.

rchl:
> > * "quick_panel": true/false
> 
> Don't like this since it's not clear what "false" means. I don't know if I have good suggestions... `output` or `target` come to mind

`show_mode`?

jwortmann:
> `show_mode`?

Hm... this sounds a bit like a "TV show" or something. I like your "output" suggestion, or "output_mode", or "display_mode".

The command caption still has "Bottom Panel" instead of "Output Panel" currently.

plugin/__init__.py
```diff
@@ -12,6 +12,7 @@
 from .core.registry import LspTextCommand
 from .core.registry import LspWindowCommand
 from .core.sessions import AbstractPlugin
+from .core.sessions import AbstractPluginV2
```
predragnikolic, line 15: accidentally committed? :)

plugin/references.py
```diff
@@ -113,17 +130,27 @@ def _handle_response(
...
+        if output_mode is None:
+            show_in_quick_panel = userprefs().show_references_in_quick_panel
+            if modifier_keys.get('shift'):
+                show_in_quick_panel = not show_in_quick_panel
+        else:
+            show_in_quick_panel = output_mode == 'quick_panel'
+        if show_in_quick_panel:
+            if modifier_keys.get('primary'):
+                side_by_side = True
+            self._show_references_in_quick_panel(word, session, response, side_by_side, force_group, group, position)
```
predragnikolic, line 151: this part was a bit hard to look a and understand, but I guess it is necessary if we want to support `primary` and `shift`

plugin/__init__.py
rchl: Yeah, hard at work on the new API ;)

jwortmann (approved):

PR #2424, opened 2024-02-18
docs: add info about typst-lsp commands

istudyatuni:


docs/src/language_servers.md
```diff
@@ -751,6 +752,33 @@ See [Javascript/TypeScript](#javascripttypescript).
...
     ```
 
+5. Optional: to enable some useful commands provided by language server, add the following to `*.sublime-commands` (see more [here](commands.md#execute-server-commands)):
+
+    <!-- how to call: see https://github.com/nvarner/typst-lsp/blob/master/editors/vscode/src/extension.ts -->
+    ```jsonc title="Packages/User/Default.sublime-commands"
+    [
+        // ...
+        {
+            "caption": "typst-lsp - Pin the main file to the currently openning document",
```
rchl, line 762:
```suggestion
            "caption": "typst-lsp - Pin the main file to the currently opened document",
```

?

docs/src/language_servers.md
```diff
@@ -751,6 +752,33 @@ See [Javascript/TypeScript](#javascripttypescript).
     }
     ```
 
+5. Optional: to enable some useful commands provided by language server, add the following to `*.sublime-commands` (see more [here](commands.md#execute-server-commands)):
```
rchl, line 755: I'm not sure if this link is relevant here... Why would someone who is adding this command need an explanation on executing server commands?

docs/src/language_servers.md
```diff
@@ -751,6 +752,33 @@ See [Javascript/TypeScript](#javascripttypescript).
...
+        },
+        {
+            "caption": "typst-lsp - Unpin the main file",
+            "command": "lsp_execute",
+            "args": {
+                "session_name": "typst-lsp",
+                "command_name": "typst-lsp.doPinMain",
+                "command_args": ["detached"]
+            }
+        },
```
rchl, line 778: I don't really understand this functionality but it looks like in VSCode this command unpins and then pins the current file. To be fair, based on the name, I would expect it to just unpin but still weird to see this behavior there.

docs/src/language_servers.md
istudyatuni: Makes sense, then I will copy info about "... by adding an entry to your existing `*.sublime-commands` file or by creating a new one" here

docs/src/language_servers.md
istudyatuni: Hmm, I just copied this from original extension :smile:

docs/src/language_servers.md
istudyatuni: > unpins and then pins the current file

Probably you're right, I missed that, but I will check

istudyatuni: Ok, I'm not sure what's really going on, but it should work, at least in the same way as in the vscode extension

rchl (approved):

PR #2426, opened 2024-02-24
Add support for SnippetTextEdits

jwortmann:
*Untested, but I think something like this could work*

CI will obviously fail until the protocol types are updated to the latest spec version.

This would probably make https://github.com/sublimelsp/LSP/commit/29100425f8c2635d7b33524519c4bf2bbd6ee86f obsolete at some point in the future, assuming that rust-analyzer switches from its custom SnippetTextEdit to the one from the LSP 3.18 specs.

### DIFF
--- a/plugin/core/edit.py
+++ b/plugin/core/edit.py
@@ -1,13 +1,18 @@
 from .logging import debug
 from .protocol import Position
+from .protocol import SnippetTextEdit
 from .protocol import TextEdit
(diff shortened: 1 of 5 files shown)


rchl:
We have `SnippetTextEdit` type now.

jwortmann:
I think I will probably revisit this PR when there is a server which supports the SnippetTextEdit, so that it can be tested.

PR #2427, opened 2024-02-28
Fix usage of sublime.score_selector

BenjaminSchaaf:
ST 4173 tweaks how selector scoring works resulting in LSP no longer functioning.

As per the comment there seems to be a misunderstanding of how `sublime.score_selector` works. The entire selector must always match for a non-zero score, so if the goal is to check if it's not empty and matches then checking for `>= 8` is simply wrong.

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -840,11 +840,10 @@ class ClientConfig:
         syntax = view.syntax()
         if not syntax:
             return False
-        # Every part of a x.y.z scope seems to contribute 8.
-        # An empty selector result in a score of 1.
-        # A non-matching non-empty selector results in a score of 0.
-        # We want to match at least one part of an x.y.z, and we don't want to match on empty selectors.
-        return scheme in self.schemes and sublime.score_selector(syntax.scope, self.selector) >= 8
+        # We don't want to match on empty selectors.
+        if len(self.selector.strip()) == 0:
+            return False
+        return scheme in self.schemes and sublime.score_selector(syntax.scope, self.selector)
 
     def map_client_path_to_server_uri(self, path: str) -> str:
         if self.path_maps:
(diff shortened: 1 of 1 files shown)


jfcherng:
Fwiw, in `<4173`:

- sublime.score_selector('a.b', 'b') == 0
- sublime.score_selector('a.b', '')  == 1
- sublime.score_selector('a.b', 'a') == 8

In `=4173`:


- sublime.score_selector('a.b', 'b') == 0
- sublime.score_selector('a.b', '')  == 1
- sublime.score_selector('a.b', 'a') == 1

rchl:
@jfcherng I think you've mislabeled the columns which confused me on how this could cause things to break. The new build returns 2 for `sublime.score_selector("a.b", "a.b")`, not the old one.

plugin/core/types.py
```diff
@@ -840,11 +840,10 @@ def match_view(self, view: sublime.View, scheme: str) -> bool:
...
             return False
-        # Every part of a x.y.z scope seems to contribute 8.
-        # An empty selector result in a score of 1.
-        # A non-matching non-empty selector results in a score of 0.
-        # We want to match at least one part of an x.y.z, and we don't want to match on empty selectors.
-        return scheme in self.schemes and sublime.score_selector(syntax.scope, self.selector) >= 8
+        # We don't want to match on empty selectors.
+        if len(self.selector.strip()) == 0:
+            return False
+        return scheme in self.schemes and sublime.score_selector(syntax.scope, self.selector)
```
rchl, line 846:
```suggestion
        selector = self.selector.strip()
        if not selector:
            return False
        return scheme in self.schemes and sublime.score_selector(syntax.scope, selector) > 0
```

jfcherng:
> @jfcherng I think you've mislabeled the columns which confused me on how this could cause things to break. The new build returns 2 for `sublime.score_selector("a.b", "a.b")`, not the old one.

Oh. I use a wrong column title... Fixed.

rchl (approved):

Issue #2161, opened 2023-01-02
Feature request: Temporarily dismiss LSP


mcclure:
**Is your feature request related to a problem? Please describe.**

I have a tendency to hit ctrl-S to save very frequently and without thinking about it. Sometimes this results in LSP coming up before its help is particularly helpful. Sometimes, the annotations it provides are obtrusive:

![image](https://user-images.githubusercontent.com/277318/210272091-51527832-c9ff-451b-9dce-969205774bb2.png)

If I was looking for a bug, this outline would be quite helpful, so I don't want to turn off these annotations. But sometimes I don't need it **yet**. If I am still working on the function it is just distracting.

**Describe the solution you'd like**

I would like a Sublime Text "command" I could map to a key shortcut, which temporarily dismisses everything LSP, as if I had not yet saved. Remove all annotations, and hide the diagnostics panel. However I don't want this dismissal to be modal, IE, I don't want to have to turn it back on again. I just want it to remove itself until I next hit save and it brings itself back up again.

**Describe alternatives you've considered**

LSP was initially my alternative :) I used to use rust-enhanced which has even more obtrusive annotations.

I guess one possible alternative would be a keystroke to hide LSP annotations altogether, a feature I think already exists. But if I did this I am afraid (ok, I am certain) I would forget to turn them back on again and then they would not be there when I needed them.

rchl:
> I have a tendency to hit ctrl-S to save very frequently and without thinking about it. Sometimes this results in LSP coming up before its help is particularly helpful.

But diagnostics typically come up on typing so I'm not sure why is there an implication here that this is a "on save"-only case.

In any case, your suggested solution sounds very corner-case and manual and not that universally useful to be honest. I think you could look at suggestion in #2117 which might also solve your use case if implemented.

rwols:
I think you could achieve with a plugin in your User/ folder that could, for example, toggle the `show_diagnostics_severity_level` setting in the User/LSP.sublime-settings file between `0` and `4`.

rchl:
Though that would result in restarting the server which can be slow for Rust.

mcclure:
> But diagnostics typically come up on typing so I'm not sure why is there an implication here that this is a "on save"-only problem.

I only started using LSP today. So this sounds like the problem is actually worse than I thought? D: I assumed diagnostics were shown on save because of the `show_diagnostics_panel_on_save` setting.
 
[...]

PR #2429, opened 2024-03-05
feat: toggle show diagnostics without server restart

ta946:
referencing https://github.com/sublimelsp/LSP/issues/2161#issuecomment-1935350667

Introduce the ability to toggle the display of diagnostics without restarting the server

LSP-pyright's output can be quite distracting when trying to read the code to understand it before making changes.

Current method to hide the output, setting `show_diagnostics_severity_level` to `0`, causes the server to restart which can take quite a while with lsp-pyright. Which can be very frustrating and slows down development.

Even worse is that there is no indication that the server is still starting up, so i have to wait longer to make sure there are no errors as opposed to the server still starting up

How it works is by caching the latest diagnostic information. When toggled off, it removes the regions, and when toggled on again, it re-adds the regions by mimicking that new diagnostic information was received

This PR is more a low-quality proof of concept as I don't know the LSP code-base enough nor whats required to contribute correctly

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -250,4 +250,10 @@
         "args": {"async": true},
         "context": [{"key": "lsp.session_with_capability", "operand": "textDocumentSync.willSave | textDocumentSync.willSaveWaitUntil | codeActionProvider.codeActionKinds | documentFormattingProvider | documentRangeFormattingProvider"}]
     },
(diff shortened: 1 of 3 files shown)


rchl:
Thanks for the work but I think it would make sense to just create a feature request for it rather than a pull request that is not really done properly.

PR #2431, opened 2024-03-09
Keep tab selection and focus when applying WorkspaceEdit

jwortmann:
Prevent that the tab selection changes if a WorkspaceEdit affects files that are not currently open in the window (because `window.open_file()` focuses the new tab).

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1779,8 +1779,8 @@ class Session(TransportCallbacks):
                 self.open_uri_async(uri).then(functools.partial(self._apply_text_edits, edits, view_version, uri))
             )
         return Promise.all(promises).then(
-            lambda _: self._check_select_sheets(selected_sheets)).then(
-            lambda _: self._check_focus_sheet(active_sheet))
+            lambda _: self._set_selected_sheets(selected_sheets)).then(
+            lambda _: self._set_focused_sheet(active_sheet))
 
     def _apply_text_edits(
         self, edits: List[TextEdit], view_version: Optional[int], uri: str, view: Optional[sublime.View]
@@ -1790,11 +1790,11 @@ class Session(TransportCallbacks):
             return
         apply_text_edits(view, edits, required_view_version=view_version)
 
-    def _check_select_sheets(self, sheets: List[sublime.Sheet]) -> None:
+    def _set_selected_sheets(self, sheets: List[sublime.Sheet]) -> None:
         if len(sheets) > 1 and len(self.window.selected_sheets()) != len(sheets):
             self.window.select_sheets(sheets)
 
-    def _check_focus_sheet(self, sheet: Optional[sublime.Sheet]) -> None:
+    def _set_focused_sheet(self, sheet: Optional[sublime.Sheet]) -> None:
         if sheet and sheet != self.window.active_sheet():
             self.window.focus_sheet(sheet)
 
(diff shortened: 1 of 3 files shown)


plugin/core/sessions.py
```diff
@@ -1771,12 +1771,16 @@ def apply_workspace_edit_async(self, edit: WorkspaceEdit) -> Promise[None]:
...
+        selected_sheets = self.window.selected_sheets()
         promises = []  # type: List[Promise[None]]
         for uri, (edits, view_version) in changes.items():
             promises.append(
                 self.open_uri_async(uri).then(functools.partial(self._apply_text_edits, edits, view_version, uri))
             )
-        return Promise.all(promises).then(lambda _: None)
+        return Promise.all(promises).then(
+            lambda _: self._check_select_sheets(selected_sheets)).then(
+            lambda _: self._check_focus_sheet(active_sheet))
```
rchl, line 1783: Would prefer this style:

```suggestion
        return Promise.all(promises)
            .then(lambda _: self._check_select_sheets(selected_sheets))
            .then(lambda _: self._check_focus_sheet(active_sheet))
```

plugin/core/sessions.py
```diff
@@ -1786,6 +1790,14 @@ def _apply_text_edits(
             return
         apply_text_edits(view, edits, required_view_version=view_version)
 
+    def _check_select_sheets(self, sheets: List[sublime.Sheet]) -> None:
+        if len(sheets) > 1 == len(self.window.selected_sheets()):
+            self.window.select_sheets(sheets)
+
+    def _check_focus_sheet(self, sheet: Optional[sublime.Sheet]) -> None:
+        if sheet and sheet != self.window.active_sheet():
+            self.window.focus_sheet(sheet)
```
rchl, line 1799: Maybe name `_set_selected_sheets` and `_set_focused_sheet`. They do some checks too but is it important to convey? I think not.

plugin/core/sessions.py
```diff
@@ -1786,6 +1790,14 @@ def _apply_text_edits(
             return
         apply_text_edits(view, edits, required_view_version=view_version)
 
+    def _check_select_sheets(self, sheets: List[sublime.Sheet]) -> None:
+        if len(sheets) > 1 == len(self.window.selected_sheets()):
```
rchl, line 1794: oh man. I don't know what this syntax does exactly and don't know how to google it :)
It's probably just me and maybe we've already discussed it before but still.

plugin/core/sessions.py
jwortmann: I initially wanted to write it like that, but this is not valid Python code.

It would need to have explicit line continuations `\` on the two lines, but then it looks even more ugly as it is now (IMO).

plugin/core/sessions.py
jwortmann: Haha. Okay I'll reconsider tomorrow.

predragnikolic: https://github.com/sublimelsp/LSP/pull/2311#discussion_r1299435330 :⁠-⁠)

rchl: thanks. Wasn't sure if it was the same case :)

plugin/core/sessions.py
rchl: We do use that syntax with `\` and I think it's more readable since then you see each chain start on the new line.
But I won't fight over it.

jwortmann: There are a few places in the code where the style from this PR is used, but the explicit line continuation is more common. I'll unify the styles used.

PR #2435, opened 2024-03-21
Draft/py38 ignore this PR

predragnikolic:
I opened this PR just to see if I finally pinned the issue why the CI fails on Mac


predragnikolic:
Great, 
I can finally say that the test on MacOS because of this line `view.settings().set("lsp_uri", "file:///foo/bar.txt")`

Why? I do not know... but at least we have a starting point...

PR #2436, opened 2024-03-24
Fix failing python 3.8 unittests on MacOS

deathaxe:
This commit adopts `unittesting.ViewTestCase` to run view related test cases.

ViewTestCase ...

1. provides `view` and `window` members pointing to a dedicated view, being created for testing.
2. ensures not to close empty windows, which was probably the most likely reason for unittests failing before on MacOS.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,8 +24,6 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
-      - if: ${{ matrix.os == 'macOS-latest' }}
-        run: brew unlink openssl
       - uses: SublimeText/UnitTesting/actions/setup@v1
       - uses: SublimeText/UnitTesting/actions/run-tests@v1
         with:
--- a/tests/test_configurations.py
+++ b/tests/test_configurations.py
@@ -1,12 +1,15 @@
+import sublime
+
+from unittest import TestCase
+from unittest.mock import MagicMock
+from unittesting import ViewTestCase
+
 from LSP.plugin.core.configurations import WindowConfigManager
 from test_mocks import DISABLED_CONFIG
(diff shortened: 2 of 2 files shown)


deathaxe:
Note: Various test cases don't seem to properly create dedicated windows for testing.

I therefore managed to destroy my open workspace due to running LSP unittests locally.

tests/test_configurations.py
```diff
@@ -24,35 +27,29 @@ def test_override_config(self):
...
+        self.assertIsNotNone(self.view)
+        self.assertIsNotNone(self.window)
+        self.view.settings().set("lsp_uri", "file:///foo/bar.txt")
+        self.view.syntax = MagicMock(return_value=sublime.Syntax(
             path="Packages/Text/Plain text.tmLanguage",
             name="Plain Text",
             scope="text.plain",
             hidden=False
         ))
-        view.settings().set("lsp_uri", "file:///foo/bar.txt")
```
predragnikolic, line 48: any particular reason why `view.settings().set("lsp_uri", "file:///foo/bar.txt")` was moved before `manager = WindowConfigManager(self.window, {TEST_CONFIG.name: TEST_CONFIG})` ?

.github/workflows/main.yml
```diff
@@ -24,8 +24,6 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
-      - if: ${{ matrix.os == 'macOS-latest' }}
-        run: brew unlink openssl
```
predragnikolic, line 28: thanks for removing this as well

predragnikolic (approved):
Thanks,

I left this comment here https://github.com/sublimelsp/LSP/pull/2436/files#r1536919851
If we conclude that the order needs to be as it was, I will change it my PR.

PR #2438, opened 2024-03-26
Ensure didChange notification is never sent after didClose

jwortmann:
This should fix for example the Pyright warning
> LSP-pyright: Received change text document command for closed file <URI>

when a file is saved and closed immediately after changes were applied. This bug wasn't noticed yet, because there is no automatic saving and closing of files implemented yet, and it's unlikely to happen within the debounce time if done manually.

I cherry picked from #2433 because it can be its own PR and to check whether this is the cause for a dubious test failure on Linux.


jwortmann:
The described bug can be observed on the main branch using a key binding like
```jsonc
    {
        "keys": ["f1"],
        "command": "chain",
        "args": {
            "commands": [
                ["insert", {"characters": "TEST"}],
                ["save", {"async": false}],
                ["close", {}]
            ]
        }
    },
```

It doesn't seem to happen always, but most of the time. And keep another file open so the server doesn't shutdown afterwards.

---

I don't know why the test on Linux fails; the change from a previous `textDocument/didChange` seems to be included again in the next notification... seems to be another bug either here or from ST on Linux.

plugin/session_buffer.py
```diff
@@ -308,15 +309,15 @@ def on_revert_async(self, view: sublime.View) -> None:
 
     on_reload_async = on_revert_async
 
-    def purge_changes_async(self, view: sublime.View) -> None:
+    def purge_changes_async(self, view: sublime.View, suppress_requests: bool = False) -> None:
```
predragnikolic, line 312: can we remove `suppress_requests`? 
As it is not used currently (it is always `False`).

jwortmann: Sorry, yes I already noticed that I missed to use this argument in the function call, but I forgot to push the corrected version. The intention of this argument is that we don't need to send all the requests to the server when we know that the view is closing.

predragnikolic:
I am on Linux and I can't reproduce the error in pyright with the above keybinding.

When I add the keybinding,
and trigger it,
didChange is always send before didClose,
regardless if I brute force the keybinding or not...
so I do not see the above error in pyright in the ST console.

I'm also not sure if this PR is the correct way to address the issue.

It would make sense to asset and make sure that didChange is always send before didClose
instead of suppressing the request?

jwortmann:
I can reproduce it quite easily on Windows and the log looks like this:
```
:: [22:05:27.897]  -> LSP-pyright textDocument/didClose: {'textDocument': {'uri': 'file:///D:/code/Python/LspTest/main.py'}}
:: [22:05:27.899] <-  LSP-pyright textDocument/publishDiagnostics: {'uri': 'file:///d%3A/code/Python/LspTest/main.py', 'diagnostics': []}
:: [22:05:28.184]  -> LSP-pyright textDocument/didChange: {'contentChanges': [{'text': 'TEST', 'range': {'start': {'line': 0, 'character': 0}, 'end': {'line': 0, 'character': 0}}, 'rangeLength': 0}], 'textDocument': {'version': 1, 'uri': 'file:///D:/code/Python/LspTest/main.py'}}
LSP-pyright: Received change text document command for closed file file:///d%3A/code/Python/LspTest/main.py
:: [22:05:28.199] <-  LSP-pyright window/logMessage: {'message': 'Received change text document command for closed file file:///d%3A/code/Python/LspTest/main.py', 'type': 1}
```

It also makes sense to me that this bug can be observed, because `purge_changes_async` is called with a 300 ms delay after buffer modifications
https://github.com/sublimelsp/LSP/blob/550d5910a14e7a2943d4aacd1f58f6049f3917d6/plugin/session_buffer.py#L300

but there is no such debouncing for `on_close`.

The request suppression is not directly related to this fix, it is just a further optimization. I just didn't want to create another extra PR for that. The fix is that now `purge_changes_async` is also called within `_check_did_close` (and before `self.session.send_notification(did_close(...))`).

jwortmann:
According to failing CI in #2439, `didChange` is not sent as the first notification, even on Ubuntu 🤷 

https://github.com/sublimelsp/LSP/actions/runs/8493731798/job/23268129591?pr=2439

```
ERROR: test_did_change_before_did_close (test_single_document.AnotherDocumentTestCase)
  File "/home/runner/.config/sublime-text/Packages/UnitTesting/unittesting/core/py33/case.py", line 39, in _executeTestPart
    yield from deferred
  File "/home/runner/.config/sublime-text/Packages/LSP/tests/test_single_document.py", line 418, in test_did_change_before_did_close
    yield from self.await_message('textDocument/didChange')
  File "/home/runner/.config/sublime-text/Packages/LSP/tests/setup.py", line 184, in await_message
    yield from cls.await_promise(promise)
  File "/home/runner/.config/sublime-text/Packages/LSP/tests/setup.py", line 207, in await_promise
    yield {"condition": yielder, "timeout": TIMEOUT_TIME}
```

predragnikolic:
Ok i think the CI tests are flaky.
Locally I have 0fc0e1e67817ea366acaac89942ccdc39c2a0a7e checkout, that is the commit before [Exempt Linux](https://github.com/sublimelsp/LSP/pull/2438/commits/4dd2e91ee9f9bc6bf8b93082b5be4df7243b0c75)

And all the test pass on Linux locally.
I will try to dig into it now.

plugin/session_buffer.py
```diff
@@ -329,23 +330,28 @@ def purge_changes_async(self, view: sublime.View) -> None:
         finally:
             self._pending_changes = None
         self.session.notify_plugin_on_session_buffer_change(self)
-        sublime.set_timeout_async(lambda: self._on_after_change_async(view, version))
+        sublime.set_timeout_async(lambda: self._on_after_change_async(view, version, suppress_requests))
```
predragnikolic, line 333: instead of passing suppress_requests,
can we just do an early return one line above:
```py
if suppress_requests:
   return
sublime.set_timeout_async(lambda: self._on_after_change_async(view, version))
```

I am wondering if we can avoid the `suppress_requests` entirely, 
by relying on the `self.opened` property?

```diff
        if self.opened and self.should_notify_did_close():
+           self.opened = False
+           self.purge_changes_async(view)
-           self.purge_changes_async(view, suppress_requests=True)
            self.session.send_notification(did_close(uri=self._last_known_uri))
-           self.opened = False
```

```diff
    def _on_after_change_async(self, view: sublime.View, version: int, suppress_requests: bool = False) -> None:
        if self._is_saving:
            self._has_changed_during_save = True
            return
       if not self.opened:
            return
```

predragnikolic: if we can't rely on self.opened, 
can we than use view.is_valid() to not send requests for a closed view?

predragnikolic: if I understood correctly suppress_requests is only True when we send the didClose notification
to not send requests for a view that is closing(or closed, didn't test to be 100% sure about this statement)

So it kind of make sense to rely on self.opened or view.is_valid().

jwortmann: 1. `view.is_valid()`:

    I think this is not possible to use. I use this test file to simulate the EventListener callbacks from ST that LSP uses:
    ```python
    import sublime
    import sublime_plugin
    
    
    class TestTextChangeListener(sublime_plugin.TextChangeListener):
    
        def on_text_changed(self, changes):
            print('on_text_changed')
    
    
    class TestViewEventListener(sublime_plugin.ViewEventListener):
    
        def on_post_save_async(self):
            print('on_post_save_async')
    
        def on_close(self):
            print('on_close', self.view.is_valid())
            sublime.set_timeout_async(lambda: print('async after close', self.view.is_valid()))
    ```
    When I close a view, I get this result:
    ```
    on_close True
    async after close True
    ```
    The view is still reported as valid here, and therefore we can't use it as a check whether to suppress the requests. In the real code, it will go through various other functions from DocumentSyncListener and SessionView before calling `purge_changes_async`. If I see it right, this is the line which puts in on the async thread: https://github.com/sublimelsp/LSP/blob/550d5910a14e7a2943d4aacd1f58f6049f3917d6/plugin/documents.py#L1016
    So it could mean that the result of `view.is_valid()` might even depend on how fast all the functions run and how busy the async thread is.

2. `self.opened`:
  
    I think this should work with your suggested diff. But it's a bit subjective which solution is the better one. If it checks the `self.opened` property, then the code implicitly relies on it being set to `False` before the function is called. If we forget about that and in a few years change the order of the lines again for some reasons, it might not work correctly anymore. So I think passing the explicit `suppress_requests` parameter would be the better alternative and makes the code easier to understand.

predragnikolic: The approach with `suppress_requests` is the most explicit.
 suppress_requests is effectively used just in `_on_after_change_async`
but we also need to pollute the purge_changes_async with an additional argument
`def purge_changes_async(self, view: sublime.View, suppress_requests: bool = False) -> None:`.

The first time I sat the suppress_requests in purge_changes_async,
it wasn't that clear what requests will be suppressed.
I thought that it would prevent sending the didChange **notifications**... but I was wrong.


I would be fine with `suppress_requests` or with `self.opened`... So I will resolve this thread.

tests/test_single_document.py
```diff
@@ -371,6 +346,54 @@ def test_progress(self) -> 'Generator':
...
+        assert self.view
+        self.view.window().run_command("chain", {
+            "commands": [
+                ["insert", {"characters": "TEST"}],
+                ["save", {"async": False}],
+                ["close", {}]
+            ]
+        })
+        yield from self.await_message('textDocument/didChange')
+        # yield from self.await_message('textDocument/didSave')  # TODO why is this not sent?
```
predragnikolic, line 393: now just to see why this happens
and than we can merge this

predragnikolic: test pass locally, but not on the CI
I am not sure what to do here, except to ignore this
and to report the issues to UnitTesting package.

predragnikolic (approved):
I will wait a bit, to see if others will take a look, before I merge this.

tests/test_single_document.py
jwortmann: I can also see it happen locally, when I manually use the key binding in a file. Only `didChange` and `didClose` are sent, but not `didSave`. But this is another bug and not directly related to this PR, so we can investigate another time.

predragnikolic: When I uncomment the line.
and test the current test file, 
the tests pass on Linux and Mac locally:
```
test_apply_formatting (test_single_document.SingleDocumentTestCase) ... ok
test_apply_formatting_and_preserve_order (test_single_document.SingleDocumentTestCase) ... ok
test_declaration (test_single_document.SingleDocumentTestCase) ... ok
test_declaration_location_link (test_single_document.SingleDocumentTestCase) ... ok
test_definition (test_single_document.SingleDocumentTestCase) ... ok
test_definition_location_link (test_single_document.SingleDocumentTestCase) ... ok
test_did_close (test_single_document.SingleDocumentTestCase) ... ok
test_did_open (test_single_document.SingleDocumentTestCase) ... ok
test_expand_selection (test_single_document.SingleDocumentTestCase) ... ok
test_formats_on_save (test_single_document.SingleDocumentTestCase) ... ok
test_hover_info (test_single_document.SingleDocumentTestCase) ... ok
test_implementation (test_single_document.SingleDocumentTestCase) ... ok
test_implementation_location_link (test_single_document.SingleDocumentTestCase) ... ok
test_out_of_bounds_column_for_text_document_edit (test_single_document.SingleDocumentTestCase) ... ok
test_progress (test_single_document.SingleDocumentTestCase) ... ok
test_remove_line_and_then_insert_at_that_line_at_end (test_single_document.SingleDocumentTestCase) ... ok
test_rename (test_single_document.SingleDocumentTestCase) ... ok
test_run_command (test_single_document.SingleDocumentTestCase) ... ok
test_sends_save_with_purge (test_single_document.SingleDocumentTestCase) ... ok
test_tabs_are_respected_even_when_translate_tabs_to_spaces_is_set_to_true (test_single_document.SingleDocumentTestCase) ... ok
test_type_definition (test_single_document.SingleDocumentTestCase) ... ok
test_type_definition_location_link (test_single_document.SingleDocumentTestCase) ... ok
test_did_change (test_single_document.SingleDocumentTestCase2) ... ok
test_did_change_before_did_close (test_single_document.SingleDocumentTestCase3) ... ok
test_will_save_wait_until (test_single_document.WillSaveWaitUntilTestCase) ... ok

Ran 25 tests in 6.032s

OK
```

predragnikolic:
So I decided to test this once more.

On Linux when I use the following keybinding:
```
    {
        "keys": ["f1"],
        "command": "chain",
        "args": {
            "commands": [
                ["insert", {"characters": "TEST"}],
                ["save", {"async": false}],
                ["close", {}]
            ]
        }
    },
```
The order notifications is the following:  `didClose` is send first, followed by two `workspace/didChangeWatchedFiles` notifications.
Which I find unexpected.

I expected `didChange` -> `didSave` -> `didClose`.


```
:: [11:13:39.568]  -> LSP-pyright textDocument/didClose: {'textDocument': {'uri': 'file:///home/predragnikolic/.config/sublime-text/Packages/LSP/plugin/color.py'}}
:: [11:13:39.946]  -> LSP-pyright workspace/didChangeWatchedFiles: {'changes': [{'type': 2, 'uri': 'file:///home/predragnikolic/.config/sublime-text/Packages/LSP/plugin/color.py'}]}
:: [11:13:39.947]  -> LSP-pyright workspace/didChangeWatchedFiles: {'changes': [{'type': 2, 'uri': 'file:///home/predragnikolic/.config/sublime-text/Packages/LSP/plugin/color.py'}]}
```

I also tested with other servers.
```
:: [11:20:46.726] --> LSP-typescript textDocument/documentHighlight (14): {'textDocument': {'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/appAdmin/src/i18n/i18n.ts'}, 'position': {'line': 6, 'character': 0}}
:: [11:20:46.727] --> LSP-tailwindcss textDocument/codeAction (8): {'range': {'start': {'line': 6, 'character': 0}, 'end': {'line': 6, 'character': 0}}, 'textDocument': {'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/appAdmin/src/i18n/i18n.ts'}, 'context': {'triggerKind': 2, 'diagnostics': []}}
:: [11:20:46.727] --> LSP-eslint textDocument/codeAction (8): {'range': {'start': {'line': 6, 'character': 0}, 'end': {'line': 6, 'character': 0}}, 'textDocument': {'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/appAdmin/src/i18n/i18n.ts'}, 'context': {'triggerKind': 2, 'diagnostics': []}}
:: [11:20:46.728] --> LSP-typescript textDocument/codeAction (15): {'range': {'start': {'line': 6, 'character': 0}, 'end': {'line': 6, 'character': 0}}, 'textDocument': {'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/appAdmin/src/i18n/i18n.ts'}, 'context': {'triggerKind': 2, 'diagnostics': []}}
:: [11:20:46.730] <<< LSP-typescript (14) (duration: 4ms): []
:: [11:20:46.730] <<< LSP-eslint (8) (duration: 2ms): []
:: [11:20:46.730] <<< LSP-tailwindcss (8) (duration: 3ms): []
:: [11:20:46.732] <<< LSP-typescript (15) (duration: 4ms): []
:: [11:20:47.301]  -> LSP-tailwindcss textDocument/didClose: {'textDocument': {'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/appAdmin/src/i18n/i18n.ts'}}
:: [11:20:47.302]  -> LSP-eslint textDocument/didClose: {'textDocument': {'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/appAdmin/src/i18n/i18n.ts'}}
:: [11:20:47.303]  -> LSP-typescript textDocument/didClose: {'textDocument': {'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/appAdmin/src/i18n/i18n.ts'}}
:: [11:20:47.303] <-  LSP-eslint textDocument/publishDiagnostics: {'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/appAdmin/src/i18n/i18n.ts', 'diagnostics': []}
:: [11:20:47.306] <-  LSP-typescript textDocument/publishDiagnostics: {'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/appAdmin/src/i18n/i18n.ts', 'diagnostics': []}
:: [11:20:47.557] <-  LSP-typescript textDocument/publishDiagnostics: {'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/appAdmin/src/i18n/i18next.d.ts', 'diagnostics': []}
:: [11:20:47.675]  -> LSP-tailwindcss workspace/didChangeWatchedFiles: {'changes': [{'type': 2, 'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/appAdmin/src/i18n/i18n.ts'}]}
:: [11:20:47.676]  -> LSP-tailwindcss workspace/didChangeWatchedFiles: {'changes': [{'type': 2, 'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/appAdmin/src/i18n/i18n.ts'}]}
```

Note this is Linux. Will test on a Mac tonight.


I also get the same order of notifications in Windows:
 `didClose` is send first, followed by two `workspace/didChangeWatchedFiles` notifications.

jwortmann:
Are you sure that you are on the correct branch?
It works for me (excluding the missing `didSave`, but this is a different matter) on Windows and it also seems to work on CI. I have tested with different servers too.

predragnikolic:
Yes, can confirm that I am on the correct branch `fix-notification-order`

predragnikolic:
I tried to test a few more times, 
and now the execution order is as expected.
```
:: [19:52:41.928]  -> LSP-typescript textDocument/didChange: {'textDocument': {'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/server/src/index.ts', 'version': 2}, 'contentChanges': [{'rangeLength': 0, 'range': {'end': {'character': 0, 'line': 6}, 'start': {'character': 0, 'line': 6}}, 'text': 'TEST'}, {'rangeLength': 0, 'range': {'end': {'character': 4, 'line': 6}, 'start': {'character': 4, 'line': 6}}, 'text': '\n'}]}
:: [19:52:41.928]  -> LSP-typescript textDocument/didSave: {'textDocument': {'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/server/src/index.ts'}}
:: [19:52:41.929]  -> LSP-typescript textDocument/didClose: {'textDocument': {'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/server/src/index.ts'}}
```

So I dig into it.

If I have two files hello.py and hello2.py
1. if hello.py is open 
2. and with `ctrl+p` open hello2.py 
3. if I trigger the `chain` keybinding
4. I is a high changes to get the following logs:
```
:: [20:08:14.402]  -> LSP-pyright textDocument/didOpen: {'textDocument': {'text': '', 'version': 0, 'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/hello2.py', 'languageId': 'python'}}
:: [20:08:14.423] <-  LSP-pyright textDocument/publishDiagnostics: {'version': 0, 'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/hello2.py', 'diagnostics': []}
:: [20:08:16.905]  -> LSP-pyright textDocument/didClose: {'textDocument': {'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/hello2.py'}}
:: [20:08:16.906] <-  LSP-pyright textDocument/publishDiagnostics: {'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/hello2.py', 'diagnostics': []}
:: [20:08:17.285]  -> LSP-pyright workspace/didChangeWatchedFiles: {'changes': [{'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/hello2.py', 'type': 2}]}
:: [20:08:17.285]  -> LSP-pyright workspace/didChangeWatchedFiles: {'changes': [{'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/hello2.py', 'type': 2}]}
```
...but not always.

If i focus the `hello2.py` file or leave it open for longer, (1s, 2s),
I will see the expected logs, always.
```
:: [20:09:41.193] <<< LSP-pyright (9) (duration: 1ms): None
:: [20:09:41.193] <<< LSP-pyright (10) (duration: 0ms): []
:: [20:09:42.269]  -> LSP-pyright textDocument/didChange: {'contentChanges': [{'text': 'TEST', 'range': {'start': {'line': 1, 'character': 0}, 'end': {'line': 1, 'character': 0}}, 'rangeLength': 0}, {'text': '\n', 'range': {'start': {'line': 1, 'character': 4}, 'end': {'line': 1, 'character': 4}}, 'rangeLength': 0}], 'textDocument': {'version': 2, 'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/hello2.py'}}
:: [20:09:42.269]  -> LSP-pyright textDocument/didSave: {'textDocument': {'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/hello2.py'}}
:: [20:09:42.270]  -> LSP-pyright textDocument/didClose: {'textDocument': {'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/hello2.py'}}
:: [20:09:42.272] <-  LSP-pyright textDocument/publishDiagnostics: {'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/hello2.py', 'diagnostics': []}
:: [20:09:42.657]  -> LSP-pyright workspace/didChangeWatchedFiles: {'changes': [{'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/hello2.py', 'type': 2}]}
:: [20:09:42.658]  -> LSP-pyright workspace/didChangeWatchedFiles: {'changes': [{'uri': 'file:///home/predragnikolic/Documents/sandbox/vite-app/hello2.py', 'type': 2}]}

``` 

Here is a gif, that can demonstrate this.
output

jwortmann:
Interesting... but I have no explanation for that, nor can I reproduce it on my PC :D 

I've tried the same a few times (open file with <kbd>Ctrl</kbd>+<kbd>P</kbd>, then immediately press the key binding), but I always get the same result:
```
:: [20:31:10.078]  -> LSP-pyright textDocument/didOpen: {'textDocument': {'version': 0, 'text': '\n', 'uri': 'file:///D:/code/Python/LspTest/main.py', 'languageId': 'python'}}
:: [20:31:10.112] <-  LSP-pyright textDocument/publishDiagnostics: {'diagnostics': [], 'uri': 'file:///d%3A/code/Python/LspTest/main.py', 'version': 0}
:: [20:31:10.403]  -> LSP-pyright textDocument/didChange: {'contentChanges': [{'range': {'start': {'character': 0, 'line': 0}, 'end': {'character': 0, 'line': 0}}, 'rangeLength': 0, 'text': 'TEST'}], 'textDocument': {'uri': 'file:///D:/code/Python/LspTest/main.py', 'version': 1}}
:: [20:31:10.403]  -> LSP-pyright textDocument/didClose: {'textDocument': {'uri': 'file:///D:/code/Python/LspTest/main.py'}}
:: [20:31:10.405] <-  LSP-pyright textDocument/publishDiagnostics: {'uri': 'file:///d%3A/code/Python/LspTest/main.py', 'diagnostics': []}
```

Another difference is that I never see the `didSave` notification, while it seems to work for you when you wait a bit longer before using the key binding. I've tried also with the chokidar file watcher enabled or disabled, but that makes no difference here.

Actually it is not good that it doesn't always work correctly when the open and closing happen fast, because this was exactly the use case with the autosaving & close when applying workspace edits, from the other PR. But I can't see where the code could possibly go wrong.

predragnikolic:
I would suggest merging changes from this PR to `jwortmann:workspace-edit-preserve-tabs` https://github.com/sublimelsp/LSP/pull/2433 
instead of `main`. 

To keep things going,
I will continue investigating this strange behavior from this PR
while also testing #2433. 

Maybe we will find out why it happens.

plugin/session_buffer.py
```diff
@@ -329,23 +330,28 @@ def purge_changes_async(self, view: sublime.View) -> None:
...
+            if self.session.config.diagnostics_mode == "workspace" and \
+                    not self.session.workspace_diagnostics_pending_response and \
+                    self.session.has_capability('diagnosticProvider.workspaceDiagnostics'):
+                self._workspace_diagnostics_debouncer_async.debounce(
+                    self.session.do_workspace_diagnostics_async, timeout_ms=WORKSPACE_DIAGNOSTICS_TIMEOUT)
+            self.do_semantic_tokens_async(view)
+            if userprefs().link_highlight_style in ("underline", "none"):
+                self._do_document_link_async(view, version)
+            self.do_inlay_hints_async(view)
+        except MissingUriError:
```
predragnikolic, line 353: Before I forget again, 
was this an intentional addition?
```
except MissingUriError:
  pass
```

jwortmann: Yes, I've added the `try` block for the requests because this function is put on the async thread via `sublime.set_timeout_async` at the end of `purge_changes_async`. So it could theoretically happen that the view has closed in the meantime, and then it would raise the MissingUriError. In that case nothing else needs to happen; we could also use `return` here but I used `pass` because this is at the end of the function anyway.

jwortmann:
I merged it into the other PR, but I would prefer to merge this one seperately first. I don't know if we can find the cause for the strange and different behaviors and remaining bugs, but even if not, I think this should still be an improvement to the current behavior on the `main` branch.

And thanks for the debugging help 😄

PR #2444, opened 2024-04-13
Print messages from window/logMessage into console instead of LSP log panel

jwortmann:
I think it's not very useful to print those messages into the LSP log panel, in addition to the payload log. Those lines are just redundant there, for example
```
LSP-pyright: Received change text document command for closed file file:///d%3A/code/Python/LspTest/main.py
:: [22:05:28.199] <-  LSP-pyright window/logMessage: {'message': 'Received change text document command for closed file file:///d%3A/code/Python/LspTest/main.py', 'type': 1}
```

Instead, if `"log_debug"` is enabled, I suggest to log them into the console, similar to the other log/debug messages like
```
LSP: enabled configs: [...]
LSP: starting [...]
LSP: LSP-pyright: supported code action kinds: [...]
```
Error messages are additionally printed to the status bar (only if `"log_debug"` setting is enabled).

### DIFF
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -216,6 +216,8 @@ class MessageType(IntEnum):
     """ An information message. """
     Log = 4
     """ A log message. """
+    Debug = 5
+    """ A debug message. """
 
 
 class TextDocumentSyncKind(IntEnum):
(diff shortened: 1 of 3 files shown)


predragnikolic (approved):
:+1:

PR #2446, opened 2024-04-13
Import from typing, typing_extensions and enum modules on Python 3.8

jwortmann:
On Python 3.8 `enum` and `typing` are part of the standard library.

`.core.typing` still contains all the fake typing classes, because many LSP-* plugins import from `LSP.plugin.core.typing`.


plugin/core/input_handlers.py
```diff
@@ -16,7 +17,7 @@
 P = ParamSpec('P')
 
 
-def debounced(user_function: Callable[P, Any]) -> Callable[P, None]:
+def debounced(user_function: 'Callable[P, Any]') -> 'Callable[P, None]':
```
jwortmann, line 20: Can't use the `Callable[P, returntype]` syntax anymore, because it was introduced in Python 3.10 ([ParamSpec](https://typing.readthedocs.io/en/latest/spec/generics.html#paramspec)) and the `Callable` which was used here was just a "fake" class.

jfcherng:
Fwiw, with `from __future__ import annotations`, type annotations are just ignored at runtime. People can use something like `tuple[str, ...] | dict[str, Any] | None` for type annotations in py38. Most of `from typing import ...` can be removed (later).

jwortmann:
I would say that we can make those changes step by step in separate PRs. So the next steps could be
- use modern type union syntax
  - replace `Union[Foo, Bar]` with `Foo | Bar`
  - replace `Optional[Foo]` with `Foo | None`
- replace generic typing classes like `Dict`, `List`, `Tuple` with builtins `dict`, `list`, `tuple`

The pipe operator was introduced in Python 3.10 and the syntax with `dict` etc. in Python 3.9, so we would need to have `from __future__ import annotations` in all files.

And this is probably significantly more work if done manually throughout the whole code base, so we should check how well it would work to do it automatically.

jwortmann:
By the way, with this PR there are still 3 classes required to be imported from the local `.core.typings`, because they were introduced after Python 3.8: TypeGuard, NotRequired, and ParamSpec.
I've noticed that `typing_extensions` is on Package Control (https://github.com/packagecontrol/channel/blob/6faa8ffb43475f829ee95936955a962cce47d6d3/repository.json#L1387), so I wonder whether we could use that to avoid those 3 remaining imports.

predragnikolic:
Nice! 

I would rather not rush into merging this in feat/py38, 
I hope that you are OK if this PR is merged sometime after the 19th April? :)

jwortmann:
Okay, I can change the target branch then and rebase.

---

Regarding typing_extensions, I can see that it was merged two weeks ago in https://github.com/packagecontrol/channel/commit/cd8ec019e2457eec41f0d2a477c7392c973e3918, but it's not (yet?) in the https://packagecontrol.github.io/channel/channel_v4.json file used by Package Control. So maybe we need to wait a bit until it gets updated.

I think with `from __future__ import annotations` we could alternatively also do
```python
if TYPE_CHECKING:
    from typing import NotRequired, TypeGuard
```
but it wouldn't work with `ParamSpec` because that one is used outside of type annotations.

predragnikolic:
Will check the `typing_extensions` , 
if it works, then `NotRequired`, `TypeGuard` could be imported from there without the `if TYPE_CHECKING:` check... I think.

jwortmann:
typing_extensions seems to be fixed upstream now. So the question would be if we want to have it as an additional dependency or not. With it, `NotRequired`, `TypeGuard` and `ParamSpec` can all be imported directly.

I missed to mention above that there is still `StrEnum` (used in `protocol.py) that we need to import from `.core.typings`, but this is regardless whether typing_extensions is available or not.

jfcherng:
I think more or less, `StrEnum` is just

```py
class StrEnum(str, Enum):
    def __str__(self) -> str:
        return self.value()
```

jwortmann:
There is another problem with LSP's fake enums. Currently we can use the syntax `SomeEnum.SomeValue` to get the value, because `SomeEnum` as an derived class from the fake enum is just a regular class, so it returns the attribute value directly. But if we import from the `enum` module, `SomeEnum` now is a real enum and `SomeEnum.SomeValue` would return something like `<SomeEnum.SomeValue: 42>` instead of just `42`. In other words we would need to use `SomeEnum.SomeValue.value` instead throughout the code.

I guess I'll revert using the "real" enums, for now...

jwortmann:
> Maybe you can try [...]

The IntEnum classes are actually not problematic, because comparisons work out of the box for them. But a problem is that some other classes in *protocol.py* inherit from `Enum` with their values being strings. Unfortunately you can't compare via `SomeEnum.SomeValue == 'something'` if `SomeEnum` is just an `Enum`. But it works if it is a `StrEnum`.

And there are some problems in the initialize request, where we access the enum members directly in the code (and not via `.value` suffix). But I'll push a fix for that.

jfcherng:
> > Maybe you can try [...]
> 
> The IntEnum classes are actually not problematic, because comparisons work out of the box for them. But a problem is that some other classes in _protocol.py_ inherit from `Enum` with their values being strings. Unfortunately you can't compare via `SomeEnum.SomeValue == 'something'` if `SomeEnum` is just an `Enum`. But it works if it is a `StrEnum`.

https://github.com/sublimelsp/LSP/pull/2446#issuecomment-2059373856 works in my test.

```py
>>> from enum import Enum


class StrEnum(str, Enum):
    def __str__(self) -> str:
        return self.value.__str__()


class Kind(StrEnum):
    KEYWORD = "1"
    CONSTANT = "2"


>>> Kind.KEYWORD == '1'
True
>>> Kind.KEYWORD < '2'
True
```

jwortmann:
> Would it be problematic if it inherits from this `class StrEnum(str, Enum)` rather than `Enum`?

No, your suggestion should work too. But it's not much different in practice from what we currently have with
```python
class StrEnum:
    pass
```

The problem was that the enums in *protocol.py* are automatically generated from https://github.com/microsoft/lsprotocol and they used `Enum` instead of `StrEnum`. The PR linked somewhere above should workaround that.

jfcherng:
But we still have `Enum`, `IntEnum` or `StrEnum` issues? What are them?

jwortmann:
Problem 1 was that for the initialize request we used for example https://github.com/sublimelsp/LSP/blob/5891ff41deaa781783effeb97c978b3e50bd0a26/plugin/core/sessions.py#L299

which would give us something like `[<InsertTextMode.AdjustIndentation: 1>]` (which is not JSON serializable) if we import IntEnum from the enum module instead of using a fake class.

Problem 2 was that comparisons of enums which have string values would not work anymore in the code, because those enums only used the `Enum` class and not `StrEnum`.

But both should be fixed now in https://github.com/sublimelsp/LSP/pull/2446/commits/8c0114794be41e07a2ecb26a6c889b8c2a0984b9

plugin/core/input_handlers.py
jwortmann: This workaround is not required anymore with the `ParamSpec` imported from typing_extensions.

jwortmann:
@predragnikolic I wonder can you restore the `feat/py38` branch temporarily? Otherwise it seems to be impossible to change the target branch of this PR

predragnikolic:
Will do, I saw what happened.
I didn't expect that.

I thought that it would just switch to the `main` branch.

jwortmann:
Ok I think now the `feat/py38` branch can be deleted.

And I think we discovered a bug in GitHub's UI 😆 

github

predragnikolic:
Well, 
I kind of died a bit when I saw UI "the bug" :D

predragnikolic (approved):

Issue #2044, opened 2022-09-06
The server is restarted when changing workspace settings


predragnikolic:
**Current:**
Currently when changing workspace "settings" key, the server is restarted.

**Expected:**
I expect the the server to not restart when chaining the "settings" key, but rather LSP to send a ‘workspace/didChangeConfiguration` to notify the server about the setting changed.

**Describe the solution you'd like**
LSP should restart the server for the following keys in the client config:
"auto_complete_selector"
"disabled_capabilities"
"enabled"
"env"
"experimental_capabilities"
"file_watcher",
"initializationOptions"
"priority_selector"
"schemes"
"selector"

LSP should NOT restart the server only if the "settings" key is modified.

predragnikolic:
Closing this in favor of #2243

PR #2448, opened 2024-04-15
Don't restart servers when userprefs change

jwortmann:
Currently, all language servers are restarted whenever the user settings in LSP.sublime-settings are saved. This is not really necessary, so with this PR they are only restarted if the `"clients"` setting (containing the server configs) is actually modified.

Note that not all UI features are immediately redrawn when the user settings change. For example if you change "diagnostics_gutter_marker", it still uses the old icons until new diagnostics arrive (usually after next buffer modification). That is because the icon is stored directly in the DiagrosticSeverityData object
https://github.com/sublimelsp/LSP/blob/5891ff41deaa781783effeb97c978b3e50bd0a26/plugin/core/views.py#L68-L78

which was maybe not the optimal design choice, but I didn't want to do huge refactorings here.

Or for example if inlay hints get enabled/disabled, they are only added or cleared after the next buffer modification, but I think that should be okay.


predragnikolic:
I have not tested this,
but based on the code that I see

Previously, 
if someone modified server settings,
because LSP restarts the server on setting change,
the server would pick it up the new changes once started again.

Now in this PR,
LSP will not restart the server, when we change server settings,
thus the server will not pick up new setting changes.

In order to address this, 
the server would now need to be notified with `didChangeConfiguration`. 

----

related to
https://github.com/sublimelsp/LSP/issues/2044

jwortmann:
The server settings are part of the `"clients"` setting, so if they are modified, servers are still restarted.

`didChangeConfiguration` would need to be implemented separately.

This PR is only for the userprefs, i.e. UI settings and other LSP settings. It is not directly related to #2044.

plugin/session_view.py
```diff
@@ -301,22 +302,23 @@ def diagnostics_tag_scope(self, tag: int) -> str | None:
     def present_diagnostics_async(
         self, is_view_visible: bool, data_per_severity: dict[tuple[int, bool], DiagnosticSeverityData]
     ) -> None:
+        self._diagnostics_data_per_severity = data_per_severity
```
rchl, line 305: Wouldn't it be more correct to store this in `SessionBuffer` since this data is the same for each SessionView belonging to the same SessionBuffer?

plugin/core/windows.py
```diff
@@ -522,12 +522,27 @@ class WindowRegistry:
...
+        for wm in self._windows.values():
+            wm.on_diagnostics_updated()
+            for session in wm.get_sessions():
+                session.redraw_config_status_async()
+                for sb in session.session_buffers_async():
+                    sb.redraw_document_links_async()
+                    if not userprefs().semantic_highlighting:
+                        sb.clear_semantic_tokens_async()
+                for sv in session.session_views_async():
+                    sv.redraw_diagnostics_async()
```
rchl, line 544: I wonder if it would be better to delegate all those actions to Session/SessionBuffer/SessionView. We'd trigger `on_settings_changed` on `Session` and it would propagate this to SessionBuffers->SessionViews by calling public `on_settings_changed` on those.

We wouldn't need to expose public methods for each feature separately and I feel like the logic be contained better, without having to delegate responsibility to trigger correct actions to other places.

plugin/core/settings.py
```diff
@@ -15,11 +16,17 @@ class ClientConfigs:
     def __init__(self) -> None:
         self.all: dict[str, ClientConfig] = {}
         self.external: dict[str, ClientConfig] = {}
-        self._listener: Callable[[str | None], None] | None = None
+        self._clients_listener: Callable[[str | None], None] | None = None
+        self._userprefs_listener: Callable[[], None] | None = None
+        self._clients_hash: int | None = None
 
     def _notify_listener(self, config_name: str | None = None) -> None:
```
rchl, line 23: Would be good to rename to `_notify_clients_listener` for consistency.

plugin/core/windows.py
```diff
@@ -522,12 +522,27 @@ class WindowRegistry:
     def __init__(self) -> None:
         self._enabled = False
         self._windows: dict[int, WindowManager] = {}
-        client_configs.set_listener(self._on_client_config_updated)
+        client_configs.set_listeners(self._on_client_config_updated, self._on_userprefs_updated)
```
rchl, line 525: Now that we are passing two function references, it might make sense to create an interface with `on_client_config_updated` and `on_userprefs_updated` that `WindowRegistry` would inherit and we'd just pass `self` here.

(just a suggestion)

plugin/session_buffer.py
```diff
@@ -428,14 +428,20 @@ def _do_document_link_async(self, view: sublime.View, version: int) -> None:
...
-                [range_to_region(link["range"], view) for link in self._document_links],
-                scope="markup.underline.link.lsp",
-                flags=DOCUMENT_LINK_FLAGS)
+            view = self.some_view()
+            if not view:
+                return
+            regions = [range_to_region(link["range"], view) for link in self._document_links]
+            for sv in self.session_views:
+                sv.view.add_regions(
+                    "lsp_document_link", regions, scope="markup.underline.link.lsp", flags=DOCUMENT_LINK_FLAGS)
```
rchl, line 441: Nice fix to region rendering. Ideally would be its own PR to keep records clear but no pressure.

plugin/session_view.py
jwortmann: https://github.com/sublimelsp/LSP/pull/2448/commits/b2bab2f146c06da56f19a80ac574c4c3ac223005

plugin/core/windows.py
jwortmann: https://github.com/sublimelsp/LSP/pull/2448/commits/32d3bef74016697caeeb5d5ff6a905814425203b

I named it `on_userprefs_changed_async`

plugin/core/settings.py
jwortmann: https://github.com/sublimelsp/LSP/pull/2448/commits/3dd85ec876ab1ccf2288cff8a22f5b951a88fc0e

plugin/session_buffer.py
jwortmann: On the other hand if we keep it out of the commit history, maybe we can keep it a secret that this bug existed ;)

rchl:
> Or for example if inlay hints get enabled/disabled, they are only added or cleared after the next buffer modification, but I think that should be okay.

Honestly, I'm not sure whether that's OK. That means that you touch some setting and now all the open views start missing something (semantic highlighting, for example). Now you have to either poke all the views or restart ST which kinda defeats the point of this fix. I think we would have to do the right thing and update every functionality that might be affected.

plugin/core/windows.py
jwortmann: https://github.com/sublimelsp/LSP/pull/2448/commits/bc61186f5f89a681e3b169d4b9d2309aa33120e6

jwortmann:
Actually for inlay hints we were already doing almost the right thing. The `"show_inlay_hints"` setting is different in the way that it is only used as the default value for new windows. But we have a command and a menu item to toggle the inlay hints dynamically per window, so I assume the states for all windows that are already opened should not be affected when the global LSP setting is changed.

I think there was a small inconsistency with the settings description in the LspToggleInlayHintsCommand that it would only set the initial state when it was toggled the first time. But this should be fixed now.

---

In contrast to what I wrote above, I saw that semantic highlighting is currently only removed when the setting is turned off, but not added when turned on. I'll take another look later how it can be redrawn directly.

jwortmann:
Semantic tokens should now be requested/updated if the userprefs change and semantic highlighting is enabled (regardless whether that setting actually changed or not), *as soon as the view gets activated*. I think this is a good compromise between having the rendering state up to date and not overwhelming the server with many expensive requests.

rchl (approved):

PR #2449, opened 2024-04-16
Fix Discord link

predragnikolic:
Reported on Discord.

> FichteFoll — Yesterday at 11:31 AM
FYI, I just noticed that you have a discord link to a channel in the footer of the docs (i.e. https://lsp.sublimetext.io/). This does not work for people not already part of the server. Please use the invite link in the channel topic instead (also mentioned in

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -66,4 +66,4 @@ nav:
 extra:
   social:
     - icon: fontawesome/brands/discord
-      link: https://discordapp.com/channels/280102180189634562/645268178397560865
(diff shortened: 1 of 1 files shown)


jwortmann (approved):

predragnikolic:
> I may suggest to use a "fixed" link like https://lsp-discord.sublimetext.io so that there won't be outdated links spread on the Internet.

I think it is fine to have broken links from time to time, it is not that crucial.
I also have nothing agains having a "fixed" link, but I do not see a huge value in it.

It is easy to fix broken links :)

PR #2450, opened 2024-04-18
Use variable annotations instead of type comments on Python 3.8

jwortmann:
https://peps.python.org/pep-0526/

I used this regex to replace them:

regex

But I did a little bit of fine tuning by hand afterwards, because the regex didn't catch multi-line expressions, and a few type comments were not valid syntax.

I can change the target branch if we want to wait for the Python 3.8 release first.


plugin/core/transports.py
```diff
@@ -281,7 +281,7 @@ def start_subprocess() -> subprocess.Popen:
         config.name, process, sock, reader, writer, stderr, json_rpc_processor, callback_object)  # type: ignore
 
 
-_subprocesses = weakref.WeakSet()  # type: weakref.WeakSet[subprocess.Popen]
+_subprocesses: 'weakref.WeakSet[subprocess.Popen]' = weakref.WeakSet()
```
predragnikolic, line 284: note to myself to check why it was necessary to use `'weakref.WeakSet[subprocess.Popen]'` instead of `weakref.WeakSet[subprocess.Popen]`

plugin/documents.py
```diff
@@ -91,7 +91,7 @@ def previous_non_whitespace_char(view: sublime.View, pt: int) -> str:
 
 class TextChangeListener(sublime_plugin.TextChangeListener):
 
-    ids_to_listeners = WeakValueDictionary()  # type: WeakValueDictionary[int, TextChangeListener]
+    ids_to_listeners: 'WeakValueDictionary[int, TextChangeListener]' = WeakValueDictionary()
```
predragnikolic, line 94: note to myself, check why we need quotes for this type.

predragnikolic (approved):

predragnikolic:
> I can change the target branch if we want to wait for the Python 3.8 release first.

LSP v2.0.0 will be released tomorrow
No need to change the target branch, Github will automatically update it to `main` after `sublimelsp:feat/py38` is merged.

I do plan to merge your py38 PRs at the beginning of next week. Tuesday.

I plan to release python 3.8 tomorrow,
but give it a few days, till Monday (just in case if something unexpected bad happens, to easily revert to pyton 3.3 without reverting much... This option shouldn't happen, but if it happens, I will be ready (: )

After that I plan to continue with merging the new PR's.

plugin/core/transports.py
jwortmann: `WeakSet` and `WeakValueDictionary` are not generic types like classes from typing like for example `List`, so you can't use the notation of having another type inside square brackets on them.

It would work with `from __future__ import annotations`, because then type hints are automatically turned into strings.

predragnikolic: Thanks!

predragnikolic:
Test fail on Windows :)
```
ERROR: test_did_change_before_did_close (test_single_document.SingleDocumentTestCase3)
  File "C:\st\Data\Packages\UnitTesting\unittesting\core\py38\case.py", line 80, in run
    yield from deferred
  File "C:\st\Data\Packages\UnitTesting\unittesting\core\py38\case.py", line 23, in _callTestMethod
    yield from deferred
  File "C:\st\Data\Packages\LSP\tests\test_single_document.py", line 392, in test_did_change_before_did_close
    yield from self.await_message('textDocument/didChange')
  File "C:\st\Data\Packages\LSP\tests\setup.py", line 184, in await_message
    yield from cls.await_promise(promise)
  File "C:\st\Data\Packages\LSP\tests\setup.py", line 207, in await_promise
    yield {"condition": yielder, "timeout": TIMEOUT_TIME}
```

PR #2456, opened 2024-04-20
Fix LSP.plugin.core.typing with py38

jfcherng:
Since we are now in py38, we don't need most of type polyfills.

`# noqa: F401` is to suppress `import but no used` linter warnings.

### DIFF
--- a/plugin/core/typing.py
+++ b/plugin/core/typing.py
@@ -36,11 +36,12 @@ if sys.version_info >= (3, 11):
         TypeGuard,  # noqa: F401
     )
 else:
+    _T = TypeVar("_T")
 
     class StrEnum(Type):  # type: ignore
         pass
 
-    class NotRequired(Type):  # type: ignore
+    class NotRequired(Type, Generic[_T]):  # type: ignore
         pass
 
     class ParamSpec(Type):  # type: ignore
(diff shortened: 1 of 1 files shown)


predragnikolic:
Hello,

I tested this, and it looks like the `NotRequired` needs more work.

Some LSP-jdtls, LSP-julia, LSP-volar import NotRequired from LSP.plugin.core.typing.

This PR breaks them:
`TypeError: <class 'LSP.plugin.core.typing.NotRequired'> is not a generic class`

jfcherng:
> TypeError: <class 'LSP.plugin.core.typing.NotRequired'> is not a generic class

@predragnikolic Wondering whether [`dfbde05` (#2456)](https://github.com/sublimelsp/LSP/pull/2456/commits/dfbde0581cbd97760bb781d1e503e9414ed7510b) works for that.

predragnikolic:
It works.

Ideally LSP-* plugins can now import types from `typings` or `typing_extensions` instead of importing from `LSP.plugins.core.typing`. But `LSP.plugins.core.typing` can exist as long as there is something using it. 

And instead of List, Dict, we can use `list, dict, tuple,... from __future__ import annotations`

jfcherng:
> Ideally LSP-* plugins can now import types from `typings` or `typing_extensions` instead of importing from `LSP.plugins.core.typing`. But `LSP.plugins.core.typing` can exist as long as there is something using it.
> 
> And instead of List, Dict, we can use `list, dict, tuple,... from __future__ import annotations`

Yeah. That's what I encountered from using `lsp_utils`. Get complaints like `tuple` is not `Tuple` (from `LSP.plugins.core.typing`).

predragnikolic (approved):
Tested works

predragnikolic:
Thanks!

PR #2458, opened 2024-04-21
Import __future__ annotations

deathaxe:
This commit...

1. adds `from __future__ import annotations` to each relevant module in order to enable language level type annotation support.
2. as a result most quotation marks can be removed from type annotations.

### DIFF
--- a/boot.py
+++ b/boot.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 import os
 import sublime
 import sublime_plugin
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 from .core.promise import Promise
 from .core.protocol import CodeAction
 from .core.protocol import CodeActionKind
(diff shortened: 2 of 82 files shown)


predragnikolic:
this PR triggers an error in LSP-elm ->

```
  File "/Users/predrag/Library/Application Support/Sublime Text/Packages/LSP-elm/types.py", line 29, in <module>
    'destination': NotRequired[MoveDestination]
  File "/Users/predrag/Library/Application Support/Sublime Text/Lib/python38/typing_extensions.py", line 2071, in __getitem__
    item = typing._type_check(parameters,
  File "./python3.8/typing.py", line 149, in _type_check
TypeError: NotRequired accepts only a single type. Got {}.
reloading plugin LSP-elm.types
```

predragnikolic:
I will fix this in LSP-elm, 
but that error can be also happen in other LSP-*, although I have most checked out locally.

predragnikolic (approved):

predragnikolic:
Thanks!

PR #2460, opened 2024-04-21
Make more use of typing_extensions module

deathaxe:
This commit replaces various homegrown type definitions by imports from typing_extensions module.


predragnikolic (approved):
👍

PR #2468, opened 2024-05-03
Check CI status on Mac

jwortmann:
Here we can track whether CI is still broken on Mac (use Re-Run jobs button from the "checks" tab) and re-enable it once it gets fixed upstream.


predragnikolic:
deathaxe fixed it :)

PR #2469, opened 2024-05-03
Mark deprecated functions

jwortmann:


predragnikolic (approved):

plugin/core/url.py
```diff
@@ -31,6 +32,7 @@ def view_to_uri(view: sublime.View) -> str:
     return filename_to_uri(file_name)
 
 
+@deprecated("Use parse_uri() instead")
```
predragnikolic, line 35: is there any visible benefit in using the deprecated decorator?

I expected to see a log in the Sublime Text console, 
when a function marked with this decorator is called.

jwortmann: Not in the console, but for example Pyright reports a diagnostic if the deprecated functions are used (I have the severity set to "warning" here via `python.analysis.diagnosticSeverityOverrides`):

deprecated

PR #2473, opened 2024-05-07
docs: add info about tinymist

istudyatuni:


docs/src/language_servers.md
```diff
@@ -760,21 +760,35 @@ See [Javascript/TypeScript](#javascripttypescript).
...
+    If you want to provide some initialization options to tinymist, like exportPdf of typstExtraArgs, set it like
+
+    ```jsonc
+    "tinymist": {
+        // ...
+        "initializationOptions": {
+            "exportPdf": "never",
+            "typstExtraArgs": [],
+        }
+    }
```
rchl, line 789: It can be confusing to show the `tinymist` object as a root object. You should have the parent `clients` here also, like in the previous snippet.

docs/src/language_servers.md
```diff
@@ -793,7 +807,7 @@ See [Javascript/TypeScript](#javascripttypescript).
     <!-- how to call: see https://github.com/nvarner/typst-lsp/blob/master/editors/vscode/src/extension.ts -->
     ```jsonc title="Packages/User/Default.sublime-commands"
     [
-        // ...
+        // typst-lsp:
```
rchl, line 810: `// ...` were supposed to signify that there might be other content in this file also.

docs/src/language_servers.md
rchl, line 789: If there exist two servers then this section should be refactored to have the `## Typst` header and two separate `### ...` sub headers for each server.

istudyatuni: fixed

docs/src/language_servers.md
istudyatuni: fixed

docs/src/language_servers.md
istudyatuni: fixed

docs/src/language_servers.md
```diff
@@ -778,35 +778,90 @@ See [Javascript/TypeScript](#javascripttypescript).
...
+            "caption": "tinymist - Unpin the main file",
+            "command": "lsp_execute",
+            "args": {
+                "session_name": "tinymist",
+                "command_name": "tinymist.doPinMain",
+                "command_args": [null]
+            }
+        },
+    ]
+    ```
```
rchl, line 845: This seems like a copy-paste from the `Typst-lsp` section with just the server name changed. Are you sure that `Tinymist` supports same commands?

istudyatuni: Yes. Also, the argument for the second command is different

rchl: Then why can I not find any references to `doPinMain` in its repository (https://github.com/Myriad-Dreamin/tinymist/blob/main/editors/vscode/src/extension.ts)

istudyatuni: Thanks for noticing, I fixed this in my config, but forgot about documentation

rchl (approved):

docs/src/language_servers.md
Myriad-Dreamin: https://github.com/Myriad-Dreamin/tinymist/blob/ee50b24ce43d98d560e38c49ad8c34769a97a64e/crates/tinymist/src/server.rs#L262

@istudyatuni it is a lsp command provided by server.

```rs
.with_command("tinymist.pinMain", State::pin_document)
```

PR #2478, opened 2024-05-20
Remove logging.basicConfig()

### LINKED ISSUE #2477: `websocket_server` installs a root logger
Hi 😉 ,  I from time to time get to see logs with double logging in SublimeLinter-land.  E.g.

```
SublimeLinter: log_handler.py:77      Logging installed; log level INFO
INFO:SublimeLinter:Logging installed; log level INFO
```

The first line comes from the SublimeLinter handler/formatter but I never figured out where the second line comes from.
Just by luck, I finally found it:

https://github.com/sublimelsp/LSP/blob/723b129f73d65982ad88215208cc993c9b55b0f3/third_party/websocket_server/websocket_server.py#L18-L18

This installs a root logger.  Don't think that's even intended here.  Probably just deleting the line will be 🆗 .

predragnikolic:
Avoid polluting other packages.

closes #2477

### DIFF
--- a/third_party/websocket_server/websocket_server.py
+++ b/third_party/websocket_server/websocket_server.py
@@ -15,7 +15,6 @@ else:
     from socketserver import ThreadingMixIn, TCPServer, StreamRequestHandler
 
 logger = logging.getLogger(__name__)
-logging.basicConfig()
 
 '''
 +-+-+-+-+-------+-+-------------+-------------------------------+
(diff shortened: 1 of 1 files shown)


third_party/websocket_server/websocket_server.py
```diff
@@ -15,7 +15,6 @@
     from socketserver import ThreadingMixIn, TCPServer, StreamRequestHandler
 
 logger = logging.getLogger(__name__)
-logging.basicConfig()
```
rchl, line 18: Comment it out and add a comment that explains the reasoning in case we end at one point end up updating this dependency, so that we know why it was removed.

kaste:
Upstream issue is https://github.com/Pithikos/python-websocket-server/issues/85

PR #2484, opened 2024-05-28
Fix CodeActionTriggerKind enum access

jwortmann:
A bug that was missed, resulting from importing the `enum` stdlib on Python 3.8 (compare https://github.com/sublimelsp/LSP/pull/2446#issuecomment-2059553794).

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -403,9 +403,10 @@ def text_document_code_action_params(
     only_kinds: list[CodeActionKind] | None = None,
     manual: bool = False
 ) -> CodeActionParams:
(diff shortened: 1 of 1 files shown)


predragnikolic (approved):
> A bug that was missed

I do not consider the current behavior a bug.

With the current `main` the server still receives the correct int.
Screenshot 2024-05-29 at 18 54 59

I will merge this regardless, 
because this PR makes the triggerKind easer to read in the LSP log panel.

jwortmann:
You're right, technically the server communication wasn't affected by this, so maybe it's not a bug. I think I had earlier seen errors in the console about not being able to serialize the enums to JSON, but apparently it does still work and perhaps I remembered that wrong (or there was a different cause).

Issue #2381, opened 2023-12-21
setting for 'diagnostics' hover popup


HJX-001:
Settings for 'show_code_actions_in_hover' and 'show_symbol_action_links' are provided and known but somehow a new setting for just 'diagnostics(warnings etc.)' would be great for some class of users who dont want to fully disable 'hover' in 'disabled_functionalities'.

And hover popup may be made small (which can be extended via **read more** section)

ignore and close if not suitable

jwortmann:
From the next release there is a popup shown with the diagnostics from the line when you hover over a gutter icon.
Maybe that will be a sufficient alternative?

HJX-001:
the new feature of showing popup on hovering gutter icon looks awesome, but original problem still there.
![lsp_feature](https://github.com/sublimelsp/LSP/assets/116998733/a641d0e4-2909-4eea-b975-df7f865626e1)
basically a setting for enable/disable the area in red line is wanted (have a look at pic) as it looks too big and if require a look for 'diagnostics', gutter icon may be hovered (thanks for this awesome update).

PR #2488, opened 2024-05-31
feat: add option to disable/enable diagnostics in hover

HJX-001:
fixes issue #2381 
turn down if not suitable


HJX-001:
had a miss-spalled variable

PR #2489, opened 2024-05-31
feat: add option to disable/enable diagnostics in hover

HJX-001:
fixes issue #2381


plugin/hover.py
```diff
@@ -278,10 +278,11 @@ def show_hover(self, listener: AbstractViewListener, point: int, only_diagnostic
 
     def _show_hover(self, listener: AbstractViewListener, point: int, only_diagnostics: bool) -> None:
         hover_content = self.hover_content()
-        contents = self.diagnostics_content() + hover_content + code_actions_content(self._actions_by_config)
+        prefs = userprefs()
+        diagnostics_content = "" if not prefs.show_diagnostics_in_hover else self.diagnostics_content()
```
predragnikolic, line 282:
```suggestion
        diagnostics_content =  self.diagnostics_content() if prefs.show_diagnostics_in_hover else ""
```

predragnikolic:
Hello, 

The new setting should also be added to https://github.com/sublimelsp/LSP/blob/main/sublime-package.json

predragnikolic:
LSP has goto next/previous diagnostics command:

https://github.com/sublimelsp/LSP/blob/cd7cb682e634657c63105b906bacda99611846fd/Default.sublime-keymap#L147

https://github.com/sublimelsp/LSP/blob/cd7cb682e634657c63105b906bacda99611846fd/Default.sublime-keymap#L141

Currently, when we navigate to a diagnostics, the hover popup is shown with only diagnostics.
https://github.com/sublimelsp/LSP/blob/cd7cb682e634657c63105b906bacda99611846fd/plugin/core/registry.py#L265

I'm on a phone now, 
So I can't test this at the moment, but I assume that the diagnostics will not be show in the hover popup in that case with this PR.

I would suggest still showing the diagnostics in the hover popup
when navigating diagnostics with the two commands above.

HJX-001:
I have near zero python experience, and I dont have much idea of this codebase. Turn down this pr if not suitable and the issue can be closed now as It is working for me at least.

HJX-001:
Its my first pr so i m not very aware of workflow. It is requested to take over or guide me about 'suggested change'/ any other thing.

predragnikolic:
The PR is good, just needs some additional things.

It is ok if you don't know the codebase, 
And if you missed a few spots,
that's why code review exist 🙂

Fell free just to push additional commits.(There is no need to close and open a new PR)

If you have additional questions, feel free to ask.

HJX-001:
> LSP has goto next/previous diagnostics command:
> 
> https://github.com/sublimelsp/LSP/blob/cd7cb682e634657c63105b906bacda99611846fd/Default.sublime-keymap#L147
> 
> https://github.com/sublimelsp/LSP/blob/cd7cb682e634657c63105b906bacda99611846fd/Default.sublime-keymap#L141
> 
> Currently, when we navigate to a diagnostics, the hover popup is shown with only diagnostics.
> https://github.com/sublimelsp/LSP/blob/cd7cb682e634657c63105b906bacda99611846fd/plugin/core/registry.py#L265
> 
> I'm on a phone now, 
> So I can't test this at the moment, but I assume that the diagnostics will not be show in the hover popup in that case with this PR.
> 
> I would suggest still showing the diagnostics in the hover popup
> when navigating diagnostics with the two commands above.

I tested 'next/prev diagnostics', seems like you were right about it. It is not working. It feels intutive have this working. But on the other hand someone turning off 'diagnostics' in hover should know about it. I'll try to figure out a way for turning 'diagnostics' off for hover only, not for 'next/prev diagnostics'.

HJX-001:
added a new commit that seems resolving 'lsp prev/next diagnostics cmd' not working and added suggested changes and added entry in 'sublime-package.json'. It is requested to review updated changes

predragnikolic (approved):
Looks good

plugin/hover.py
```diff
@@ -278,10 +278,13 @@ def show_hover(self, listener: AbstractViewListener, point: int, only_diagnostic
 
     def _show_hover(self, listener: AbstractViewListener, point: int, only_diagnostics: bool) -> None:
         hover_content = self.hover_content()
-        contents = self.diagnostics_content() + hover_content + code_actions_content(self._actions_by_config)
+        prefs = userprefs()
+        diagnostics_content = ""
+        if only_diagnostics or prefs.show_diagnostics_in_hover:
+            diagnostics_content = self.diagnostics_content()
```
rchl, line 284:
```suggestion
        diagnostics_content = self.diagnostics_content() if only_diagnostics or prefs.show_diagnostics_in_hover else ""
```

HJX-001 (commented):
looks good

predragnikolic:
Thank you for the PR

Issue #2494, opened 2024-06-21
Allow `AbstractPlugin` to handle `$/progress` notification


jfcherng:
## Description

Similar to https://github.com/sublimelsp/LSP/issues/2493 but this time is asking for `$/progress` notification, whose payload is more structured than `window/logMessage`.

Currently `$/progress` notification is handled by LSP automatically at https://github.com/sublimelsp/LSP/blob/f512c96ec53cbd668dab29ab9b013423ae1300e5/plugin/core/sessions.py#L2177 and not propagated to the client (LSP-copilot in our case).

## Use Case

While [implementing LSP-copilot's chat](https://github.com/TerminalFi/LSP-copilot/pull/168) (or say "conversation" in its official term) feature, 

- We asked `Am I working in Sublime Text?`.
- Eventually, the server answered `No, you are working in an IDE called vscode.` in a `$/progress` notification.

However, all responses from the [copilot.vim](https://github.com/github/copilot.vim) server in this conversation are `$/progress`. Currently, there is no way for a client to handle any of them and hence LSP-copilot can't get the answer from the server.

## Communication Payloads

```
:: [15:00:05.997] --> LSP-copilot conversation/turn (8): {'conversationId': 'f4302595-dbbe-4411-9d82-e2367aa74217', 'message': 'Am I working in Sublime Text?', 'workDoneToken': '5'}
:: [15:00:06.011] <-  LSP-copilot $/progress: {'token': '5', 'value': {'kind': 'begin', 'title': 'Conversation f4302595-dbbe-4411-9d82-e2367aa74217 Turn e462ef0d-09c8-4e66-bb5e-5cf5fc0d99d7', 'conversationId': 'f4302595-dbbe-4411-9d82-e2367aa74217', 'turnId': 'e462ef0d-09c8-4e66-bb5e-5cf5fc0d99d7'}}
:: [15:00:06.011] <-  LSP-copilot $/progress: {'token': '5', 'value': {'kind': 'report', 'conversationId': 'f4302595-dbbe-4411-9d82-e2367aa74217', 'turnId': 'e462ef0d-09c8-4e66-bb5e-5cf5fc0d99d7', 'steps': [{'id': 'collect-context', 'title': 'Collecting context', 'status': 'running'}]}}
[...]

PR #2496, opened 2024-06-21
Add on_server_notification_async API

predragnikolic:
closes https://github.com/sublimelsp/LSP/issues/2494

This API allows LSP-* to hook into when a notification is received form the language server.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1047,6 +1047,15 @@ class AbstractPlugin(metaclass=ABCMeta):
         """
         pass
 
+    def on_server_notification_async(self, method: str, notification: Notification) -> None:
+        """
+        Notifies about a notification message that has been received from the language server.
+
+        :param    method:    The method of the notification.
+        :param    notification:  The notification object.
+        """
+        pass
+
     def on_open_uri_async(self, uri: DocumentUri, callback: Callable[[str, str, str], None]) -> bool:
         """
         Called when a language server reports to open an URI. If you know how to handle this URI, then return True and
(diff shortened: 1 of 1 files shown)


plugin/core/sessions.py
```diff
@@ -1047,6 +1047,14 @@ def on_server_response_async(self, method: str, response: Response) -> None:
         """
         pass
 
+    def on_server_notification_async(self, notification: Notification) -> None:
+        """
+        Notifies about a notification message that has been received from the language server.
+
+        :param    notification:  The notification object.
```
predragnikolic, line 1054: > ... "The notification object." 

I couldn't think of better 🙈

TerminalFi:
Although Copilot is doing some unconventional ideas. The use case is strong, in that helpers could also us it to raise information

predragnikolic:
Copilot is adhering to the spec here 

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#progress

> The base protocol offers also support to report progress in a generic fashion. This mechanism can be used to report any kind of progress including [work done progress](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workDoneProgress) (usually used to report progress in the user interface using a progress bar) **and partial result progress to support streaming of results.**

```
interface [ProgressParams](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#progress)<T> {
	/**
	 * The progress token provided by the client or server.
	 */
	token: [ProgressToken](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#progress);

	/**
	 * The progress data.
	 */
	value: T;
}
```

rchl (approved):

predragnikolic:
Just for leaving a clue on how to implement it in packages:

Here example with `NpmClientHandler`
```py
from lsp_utils import NpmClientHandler

class LspMyPackage(NpmClientHandler):

    def on_server_notification_async(self, notification) -> None:
        if notification.method == '$/progress':
            print('read this notification', notification.params)
```

Here example with `AbstractPlugin`

```py
from LSP.plugin import AbstractPlugin

class LspMyPackage(AbstractPlugin):

    def on_server_notification_async(self, notification) -> None:
        if notification.method == '$/progress':
            print('read this notification', notification.params)
```

PR #2500, opened 2024-06-27
Use generated types for python 38

predragnikolic:
- Switch from using [lsp_types_sublime_text_33.py](https://github.com/sublimelsp/lsp-python-types/blob/main/lsp_types_sublime_text_33.py) to using [lsp_types.py](https://github.com/sublimelsp/lsp-python-types/blob/main/lsp_types.py)
- Update the LSP types to match the latest [metaModel.schema.json](https://raw.githubusercontent.com/microsoft/vscode-languageserver-node/main/protocol/metaModel.schema.json)
- These two helper types are now removed, because they are no longer necessary:
```
CompletionItemDefaults = __CompletionList_itemDefaults_Type_1
CompletionEditRange = __CompletionList_itemDefaults_editRange_Type_1
```

### DIFF
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -6281,7 +6281,5 @@ CodeLensExtended = TypedDict('CodeLensExtended', {
     'session_name': str
 })
 
-CompletionEditRange = EditRangeWithInsertReplace
-
 # Temporary for backward compatibility with LSP packages.
 RangeLsp = Range
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
 from .typing import StrEnum
-from enum import IntEnum, IntFlag
+from enum import Enum, IntEnum, IntFlag
 from functools import total_ordering
 from typing import Any, Dict, Generic, Iterable, List, Literal, Mapping, TypedDict, TypeVar, Union
 from typing_extensions import NotRequired
 import sublime
+from typing_extensions import NotRequired
+from typing import Dict, List, Literal, TypedDict, Union
 
 INT_MAX = 2**31 - 1
 UINT_MAX = INT_MAX
@@ -220,7 +222,10 @@ class MessageType(IntEnum):
     Log = 4
     """ A log message. """
     Debug = 5
-    """ A debug message. """
+    """ A debug message.
+
+    @since 3.18.0
+    @proposed """
 
 
 class TextDocumentSyncKind(IntEnum):
@@ -360,6 +365,18 @@ class CodeActionKind(StrEnum):
     - Inline variable
     - Inline constant
     - ... """
+    RefactorMove = 'refactor.move'
+    """ Base kind for refactoring move actions: `refactor.move`
+
+    Example move actions:
+
+    - Move a function to a new file
+    - Move a property between classes
+    - Move method to base class
+    - ...
+
+    @since 3.18.0
+    @proposed """
     RefactorRewrite = 'refactor.rewrite'
     """ Base kind for refactoring rewrite actions: 'refactor.rewrite'
 
@@ -384,9 +401,14 @@ class CodeActionKind(StrEnum):
     They should not suppress errors or perform unsafe fixes such as generating new types or classes.
 
     @since 3.15.0 """
+    Notebook = 'notebook'
+    """ Base kind for all code actions applying to the entire notebook's scope. CodeActionKinds using
+    this should always begin with `notebook.`
+
+    @since 3.18.0 """
 
 
-class TraceValues(StrEnum):
+class TraceValue(Enum):
     Off = 'off'
     """ Turn tracing off. """
     Messages = 'messages'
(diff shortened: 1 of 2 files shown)


plugin/core/sessions.py
```diff
@@ -330,6 +330,7 @@ def get_initialize_params(variables: dict[str, str], workspace_folders: list[Wor
             "contextSupport": True,
             "signatureInformation": {
                 "activeParameterSupport": True,
+                "noActiveParameterSupport": False,
```
rchl, line 333: Shouldn't be needed if it's the same as default?

plugin/core/edit.py
```diff
@@ -1,6 +1,6 @@
 from __future__ import annotations
 from .logging import debug
-from .protocol import Position
+from .protocol import Position, is_text_edit
```
rchl, line 3: convention is to have each import on separate line

plugin/core/protocol.py
```diff
@@ -384,9 +399,14 @@ class CodeActionKind(StrEnum):
...
     @since 3.15.0 """
+    Notebook = 'notebook'
+    """ Base kind for all code actions applying to the entire notebook's scope. CodeActionKinds using
+    this should always begin with `notebook.`
+
+    @since 3.18.0 """
 
 
-class TraceValues(StrEnum):
+class TraceValue(Enum):
```
rchl, line 409: Why did this change from `StrEnum` to `Enum`?

plugin/core/protocol.py
```diff
@@ -720,5298 +813,5150 @@ class TokenFormat(StrEnum):
...
+    """ An optional token that a server can use to report partial results (e.g. streaming) to
+    the client. """
 
 
-DidChangeWorkspaceFoldersParams = TypedDict('DidChangeWorkspaceFoldersParams', {
-    # The actual workspace folder change event.
-    'event': 'WorkspaceFoldersChangeEvent',
-})
-""" The parameters of a `workspace/didChangeWorkspaceFolders` notification. """
+class SelectionRange(TypedDict):
```
rchl, line 1035: Is using `Class` equivalent to the old way in terms of overhead (memory, speed)?

plugin/core/protocol.py
predragnikolic: It was always Enum, but was changed to StrEnum in this PR https://github.com/sublimelsp/LSP/pull/2446/files#diff-049653877a86e4c220cd537d174655150e80b4ce0077d48742cc72b4654e2112R388

In the metaModel.json schema it is typed as Enum.
I'll leave it as Enum, because `TraceValue` is not used anywhere in code yet.

plugin/core/protocol.py
predragnikolic: One noticable reason on why I want to switch to `class SelectionRange(TypedDict):` form `SelectionRange = TypedDict('SelectionRange', {` syntax is because classes: allows for doc comments:
Screenshot 2024-07-02 at 12 46 32

The previous syntax that we used was an alternative syntax for TypedDict for python version less than 3.5.

I haven't done any benchmarks. I didn't notice any noticeable improvements or slowness.

plugin/core/protocol.py
rchl: > In the metaModel.json schema it is typed as Enum.

In metaModel.json there is no concept of StrEnum or IntEnum so that's not really an argument. :)

I think the answer to my question is that the name of it has changed but the https://github.com/sublimelsp/lsp-python-types/blob/f48ed1acee6040d64fc2d69df4eee77667008fcf/generate.py#L26 wasn't updated.

I think `StrEnum` would be correct here but if we are not using it then I guess it doesn't matter.

plugin/core/edit.py
rchl: There is more places in this PR where this is done.

plugin/core/protocol.py
predragnikolic: you are correct, thanks for the lookup

predragnikolic: I've pushed a fixed https://github.com/sublimelsp/lsp-python-types/commit/0da4d7c4d7543404d9a322f487280990d04980af

plugin/core/signature_help.py
```diff
@@ -168,7 +170,8 @@ def _parameter_documentation(self, view: sublime.View, signature: SignatureInfor
         if not parameters:
             return None
         try:
-            parameter = parameters[signature.get("activeParameter", self._active_parameter_index)]
+            active_parameter = cast(Uint, signature.get("activeParameter"))
```
rchl, line 173: It doesn't look like you need to cast here.

plugin/core/protocol.py
```diff
@@ -3,8 +3,8 @@
 from enum import IntEnum, IntFlag
 from functools import total_ordering
 from typing import Any, Dict, Generic, Iterable, List, Literal, Mapping, TypedDict, TypeVar, Union
-from typing_extensions import NotRequired
 import sublime
+from typing_extensions import NotRequired, TypeGuard
```
rchl, line 7: sorting :)

plugin/core/protocol.py
```diff
@@ -720,5298 +813,5150 @@ class TokenFormat(StrEnum):
...
 
 
-FoldingRangeRegistrationOptions = TypedDict('FoldingRangeRegistrationOptions', {
-    # A document selector to identify the scope of the registration. If set to null
-    # the document selector provided on the client side will be used.
-    'documentSelector': Union['DocumentSelector', None],
-    # The id used to register the request. The id can be used to deregister
-    # the request again. See also Registration#id.
-    'id': NotRequired[str],
+CallHierarchyIncomingCall = TypedDict('CallHierarchyIncomingCall', {
```
rchl, line 1124: Why is this still using `TypedDict`

predragnikolic: https://github.com/sublimelsp/LSP/pull/2500/files#diff-049653877a86e4c220cd537d174655150e80b4ce0077d48742cc72b4654e2112R1126

It has a property "from" which is a reserved keyword and cannot be used as a valid class property name. the generator will not use the class syntax in that case.

rchl (approved):

plugin/core/edit.py
```diff
@@ -24,7 +25,11 @@ def parse_workspace_edit(workspace_edit: WorkspaceEdit) -> WorkspaceChanges:
             uri = text_document['uri']
             version = text_document.get('version')
             edits = document_change.get('edits')
-            changes.setdefault(uri, ([], version))[0].extend(edits)
+            changes.setdefault(uri, ([], version))[0].extend(
+                # currently LSP code supprots only `TextEdit`s
+                # TODO: Extend WorkspaceEdit to support `AnnotatedTextEdit` and `SnippetTextEdit`
+                [edit for edit in edits if is_text_edit(edit)]
+            )
```
rchl, line 32: Not sure if it's this part but this PR seems to break something related to applying code edits.

If you try to apply fixes in this file by saving it with LSP-eslint and `"source.fixAll.eslint": true,` https://github.com/rchl/test-admin-js/blob/main/src/admin/options.ts then the syntax somewhat breaks

plugin/core/protocol.py
```diff
@@ -6335,8 +6280,9 @@ def to_lsp(self) -> Position:
...
 
-CompletionItemDefaults = __CompletionList_itemDefaults_Type_1
-CompletionEditRange = __CompletionList_itemDefaults_editRange_Type_1
-
 # Temporary for backward compatibility with LSP packages.
 RangeLsp = Range
+
+
+def is_text_edit(value: Any) -> TypeGuard[TextEdit]:
+    return isinstance(value, dict) and bool(value.get('range')) and bool(value.get('newText'))
```
rchl, line 6288: yeah, that's not correct. Use `in` to check for existance of property and not `bool`

rchl: But also this does not exclude `AnnotatedTextEdit` if that was your goal. You'd need to explicitly check for lack of `annotationId`

predragnikolic: Also true, I did the filtering to make pyright happy. There is probably a better way to handle this, than the way I did.

rchl: Then we can probably just extend the type of `changes` so that it allows `AnnotatedTextEdit`. We are not supposed to get them anyway unless we say so.

plugin/core/edit.py
predragnikolic: Yes, I can verify that it breaks the syntax.

plugin/core/protocol.py
predragnikolic: > yeah, that's not correct. Use in to check for existance of property and not bool

Thanks, this is the reason for this bug reported here https://github.com/sublimelsp/LSP/pull/2500#discussion_r1666795122

predragnikolic: I tried to extend the type:
```diff
- WorkspaceChanges = dict[str, tuple[list[TextEdit], Optional[int]]]
+ WorkspaceChanges = dict[str, tuple[list[TextEdit | AnnotatedTextEdit | SnippetTextEdit], Optional[int]]]
```

I quit from changing the code because:
- there is this error which I cant seem to figure out how to solve, and not cause type issues on other places  -> 
```
// Packages/LSP/plugin/core/edit.py:
32:17  error   Argument of type "tuple[List[TextEdit], None]" cannot be assigned to parameter "value" of type "tuple[list[TextEdit | AnnotatedTextEdit | SnippetTextEdit], int | None]" in function "__setitem__" ​Pyright
                    "tuple[List[TextEdit], None]" is incompatible with "tuple[list[TextEdit | AnnotatedTextEdit | SnippetTextEdit], int | None]"
                      Tuple entry 1 is incorrect type
                        "List[TextEdit]" is incompatible with "list[TextEdit | AnnotatedTextEdit | SnippetTextEdit]"
                          Type parameter "_T@list" is invariant, but "TextEdit" is not the same as "TextEdit | AnnotatedTextEdit | SnippetTextEdit"
                          Consider switching from "list" to "Sequence" which is covariant
```

AnnotatedTextEdit don't require code changes, however SnippetTextEdit do require code changes because they doesn't have `newText`, instead they have a `snippet` property. I rather not extend the PR to support SnippetTextEdit, at least not in this PR. I would leave the filtering of Text edits `[edit for edit in edits if is_text_edit(edit)]`. I just wanted to update the types with this PR.

rchl: `AnnotatedTextEdit` is defined as `interface AnnotatedTextEdit extends TextEdit` so it does include `newText`, same as `TextEdit`.

I can look into the error later.

rchl: I've pushed a change that should address it in a different way

rchl (approved):

plugin/core/protocol.py
predragnikolic: Thanks, I like the way it was addressed

PR #2513, opened 2024-09-13
Use orjson to de-/serialize json-rpc messages

deathaxe:
After a recent discussion about json serialization performance, I was curious about impact orjson makes on LSP processing performance.

This PR showcases how orjson as transport de-/serializer.

During recent 2 weeks of daily use, I haven't realized any issues, but also my projects are small enough to not realize any significant impact on overall performance.

Maybe someone using heavy language servers such as typescript can try to benchmark it.

### DIFF
--- a/plugin/core/transports.py
+++ b/plugin/core/transports.py
@@ -7,7 +7,6 @@ from functools import partial
 from queue import Queue
 from typing import Any, Callable, Dict, Generic, IO, Protocol, TypeVar
 import http
-import json
 import os
 import shutil
 import socket
@@ -17,6 +16,11 @@ import threading
 import time
 import weakref
 
+try:
+    import orjson
+except ImportError:
+    import json
+    orjson = None
 
 T = TypeVar('T')
 T_contra = TypeVar('T_contra', contravariant=True)
@@ -91,6 +95,17 @@ class JsonRpcProcessor(AbstractProcessor[Dict[str, Any]]):
     @staticmethod
     def _decode(message: bytes) -> dict[str, Any]:
         return json.loads(message.decode('utf-8'))
+
+
+class OrjsonRpcProcessor(JsonRpcProcessor):
+
+    @staticmethod
+    def _encode(data: dict[str, Any]) -> bytes:
+        return orjson.dumps(data)
+
+    @staticmethod
+    def _decode(message: bytes) -> dict[str, Any]:
+        return orjson.loads(message)
 
 
 class ProcessTransport(Transport[T]):
@@ -231,7 +246,10 @@ class ProcessTransport(Transport[T]):
 
 
 # Can be a singleton since it doesn't hold any state.
-json_rpc_processor = JsonRpcProcessor()
+if orjson:
+    json_rpc_processor = OrjsonRpcProcessor()
+else:
+    json_rpc_processor = JsonRpcProcessor()
 
 
 def create_transport(config: TransportConfig, cwd: str | None,
--- a/dependencies.json
+++ b/dependencies.json
@@ -3,6 +3,7 @@
         ">=4096": [
             "bracex",
             "mdpopups",
+            "orjson",
             "typing_extensions",
             "wcmatch"
         ]


predragnikolic:
It can be tested with this package that Rafal wrote
https://github.com/sublimelsp/LSP-performance-test

deathaxe:
According to this plugin, orjson doesn't make any difference. Tests results in 0.189 - 0.192 with both json and orjson on my machine.

predragnikolic:
This thread has some information of where bottle necks are located, and how they were solved
https://github.com/sublimelsp/LSP/pull/2190

deathaxe:
Yeah, and those include json serialization performance being an aspect. With promised performance of orjson I would have expected a little improvement with regards to deserialize messages from language servers.

rchl:
That test tests performance of our `format_completion` and not json deserialization. It would have to be modified.

rchl:
I've added relevant tests in https://github.com/sublimelsp/LSP-performance-test

It does appear to be twice as fast:

```
orjson
Best of 3: 0.09664862500000027 usec. All times: [0.0971984999999993, 0.09785637499999922, 0.09664862500000027]

json:
Best of 3: 0.197854125000001 usec. All times: [0.2019498329999987, 0.19887404199999992, 0.197854125000001]
```

@deathaxe are you able to get this PR into a state that passes CI?

rchl:
For the record, that was parsing of 50MB payload.

Here are also results for parsing a tiny payload 10000 times:

```
Best of 3: 0.02308479100000227 usec. All times: [0.023518416999991132, 0.02308479100000227, 0.023143875000002367]

orjson
Best of 3: 0.0058567079999996 usec. All times: [0.006264167000001208, 0.006002708000000467, 0.0058567079999996]

**Comment by deathaxe** [comment]
No idea how to satisfy pyright's limitations.

**Comment by rchl** [comment]
Add `orjson==3.10.7` in:
 - https://github.com/sublimelsp/LSP/blob/603632ea703d6296348680890859c148ac85229a/.github/workflows/main.yml#L44-L44
 - https://github.com/sublimelsp/LSP/blob/603632ea703d6296348680890859c148ac85229a/tox.ini#L22-L25

File: plugin/core/transports.py [hunk 1]
```diff
@@ -7,7 +7,6 @@
 from queue import Queue
 from typing import Any, Callable, Dict, Generic, IO, Protocol, TypeVar
 import http
-import json
```
rchl, line 10: Should be fine to always import it and remove the import from the `except` block.

rchl:
> I would prefer an import-time decision about using orjson or json. But pyright doesn't get that right.

Probably good enough as is. Getting the types right otherwise does indeed seem tricky.

jwortmann:
> ```
> orjson
> Best of 3: 0.09664862500000027 usec. All times: [0.0971984999999993, 0.09785637499999922, 0.09664862500000027]
> 
> json:
> Best of 3: 0.197854125000001 usec. All times: [0.2019498329999987, 0.19887404199999992, 0.197854125000001]
> ```
> For the record, that was parsing of a 50MB payload.

Just for comparison, the time for a single frame on a 120 Hz display is 8333.3 µs, so the performance improvement of parsing a 50 MB payload would make around 0.0012 % of a single frame time on a 120 Hz display. Not sure if that is worth to introduce another dependency...

---

The Python code could be simplified to single lines:
```python
return orjson.loads(message) if orjson else json.loads(message.decode('utf-8'))
```

deathaxe:
> The Python code could be simplified to single lines:

That's just another flavor, but no real simplification. Actually such one-liners are often harder to read and don't execute faster by any means.

rchl:
> _Edit:_ [#2190 (comment)](https://github.com/sublimelsp/LSP/pull/2190#issuecomment-1421622081) says current JSON parsing takes 0.293 seconds, so there is probably an error in your reported times? Should it be seconds instead of `usec`?

The output of the test is actually in seconds.
I guess I got confused because according to the documentation, the CLI variant of timeout outputs in usec.

jwortmann (approved):

PR #2514, opened 2024-09-15
Feature: Truncate inlay hint label

bivashy:
### Description
This pull request allows inlay hints to be truncated, which could be useful if inlay hints are too large. For example: TypeScript has absurdly long inlay hints (see screenshot 1). 
The default truncate limit is 1000. 
~~Truncated content reveal is not implemented, which means you cannot see the full content of the inlay hint if it is truncated.~~

Tested with:
`LSP-typescript`
`LSP-gopls`
`LSP-rust-analyzer`

---

<h6 align="center">1. Long inlay hint</h6>

long-inlay-hint

<h6 align="center">2. `LSP-typescript` (limit 30)</h6>

lsp-typescript-inlay-hint

<h6 align="center">3. `LSP-gopls` (limit 30)</h6>

lsp-gopls

<h6 align="center">4. `LSP-rust-analyzer` (limit 30)</h6>

lsp-rust-analyzer


plugin/inlay_hint.py
```diff
@@ -162,7 +166,10 @@ def format_inlay_hint_label(inlay_hint: InlayHint, session: Session, phantom_uui
                 }
             })
             value += f'<a href="{inlay_hint_click_command}">'
-        value += html.escape(label_part['value'])
+        raw_label = label_part['value']
+        truncated_label = raw_label[:truncate_limit] + '...' if len(raw_label) >= truncate_limit else raw_label
```
rchl, line 170: `if len(raw_label) == truncate_limit` then we shouldn't be adding elipsis

plugin/inlay_hint.py
```diff
@@ -142,12 +143,15 @@ def format_inlay_hint_label(inlay_hint: InlayHint, session: Session, phantom_uui
...
-        result += f'<span title="{(tooltip + instruction_text).strip()}">{html.escape(label)}</span>'
+        truncated_label = label[:truncate_limit] + '...' if len(label) >= truncate_limit else label
+        result += f'<span title="{(tooltip + instruction_text).strip()}">{html.escape(truncated_label)}</span>'
         if is_clickable:
             result += "</a>"
         return result
 
     for label_part in label:
+        if truncate_limit <= 0:
+            break
```
rchl, line 154: kinda weird to completely skip this part of the code if limit is out of range. I would expect falling back to default value in this case.

plugin/inlay_hint.py
jfcherng: probably just use textwrap.shorten(...)

rchl: It looks like that one seems to collapse spaces but maybe that's OK...

> First the whitespace in text is collapsed (all whitespace is replaced by single spaces)

https://docs.python.org/3/library/textwrap.html#textwrap.shorten

LSP.sublime-settings
```diff
@@ -180,6 +180,9 @@
    // or a custom keybinding.
   "show_inlay_hints": false,
 
+  // Truncate limit for inlay hints. Truncates and adds an ellipsis at the end.
+  "inlay_truncate_limit": 1000,
```
rchl, line 184: Call it `inlay_hints_truncate_limit` to make it clear that it relates to the `show_inline_hints` above.

bivashy: Done in [528a430](https://github.com/sublimelsp/LSP/pull/2514/commits/528a43071c4d24d566fc9b251fcd77b05a39a88b)

plugin/inlay_hint.py
bivashy: We iterate over `label` which contains "parts". And we need to "exit" the loop if we cannot cut any more. 

For example: `: { session: { user: AdapterUser } }` will be split into parts like:
- `:`
- `{`
- ` `
- `session`
- `:`
- ` `
- `{`
...

We need to decrement `truncate_limit` each iteration and stop iterating when we reach the end of the truncation.

I think decrementing the method parameter can lead to confusion, so I will move truncate_limit to a local variable to make it more intuitive.

bivashy: Done (probably) in: [6335b74](https://github.com/sublimelsp/LSP/pull/2514/commits/6335b74522c5de8803f2c0349a45fe62392fd5b0)

I hope it will make sense now, please let me know if you have better approach

jwortmann:
Could you use `…` instead of `...`? I think that would look nicer.

I wonder whether this really needs a setting, maybe long labels could instead always be truncated if we pick a reasonable cutoff length? And if the label gets truncated, it could be added to the tooltip, so you can easily see the full inlay hint on mouse hover. So in that case the full label and probably also a `\n` (or `&lt;br&gt;`?) should be prepended to the `title` attribute at
https://github.com/sublimelsp/LSP/blob/f4223d6d13eb941370c0ac40b47a185befa3a852/plugin/inlay_hint.py#L145
and
https://github.com/sublimelsp/LSP/blob/f4223d6d13eb941370c0ac40b47a185befa3a852/plugin/inlay_hint.py#L170

Could be useful to try whether that works and how it would look like.

bivashy:
> Could you use `…` instead of `...`? I think that would look nicer.

Maybe we should make this configurable from the preferences? Or is this redundant?

plugin/inlay_hint.py
bivashy: > It looks like that one collapses spaces but maybe that's OK...
> 
> > First the whitespace in text is collapsed (all whitespace is replaced by single spaces)
> 
> https://docs.python.org/3/library/textwrap.html#textwrap.shorten

I made a quick prototype with `textwrap.shorten`, but it drops whitespace, and it affects the readability of the inlay hint:


<h6 align="center">With `textwrap.shorten`</h6>

textwrap-shorten-usage-example


<h6 align="center">With manual implementation</h6>

manual-truncation-usage-example

I think I will stick with manual truncation because it is easier to manage in the for loop. I think `textwrap.shorten` might not work properly if we're using "parts" of the label.

rchl:
> > Could you use `…` instead of `...`? I think that would look nicer.
> 
> Maybe we should make this configurable from the preferences? Or is this redundant?

We strive to limit the number of exposed options so no. But also `…` is the correct one to use semantically. Three separate dots don't mean much semantically. And also it's slightly shorter.

bivashy:
I've used suggested character instead of plain `...`. I think it looks nice.

more-compact-dots

plugin/inlay_hint.py
```diff
@@ -97,7 +97,8 @@ def inlay_hint_to_phantom(view: sublime.View, inlay_hint: InlayHint, session: Se
 
 
 def get_inlay_hint_html(view: sublime.View, inlay_hint: InlayHint, session: Session, phantom_uuid: str) -> str:
-    label = format_inlay_hint_label(inlay_hint, session, phantom_uuid)
+    truncate_limit = view.settings().get("inlay_hints_truncate_limit", userprefs().inlay_hints_truncate_limit)
```
jwortmann, line 100: What is the purpose to try to read this setting from the view settings first - so that you could set different values per syntax? This is not really something which is done this way with any of the other LSP settings I think, and I don't imagine it to be very useful.

plugin/inlay_hint.py
```diff
@@ -142,12 +143,16 @@ def format_inlay_hint_label(inlay_hint: InlayHint, session: Session, phantom_uui
             })
             result += f'<a href="{inlay_hint_click_command}">'
         instruction_text = '\nDouble-click to insert' if has_text_edits else ""
-        result += f'<span title="{(tooltip + instruction_text).strip()}">{html.escape(label)}</span>'
+        truncated_label = label[:truncate_limit] + '…' if len(label) > truncate_limit else label
+        result += f'<span title="{(tooltip + instruction_text).strip()}">{html.escape(truncated_label)}</span>'
```
jwortmann, line 147: This one should have the `truncation_tooltip` too I guess

plugin/inlay_hint.py
bivashy: Done in [1ba72a6](https://github.com/sublimelsp/LSP/pull/2514/commits/1ba72a66d41777c5b0f4ad237b149efba63b16ad)

plugin/inlay_hint.py
```diff
@@ -97,7 +97,8 @@ def inlay_hint_to_phantom(view: sublime.View, inlay_hint: InlayHint, session: Se
 
 
 def get_inlay_hint_html(view: sublime.View, inlay_hint: InlayHint, session: Session, phantom_uuid: str) -> str:
-    label = format_inlay_hint_label(inlay_hint, session, phantom_uuid)
+    truncate_limit = userprefs().inlay_hints_truncate_limit
+    label = format_inlay_hint_label(inlay_hint, session, phantom_uuid, truncate_limit)
```
rchl, line 101: Seems like it would be fine to read the pref in `format_inlay_hint_label` directly since we don't intend for `format_inlay_hint_label` to get called with any other value than the one from the pref.

plugin/inlay_hint.py
```diff
@@ -142,12 +143,19 @@ def format_inlay_hint_label(inlay_hint: InlayHint, session: Session, phantom_uui
             })
             result += f'<a href="{inlay_hint_click_command}">'
         instruction_text = '\nDouble-click to insert' if has_text_edits else ""
-        result += f'<span title="{(tooltip + instruction_text).strip()}">{html.escape(label)}</span>'
+        truncated = len(label) > truncate_limit
+        truncated_label = label[:truncate_limit] + '…' if truncated else label
+        tooltip_label = "".join(label_part['value'] for label_part in label)
```
rchl, line 148: Typechecker like LSP-pyright should tell you that there is an error here `label` is a string so `label_part['value']` doesn't exist in this case.

plugin/inlay_hint.py
```diff
@@ -162,10 +170,20 @@ def format_inlay_hint_label(inlay_hint: InlayHint, session: Session, phantom_uui
...
             })
             value += f'<a href="{inlay_hint_click_command}">'
-        value += html.escape(label_part['value'])
+        raw_label = label_part['value']
+        truncated = len(raw_label) > remaining_truncate_limit
+        if truncated:
+            truncated_label = raw_label[:remaining_truncate_limit] + '…'
+        else:
+            truncated_label = raw_label
+
```
rchl, line 179: more python-like

```suggestion
        truncated_label = raw_label[:remaining_truncate_limit] + '…' if truncated else raw_label
```

Also we have convention in the code to not add blank links in function bodies (with some exceptions and some places where someone forgot to follow it ;)).

bivashy: Ternary operator: [9f0f3c7](https://github.com/sublimelsp/LSP/pull/2514/commits/9f0f3c712b4be52fe5f5d38a8310f223e3e6e871)
Removal of blank lines: [1a5bae4](https://github.com/sublimelsp/LSP/pull/2514/commits/1a5bae46dcac82e7cc0336f091407c5da3d528f1)

plugin/inlay_hint.py
bivashy: Yup, done in: [08b9795](https://github.com/sublimelsp/LSP/pull/2514/commits/08b9795cdb829d43d6b4d6716f75a3eb38ce22ca)

bivashy:
Note: We need to hover three dots to show the tooltip, hovering text will not work. Should I be worried about this?

rchl:
> Note: We need to hover three dots to see the tooltip, hovering text will not work. Should I be concerned about this? Should I wrap result in `span` with title if input is truncated, but that might break other tooltips.

It doesn't sound like a common behavior that people would be familiar with so it would make the title hard to discover.

I haven't tried the changes yet, does this feel like an ST bug to you?

bivashy:
> > Note: We need to hover three dots to see the tooltip, hovering text will not work. Should I be concerned about this? Should I wrap result in `span` with title if input is truncated, but that might break other tooltips.
> 
> It doesn't sound like a common behavior that people would be familiar with so it would make the title hard to discover.
> 
> I haven't tried the changes yet, does this feel like an ST bug to you?

Nope, I don't know how to implement that properly.
I mean, if I wrap result in `span`, will it break something?
I can send commit with such implementation so you can take a look at it.

rchl:
I would have to look exactly what content is possible in this case but I would think that the inlay hints are plain text (they are in the spec) so adding an inline wrapper shouldn't change anything. It would only be an issue if it would be possible for there to be a child block element but I don't think it is in this case.

bivashy:
> I would have to look exactly what content is possible in this case but I would think that the inlay hints are plain text (they are in the spec) so adding an inline wrapper shouldn't change anything. It would only be an issue if it would be possible for there to be a child block element but I don't think it is in this case.

Tested, `span` inside of `span` works as expected:
```html
<span title="test">Some text is <span title="this is subspan?">subunder</span></span>
```

Results:
hover-out-span
hover-inner-span

bivashy:
> Note: We need to hover three dots to see the tooltip, hovering text will not work. Should I be concerned about this? Should I wrap result in `span` with title if input is truncated, but that might break other tooltips.

Now it works as expected, and doesn't conflict with other tooltips

plugin/inlay_hint.py
```diff
@@ -162,10 +174,18 @@ def format_inlay_hint_label(inlay_hint: InlayHint, session: Session, phantom_uui
...
+        remaining_truncate_limit -= len(raw_label)
+        value += html.escape(truncated_label)
         if has_command:
             value += "</a>"
         # InlayHintLabelPart.location is not supported
         instruction_text = '\nDouble-click to execute' if has_command else ""
         result += f"<span title=\"{(tooltip + instruction_text).strip()}\">{value}</span>"
+    if truncated:
+        tooltip_label = "".join(label_part['value'] for label_part in label)
+        truncation_tooltip = html.escape('\n' + tooltip_label)
```
rchl, line 189: It doesn't need to start with a newline here.

plugin/inlay_hint.py
```diff
@@ -142,12 +145,21 @@ def format_inlay_hint_label(inlay_hint: InlayHint, session: Session, phantom_uui
...
         if is_clickable:
             result += "</a>"
+        if truncated:
+            truncation_tooltip = html.escape('\n' + label)
+            result = f"<span title=\"{truncation_tooltip}\">{result}</span>"
         return result
 
+    remaining_truncate_limit = truncate_limit
     for label_part in label:
+        if remaining_truncate_limit < 0:
```
rchl, line 161:
```suggestion
        if remaining_truncate_limit <= 0:
```

plugin/inlay_hint.py
bivashy: Done in [bc41741](https://github.com/sublimelsp/LSP/pull/2514/commits/bc41741055fed8ccac2f7b91a2372baea218e214)

plugin/inlay_hint.py
bivashy: We need to change 

`truncated = len(raw_label) > remaining_truncate_limit`

 to
 
`truncated = len(raw_label) >= remaining_truncate_limit`

Otherwise, if the inlay hint length matches the truncate limit, no dots are applied (limit is 10 characters):

possible-bug

bivashy: Done in [abe0987](https://github.com/sublimelsp/LSP/pull/2514/commits/abe09875b6b058e45fd97fda86b92df0ab838881)

LSP.sublime-settings
```diff
@@ -180,6 +180,9 @@
    // or a custom keybinding.
   "show_inlay_hints": false,
 
+  // Truncate limit for inlay hints. Truncates and adds an ellipsis at the end.
+  "inlay_hints_truncate_limit": 1000,
```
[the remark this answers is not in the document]
bivashy: I think we need to treat 0 or negative values as "no limit". I'll implement that.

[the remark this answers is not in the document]
bivashy: You're right, I thought this feature shouldn't affect if someone is okay with long inlay hints.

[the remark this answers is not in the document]
bivashy: Done in [17591c1](https://github.com/sublimelsp/LSP/pull/2514/commits/17591c1aa66cf73764684a89250ef0a817b18112)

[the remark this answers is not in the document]
bivashy: Done in [09162a2](https://github.com/sublimelsp/LSP/pull/2514/commits/09162a2561632c20226e57d41686cd8076870f13)

plugin/inlay_hint.py
```diff
@@ -142,12 +145,21 @@ def format_inlay_hint_label(inlay_hint: InlayHint, session: Session, phantom_uui
...
-        result += f'<span title="{(tooltip + instruction_text).strip()}">{html.escape(label)}</span>'
+        truncated = len(label) >= truncate_limit and truncate_limit > 0
+        truncated_label = label[:truncate_limit] + '…' if truncated else label
+        result_tooltip = (tooltip + instruction_text).strip()
+        result += f'<span title="{result_tooltip}">{html.escape(truncated_label)}</span>'
         if is_clickable:
             result += "</a>"
+        if truncated:
+            truncation_tooltip = html.escape(label)
+            result = f"<span title=\"{truncation_tooltip}\">{result}</span>"
```
jwortmann, line 156: I have not tested it but this looks like it would result in something like
```html
<span title="Truncation tooltip"><span title="Regular tooltip">Label</span></span>
```
I don't think this would work, i.e. it would only show `Truncation tooltip` on hover, instead of both of them, or does it?

jwortmann:
I think a better default length would be something like 30 or 40, and I would probably name the setting "inlay_hints_truncate_length" or "inlay_hints_max_length" (just a suggestion).

plugin/inlay_hint.py
bivashy: You are right, but it works as expected. See comment: https://github.com/sublimelsp/LSP/pull/2514#issuecomment-2379405302

jwortmann: Then let's first clarify what is expected. In your example "test" would be the full `label` of the [InlayHint](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#inlayHint), and "this is subspan?" would be the (optional) `tooltip`. I would expect that the `tooltip` is always visible when hovering (if it exists), and the full `label` is additionally visible if it is truncated. Your screenshot suggests that `tooltip` would be hidden if the label gets truncated.

What was wrong with only having a single `title` attribute and appending/prepending the `truncation_tooltip` content plus linebreak?

(Besides that, it seems the `tooltip` can be [MarkupContent](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#markupContentInnerDefinition), so I'm not sure whether the assumption from https://github.com/sublimelsp/LSP/pull/2514#issuecomment-2376144994 is correct. But in this case we're probably processing it not quite correctly anyways, which is a different issue.)

bivashy: > visible

Sorry, I missed your point earlier. Now I understand. The label for truncated content won't be shown if a label_part has a tooltip, but it will be displayed in areas where the tooltip is missing. I didn't see it as an issue, but I get your concern now.

For appending/prepending the tooltip, I'm not sure yet. One idea is to append the tooltip on a new line with the full label. Does that sound good?

jwortmann: Yes, this was my initial idea. I'm not aware of any server which uses the optional `tooltip`, so perhaps we cannot test it directly. But it looks like you already had implemented it in this way before https://github.com/sublimelsp/LSP/pull/2514/commits/0fcdf798ecc1a97dd7bf44b4275d317a2f1548f5. Didn't it work correctly?

rchl: > I have not tested it but this looks like it would result in something like
> 
> ```
> <span title="Truncation tooltip"><span title="Regular tooltip">Label</span></span>
> ```
> 
> I don't think this would work, i.e. it would only show `Truncation tooltip` on hover, instead of both of them, or does it?

As far as I remember, this will show the "Regular tooltip" when hovering `Label`.

Whatever we do, it won't be very apparent to the user where each tooltip starts and ends unless we maybe add some additional styling.

bivashy: > Yes, this was my initial idea. I'm not aware of any server which uses the optional `tooltip`, so perhaps we cannot test it directly. But it looks like you already had implemented it in this way before [0fcdf79](https://github.com/sublimelsp/LSP/commit/0fcdf798ecc1a97dd7bf44b4275d317a2f1548f5). Didn't it work correctly?

I thought that wrapping with `<span>' would be more appropriate. I guess I'll implement with append/prepend.

bivashy: Done in [c6e0055](https://github.com/sublimelsp/LSP/pull/2514/commits/c6e0055b849610f937968c7ed8c80b6fb599124f)

plugin/inlay_hint.py
```diff
@@ -140,13 +142,23 @@ def format_inlay_hint_label(inlay_hint: InlayHint, session: Session, phantom_uui
                 }
             })
             result += f'<a href="{inlay_hint_click_command}">'
+        truncated = len(label) >= truncate_limit and truncate_limit > 0
+        truncated_label = label[:truncate_limit] + '…' if truncated else label
+
```
jwortmann, line 147: Shouldn't it be `>` instead of `>=`? If the label length is exactly `truncate_limit`, you don't need to add the dots. Also I think it should be `label[:truncate_limit - 1]`, so that the length with added dots doesn't exceed the `"inlay_hints_max_length"` value.

The empty line should be removed (also in line 127).

bivashy: > Shouldn't it be `>` instead of `>=`? If the label length is exactly `truncate_limit`, you don't need to add the dots. Also I think it should be `label[:truncate_limit - 1]`, so that the length with added dots doesn't exceed the `"inlay_hints_max_length"` value.
> 
> The empty line should be removed (also in line 127).

Please see: https://github.com/sublimelsp/LSP/pull/2514#discussion_r1779999461

bivashy: Empty lines are removed in: [b83a147](https://github.com/sublimelsp/LSP/pull/2514/commits/b83a1474059562bc0a72c7c68947aa3777b8822a)

I am unsure if we should change `>` into `>=`, because it was already reviewed, and requested to change into `>=`.

bivashy: Done in: [9be92d9](https://github.com/sublimelsp/LSP/pull/2514/commits/9be92d9085791260140dc74a96ec3670a313fbc0)

plugin/inlay_hint.py
```diff
@@ -140,13 +141,21 @@ def format_inlay_hint_label(inlay_hint: InlayHint, session: Session, phantom_uui
...
+        truncation_tooltip = f'\n{html.escape(label)}' if truncated else ""
+        result_tooltip = (tooltip + instruction_text + truncation_tooltip).strip()
+        result += f'<span title="{result_tooltip}">{html.escape(truncated_label)}</span>'
         if is_clickable:
             result += "</a>"
         return result
-
+    remaining_truncate_limit = truncate_limit
+    full_label = "".join(label_part['value'] for label_part in label)
+    full_label_truncated = len(full_label) >= truncate_limit and truncate_limit > 0
```
jwortmann, line 155: Should this also be `>` here or not?

---

I have tested the PR with a server that uses the regular label (no label parts) and I think it works well with the tooltips.

My only suggestion left would be to use a lower default value for the setting. I think 30 would be a sane default. I would say the truncation is actually a good thing to do, and some users might not check settings or read update notes, so it should be applied by default. Opinions?

sublime-package.json
```diff
@@ -739,6 +739,12 @@
               "default": false,
               "markdownDescription": "Show inlay hints in the editor. Inlay hints are short annotations within the code, which show variable types or parameter names.\nThis is the default value used for new windows but can be overriden per-window using the `LSP: Toggle Inlay Hints` command from the Command Palette, Main Menu or a custom keybinding."
             },
+            "inlay_hints_max_length": {
+              "type": "integer",
+              "default": 100,
+              "minimum": 0,
+              "markdownDescription": "Truncate limit for inlay hints. Truncates and adds an ellipsis at the end."
```
jwortmann, line 746: Update to match description in settings?

plugin/inlay_hint.py
bivashy: > Should this also be `>` here or not?
> 
> I have tested the PR with a server that uses the regular label (no label parts) and I think it works well with the tooltips.
> 
> My only suggestion left would be to use a lower default value for the setting. I think 30 would be a sane default. I would say the truncation is actually a good thing to do, and some users might not check settings or read update notes, so it should be applied by default. Opinions?

I'll check it out. Thanks for pointing that out.

---

I think inlay hint is not enabled by default, so most users will not be affected. So I agree that this feature should be enabled by default. I'll wait for more opinions.

sublime-package.json
bivashy: Done in [dc7d5ca](https://github.com/sublimelsp/LSP/pull/2514/commits/dc7d5ca31f0918a586659d12ee735ca17c18a215)

plugin/inlay_hint.py
bivashy: > I'll check it out. Thanks for pointing that out.

Done in [d6a544a](https://github.com/sublimelsp/LSP/pull/2514/commits/d6a544a1fa3a558b16bb485fd454fed6a07d5777).

---

Done in [a43e68a](https://github.com/sublimelsp/LSP/pull/2514/commits/a43e68adbd1c500e8f887f626bf5282e6a72b11e)

I will not resolve this conversation in case of reconsideration, should I?

sublime-package.json
```diff
@@ -739,6 +739,12 @@
               "default": false,
               "markdownDescription": "Show inlay hints in the editor. Inlay hints are short annotations within the code, which show variable types or parameter names.\nThis is the default value used for new windows but can be overriden per-window using the `LSP: Toggle Inlay Hints` command from the Command Palette, Main Menu or a custom keybinding."
             },
+            "inlay_hints_max_length": {
+              "type": "integer",
+              "default": 30,
+              "minimum": 0,
+              "markdownDescription": "Maximum length for inlay hints. Truncates exceeding characters and adds an ellipsis."
```
rchl, line 746: 0 means "no truncation", right? We should specify that.

bivashy: Done in [572b1cd](https://github.com/sublimelsp/LSP/pull/2514/commits/572b1cd7699544fc92a1deedcdd550d190254818)

jwortmann (approved):

LSP.sublime-settings
```diff
@@ -180,6 +180,9 @@
    // or a custom keybinding.
   "show_inlay_hints": false,
 
+  // Maximum length for inlay hints. Truncates exceeding characters and adds an ellipsis. Zero or less means no truncation.
```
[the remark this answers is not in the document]
jwortmann: I pushed a small change for the description.

Is there anything more or can we merge the PR? (I have tested only for the regular inlay hints without label parts)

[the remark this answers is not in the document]
bivashy: > I pushed a small change for the description.
> 
> Is there anything more or can we merge the PR? (I have tested only for the regular inlay hints without label parts)

I think it is ready to use. I don't have any more changes. 

---

Thanks to all reviewers for their patience!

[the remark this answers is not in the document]
bivashy: > A linebreak would be nice here. Less than zero would violate the JSON schema btw.

Linebreak after each sentence, if I understand correctly

rchl (approved):

PR #2518, opened 2024-09-21
fix: ensure ending a session triggers listeners

rchl:
I was looking into introducing some global session listener concept for plugin use but while doing so I found out that on closing a session we don't trigger various lifecycle listeners that we should be triggering. It appears that `Session` is deleted before `Transport` is closed so `Session.on_transport_close` and following logic is not executed.

The one-line fix is to change `self._sessions` in `WindowManager` from `WeakSet[Session]` to `set[Session]`.

I'm not sure why we are using `WeakSet` here. There could be a valid reason when this makes sense or it could be due to old convention where we've relied more on GC than on explicitly notifying when Session/SessionView/SessionBuffer are closed.

(I've also left a bunch of logs that I'm of course planning to remove but those can be used to see the difference with and without this fix when closing the last file of some session, for example.)

### DIFF
--- a/docs/src/commands.md
+++ b/docs/src/commands.md
@@ -43,14 +43,15 @@ You can include special variables in the `command_args` array that will be autom
 
 | Variable | Type | Description |
 | -------- | ---- | ----------- |
-| `"$document_id"` | object | JSON object `{ 'uri': string }` containing the file URI of the active view, see [Document Identifier](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentIdentifier) |
+| `"$document_id"` | object | JSON object `{ "uri": string }` containing the URI of the active view, see [TextDocumentIdentifier](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentIdentifier) |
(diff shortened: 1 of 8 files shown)


rchl:
Logs on closing the last file of a session

Without the fix:

```
[DocumentListener] _clear_session_views_async
[Session] end_async False
[Session] Sending shutdown
[Session] Closing transport
[Session] __del__ LSP-typescript
[ProcessTransport] end Transport None
```

With the fix:

```
[DocumentListener] _clear_session_views_async
[Session] end_async False
[Session] Sending shutdown
[Session] Closing transport
[ProcessTransport] end Transport <LSP.plugin.core.sessions.Session object at 0x1058472b0>
[Session] on_transport-close
[WindowManager] on_post_exit_async LSP-typescript
[Session] __del__ LSP-typescript
```

rchl:
Even after the changes:
- On closing a window the `WindowManager.destroy() is called and it ends and removes all sessions from the set so the listeners are not called in this case. This is perhaps OK.
- On restarting a session the `WindowManager.restart_sessions_async()` is called and it also ends the session and removes it from the set immediately so the listeners are also not called. This might not be ideal since a proper cleanup might not happen.

PR #2523, opened 2024-09-29
feat: add GlobalLspListener

rchl:
I've been thinking a bit on how to support support https://github.com/sublimelsp/LSP-volar/issues/221

My latest idea is to not do any magic to make LSP-typescript initialize for Vue files and instead when LSP-volar is initialized on a vue file, and LSP-typescript is not running, add a Phantom at the top of the file that tells the user to initialize LSP-typescript to get the full experience. Clicking on the phantom triggers a window command that modifies project settings that enables LSP-typescript for Vue files and also enables the vue plugin by referencing it from the LSP-volar storage path. Would look something like:

Screenshot 2024-09-29 at 22 02 23

To implement that idea I need some global listener that will tell me whenever a session initializes or shuts down on a view thus this API. It would be used like this:

```py
from LSP.plugin import GlobalLspListener
from LSP.plugin import Session
from LSP.plugin import windows

class SessionListener(GlobalLspListener):

    def __init__(self) -> None:
        self._phantoms: dict[int, sublime.PhantomSet] = dict()

    def on_session_initialized_async(self, view_listener: AbstractViewListener, session: Session) -> None:
        pass

    def on_session_shutdown_async(self, view_listener: AbstractViewListener, session: Session) -> None:
        pass


session_listener = SessionListener()


def plugin_loaded():
    LspVolarPlugin.setup()
    windows.add_global_listener(session_listener)


def plugin_unloaded():
    windows.remove_global_listener(session_listener)
    LspVolarPlugin.cleanup()
```

Let me know if you have any better idea on how this could be handled.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,7 +41,7 @@ jobs:
           python-version: '3.8'
       - run: sudo apt update
       - run: sudo apt install --no-install-recommends -y x11-xserver-utils
-      - run: pip3 install mypy==1.7.1 flake8==5.0.4 pyright==1.1.339 --user
+      - run: pip3 install mypy==1.7.1 flake8==5.0.4 pyright==1.1.381 --user
       - run: echo "$HOME/.local/bin" >> $GITHUB_PATH
       - run: mypy stubs
       - run: flake8 plugin tests
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,7 +41,7 @@ jobs:
           python-version: '3.8'
       - run: sudo apt update
       - run: sudo apt install --no-install-recommends -y x11-xserver-utils
-      - run: pip3 install mypy==1.7.1 flake8==5.0.4 pyright==1.1.381 --user
+      - run: pip3 install mypy==1.7.1 flake8==5.0.4 pyright==1.1.381 orjson==3.10.7 --user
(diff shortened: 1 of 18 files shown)


jwortmann:
I think `LspSessionListener` could be a better name than `GlobalLspListener`.

I'm not a big fan of exposing `windows` and doing the add/remove manually. Would it be possible to handle it automatically, like the event listeners from the ST API, maybe similar to this?
https://github.com/sublimelsp/LSP/blob/d71e11a9d428462a5994508f7891914c3fa74258/boot.py#L175-L184

Alternatively I would define two new public functions `register_listener` / `unregister_listener` instead of using class methods on `windows`.

But in general, if the use case is only the interaction between LSP-volar and LSP-typescript, shouldn't this whole logic better be handled in LSP-volar? The `__init__` / `on_post_start` / `on_pre_send_request_async` methods of AbstractPlugin should be enough to know when a session initializes or is about to shutdown.

rchl:
> But in general, if the use case is only the interaction between LSP-volar and LSP-typescript, shouldn't this whole logic better be handled in LSP-volar? The `__init__` / `on_post_start` / `on_pre_send_request_async` methods of AbstractPlugin should be enough to know when a session initializes or is about to shutdown.

In this case LSP-volar needs to know when LSP-typescript initializes or shuts down.

I'm still thinking about this and experimenting with different approaches though.

PR #2524, opened 2024-09-30
remove unnecessary use of generators for session retrieval

rchl:
Use of generators for retrieving the session is IMO completely unnecessary (correct me if I'm missing something).

1. This is not an expensive operation that would benefit from being interrupted early. Getting all sessions and returning a list is not gonna be slower. It will potentially even be faster as I imagine that generators have some overhead.
2. Passing generators around can lead to errors or at least confusion since the values can only be retrieved once from it. In cases like https://github.com/sublimelsp/LSP/blob/603632ea703d6296348680890859c148ac85229a/plugin/core/registry.py#L24-L33 where we pass generators and disguise it as `Iterable[Session]`, there is no way of knowing that the code will fail if trying to iterate sessions twice.

### DIFF
--- a/plugin/core/registry.py
+++ b/plugin/core/registry.py
@@ -12,7 +12,7 @@ from .views import uri_from_view
 from .windows import WindowManager
 from .windows import WindowRegistry
 from functools import partial
-from typing import Any, Generator, Iterable
+from typing import Any, Generator
 import operator
 import sublime
 import sublime_plugin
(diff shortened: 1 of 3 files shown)


plugin/core/registry.py
```diff
@@ -21,14 +21,14 @@
 windows = WindowRegistry()
 
 
-def best_session(view: sublime.View, sessions: Iterable[Session], point: int | None = None) -> Session | None:
+def best_session(view: sublime.View, sessions: list[Session], point: int | None = None) -> Session | None:
```
jwortmann, line 24: I think this change for the parameter type is unnecessary here because `list` is still an `Iterable`, so this just makes the function more restrictive where it doesn't need to be. And `Iterable` would be more inline with the usage in `max`, which takes an iterable https://docs.python.org/3/library/functions.html#max

But frankly it doesn't really matter and I think nobody payed much intention of this throughout the rest of the code base.

jwortmann (approved):

PR #2527, opened 2024-10-10
Attempt to open file as fallback for URIs without scheme

ryuukk:
I kept getting `error: Unable to open URI`, i didn't know why

Turns out the plugin requires the path to have a scheme `file://`

I think it's fine to still try to open it as a fallback

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1702,7 +1702,8 @@ class Session(TransportCallbacks):
         # There is no pre-existing session-buffer, so we have to go through AbstractPlugin.on_open_uri_async.
         if self._plugin:
             return self._open_uri_with_plugin_async(self._plugin, uri, r, flags, group)
-        return None
+        # Server didn't provide a URI scheme, let's try to open the file, it should be a path anyways
+        return self._open_file_uri_async(uri, r, flags, group)
 
     def open_uri_async(
         self,


jwortmann:
This function takes a `DocumentUri` as input parameter and per definition an URI must always have a scheme because the scheme part is mandatory.

When does this error happen? For example for links in a hover popup there is a fallback to open it in a browser if the link doesn't have a scheme:
https://github.com/sublimelsp/LSP/blob/d71e11a9d428462a5994508f7891914c3fa74258/plugin/hover.py#L369-L372

predragnikolic:
> I think it's fine to still try to open it as a fallback

The URI spec says the following: 
> The scheme and path components are required, though the path may be
   empty

Ref: https://datatracker.ietf.org/doc/html/rfc3986#section-3.1

> a fallback or a proper message is better than returning None and giving a generic error message

When I saw the issue, I immediately knew that the error was due to a wrong document URI, I deduced that by the generic error message. So I'm fine withe the error message. if you have a suggestion on how the error message could be improved, do tell.

rchl:
Feel free to submit a PR that improves the error message.

But note that there is an infinite amount of non-compliant cases (not only for this case but in general) that the server could trigger and it's not very reasonable to expect that the client will handle every one of those in a way that provides detailed and very relevant response.

PR #2531, opened 2024-10-15
Don't attach LSP to syntax test files

### LINKED ISSUE #2504: Disable language server for syntax test files by default
## TL;DR

By default, disable servers for

- syntax test files
- SublimeREPL views

## Description

I think enabling servers on syntax test files only cause (linting) issues. Only very few ones (e.g., LSP-copilot) will work well. [By suggested rules](https://www.sublimetext.com/docs/syntax.html#testing), the name of a syntax test file starts with `syntax_test_` (but in `Default/run_syntax_tests.py`, it finds filenames start with `syntax_test`).

The same reason for disabling for SublimeREPL views. See https://github.com/sublimelsp/LSP-pyright/issues/343
A SublimeREPL view has the view setting `"repl": True,`.

## Suggestion

Something equivalent to

```py
    @classmethod
    def should_ignore(cls, view: sublime.View) -> bool:
        return bool(
            # SublimeREPL views
            view.settings().get("repl")
            # syntax test files
            or os.path.basename(view.file_name() or "").startswith("syntax_test")
        )
```

jwortmann:
Closes #2504

Maybe having this check in `is_regular_view` is not really accurate (regarding the function name), but I think it is the best place for reusability.


jwortmann:
@jfcherng
> Only very few ones (e.g., LSP-copilot) will work well.

Merging this PR would of course also prevent LSP-copilot to run on syntax test files. I don't know how useful LSP-Copilot is for those files; but since you were the one who opened the linked issue, I assume that you are okay with it?

jfcherng:
> @jfcherng
> 
> > Only very few ones (e.g., LSP-copilot) will work well.
> 
> Merging this PR would of course also prevent LSP-copilot to run on syntax test files. I don't know how useful LSP-Copilot is for those files; but since you were the one who opened the linked issue, I assume that you are okay with it?

I hardly write syntax test :D

rchl (approved):

PR #2532, opened 2024-10-16
Make DocumentSyncListener more efficient if no server is running

jwortmann:
Currently if LSP (this package) is installed, it runs quite some code in the background whenever the callbacks from [`DocumentSyncListener`](https://github.com/sublimelsp/LSP/blob/a6b17a4aa4d10d8ddc225413d221c4241b877f8b/plugin/documents.py#L139), which is a subclass of `sublime.ViewEventListener`, are triggered. This happens for *all* views, not only when a server is active. For example it creates an empty (if no server is running) `sublime.CompletionList` and schedules to run more code on the async thread in https://github.com/sublimelsp/LSP/blob/a6b17a4aa4d10d8ddc225413d221c4241b877f8b/plugin/documents.py#L555-L560
or it runs the debouncing in https://github.com/sublimelsp/LSP/blob/a6b17a4aa4d10d8ddc225413d221c4241b877f8b/plugin/documents.py#L384 via `sublime.set_timeout_async` each time when the cursor position changes.

All of this runs on the async thread, so you won't immediately experience any lag from this. But still it feels very inefficient and I wondered how it could be prevented or improved. We can't use the `is_applicable` classmethod for `DocumentSyncListener`, because if the view's syntax is changed later, an LSP session might still get attached to this view. So my idea would be to add a check in each of the ST callback methods and return early if there is no session running for this view. Does it make sense or is there a better solution?

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -328,6 +328,8 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         return list(self._session_views.values())
 
     def on_text_changed_async(self, change_count: int, changes: Iterable[sublime.TextChange]) -> None:
+        if not self.sessions_async():
+            return
         if self.view.is_primary():
             for sv in self.session_views_async():
                 sv.on_text_changed_async(change_count, changes)
@@ -364,6 +366,8 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
             return
         if not self._registered:
             self._register_async()
+        if not self.sessions_async():
+            return
         if userprefs().show_code_actions:
             self._do_code_actions_async()
         for sv in self.session_views_async():
@@ -382,6 +386,8 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
                 sb.do_inlay_hints_async(self.view)
 
     def on_selection_modified_async(self) -> None:
+        if not self.sessions_async():
+            return
         first_region, _ = self._update_stored_selection_async()
         if first_region is None:
             return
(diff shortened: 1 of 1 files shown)


plugin/documents.py
```diff
@@ -328,6 +328,8 @@ def session_views_async(self) -> list[SessionView]:
         return list(self._session_views.values())
 
     def on_text_changed_async(self, change_count: int, changes: Iterable[sublime.TextChange]) -> None:
+        if not self.sessions_async():
+            return
         if self.view.is_primary():
             for sv in self.session_views_async():
                 sv.on_text_changed_async(change_count, changes)
```
rchl, line 335: So I guess this could be considered a bit more efficient since we don't have to get sessions twice.
The con is that the code has to be case-specific and be less consistent.

```suggestion
    def on_text_changed_async(self, change_count: int, changes: Iterable[sublime.TextChange]) -> None:
        sessons = self.session_views_async()
        if not sessions:
            return
        if self.view.is_primary():
            for sv in sessions:
                sv.on_text_changed_async(change_count, changes)
```

The alternative that would look clean is to have a decorator that would be attached to a method. But that's obviously less flexible and wouldn't allow us to create fully optimized early exit.

jwortmann: Actually we can always check for SessionViews, because `self.sessions_async` uses `self.session_buffers_async` -> `self.session_views_async` anyway. https://github.com/sublimelsp/LSP/pull/2532/commits/7a8b4821c38f5a4de25e021963035d67429e684f

I do like the idea with the decorator though.

rchl (commented):
One thing that is a potential concern is that there could be a case where we rely on given method cleaning up stuff after session is destroyed. For example on restarting session. Though I know that there already exist issues with that so if we noticed something wrong with this change, it might not be due to it actually...

rchl (approved):

PR #2537, opened 2024-10-25
Various small code style improvements

jwortmann:
See individual commits for details. The biggest change is to use the namespaced enum constants from the Python 3.8 ST API, instead of global constants from the `sublime` module. This has the advantage that only the relevant values are shown when auto-completing, for example after typing `sublime.RegionFlags.`.

There is no functional change.


rchl:
(just squashed two last follow-up commits to the original ones)

rchl (approved):

PR #2538, opened 2024-10-25
fix: consider priority_selector when using best_session during rename

rchl:
Pass point to `best_session()` so that `priority_selector` is respected.

I've noticed that having LSP-some-sass and LSP-css running at once sends `prepareRename` to LSP-some-sass but `rename` to `LSP-css`.

Having this fix + `prioriority_selector` defined in LSP-some-sass will fix the issue.

Related to https://github.com/sublimelsp/LSP-some-sass/pull/4


jwortmann:
I think the correct way to fix this would be to pass the session name from `prepare_provider_session` to `_on_prepare_result` and from there further as an additional argument to when the `lsp_symbol_rename` command is called a second time. Then if this name is not `None`, use `self.session_by_name` instead of `self.best_session` (in `_do_rename`) to select the rename session. This would guarantee that it is really the same session.

Otherwise there might be only one of the servers having support for `renameProvider.prepareProvider`, but another session has a higher `priority_selector`. In that case, the first session would still be used for the prepare rename request, and then the other session for the rename request.

Passing the point to `self.best_session` for the `renameProvider.prepareProvider` is certainly still useful (probably also for various other commands in LSP). I think the code block could be simplified to
```python
point = get_position(self.view, point=args.get('point'))
if not point:
    return None
```
though I see that in this PR it was just moved from one place to another.

---

> Related to https://github.com/sublimelsp/LSP/pull/2538

Infinite loop from self reference detected.

rchl:
> Passing the point to `self.best_session` for the `renameProvider.prepareProvider` is certainly still useful (probably also for various other commands in LSP). I think the code block could be simplified to
> 
> ```python
> point = get_position(self.view, point=args.get('point'))
> if not point:
>     return None
> ```

Good that someone still remembers what kind of utility functions we have :)

Though in this case I think we should actually not return since the prepare rename request should not need the point. (Not that I ever see `point` being `None` here...)

rchl:
We the addition of `session_name` for `lsp_rename`, now the whole fix for the `point` doesn't even matter for the https://github.com/sublimelsp/LSP-some-sass/pull/4 case anymore. But of course it's still a good fix.

plugin/rename.py
```diff
@@ -132,20 +134,17 @@ def is_visible(
     def input(self, args: dict) -> sublime_plugin.TextInputHandler | None:
         if "new_name" in args:
             # Defer to "run" and trigger rename.
-            return None
-        prepare_provider_session = self.best_session("renameProvider.prepareProvider")
+            return
```
jwortmann, line 137: I'd prefer this to explicitly return `None`, unless `None` is the only possible return value of the function (which is not the case here).

See [PEP8](https://peps.python.org/pep-0008/#programming-recommendations):

> Be consistent in return statements. Either all return statements in a function should return an expression, or none of them should. If any return statement returns an expression, any return statements where no value is returned should explicitly state this as return None, and an explicit return statement should be present at the end of the function (if reachable):

plugin/rename.py
```diff
@@ -154,29 +153,35 @@ def run(
...
-                self._do_rename(location, new_name)
+                self._do_rename(location, new_name, prepare_provider_session)
                 return
             # Trigger InputHandler manually.
             raise TypeError("required positional argument")
         if location is None:
             return
         params = cast(PrepareRenameParams, text_document_position_params(self.view, location))
         request = Request.prepareRename(params, self.view, progress=True)
+        prepare_provider_session_name = prepare_provider_session.config.name
```
jwortmann, line 175: Maybe a bit pedantic, but this variable could be just `session_name`. Otherwise it might imply that there are different session names for prepare provider and rename provider, but this is exactly what should be prevented with this.

Similarly we could/should just use `session` instead of `prepare_provider_session` above.

jwortmann (commented):
> Though in this case I think we should actually not return since the prepare rename request should not need the point.

Not sure about that. Isn't the whole point of the prepare rename request to check whether a given location is valid to rename?

The control flow of the rename command is quite complicated, but I think here it probably doesn't matter because it would still return None because "placeholder" is not in `args`, and then it will eventually return from `run` if the point is None (which means that there is no cursor in the view).

plugin/rename.py
```diff
@@ -133,19 +135,19 @@ def input(self, args: dict) -> sublime_plugin.TextInputHandler | None:
...
+            return None
+        session = self._get_prepare_rename_session(point, args.get('session_name'))
+        if session and "placeholder" not in args:
             # Defer to "run" and trigger "prepare" request.
             return None
         placeholder = args.get("placeholder", "")
         if not placeholder:
-            point = args.get("point")
+            if point is None:
+                return None
```
jwortmann, line 149: Can be removed

jwortmann (approved):

PR #2540, opened 2024-10-30
fix: code actions on save not fixing issues if saving quickly

### LINKED ISSUE #2511: Quickly saving after editing a file doesn't return text edits
**Describe the bug**

I've noticed recently that editing and immediately saving a file while using LSP-eslint does not fix the linting issue that this edit introduced.

After analysing it a bit I've noticed that we don't trigger `textDocument/diagnostic` request immediately after change and before requesting code actions on save while in VSCode it's the case. Given that vscode-eslint seems to be [requesting document validation from `textDocument/diagnostics` handler](https://github.com/microsoft/vscode-eslint/blob/700f0b6d1beb88de7706c3237dca683c3aed15b7/server/src/eslintServer.ts#L263-L263), it doesn't respond with fixes unless that handler is triggered before requesting code actions.

(I haven't checked the spec to see if it's a quirk of vscode-eslint or something expected but it does cause inferior behavior in this case)

**To Reproduce**

1. Install LSP-eslint
2. Enable `"lsp_code_actions_on_save": { "source.fixAll.eslint": true }`
3. Introduce a lint issue (for example extra indent) and immediately save the file

**Expected behavior**

The indentation fix should get applied.

**Logs**

VSCode:

```
[Trace - 14:18:14] Sending notification 'textDocument/didChange'.
Params: {
    "textDocument": {
        "uri": "file:///somefile.ts",
        "version": 4
    },
    "contentChanges": [
        {
            "range": {
                "start": {
                    "line": 71,
                    "character": 4
                },
                "end": {
                    "line": 71,
                    "character": 4
                }
            },
            "rangeLength": 0,
            "text": "    "
        }
    ]
}


[Trace - 14:18:14] Sending request 'textDocument/diagnostic - (7)'.
Params: {
    "identifier": "eslint",
    "textDocument": {
        "uri": "file:///somefile.ts"
    }
}


[Trace - 14:18:14] Sending request 'textDocument/codeAction - (8)'.
Params: {
    "textDocument": {
        "uri": "file:///somefile.ts"
    },
    "range": {
        "start": {
            "line": 0,
            "character": 0
        },
        "end": {
            "line": 73,
            "character": 0
        }
    },
    "context": {
        "diagnostics": [],
        "only": [
            "source.fixAll.eslint"
        ],
        "triggerKind": 2
    }
}

...

[Trace - 14:18:15] Received response 'textDocument/codeAction - (8)' in 919ms.
Result: [
    {
        "title": "Fix all fixable ESLint issues",
        "edit": {
            "documentChanges": [
                {
                    "textDocument": {
                        "uri": "file:///somefile.ts",
                        "version": 4
                    },
                    "edits": [
                        {
                            "range": {
                                "start": {
                                    "line": 71,
                                    "character": 4
                                },
                                "end": {
                                    "line": 71,
                                    "character": 8
                                }
                            },
                            "newText": ""
                        }
                    ]
                }
            ]
        },
        "kind": "source.fixAll.eslint"
    }
]
```

- OS: macOS
- Sublime Text version: 4140
- LSP version: latest
- Language servers used: LSP-eslint

rchl:
Fix issue with code actions on save not fixing diagnostic errors that were just introduced, when saving immediately after changing the document.

For servers that support the pull diagnostics model we have to assume that the server doesn't re-compute diagnostics until client asks for them. Thus we have to ask just before sending code actions request or otherwise it might not provide fixes for issues that were just introduced.

To test, with a LSP-eslint running on a file, create an issue (like a trailing space) and save the file immediately. The issue is often not fixed on the first try.

Fixes #2511

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -130,6 +130,8 @@ class CodeActionsManager:
                 if diag_sb == sb:
                     diagnostics = diags
                     break
+            # Pull for diagnostics to ensure that server computes them before receiving code action request.
+            sb.do_document_diagnostic_async(view)
             params = text_document_code_action_params(view, region, diagnostics, matching_kinds, manual=False)
             return Request.codeAction(params, view)
 
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -130,8 +130,6 @@ class CodeActionsManager:
                 if diag_sb == sb:
                     diagnostics = diags
                     break
-            # Pull for diagnostics to ensure that server computes them before receiving code action request.
-            sb.do_document_diagnostic_async(view)
             params = text_document_code_action_params(view, region, diagnostics, matching_kinds, manual=False)
             return Request.codeAction(params, view)
 
@@ -165,6 +163,8 @@ class CodeActionsManager:
             session = sb.session
             request = request_factory(sb)
             if request:
+                # Pull for diagnostics to ensure that server computes them before receiving code action request.
+                sb.do_document_diagnostic_async(listener.view)
                 response_handler = partial(on_response, sb)
                 task: Promise[list[CodeActionOrCommand] | None] = session.send_request_task(request)
                 tasks.append(task.then(response_handler))
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -81,6 +81,15 @@ class PendingChanges:
         self.changes.extend(changes)
 
 
+class PendingDocumentDiagnosticRequest:
+
+    __slots__ = ('version', 'request_id')
+
+    def __init__(self, version: int, request_id: int) -> None:
+        self.version = version
+        self.request_id = request_id
+
+
 class SemanticTokensData:
 
     __slots__ = (
@@ -122,7 +131,7 @@ class SessionBuffer:
         self.diagnostics_flags = 0
         self._diagnostics_are_visible = False
         self.document_diagnostic_needs_refresh = False
-        self._document_diagnostic_pending_response: int | None = None
+        self._document_diagnostic_pending_request: PendingDocumentDiagnosticRequest | None = None
         self._last_synced_version = 0
         self._last_text_change_time = 0.0
         self._diagnostics_debouncer_async = DebouncerNonThreadSafe(async_thread=True)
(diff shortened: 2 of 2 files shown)


plugin/session_buffer.py
```diff
@@ -471,30 +480,32 @@ def update_document_link(self, new_link: DocumentLink) -> None:
...
+            params['identifier'] = identifier
+        result_id = self.session.diagnostics_result_ids.get(self._last_known_uri)
+        if result_id is not None:
+            params['previousResultId'] = result_id
+        request_id = self.session.send_request_async(
+            Request.documentDiagnostic(params, view),
+            partial(self._on_document_diagnostic_async, version),
+            partial(self._on_document_diagnostic_error_async, version)
+        )
+        self._document_diagnostic_pending_request = PendingDocumentDiagnosticRequest(version, request_id)
```
rchl, line 506: Moved check for `self.has_capability("diagnosticProvider")` to the top which changed indentation quite a bit and made it harder to review.

Apart from that, added a check and bail for the case when requesting diagnostics for document version for which there already is a pending request. Without the added check it would cancel existing and trigger a new one which is a waste if the document hasn't changed since.

predragnikolic:
> To test, with a LSP-eslint running on a file, create an issue (like a trailing space) and save the file immediately. The issue is often not fixed on the first try.

The code looks ok, but I cannot test it.
I didn't manage to reproduce this on my Laptop with that step with the code from `main`. LSP managed to fix apply the correct fix for the diagnostic every time, regardless of how fast I press save after making the error... I guess the fast laptop I use is to blame.

predragnikolic:
I guess it depends on how fast the diagnostics arrive. I get them almost instant even on the main branch. Looking at this code, this will make sure to pull diagnostics(only applicable to servers that support the pull model) before requesting code actions.

rchl:
The issue is that LSP didn't even request diagnostics before triggering code actions on save. It's not important that the diagnostics response is received, only that it is sent.

(Of course it's also necessary to enable code action on save which I forgot to mention - `"lsp_code_actions_on_save": { 	"source.fixAll.eslint": true }`.

predragnikolic:
yes I had code actions on save already enabled. 

Oh, I knew that the LSP spec allows to send diagnostics to in the CodeActionContext. I thought the we implemented to send diagnostics when requesting code actions.... I haven't seen that code in a while, and that topic is a bit blurry.

rchl:
It's not relevant for this issue which diagnostics are included in code actions request. Only that we trigger the diagnostic request itself.

plugin/code_actions.py
```diff
@@ -130,6 +130,8 @@ def request_factory(sb: SessionBufferProtocol) -> Request | None:
                 if diag_sb == sb:
                     diagnostics = diags
                     break
+            # Pull for diagnostics to ensure that server computes them before receiving code action request.
+            sb.do_document_diagnostic_async(view)
```
jwortmann, line 134: I probably would have put this part under `if request:` in `_collect_code_actions_async` (line 165), because it's not strictly necessary for creating the Request object.

jwortmann (commented):
I tried to test the bug, but for me LSP-eslint doesn't report any diagnostics at all. Perhaps I'm missing some required configuration file?

I guess to make it easier to reproduce, you could increase the FEATURES_TIMEOUT debounce time.

The code mostly looks good to me (besides see the comment below, maybe).

rchl:
Can try on https://github.com/rchl/lsp-log-parser - clone the project and run `yarn` or `npm i`.

plugin/code_actions.py
rchl: makes sense

Ideally perhaps this should be trigger before any code action request. While normal usage should not require that, if someone uses a macro or some sort of automation then it could become a problem then too.

rchl: Actually this code is used for any code action request so all good.

predragnikolic:
> I guess to make it easier to reproduce, you could temporarily increase the FEATURES_TIMEOUT debounce time in the code.

Yes, increasing the timeout helped to test this out.

Tested, works ✅

Issue #1386, opened 2020-10-22
Cancel inflight view requests when the view closes or when the view state changes


rwols:
**Is your feature request related to a problem? Please describe.**
When there is a request in flight for a particular view, say textDocument/documentSymbol, then the request is not cancelled when the view is closed.

**Describe the solution you'd like**
Cancel any view-specific request with the $/cancel notification once:

- The view state changes (i.e. once the buffer is modified)
- The view is closed

rwols:
We should cancel textDocument/documentHighlight when the selection changes.

rwols:
We now have an official client capability for this as well https://github.com/microsoft/language-server-protocol/commit/8aec48a51caf88e5c58cb082828abe028eb08396

PR #2543, opened 2024-11-02
fix: cancel pending requests as early as possible

rchl:
On document change, it makes sense to cancel pending semantic tokens/pull diagnostics requests as early as possible rather than waiting for the debounce timeout and potentially failing to cancel request on time and wasting resources on handling those.

### DIFF
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -311,8 +311,17 @@ class SessionBuffer:
                 self._pending_changes.update(change_count, changes)
                 purge = True
             if purge:
+                self._cancel_pending_requests_async()
                 debounced(lambda: self.purge_changes_async(view), FEATURES_TIMEOUT,
                           lambda: view.is_valid() and change_count == view.change_count(), async_thread=True)
+
+    def _cancel_pending_requests_async(self) -> None:
+        if self._document_diagnostic_pending_request:
+            self.session.cancel_request(self._document_diagnostic_pending_request.request_id)
+            self._document_diagnostic_pending_request = None
+        if self.semantic_tokens.pending_response:
+            self.session.cancel_request(self.semantic_tokens.pending_response)
+            self.semantic_tokens.pending_response = None
 
     def on_revert_async(self, view: sublime.View) -> None:
         self._pending_changes = None  # Don't bother with pending changes
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -487,7 +487,9 @@ class SessionBuffer:
 
     # --- textDocument/diagnostic --------------------------------------------------------------------------------------
 
-    def do_document_diagnostic_async(self, view: sublime.View, version: int | None = None) -> None:
+    def do_document_diagnostic_async(
+        self, view: sublime.View, version: int | None = None, ignore_pending: bool = False
+    ) -> None:
         mgr = self.session.manager()
         if not mgr or not self.has_capability("diagnosticProvider"):
             return
@@ -496,7 +498,7 @@ class SessionBuffer:
         if version is None:
             version = view.change_count()
         if self._document_diagnostic_pending_request:
-            if self._document_diagnostic_pending_request.version == version:
+            if self._document_diagnostic_pending_request.version == version and not ignore_pending:
                 return
             self.session.cancel_request(self._document_diagnostic_pending_request.request_id)
         params: DocumentDiagnosticParams = {'textDocument': text_document_identifier(view)}
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -488,7 +488,7 @@ class SessionBuffer:
     # --- textDocument/diagnostic --------------------------------------------------------------------------------------
 
     def do_document_diagnostic_async(
-        self, view: sublime.View, version: int | None = None, ignore_pending: bool = False
+        self, view: sublime.View, version: int | None = None, forced_update: bool = False
     ) -> None:
         mgr = self.session.manager()
         if not mgr or not self.has_capability("diagnosticProvider"):
@@ -498,7 +498,7 @@ class SessionBuffer:
         if version is None:
             version = view.change_count()
         if self._document_diagnostic_pending_request:
-            if self._document_diagnostic_pending_request.version == version and not ignore_pending:
+            if self._document_diagnostic_pending_request.version == version and not forced_update:
                 return
             self.session.cancel_request(self._document_diagnostic_pending_request.request_id)
         params: DocumentDiagnosticParams = {'textDocument': text_document_identifier(view)}
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -671,7 +671,7 @@ class SessionBufferProtocol(Protocol):
     def do_semantic_tokens_async(self, view: sublime.View) -> None:
         ...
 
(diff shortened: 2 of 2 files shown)


plugin/session_buffer.py
```diff
@@ -311,9 +311,18 @@ def on_text_changed_async(self, view: sublime.View, change_count: int,
...
+                self._cancel_pending_requests_async()
                 debounced(lambda: self.purge_changes_async(view), FEATURES_TIMEOUT,
                           lambda: view.is_valid() and change_count == view.change_count(), async_thread=True)
 
+    def _cancel_pending_requests_async(self) -> None:
+        if self._document_diagnostic_pending_request:
+            self.session.cancel_request(self._document_diagnostic_pending_request.request_id)
+            self._document_diagnostic_pending_request = None
+        if self.semantic_tokens.pending_response:
+            self.session.cancel_request(self.semantic_tokens.pending_response)
```
rchl, line 323: I have an urge to rename `pending_response` to `pending_request`.

`pending_response` sounds like a boolean state (waiting for response) rather than what it actually is.

jwortmann:
In that case I guess we don't need these parts anymore:

https://github.com/sublimelsp/LSP/blob/04827c78d04cc6f055e8e061a6e3c8a6c1b882c3/plugin/session_buffer.py#L489-L492

https://github.com/sublimelsp/LSP/blob/04827c78d04cc6f055e8e061a6e3c8a6c1b882c3/plugin/session_buffer.py#L605-L606

Or is there maybe an edge case that I'm missing?

plugin/session_buffer.py
jwortmann: My thinking was that the request has already happened (was sent), so we're indeed only still waiting for the response now. But maybe we use the word "pending" in a sligthly different meaning. I guess it can both mean something like "upcoming" (that's how I used it here), or "unanswered". Of course the server response is directly coupled to the request, so feel free to rename it if you like to.

rchl:
> In that case I guess we don't need these parts anymore:
> 
> https://github.com/sublimelsp/LSP/blob/04827c78d04cc6f055e8e061a6e3c8a6c1b882c3/plugin/session_buffer.py#L489-L492
> 
> https://github.com/sublimelsp/LSP/blob/04827c78d04cc6f055e8e061a6e3c8a6c1b882c3/plugin/session_buffer.py#L605-L606
> 
> Or is there maybe an edge case that I'm missing?

I'd have to check. I thought that we still have to cancel but upon more thinking I've ruled out what I was thinking about before.

jwortmann:
Related issue: #1386

Should we set `general.staleRequestSupport.cancel` client capability to `true`?

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#clientCapabilities

```typescript
		/**
		 * Client capability that signals how the client
		 * handles stale requests (e.g. a request
		 * for which the client will not process the response
		 * anymore since the information is outdated).
		 *
		 * @since 3.17.0
		 */
		staleRequestSupport?: {
			/**
			 * The client will actively cancel the request.
			 */
			cancel: boolean;

			/**
			 * The list of requests for which the client
			 * will retry the request if it receives a
			 * response with error code `ContentModified``
			 */
			 retryOnContentModified: string[];
		}
```

rchl:
I'm fine with that.

I'm not sure what's the purpose of this capability though and whether it makes a difference that we don't consistently do canceling for every feature that could benefit from it.

rchl:
The only reference to `staleRequestSupport` that I found is https://github.com/swiftlang/sourcekit-lsp/issues/1650

I wonder if explicitly stating that we cancel requests would mean that we would potentially miss server canceling some requests that we don't currently handle.

rchl:
LSP-vue (new volar version) does something that makes the cancel in `do_semantic_tokens_async` still have a purpose.

```
:: [22:17:13.766] --> LSP-vue textDocument/semanticTokens/range (3): {'textDocument': {'uri': 'file:///usr/local/workspace/lsp/lsp-log-parser/components/app-bar.vue'}, 'range': {'start': {'line': 0, 'character': 0}, 'end': {'line': 162, 'character': 0}}}
:: [22:17:13.766] --> LSP-vue textDocument/documentLink (4): {'textDocument': {'uri': 'file:///usr/local/workspace/lsp/lsp-log-parser/components/app-bar.vue'}}
:: [22:17:13.780] <-- LSP-vue client/registerCapability (0): {'registrations': [{'id': 'a57c161c-26ab-45c1-821c-653beb6c96bc', 'method': 'workspace/didChangeConfiguration', 'registerOptions': {}}]}
:: [22:17:13.780] >>> LSP-vue (0) (duration: 0ms): None
:: [22:17:13.780] <-- LSP-vue workspace/configuration (1): {'items': [{'section': 'http'}]}
:: [22:17:13.780] >>> LSP-vue (1) (duration: 0ms): [None]
:: [22:17:13.780] <-- LSP-vue workspace/configuration (2): {'items': [{'section': 'http'}]}
:: [22:17:13.781] >>> LSP-vue (2) (duration: 0ms): [None]
:: [22:17:14.750] <-- LSP-vue client/registerCapability (3): {'registrations': [{'id': '7608602c-8896-4727-a2b6-dbf4ea45dbcb', 'method': 'workspace/didChangeWatchedFiles', 'registerOptions': {'watchers': [{'globPattern': '**/*.{js,cjs,mjs,ts,cts,mts,jsx,tsx,json,vue}'}]}}]}
:: [22:17:14.752] >>> LSP-vue (3) (duration: 1ms): None
:: [22:17:14.752] <-- LSP-vue workspace/configuration (4): {'items': [{'section': 'typescript.preferences'}]}
:: [22:17:14.752] >>> LSP-vue (4) (duration: 0ms): [{'autoImportFileExcludePatterns': [], 'importModuleSpecifier': 'shortest', 'importModuleSpecifierEnding': 'auto', 'jsxAttributeCompletionStyle': 'auto', 'quoteStyle': 'auto'}]
:: [22:17:14.752] <<< LSP-vue (2) (duration: -): None
:: [22:17:14.753] <-- LSP-vue workspace/inlayHint/refresh (5): None
:: [22:17:14.753] >>> LSP-vue (5) (duration: 0ms): None
:: [22:17:14.754] <-- LSP-vue workspace/semanticTokens/refresh (6): None
:: [22:17:14.988] <<< LSP-vue (3) (duration: -): {'resultId': '1730582234987', 'data': [...]}
```

Summary

```
:: [22:17:13.766] --> LSP-vue textDocument/semanticTokens/range
:: [22:17:14.754] <-- LSP-vue workspace/semanticTokens/refresh (6): None
:: [22:17:14.988] <<< LSP-vue (3) (duration: -): {'resultId': '1730582234987', 'data': [...]}
...

rchl:
And since with pull diagnostics the server can also trigger `workspace/diagnostic/refresh`, in theory the same can happen with those.

Though that made me realize that ignoring a new request when the view version hasn't changed could in theory be incorrect if the server can return different diagnostics at different point in time (because it initialized late or something). That said, it might be a rare corner case.

jwortmann (approved):
> And since with pull diagnostics the server can also trigger workspace/diagnostic/refresh, in theory the same can happen with those.

Right, good that you thought of that.

I wonder whether `ignore_pending` is the best name for this parameter, it's a bit difficult to understand in my opinion. Maybe something like `forced` or `forced_update` would be better?

Issue #2545, opened 2024-11-03
Disable blocking popup modal errors by default


eugenesvk:
**Is your feature request related to a problem? Please describe.**

After debugging some packages and adding a few to the list of ignored packages I get an infninite notification window

infinite mean that as soon as I close one, another one immediately pops up, so I can't even quit the editor properly!!! (if spam the close button fast enough, I am able to do)

And even outside of such pathological cases I don't care about a failing JSON parsing enough to even have a full UI blocking modal thrown in my face, so please disable it by default.

**Describe the solution you'd like**

UI blocking operations on errors are disabled by default and people who like them can toggle them on.

Instead you could show an error in the status bar, and even add some ❗❗❗ symbols to make it stand out but without having to block the UI

**Describe alternatives you've considered**
n/a

**Additional context**

That notification modal window:

<img width="253" alt="lsp-json" src="https://github.com/user-attachments/assets/662bbac6-ade8-4fb5-87d0-8a47721689ff">

(I think this is a generic LSP window, but not sure, so sorry if it's LSP-json-package specific)

rchl:
There are different kind of errors and for some modal dialog works better than other options. Ideally ST would have like a permanent popup that would stay until dismissed manually but it doesn't so I don't really see a better option for errors like when server fails to start.

Status field doesn't really work well for those kind of errors. For many reasons. For example it doesn't have that much space to work with, it's not possible to trigger any action on clicking on it and it's not clear when such error should be cleared in such case.

Ideally, the errors related to server failing to start should be very rare.

Infinite errors is a bug and should be fixed but we need some steps to reproduce those.

eugenesvk:
> for some a modal dialog works better

Which ones? You could also have a different default for those, and let the users replace with less intrusive error type via a separate config

>  permanent popup that would stay until dismissed manually

But that's exactly what modal is! It's just that you open a new popup every time the old one is closed.

> I don't really see a better option

The better option is to have user configuration to not show this dialog in the first place. It's a simple notification that requires no useful action from the user, and re. your other concerns:
  - there is an infinite amount of space in the console
  - and you can have user configurable timer after which the error goes away
  - pressing "OK" to acknowledge an error isn't a useful action

> Ideally, errors related to server failing to start should be very rare.

[...]

PR #2546, opened 2024-11-06
Allow suppressing UI blocking dialogs

eugenesvk:
Add 2 user configuration options that would allow to suppress the attention-demanding UI blocking dialogs, instead redirecting warnings/errors to a status message and the full console message 

Simple way to close https://github.com/sublimelsp/LSP/issues/2545 while preserving the current defaults

Though not all modals are affected since some are invoked by the lsp_utils dependency


plugin/tooling.py
```diff
@@ -130,15 +132,19 @@ def run(self, base_package_name: str) -> None:
         try:
             urllib.parse.urlparse(base_url)
         except Exception:
-            sublime.error_message("The clipboard content must be a URL to a package.json file.")
+            msg = "The clipboard content must be a URL to a package.json file."
+            status = "Clipboard must be a URL to package.json"
+            notify_err(msg, status)
```
rchl, line 137: I honestly don't think we are gonna accept any of these changes (not in this form). It doesn't make sense to just blindly change all instance of modals. Especially in this file it's a case of explicit user action and it makes sense to report the error in the dialog (the user won't be surprised by it) instead of having to dig for more info the console or potentially miss status field message.

eugenesvk: The users to whom it makes sense to see these kind of dialogs or who worry about missing a status field message can simply continue to use the defaults. Or they can set the regular dialog suppression setting, but leave the error setting on.

So this setting is simply not for them, but for the other group who have never had an issue of missing a status field message (by the way, you could add ❗ symbol if you think this is a very important message, that's harder to miss), don't see the point of a blocking modal for a simple error message. For them these bad interface design patterns don't make sense because they do not reflect the importance of the messages.

> explicit user action and it makes sense to report the error in the dialog (the user won't be surprised by it)

The issue here isn't surprise, but, requiring additional user action for no good reason.

What's your alternative - group/code all the messages and ask the user to provide a list of those that should be suppressed like in lints?

rchl: In this case the error is a direct response to user action so it's fine to present a dialog that can be dismissed easily with a single button press.

If you don't like dialogs so much then maybe just create a plugin that overrides `sublime.error_message` (if that's possible but I think it should)? It's not sensible to expect every package to introduce an option to switch between error dialogs and status field messages.

For some cases I would also prefer to not have dialogs but as discussed in related conversation, there is not necessarily a better way due to ST's limitations.

eugenesvk: > In this case the error is a direct response to user action

Showing a status message is also a direct response

>  so it's fine to present a dialog that can be dismissed easily with a single button press.

It's not fine, you need to provide value to the user to demand action from every single user (this is an **optional** config, disabled by default, so most of the users will continue to suffer through needless actions the way you want them), and in this case you don't - this is simply not an important error message.

Also for some of these cases a blocking message prevents a much better UI: for example, in the next error that appears after the`parse VSCode package JSON` action instead of a blocking `URL must end with 'package.json'"` it'd be better to show the same input field again with the old user value and show a non-blocking error message (e.g., in the status bar or maybe in some popup) so you don't lose valuable user input data and allow for the next user action to be error correction instead of UI unblocking

> If you don't like dialogs so much then maybe just create a plugin that overrides sublime.error_message (if that's possible but I think it should)?

I don't like needless user actionn, not dialogs, so I don't want to override all the error messages, just the ones that aren't important. None in your plugin are, so I'd be fine with overriding all of them.

> It's not sensible to expect every package to introduce an option to switch between error dialogs and status field messages.

I don't expect that, there are plenty of plugins that don't show dialogs, or use less intrusive notifications, and this is a PR for a single package optionally supressing unimportant error messages, not for Sublime core changing everything

jwortmann:
Here is my opinion:

- I would probably be okay with such a change to remove (most of) the dialog windows, which don't require user input.
- *if* we want to make this configurable, it should be a single setting and use a better name. Maybe something like `suppress_error_dialogs`. No need to distinguish between `message_dialog` and `error_dialog` imo. But I would probably favor not to add a setting at all for this, unless you want to implement a "don't show this dialog again" checkbox/button.
- the status text should give a reference to the console, like "Failed to start LSP-json - see console for details"
- status message should use the `Window.status_message` method if possible, not `sublime.status_message`. And there are likely a few more things to clean up in the code, e.g. don't use local imports. But I haven't looked too closely yet.
- I don't really like emojis as ⚠️ in the status bar or console
- I found the package for notifications: https://facelessuser.github.io/SubNotify/ but unfortunately it is a separate package and not a dependency. While it could be implemented with a fallback (https://facelessuser.github.io/SubNotify/usage/#tips-and-tricks-for-developers), probably no user has that package installed, so it seems more or less useless to use as it is now. Probably out of scope for this PR, but maybe it could be converted to a regular dependency?

jwortmann:
> **Question**: is it possible to make that dependency respect config options from this plugin? Or do dependencies "by design" ignore their clients?

Yes, should be possible. Just load settings via `sublime.load_settings('LSP.sublime-settings')`.

LSP.sublime-settings
```diff
@@ -49,8 +49,7 @@
   // Show errors and warnings count in the status bar
   "show_diagnostics_count_in_view_status": false,
 
-  "show_ui_blocking_message": true, // Show UI blocking modals or the full console message and short status notification
-  "show_ui_blocking_err_message": true, // Show UI blocking error modals or the full console message and short status notification
+  "suppress_error_dialogs": false, // Replace UI blocking modals with the full console message and a short status notification
```
rchl: - Follow conventions in the file (the way the comment is placed)
- Find a better place for the option - right now it's inserted in the middle of diagnostics-related options.

plugin/code_actions.py
```diff
@@ -352,7 +353,8 @@ def run_async() -> None:
 
     def _handle_response_async(self, session_name: str, response: Any) -> None:
         if isinstance(response, Error):
-            sublime.error_message(f"{session_name}: {str(response)}")
+            msg = f"{session_name}: {str(response)}"
+            notify_err(msg, msg)
```
rchl, line 357: Don't use abbreviations like `err`, `msg`.

eugenesvk:
Updated the PR per the following:

> * I would probably be okay with such a change to remove (most of) the dialog windows, which don't require user input.

switched to suppress by default

> * _if_ we want to make this configurable, it should be a single setting and use a better name. Maybe something like `suppress_error_dialogs`. No need to distinguish between `message_dialog` and `error_dialog` imo. But I would probably favor not to add a setting at all for this, unless you want to implement a "don't show this dialog again" checkbox/button.

Left the setting given that it's suppressed by default, also re the extra button: Sublime still hasn't learned to properly roundtrip user configs, and this button would change files in the User folder, thus breaking subtle formatting, so I'd just leave it in the currently less convenient format

> * the status text should give a reference to the console, like "Failed to start LSP-json - see console for details"

Added a shorter text, otherwise it might not fit

> * status message should use the `Window.status_message` method if possible, not `sublime.status_message`. And there are likely a few more things to clean up in the code, e.g. don't use local imports. But I haven't looked too closely yet.

switched

> * I don't really like emojis as ⚠️ in the status bar or console

They address @rchl concert of users not noticing notifications, ⚠️ make them stand out. I don't mind either way, so removed ⚠️, but left ❗ as I presume those are supposed to still be differentiated somehow

rchl:
> > * I don't really like emojis as ⚠️ in the status bar or console
> 
> They address @rchl concert of users not noticing notifications, ⚠️ make them stand out. I don't mind either way, so removed ⚠️, but left ❗ as I presume those are supposed to still be differentiated somehow

I do think that there will be a good portion of users that will not notice something showing up in the status field for a few seconds. Or if they do notice something changing, they might not realize what has changed (if the status field is busy).

So as I said, I don't think that status messages are a good solution in general.

I suppose when user is doing an explicit action then he/she might be more looking for some feedback/result and might pay more attention.

If the user doesn't expect anything then it might be easy to miss status change completely.

LSP.sublime-settings
eugenesvk: moved to others

plugin/code_actions.py
eugenesvk: renamed

eugenesvk:
After checking the manually shortened messages, there is actually no ❗ left, only in the short template when the only thing you see is `❗LSP: see console log…` instead of any substatntive message, and since you both don't like them won't adde them to the actual status error messages

> I do think that there will be a good portion of users that will not notice something showing up in the status field for a few seconds

Have you actually tried it? It's harder to miss colored ❗ motion in your vision periphery

> So as I said, I don't think that status messages are a good solution in general.

By the way, for very important messages you could also use these popups, that's in user's focus (at the cursor) (and of course it could fit the whole message)

not

plugin/tooling.py
```diff
@@ -130,15 +132,19 @@ def run(self, base_package_name: str) -> None:
...
-            sublime.error_message("URL must end with 'package.json'")
+            message = "URL must end with 'package.json'"
+            notify_error(sublime.active_window(), message, message)
             return
         try:
             package = json.loads(urllib.request.urlopen(base_url).read().decode("utf-8"))
         except Exception as ex:
-            sublime.error_message(f'Unable to load "{base_url}": {ex}')
+            msg = f'Unable to load "{base_url}": {ex}'
+            notify_error(sublime.active_window(), message, message)
```
rchl, line 147: Get window from `self.view`

plugin/core/logging.py
```diff
@@ -39,3 +39,33 @@ def exception_log(message: str, ex: Exception) -> None:
...
+        if window:
+            window.status_message(status)  # print short message to statusbar
+        else:
+            sublime.status_message(status)
+        print(message)  # print full message to console log
+    else:
+        sublime.message_dialog(message)
+
+
+def notify_error(window: Optional[sublime.Window], message: str, status: str = '❗LSP: see console log…') -> None:
```
jwortmann, line 59: I think we don't need 2 separate functions. Current use of `sublime.mesage_dialog` and `sublime.error_message` looks pretty arbitrary. I think all of them are errors, so I'd use only `sublime.error_message`. If the severity is important it could be another parameter `severety`, but I don't think that's necessary. `notify_error` sounds like the appropriate function name, so I would remove the other function.

plugin/core/logging.py
```diff
@@ -39,3 +39,33 @@ def exception_log(message: str, ex: Exception) -> None:
...
+def notify(window: Optional[sublime.Window], message: str, status: str = 'LSP: see console log…') -> None:
+    """Pick either of the 2 ways to show a user notification message:
+      - via a detailed console message and a short status message
+      - via a blocking modal dialog"""
+    from .settings import userprefs
+    if userprefs().suppress_error_dialogs:
+        if window:
+            window.status_message(status)  # print short message to statusbar
+        else:
+            sublime.status_message(status)
```
jwortmann, line 53: I assume you used this to satisfy the type checker. I believe window = `None` is only a theoretical case, and if there is no (active) window, I think we don't need to log anyway. So you can just put
```
if not window:
    return
```
at the top of the function.

plugin/core/windows.py
```diff
@@ -295,12 +296,13 @@ def start_async(self, config: ClientConfig, initiating_view: sublime.View) -> No
                 "Re-enable by running \"LSP: Enable Language Server In Project\" from the Command Palette.",
                 "\n\n--- Error: ---\n{1}"
             )).format(config.name, str(e))
+            status = f"LSP: Failed to start {config.name}… See console"
```
jwortmann, line 299: Do we need the `LSP:` prefix in the status messages? It should be obvious if there is either the `config.name` or when it was something directly triggered by the user.

plugin/core/logging.py
```diff
@@ -39,3 +39,33 @@ def exception_log(message: str, ex: Exception) -> None:
...
+
+
+def notify(window: Optional[sublime.Window], message: str, status: str = 'LSP: see console log…') -> None:
+    """Pick either of the 2 ways to show a user notification message:
+      - via a detailed console message and a short status message
+      - via a blocking modal dialog"""
+    from .settings import userprefs
+    if userprefs().suppress_error_dialogs:
+        if window:
+            window.status_message(status)  # print short message to statusbar
```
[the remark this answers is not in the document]
eugenesvk: removed, but the non-obvious part was to differentiate status_message vs print (short vs full)

plugin/core/logging.py
```diff
@@ -39,3 +39,33 @@ def exception_log(message: str, ex: Exception) -> None:
 def printf(*args: Any, prefix: str = 'LSP') -> None:
     """Print args to the console, prefixed by the plugin name."""
     print(prefix + ":", *args)
+
+
+def notify(window: Optional[sublime.Window], message: str, status: str = 'LSP: see console log…') -> None:
```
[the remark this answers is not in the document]
eugenesvk: renamed, but the defaults are for those future cases

plugin/core/logging.py
eugenesvk: done

plugin/core/windows.py
eugenesvk: ok, there is also the bigger console message with more detailed errors to fall back to

plugin/core/logging.py
```diff
@@ -39,3 +39,33 @@ def exception_log(message: str, ex: Exception) -> None:
...
+    """Pick either of the 2 ways to show a user notification message:
+      - via a detailed console message and a short status message
+      - via a blocking modal dialog"""
+    from .settings import userprefs
```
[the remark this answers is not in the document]
eugenesvk: yes, circular import was the reason, moved these to a separate file

plugin/core/logging.py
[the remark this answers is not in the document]
jwortmann: When it's used in the future, I'd expect that you still pass a message to this function and don't use that default text.

How about `status_message: str | None = None` and then
```python
if status_message is None:
    status_message = message
```
because a few times in the code you used the same console and status message.

Also I think now that for the console message we should add the `LSP:` prefix, to make it easy to see which lines come from LSP and because the space is not so limited in the console.

plugin/core/logging_notify.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import annotations
+import sublime
+from .settings import userprefs
+
+
+def notify_error(window: sublime.Window | None, message: str, status_message: str = 'LSP: see console log…',
+                 severe: bool = True) -> None:
```
jwortmann, line 7: You never used that argument in the function calls, so I think we don't need that and it can be removed. The difference between `sublime.error_message` and `sublime.message_dialog` is just an additional icon anyways (at least on Windows), so it's probably not very important.

plugin/core/logging.py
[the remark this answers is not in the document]
eugenesvk: ok, added print prefix, deduped and made it optional

plugin/core/logging_notify.py
eugenesvk: k, removed

I've never used it because you said that all current messages are the same, so they all got bumped to "severe" in the process of converting to a single function.

I haven't reviewed all the messages to see whether a different type is justified. This again was more future-proof in case some messages will be deemed less severe

jwortmann: Yeah but if we need it in the future, we can simply add it then. I'm not gonna say that I have never done this myself, but in general it's probably not very useful to add unused code with the reasoning that it could potentionally be used in the future.

sublime-package.json
```diff
@@ -449,6 +449,11 @@
               "default": false,
               "markdownDescription": "Show errors and warnings count in the status bar."
             },
+            "suppress_error_dialogs": {
+              "type": "boolean",
+              "default": false,
```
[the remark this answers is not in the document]
eugenesvk: fixed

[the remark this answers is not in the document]
eugenesvk: >  I would just print the messages to the status bar and console.

So let's revert it back to true then here and in the settings

eugenesvk:
> Perhaps as a compromise the error messages from the tooling / troubleshoot command

But user configuration is that compromise. I, for one, would never want to see these as blocking either, showing informational error messages doesn't warrant this. And this package then can define whatever compromising defaults you want, it won't affect others permanently

Besides this goes back to the 2-tier error message system that was just removed, where you had the flexibility of suppressing dialogs of only 1 type, both at the level of the plugin, and individual user

PR #2547, opened 2024-11-06
Add commands to open user config/keybinds only

eugenesvk:
The default Sublime way to open settings side-by-side is rather unergonomic as it wastes a lot of space, so it's sometimes easier to copy/paste/comment then edit the settings you want.

However, there is no command to open your user settings later without the defaults.

This PR adds such commands

### DIFF
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -8,12 +8,28 @@
         }
     },
     {
+        "caption": "Preferences: LSP Settings User",
+        "command": "open_file",
+        "args": {
+            "file": "${packages}/User/LSP.sublime-settings",
+            "contents": "// Settings in here override those in \"LSP/LSP.sublime-settings\"\n{\n\t$0\n}\n"
+        }
+    },
+    {
         "caption": "Preferences: LSP Key Bindings",
         "command": "edit_settings",
         "args": {
             "base_file": "${packages}/LSP/Default.sublime-keymap",
             "user_file": "${packages}/User/Default ($platform).sublime-keymap",
             "default": "[\n\t$0\n]\n",
+        }
+    },
+    {
+        "caption": "Preferences: LSP Key Bindings User",
+        "command": "open_file",
+        "args": {
+            "file": "${packages}/User/Default ($platform).sublime-keymap",
+            "contents": "[\n\t$0\n]\n",
         }
     },
     {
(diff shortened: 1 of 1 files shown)


rchl:
Not a fan of that. Means that those new settings will get in the way when trying to find the original one (which is IMO more practical for most people).

Also, it's a common convention in new packages and ST itself to open settings side-by-side.

I think you would have to just extend your personal commands with those.

eugenesvk:
The User settings are always sorted 2nd to the main side-by-side setting, so don't stand in the way. But you could move `User` to the begginning so that it's not interspersed with the defaults, but grouped separately

And yes, it's a common mistake packages make following Sublime. The side-by-side view could justify itself over the much more common tabbed navigation if the setting values "synced" in view, e.g., if you edit `show_code_lens` in user config, the default config would jump to `show_code_lens` to show you the options

jwortmann:
I agree to what @rchl wrote. The side by side settings is a standard for a very long time now, and I don't think there is any other package which provides the option to additionally show only the user settings if it already has the side by side option.

Just copy those changes into your user commands file and you should be fine.

That "synched" view would be something you would need to propose to SublimeHQ, for it to be implemented directly in ST. I doubt that they will do it, though.

jfcherng:
Just want to say that, in general, I consider copying everything to your own settings is not a good practice. 

1. From the plugin's viewpoint, you use nothing default. You override everything with your customized values, which are coincidentally the same value with defaults defined by the plugin, at that moment. But current defaults may not be future defaults. 
2. Due to `0.`, if a value is a list/array and the plugin author changes (add/remove/modify) its element, you will still use the old value.
3. You very likely don't know what you have customized.

rchl:
> The User settings are always sorted 2nd to the main side-by-side setting, so don't stand in the way. But you could move `User` to the begginning so that it's not interspersed with the defaults, but grouped separately

With the fuzzy filtering and ST's sorting algorithm you can't state any of that for sure. There are many factors that go into that.

eugenesvk:
> With the fuzzy filtering and ST's sorting algorithm you can't state any of that for sure. There are many factors that go into that.

It's not some deep mystery either (and why would it be in such a short text? What would the actual query be to make the last word rank higher?), and I've been using this for a while in other plugins (in the current form with Default/User at the end), it's never been an issue

eugenesvk:
> 1. From the plugin's viewpoint, you use nothing default.

You ignored the `comment` part, so this is simply not true as is the rest. You only use the values that you change

eugenesvk:
> The side by side settings is a standard for a very long time now

Sure, defaults matter, and bad defaults can persist for generations. Still, this doesn't remove those

> any other package which provides the option to additionally show only the user settings if it already has the side by side option.

There are plugins that offer such an option (separating Default and User), but mostly they tuck it in the Settings menu, where it takes 5(!) clicks to reach! That's why I just move it to the more convenient searchable place

And I don't need the synced view myself, it's still inferior to a copy/paste/edit in 1 fullscreen file or help via LSP, so wouldn't propose it

jfcherng:
> There are plugins that offer such an option (separating Default and User), but mostly they tuck it in the Settings menu, where it takes 5(!) clicks to reach!

Fwiw... most likely old plugins

jwortmann:
> There are plugins that offer such an option (separating Default and User), but mostly they tuck it in the Settings menu, where it takes 5(!) clicks to reach!

Seperate Default & User is different to what I wrote above (having *both* side-by-side and also single user settings itmes). If you know of any package which is registered on Package Control and which does that (similar to this PR), then please provide a link because otherwise I don't want to believe there exists any package doing such a thing.

---

@jfcherng 

> Just want to say that, in general, I consider copying everything to your own settings a bad practice.
> 
>     1. From the plugin's viewpoint, you use nothing default. You override everything with your customized values, which are coincidentally the same value with defaults defined by the plugin, at that moment. But current defaults may not be future defaults.
> 
>     2. Due to `1.`, if a value is a list/array and the plugin author changes (add/remove/modify) its elements, you will still use the old value.
> 
>     3. You very likely don't know what you have customized, one month later.

I'm not sure if you wrote this with reference to my comment above

> Just copy those changes into your user commands file and you should be fine.

but if so, I don't see why those three points would have any relevance here. If you click on "files" in this PR you see that it only provides new command palette entries for the built-in `open_file` command. This should be absolutely safe to add to the user commands file, regardless of if the package settings change later or not.

If unrelated, it still shouldn't really matter much for the LSP package, because there is code to ensure backwards compatibility with old types of setting values:
https://github.com/sublimelsp/LSP/blob/1d2e995983491d073997315299848f128c76968e/plugin/core/types.py#L288-L325

Issue #2520, opened 2024-09-23
Sublime-LSP Local Code-Assistant Support


fruffy:
There is currently no first-class code local assistant plugin for Sublime-LSP. [Github Copilot](https://github.com/TerminalFi/LSP-copilot) and [Codeium](https://codeium.com/sublime_tutorial) plugins exist but they require a consistent internet connection and use proprietary models. 

It would be nice to have good support for a local model. There seems to be some interest in integrating tabby with Sublime-LSP: https://github.com/TabbyML/tabby/issues/219

Unfortunately, I do not know how to build such a plugin. How easy is it to add support or integrate SublimeLSP with a code assistant?

predragnikolic:
Is your request is too have a LSP-tabby package?
Does tabby support the LSP spec?

Not so related, but have you maybe seen https://github.com/yaroslavyaroslav/OpenAI-sublime-text ? I guess OpenAi package doesn't have ghost text completion items. But it does support having a local model.

fruffy:
> Is your request is to have a LSP-tabby package?

Yes, tabby could be a good target. It looks like they have LSP support https://github.com/TabbyML/tabby/tree/main/clients/example-vscode-lsp

>Not so related, but have you maybe seen https://github.com/yaroslavyaroslav/OpenAI-sublime-text ? I guess OpenAi package doesn't have ghost text completion items. But it does support having a local model.

Yes, I have looked at this repository and I think it is neat. But as you said it doesn't offer the ghost text completion, which imho is a huge value-add.

jwortmann:
It seems like the LSP server is implemented via [tabby-agent](https://github.com/TabbyML/tabby/tree/main/clients/tabby-agent) (https://www.npmjs.com/package/tabby-agent) and it is a server (the first one?) with support for `textDocument/inlineCompletion` 🚀 

This client currently doesn't have support for inline completions ("ghost text completion"), but it would be interesting to see how it could be implemented and how the UI for that could work (but regular `textDocument/completion` is apparently supported by the server too).

So I tried to set up the server on a Windows PC for testing
1. From https://github.com/TabbyML/tabby/releases downloaded and extracted tabby_x86_64-windows-msvc.zip
2. Downloaded model: `.\tabby.exe download --model StarCoder-1B` (1.3 GB, also see https://tabby.tabbyml.com/docs/models/)
3. Installed tabby-agent: `npm install -g tabby-agent`
4. Added config in `LSP.sublime-settings`, e.g.
[...]

PR #2552, opened 2024-11-11
Add support for inline completions (WIP)

jwortmann:
Experimental / work in progress

Implementation details are open for discussion and it would be nice to gather some opinions how inline completions should work. 

- For now in this PR they are only requested when you manually trigger autocomplete. And then they are immediately hidden when you either type something or move the caret. It would be nice to show the inline completions while typing, but I think the phantom behavior is too buggy for that (https://github.com/sublimehq/sublime_text/issues/5536). Another problem is that the regular completions popup often hides a big part of the inline completion phantom. But that is independent of whether it would be on typing or on manual trigger.
- The server can return multiple inline completion items, but currently only the first one is shown and the other ones are ignored. Maybe there should be some way to switch between items, like in the signature help popup.
- Maybe add a setting to configure the display mode; full or only single line (see https://github.com/sublimelsp/LSP/issues/2520#issuecomment-2439625978)?
- We can't provide [SelectedCompletionInfo](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#selectedCompletionInfo) because it is not available from ST.

It can be tested with Tabby; for setup see #2520 or the docs description which is added in this PR. Tabby requires a compatible GPU.

For this PR I hijacked the <kbd>Alt</kbd> + <kbd>Enter</kbd> key binding which is currently used for "insert completion with opposite insert mode", because there aren't really any good key bindings left (everything with <kbd>Ctrl</kbd> or <kbd>Shift</kbd> is already taken in the ST default key bindings). This was mostly for testing and is up for discussion I guess.

I haven't really looked at the implementations in e.g. LSP-copilot yet. Perhaps there are some improvements that we can adopt.

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -9,13 +9,19 @@
     //     "args": {"overlay": "command_palette", "text": "LSP: "}
     // },
     // Insert/Replace Completions
+    // {
+    //     "keys": ["UNBOUND"],
+    //     "command": "lsp_commit_completion_with_opposite_insert_mode",
+    //     "context": [
+    //         {"key": "lsp.session_with_capability", "operand": "completionProvider"},
+    //         {"key": "auto_complete_visible"}
+    //     ]
+    // },
+    // Insert Inline Completion
     {
         "keys": ["alt+enter"],
-        "command": "lsp_commit_completion_with_opposite_insert_mode",
-        "context": [
-            {"key": "lsp.session_with_capability", "operand": "completionProvider"},
-            {"key": "auto_complete_visible"}
-        ]
+        "command": "lsp_commit_inline_completion",
+        "context": [{"key": "lsp.inline_completion_visible"}]
     },
     // Save all open files that have a language server attached with lsp_save
     // {
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -23,6 +23,12 @@
         "command": "lsp_commit_inline_completion",
         "context": [{"key": "lsp.inline_completion_visible"}]
     },
+    // Show next Inline Completion
+    // {
+    //     "keys": ["UNBOUND"],
+    //     "command": "lsp_next_inline_completion",
+    //     "context": [{"key": "lsp.inline_completion_visible"}]
+    // },
     // Save all open files that have a language server attached with lsp_save
     // {
     //     "keys": ["UNBOUND"],
--- a/boot.py
+++ b/boot.py
@@ -126,6 +126,7 @@ __all__ = (
     "LspInlayHintClickCommand",
     "LspInlineCompletionCommand",
     "LspNextDiagnosticCommand",
+    "LspNextInlineCompletionCommand",
     "LspOnDoubleClickCommand",
     "LspOpenLinkCommand",
     "LspOpenLocationCommand",
--- a/boot.py
+++ b/boot.py
@@ -61,6 +61,7 @@ from .plugin.inlay_hint import LspInlayHintClickCommand
 from .plugin.inlay_hint import LspToggleInlayHintsCommand
 from .plugin.inline_completion import LspCommitInlineCompletionCommand
 from .plugin.inline_completion import LspInlineCompletionCommand
+from .plugin.inline_completion import LspNextInlineCompletionCommand
 from .plugin.panels import LspClearLogPanelCommand
 from .plugin.panels import LspClearPanelCommand
 from .plugin.panels import LspShowDiagnosticsPanelCommand
--- a/boot.py
+++ b/boot.py
@@ -59,6 +59,8 @@ from .plugin.hover import LspHoverCommand
 from .plugin.hover import LspToggleHoverPopupsCommand
 from .plugin.inlay_hint import LspInlayHintClickCommand
 from .plugin.inlay_hint import LspToggleInlayHintsCommand
+from .plugin.inline_completion import LspCommitInlineCompletionCommand
+from .plugin.inline_completion import LspInlineCompletionCommand
 from .plugin.panels import LspClearLogPanelCommand
 from .plugin.panels import LspClearPanelCommand
 from .plugin.panels import LspShowDiagnosticsPanelCommand
@@ -99,6 +101,7 @@ __all__ = (
     "LspCollapseTreeItemCommand",
     "LspColorPresentationCommand",
     "LspCommitCompletionWithOppositeInsertMode",
+    "LspCommitInlineCompletionCommand",
     "LspCopyToClipboardFromBase64Command",
     "LspDisableLanguageServerGloballyCommand",
     "LspDisableLanguageServerInProjectCommand",
@@ -120,6 +123,7 @@ __all__ = (
     "LspHierarchyToggleCommand",
     "LspHoverCommand",
     "LspInlayHintClickCommand",
+    "LspInlineCompletionCommand",
     "LspNextDiagnosticCommand",
     "LspOnDoubleClickCommand",
     "LspOpenLinkCommand",
--- a/docs/src/features.md
+++ b/docs/src/features.md
@@ -174,7 +174,10 @@ Inlay hints are disabled by default and can be enabled with the `"show_inlay_hin
 
 Inline completions are typically provided by an AI code assistant.
 They can span multiple lines and are rendered directly in the source code as grayed out text ("ghost text").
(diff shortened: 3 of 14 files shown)


plugin/inline_completion.py
```diff
@@ -0,0 +1,215 @@
...
+        sublime.set_timeout_async(partial(self._run_async, event, point))
+
+    def _run_async(self, event: dict | None = None, point: int | None = None) -> None:
+        position = get_position(self.view, event, point)
+        if position is None:
+            return
+        session = self.best_session(self.capability, point)
+        if not session:
+            return
+        if self.view.settings().get('mini_auto_complete', False):
```
predragnikolic, line 111: hahahaha I was wondering why this PR doesn't work for me, it was due to this check.
I had `mini_auto_complete` enabled.

Default.sublime-keymap
```diff
@@ -9,14 +9,26 @@
...
+    //     "keys": ["UNBOUND"],
+    //     "command": "lsp_commit_completion_with_opposite_insert_mode",
+    //     "context": [
+    //         {"key": "lsp.session_with_capability", "operand": "completionProvider"},
+    //         {"key": "auto_complete_visible"}
+    //     ]
+    // },
+    // Insert Inline Completion
     {
         "keys": ["alt+enter"],
```
predragnikolic, line 22: What do you think of using `tab` for committing an inline completion?

docs/src/language_servers.md
```diff
@@ -12,6 +12,69 @@ If there are no setup steps for a language server on this page, but a [language
...
+[Tabby](https://tabby.tabbyml.com/) is a self-hosted AI coding assistant which can provide inline completions for [various programming languages](https://tabby.tabbyml.com/docs/references/programming-languages/).
+
+In order to use Tabby you need a sufficiently fast GPU; the CPU version which can also be downloaded from the GitHub releases page is much too slow and it will result in timeouts for the completion requests.
+Alternatively, Tabby can be setup on a separate server with capable hardware; see the [Configuration docs](https://tabby.tabbyml.com/docs/extensions/configurations/) for the required configuration details.
+The following steps describe a local installation on a Windows PC with compatible Nvidia GPU. More installation methods and the steps for other operation systems are listed in the [Tabby docs](https://tabby.tabbyml.com/docs/quick-start/installation/docker/).
+
+1. Download and install the CUDA Toolkit from <https://developer.nvidia.com/cuda-downloads>
+
+2. Download and extract a CUDA version of Tabby from the [GitHub releases page](https://github.com/TabbyML/tabby/releases) (click on "Assets"); e.g. `tabby_x86_64-windows-msvc-cuda122.zip`
+
```
predragnikolic, line 28: There is an additional step here.
Extract the zip, and change the access permissions of `llama-server` and `tabby` to be executables:
```
$  tabby_aarch64-apple-darwin ls
llama-server tabby
$ tabby_aarch64-apple-darwin chmod +x llama-server
$  tabby_aarch64-apple-darwin chmod +x tabby
```

> [!NOTE] 
> On MacOS you might get an error `“tabby” cannot be opened because it is from an unidentified developer.`
> After changing the permission to executable, right click on `tabby` and select `Open`, that will get rid of the error.

predragnikolic:
> We can't provide [SelectedCompletionInfo](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#selectedCompletionInfo) because it is not available from ST.

This is true.

plugin/inline_completion.py
predragnikolic: I am thinking if this should be documented.

docs/src/features.md
```diff
@@ -170,6 +170,12 @@ Inlay hints are disabled by default and can be enabled with the `"show_inlay_hin
 
 !!! info "Some servers require additional settings to be enabled in order to show inlay hints."
 
+## Inline Completions
+
+Inline completions are typically provided by an AI code assistant.
+They can span multiple lines and are rendered directly in the source code as grayed out text ("ghost text").
+Currently inline completions are only requested when you manually trigger auto-completions (<kbd>Ctrl</kbd> + <kbd>Space</kbd>).
+
```
predragnikolic, line 178: I would suggest adding here that Inline completions will not be triggered if `mini_auto_complete` is enabled.

plugin/inline_completion.py
```diff
@@ -0,0 +1,215 @@
...
+            range_ = item.get('range')
+            region = range_to_region(range_, self.view) if range_ else sublime.Region(position)
+            region_length = len(region)
+            if region_length > len(insert_text):
+                continue
+            listener.inline_completion.items.append((session_name, region, insert_text, item.get('command')))
+        listener.inline_completion.position = position
+        listener.inline_completion.render_async(0)
+
+        # filter_text = item.get('filterText', insert_text)  # ignored for now
```
predragnikolic, line 160: 👀

plugin/inline_completion.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def is_enabled(self, event: dict | None = None, point: int | None = None) -> bool:
+        if not super().is_enabled(event, point):
+            return False
+        listener = self.get_listener()
+        if not listener:
+            return False
+        return listener.inline_completion.visible
+
+    def run(self, edit: sublime.Edit, event: dict | None = None, point: int | None = None) -> None:
```
predragnikolic, line 200: point is unused

boot.py
```diff
@@ -120,7 +124,9 @@
     "LspHierarchyToggleCommand",
     "LspHoverCommand",
     "LspInlayHintClickCommand",
+    "LspInlineCompletionCommand",
     "LspNextDiagnosticCommand",
+    "LspNextInlineCompletionCommand",
```
predragnikolic, line 129: I am not sure how to test the `LspNextInlineCompletionCommand`

predragnikolic: Currently I trigger the autocomple,
that also triggers the inline completions request,
and I will see the inline completion phantom,
When I trigger the `LspNextInlineCompletionCommand`, I expect to see a different inline completion phantom, but I still see the same phantom. Do you see a different phantom when triggering the `LspNextInlineCompletionCommand`?

jwortmann: In my experiments, Tabby always provides only a single inline completion item. So it's expected that currently nothing happens when you trigger that command.

Default.sublime-keymap
jwortmann: I'm aware that <kbd>Tab</kbd> is used for it in VSCode and I think also in other IDEs like IntelliJ. But I doubt that it would be a good choice in Sublime; personally I use <kbd>Tab</kbd> to commit regular auto-complete (`"auto_complete_commit_on_tab": true`) and there are probably many users who do the same. I think we should use a key binding which is definitely not already in use by Sublime Text. However <kbd>Alt + Enter</kbd> is not the most ergonomic to use though; here I used it mostly for testing but we can discuss whether we find something better. Or maybe it should be left without any key binding by default and require users to set up an own key binding for it.

predragnikolic: there are so many use cases where I would (ab)use a `tab`, but I agree... Each time a cool feature shows up and it has to do with autocomplete, `tab` is the first shortcut that I will head to. Obviously a `tab` can't cover every possible use case.

docs/src/language_servers.md
```diff
@@ -12,6 +12,79 @@ If there are no setup steps for a language server on this page, but a [language
...
+3. On macOS and Linux it might be necessary to change the access permissions of `lama-server` and `tabby` to be executable:
+
+    ```sh
+    $ tabby_aarch64-apple-darwin chmod +x llama-server
+    $ tabby_aarch64-apple-darwin chmod +x tabby
```
predragnikolic, line 33:
```suggestion
    $ chmod +x llama-server
    $ chmod +x tabby
```

plugin/inline_completion.py
predragnikolic: this was addressed with [3063b62](https://github.com/sublimelsp/LSP/pull/2552/commits/3063b628d698339d3641d029caf64fbb2bd6d5dd)

jwortmann:
I think we would need first-class support from Sublime Text to handle and render inline completions properly (instead of the workaround with phantoms), so I'll close this old PR for now. Maybe we can revisit it when the inline completion capability gets supported by more servers, or if ST adds better API functionality for this.

PR #2563, opened 2024-11-27
Set completions from the async thread

jwortmann:
According to https://github.com/sublimehq/sublime_text/issues/6249#issuecomment-2502804237 this should prevent lag/freezing if there is an excessive amount of completion items.

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -7,6 +7,7 @@ from .core.constants import DOCUMENT_HIGHLIGHT_KIND_NAMES
 from .core.constants import DOCUMENT_HIGHLIGHT_KIND_SCOPES
 from .core.constants import HOVER_ENABLED_KEY
 from .core.constants import RegionKey
+from .core.constants import ST_VERSION
 from .core.logging import debug
 from .core.open import open_in_browser
 from .core.panels import PanelName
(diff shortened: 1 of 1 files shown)


rchl (approved):

predragnikolic:
Here is how I tested this.
I cloned the typescript-language-server.
And I have changed this code:
```diff
    async completion(params: lsp.CompletionParams, token?: lsp.CancellationToken): Promise<lsp.CompletionList | null> {
        
        const completions = asCompletionItems(entries, this.completionDataCache, filepath, params.position, document, this.tsClient, completionOptions, this.features, completionContext);
+        // added 100000 on top of typescript the returned typescript completion items, to mock a really huge response
+        for (let i = 0; i < 100000; i++) { 
+          completions.push({
+            ...completions[0],
+            label: `Item ${i}`
+          })
+        }
        return lsp.CompletionList.create(completions, isIncomplete);
    }
```

Then I compared built the server.
And tested with
 - ST build 4183
 - ST build 4184
 - ST build 4184 with this PR.

The conclusion is that there is still a noticeable delay when typing.

https://github.com/user-attachments/assets/29f0f3f0-89cc-4dc4-8d76-654ece8d52ae

https://github.com/user-attachments/assets/b4167c8d-348b-4904-9c2c-fd7b650973d5

https://github.com/user-attachments/assets/ea95d97c-c27f-4b11-abdc-3a33670f12fd

predragnikolic:
I also tested this ->
https://github.com/sublimehq/sublime_text/issues/6249
There seems to be an noticeable improvement in a good way, but I can still notice lag.

predragnikolic:
Here is one theory.
If a server returns 10mb of data,
this line will not be the bottleneck:
```
body = reader.read(int(headers.get("Content-Length")))
```
but this one will:
```
return orjson.dumps(data)
```

`orjson` is fast, but in my discoveries the bottleneck on why the lag exists.

I think the issue is similar to this article "Python’s memory-inefficient JSON loading"
https://pythonspeed.com/articles/json-memory-streaming/

predragnikolic:
Here is the minimal plugin code to reproduce the issue. 
[Advent of Code - with a bit of an LSP issue.zip](https://github.com/user-attachments/files/17966962/Advent.of.Code.-.with.a.bit.of.an.LSP.issue.zip)

jwortmann:
My assumption was (is) that the UI thread and the async thread from ST are created in C++. So they are "real" threads and don't block each other.

rchl:
Not sure. Those still have to run within the same python environment so I would think that GIL would still apply.

Also, if async thread is blocked then the `*_async` event handles can't fire and that can also cause some form of blocking (not in a complete UI block way but certain actions can become unresponsive).

rchl:
This looks good, I would say.
Any potential improvements should be handled in a separate issue.

PR #2567, opened 2024-12-04
Replace Quick Fix label in hover popup with lightbulb icon

jwortmann:
Currently there is a "Quick Fix" prefix for when code actions are shown in the hover popup.

This is a suggestion to replace that label with the lightbulb icon, because
- it saves a little bit of space and subjectively looks nicer
- we filter out non-quickfix actions for the popups, but some servers omit the code action kind and then there might still be actions included which are no quickfixes. I think the lightbulb symbol is more generic and fits better in that case

I used the orange color from Mariana, but I think it should look okay in both dark and light color schemes.

lightbulb


icons/lightbulb_colored@2x.png
jwortmann, line 1: Actually the higher DPI images don't seem to work in minihtml...

plugin/hover.py
```diff
@@ -83,7 +83,13 @@ def link(self, point: int, view: sublime.View) -> str:
 ]
 
 
-def code_actions_content(actions_by_config: list[CodeActionsByConfigName]) -> str:
+LIGHTBULB_ICON_HTML = """
+<span class="lightbulb">
+    <img width="13" height="13" src="res://Packages/LSP/icons/lightbulb_colored.png">
+</span>"""
```
rchl, line 89: - we don't really need to wrap it in `span`
- Use em units so that the icon size better adapts to whatever font size the user is using

```suggestion
    <img src="res://Packages/LSP/icons/lightbulb_colored.png" style="width: 1em; height: 1em">
"""
```

(just an example, you can of course move the styles to css)

jwortmann: I tried to set the style (padding-right) directly on the img element, but then it seems to be ignored. That's probably a limitation/bug in ST minihtml. That was the reason for the span wrapper.

I will update the units.

rchl: Try margin instead

jwortmann: Tried that as well, but without success.

rchl: Since it's an inline (or inline-block) element it should be then enough to create an extra spacing by having a space characters between the icon and the text

jwortmann: Hm, not a big fan of using space characters or `&nbsp` as padding/spacers. I hate when websites do this and you accidentally select them when you want to select some text. And maybe one day ST might make the contents of popups selectable. The CSS styling on the span is also a more flexible solution.

rchl (approved):

PR #2571, opened 2024-12-06
Fix LspToggleInlayHintsCommand not initializing

### LINKED ISSUE #2569: LSP.plugin.inlay_hint.LspToggleInlayHintsCommand init failure
Sublime-LSP fails to init inlay hints on startup

```
  File "C:\Program Files\Sublime Text\Lib\python38\sublime_plugin.py", line 535, in create_window_commands
    o = cls(window)
  File "C:\Users\user\AppData\Roaming\Sublime Text\Installed Packages\LSP.sublime-package\plugin/inlay_hint.py", line 24, in __init__
    window.settings().set('lsp_show_inlay_hints', userprefs().show_inlay_hints)
AttributeError: 'NoneType' object has no attribute 'show_inlay_hints'

The above exception was the direct cause of the following exception:

RuntimeError: unable to instantiate 'LSP.plugin.inlay_hint.LspToggleInlayHintsCommand'

```
- OS: Win10
- Sublime Text version: 4180
- LSP version: 2.3.0

Re-install does not help, reproducible on every startup.

jwortmann:
Should fix #2569 I guess.

### DIFF
--- a/plugin/inlay_hint.py
+++ b/plugin/inlay_hint.py
@@ -22,8 +22,7 @@ class LspToggleInlayHintsCommand(LspWindowCommand):
     def __init__(self, window: sublime.Window) -> None:
         super().__init__(window)
         settings = userprefs()
-        default = settings.show_inlay_hints if settings is not None else False
-        window.settings().set('lsp_show_inlay_hints', default)
+        window.settings().set('lsp_show_inlay_hints', settings.show_inlay_hints if settings is not None else False)
 
     def run(self, enable: bool | None = None) -> None:
         window_settings = self.window.settings()
--- a/plugin/inlay_hint.py
+++ b/plugin/inlay_hint.py
@@ -21,7 +21,9 @@ class LspToggleInlayHintsCommand(LspWindowCommand):
 
     def __init__(self, window: sublime.Window) -> None:
         super().__init__(window)
-        window.settings().set('lsp_show_inlay_hints', userprefs().show_inlay_hints)
+        settings = userprefs()
+        default = settings.show_inlay_hints if settings is not None else False
+        window.settings().set('lsp_show_inlay_hints', default)
 
     def run(self, enable: bool | None = None) -> None:
         window_settings = self.window.settings()
--- a/plugin/inlay_hint.py
+++ b/plugin/inlay_hint.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+from .core.constants import ST_VERSION
 from .core.css import css
 from .core.edit import apply_text_edits
 from .core.protocol import InlayHint
@@ -21,8 +22,16 @@ class LspToggleInlayHintsCommand(LspWindowCommand):
 
     def __init__(self, window: sublime.Window) -> None:
         super().__init__(window)
+        window.settings().setdefault('lsp_show_inlay_hints', self._get_default_value())
+
+    @staticmethod
+    def _get_default_value() -> bool:
         settings = userprefs()
-        window.settings().setdefault('lsp_show_inlay_hints', False if settings is None else settings.show_inlay_hints)
+        if settings is not None:
+            return settings.show_inlay_hints
+        if ST_VERSION >= 4171:  # sublime.load_settings is available at import time
+            return sublime.load_settings('LSP.sublime-settings').get('show_inlay_hints')
+        return False
 
     def run(self, enable: bool | None = None) -> None:
         window_settings = self.window.settings()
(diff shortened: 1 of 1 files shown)


plugin/inlay_hint.py
```diff
@@ -21,7 +21,9 @@ class LspToggleInlayHintsCommand(LspWindowCommand):
 
     def __init__(self, window: sublime.Window) -> None:
         super().__init__(window)
-        window.settings().set('lsp_show_inlay_hints', userprefs().show_inlay_hints)
+        settings = userprefs()
+        default = settings.show_inlay_hints if settings is not None else False
```
jwortmann, line 25: Note that I put a literal `False` here if `userprefs()` is None, because I assume that we cannot use `sublime.load_settings('LSP.sublime-settings')` before `plugin_loaded()`.

This would only affect the very first window(s) which is/are open when Sublime Text starts.

plugin/inlay_hint.py
```diff
@@ -21,7 +21,9 @@ class LspToggleInlayHintsCommand(LspWindowCommand):
...
+        window.settings().set('lsp_show_inlay_hints', default)
```
rchl, line 26: Might not be a new thing but shouldn't it use the already saved setting from window, if set? It seems like it's potentially overriding what is stored there already.

plugin/inlay_hint.py
rchl: It's kinda not great to assume false when it might not be false.
Could we just not store anything on `init` and only do so when the command is run?

plugin/inlay_hint.py
jwortmann: But this is the first place where it gets saved in the window settings, so I think it can't be already set. The only other place where it would be saved in the window settings would be from the `run` method.

rchl: When starting ST the windows are restored including the window settings so it could be stored already.

plugin/inlay_hint.py
jwortmann: Yeah, actually that should work

jwortmann: We could also delay reading the default and setting the window setting via using a `sublime.set_timeout` in `__init__`.

But I'm not sure if it is allowed to use `set_timeout` before `plugin_loaded`

rchl: I'm sure set_timeout is allowed but it doesn't sound very robust.

I don't feel like the case you've mentioned is something that we should need to avoid. It sounds fine to me to follow the global setting until explicitly changing the window setting.

plugin/inlay_hint.py
```diff
@@ -21,7 +21,8 @@ class LspToggleInlayHintsCommand(LspWindowCommand):
 
     def __init__(self, window: sublime.Window) -> None:
         super().__init__(window)
-        window.settings().set('lsp_show_inlay_hints', userprefs().show_inlay_hints)
+        settings = userprefs()
+        window.settings().setdefault('lsp_show_inlay_hints', False if settings is None else settings.show_inlay_hints)
```
rchl, line 25: So this still seems to be setting `False` in the case when settings are `None` (and no previous value is stored). Which can different from actual value from userprefs.

jwortmann: Yes, but it only happens if the command is initialized before `plugin_loaded` and if the window doesn't already have a value stored. I think it's not safe to use API methods before plugin_loaded unless they are explicitly documented to be available on import time. Though iirc there might have been some changes to that in a past ST update (?)

plugin/inlay_hint.py
jwortmann: I think I would not quite like that, because it might appear very inconsistent to users. If you have multiple windows open and change the value in the settings, then it might be toggled in some of the windows but not in all. And since with my latest commit the window setting will be preserved when initializing, the manual toggle could have happened a long time ago and it would still have an effect here.

So I'd say the toggles should either be a global over all windows (I suggested that earlier when it was implemented, but we decided to not do this), or keep the current behavior and use the setting value only for newly created windows. The latter is also explained in the setting description:
https://github.com/sublimelsp/LSP/blob/8b7afad0527400e113dd6cd0f1f63175bcc599c7/LSP.sublime-settings#L176-L178

plugin/inlay_hint.py
jwortmann: I found it in the changelog that from ST 4171 all functions are available at import time (also tested and confirmed).

So now it uses the API method to read from the settings file if the build is new enough: https://github.com/sublimelsp/LSP/pull/2571/commits/7f3da69bffa93d3d38afcf746b6610dd68f86efc

rchl (approved):

PR #2577, opened 2024-12-10
Update url.py to fit WSL path that starts with "\\wsl.local"

95833:


plugin/core/url.py
```diff
@@ -56,7 +56,8 @@ def parse_uri(uri: str) -> tuple[str, str]:
         path = url2pathname(parsed.path)
         if os.name == 'nt':
             netloc = url2pathname(parsed.netloc)
-            path = path.lstrip("\\")
+            if path[0] == '\\':
```
jwortmann, line 59:
```suggestion
            if path.startswith('\\'):
```

jwortmann:
Could you add a test in the `tests/test_url.py` file with an example, i.e. some URL that is not handled correctly at the moment but would be fixed with this pull request?

I see there is already a test for a WSL URL at https://github.com/sublimelsp/LSP/blob/c64ac00a1295fdc8461811033e0974443fd9ea55/tests/test_url.py#L39-L42
but you could add another one after that which would be handled by the change in this pull request.

95833:
Can you teach me how to use tests?

95833:
I don't know how to run this test.

jwortmann:
The CI check on GitHub will run it after you pushed another commit.

95833:
Can't I run it locally? It's a bit annoying to have to push to GitHub every time I test the code.

jwortmann:
To manually run tests locally you need to install the UnitTesting package from package control. See https://github.com/sublimelsp/LSP/blob/main/CONTRIBUTING.md#testing

I haven't done this though, but usually you know what the expected outcome should be, so it should be fine to just push new commits until it works. The commits will be squashed into a single one when the PR gets merged anyway.

95833:
Okay

95833:
To provide support for WSL paths, I think it would be better to update the underlying packages related to path handling rather than changing the program logic; the issue seems to have become somewhat complex.

PR #2579, opened 2024-12-12
add path-mapping function

95833:


rwols:
Why did you close this? It seems useful. Note that it can be hard for maintainers to make time to review every pull request. I will re-open this just in case.

95833:
I am fixing some bugs, so I have temporarily closed it.

rwols:
I have converted the PR status to "draft" then. You can remove the "draft" status when you think it's ready.

95833:
Alright, it should now be usable to some extent. At least on the surface, it seems that no new issues have been introduced.

jwortmann:
Could you describe the use case for this; there isn't any linked issue or explanation in the opening comment, and I don't have any intuition just from the config description
> Map the paths of URIs in the data exchanged between the server and the client.

why I would want to map paths like that. Apparently it has something to do with WSL. I think from the regular contributors I'm the only Windows user, but I've never really used WSL, so if this requires a certain setup then a step by step description would be useful if we want to test this PR.

Ideally this PR would have a test with something that currently doesn't work, but will be fixed by this PR (if this is applicable here).

95833:
I use it in the WSL environment, but it is a general feature. clangd also has a path mapping option, and their functions should be consistent. However, clangd also has bugs and cannot meet my needs. @rwols Should be able to provide more explanation about this feature.
 
In my use case, I code using an editor on Windows and compile it with the toolchain and libraries in WSL.  Since the project is compiled on Linux, the paths in the compile_commands used by clangd are all Linux paths. To correctly navigate to files, the LSP needs to map the Linux paths provided by clangd to paths in Windows.

jwortmann:
Thanks for the explanation.

I see now that the `"path_maps"` config already existed in the code, it was just not documented (unfinished implementation?). Then it should also be added into the JSON schema under https://github.com/sublimelsp/LSP/blob/fb0fc753ed9b918d1903d8efa9ea084866ef4281/sublime-package.json#L323-L327

I wonder why path_maps is a list of dicts, and not just simply a single dict. But I'm also not very familiar with that part of the code, and it might not be a good idea to change it now if that would be breaking.

95833:
Most of the code should be provided by @rwols  , I am just using his code to meet some of my needs.  He is more familiar with this part of the code than we are, so he should decide how to handle this PR.

PR #2582, opened 2024-12-18
Disable workaround for on_post_move_async listener not triggering

jwortmann:
It was fixed upstream, so we can get rid of accessing the undocumented (private) `sublime_plugin.view_event_listeners` dict on current builds.

### DIFF
--- a/boot.py
+++ b/boot.py
@@ -17,6 +17,7 @@ from .plugin.configuration import LspDisableLanguageServerGloballyCommand
 from .plugin.configuration import LspDisableLanguageServerInProjectCommand
 from .plugin.configuration import LspEnableLanguageServerGloballyCommand
 from .plugin.configuration import LspEnableLanguageServerInProjectCommand
+from .plugin.core.constants import ST_VERSION
 from .plugin.core.css import load as load_css
 from .plugin.core.open import opening_files
 from .plugin.core.panels import PanelName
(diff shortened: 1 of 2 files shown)


jwortmann:
I wonder if the previous (current) code was even correct in case there are multiple simultaneous sessions for the view, because of the return statement at line https://github.com/sublimelsp/LSP/blob/c64ac00a1295fdc8461811033e0974443fd9ea55/boot.py#L240
(which was even active before rewriting that line in my PR #1927 and) which presumably would cause to miss re-registering of additional sessions.

rchl (approved):

rchl:
> I wonder if the previous (current) code was even correct in case there are multiple simultaneous sessions for the view, because of the return statement at line
> 
> https://github.com/sublimelsp/LSP/blob/c64ac00a1295fdc8461811033e0974443fd9ea55/boot.py#L240
> 
> (which was even active before rewriting that line in my PR #1927 and) which presumably would cause to miss re-registering of additional sessions.

But `DocumentSyncListener` is per view, not per session, so calling `listener.on_post_move_window_async` once should be enough?

jwortmann:
> But `DocumentSyncListener` is per view, not per session, so calling `listener.on_post_move_window_async` once should be enough?

Ah, you're right. I got confused, but of course it's correct to call it once.

jwortmann:
I think this PR can be merged

PR #2586, opened 2025-01-14
Remove language IDs configuration file

jwortmann:
As previously mentioned in https://github.com/sublimelsp/LSP/discussions/2568#discussioncomment-11470112 I see no use case for this configuration file. If there is some entry missing in the current language ID mapping (like a new syntax with unconventional base scope), it would be better for everyone to fix it directly in the LSP package, rather than allowing users to add individual overrides. If anyone sees any use case that I missed, please correct me if I'm wrong.


predragnikolic (approved):

predragnikolic:
I will see to merge this if no other reviews appear.

jwortmann:
Maybe wait a day or two to see if there is anything that would be against this change. Because it is a somewhat breaking change. I wonder if this configuration file was ever used by anybody.

predragnikolic:
@jwortmann [f8c2ee9](https://github.com/sublimelsp/LSP/pull/2586/commits/f8c2ee98d11e1d4f9edb3e80a412a2fdbc5a7faa) will make this a not breaking change for now:

LSP will function as before,
but a deprecation warning will be shown to the user,
The warning in the sublime console will redirect them to the migration guide https://github.com/sublimelsp/LSP/issues/2592
This will give time LSP users to migrate before we actually remove the file in the next **minor** release. (I took the liberty to choose the time, but I'm fine with if you want a differently date)
---
More importantly, I will revert the change if you do not find it appropriate.

jwortmann:
Okay the change looks fine, good idea.

PR #2589, opened 2025-02-03
Add instructions for helm-ls to docs

FichteFoll:


docs/src/language_servers.md
```diff
@@ -235,6 +235,23 @@ Follow installation instructions on [LSP-graphql](https://github.com/sublimelsp/
...
+
+1. Install [helm-ls](https://github.com/mrjosh/helm-ls).
+2. (Optional & recommended) Install [yaml-language-server](https://github.com/redhat-developer/yaml-language-server).
+3. Open `Preferences > Package Settings > LSP > Settings` and add the `"helm-ls"` client configuration to the `"clients"`:
+
+    ```jsonc
+    {
+        "clients": {
+            "helm-ls": {
+                "command": ["helm_ls", "serve"],
```
predragnikolic, line 248: missing "enabled": true

docs/src/language_servers.md
```diff
@@ -235,6 +235,23 @@ Follow installation instructions on [LSP-graphql](https://github.com/sublimelsp/
     }
     ```
 
+## Helm
+
+1. Install [helm-ls](https://github.com/mrjosh/helm-ls).
+2. (Optional & recommended) Install [yaml-language-server](https://github.com/redhat-developer/yaml-language-server).
```
predragnikolic, line 241: I would suggest removing step 2, and instead add a text that `helm-ls` could be used together with `LSP-yaml` (and add a link to LSP-yaml)

FichteFoll: Using both language servers from ST is a subpar experience because the helm LS cannot manage the YAML-specific parts on its own and the YAML LS chokes on Helm templates. 

It is a far better experience to have the helm LS also run the YAML LS in the background, which requires it to be installed and available on the PATH. There is probably a configuration option for the helm LS to specify a concrete path as well.

FichteFoll: I am also using the `YAML (Go)` with helm files, of course, but  they use the `.yml` extension, which is why I need to use a plugin to assign the correct syntax to template files following a specific path schema. I should probably tweak the selector.

predragnikolic: > Using both language servers from ST is a subpar experience because the helm LS cannot manage the YAML-specific parts on its own and the YAML LS chokes on Helm templates.
> It is a far better experience to have the helm LS also run the YAML LS in the background, which requires it to be installed and available on the PATH. There is probably a configuration option for the helm LS to specify a concrete path as well.
> I should probably also add a note that using LSP-yaml in parallel is not recommended.

That functionality isn't immediately obvious, until I have read what you wrote I had no idea. 
So yes, it would make sense to mention that in the docs.

docs/src/language_servers.md
```diff
@@ -245,13 +245,18 @@ Follow installation instructions on [LSP-graphql](https://github.com/sublimelsp/
...
-                "selector": "source.yaml",
+                "selector": "source.yaml.go", // Requires ST 4181+. Use `source.yaml` otherwise.
             },
         },
     }
     ```
 
+Note that the YAML language server on its own does not function properly for Helm files
+and should thus be disabled –
+ideally by adjusting its selector to `source.yaml - source.yaml.go`.
```
FichteFoll, line 258: I made this change in my local settings already, but imo it makes sense to upstream it as the default. What do you say?

jfcherng: The corresponding change has been merged into LSP-yaml.

docs/src/language_servers.md
```diff
@@ -235,6 +235,28 @@ Follow installation instructions on [LSP-graphql](https://github.com/sublimelsp/
     }
     ```
 
+## Helm
+
+1. Install [helm-ls](https://github.com/mrjosh/helm-ls).
+2. (Optional & recommended) Install [yaml-language-server](https://github.com/redhat-developer/yaml-language-server).
```
predragnikolic, line 241:
```suggestion
2. (Optional & recommended) Install integration with [yaml-language-server](https://github.com/mrjosh/helm-ls?tab=readme-ov-file#integration-with-yaml-language-server). 
```

FichteFoll: Your suggested wording doesn't make sense imo, but I replaced the link.

predragnikolic: Yes, it really doesn't make sense. Sorry about that.

predragnikolic (approved):

predragnikolic:
Thanks @FichteFoll for the PR.
---
I also want to communicate a mistake on my part.
Wanted to squash and merge the commits, but incidentally selected rebase and merge. (I always forget... I know that some people do not like that :) )

qvalentin:
Hi @FichteFoll, with version v.0.3.0 helm-ls should now also be enabled for values.yaml files (https://github.com/mrjosh/helm-ls/issues/61#issuecomment-2927585818).
Do you have an idea how this can be configured for sublime?

FichteFoll:
An LSP expert would need to comment on whether an LSP plugin can be active for files with specific names as I'm currently not aware of a way to do so (and don't have the time to look into it in detail). That does not mean there isn't one, however.

jwortmann:
The Sublime LSP client identifies the applicable documents based on the syntax scope instead of file patterns. So the best way would probably be to create a small "dummy" syntax that only applies to files with the name `values.yaml` (assuming this should be the exact filename). For example
```yaml
%YAML 1.2
---
name: YAML (values.yaml)
scope: source.yaml.values
hidden: true
hidden_file_extensions:
  - values.yaml

contexts:
  main:
    - include: scope:source.yaml
      apply_prototype: true
```

Then you can add the new base scope to the `selector` option in the configuration:
```jsonc
"selector": "source.yaml.go, source.yaml.values",
```

PR #2618, opened 2025-08-16
Copy hover response content and diagnostics text from the hover popup

### LINKED ISSUE #2600: Enable text selection and copying in LSP pop-up
As a developer, I am not able to select and copy errors brought up in the LSP pop-up. Oftentimes, I need to ask AI for help but not being able to copy & paste the error to AI frustrates me.

Allow users to select text using mouse or copy all through a button

Currently the workaround is by toggling diagnostics panel which shows pop-up content so you can copy

**Additional context**
This feature has been requested in https://github.com/sublimehq/sublime_text/issues/1649

predragnikolic:
This is a POC that could address #2600

https://github.com/user-attachments/assets/ca73753f-6497-43c3-9a33-9f4a4a27bf2c

This enables coping hover responses from server,
and diagnostics text, at the moment.

TODO:
- [x] Error diagnostics look broken, need to fix that.()
Screenshot 2025-08-16 at 11 14 27
-[] Add logic to copy signature help as well.

closes #2600

### DIFF
--- a/boot.py
+++ b/boot.py
@@ -56,6 +56,7 @@ from .plugin.goto_diagnostic import LspGotoDiagnosticCommand
 from .plugin.hierarchy import LspCallHierarchyCommand
 from .plugin.hierarchy import LspHierarchyToggleCommand
 from .plugin.hierarchy import LspTypeHierarchyCommand
+from .plugin.hover import LspCopyTextCommand
 from .plugin.hover import LspHoverCommand
 from .plugin.hover import LspToggleHoverPopupsCommand
 from .plugin.inlay_hint import LspInlayHintClickCommand
@@ -158,6 +159,7 @@ __all__ = (
     "LspUpdateLogPanelCommand",
     "LspUpdatePanelCommand",
     "LspWorkspaceSymbolsCommand",
+    'LspCopyTextCommand',
     "TextChangeListener",
     "plugin_loaded",
     "plugin_unloaded",
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -22,7 +22,6 @@ from .core.sessions import Session
 from .core.settings import userprefs
 from .core.views import copy_text_html
 from .core.views import FORMAT_STRING, FORMAT_MARKUP_CONTENT
-from .core.views import markup_to_string
 from .core.views import MarkdownLangMap
 from .core.views import minihtml
 from .core.views import range_to_region
@@ -308,8 +307,7 @@ class LspResolveDocsCommand(LspTextCommand):
             minihtml_content += copy_text_html(f"<div class='highlight'>{detail}</div>",
                                                copy_text=item.get('detail') or "")
         if documentation:
-            minihtml_content += copy_text_html(documentation,
-                                               copy_text=markup_to_string(item.get("documentation") or ""))
+            minihtml_content += copy_text_html(documentation, copy_text=item.get("documentation") or "")
 
         def run_main() -> None:
             if not self.view.is_valid():
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -305,9 +305,9 @@ class LspResolveDocsCommand(LspTextCommand):
             documentation = self._format_documentation(markdown, None)
         minihtml_content = ""
         if detail:
-            minihtml_content += copy_text_html(f"<div class='highlight'>{detail}</div>", copy_text=item.get('detail'))
+            minihtml_content += copy_text_html(f"<div class='highlight'>{detail}</div>", copy_text=item.get('detail') or "")
         if documentation:
-            minihtml_content += copy_text_html(documentation, markup_to_string(item.get("documentation") or ""))
+            minihtml_content += copy_text_html(documentation, copy_text=markup_to_string(item.get("documentation") or ""))
 
         def run_main() -> None:
             if not self.view.is_valid():
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -3,13 +3,13 @@ from .core.constants import COMPLETION_KINDS
 from .core.edit import apply_text_edits
 from .core.logging import debug
 from .core.promise import Promise
-from .core.protocol import EditRangeWithInsertReplace
 from .core.protocol import CompletionItem
 from .core.protocol import CompletionItemDefaults
 from .core.protocol import CompletionItemKind
 from .core.protocol import CompletionItemTag
 from .core.protocol import CompletionList
 from .core.protocol import CompletionParams
+from .core.protocol import EditRangeWithInsertReplace
 from .core.protocol import Error
 from .core.protocol import InsertReplaceEdit
 from .core.protocol import InsertTextFormat
@@ -20,7 +20,9 @@ from .core.protocol import TextEdit
 from .core.registry import LspTextCommand
 from .core.sessions import Session
 from .core.settings import userprefs
+from .core.views import copy_text_html_element
 from .core.views import FORMAT_STRING, FORMAT_MARKUP_CONTENT
+from .core.views import get_copy_text_from_markup
 from .core.views import MarkdownLangMap
 from .core.views import minihtml
 from .core.views import range_to_region
@@ -303,9 +305,9 @@ class LspResolveDocsCommand(LspTextCommand):
             documentation = self._format_documentation(markdown, None)
         minihtml_content = ""
         if detail:
-            minihtml_content += f"<div class='highlight'>{detail}</div>"
+            minihtml_content += copy_text_html_element(f"<div class='highlight'>{detail}</div>", copy_text=item.get('detail'))
         if documentation:
-            minihtml_content += documentation
+            minihtml_content += copy_text_html_element(documentation, get_copy_text_from_markup(item.get("documentation") or ""))
 
         def run_main() -> None:
             if not self.view.is_valid():
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -20,9 +20,9 @@ from .core.protocol import TextEdit
 from .core.registry import LspTextCommand
 from .core.sessions import Session
 from .core.settings import userprefs
-from .core.views import copy_text_html_element
+from .core.views import copy_text_html
 from .core.views import FORMAT_STRING, FORMAT_MARKUP_CONTENT
-from .core.views import get_copy_text_from_markup
+from .core.views import markup_to_string
 from .core.views import MarkdownLangMap
 from .core.views import minihtml
 from .core.views import range_to_region
@@ -305,9 +305,9 @@ class LspResolveDocsCommand(LspTextCommand):
             documentation = self._format_documentation(markdown, None)
         minihtml_content = ""
         if detail:
-            minihtml_content += copy_text_html_element(f"<div class='highlight'>{detail}</div>", copy_text=item.get('detail'))
+            minihtml_content += copy_text_html(f"<div class='highlight'>{detail}</div>", copy_text=item.get('detail'))
         if documentation:
-            minihtml_content += copy_text_html_element(documentation, get_copy_text_from_markup(item.get("documentation") or ""))
+            minihtml_content += copy_text_html(documentation, markup_to_string(item.get("documentation") or ""))
 
         def run_main() -> None:
             if not self.view.is_valid():
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -305,9 +305,11 @@ class LspResolveDocsCommand(LspTextCommand):
             documentation = self._format_documentation(markdown, None)
         minihtml_content = ""
         if detail:
-            minihtml_content += copy_text_html(f"<div class='highlight'>{detail}</div>", copy_text=item.get('detail') or "")
+            minihtml_content += copy_text_html(f"<div class='highlight'>{detail}</div>",
+                                               copy_text=item.get('detail') or "")
         if documentation:
-            minihtml_content += copy_text_html(documentation, copy_text=markup_to_string(item.get("documentation") or ""))
+            minihtml_content += copy_text_html(documentation,
+                                               copy_text=markup_to_string(item.get("documentation") or ""))
 
         def run_main() -> None:
             if not self.view.is_valid():
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -332,7 +332,8 @@ class LspResolveDocsCommand(LspTextCommand):
         return minihtml(self.view, content, FORMAT_STRING | FORMAT_MARKUP_CONTENT, language_map)
 
     def _on_navigate(self, url: str) -> None:
-        webbrowser.open(url)
+        if url.startswith("http"):
+            webbrowser.open(url)
 
 
 class LspCommitCompletionWithOppositeInsertMode(LspTextCommand):
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -822,7 +822,10 @@ def _html_element(name: str, text: str, class_name: str | None = None, escape: b
 def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, base_dir: str | None = None) -> str:
     code = diagnostic.get("code")
     source = diagnostic.get("source") or ""
-    html = copy_text_html(_html_element('span', diagnostic["message"]), copy_text=f"{source} {diagnostic['message']}")
+    html = copy_text_html(
+        _html_element('span', diagnostic["message"]),
+        copy_text=f"{source} {diagnostic['message']}"
+    )
     if source or code is not None:
         meta_info = ""
         if source:
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -844,7 +844,7 @@ def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, bas
     return _html_element("pre", html, class_name=severity_class, escape=False)
 
 
-def markup_to_string(content: MarkupContent | MarkedString | list[MarkedString]) -> str:
+def _markup_to_string(content: MarkupContent | MarkedString | list[MarkedString]) -> str:
     if isinstance(content, str):
         return content
     elif isinstance(content, dict):
@@ -853,7 +853,8 @@ def markup_to_string(content: MarkupContent | MarkedString | list[MarkedString])
         return " ".join(content)   # pyright: ignore[reportCallIssue, reportUnknownVariableType, reportArgumentType]
 
 
-def copy_text_html(html_content: str, copy_text: str) -> str:
+def copy_text_html(html_content: str, copy_text: str | MarkupContent | MarkedString | list[MarkedString]) -> str:
+    copy_text = _markup_to_string(copy_text)
     if not len(copy_text):
         return html_content
     return f"""<a title="Click to Copy"
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -837,7 +837,7 @@ def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, bas
         html += " " + _html_element("span", meta_info, class_name="color-muted", escape=False)
     related_infos = diagnostic.get("relatedInformation")
     if related_infos:
-        info = "<br>".join(_format_diagnostic_related_info(config, info, base_dir) for info in related_infos)
+        info = "<br>".join(copy_text_html(_format_diagnostic_related_info(config, info, base_dir), copy_text=info['message']) for info in related_infos)
         html += '<br>' + _html_element("pre", info, class_name="related_info", escape=False)
     severity_class = DIAGNOSTIC_SEVERITY[diagnostic_severity(diagnostic) - 1][1]
     return _html_element("pre", html, class_name=severity_class, escape=False)
@@ -852,9 +852,9 @@ def markup_to_string(content: MarkupContent | MarkedString | list[MarkedString])
         return " ".join(content)   # pyright: ignore[reportCallIssue, reportUnknownVariableType, reportArgumentType]
 
 
-def copy_text_html(html_content: str, copy_text: str | None = None) -> str:
-    if copy_text is None:
-        copy_text = html_content
+def copy_text_html(html_content: str, copy_text: str) -> str:
+    if not len(copy_text):
+        return html_content
     return f"""<a title="Click to Copy"
        style='text-decoration: none; display: block; color: var(--foreground)'
        href='{sublime.command_url('lsp_copy_text', {
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -852,7 +852,7 @@ def get_copy_text_from_markup(content: MarkupContent | MarkedString | list[Marke
 def copy_text_html_element(html_content: str, copy_text: str | None = None) -> str:
     if copy_text is None:
         copy_text = html_content
-    return f"""<a title="Click to copy"
+    return f"""<a title="Click to Copy"
        style='text-decoration: none; display: block; color: var(--foreground)'
        href='{sublime.command_url('lsp_copy_text', {
         'text': html.unescape(copy_text.replace(' ', ' '))
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -843,14 +843,11 @@ def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, bas
 def copy_text_html_element(html_content: str, copy_text: str | None = None) -> str:
     if copy_text is None:
         copy_text = html_content
-    return f"""
-    <a title="Click to copy"
+    return f"""<a title="Click to copy"
        style='text-decoration: none; display: block; color: var(--foreground)'
        href='{sublime.command_url('lsp_copy_text', {
         'text': html.unescape(copy_text.replace(' ', ' '))
-       })}'>
-        {html_content}
-    </a>"""
+       })}'>{html_content}</a>"""
 
 
 def format_code_actions_for_quick_panel(
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -822,10 +822,8 @@ def _html_element(name: str, text: str, class_name: str | None = None, escape: b
 def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, base_dir: str | None = None) -> str:
     code = diagnostic.get("code")
     source = diagnostic.get("source") or ""
-    html = copy_text_html(
-        _html_element('span', diagnostic["message"]),
-        copy_text=f"{source} {diagnostic['message']}"
-    )
+    html = copy_text_html(_html_element('span', diagnostic["message"]),
+                          copy_text=f"{source} {diagnostic['message']}")
     if source or code is not None:
         meta_info = ""
         if source:
@@ -837,7 +835,10 @@ def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, bas
         html += " " + _html_element("span", meta_info, class_name="color-muted", escape=False)
     related_infos = diagnostic.get("relatedInformation")
     if related_infos:
-        info = "<br>".join(copy_text_html(_format_diagnostic_related_info(config, info, base_dir), copy_text=info['message']) for info in related_infos)
+        info = "<br>".join(copy_text_html(
+            _format_diagnostic_related_info(config, info, base_dir),
+            copy_text=info['message']
+        ) for info in related_infos)
         html += '<br>' + _html_element("pre", info, class_name="related_info", escape=False)
     severity_class = DIAGNOSTIC_SEVERITY[diagnostic_severity(diagnostic) - 1][1]
     return _html_element("pre", html, class_name=severity_class, escape=False)
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -820,9 +820,9 @@ def _html_element(name: str, text: str, class_name: str | None = None, escape: b
 
 
 def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, base_dir: str | None = None) -> str:
-    html = copy_text_html_element(_html_element('span', diagnostic["message"]), copy_text=diagnostic["message"])
     code = diagnostic.get("code")
-    source = diagnostic.get("source")
+    source = diagnostic.get("source") or ""
+    html = copy_text_html_element(_html_element('span', diagnostic["message"]), copy_text=f"{source} {diagnostic['message']}")
     if source or code is not None:
         meta_info = ""
         if source:
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -860,7 +860,7 @@ def copy_text_html(html_content: str, copy_text: str | MarkupContent | MarkedStr
     return f"""<a title="Click to Copy"
        style='text-decoration: none; display: block; color: inherit'
        href='{sublime.command_url('lsp_copy_text', {
-        'text': html.unescape(copy_text.replace(' ', ' '))
+        'text': copy_text.replace(' ', ' ')
        })}'>{html_content}</a>"""
 
 
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -820,7 +820,7 @@ def _html_element(name: str, text: str, class_name: str | None = None, escape: b
 
 
 def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, base_dir: str | None = None) -> str:
-    html = _html_element('span', diagnostic["message"])
+    html = copy_text_html_element(_html_element('span', diagnostic["message"]), copy_text=diagnostic["message"])
     code = diagnostic.get("code")
     source = diagnostic.get("source")
     if source or code is not None:
@@ -838,6 +838,19 @@ def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, bas
         html += '<br>' + _html_element("pre", info, class_name="related_info", escape=False)
     severity_class = DIAGNOSTIC_SEVERITY[diagnostic_severity(diagnostic) - 1][1]
     return _html_element("pre", html, class_name=severity_class, escape=False)
+
+
+def copy_text_html_element(html_content: str, copy_text: str | None = None) -> str:
+    if copy_text is None:
+        copy_text = html_content
+    return f"""
+    <a title="Click to copy"
+       style='text-decoration: none; display: block; color: var(--foreground)'
+       href='{sublime.command_url('lsp_copy_text', {
+        'text': html.unescape(copy_text.replace(' ', ' '))
+       })}'>
+        {html_content}
+    </a>"""
 
 
 def format_code_actions_for_quick_panel(
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -840,6 +840,15 @@ def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, bas
     return _html_element("pre", html, class_name=severity_class, escape=False)
 
 
+def get_copy_text_from_markup(content: MarkupContent | MarkedString | list[MarkedString]) -> str:
+    if isinstance(content, str):
+        return content
+    elif isinstance(content, dict):
+        return content.get('value', '')
+    elif isinstance(content, list):
+        return " ".join(content)   # pyright: ignore[reportCallIssue, reportUnknownVariableType, reportArgumentType]
+
+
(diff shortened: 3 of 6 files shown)


plugin/hover.py
```diff
@@ -420,3 +425,20 @@ def _update_views_async(self, enable: bool) -> None:
...
+def get_copy_text_from_hover_contents(content: MarkupContent | MarkedString | list[MarkedString]) -> str:
+    if isinstance(content, str):
+        return content
+    elif isinstance(content, dict):
+        return content.get('value', '')
+    elif isinstance(content, list):
+        return " ".join(content)   # pyright: ignore[reportCallIssue, reportUnknownVariableType, reportArgumentType]
+
+
+class LspCopyTextCommand(sublime_plugin.TextCommand):
```
rchl: Would using the built-in `copy` command be enough? Not sure we need the copied text to show in the status field.

rchl (commented):
What about a case when the popup also shows other info like symbol documentation? Is the whole contents of the popup copied or each section is copied separately. If the latter then it would probably be hard to convey what is going to get copied.

plugin/hover.py
predragnikolic: It would, do you know if the copy command accepts text as an arg?

I tried to find it but no success.

predragnikolic:
> or each section is copied separately

Yes, it is the later.
And it is visually hard to distinguish what gets copied sometimes. (I did enable pyright, basepyrigh and pylsp to try out this use case). It is a limitation I am not sure how to address yet.

plugin/hover.py
rchl: I suppose it might not. It might just take current selection from the active view.

But then I would mimic its behavior regarding what is shown in the status field because showing some random text can look messy and unnecessary (can also push things around)

plugin/hover.py
```diff
@@ -420,3 +426,8 @@ def _update_views_async(self, enable: bool) -> None:
                     session_view.view.settings().set(SHOW_DEFINITIONS_KEY, False)
                 else:
                     session_view.reset_show_definitions()
+
+
+class LspCopyTextCommand(sublime_plugin.TextCommand):
+    def run(self, edit, text: str) -> None:
+        sublime.set_clipboard(text)
```
rchl: Should update status with number of characters copied like the original copy does?

plugin/completion.py
```diff
@@ -303,9 +305,9 @@ def _handle_resolve_response_async(self, language_map: MarkdownLangMap | None, i
             documentation = self._format_documentation(markdown, None)
         minihtml_content = ""
         if detail:
-            minihtml_content += f"<div class='highlight'>{detail}</div>"
+            minihtml_content += copy_text_html(f"<div class='highlight'>{detail}</div>", copy_text=item.get('detail'))
         if documentation:
-            minihtml_content += documentation
+            minihtml_content += copy_text_html(documentation, markup_to_string(item.get("documentation") or ""))
```
rchl: Do existing links within those markdown blocks still work?

predragnikolic: I did hardcode `"[label](http://example.com) asdasdas"` to try it out, and it does work.

Screenshot 2025-09-02 at 22 39 46

I also did test other examples, and they do seem to work.

https://github.com/user-attachments/assets/3b76c6a7-5c65-43be-8cf6-c603343db920


I only stumble upon one issue with the signature help code and LSP-pyright,
for some reason, trying to copy the signature help there, will move the ST window in the background (this doesn't happen with LSP-typescript)

https://github.com/user-attachments/assets/d7ad8904-9edc-4b71-b2ee-23096717c6be
> notice that the ST window disappears once I click to copy the text in the hover popup

I will try to see what is wrong.

predragnikolic: > I will try to see what is wrong.

fixed with [ab65e90](https://github.com/sublimelsp/LSP/pull/2618/commits/ab65e90d109df1149d648763fa2f290db01a130a)

plugin/hover.py
predragnikolic: [1d8a53a](https://github.com/sublimelsp/LSP/pull/2618/commits/1d8a53a732da98121fd6e15d41d0880bf895176f)

predragnikolic:
The functionality is implemented and it should work fine.
I have problems updating the signature help test. I did locally update:

```diff
def copy_text_html(html_content: str, copy_text: str | MarkupContent | MarkedString | list[MarkedString]) -> str:
    copy_text = _markup_to_string(copy_text)
    if not len(copy_text):
        return html_content
    return f"""<a title="Click to Copy" style='text-decoration: none; display: block; color: inherit' href='{sublime.command_url('lsp_copy_text', {
        'text': copy_text.replace(' ', ' ')
       })}'>{html_content}</a>"""
```


```diff
 r'''
+ <a [^>]+>
 <div class="highlight"><pre>
 <span style="color: #\w{6}">RUN </span>
 <span style="color: #\w{6}">\[</span>
 <span style="color: #\w{6}"> </span>
 <span style="color: #\w{6}">"command"</span>
 <span style="color: #\w{6}"> </span>
 <span style="color: #\w{6}; font-weight: bold; text-decoration: underline">"parameters"</span>
 <span style="color: #\w{6}">, </span>
 <span style="color: #\w{6}">\.\.\.</span>
 <span style="color: #\w{6}"> </span>
 <span style="color: #\w{6}">\]</span>
- </pre></div>
+ </pre></div></a>
 '''
```

predragnikolic:
The functionality is implemented and it should work fine.
I just need to fix the failing tests, before I mark the PR for review.

---

I have problems fixing the signature help tests. 

I did update `copy_text_html` from:
```py
def copy_text_html(html_content: str, copy_text: str | MarkupContent | MarkedString | list[MarkedString]) -> str:
    copy_text = _markup_to_string(copy_text)
    if not len(copy_text):
        return html_content
    return f"""<a title="Click to Copy"
       style='text-decoration: none; display: block; color: inherit'
       href='{sublime.command_url('lsp_copy_text', {
        'text': copy_text.replace(' ', ' ')
       })}'>{html_content}</a>"""
```

to: (to remove new lines)
```py
def copy_text_html(html_content: str, copy_text: str | MarkupContent | MarkedString | list[MarkedString]) -> str:
    copy_text = _markup_to_string(copy_text)
    if not len(copy_text):
        return html_content
    return f"""<a title="Click to Copy" style='text-decoration: none; display: block; color: inherit' href='{sublime.command_url('lsp_copy_text', {
        'text': copy_text.replace(' ', ' ')
       })}'>{html_content}</a>"""
```

and I have updated the regex to be:

```diff
 r'''
+ <a [^>]+>
 <div class="highlight"><pre>
 <span style="color: #\w{6}">RUN </span>
 <span style="color: #\w{6}">\[</span>
 <span style="color: #\w{6}"> </span>
 <span style="color: #\w{6}">"command"</span>
 <span style="color: #\w{6}"> </span>
 <span style="color: #\w{6}; font-weight: bold; text-decoration: underline">"parameters"</span>
 <span style="color: #\w{6}">, </span>
 <span style="color: #\w{6}">\.\.\.</span>
 <span style="color: #\w{6}"> </span>
 <span style="color: #\w{6}">\]</span>
- </pre></div>
+ </pre></div></a>
 '''
```

But this does not work as I still get this error:
```
AssertionError: Regex didn't match: '<a  title="Click to Copy" style=\'text-decoration: none; display: block; color: inherit\' href=\'subl:lsp_copy_text {&quot;text&quot;:&quot;RUN [ \\&quot;command\\&quot; \\&quot;parameters\\&quot;, ... ]&quot;}\'><div class="highlight"><pre><span style="color: #\\w{6}">RUN </span><span style="color: #\\w{6}">\\[</span><span style="color: #\\w{6}"> </span><span style="color: #\\w{6}">"command"</span><span style="color: #\\w{6}"> </span><span style="color: #\\w{6}; font-weight: bold; text-decoration: underline">"parameters"</span><span style="color: #\\w{6}">, </span><span style="color: #\\w{6}">\\.\\.\\.</span><span style="color: #\\w{6}"> </span><span style="color: #\\w{6}">\\]</span></pre></div></a>' not found in '<a title="Click to Copy" style=\'text-decoration: none; display: block; color: inherit\' href=\'subl:lsp_copy_text {&quot;text&quot;:&quot;RUN [ \\&quot;command\\&quot; \\&quot;parameters\\&quot;, ... ]&quot;}\'><div class="highlight"><pre><span style="color: #ffffff">RUN </span><span style="color: #8c8a90">[</span><span style="color: #ffffff"> </span><span style="color: #8c8a90">"command"</span><span style="color: #ffffff"> </span><span style="color: #f6ac90">"parameters"</span><span style="color: #ffffff">, </span><span style="color: #8c8a90">...</span><span style="color: #ffffff"> </span><span style="color: #8c8a90">]</span></pre></div></a>'
```

I will try to fix the failing tests on another day.
If someone want to take a look at the failing tests in the mean while, feel free. Any help is welcome.

predragnikolic:
I probably closed the PR by accident (probably pressed the keyboard shortcut)

predragnikolic:
I pushed [18b26db](https://github.com/sublimelsp/LSP/pull/2618/commits/18b26db79879d314f1e8ba638a8007464b2d3663) because the generated HTML looked wrong -> "<a><div>...</a></div>" (a seemed to be closed before the div)... but it was my lack of concentration. The generated html was OK.

Because I wanted to support signature as well, 
I dig into the issue with signature help and discovered that the regexes in tests I had locally were not right. So I just updated the tests.

---

Now, I consider this PR read for review.

boot.py
```diff
@@ -158,6 +159,7 @@
     "LspUpdateLogPanelCommand",
     "LspUpdatePanelCommand",
     "LspWorkspaceSymbolsCommand",
+    'LspCopyTextCommand',
```
rchl: In sorted order please.

plugin/core/views.py
```diff
@@ -832,14 +832,38 @@ def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, bas
...
     severity_class = DIAGNOSTIC_SEVERITY[diagnostic_severity(diagnostic) - 1][1]
     return _html_element("pre", html, class_name=severity_class, escape=False)
 
 
+def _markup_to_string(content: MarkupContent | MarkedString | list[MarkedString]) -> str:
+    if isinstance(content, str):
+        return content
+    elif isinstance(content, dict):
+        return content.get('value', '')
+    elif isinstance(content, list):
```
rchl: just a style nit but no need for `elif` after `return`. `if` would be enough.

plugin/core/views.py
```diff
@@ -832,14 +832,38 @@ def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, bas
...
+        return " ".join(content)   # pyright: ignore[reportCallIssue, reportUnknownVariableType, reportArgumentType]
```
rchl: Why hacking this with type ignores? `MarkedString` is not necessarily a plain string so this seems wrong.

Possibly:

```suggestion
        return " ".join([_markup_to_string(text) for text in content])
```

plugin/core/views.py
```diff
@@ -832,14 +832,38 @@ def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, bas
...
+
+
+def copy_text_html(html_content: str, copy_text: str | MarkupContent | MarkedString | list[MarkedString]) -> str:
+    copy_text = _markup_to_string(copy_text)
+    if not len(copy_text):
+        return html_content
+    return f"""<a title="Click to Copy"
+       style='text-decoration: none; display: block; color: inherit'
+       href='{sublime.command_url('lsp_copy_text', {
+        'text': copy_text.replace(' ', ' ')
```
rchl: Why do we need to replace non-breaking spaces?

plugin/core/views.py
```diff
@@ -832,14 +832,38 @@ def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, bas
...
+def _markup_to_string(content: MarkupContent | MarkedString | list[MarkedString]) -> str:
+    if isinstance(content, str):
+        return content
+    elif isinstance(content, dict):
+        return content.get('value', '')
+    elif isinstance(content, list):
+        return " ".join(content)   # pyright: ignore[reportCallIssue, reportUnknownVariableType, reportArgumentType]
+
+
+def copy_text_html(html_content: str, copy_text: str | MarkupContent | MarkedString | list[MarkedString]) -> str:
```
rchl: This naming makes it hard to understand the purpose IMO.

Perhaps `wrap_in_copy_to_clipboard_link` with arguments `link_text` and `text_to_copy`?

And maybe the `_markup_to_string` should also be done separately so that this function is more specialized and less do-it-all.

plugin/core/views.py
```diff
@@ -832,14 +832,38 @@ def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, bas
...
+        info = "<br>".join(copy_text_html(
+            _format_diagnostic_related_info(config, info, base_dir),
+            copy_text=info['message']
+        ) for info in related_infos)
         html += '<br>' + _html_element("pre", info, class_name="related_info", escape=False)
     severity_class = DIAGNOSTIC_SEVERITY[diagnostic_severity(diagnostic) - 1][1]
     return _html_element("pre", html, class_name=severity_class, escape=False)
 
 
+def _markup_to_string(content: MarkupContent | MarkedString | list[MarkedString]) -> str:
```
rchl: It's a good practice to place the functions in the order they are used. Helps when reading the code. So move this one to after `copy_text_html`

plugin/core/views.py
```diff
@@ -832,14 +832,38 @@ def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, bas
...
 
+def copy_text_html(html_content: str, copy_text: str | MarkupContent | MarkedString | list[MarkedString]) -> str:
+    copy_text = _markup_to_string(copy_text)
+    if not len(copy_text):
+        return html_content
+    return f"""<a title="Click to Copy"
+       style='text-decoration: none; display: block; color: inherit'
+       href='{sublime.command_url('lsp_copy_text', {
+        'text': copy_text
+       })}'>{html_content}</a>"""
```
rchl: Seems to create an extra padding in the popups (the bottom one should be the same as all other):

Image

I have not looked into it too much. Perhaps there is an empty `a` element there?
Or perhaps it has some margin collapsing bug which ST is known for.

predragnikolic: Seems to be a <br>
When I hover:
```ts
const foo = "bar"
```
the html content with this branch is:
```
mdpopups: =====HTML OUTPUT=====
mdpopups: 
<body><a title="Click to Copy"
       style='text-decoration: none; display: block; color: inherit'
       href='subl:lsp_copy_text {&quot;text&quot;:&quot;\n```typescript\nconst foo: \&quot;bar\&quot;\n```\n&quot;}'><div class="highlight"><pre><span style="color: #fcb6b6;">const</span><span style="color: #ffffff;"> </span><span style="color: #ffffff;">foo</span><span style="color: #ffffff;">:</span><span style="color: #ffffff;"> </span><span style="color: #fcd49e;">"</span><span style="color: #fcd49e;">bar</span><span style="color: #fcd49e;">"</span><br></pre></div></a><div class="lsp_popup--spacer"></div></body>
```

while the html with the main branch still contains a `br`
```
mdpopups: =====HTML OUTPUT=====
mdpopups: 
<body><div class="highlight"><pre><span style="color: #fcb6b6;">const</span><span style="color: #ffffff;"> </span><span style="color: #ffffff;">foo</span><span style="color: #ffffff;">:</span><span style="color: #ffffff;"> </span><span style="color: #fcd49e;">"</span><span style="color: #fcd49e;">bar</span><span style="color: #fcd49e;">"</span><br></pre></div><div class="lsp_popup--spacer"></div></body>
```
but there is margin bottom on main.

I will see what is going on.

jwortmann:
In my opinion the copying should be restricted to diagnostics messages and to the content of (fenced) code blocks.
I don't think that the current choice of just copying the entire hover content as Markdown is extremely useful, and I don't like that if you accidentally click on a hover popup, you will lose your previous clipboard content. Also the tooltip is misleading (wrong) when you hover over links in the hover content. If restricting the copy functionality to code blocks is not possible, then I would move this functionality to an additional link "copy" in the symbol action links at the bottom of the popup.

predragnikolic:
While copying content from the whole hover popup looked useful, this comment made me rethink it:
 
> In my opinion the copying should be restricted to diagnostics messages and to the content of (fenced) code blocks.
I don't think that the current choice of just copying the entire hover content as Markdown is extremely useful, and I don't like that if you accidentally click on a hover popup, you will lose your previous clipboard content. Also the tooltip is misleading (wrong) when you hover over links in the hover content. If restricting the copy functionality to code blocks is not possible, then I would move this functionality to an additional link "copy" in the symbol action links at the bottom of the popup.

So, I've removed coping content from most hover popups,
but I only left coping diagnostics and tweaked the UI (there is a `copy` button now)
Screenshot 2025-10-15 at 21 13 32

Would you still find this PR useful if it only adds coping the diagnostics in the hover popup?

I know I've reduced the scope of the PR, 
but any approach I try to come up to allow copying text in popup will not match the UX that we could have if ST supported selecting text from hover popups natively.

jwortmann:
> Would you still find this PR useful if it only adds coping the diagnostics in the hover popup?

Yes, I like it. Would it be possible to make the copy button aligned to the right side of the popup?

Alternatively we could use icons instead of the "copy" button. For example:

clipboard

clipboard@2x

clipboard@3x

copy

copy@2x

copy@3x

It is possible to include the images in minihtml via `res://` URLs like in
https://github.com/sublimelsp/LSP/blob/aec423a51dda1c12a9c44aef834133a5abd732c5/plugin/hover.py#L98

The icons would probably also need a dark variant for light color schemes.

But the "copy" button looks good too. I'm just throwing in some ideas.

---

I think if we want a copy button for the fenced code blocks too in the future, we would need to parse the html from mdpopups after the conversion from Markdown and then inject additional html for the buttons. I think it's not impossible, but probably not very easy and there could be various edge cases. Also the minihtml is quite limited, so I don't know where the best placement for the copy buttons would be in that case. A better way might be if this could be handled directly by mdpopups. But probably this is out of scope for mdpopups, and an implementation there might also be complicated.

docs_copy

predragnikolic:
Just some small update regarding the image icon.

> Would it be possible to make the copy button aligned to the right side of the popup?

I don't think it is possible with the current minihtml API.
That could be achieved, if minihtml supported, "position: absolute". 

FWIW, when I try to put the icon closer to the text.
I didn't even manage to perfectly align the icon even in such case.
Minihtml CSS and HTML just behaves differently to what I expect it to behave. 

I had a couple of failed attempts.
When I added the image,
the error box height changed so elements because miss-aligned
a

One of the ways to address it is to set inline-block on the image, that fixed the height of the error box, but the copy image disappeared... 
b

Other attempts I tried didn't have success either.

---

I think I am gonna leave out adding the copy image, just because it causes the error box to look miss-aligned(as in the first image) and I have not found a way to make the error popup look aligned.

predragnikolic:
I tried to add images.
I've downloaded them, created a dark and light png image of the icon.
I knew that ST automatically adds the `light` OR `dark` class to the root html tag,
so I wanted to use it like this:

```html
<a class='copy_button'>
  Copy
  <img class='light_img' src='.../copy-light.png'> 
  <img class='dark_img' src='.../copy-dark.png'> 
</a>
```
and do the following in CSS, to show the appropriate image based on the light/dark theme. 
```css
.copy_button_light_img, .copy_button_dark_img { display: none }

// ST automatically adds light/dark classes to the root html tag
// https://www.sublimetext.com/docs/minihtml.html#predefined-classes
.light .light_img, .dark .dark_img  {
  display: inline-block; 
}
```
This would have been nice,
but minihtml didn't allow me to position images correctly. (see my previous message)
I tried a couple of ways. I just didn't find a way.

---

Then I spend time trying to find an UTF char that looked like a clipboard. 
I finally found one - https://www.compart.com/en/unicode/U+23CD (I think it is good and simple looking)

Screenshot 2025-11-11 at 21 05 48

rchl:
> Then I spend time trying to find an UTF char that looked like a clipboard.
> I finally found one - https://www.compart.com/en/unicode/U+23CD (I think it is good and simple looking)

Think that ⧉ might be more recognizable.

Screenshot 2025-11-11 at 21 14 02

rchl:
Some observations:
1. do we need the "copy" text or would just the icon be enough? It looks busy, especially with multiple diagnostics.
2. We don't currently copy the name of the server and optional error code. Should we? In case of ruff the code is even a link. EDIT: I see now that we do copy `source` but put it at the beginning. Maybe better to keep it at the end (in parenthesis or something)?
3. The extra text/icon actually makes the line height slightly taller. Ideally it wouldn't.

plugin/core/views.py
```diff
@@ -831,13 +831,22 @@ def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, bas
...
     if related_infos := diagnostic.get("relatedInformation"):
         info = "<br>".join(_format_diagnostic_related_info(config, info, base_dir) for info in related_infos)
         html += '<br>' + _html_element("pre", info, class_name="related_info", escape=False)
     severity_class = DIAGNOSTIC_SEVERITY[diagnostic_severity(diagnostic) - 1][1]
     return _html_element("pre", html, class_name=severity_class, escape=False)
 
 
+def copy_icon_html(text_to_copy: str) -> str:
+    return f"""<a class="copy_button"
+       style="color: inherit"
```
rchl: This inline style can be moved to the css class.

plugin/core/views.py
```diff
@@ -831,13 +831,22 @@ def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, bas
             meta_info += "({})".format(
                 make_link(code_description["href"], str(code)) if code_description else text2html(str(code)))
         html += " " + _html_element("span", meta_info, class_name="color-muted", escape=False)
+    html += "&nbsp;" + copy_icon_html(f"{source} {diagnostic['message']}")
```
rchl: If source is empty then there will be a stray space at the beginning.

plugin/core/views.py
predragnikolic: I needed to leave it in inlined styles for the `color: inherit` to work.
If I move it to popups.css, "inherit" will not work in that case as the color will be "blue". (even if I use !important)

Screenshot 2025-11-11 at 21 33 44

It is probably a minor bug in minihtml.

rchl: Will work if you change selector to `a.copy_button`

predragnikolic:
> - Think that ⧉ might be more recognizable.
> - do we need the "copy" text or would just the icon be enough? It looks busy, especially with multiple diagnostics.

If I use ⧉ and remove the "Copy" text, but still keep some padding and transparent bg, it can look like this: 

Screenshot 2025-11-11 at 21 49 16
Screenshot 2025-11-11 at 21 54 35
Screenshot 2025-11-11 at 21 53 32

plugin/core/views.py
predragnikolic: It will :) thanks, I will push a commit

plugin/core/views.py
predragnikolic: [3fbf73d](https://github.com/sublimelsp/LSP/pull/2618/commits/3fbf73d8bb8fb632915c666c064a702bb273bfa7)

plugin/core/views.py
predragnikolic: [4e6d21f](https://github.com/sublimelsp/LSP/pull/2618/commits/4e6d21f84572e7c5e73b8dd157ff22d1f00bd4bf)

plugin/core/views.py
```diff
@@ -821,7 +821,7 @@ def _html_element(name: str, text: str, class_name: str | None = None, escape: b
 def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, base_dir: str | None = None) -> str:
     html = _html_element('span', diagnostic["message"])
     code = diagnostic.get("code")
-    source = diagnostic.get("source")
+    source = diagnostic.get("source") or ""
```
rchl: can revert this now

predragnikolic:
Should the copy button be on the start or the end?

start
Screenshot 2025-11-11 at 22 41 39

end
Screenshot 2025-11-11 at 22 32 47

plugin/core/views.py
predragnikolic: [09113ba](https://github.com/sublimelsp/LSP/pull/2618/commits/09113babd9d76ca8e287412dd8f4a12c641d885e)

predragnikolic:
I pushed [6f9be2e](https://github.com/sublimelsp/LSP/pull/2618/commits/6f9be2e772a9e780fc9fdcc60c46a39bff001e50) - it use the png icon that @jwortmann provided and it looks like this:

Screenshot 2025-11-12 at 18 59 26

Screenshot 2025-11-12 at 19 02 19


I don't have an explanation why this didn't work properly before. (Probably because I haven't set the image height to be 1rem)

rchl:
> I pushed [6f9be2e](https://github.com/sublimelsp/LSP/pull/2618/commits/6f9be2e772a9e780fc9fdcc60c46a39bff001e50) - it use the png icon that @jwortmann provided and it looks like this:

Is there any benefit to doing it that way? To me it seems more complex and I don't really associate that icon with copying personally.

EDIT: well, the icon is fine but the question about complexity still stands.

rchl:
> Should the copy button be on the start or the end?
> 
> start Screenshot 2025-11-11 at 22 41 39
> 
> end Screenshot 2025-11-11 at 22 32 47

I feel like having it on the left makes it too prominent for something that in practice should rarely be used. I'd probably even make it to stand out less (no background and less opaque).

predragnikolic:
a)
Screenshot 2025-11-12 at 21 31 30

b)
Screenshot 2025-11-12 at 21 34 14

c)
Screenshot 2025-11-12 at 21 27 27

d)
Screenshot 2025-11-12 at 21 27 42

e)
Screenshot 2025-11-12 at 21 32 47

f)
Screenshot 2025-11-12 at 21 33 54

predragnikolic:
> EDIT: well, the icon is fine but the question about complexity still stands.

The png approach requires 10 lines more and 2 new png icons (dark and light variants).

I will try to explain [6f9be2e](https://github.com/sublimelsp/LSP/pull/2618/commits/6f9be2e772a9e780fc9fdcc60c46a39bff001e50), (but it still doesn't change the fact that it is more code and that is more complex than the utf-8 one char approach).

ST automatically adds `.dark` and `.light` class names to the root html element in the popup
(depending on the selected color scheme).

Two helper css classes are added by me:
- `.light_only` - the element will only be displayed if the ST color scheme is `.light`
- `.dark_only`  - the element will only be displayed if the ST color scheme is `.dark`

We define `copy_light_icon` and `copy_dark_icon` 
```
copy_light_icon = '<img class="light_only" src="res://Packages/LSP/icons/copy_light_theme.png" />'
copy_dark_icon = '<img class="dark_only" src="res://Packages/LSP/icons/copy_dark_theme.png" />'
```
and we render both in the HTML `<a ...>{copy_light_icon}{copy_dark_icon}</a>`
but because of the .light_only and .dark_only class names, only one image will be shown.

rchl:
To me the most subtle example "d" looks good.

Though versions using the utf8 character do actually enlarge the line height (due to font-size being over 1rem). So for that reason it might be better to go with an image (ideally slightly dimmed).

jwortmann:
While I think that the icon style from the images looks a tiny bit better than the Unicode symbol, I agree that it's probably not worth to add the image files and the additional required code. And the unicode symbol would make it easy to adjust the icon color via CSS. Couldn't you just reduce the font size just for the icon class, to prevent enlarging the line hight with the Unicode symbol?

I'm undecided about the dimmed background, I would say both variants c) and d) look fine.

I would probably use "Copy to clipboard" in the title (tooltip) text. And could you change the CSS class name from "copy_button" to "copy-icon"?

---

By the way, it looks like the people at VSCode just had the same idea: https://code.visualstudio.com/updates/v1_106#_copy-diagnostic-hover-text

plugin/hover.py
```diff
@@ -409,3 +409,10 @@ def _update_views_async(self, enable: bool) -> None:
                         session_view.view.settings().set(SHOW_DEFINITIONS_KEY, False)
                     else:
                         session_view.reset_show_definitions()
+
+
+class LspCopyTextCommand(sublime_plugin.TextCommand):
+    def run(self, edit, text: str) -> None:
+        sublime.set_clipboard(text)
+        text_length = len(text)
+        sublime.status_message(f"Copied {text_length} character{'s' if text_length > 1 else ''}")
```
jwortmann: I think the diagnostic message will always be longer than 1 character, so not sure if this is needed.

predragnikolic:
@jwortmann, sorry, I will revert the changes I pushed just now in [87cf18b](https://github.com/sublimelsp/LSP/pull/2618/commits/87cf18b481c63e818347ee319860ef96bc517647) :D 

Screenshot 2025-11-13 at 18 58 51
Screenshot 2025-11-13 at 18 58 30

plugin/hover.py
```diff
@@ -409,3 +409,10 @@ def _update_views_async(self, enable: bool) -> None:
                         session_view.view.settings().set(SHOW_DEFINITIONS_KEY, False)
                     else:
                         session_view.reset_show_definitions()
+
+
+class LspCopyTextCommand(sublime_plugin.TextCommand):
```
jwortmann: Should this be a WindowCommand, so that it could also be used in a HtmlSheet?
In that case I would also use `self.window.status_message` instead of `sublime.status_message`

predragnikolic:
Latest update after [c0957cb](https://github.com/sublimelsp/LSP/pull/2618/commits/c0957cb31a3fd4fa1d0c5506603d51dc255c0b2a)

Screenshot 2025-11-13 at 19 18 55
Screenshot 2025-11-13 at 19 18 42

predragnikolic:
FWIW, I do find the pngs approach nicer looking https://github.com/sublimelsp/LSP/pull/2618#issuecomment-3529043125
then the utf-char https://github.com/sublimelsp/LSP/pull/2618#issuecomment-3529109035

plugin/hover.py
predragnikolic: Done [226e125](https://github.com/sublimelsp/LSP/pull/2618/commits/226e1256c07e3a9b01c3a162077fc6850d9b41b4)

plugin/hover.py
predragnikolic: Done [226e125](https://github.com/sublimelsp/LSP/pull/2618/commits/226e1256c07e3a9b01c3a162077fc6850d9b41b4)

popups.css
```diff
@@ -86,3 +86,10 @@ pre.related_info {
     margin-top: 0.7rem;
     padding-top: 0.7rem;
 }
+a.copy-icon {
+    font-size: 1rem;
```
rchl: That's the default so doesn't need to be set.

predragnikolic:
We are getting close with this, we just need to decide 
png or utf-8

popups.css
predragnikolic: [fa851c0](https://github.com/sublimelsp/LSP/pull/2618/commits/fa851c06772f66c8175f5038823c0db4fab039fb)

predragnikolic:
I do find the png approach nicer looking (because the icon is a bit bigger, a nicer looking, but it cannot be customized with css, and it does require more code)

Based on the above messages, Rafal and Janos are leaning towards utf-8.
1 person is for png, 2 are for utf, so I guess the utf-8 char approach wins. But please confirm.

(also no rush in the decision)

plugin/core/views.py
```diff
@@ -831,13 +831,22 @@ def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, bas
...
     severity_class = DIAGNOSTIC_SEVERITY[diagnostic_severity(diagnostic) - 1][1]
     return _html_element("pre", html, class_name=severity_class, escape=False)
 
 
+def copy_icon_html(text_to_copy: str) -> str:
+    return f"""<a class='copy-icon'
+       title='Copy to clipboard'
+       href='{sublime.command_url('lsp_copy_text', {
+        'text': text_to_copy
+       })}'><span>⧉</span></a>"""
```
rchl: The extra span is not needed.

rchl:
> I do find the png approach nicer looking (because the icon is a bit bigger, a nicer looking, but it cannot be customized with css, and it does require more code)
> 
> Based on the above messages, Rafal and Janos are leaning towards utf-8. 1 person is for png, 2 are for utf, so I guess the utf-8 char approach wins. But please confirm.
> 
> (also no rush in the decision)

I agree that utf-8 character looks a bit worse due to the fact that at least on Mac it looks a bit blurry at default 1rem. It looks good at 1.2rem but then it makes line height taller... I don't see a way to fix that with limited minihtml capabilities.
Then font rendering is also very OS/font specific so it might look better (or worse ;)) for others.

I would still take utf-8 version for its simplicity and versatility.

plugin/core/views.py
predragnikolic: [478f64c](https://github.com/sublimelsp/LSP/pull/2618/commits/478f64c7c9fc134a62c9a426e3658b0d1c777779)

removed.

it was a leftover from a final attempt to increase the char font-size but not affect the line height of the outer box.

I wanted to make the `a` tag position relative, and the span absolute, 
and just increase the span font size, but it didn't work.

rchl: Right.

minihtml doesn't support position absolute. Or overflow.
https://www.sublimetext.com/docs/minihtml.html

rchl (approved):

jwortmann:
If I click the copy icon in the popup when hovering the gutter icon, a webbrowser (which is not even my default browser) opens...

predragnikolic:
Nice catch. Found the reason ->

`LSP/plugin/documents.py`
```
    def _on_navigate(self, href: str, point: int) -> None:
        if href.startswith('code-actions:'):
            ...
            else:
                self.handle_code_action_select(config_name, actions, 0)
        else:
            open_in_browser(href)
```

jwortmann:
The icon is a bit blurry on Windows too (I assume Segoe UI Symbol font), but I'd say it looks okay.

U+29C9

I tried some alternatives:

U+2BBA (less blurry, but slightly smaller):

U+2BBA

U+2BBB (hm... too bright):

U+2BBB

predragnikolic:
I did try one other approach. but no success. 
But for documentation purposes here it is:

To avoid the blur I did try to create the icon with css, 
with a combination of borders/ padding/inline-boxes/border radius... 

But minihtml doesn't behave as html and css, so I either break height of the outer box, or I break the look of the boxes itself.

Screenshot 2025-11-13 at 21 20 09

Screenshot 2025-11-13 at 21 19 54

plugin/core/views.py
```diff
@@ -831,13 +831,22 @@ def format_diagnostic_for_html(config: ClientConfig, diagnostic: Diagnostic, bas
             meta_info += "({})".format(
                 make_link(code_description["href"], str(code)) if code_description else text2html(str(code)))
         html += " " + _html_element("span", meta_info, class_name="color-muted", escape=False)
+    html += copy_icon_html(f"{diagnostic['message']} {f'({source})' if source else ''}".strip())
```
jwortmann: This function could be inlined, since it consists of only a single expression and is not used elsewhere:

```python
copy_text = f"{diagnostic['message']} {f'({source})' if source else ''}".strip()
html += f"""<a class='copy-icon' title='Copy to clipboard' href='{sublime.command_url(
        'lsp_copy_text', {'text': copy_text}
    )}'>⧉</a>"""
```

jwortmann (approved):
One thing I noticed is that the copied text can contain for example non-breaking space characters. But this happens because the server (in this case Pyright) already contains these characters as `\n\xa0\xa0` in the JSON payload, assumingly to enforce indentation after a linebreak, which would otherwise not get through the HTML rendering in the popup. Perhaps we should filter out/replace such characters? On the other hand that would feel a bit arbitrary and ultimatively it's the responsibility of the server what it puts into the diagnostic message.

predragnikolic:
>  Perhaps we should filter out/replace such characters? 

I had code for this, but I reverted it because later I could not find the exact example on why I added the code in the first place [baed32c](https://github.com/sublimelsp/LSP/pull/2618/commits/baed32c9920316e20a2ea5778d9c82ababb1229e) 

I am in favor of replacing such chars, I think it benefits the user.

predragnikolic:
Hello, if there are no further comments, I will merge this PR tonight.

PR #2619, opened 2025-08-19
Delete gello.graphql

predragnikolic:
I used this one for testing LSP-graphql and accidentally commited it.


predragnikolic:
Instead of a release, I did create a tag 4070-2.4.1
https://github.com/sublimelsp/LSP/releases/tag/4070-2.4.1

jwortmann:
I think Package Control only uses the tags to distribute package updates to users. It's not necessary to bother users with a new update just for this change, so I just deleted this tag, I hope that's okay. (According to the website the crawler shouldn't have seen that tag yet.)

PR #2620, opened 2025-08-21
Add on_configuration_loaded API in AbstractPlugin

rchl:
Add a better API for modifying the client connfiguration programmatically.

Before it was possible to modify the client configuration programmatically from `on_pre_start` but that's too late in case we want to modify the `selector` which decides whether plugin can get started in the first place.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -863,6 +863,25 @@ class AbstractPlugin(metaclass=ABCMeta):
         return sublime.load_settings(basename), filepath
 
     @classmethod
+    def on_configuration_loaded(cls, window: sublime.Window, workspace_folders: list[WorkspaceFolder],
+                                configuration: ClientConfig) -> None:
+        """
+        Notify when client configuration has loaded to allow for last-minute "configuration" adjustments.
+
+        Triggered initially on plugin initialization and later when the user explicity changes plugin configuration
+        within the Sublime settings or the project file. Note that the provided configuration object does not contain
+        any changes that might have been done previously from within this notification. Every time it's called the
+        configuration is re-read from scratch.
+
+        Prefer this notification over `on_pre_start` for adjusting "configuration".
+
+        :param      window:             The window
+        :param      workspace_folders:  The workspace folders
+        :param      configuration:      The configuration
+        """
+        return
+
+    @classmethod
     def additional_variables(cls) -> dict[str, str] | None:
         """
         In addition to the above variables, add more variables here to be expanded.
@@ -933,9 +952,10 @@ class AbstractPlugin(metaclass=ABCMeta):
                      workspace_folders: list[WorkspaceFolder], configuration: ClientConfig) -> str | None:
         """
         Callback invoked just before the language server subprocess is started. This is the place to do last-minute
-        adjustments to your "command" or "init_options" in the passed-in "configuration" argument, or change the
-        order of the workspace folders. You can also choose to return a custom working directory, but consider that a
-        language server should not care about the working directory.
+        adjustment to the order of the workspace folders. You can also choose to return a custom working directory,
+        but consider that a language server should not care about the working directory.
+
+        Prefer `on_configuration_loaded` for adjusting the "configuration" object.
 
         :param      window:             The window
         :param      initiating_view:    The initiating view
--- a/plugin/core/configurations.py
+++ b/plugin/core/configurations.py
@@ -21,8 +21,8 @@ RETRY_COUNT_TIMEDELTA = timedelta(minutes=3)
 class WindowConfigChangeListener(metaclass=ABCMeta):
 
     @abstractmethod
-    def on_configs_changed(self, config_name: str | None = None) -> None:
-        raise NotImplementedError()
+    def on_configs_changed(self, configs: list[ClientConfig]) -> None:
+        raise NotImplementedError
 
 
 class WindowConfigManager:
@@ -65,6 +65,7 @@ class WindowConfigManager:
     def _reload_configs(self, updated_config_name: str | None = None, notify_listeners: bool = False) -> None:
         project_data = self._window.project_data()
         project_settings = project_data.get("settings", {}).get("LSP", {}) if isinstance(project_data, dict) else {}
+        updated_configs: list[ClientConfig] = []
         if updated_config_name is None:
             self.all.clear()
         for name, config in self._global_configs.items():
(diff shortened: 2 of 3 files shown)


plugin/core/sessions.py
```diff
@@ -862,6 +862,25 @@ def configuration(cls) -> tuple[sublime.Settings, str]:
...
+        configuration is re-read from scratch.
+
+        Prefer this notification over `on_pre_start` for adjusting "configuration".
+
+        :param      window:             The window
+        :param      workspace_folders:  The workspace folders
+        :param      configuration:      The configuration
+        """
+        return
+
```
rchl: I've called it `on_configuration_loaded` instead of `on_configuration_changed`, for example, since it won't be called when configuration is changed programmatically (that would cause an infinite loop).

plugin/core/sessions.py
```diff
@@ -862,6 +862,25 @@ def configuration(cls) -> tuple[sublime.Settings, str]:
...
+                                configuration: ClientConfig) -> None:
+        """
+        Notify when client configuration has loaded to allow for last-minute "configuration" adjustments.
+
+        Triggered initially on plugin initialization and later when the user explicity changes plugin configuration
+        within the Sublime settings or the project file. Note that the provided configuration object does not contain
+        any changes that might have been done previously from within this notification. Every time it's called the
+        configuration is re-read from scratch.
+
+        Prefer this notification over `on_pre_start` for adjusting "configuration".
```
jwortmann: To me it looks like `on_pre_start` and this new method have a different purpose. As far as I can tell, this new method does not get called if I open another window with a different project. So if I want to adjust some config options that depend on the particular project, I still have to use `on_pre_start` (I use this for example in LSP-julia to let the language server know what the current project configuration file is - which for example declares the project's dependencies - or fallback to a "default" project if there is no such configuration file).

Also I wonder why a window and workspace folders get passed to this method - actually which exact window and workspace folders are these if I have multiple windows open and I change the config in my *LSP-foo.sublime-settings* file?

rchl: This is meant to get called separately for each window (and each AbstractPlugin instance of course). And it does seem to be the case here when opening ST with multiple windows open or opening a new project. Try adding `print(cls, window.id())` to this implementation and check the console. One case when it doesn't seem to happen is when opening a new blank window. Perhaps I can make this work too...

It also should get called for every open window when changing LSP-foo.sublime-settings file.

jwortmann: Okay, thanks. So if the workflow is that someone opens a new window via *File > New Window* and after that opens a project in that window, this function doesn't get called? Or would it be called "too early" with empty workspace folders?

Note that `on_pre_start` does also get called when the config in the settings file is modified (because the server gets restarted). The advantage of `on_configuration_loaded` seems to be that it allows to modify the "selector". But for modifying the other properties `on_pre_start` seems more convenient to me, because it has the workspace folders at the time when the server starts. So I would probably remove this line in the docstring (or better describe that its purpose is to allow modifications of "selector").

jwortmann: Or perhaps would it be better to just add another check after the `on_pre_start` call, to see if the selector is still applicable, and cancel the server start if it isn't? Then there wouldn't be two API methods needed which do almost the same thing.

rchl: > Or perhaps would it be better to just add another check after the `on_pre_start` call, to see if the selector is still applicable, and cancel the server start if it isn't? Then there wouldn't be two API methods needed which do almost the same thing.

I was going to write a response to address your points but perhaps I should cave in and do that if there are no downsides.

For the record, here is [my use case](https://github.com/sublimelsp/LSP-typescript/pull/274/files#diff-08655555cc53ca79554ba180d1ae7e6ebe5c9b0a25e1772ce248fdd8b83dab34R74)

One immediate thing that can't be done with the `on_pre_start` approach is modifying "selector" programmatically which would make my approach less flexible (currently it's read from a different plugin/package using `ResourcePath.glob_resources`)

jwortmann: I see. Right, that approach would not work if you want to make "selector" more general than what is specified in the settings file.

Then maybe an alternative would be to have a new method that can just modify the selector and gets called right at the time when we check which servers are applicable to a view. Something like
```python
@classmethod
def selector(cls, window: sublime.Window, config: ClientConfig) -> str:
    return config.selector
```

And then in
https://github.com/sublimelsp/LSP/blob/4651aa1b03f64363769b25f8ad685e4a7219e42f/plugin/core/types.py#L851-L855

we do something like
```python
from .sessions import get_plugin

def match_view(self, view: sublime.View, scheme: str) -> bool:
    # ...
    plugin = get_plugin(self.name)
    selector = plugin.selector(view.window(), self) if plugin else self.selector.strip()
```

rchl: did that

jwortmann: Okay, I think this is better than having two methods with almost the exact same purpose. I now realize that this still has some overlap with the `should_ignore` method. Probably the best would be to simply move the whole matching logic into the API method, so that plugins can have full control over it.

```python
def is_applicable(self, view: sublime.View, config: ClientConfig) -> bool:
    """ Override whether the language server should attach to the given view. """
    return view.match_selector(config.selector) and get_uri_scheme(view) in config.schemes
``` 

But I guess it's too late to change this now. Maybe that would be something to consider for an `AbstractPluginV2`.

plugin/core/types.py
```diff
@@ -849,10 +849,12 @@ def erase_view_status(self, view: sublime.View) -> None:
         view.erase_status(self.status_key)
 
     def match_view(self, view: sublime.View, scheme: str) -> bool:
+        from .sessions import get_plugin
```
jwortmann, line 852: I assume this is here to avoid cyclic imports?

rchl: Yep.
I've tried to fix it by moving code around but it's just impossible without more serious refactoring.

plugin/core/sessions.py
rchl: Would agree to leave it for next API version.

The logic should probably not live in plugin-specific code though since it's not only used by plugins.

jwortmann: Or do you think that approach would be better? I assume that `should_ignore` is currently only used by LSP-Copilot (to check whether a file is listed in a `.copilotignore` configuration file). So it might be possible to still deprecate `should_ignore` and strive for a cleaner and most flexible API.

jwortmann: Okay then I guess we can try with the `selector` method for now

predragnikolic: > So it might be possible to still deprecate should_ignore and strive for a cleaner and most flexible API.

+1 , if one plugin is affected can create a PR for LSP-copilot

PR #2623, opened 2025-08-25
fix: skip diagnostics pull if diagnostics already received

### LINKED ISSUE #2609: LSP server goes into infinite loop with multiple views of the same file
**Describe the bug**

As reported in [LSP-ruff issue sublimelsp/LSP#2621](https://github.com/sublimelsp/LSP/issues/2621), the LSP server can get into an infinite loop on occasion. This seems to be easily triggered by opening the same file multiple times. I can see this using either the LSP-json or LSP-ruff servers, so it doesn't appear to be server-specific.

**To Reproduce**

1. Install either the LSP-ruff or LSP-json server.
2. Open an appropriate file (.py or .json).
3. Open a second view to that file (I typically do this using the Origami plugin, but you can also use the _File > Split View_ menu option).
4. Open the LSP log panel.
5. Note the output going crazy, looping forever.

**Expected behavior**

LSP should be capable of handling multiple views of the same file without getting into an infinite loop.

**Screenshots**

See the screencast shared in [LSP-ruff issue sublimelsp/LSP#2621](https://github.com/sublimelsp/LSP/issues/2621) for an example.

**Environment**

- OS: Windows 11 Pro
- Sublime Text version: 4200
- LSP version: 2.3.0
- Language servers used: ruff, json

rchl:
Skip `textDocument/diagnostic` request if we have already received diagnostics for the current view version.

For servers that support pull diagnostics, we would re-ask for diagnostics on each selection change (through the code that requests code actions), even if the view hasn't changed.

In the case of the bug linked below, this would even trigger an infinite loop as the cache in the `CodeActionsManager` that caches the response would not cut the loop short if two views into the same file are opened and selection range is different in each.

Fixes #2609

### DIFF
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -128,7 +128,7 @@ class SessionBuffer:
         self._pending_changes: PendingChanges | None = None
         self.diagnostics: list[tuple[Diagnostic, sublime.Region]] = []
         self.diagnostics_data_per_severity: dict[tuple[int, bool], DiagnosticSeverityData] = {}
-        self.diagnostics_version = -1
+        self._diagnostics_version = -1
         self.diagnostics_flags = 0
         self._diagnostics_are_visible = False
         self.document_diagnostic_needs_refresh = False
@@ -496,8 +496,13 @@ class SessionBuffer:
             return
         if mgr.should_ignore_diagnostics(self._last_known_uri, self.session.config):
             return
+        change_count = view.change_count()
         if version is None:
-            version = view.change_count()
+            version = change_count
+        elif version < change_count:
+            return
+        if version == self._diagnostics_version:
+            return
         if self._document_diagnostic_pending_request:
             if self._document_diagnostic_pending_request.version == version and not forced_update:
                 return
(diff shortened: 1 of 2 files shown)


plugin/session_buffer.py
```diff
@@ -496,8 +496,13 @@ def do_document_diagnostic_async(
...
         if mgr.should_ignore_diagnostics(self._last_known_uri, self.session.config):
             return
+        change_count = view.change_count()
         if version is None:
-            version = view.change_count()
+            version = change_count
+        elif version < change_count:
+            return
+        if version == self._diagnostics_version:
+            return
```
rchl, line 505: I've identified cases (by furiously editing the document and undoing changes) where this code would run with `version` that is already behind `view.change_count()`. Thus the added `version < change_count` check. Happens because there can be some delay between document change and this triggering.

plugin/session_buffer.py
```diff
@@ -520,18 +525,17 @@ def _on_document_diagnostic_async(self, version: int, response: DocumentDiagnost
...
-            sb = self.session.get_session_buffer_for_uri_async(uri)
-            if sb:
-                cast(SessionBuffer, sb)._apply_document_diagnostic_async(
-                    None, cast(DocumentDiagnosticReport, diagnostic_report))
+        if 'relatedDocuments' in response:
+            for uri, diagnostic_report in response['relatedDocuments'].items():
+                sb = self.session.get_session_buffer_for_uri_async(uri)
+                if sb:
+                    cast(SessionBuffer, sb)._apply_document_diagnostic_async(
+                        None, cast(DocumentDiagnosticReport, diagnostic_report))
```
rchl, line 538: This code was just plain wrong before (missing `.items()`) and would crash on non-empty `relatedDocuments`.

jwortmann (approved):

plugin/session_buffer.py
predragnikolic: Even if `version` is less then `change_count`
Instead of returning early, should LSP still make a pull diagnostic request, because the change count differ?

PR #2634, opened 2025-09-05
Don't show error dialog for failed workspace command

jwortmann:
The error dialog that occurs when trying to run `workspace/executeCommand` (for example from a code lens) with a command that is not implemented on the server side is pretty annoying, so let's print to the status bar instead.

I wonder if we should maybe even prevent the request for commands that are not listed in the server capabilities (from initialize response or dynamic registration), and directly print a "command not implemented" message instead. Or are there any servers which handle commands that are not advertised its capabilities?

### DIFF
--- a/plugin/execute_command.py
+++ b/plugin/execute_command.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+from .core.logging import debug
 from .core.protocol import Error
 from .core.protocol import ExecuteCommandParams
 from .core.registry import LspTextCommand
@@ -58,7 +59,11 @@ class LspExecuteCommand(LspTextCommand):
         :param error: The Error object.
         :param command_name: The name of the command that was executed.
         """
-        sublime.message_dialog(f"command {command_name} failed. Reason: {str(error)}")
+        msg = f"command {command_name} failed: {str(error)}"
+        debug(msg)
+        window = self.view.window()
+        if window:
+            window.status_message(msg)
 
     def _expand_variables(self, command_args: list[Any]) -> list[Any]:
         view = self.view


rchl (approved):

predragnikolic:
> I wonder if we should maybe even prevent the request for commands that are not listed in the server capabilities (from initialize response or dynamic registration), and directly print a "command not implemented" message instead. Or are there any servers which handle commands that are not advertised its capabilities?

What you wrote doesn't violate the spec.

I don't mind seeing text in the status bar if a server cannot handle something,
as the status bar is less distracting than a dialog.

There might be some servers that have forgot to announce `commands`... (but I would consider that the server in that case is not following the spec).

If I compare the strict approach vs the (now) less strict approach. 
I would pick the less strict approach, 
but will not interfere that much if you created a PR that introduces the strict approach.

jwortmann:
Yes, I think according to the specs, the server should list all commands that it can handle server side in the [ExecuteCommandOptions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#executeCommandOptions):
```ts
export interface ExecuteCommandOptions extends WorkDoneProgressOptions {
	/**
	 * The commands to be executed on the server
	 */
	commands: string[];
}
```

So we actually know which commands are handled by the server and which are expected to be implemented on the plugin side. And therefore we don't really need to try to run `executeCommand` from a code lens (or code action) with a command that is not supported directly by the server, because it will just return the response error.

Currently I see "command ... failed: method not found (-32601)" in the status bar.
And I thought that something like "command ... not implemented" would be less confusing.

But it's not very important, so I agree that we can keep using the less strict approach for now.

PR #2636, opened 2025-09-07
Limit length of text in completion popup details pane

jwortmann:
If the `CompletionItem.detail` returned from the server is very long, the completion popup can span the entire window and can drastically change its size while typing, which is quite distracting. This PR limits the maximum length of the text to 80 characters.

Tested with https://github.com/sublimelsp/LSP-Tinymist

before:

before

after:

after


jwortmann:
It seems like some time in the past ST would automaticall fade out long text in the details pane: see e.g. screenshots in https://github.com/sublimehq/sublime_text/issues/4048. But apparently this is not the case anymore.

*Edit:* Oh I see that the fade out still happens if the window is very small. So I think the popup behavior was always like this, and manually restricting the maximum length should be the right thing to do.

rchl (approved):

PR #2643, opened 2025-09-18
fix: respect `refactoring_auto_save` in more cases

rchl:
### `code_actions.py`
- the logic should be unchanged here. I just had to extract the "kind includes other kind" logic out of this code and place it somewhere where `Session` could access it.
### `protocol.py`
- added `kind_includes_other_kind` function
### `sessions.py`
  - Fixed a bug where a refactoring code action would not always trigger save with `refactoring_auto_save` enabled. It only worked if the code action had `refactoring` kind exactly and not if it had more specific one like `refactoring.move`. This is fixed by the new `kind_includes_other_kind` function.
  - Allow passing `is_refactoring` to `Session.execute_command` to allow for all edits triggered within this request to be treated as refactoring and trigger the `refactoring_auto_save` logic.  (see https://github.com/sublimelsp/LSP-typescript/pull/278 to see the code that uses that)

### DIFF
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -6519,16 +6519,3 @@ CodeLensExtended = TypedDict('CodeLensExtended', {
 
 # Temporary for backward compatibility with LSP packages.
 RangeLsp = Range
-
-
-def kind_contains_other_kind(kind: str, other_kind: str) -> bool:
-    """
-    Check if `other_kind` is a sub-kind of `kind`.
-
-    The kind `"refactor.extract"` for example contains `"refactor.extract"` and ``"refactor.extract.function"`,
-    but not `"unicorn.refactor.extract"`, or `"refactor.extractAll"` or `refactor`.
-    """
-    if kind == other_kind:
-        return True
-    kind_len = len(kind)
-    return len(other_kind) > kind_len and other_kind.startswith(kind + '.')
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -6521,11 +6521,14 @@ CodeLensExtended = TypedDict('CodeLensExtended', {
 RangeLsp = Range
 
 
-def kind_includes_other_kind(kind: str, other_kind: str) -> bool:
+def kind_contains_other_kind(kind: str, other_kind: str) -> bool:
     """
-    Return True if "kind" includes "other_kind". For example: kind "a.b" includes kind "a" and "a.b" but not "a.b.c".
+    Check if `other_kind` is a sub-kind of this `kind`.
+
+    The kind `"refactor.extract"` for example contains `"refactor.extract"` and ``"refactor.extract.function"`,
+    but not `"unicorn.refactor.extract"`, or `"refactor.extractAll"` or `refactor`.
     """
     if kind == other_kind:
         return True
-    other_kind_len = len(other_kind)
-    return len(kind) > other_kind_len and kind.startswith(other_kind) and kind[other_kind_len] == '.'
+    kind_len = len(kind)
+    return len(other_kind) > kind_len and other_kind.startswith(kind + '.')
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -6519,3 +6519,14 @@ CodeLensExtended = TypedDict('CodeLensExtended', {
 
 # Temporary for backward compatibility with LSP packages.
 RangeLsp = Range
+
+
+def kind_includes_other_kind(kind: str, other_kind: str) -> bool:
+    """
+    The Code Action "kind" includes "other_kind" if "kind" matches "other_kind" exactly or "other_kind" is a subset
+    of "kind".
+    """
+    if kind == other_kind:
+        return True
+    other_kind_len = len(other_kind)
+    return len(kind) > other_kind_len and kind.startswith(other_kind) and kind[other_kind_len] == '.'
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -6523,7 +6523,7 @@ RangeLsp = Range
 
 def kind_contains_other_kind(kind: str, other_kind: str) -> bool:
     """
-    Check if `other_kind` is a sub-kind of this `kind`.
+    Check if `other_kind` is a sub-kind of `kind`.
 
     The kind `"refactor.extract"` for example contains `"refactor.extract"` and ``"refactor.extract.function"`,
     but not `"unicorn.refactor.extract"`, or `"refactor.extractAll"` or `refactor`.
--- a/plugin/core/protocol.py
+++ b/plugin/core/protocol.py
@@ -6523,8 +6523,7 @@ RangeLsp = Range
 
 def kind_includes_other_kind(kind: str, other_kind: str) -> bool:
     """
-    The Code Action "kind" includes "other_kind" if "kind" matches "other_kind" exactly or "other_kind" is a subset
-    of "kind".
+    Return True if "kind" includes "other_kind". For example: kind "a.b" includes kind "a" and "a.b" but not "a.b.c".
     """
     if kind == other_kind:
         return True
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1680,7 +1680,8 @@ class Session(TransportCallbacks):
 
             sublime.set_timeout_async(run_async)
             return Promise.resolve(None)
-        self._is_executing_refactoring_command = is_refactoring
+        if is_refactoring:
+            self._is_executing_refactoring_command = True
         # TODO: Our Promise class should be able to handle errors/exceptions
         execute_command = Promise(
             lambda resolve: self.send_request(
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -46,7 +46,7 @@ from .protocol import InitializeError
 from .protocol import InitializeParams
 from .protocol import InitializeResult
 from .protocol import InsertTextMode
-from .protocol import kind_includes_other_kind
+from .protocol import kind_contains_other_kind
 from .protocol import Location
 from .protocol import LocationLink
 from .protocol import LogMessageParams
@@ -1706,7 +1706,7 @@ class Session(TransportCallbacks):
             arguments = code_action.get('arguments', None)
             if isinstance(arguments, list):
                 command_params['arguments'] = arguments
-            is_refactoring = kind_includes_other_kind(code_action.get('kind', ''), CodeActionKind.Refactor)
+            is_refactoring = kind_contains_other_kind(CodeActionKind.Refactor, code_action.get('kind', ''))
             return self.execute_command(command_params, progress=progress, view=view, is_refactoring=is_refactoring)
         # At this point it cannot be a command anymore, it has to be a proper code action.
         # A code action can have an edit and/or command. Note that it can have *both*. In case both are present, we
@@ -1836,7 +1836,7 @@ class Session(TransportCallbacks):
             self.window.status_message(f"Failed to apply code action: {code_action}")
             return Promise.resolve(None)
         edit = code_action.get("edit")
-        is_refactoring = kind_includes_other_kind(code_action.get('kind', ''), CodeActionKind.Refactor)
+        is_refactoring = kind_contains_other_kind(CodeActionKind.Refactor, code_action.get('kind', ''))
         promise = self.apply_workspace_edit_async(edit, is_refactoring) if edit else Promise.resolve(None)
         command = code_action.get("command")
         if command is not None:
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1689,7 +1689,8 @@ class Session(TransportCallbacks):
                 lambda err: resolve(Error(err["code"], err["message"], err.get("data")))
             )
         )
-        execute_command.then(lambda _: self._reset_is_executing_refactoring_command())
+        if is_refactoring:
+            execute_command.then(lambda _: self._reset_is_executing_refactoring_command())
         return execute_command
 
     def _reset_is_executing_refactoring_command(self) -> None:
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1680,8 +1680,6 @@ class Session(TransportCallbacks):
 
             sublime.set_timeout_async(run_async)
             return Promise.resolve(None)
-        if is_refactoring:
-            self._is_executing_refactoring_command = True
         # TODO: Our Promise class should be able to handle errors/exceptions
         execute_command = Promise(
             lambda resolve: self.send_request(
@@ -1691,6 +1689,7 @@ class Session(TransportCallbacks):
             )
         )
         if is_refactoring:
+            self._is_executing_refactoring_command = True
             execute_command.then(lambda _: self._reset_is_executing_refactoring_command())
         return execute_command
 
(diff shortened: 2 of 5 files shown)


plugin/core/sessions.py
```diff
@@ -1650,7 +1652,8 @@ def _template_variables(self) -> dict[str, str]:
         return variables
 
     def execute_command(
-        self, command: ExecuteCommandParams, progress: bool, view: sublime.View | None = None
+        self, command: ExecuteCommandParams, *, progress: bool, view: sublime.View | None = None,
+        is_refactoring: bool = False,
```
rchl: Using `*` here is technically a breaking change but my searching shows that nobody uses it currently in a way that would cause it to break.

Feel free to double check (especially third party LSP packages).

jwortmann: I think `progress` should have a default value (preferrably `False`).

I just recently used this a few times in a new plugin https://github.com/sublimelsp/LSP-Tinymist/blob/main/plugin.py and was surprised that this function requires passing at least two arguments. That code would break when changing to keyword-only arguments, but I can easily adjust it if all keyword arguments get default values.

I agree that it feels "right" that the other arguments should be keyword-only arguments.

plugin/core/sessions.py
```diff
@@ -1317,6 +1318,7 @@ def __init__(self, manager: Manager, logger: Logger, workspace_folders: list[Wor
         self._plugin: AbstractPlugin | None = None
         self._status_messages: dict[str, str] = {}
         self._semantic_tokens_map = get_semantic_tokens_map(config.semantic_tokens)
+        self._is_executing_refactoring_command = False
```
jwortmann, line 1321: I've only taken a first glance over the code in this PR but this kind of screams for a race condition. This value gets set before a request gets send and is reset when handling the response (I think). But the requests are async and there could be multiple open requests of this type at the same time. So this should probably be a local variable for each particular request instead.

rchl: I think that it's not likely that there would be multiple execute commands running at the same time.

Also I don't see how it would work to have this variable per-request. This is relying on the fact that server will trigger workspace edit during execution of executeCommand but there is no real way to tell that specific edits belong to that executeCommand other than just assuming that anything executed before the executeCommand has finished is a part of it.

Here is example log to paint a picture of what happens:

```
:: [20:13:56.399] --> LSP-typescript workspace/executeCommand (17): {'command': '_typescript.applyRefactoring', 'arguments': [{'file': '/usr/local/workspace/github/typescript-language-server/src/tsServer/logDirectoryProvider.ts', 'startLine': 15, 'startOffset': 30, 'endLine': 15, 'endOffset': 30, 'refactor': 'Move to a new file', 'action': 'Move to a new file'}]}
:: [20:13:56.418] <-- LSP-typescript workspace/applyEdit (2): {'edit': {'changes': {'file:///usr/local/workspace/github/typescript-language-server/src/tsServer/logDirectoryProvider.ts': [{'range': {'start': {'line': 13, 'character': 0}, 'end': {'line': 13, 'character': 0}}, 'newText': "import { ILogDirectoryProvider } from './ILogDirectoryProvider.js';\n"}, {'range': {'start': {'line': 14, 'character': 0}, 'end': {'line': 18, 'character': 0}}, 'newText': ''}], 'file:///usr/local/workspace/github/typescript-language-server/src/tsServer/spawner.ts': [{'range': {'start': {'line': 17, 'character': 0}, 'end': {'line': 18, 'character': 0}}, 'newText': ''}, {'range': {'start': {'line': 18, 'character': 0}, 'end': {'line': 18, 'character': 0}}, 'newText': "import type { ILogDirectoryProvider } from './ILogDirectoryProvider.js';\n"}], 'file:///usr/local/workspace/github/typescript-language-server/src/ts-client.ts': [{'range': {'start': {'line': 22, 'character': 0}, 'end': {'line': 23, 'character': 0}}, 'newText': ''}, {'range': {'start': {'line': 23, 'character': 0}, 'end': {'line': 23, 'character': 0}}, 'newText': "import type { ILogDirectoryProvider } from './tsServer/ILogDirectoryProvider.js';\n"}], 'file:///usr/local/workspace/github/typescript-language-server/src/tsServer/ILogDirectoryProvider.ts': [{'range': {'start': {'line': 0, 'character': 0}, 'end': {'line': 0, 'character': 0}}, 'newText': '\nexport interface ILogDirectoryProvider {\n    getNewLogDirectory(): string | undefined;\n}\n'}]}}}
:: [20:13:56.492]  -> LSP-typescript textDocument/didOpen: <params with 9655 characters>
:: [20:13:56.600]  -> LSP-typescript textDocument/didOpen: {'textDocument': {'uri': 'file:///usr/local/workspace/github/typescript-language-server/src/tsServer/ILogDirectoryProvider.ts', 'languageId': 'typescript', 'version': 0, 'text': ''}}
:: [20:13:56.709]  -> LSP-typescript textDocument/didOpen: <params with 25123 characters>
:: [20:13:56.775] >>> LSP-typescript (2) (duration: 356ms): {'applied': True}
:: [20:13:56.786]  -> LSP-typescript textDocument/didChange: {'textDocument': {'uri': 'file:///usr/local/workspace/github/typescript-language-server/src/tsServer/logDirectoryProvider.ts', 'version': 6}, 'contentChanges': [{'range': {'start': {'line': 14, 'character': 0}, 'end': {'line': 18, 'character': 0}}, 'rangeLength': 90, 'text': ''}, {'range': {'start': {'line': 13, 'character': 0}, 'end': {'line': 13, 'character': 0}}, 'rangeLength': 0, 'text': "import { ILogDirectoryProvider } from './ILogDirectoryProvider.js';\n"}]}
:: [20:13:56.786]  -> LSP-typescript textDocument/didSave: {'textDocument': {'uri': 'file:///usr/local/workspace/github/typescript-language-server/src/tsServer/logDirectoryProvider.ts'}}
:: [20:13:56.802] --> LSP-typescript textDocument/codeAction (18): {'textDocument': {'uri': 'file:///usr/local/workspace/github/typescript-language-server/src/tsServer/logDirectoryProvider.ts'}, 'range': {'start': {'line': 15, 'character': 0}, 'end': {'line': 15, 'character': 0}}, 'context': {'diagnostics': [], 'triggerKind': 2}}
:: [20:13:56.804]  -> LSP-typescript textDocument/didClose: {'textDocument': {'uri': 'file:///usr/local/workspace/github/typescript-language-server/src/tsServer/spawner.ts'}}
:: [20:13:56.822]  -> LSP-typescript textDocument/didClose: {'textDocument': {'uri': 'file:///usr/local/workspace/github/typescript-language-server/src/tsServer/ILogDirectoryProvider.ts'}}
:: [20:13:56.833]  -> LSP-typescript textDocument/didChange: {'textDocument': {'uri': 'file:///usr/local/workspace/github/typescript-language-server/src/ts-client.ts', 'version': 2}, 'contentChanges': [{'range': {'start': {'line': 23, 'character': 0}, 'end': {'line': 23, 'character': 0}}, 'rangeLength': 0, 'text': "import type { ILogDirectoryProvider } from './tsServer/ILogDirectoryProvider.js';\n"}, {'range': {'start': {'line': 22, 'character': 0}, 'end': {'line': 23, 'character': 0}}, 'rangeLength': 81, 'text': ''}]}
:: [20:13:56.833]  -> LSP-typescript textDocument/didSave: {'textDocument': {'uri': 'file:///usr/local/workspace/github/typescript-language-server/src/ts-client.ts'}}
:: [20:13:56.882]  -> LSP-typescript textDocument/didClose: {'textDocument': {'uri': 'file:///usr/local/workspace/github/typescript-language-server/src/ts-client.ts'}}
:: [20:13:56.884] <<< LSP-typescript (17) (duration: 484ms): None
```

jwortmann: Hm, okay then it's not so trivial and it looks like this is the only way to handle it. But it also assumes that the `workspace/applyEdit` request is received before the response for `workspace/executeCommand`. I don't see anything in the specs that guarantees this order. So the implemented logic would not work if the order is different. But I've also just tested with the Julia server and it uses the same order as LSP-typescript.

plugin/core/protocol.py
```diff
@@ -6519,3 +6519,14 @@ def to_lsp(self) -> Position:
 
 # Temporary for backward compatibility with LSP packages.
 RangeLsp = Range
+
+
+def kind_includes_other_kind(kind: str, other_kind: str) -> bool:
+    """
+    The Code Action "kind" includes "other_kind" if "kind" matches "other_kind" exactly or "other_kind" is a subset
+    of "kind".
```
jwortmann: This description is somewhat misleading. I would say "refactor.rewrite" is a subset of "refactor" (because "refactor.rewrite" is more specific). So it's the other way round. It's a bit confusing because the "bigger" kind string is a substring of the more specific kind string.

plugin/core/sessions.py
rchl: As long as it works in more cases I'm happy :)

I think it should be pretty natural to assume that server will handle the whole command and then send the response. It would probably have to get out of its way to do otherwise.

plugin/core/protocol.py
rchl: I struggle with describing this properly. Hard to make it clear without also describing the whole code action matching logic and how code actions are structured...

rchl: Actually the naming for this function might be wrong/confusing.

In JS the language-server libraries have similar logic and they have `contains` function which actually does the opposite of this function.

```ts
    /**
     * Checks if `other` is a sub-kind of this `CodeActionKind`.
     *
     * The kind `"refactor.extract"` for example contains `"refactor.extract"` and ``"refactor.extract.function"`,
     * but not `"unicorn.refactor.extract"`, or `"refactor.extractAll"` or `refactor`.
     *
     * @param other Kind to check.
     */
    public contains(other: CodeActionKind): boolean {
        return this.equals(other) || this.value === '' || other.value.startsWith(this.value + CodeActionKind.sep);
    }


CodeActionKind.QuickFix.contains('quickfix.foo') === true
```

rchl: Changed the code to match those libs.

jwortmann: I think the "includes" was from the previous function which had an array of kinds for the input.

Maybe rename like this

```python
def is_subkind(base_kind: str, kind: str) -> bool:
    """
    Return `True` if `kind` is a subkind of `base_kind`.
    For example "a.b" is a subkind of "a" and of "a.b".
    """
```

rchl: Are you happy with matching the Microsoft's libraries?

I find the short `is_subkind` not ideal for two reasons:
- it doesn't make it clear that it's related to code action kinds unless you see the parameters
- "subkind" is a bit confusing wording in case of kinds. One would think that "a" is a subkind of "a.b" (in the sense that "a.b" includes "a" when thinking about it as a string) but it's the opposite.

jwortmann: Yeah it's okay. But could you move this function out of `protocol.py`? In my opinion ideally this file should only contain the auto-generated types from the specs (I know there is some other stuff too, but let's not keep it growing if possible). Maybe put it in the "utils" file `views.py` or in `sessions.py`.

rchl: `utils` is exactly what I was planning to do first but was a bit afraid of creating a file for any random stuff. Alternatively was also thinking about creating `core/code_actions` but then there would be two `code_actions`.

Maybe I'm overthinking it.

jwortmann: We already have the `utils` file, it is just named `views.py`. So let's just use that for now.

Perhaps some time in the future we can do a bigger refactoring of the files structure. But first we need a proper API for plugins (what currently is `LSP/plugin/__init__.py`, but then including all of the functionality that can be used by plugins - and then fix all the imports in plugins that currently import some random things from various files directly). And then I would also move (only) the auto-generated types from `LSP/plugin/core/protocol.py` to `LSP/protocol/__init__.py` and add a line `from ...protocol import *` into `LSP/plugin/core/protocol.py` for backward compatibility. So that we have a clean didvision between "official" LSP protocol types and the custom classes and types defined by this package. Then there would only be two API "endpoints" for imports in helper plugins: e.g. `from LSP.protocol import TextDocumentPositionParams` for the LSP types and `from LSP.plugin import Request` for all other things.

plugin/core/sessions.py
```diff
@@ -1677,14 +1680,20 @@ def run_async() -> None:
 
             sublime.set_timeout_async(run_async)
             return Promise.resolve(None)
+        self._is_executing_refactoring_command = is_refactoring
```
jwortmann:
```suggestion
        if is_refactoring:
            self._is_executing_refactoring_command = True
```

This should be a bit more correct in case there would be multiple different commands open at the same time. So if another non-refactoring command is run very quickly after the refactoring command (before the refactoring response), this variable doesn't get reset accidentally.

For example LSP-Tinymist sends `workspace/executeCommand` whenever the cursor position changes, to synchronize a browser preview scrolling position with the cursor position in the file. Though afaik the server doesn't have any refactoring commands, so I can't create a concrete example to test this.

plugin/core/sessions.py
```diff
@@ -1677,14 +1680,20 @@ def run_async() -> None:
...
                 Request("workspace/executeCommand", command, None, progress),
                 resolve,
                 lambda err: resolve(Error(err["code"], err["message"], err.get("data")))
             )
         )
+        execute_command.then(lambda _: self._reset_is_executing_refactoring_command())
+        return execute_command
+
+    def _reset_is_executing_refactoring_command(self) -> None:
+        self._is_executing_refactoring_command = False
```
jwortmann, line 1696:
```suggestion
        execute_command.then(lambda _: self._reset_is_executing_refactoring_command(is_refactoring))
        return execute_command

    def _reset_is_executing_refactoring_command(self, is_refactoring: bool) -> None:
        if is_refactoring:
            self._is_executing_refactoring_command = False
```

Only need to reset if this value was activated by this command (see comment above).

plugin/core/sessions.py
```diff
@@ -1677,14 +1680,20 @@ def run_async() -> None:
...
         # TODO: Our Promise class should be able to handle errors/exceptions
-        return Promise(
+        execute_command = Promise(
             lambda resolve: self.send_request(
                 Request("workspace/executeCommand", command, None, progress),
                 resolve,
                 lambda err: resolve(Error(err["code"], err["message"], err.get("data")))
             )
         )
+        execute_command.then(lambda _: self._reset_is_executing_refactoring_command())
```
jwortmann:
```suggestion
        if is_refactoring:
            execute_command.then(lambda _: self._reset_is_executing_refactoring_command())
```

Only need to reset if this value was activated by this command (see comment above).

plugin/core/sessions.py
```diff
@@ -1677,14 +1680,22 @@ def run_async() -> None:
...
-        return Promise(
+        execute_command = Promise(
             lambda resolve: self.send_request(
                 Request("workspace/executeCommand", command, None, progress),
                 resolve,
                 lambda err: resolve(Error(err["code"], err["message"], err.get("data")))
             )
         )
+        if is_refactoring:
+            execute_command.then(lambda _: self._reset_is_executing_refactoring_command())
```
jwortmann, line 1693: And a small style simplification: move the other line to here too and remove the two lines from above
```suggestion
        if is_refactoring:
            self._is_executing_refactoring_command = True
            execute_command.then(lambda _: self._reset_is_executing_refactoring_command())
```

Feel free to merge afterwards

plugin/core/sessions.py
```diff
@@ -1677,14 +1680,22 @@ def run_async() -> None:
 
             sublime.set_timeout_async(run_async)
             return Promise.resolve(None)
+        if is_refactoring:
+            self._is_executing_refactoring_command = True
```
jwortmann:
```suggestion
```

jwortmann (approved):

PR #2648, opened 2025-09-29
Ensure document state is synced before requesting diagnostics

rchl:
I've noticed a bug where we would see stale diagnostics. This has triggered for me with LSP-eslint (pull diagnostics is a prerequisite for being able to reproduce).

Quickly making few edits could result in seeing outdated diagnostics. This was because of the code actions request triggering on selection change. That code has forced diagnostic request without specifying version so it has always requested diagnostics for the latest document version but at that point the latest changes might not have been synced to the server yet (they were still pending). That meant that once document state was synced later, the corresponding diagnostics request was ignored since we've already requested diagnostics for that version.

Force a `purge_changes_async()` call before getting diagnostics to ensure latest state.
Also made `version` argument required in `do_document_diagnostic_async()` so that one has to think about passing it correctly when changing/adding calls.


plugin/session_buffer.py
```diff
@@ -488,18 +488,14 @@ def update_document_link(self, new_link: DocumentLink) -> None:
...
         mgr = self.session.manager()
         if not mgr or not self.has_capability("diagnosticProvider"):
             return
         if mgr.should_ignore_diagnostics(self._last_known_uri, self.session.config):
             return
         change_count = view.change_count()
-        if version is None:
-            version = change_count
-        elif version < change_count:
+        if version < change_count:
```
jwortmann: You can combine these 2 lines into 1 now (no need for a variable if it is only used once).

jwortmann: Could also combine it with the next condition below to save 2 more lines

Issue #2649, opened 2025-09-30
helm-ls crashing constantly with could not find document file


shubham-dogra-s1:
# Troubleshooting: helm-ls
## Version
 - LSP: 2.6.0
 - Sublime Text: 4200
## Server Test Run
 - exit code: 0
 - output
```
{"@level":"info","@message":"helm-ls: connections opened","@timestamp":"2025-09-30T10:07:41+05:30"}
```
## Server Configuration
 - command
```json
[
  "helm_ls",
  "serve"
]
```
 - shell command
```sh
helm_ls serve
```
 - selector
```
source.yaml.go
```
 - priority_selector
```
source.yaml.go
```
 - init_options
```json
{}
```
 - settings
```json
{}
```
 - env
```json
{}
```

## Active view
 - File name
```
/Users/shubham.dogra/Desktop/test/charts/alerts/templates/configmap.yaml
```
 - Settings
```json
{
  "auto_complete_selector": "source - source.go - source.java - source.js - source.py",
  "lsp_active": true,
  "syntax": "Packages/Go/YAML (Go).sublime-syntax"
}
```
 - base scope
```
source.yaml.go
```

## Project / Workspace
 - folders
```json
[
  "/Users/shubham.dogra/Desktop/test"
]
```
 - is project: False

## LSP configuration

```json
{
  "clients": {
    "LSP-typescript": {
      "enabled": true
    },
    "gopls": {
      "disabled_capabilities": {
        "hover": "true",
        "signatureHelpProvider": true
      },
      "enabled": true,
      "selector": "source.go",
      "settings": {
        "gopls": {
          "completeUnimported": false
        }
      }
    },
    "helm-ls": {
      "command": [
        "helm_ls",
        "serve"
      ],
      "enabled": true,
      "helmLint": {
        "enabled": true
      },
      "selector": "source.yaml.go"
    },
    "jdtls": {
      "disabled_capabilities": {
        "signatureHelpProvider": true
      },
      "enabled": true,
      "selector": "source.java",
      "settings": {
        "java.compile.outputPath": "build",
        "java.completion.enabled": true,
        "java.completion.guessMethodArguments": true,
        "java.completion.maxResults": 50,
        "java.project.outputPath": "build"
      }
    }
  },
  "diagnostics_delay_ms": 2000,
  "document_highlight_style": "",
  "inhibit_snippet_completions": true,
  "inhibit_word_completions": true,
  "link_highlight_style": "none",
  "log_debug": true,
[...]

PR #2653, opened 2025-10-02
Add source.yaml.go => helm language ID mapping

rchl:


jwortmann:
See my comment in https://github.com/sublimelsp/LSP/issues/2649#issuecomment-3363177688

Ideally someone would provide a special syntax for helm template files with base scope `source.yaml.pipeline.helm`, and then we use that scope here in the language ID mapping.

rchl:
I agree. Dropping this in favor of someone creating a dedicated helm syntax.

predragnikolic:
source.yaml.go is the correct syntax for helm .
I just checked this on Sublime Text discord.

rchl:
While it's true that the `Yaml (Go)` syntax is correct one to use, I think it would still be a good idea to create a dedicated Helm syntax for both LSP and discoverability sake. Even if it just `includes` the other syntax.

jwortmann:
To my understanding, Helm template files are just a subset of all possible files that could use the "YAML (Go)" syntax. That syntax has some other predefined filenames which probably should not start the Helm language server or use the `helm` language ID: https://github.com/sublimehq/Packages/blob/ec775d84c77d5649644b7adf4e9e5a51f973e0be/Go/YAML%20(Go).sublime-syntax#L10-L14

Example Helm syntax:
```yaml
%YAML 1.2
---
name: YAML (Helm Template)
scope: source.yaml.pipeline.helm

contexts:
  main:
    - include: scope:source.yaml.go
      apply_prototype: true
```

And then add another line like https://github.com/SublimeText/YamlPipelines/blob/73d42b95e9924b11382687ea8e437ce28bd34a25/set_syntax_on_load_or_save.py#L10 adjusted with the Helm template path to that plugin.

PR #2654, opened 2025-10-02
Fix another case of outdated diagnostics

rchl:
Somehow I still keep running into outdated diagnostics ;) I've identified one more problem.

When the response for pull diagnostic comes back, in `SessionBuffer._on_document_diagnostic_async` the `version` can be 5 (for example) but by the time the code gets to `SessionBuffer.on_diagnostics_async` the `view.change_count()` can increase. Since we didn't pass the actual diagnostics `version` to `SessionBuffer.on_diagnostics_async`, we assumed that those diagnostics are for the latest version of the view (6 in this example) and then never requested the more recent diagnostics (actual version 6 diagnostics) later. Fix by passing the actual version of the diagnostics around so that we never show outdated diagnostics due to not getting latest diagnostics.

Also reset `_diagnostics_version` on diagnostics request getting canceled. This might not make any difference in practice since if we did cancel the request then it likely means that the view version has changed and we wouldn't ignore next request anyway but in any case, this change feels right.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -628,6 +628,10 @@ class SessionBufferProtocol(Protocol):
         ...
 
     @property
+    def last_synced_version(self) -> int:
+        ...
+
+    @property
     def version(self) -> int | None:
         ...
 
@@ -667,7 +671,7 @@ class SessionBufferProtocol(Protocol):
         ...
 
     def on_diagnostics_async(
-        self, raw_diagnostics: list[Diagnostic], version: int | None, visible_session_views: set[SessionViewProtocol]
+        self, raw_diagnostics: list[Diagnostic], version: int, visible_session_views: set[SessionViewProtocol]
     ) -> None:
         ...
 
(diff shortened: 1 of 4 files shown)


rchl:
Will look into the failing test tomorrow.

plugin/core/sessions.py
```diff
@@ -1394,13 +1398,14 @@ def register_session_buffer_async(self, sb: SessionBufferProtocol) -> None:
         for data in self._registrations.values():
             data.check_applicable(sb)
         uri = sb.get_uri()
-        if uri:
+        version = sb.version
```
rchl: Actually looks like we should be fine removing `SessionBuffer.version` and always relying on `last_synced_version` instead.

plugin/session_buffer.py
```diff
@@ -533,6 +532,7 @@ def _apply_document_diagnostic_async(self, view: sublime.View | None, response:
     def _on_document_diagnostic_error_async(self, version: int, error: ResponseError) -> None:
         self._document_diagnostic_pending_request = None
         if error['code'] == LSPErrorCodes.ServerCancelled:
+            self._diagnostics_version = -1
```
jwortmann: What is the idea behind resetting this value here? The most recently received diagnostics are still visible in the view. This error typically just says "I'm busy, please try again in a second" and we also don't clear the currently visible diagnostics from the views in that case (which would probably result in ugly flickering).

rchl: Right, this makes no sense. My thinking was that we have to revert it since the request was canceled but we don't since it means we didn't set it yet.

jwortmann (approved):

PR #2657, opened 2025-10-07
Add on_selection_modified_async API method

### LINKED ISSUE #2655: Expose debounced on_selection_modified_async to plugins

When the selection / caret position changes, LSP triggers document highlight and code action requests in a debounced manner for the new caret position. If a helper plugin wants to run additional requests or notifications (custom LSP extensions) when the selection changes, it either has to reinvent the wheel with its own debounce implementation or use other hacks to hook onto one of the requests mentioned above.

As a use case, LSP-Tinymist has a browser preview feature to render the active Typst document in a browser. It uses a custom `workspace/executeCommand` request to synchronize the scrolling position in the browser with the caret position in the file. Currently I (ab)use the `on_pre_send_request_async` API to hook onto the document highlight request to get notified when the caret position has changed: https://github.com/sublimelsp/LSP-Tinymist/blob/24e09e4fec31ee8e3323262b2cd98e82bcc544ff/plugin.py#L237-L241
But this would not work if the user has disabled document highlights in their client config.

Add a new plugin API method `on_selection_modified_async` that gets called (debounced) even if document highlight and code actions are disabled.

**Additional context**

Note that we already have `on_session_buffer_changed_async` for content changes in the file, which gets called with `purge_changes_async`: https://github.com/sublimelsp/LSP/blob/96b18e61b6319a5b6732e2582b90e44d4d1f0cb5/plugin/core/sessions.py#L1095-L1099
A new method `on_selection_modified_async` would complement this.

jwortmann:
This PR adds a function to notifiy plugins when the selection in a relevant view has changed, using the same debouncing that is used for `textDocument/documentHighlight` and `textDocument/codeAction` requests.

Closes #2655

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -433,8 +433,7 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         if userprefs().show_code_actions:
             self._do_code_actions_async()
         for sv in self.session_views_async():
-            plugin = sv.session.plugin
-            if plugin:
+            if plugin := sv.session.plugin:
                 plugin.on_selection_modified_async(sv)
 
     def on_post_save_async(self) -> None:


plugin/documents.py
```diff
@@ -422,9 +422,8 @@ def on_selection_modified_async(self) -> None:
         if not self._is_in_higlighted_region(first_region.b):
             self._clear_highlight_regions()
         self._clear_code_actions_annotation()
-        if userprefs().document_highlight_style or userprefs().show_code_actions:
```
jwortmann, line 425: Note that I removed this if-condition here for simplicity (otherwise we would need to check beforehand whether there is at least one session with a plugin). Iirc I added this condition a while ago to reduce unnecessary debouncing calls if neither highlights nor code actions are enabled in the user settings. But I assume that typically at least one of those features is enabled, so probably this condition was relatively useless in almost all cases anyway.

plugin/documents.py
```diff
@@ -433,6 +432,10 @@ def _on_selection_modified_debounced_async(self) -> None:
             self._do_highlights_async()
         if userprefs().show_code_actions:
             self._do_code_actions_async()
+        for sv in self.session_views_async():
+            plugin = sv.session.plugin
+            if plugin:
```
rchl:
```suggestion
            if plugin := sv.session.plugin:
```

rchl (approved):

PR #2660, opened 2025-10-12
Move protocol types to separate folder

jwortmann:
This is an idea that I mentioned in https://github.com/sublimelsp/LSP/pull/2643#discussion_r2362660066.

The goal is to have a clear division between "official" LSP types from the specs and our own type definitions and classes.
After this change, helper plugins could simply use for example
```python
from LSP.protocol import TextDocumentPositionParams
```
instead of the nested folder path
```python
from LSP.plugin.core.protocol import TextDocumentPositionParams
```

I just copied the file from https://github.com/sublimelsp/lsp-python-types/blob/main/lsp_types.py and slightly adjusted the imports. The `plugin/core/protocol.py` file still includes all LSP types via `from ...protocol import *` for backwards compatibility. But I would adjust all the import paths in the plugin files before merging this PR. Then the LSP type imports would always be positioned at the top of the alphabetically sorted import lists.

Let me know what you think and if you are in favor or against such a change.


jwortmann:
Any opinions?

predragnikolic:
This change makes copy pasting the types from lsp-python-types easier 🙂

jwortmann:
I have updated the input paths in the files too.

I'd say maybe wait for @rchl to give his okay too before merging.

rchl (approved):

jwortmann:
I don't know why unit tests are failing on macOS.

It seems to always be stuck at
```
 test_matches_more_specific_action (test_code_actions.CodeActionMatchingTestCase) ... 
Timeout: output is frozen.
```

rchl:
Appears to be failing with all the changes reverted also so we can consider it an unrelated issue.

jwortmann:
I guess it's okay to merge it then. Perhaps the test error on Mac goes away on its own after some time. I remember tests on Mac randomly failing in the past - but usually it works after re-running the test manually in the GitHub UI once or twice. If not, we could temporarily disable the test runner on Mac, or try to figure out later what's wrong.

PR #2664, opened 2025-10-23
Fix dependencies on ST 3

jwortmann:
Fix compatibility with the latest mdpopups release on ST 3 for the remaining ST 3 user(s) that are still out there.

https://forum.sublimetext.com/t/lsp-with-sublime-text-3-2-2/77338/3

Afterwards we still need to create a new tag with prefix `3154-` for this commit (I would say no need to make a release on GitHub for this).


jwortmann:
There is a `release.py` script, but I haven't used that before. Should I instead just push a new tag directly for this commit?

rchl:
Pushing new tag manually should be fine.

jwortmann:
Done: https://github.com/sublimelsp/LSP/releases/tag/3154-0.14.4

Issue #2355, opened 2023-11-04
Code Lenses duplicate on creating "Split View"


rchl:
**Describe the bug**

Code Lenses duplicate on creating "Split View"

**To Reproduce**
1. ![Screenshot 2023-11-04 at 11 48 28](https://github.com/sublimelsp/LSP/assets/153197/13518de7-3c07-49e3-b6ec-632e414950fa)

2. File -> Split View
3. ![Screenshot 2023-11-04 at 11 48 33](https://github.com/sublimelsp/LSP/assets/153197/a7e2a92c-c2fb-458d-af48-12e1d57e247d)

**Environment (please complete the following information):**
- OS: macOS 14.1
- Sublime Text version: 4160
- LSP version: 1.26.0
- Language servers used: LSP-volar

rchl:
Have been doing refactoring to fix that but got distracted by https://github.com/sublimehq/sublime_text/issues/6188 . Causes an issue with code lenses/inlay hints breaking after closing the primary view.

PR #2666, opened 2025-10-24
Fix code lenses with dynamic registration

### LINKED ISSUE #2665: Code lenses don't work with dynamic registration
**Describe the bug**

If the server registers `textDocument/codeLens` dynamically, code lenses are not requested or updated automatically. They are only requested after manually toggling "View > LSP: Show Code Lenses" (or from the command palette).

This happens because in
https://github.com/sublimelsp/LSP/blob/5ff516ca9d869f93f52e84afcaa9652a5ef91742/plugin/documents.py#L822-L827
support for code lenses is only checked using the static capabilities from the initialize response.

This logic should probably be moved into the SessionBuffer instead, to here:
https://github.com/sublimelsp/LSP/blob/5ff516ca9d869f93f52e84afcaa9652a5ef91742/plugin/session_buffer.py#L357-L376

**To Reproduce**

Use a server which uses dynamic registration for code lenses, e.g. https://github.com/aviatesk/JETLS.jl

jwortmann:
Fixes #2665

This moves some of the logic for code lens requests into the SessionBuffer class. Code lenses are still resolved in the SessionView class, because there is some coupling to the views via CodeLensView. Ideally that should be decoupled and resolving should also happen in SessionBuffer, and then only the results would be distributed to the SessionView(s).

Code lenses are now also requested if there is no plugin. But instead, in case there is no plugin, code lenses with commands that are not advertised to be supported by the language server are filtered out (currently only for the regular code lens response - ideally this should also happen after resolve).

### DIFF
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -832,7 +832,7 @@ class SessionBuffer:
 
     def _on_code_lenses_async(self, response: list[CodeLens] | None) -> None:
         if response is None:
-            supported_code_lenses = []
+            supported_code_lenses: list[CodeLens] = []
         elif self.session.uses_plugin():
             supported_code_lenses = response
         else:
@@ -850,10 +850,7 @@ class SessionBuffer:
                 else:
                     self.session.check_log_unsupported_command(command_name)
         for sv in self.session_views:
-            if supported_code_lenses:
-                sv.handle_code_lenses_async(supported_code_lenses)
-            else:
-                sv.clear_code_lenses_async()
+            sv.handle_code_lenses_async(supported_code_lenses)
 
     def set_code_lenses_pending_refresh(self, needs_refresh: bool = True) -> None:
         self.code_lenses_needs_refresh = needs_refresh
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -398,6 +398,7 @@ class SessionView:
             self.session_buffer.do_code_lenses_async(self.view)
             return
         if self._code_lenses.is_empty():
+            self.clear_code_lenses_async()
             return
         promises: list[Promise[None]] = [Promise.resolve(None)]
         if self.get_capability_async('codeLensProvider.resolveProvider'):


plugin/core/sessions.py
```diff
@@ -2097,12 +2107,11 @@ def m_workspace_applyEdit(self, params: Any, request_id: Any) -> None:
...
         self.send_response(Response(request_id, None))
-        if self.uses_plugin():
-            visible_session_views, not_visible_session_views = self.session_views_by_visibility()
-            for sv in visible_session_views:
-                sv.start_code_lenses_async()
-            for sv in not_visible_session_views:
-                sv.set_code_lenses_pending_refresh()
+        visible_session_views, not_visible_session_views = self.session_views_by_visibility()
+        for sv in visible_session_views:
+            sv.session_buffer.do_code_lenses_async(sv.view)
```
rchl, line 2118: If there are multiple views into the same file, will we trigger multiple requests for the same session buffer?

plugin/session_buffer.py
```diff
@@ -792,6 +798,49 @@ def remove_inlay_hint_phantom(self, phantom_uuid: str) -> None:
...
+
+    def _filter_code_lenses(self, code_lenses: list[CodeLens]) -> list[CodeLens]:
+        if self.session.uses_plugin():
+            # Assume that unknown commands are handled by the plugin
+            return code_lenses
+        supported_commands = self.get_capability('executeCommandProvider.commands')
+        if not supported_commands:
+            return []
+        # Filter out CodeLenses with unknown commands
+        return [code_lens for code_lens in code_lenses if self._has_supported_command(code_lens, supported_commands)]
```
rchl: Should we perhaps log something to the console when there are unsupported commands? Even in non-debug mode so that one can notice that something is missing?
(it would be quite spammy though)

plugin/session_view.py
```diff
@@ -270,9 +273,7 @@ def on_capability_added_async(self, registration_id: str, capability_path: str,
             if isinstance(trigger_chars, list) or self.session.config.auto_complete_selector:
                 self._register_auto_complete_triggers(registration_id, trigger_chars or [])
         elif capability_path.startswith("codeLensProvider"):
-            listener = self.listener()
-            if listener:
-                listener.on_code_lens_capability_registered_async()
+            self.session_buffer.do_code_lenses_async(self.view)
```
rchl: Could (should) this code just now live in `SessionBuffer.register_capability_async`?

plugin/core/sessions.py
jwortmann: Yes, I know. This is curretnly the case for all "refresh" requests. A proper way to handle it would be to add a `session_buffers_by_visibility()` and use that instead of the current `session_views_by_visibility()`. (I implemented it in form of the `session_views_by_visibility` first, because the ST API is more suited to iterate through the views in the window / group, but not iterate through buffers. Or alternatively we could collect the relevant SessionBuffers in a set first, and then skip the SessionViews with a SessionBuffer that was already in the set.

But I would say this is not a really critical issue, because the only thing is that there might be some additional useless requests that do nothing. And these "refresh" requests from the server should happen pretty rarely, e.g. when a configuration file is modified (see comment in https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeLens_refresh):

```typescript
export interface [CodeLensWorkspaceClientCapabilities](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeLensWorkspaceClientCapabilities) {
	/**
	 * Whether the client implementation supports a refresh request sent from the
	 * server to the client.
	 *
	 * Note that this event is global and will force the client to refresh all
	 * code lenses currently shown. It should be used with absolute care and is
	 * useful for situation where a server for example detect a project wide
	 * change that requires such a calculation.
	 */
	refreshSupport?: boolean;
}
```

Let's keep it on the backlog though.

plugin/session_buffer.py
jwortmann: I would probably only log to the console if debug logging is enabled. But I would not log every time when this request is made (so, after every change to the buffer). Perhaps we can collect the unknown commands and store it in a variable per Session, and then only log if there is a new unknown command.

plugin/session_view.py
jwortmann: Yes, thanks: https://github.com/sublimelsp/LSP/pull/2666/commits/cbf9d292d06b18d35c24e5f9b00045ce0598370d

plugin/session_buffer.py
jwortmann: I added debug logging for unsupported commands in https://github.com/sublimelsp/LSP/pull/2666/commits/5f202869b43b567b7fb055ee7a9408b2c90d103f.
Each unsupported command is logged only once (per session) and only if there is no plugin.

plugin/session_buffer.py
```diff
@@ -160,6 +166,10 @@ def session_views(self) -> WeakSet[SessionViewProtocol]:
     def last_synced_version(self) -> int:
         return self._last_synced_version
 
+    @property
+    def supported_commands(self) -> set[str]:
+        return self._supported_commands
```
rchl: I would think that the purpose of creating getters like this one is to make it explicit that this is a public property. In this case this is only used within this class so this seems redundant.

plugin/session_buffer.py
```diff
@@ -792,6 +818,47 @@ def remove_inlay_hint_phantom(self, phantom_uuid: str) -> None:
...
+        if not LspToggleCodeLensesCommand.are_enabled(view.window()):
+            return
+        params = {'textDocument': text_document_identifier(view)}
+        for sv in self.session_views:
+            if sv.view == view:
+                for request_id, data in sv.active_requests.items():
+                    if data.request.method == 'codeAction/resolve':
+                        self.session.cancel_request(request_id)
+                break
+        self.session.send_request_async(Request('textDocument/codeLens', params, view), self._on_code_lenses_async)
```
rchl: `params` could be moved down

```suggestion
       for sv in self.session_views:
            if sv.view == view:
                for request_id, data in sv.active_requests.items():
                    if data.request.method == 'codeAction/resolve':
                        self.session.cancel_request(request_id)
                break
        params = {'textDocument': text_document_identifier(view)}
         self.session.send_request_async(Request('textDocument/codeLens', params, view), self._on_code_lenses_async)
```

plugin/session_buffer.py
jwortmann: I think it could be useful later for code actions / executeCommand too, which are run from the Session class. In that case it's nice to ensure that it is read-only.
Do you prefer that I remove it?

plugin/session_buffer.py
jwortmann: https://github.com/sublimelsp/LSP/pull/2666/commits/035e6209a9e7536884a64fe45c5f2bc2a5de55b3

plugin/session_buffer.py
rchl: If you are at it already then lets remove. Can always add later if needed.

plugin/session_buffer.py
```diff
@@ -792,6 +818,47 @@ def remove_inlay_hint_phantom(self, phantom_uuid: str) -> None:
...
+                for request_id, data in sv.active_requests.items():
+                    if data.request.method == 'codeAction/resolve':
+                        self.session.cancel_request(request_id)
+                break
+        params = {'textDocument': text_document_identifier(view)}
+        self.session.send_request_async(Request('textDocument/codeLens', params, view), self._on_code_lenses_async)
+
+    def _on_code_lenses_async(self, response: list[CodeLens] | None) -> None:
+        if not response:
+            return
```
jwortmann: Actually I think this is wrong. I need to test what happens if the server signalizes that all code lenses should be removed. But I think this is also a bug in the current implementation in SessionView where we have `or not isinstance(response, list)`. If the response is `None`, I think we still have to remove all code lenses that are currently shown.

jwortmann: I think it should be correct now.

jwortmann: Actually @rchl could you take a look. If the response is empty, should I better call `handle_code_lenses_async([])` or `clear_code_lenses_async`? I think both paths still call `self._phantom.update` in CodeLensView. That seem a little inefficient to me, because they only need to be updated if changed, i.e. if there already were no code lenses in the view, we don't have to call `self.view.erase_regions` after each buffer change. I think there is room for improvement.

plugin/session_buffer.py
```diff
@@ -854,7 +854,10 @@ def _on_code_lenses_async(self, response: list[CodeLens] | None) -> None:
                 else:
                     self.session.check_log_unsupported_command(command_name)
         for sv in self.session_views:
-            sv.handle_code_lenses_async(supported_code_lenses)
+            if supported_code_lenses:
+                sv.handle_code_lenses_async(supported_code_lenses)
+            else:
+                sv.clear_code_lenses_async()
```
rchl: Or this `clear_code_lenses_async()` call could instead be done from `handle_code_lenses_async()`. I feel that it would make the logic more centralized and easier to follow.

plugin/session_buffer.py
jwortmann: Okay. https://github.com/sublimelsp/LSP/pull/2666/commits/7b5985b52d6f37c6f373c72e4e620e3b7f5b4a43

plugin/session_buffer.py
rchl: `CodeLensView.handle_response` doesn't seem to update view, it only changes structures. `resolve_visible_code_lenses_async()` also bails on empty code lenses so won't update view either.

So as I suggested in next issue, I would `handle_code_lenses_async([])` and then call `clear_code_lenses_async()` from there.

jwortmann: Okay I did that, but I think the logic in CodeLensView could still be improved. Ideally we don't want to call the ST APIs to update phantoms and erase regions if there were none in the view before. It might be a little tricky with that two-step strategy that separates the method calls into structure update and then render.

jwortmann: I think I leave it like that for today though. The current code should work correctly and we can do more improvements for efficiency at a later time. That would also include to move the resolving logic from SessionView into SessionBuffer as well, and then just distribute the filtered results to the CodeLensView instance(s).

jwortmann:
I noticed there is still a bug where code lenses don't get cleared correctly and old code lenses are shown multiple times. I will try to fix it in the next days.

jwortmann:
I think I found 2 more bugs that were also present before this PR:

First, on a new code lens request, the open resolve requests are intended to be cancelled. But actually it cancels the code *action* resolve, and not code *lens* resolve:
https://github.com/sublimelsp/LSP/blob/5d9f24f21b35c6959f68e1bb3f402b68c5f040eb/plugin/session_view.py#L391

And second, it seems that code lenses are only ever resolved for the visible region.
https://github.com/sublimelsp/LSP/blob/5d9f24f21b35c6959f68e1bb3f402b68c5f040eb/plugin/session_view.py#L414

Since the ST API doesn't have any callback when the view is scrolled, there could still be unresolved code lenses at some other part of the file after scrolling. And as far as I can see, the `resolve_visible_code_lenses_async` method is never called aside from directly after the regular code lens request, which only triggers after buffer modifications.

I think I will do another bigger rewrite where all unresolved code lenses are resolved immediately.
For testing, does anyone knows which servers support `codeLens/resolve` and send unresolved code lenses first?

I will also try to fix #2355 if possible.

rchl:
> For testing, does anyone knows which servers support `codeLens/resolve` and send unresolved code lenses first?

LSP-volar on https://github.com/rchl/lsp-log-parser/blob/master/components/message-payload.vue file for example.

> Since the ST API doesn't have any callback when the view is scrolled, there could still be unresolved code lenses at some other part of the file after scrolling. And as far as I can see, the `resolve_visible_code_lenses_async` method is never called aside from directly after the regular code lens request, which only triggers after buffer modifications.

At least with the older LSP version it was enough to scroll the view and change the selection but with the latest it seems like indeed some modification has to be done.

jwortmann:
> At least with the older LSP version it was enough to scroll the view and change the selection but with the latest it seems like indeed some modification has to be done.

Oh right, I see I removed the line at https://github.com/sublimelsp/LSP/blob/5d9f24f21b35c6959f68e1bb3f402b68c5f040eb/plugin/documents.py#L428 in `on_selection_modified_async` in this PR. I guess we could simply add back that line to fix it, however it still needs a selection change to trigger. So I think as long as there is no ST API for scroll events, it would probably be better to immediately resolve all code lenses in the entire file. That would also work better with filtering for supported commands. Or alternatively, if it turns out to affect performance, maybe we only resolve immediately if the total number of unresolved code lenses is below some threshold value, and otherwise fall back to resolve only visible first and after that again on selection changed.

PR #2670, opened 2025-10-29
Update CI dependencies

jwortmann:
Supersedes #2656


jwortmann:
By the way, should we disable unit tests on Mac for now? I have no idea what the reason for that failing is, or how to fix it.

predragnikolic:
I will try to experiment when I get a chance.
For now, I have noticed that the failing GitHub Actions use MacOS 15.7.1, but the last successful GitHub Action used MacOS 15.6.1.

PR #2674, opened 2025-11-02
Request code actions separately during save

### LINKED ISSUE #2510: Multiple code action kinds requests during code actions on save
**Describe the bug**

I've mentioned that (vscode interoperability) issue in some other place but I'm too lazy to find those now.

When asking servers for code actions during "code actions on save" task, we should create separate requests per code action kind and not include multiple code action kinds in a single request. Otherwise we might try to apply multiple code actions that are not compatible with each other (if server doesn't provide `version` for text edits).

rchl:
During "code actions on save" handling, request each code action kind separately from each server and apply returned code actions immediately before requesting next kind.

Before this change we would request all code actions from all servers and try to apply them all one after each other which would most likely fail since earlier document changes would make later fail to apply or apply incorrectly.

Fixes #2510

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -27,7 +27,7 @@ from typing import cast
 import sublime
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, Generator
+    from collections.abc import Callable, Generator, Iterator
     from typing import TypeGuard
 
 
@@ -247,7 +247,7 @@ class CodeActionOnSaveTask(SaveTask):
         request_iterator = actions_manager.request_on_save_async(view, on_save_actions)
         self._process_next_request(request_iterator)
 
-    def _process_next_request(self, request_iterator: Generator[Promise[CodeActionsByConfigName]]) -> None:
+    def _process_next_request(self, request_iterator: Iterator[Promise[CodeActionsByConfigName]]) -> None:
         if self._cancelled:
             return
         request = next(request_iterator, None)
@@ -257,7 +257,7 @@ class CodeActionOnSaveTask(SaveTask):
             self._on_complete()
 
     def _handle_response_async(
-        self, response: CodeActionsByConfigName, request_iterator: Generator[Promise[CodeActionsByConfigName]]
+        self, response: CodeActionsByConfigName, request_iterator: Iterator[Promise[CodeActionsByConfigName]]
     ) -> None:
         if self._cancelled:
             return


plugin/code_actions.py
```diff
@@ -222,42 +226,50 @@ class CodeActionOnSaveTask(SaveTask):
...
-                    session.run_code_action_async(action, progress=False, view=view) for action in code_actions
-                ])
-        Promise.all(tasks).then(lambda _: self._on_complete())
+        config_name, code_actions = response
+        session = self._task_runner.session_by_name(config_name, 'codeActionProvider')
+        if session and code_actions:
+            tasks.extend([
+                session.run_code_action_async(action, progress=False, view=view) for action in code_actions
+            ])
+        Promise.all(tasks).then(lambda _: self._process_next_request(request_iterator))
```
rchl, line 272: This still doesn't address the part about [running multiple code actions of the same kind at once](https://github.com/sublimelsp/LSP/issues/2510#issuecomment-2495604063) since I've only just noticed it when creating the PR. At least in case of zig, this [doesn't seem to be an issue anymore](https://github.com/zigtools/zls/issues/2076#issuecomment-3478389502) due to server change (~~pending confirmation~~).

I've looked a bit at how VSCode handles it and while I don't pretend to have comprehensive understanding of its logic, I don't quite see it re-requesting any code actions. It seems like it just applies everything it gets but I might be missing some detail. I've looked at [this code](https://github.com/microsoft/vscode/blob/f66e839a38dfe39ee66a86619a790f9c2336d698/src/vs/workbench/contrib/codeEditor/browser/saveParticipants.ts#L410-L431).

I don't think re-requesting until there are no code actions is a proper solution since for example typescript-language-server always provides an edit for `source.organizeImports`, regardless if it differs from what is already there. And in case where we would get 10 code actions of `source.fixAll` type, it would take too much time to process everything (the save task would time out). We could potentially just run the first code action which would be somewhat of a compromise.

rchl: Also checked with VSCode and the old version of `zls` that returned multiple code actions and noticed that it also had the same issue - it applied all returned code actions and that resulted in incorrect final result:


https://github.com/user-attachments/assets/30fcaa10-239a-48dd-a7f1-26978e3ca5de

So I think we are fine as is and it should be the responsibility of the server to not provide multiple code actions or provide them in order that won't cause issues.

plugin/code_actions.py
```diff
@@ -22,14 +22,19 @@
...
 from functools import partial
-from typing import Any, Callable, Dict, List, Tuple, Union
+from typing import TYPE_CHECKING, Any, List, Tuple, Union
 from typing import cast
-from typing_extensions import TypeGuard
 import sublime
 
+if TYPE_CHECKING:
+    from collections.abc import Callable, Generator
+    from typing import TypeGuard
```
rchl, line 31: Use of `TYPE_CHECKING` and some other minor changes due to me running ruff linter casually and it reporting various "best-practice" warnings. I don't think we are ready to commit to it fully since we would first have to fix all existing "issues" but I tend to follow it when changing code.

plugin/code_actions.py
```diff
@@ -222,42 +226,50 @@ class CodeActionOnSaveTask(SaveTask):
...
+        view = self._task_runner.view
+        on_save_actions = self._get_code_actions_on_save(view)
+        request_iterator = actions_manager.request_on_save_async(view, on_save_actions)
+        self._process_next_request(request_iterator)
 
-    def _request_code_actions_async(self) -> None:
-        self._purge_changes_async()
-        on_save_actions = self._get_code_actions_on_save(self._task_runner.view)
-        actions_manager.request_on_save_async(self._task_runner.view, on_save_actions).then(self._handle_response_async)
+    def _process_next_request(self, request_iterator: Generator[Promise[CodeActionsByConfigName]]) -> None:
```
jwortmann: Could be `Iterator` instead of `Generator` here.

plugin/code_actions.py
```diff
@@ -222,42 +226,50 @@ class CodeActionOnSaveTask(SaveTask):
...
+            return
+        request = next(request_iterator, None)
+        if request:
+            request.then(lambda response: self._handle_response_async(response, request_iterator))
+        else:
+            self._on_complete()
 
-    def _handle_response_async(self, responses: list[CodeActionsByConfigName]) -> None:
+    def _handle_response_async(
+        self, response: CodeActionsByConfigName, request_iterator: Generator[Promise[CodeActionsByConfigName]]
```
jwortmann: `Iterator` here too.

PR #2675, opened 2025-11-04
Reorganize PostgreSQL section in language_servers.md

predragnikolic:
Moving the PostgreSQL docs to a new SQL section.

### DIFF
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -847,7 +847,7 @@ Follow installation instructions on [LSP-metals](https://github.com/scalameta/me
 
 ## SQL
 
-### PostgresSQL
+### PostgreSQL
 
 1. Install the [Postgres Language Server](https://pg-language-server.com/latest/manual_installation/).
 2. Open `Preferences > Package Settings > LSP > Settings` and add the `"postgres-language-server"` client configuration to the `"clients"`:


docs/src/language_servers.md
```diff
@@ -861,6 +845,24 @@ Follow installation instructions on [LSP-metals](https://github.com/scalameta/me
     }
     ```
 
+## SQL
+
+### PostgresSQL
+
+1. Install the [Postgres Language Server](https://pg-language-server.com/latest/manual_installation/).
+2. Open any .sql file and select `View > Syntax > Open all with current extension as... > SQL > PostgreSQL` to set the PostgreSQL syntax as the default.
```
jwortmann: Not sure if this is necessary. It seems like PostgreSQL files can have special file extensions too: https://github.com/sublimehq/Packages/blob/9de8e396796f5013ac5dd51a189d4c4bb6537a5d/SQL/PostgreSQL.sublime-syntax#L9-L11

Perhaps just an advice like "Ensure that the PostgreSQL syntax is applied to the relevant files, for example by selecting `View > Syntax > SQL > PostgreSQL`." would be better.

jwortmann (approved):

docs/src/language_servers.md
```diff
@@ -916,7 +919,7 @@ Follow installation instructions on [LSP-terraform](https://github.com/sublimels
         "clients": {
             "jag": {
                 "enabled": true,
-                "command": ["jag" "lsp"],
+                "command": ["jag", "lsp"],
```
predragnikolic, line 922: this is an unrelated minor change

docs/src/language_servers.md
```diff
@@ -861,6 +845,25 @@ Follow installation instructions on [LSP-metals](https://github.com/scalameta/me
     }
     ```
 
+## SQL
+
+### PostgresSQL
```
predragnikolic:
```suggestion
### PostgreSQL
```

PR #2677, opened 2025-11-05
Refactor code lenses

### LINKED ISSUE #2355: Code Lenses duplicate on creating "Split View"

### LINKED ISSUE #2497: `...` annotations popping up
**Describe the bug**

With code lens disabled sometimes I observe `...` popping up in the right side of all views after I edit code.

**To Reproduce**
Prepare a file that defines a symbol (a function) that's used in some other files within the project. (I.e. a file with a function that would have "1 reference" showed if code lens were enabled) 

1. Put
   ```
     "disabled_capabilities": {
       "codeLensProvider": true
     },
   ```
   into your LSP configuration
2. Open a file with LSP enabled, wait for LSP to start.
3. Edit the name of the function
4. Undo the edit
5. Observe `...` in the right side of the view

**Expected behavior**

No `...` annotation on the screen

**Screenshots**

https://github.com/sublimelsp/LSP/assets/11963737/eafda1b5-132d-4eb8-9d02-2bbc40e992d5

**Logs**

[LSP_log.txt](https://github.com/user-attachments/files/15936090/LSP_log.txt)
[LSP_log_debug.txt](https://github.com/user-attachments/files/15936096/LSP_log_debug.txt)

<details>
<summary>
[Troubleshooting output]
</summary>

# Troubleshooting: ClojureHelper
## Version
 - LSP: 2.1.0
 - Sublime Text: 4169
## Server Test Run
 - exit code: 0
 - output
```

```
## Server Configuration
 - command
```json
[
  "clojure-lsp"
]
```
 - shell command
```sh
clojure-lsp
```
 - selector
```
source.clojure
```
 - priority_selector
```
source.clojure
```
 - init_options
```json
{}
```
 - settings
```json
{}
```
 - env
```json
{}
```

## Active view
 - File name
```
/Users/kgofhedgehogs/Work/ps/src/ps.cljc
```
 - Settings
```json
{
  "auto_complete_selector": "meta.tag, source - comment - string.quoted.double.block - string.quoted.single.block - string.unquoted.heredoc",
  "lsp_active": true,
  "syntax": "Packages/Clojure/ClojureScript.sublime-syntax"
}
```
 - base scope
```
source.clojure.clojurescript
```

## Project / Workspace
 - folders
```json
[
  "/Users/kgofhedgehogs/Work/ps"
]
```
 - is project: False

## LSP configuration

```json
{
  "log_debug": true,
  "show_code_actions": "bulb",
  "show_view_status": false
}
```
## System PATH
 - /usr/local/bin
 - /usr/bin
 - /bin
 - /usr/sbin
 - /sbin
 - /Library/Apple/usr/bin
 - /opt/homebrew/opt/arm-gcc-bin@8/bin
 - /opt/homebrew/opt/avr-gcc@8/bin
 - /opt/homebrew/opt/qt@5/bin
 - /opt/homebrew/sbin
 - /Users/kgofhedgehogs/.jenv/shims
 - /Users/kgofhedgehogs/bin
 - /opt/homebrew/bin
 - /Users/kgofhedgehogs/.orbstack/bin

</details>

- OS: macOS 12.7.4
- Sublime Text version: 4169
- LSP version: 2.1.0
- Language servers used: Clojure

**Additional context**
"ClojureHelper" source located here:
https://github.com/sublimelsp/LSP/issues/1844#issuecomment-961214237

jwortmann:
Continuing from #2666

I refactored and renamed CodeLensView and moved the handling from SessionView to SessinBuffer, so that it is not done in parallel with the same data in case there are cloned (split) views. I have kept the caching for the annotations/commands and re-enabled resolving for the visible viewport area after selection change, which was accidentally removed in #2666. Unfortunately if the server uses `codeLens/resolve` the code lenses don't work great together with the filtering of supported commands. But it should be fine at least for the current viewport area due to the implemented caching. If you scroll away, you might need to click into the viewport (to change the selection and trigger resolve) before code lenses show up. But I also made some improvements for the caching:

The previous implementation first converted the line and column based LSP `Range` to a `sublime.Region`
https://github.com/sublimelsp/LSP/blob/72e65164606c03ee08a88ef69db17ee124b5c46d/plugin/code_lens.py#L59
and then used that `Region` as the key to store code lenses grouped by region:
https://github.com/sublimelsp/LSP/blob/72e65164606c03ee08a88ef69db17ee124b5c46d/plugin/code_lens.py#L140
Then when code lenses are updated after a buffer change, it used the new regions to check if it can re-use the previous annotations temporarily until the code lenses get updated with the real data from `codeLens/resolve`. But this would only work for the code lenses between the start of the file and the cursor postion (where user is currently typing), because the offset-based regions for the code lenses that come after the cursor position would change after typing or deleting some characters in the file. So in this PR I used the line and column based LSP ranges as keys for the code lens groups instead. If you type some characters on a line, the LSP ranges for the code lenses after that line are still the same, and therefore caching should work for all code lenses in the entire file, except for the exact line where the user is typing and as long as the user didn't type or deleted a linebreak.

This rewrite also significantly reduces the number of ST API calls for `add_regions` and for updating phantoms. Now only a single region key per session is used for the code lens annotations, instead of one key per code lens and session. I wonder whether there was a reason that it was implemented as one region key per code lens before?
https://github.com/sublimelsp/LSP/blob/72e65164606c03ee08a88ef69db17ee124b5c46d/plugin/code_lens.py#L203-L205

Furthermore, if there were no code lenses in the view, and no code lenses need to be added after a buffer change, there is no unnecessary API call anymore to clear annotations and phantoms.

More fixes and changes:
- fixes to cancel `codeLens/resolve` requests and not `codeAction/resolve`, which is a bug in the current implementation: https://github.com/sublimelsp/LSP/blob/72e65164606c03ee08a88ef69db17ee124b5c46d/plugin/session_buffer.py#L831
- fixes code lenses not immediately being redrawn on user setting change (switch between phantom/annotation)
- fixes #2355
- the LspCodeLensCommand now uses the quick panel instead of a popup menu at the caret position to show options when there are multiple code lenses

Fixes #2497
Fixes #2355

### DIFF
--- a/plugin/code_lens.py
+++ b/plugin/code_lens.py
@@ -10,7 +10,6 @@ from .core.registry import LspWindowCommand
 from .core.registry import windows
 from .core.views import range_to_region
 from functools import partial
-from typing import Generator
 from typing import cast
 from typing_extensions import TypeGuard
 import itertools


plugin/session_buffer.py
```diff
@@ -821,34 +825,47 @@ def remove_all_inlay_hints(self) -> None:
...
+
+    def resolve_visible_code_lenses_async(self, view: sublime.View) -> None:
+        promises: list[Promise] = []
+        if self.has_capability('codeLensProvider.resolveProvider'):
+            for code_lens in self._code_lenses.unresolved_visible_code_lenses_async(view):
+                request = Request('codeLens/resolve', code_lens.data, view)
+                promise = self.session.send_request_task(request).then(code_lens.resolve)
+                promises.append(promise)
+        Promise.all(promises).then(lambda _: self._on_visible_code_lenses_resolved_async)  # FIXME doesn't work
+        sublime.set_timeout_async(self._on_visible_code_lenses_resolved_async, 100)  # FIXME workaround
```
jwortmann: I don't know why `Promise.all` doesn't work here. I see that the previous (current) implementation already puts one `Promise.resolve(None)` into the list (why?), but even when I try that it does not work here.

https://github.com/sublimelsp/LSP/blob/72e65164606c03ee08a88ef69db17ee124b5c46d/plugin/session_view.py#L403

rchl:
One thing I've noticed immediately is that the code lenses don't seem to appear initially until selection changes. For example when using "restart server" or restarting ST with relevant file open.

EDIT: Looks like it's related to https://github.com/sublimelsp/LSP/pull/2677/files#r2496161303 and works after getting rid of the set_timeout workaround.

plugin/session_buffer.py
rchl: I haven't tried but isn't it just a case of missing parenthesis?

```suggestion
        Promise.all(promises).then(lambda _: self._on_visible_code_lenses_resolved_async())
```

plugin/session_view.py
```diff
@@ -379,44 +380,85 @@ def on_post_save_async(self, new_uri: DocumentUri) -> None:
...
+        self.view.erase_regions(self._code_lens_key())
 
-    def resolve_visible_code_lenses_async(self) -> None:
-        if not LspToggleCodeLensesCommand.are_enabled(self.view.window()):
-            return
-        if not self._code_lenses.is_initialized():
-            self.session_buffer.do_code_lenses_async(self.view)
-            return
-        if self._code_lenses.is_empty():
+    def _redraw_code_lenses_async(self, clear: bool = False) -> None:
```
rchl: Good to force `clear` to be a keyword argument.
(Also ruff mentions this as a warning)

```suggestion
    def _redraw_code_lenses_async(self, *, clear: bool = False) -> None:
```

rchl:
When split view is opened it does not show code lenses until selection changes. Not sure if it was like that before also...

plugin/session_buffer.py
jwortmann: Ouch... I was trying all kind of other things to figure out why it wasn't working. Sometimes it's good to have two more eyes looking at the code to find the bug, thanks :)

jwortmann:
> When split view is opened it does not show code lenses until selection changes. Not sure if it was like that before also...

I will look into that tomorrow or over the weekend and do some more testing for edge cases.

*Edit:* fixed in https://github.com/sublimelsp/LSP/pull/2677/commits/e473dbfca0a7d42b3addc2b432abb97beb55edad

plugin/code_lens.py
```diff
@@ -1,259 +1,180 @@
...
-        self._init = False
-        self._phantom = sublime.PhantomSet(view, self.CODE_LENS_KEY)
-        self._code_lenses: dict[tuple[int, int], list[CodeLensData]] = {}
-
-    def clear(self) -> None:
-        self._code_lenses.clear()
+        assert is_resolved(response)
+        self.data = cast(CodeLens, response)
+        self.range_ = HashableRange(response['range'])
+        self.cached_command = cast(Command, response['command'])
```
rchl: 1. I would call it `on_resolve` or `handle_resolve` since this doesn't do resolving itself.
2. We can type the `self.data` as `CodeLens | ResolvedCodeLens` in init and remove both casts.
3. The argument in `def is_resolved(code_lens: CodeLens)` should also accept both since we don't in theory know if the input code_lens is resolved or not.

plugin/code_lens.py
```diff
@@ -1,259 +1,180 @@
...
-        self.view = view
-        self._init = False
-        self._phantom = sublime.PhantomSet(view, self.CODE_LENS_KEY)
-        self._code_lenses: dict[tuple[int, int], list[CodeLensData]] = {}
-
-    def clear(self) -> None:
-        self._code_lenses.clear()
+        assert is_resolved(response)
+        self.data = cast(CodeLens, response)
+        self.range_ = HashableRange(response['range'])
```
rchl: I find stray underscore in the name confusing. Could we call it hashable_range or h_range or something?

plugin/code_lens.py
```diff
@@ -1,259 +1,180 @@
...
+                for old, new in zip(old_group, group):
+                    if is_resolved(new.data):
+                        continue
+                    # This assignment is only temporary, the resolve call in the future will fill in the actual command
+                    # after resolve() is called on the CachedCodeLens. However reusing the previous command title if the
+                    # code lens groups are the same makes the screen not jump from the phantoms moving around too much.
+                    new.cached_command = old.data['command'] if is_resolved(old.data) else old.cached_command
+        self.code_lenses = grouped_code_lenses
+
+    def unresolved_visible_code_lenses_async(self, view: sublime.View) -> Generator[CachedCodeLens]:
```
rchl: The result is used immediately without doing anything between each yield so I would just return a plain list.

Optimizing for memory usage is IMO not important here.

plugin/session_buffer.py
```diff
@@ -828,33 +834,42 @@ def do_code_lenses_async(self, view: sublime.View) -> None:
...
+            for code_lens in code_lenses:
+                command_name = code_lens['command']['command']
                 if command_name in self._supported_commands:
-                    supported_code_lenses.append(code_lens)
+                    self._filtered_code_lenses.append(code_lens)
                 else:
                     self.session.check_log_unsupported_command(command_name)
         for sv in self.session_views:
-            sv.handle_code_lenses_async(supported_code_lenses)
+            sv.handle_code_lenses_async(self._filtered_code_lenses)
```
rchl: Can we avoid having `self._filtered_code_lenses` in SessionBuffer? It seems like we only need it when adding new SessionView but we could as well just re-filter at that point (run most of this code again).

plugin/code_lens.py
jwortmann: I renamed to just `range` since when used as a property name it cannot clash with the built-in function. The reason I opted for the underscore in the recent commit before was to be consistent with the local variable name which is used in the method below, but I guess that's not really important. (I kept the trailing underscore for that local variable name though and afaik that style is the standard convention in Python when using names that are also keywords or to avoid shadowing built-ins).

rchl:
We have to release new LSP version ASAP since there are things relying on it. Hope everyone is fine with that. :)

PR #2680, opened 2025-11-09
Purge changes before triggering formatting

### LINKED ISSUE #2578: `lsp_format_document` breaks formatting when executed right after editing code
https://github.com/user-attachments/assets/9b12a0ed-bdb2-4f17-91ec-9816d1774257

Seems like server is not fast enough to parse document after change and applies formatting to old version?

**Environment:**
- OS: Arch Linux 6.12.1
- Sublime Text version: 4180
- LSP version: 2.2.0
- Language servers used: vstls

rchl:
Fixes #2578

### DIFF
--- a/plugin/formatting.py
+++ b/plugin/formatting.py
@@ -191,6 +191,7 @@ class LspFormatDocumentRangeCommand(LspTextCommand):
         if self.view.has_non_empty_selection_region() and \
                 bool(self.best_session('documentRangeFormattingProvider.rangesSupport')):
             return True
+        return False
 
     def run(self, edit: sublime.Edit, event: dict | None = None) -> None:
         if listener := self.get_listener():


plugin/formatting.py
```diff
@@ -181,14 +186,15 @@ class LspFormatDocumentRangeCommand(LspTextCommand):
...
-            return True
-        if self.view.has_non_empty_selection_region() and \
-                bool(self.best_session('documentRangeFormattingProvider.rangesSupport')):
-            return True
-        return False
+        return bool(
+            has_single_nonempty_selection(self.view)
+            or
+            (self.view.has_non_empty_selection_region() and self.best_session('documentRangeFormattingProvider.rangesSupport'))  # noqa: E501
+        )
```
jwortmann: To be honest I'm not a big fan of these stylistic changes to return `bool(...)` with these long expressions/function names. I think the `if ... return` is easier to read here.
Also the `bool` is technically only needed around the `self.best_session` because the two other functions already return a bool. (Alternatively it looks like the bool could be removed alltogether in the `if` condition in the original code.)
And the parenthesis after `or` are not necessary because `and` has higher precedence in Python (and in every other programming language that I know).

rchl: Those were prompted by ruff warnings like:

Screenshot 2025-11-10 at 11 01 14


which I generally agree with but I can see how addressing those can end up looking a bit weird.

The parenthesis is also something it prefers in some cases:

Screenshot 2025-11-10 at 11 03 02

PR #2681, opened 2025-11-10
Use more assignment expressions

jwortmann:
... in simple if-conditions (but not when negated or in more complex expressions that would require additional parentheses). I also made a few other code simplifications and minor style adjustments.

There are no functional changes in this pull request.

### DIFF
--- a/plugin/core/input_handlers.py
+++ b/plugin/core/input_handlers.py
@@ -137,7 +137,7 @@ class DynamicListInputHandler(sublime_plugin.ListInputHandler, metaclass=ABCMeta
     def list_items(self) -> list[sublime.ListInputItem]:
         if not self.text:  # Show initial items when the command was just invoked
             return self.get_list_items() or [sublime.ListInputItem("No Results", "")]
-        elif items := getattr(self.command, '_items', None):  # Items were updated after typing
+        if items := getattr(self.command, '_items', None):  # Items were updated after typing
             if ST_VERSION >= 4157:
                 return items
             else:


predragnikolic:
Did you maybe used a tool (like ast-grep - https://ast-grep.github.io/catalog/python/use-walrus-operator-in-if.html ) to make these changes, or did you maybe did them manually? (just curious)

jwortmann:
> Did you maybe used a tool (like ast-grep - https://ast-grep.github.io/catalog/python/use-walrus-operator-in-if.html ) to make these changes, or did you maybe did them manually? (just curious)

Manually. So I may have missed one or two, but we can always make more style improvements later if we want.

plugin/code_actions.py
```diff
@@ -423,10 +414,9 @@ def _is_cache_valid(self, event: dict | None) -> bool:
         view = self.view
         if not view:
             return False
-        region = self._get_region(event)
-        if region is None:
-            return False
-        return actions_manager.menu_actions_cache_key == f"{view.buffer_id()}#{view.change_count()}:{region}"
+        if region := self._get_region(event):
+            return actions_manager.menu_actions_cache_key == f"{view.buffer_id()}#{view.change_count()}:{region}"
+        return False
```
rchl: I have not checked whether this would affect this particular case but we should theoretically be checking region against `None` since empty region also evaluates to `False`.

Same below.

plugin/core/input_handlers.py
```diff
@@ -137,16 +137,14 @@ def _detach_listener(self) -> None:
...
-            items = getattr(self.command, '_items', None)
-            if items:
-                if ST_VERSION >= 4157:
-                    return items
-                else:
-                    # Trick to select the topmost item; see https://github.com/sublimehq/sublime_text/issues/6162
-                    sublime.set_timeout(self._select_first_row)
-                    return [sublime.ListInputItem("", "")] + items
-            return [sublime.ListInputItem(f'No Symbol found: "{self.text}"', "")]
+        elif items := getattr(self.command, '_items', None):  # Items were updated after typing
```
rchl: Linters report those as "Unnecessary `elif` after `return` statement" and I tend to agree with that.

```suggestion
        if items := getattr(self.command, '_items', None):  # Items were updated after typing
```

plugin/document_link.py
```diff
@@ -16,50 +16,32 @@ class LspOpenLinkCommand(LspTextCommand):
...
-            self.open_target(target)
-        else:
-            if not session.has_capability("documentLinkProvider.resolveProvider"):
-                debug("DocumentLink.target is missing, but the server doesn't support documentLink/resolve")
-                return
-            session.send_request_async(Request.resolveDocumentLink(link, self.view), self._on_resolved_async)
+        if position := get_position(self.view, event):
+            if session := self.best_session(self.capability, position):
+                if sv := session.session_view_for_view_async(self.view):
+                    if link := sv.session_buffer.get_document_link_at_point(self.view, position):
```
rchl, line 29: Ruff suggests to merge nested if statements. I get that you are doing it for readability (avoid weird wrapping and parenthesis) so I'm not saying we should follow ruff but in the long term perhaps we should align (just asking)?

I guess the alternative would be to do:


```suggestion
        if (position := get_position(self.view, event)) \
            and (session := self.best_session(self.capability, position)) \
            and (sv := session.session_view_for_view_async(self.view)) \
            and (link := sv.session_buffer.get_document_link_at_point(self.view, position)):
```

It looks more busy for sure.
The benefit is that there is less indentation though.

plugin/code_actions.py
jwortmann: You're right. I thought empty region would still evaluate to `True`, but in fact it does not. I'll fix it.

plugin/document_link.py
jwortmann: If we do it like that, I think the subsequent lines in the if-condition should have an additional level of indentation. Otherwise the following code would be at the same indentation level as the `and`.
But not sure if I prefer it that way or with the multiple if conditions. Perhaps we should configure ruff to ignore certain lint rules?

rchl: To me multiple if conditions feel "less professional" but it might be just me.
Additional indentation would be fine by me.

PR #2686, opened 2025-11-14
Don't close signature help popup based on selector

rchl:
It was reported on Discord that the signature help closes itself when typing a single character after signature help is made visible (in Go/Ruby).

I've figured out that it's due to the code that I've removed here since it checks for specific scopes that are not used in those Go/Ruby syntaxes.

I'm not sure if we really need this code since when testing in TypeScript/Python, the signature help popup closes properly on typing over the closing parenthesis. Also it closes anyway on moving the selection. (it also closes on navigating completions with arrows which I feel is wrong/bug).

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -597,7 +597,7 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
             if format_on_paste and self.session_async("documentRangeFormattingProvider"):
                 self._should_format_on_paste = True
         elif command_name in ("next_field", "prev_field") and args is None:
-            sublime.set_timeout_async(lambda: self.do_signature_help_async(SignatureHelpTriggerKind.Invoked))
+            sublime.set_timeout_async(lambda: self.do_signature_help_async(SignatureHelpTriggerKind.ContentChange))
         if not self.view.is_popup_visible():
             return
         if self._is_documenation_popup_open and command_name in ("move", "commit_completion", "delete_word",


rchl:
Second commit also suppresses the code that forces popup to close on certain commands:

```py
        if command_name in ("hide_auto_complete", "move", "commit_completion", "delete_word", "delete_to_mark",
                            "left_delete", "right_delete"):
            # hide the popup when `esc` or arrows are pressed
            self.view.hide_popup()
```

This code was implemented in https://github.com/sublimelsp/LSP/pull/1022 and I feel like it was done too aggressively since it meant to only affect documentation popup and not all the popups. I've fixed it by explicitly only doing that when documentation popup is open. Also removed `hide_auto_complete` since ST seems to trigger `hide_popup` itself first if one is open.

This needs to be tested more extensively but to me it feels like a much better experience.

frou:
I gave this a try and really like the second part. It's much nicer for Signature Help to not be shy and disappear prematurely while the user is still in the process of choosing an argument using completion.

I think there may be something going wrong when the Signature Help popup is visible but is then replaced by the Documentation popup (by clicking "More" on a completion item): After doing that, the Documentation popup starts to linger with old content and doesn't get hidden (or updated) when focusing another completion item. Sounds like that wasn't intended? I'm experiencing that when testing with LSP-gopls.

rchl:
Good catch. Seems like when the signature popup is open, the documentation just renders inside of it (reuses it) and then we don't realize that we've opened it. I'll see what I can do.

rchl:
I've just dropped the use of "update_popup" code paths since I don't see any flickering without it.

jwortmann:
On this branch the signature help popup doesn't cloes anymore when I press <kbd>backspace</kbd>:

https://github.com/user-attachments/assets/2907f5e0-37cd-4b69-a875-6b96e29badf9

jwortmann:
Also it doesn't hide if I have removed the closing `)` first for some reason and then type it manually (it's propably a rare use case but still might happen under some circumstances and with nested function calls, like accidentally deleted one char too much).

*Edit:* a valid use case would be with setting `"auto_match_enabled": false`

https://github.com/user-attachments/assets/e7641e3d-7a90-4ec7-84f8-b9e721051756

---

By the way, I added that code with the `meta.function-call.arguments` scope in https://github.com/sublimelsp/LSP/commit/da960d56b63d88247b293b53c8841d88e89e4e0d explicitly to prevent closing the popup too eagerly. Before that, the popup would close immediately after typing any character that is not a trigger character.

Perhaps a better solution would be to improve the Go/Ruby syntaxes in https://github.com/sublimehq/Packages instead? I did file a similar issue for JavaScript some time ago at https://github.com/sublimehq/Packages/issues/3460 and it was fixed after some time.

rchl:
Relying on scopes is just a wrong approach IMO.

EDIT: Why I think so? Because if an LSP feature has to rely on something editor specific to function properly then it feels like that LSP feature is not thought out properly. But in this case I think it's designed properly but we just didn't implement it properly.

I've changed the code so that the signature request is triggered on every change when the popup is open and so that it closes once the response is empty.

plugin/documents.py
```diff
@@ -680,38 +672,35 @@ def _on_signature_help(
...
-                    self._update_sighelp_popup(content)
-                else:
-                    self._show_sighelp_popup(content, point)
-
-            sublime.set_timeout(render_sighelp_on_main_thread)
+        sighelp = SigHelp.from_lsp(response, language_map)
+        if sighelp:
+            # Render on main thread.
+            sublime.set_timeout(lambda: self._show_sighelp_popup(sighelp, point))
+        elif self._sighelp:
```
predragnikolic: These if elif conditions are confusing me.
if sighelp, then show popup
elif self._sighelp, then hide popup 

on first glance, 
sighelp and self._sighelp look like they store the same data, 
but cause opposite actions

rchl: tried something

jwortmann:
In my opinion this implementation still makes it worse than it was before. In my experience, language servers might temporarily return `None` while typing between the brackets, for example when the sighelp is requested on a keyword argument. Maybe the language server could do better here, but it is what it is. On this branch, this results in the sighelp popup closing and re-opening several times, which is a bit annoying. While on `main` it keeps open until after the closing `)` when the `meta.function-call.arguments` scope doesn't match anymore.

Here I have tested with 2 different language servers, and which produce similar results:

1. LSP-julia (and this branch):

https://github.com/user-attachments/assets/ada398ed-5805-470d-9883-4d84cd4d50ec

2. JETLS (and this branch):

https://github.com/user-attachments/assets/cdd3bd62-ff49-4168-b4e5-3eee2fce9dac

This is how it looks on main (here with JETLS, but it's the same with LSP-julia):

https://github.com/user-attachments/assets/8299a855-74d7-4846-99b8-a37aaf900b90

---

> Relying on scopes is just a wrong approach IMO.
> 
> EDIT: Why I think so? Because if an LSP feature has to rely on something editor specific to function properly then it feels like that LSP feature is not thought out properly. But in this case I think it's designed properly but we just didn't implement it properly.

I would kindly disagree here. In my opinion there is nothing wrong with using additional information from the scopes when it can improve the behavior. The concept of "trigger characters" is quite naive (or maybe simplistic would be a better word) because it has no context of the code. For example we also request signature help when typing the `(` in an expression like `x = 2 * (3 + 4)`, which is kind of useless. Similar for completion trigger characters; I think ST has a much better concept with `auto_complete_selector` and `auto_complete_triggers` settings to decide when to trigger a completion popup. But since the LSP specs cannot assume that all clients can rely on such scopes, they had to introduce something simple like the trigger characters.

Especially here I think utilizing `meta.function-call.arguments` is an elegant way to decide whether the user is still typing in the parameters or not. What you could do in this PR instead would be to use a `None` response as an additional condition which must be satisfied in order to hide the popup. So do something like
```python
        if not new_sighelp:
            if self._sighelp and not self.view.match_selector(point, 'meta.function-call.arguments'):
                self.view.hide_popup()
            return
```

rchl:
Not very impressed with how Julia behaves. It even returns signature help when not inside 
arguments And even with second argument being the "active" parameter

Screenshot 2025-11-21 at 09 02 03

---

To counter your opinion that it worked better before, I can give examples where it works better.

- In Dockerfile the signature basically didn't work before since the argument scope was not there. Now it does (though it requires one additional change that I was contemplating whether I should do in this PR but I think I will):

https://github.com/user-attachments/assets/3ad379ba-faf4-44d2-9635-3592e7871bc3

---

> for example we also request signature help when typing the `(` in an expression like `x = 2 * (3 + 4)`, which is kind of useless.

I'd rather request more often than not request enough (as long as it doesn't trigger spurious signature helps) since it works more universally over different languages. In Dockerfile the trigger character is a space and so it will trigger a lot but I'd prefer that over trying to get the scopes right which would be impossible just based on the syntax.

---

> Especially here I think utilizing `meta.function-call.arguments` is an elegant way to decide whether the user is still typing in the parameters or not. What you could do in this PR instead would be to use a `None` response as an additional condition which must be satisfied in order to hide the popup. So do something like
> 
> ```python
>         if not new_sighelp:
>             if self._sighelp and not self.view.match_selector(point, 'meta.function-call.arguments'):
>                 self.view.hide_popup()
>             return
> ```

I'm fine with that. I think it shouldn't have any downsides.
Just a note that such check doesn't make sure that the `meta.function-call.arguments` scope is at the top of the stack. So in a JS code like this:

```js
describe('some text', () => {
    |
})
```

We are still inside that scope when at `|`. Luckily typescript server doesn't return signature in that case or we would get annoying signature help.

rchl:
To make signature help work reliably in Dockerfile I've also done 03243094e460d9559814e37eed3e14908f9f6fc4 to not skip over spaces when looking for the trigger character (Dockerfile has space as one of the `triggerCharacters`). The origin of that logic is very old - https://github.com/sublimelsp/LSP/pull/163

This would hopefully not cause any issues (especially with the new logic where we don't close the popup as eagerly).

jwortmann:
> To counter your opinion that it worked better before, I can give examples where it works better.

Sure, but it should not come with the cost of regressing the behavior for other examples that worked fine before. That's why I was giving these examples for which this PR still needed some fixes. But with the latest changes it works for me.

By the way, the [registration options](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#signatureHelpOptions) not only have the `triggerCharacters`, but also `retriggerCharacters`. It seems like we ignore these completely. If we would want to follow the description from the specs closely, we should probably only retrigger on those, and not after every buffer change:
```typescript
    /**
	 * List of characters that re-trigger signature help.
	 *
	 * These trigger characters are only active when signature help is already
	 * showing. All trigger characters are also counted as re-trigger
	 * characters.
	 *
	 * @since 3.15.0
	 */
	retriggerCharacters?: string[];
```

I guess this implicitly assumes that the client uses some custom logic (like the scope) to decide how long to keep the signature help visible. But since this would not work with poor syntax packages, the current logic in this PR to retrigger after every change seems more flexible.

plugin/documents.py
```diff
@@ -971,7 +954,7 @@ def _on_view_updated_async(self) -> None:
         if userprefs().document_highlight_style:
             self._when_selection_remains_stable_async(
                 self._do_highlights_async, first_region, after_ms=self.debounce_time)
-        self.do_signature_help_async(manual=False)
+        self.do_signature_help_async(manual=self._sighelp is not None)
```
jwortmann: With this change, the parameter name `manual` loses its meaning. Maybe rename to `force`?

rchl:
> I guess this implicitly assumes that the client uses some custom logic (like the scope) to decide how long to keep the signature help visible.

Or perhaps more likely it assumes that the client won't close the signature help seemingly for no reason like ST does when auto-pairing for example? As far as I can see VSCode manages to keep it open regardless of what is being typed and also triggers it on each change (for Pylance server that doesn't advertise support for retrigger at least).

jwortmann:
https://github.com/sublimelsp/LSP/blob/4ffae29c97b002ef19def940c5219cce43bfd7d2/plugin/core/signature_help.py#L33

rchl:
> https://github.com/sublimelsp/LSP/blob/4ffae29c97b002ef19def940c5219cce43bfd7d2/plugin/core/signature_help.py#L33



thanks!

plugin/core/sessions.py
```diff
@@ -1713,7 +1717,7 @@ def run_async() -> None:
                 listener = session_view.listener()
                 if not listener:
                     return
-                listener.do_signature_help_async(manual=False)
+                listener.do_signature_help_async(force=True)
```
rchl: Changed the parameter value here but I guess this is really more of an explicit invocation?

jwortmann: I forgot that we also use this parameter for the `triggerKind` value in the request. I think the best would then be to refactor that method so that the early return is removed, and put the condition `previous_char in triggers` before the call to this method in `_on_view_updated_async`. Then the parameter name could stay as `manual`.

rchl: So move all the code related to checking trigger characters outside of the `do_signature_help_async`? I like the idea if that's what you are suggesting but we'd still have to pass the `triggerCharacter` to this function (better) or otherwise calculate it twice (worse).

That said, I also see that there is another `SignatureHelpTriggerKind` kind called `ContentChange`. It seems perfect for this case when we are re-triggering based on sighelp being active. So I'm thinking to even pass kind as an argument... I'll try something and we'll see if it looks good.

rchl: Not sure if that's what you meant but I've changed it so that you can call `do_signature_help_async` with either "TriggerCharacter" kind AND a char or with other kind and no char. A bit elaborate typing enforces that char is passed when passing TriggerCharacter kind.

plugin/documents.py
```diff
@@ -600,12 +597,11 @@ def on_post_text_command(self, command_name: str, args: dict[str, Any] | None) -
             if format_on_paste and self.session_async("documentRangeFormattingProvider"):
                 self._should_format_on_paste = True
         elif command_name in ("next_field", "prev_field") and args is None:
-            sublime.set_timeout_async(lambda: self.do_signature_help_async(manual=True))
+            sublime.set_timeout_async(lambda: self.do_signature_help_async(SignatureHelpTriggerKind.Invoked))
```
jwortmann: Shouldn't this also be a `SignatureHelpTriggerKind.ContentChange` (user jumped to next/previous tab stop in a snippet)? It's not really invoked manually or by a server command.

> "Signature help was triggered by the cursor moving [...]"

plugin/documents.py
```diff
@@ -648,98 +644,90 @@ def _on_query_completions_resolved_async(
...
+        if not new_sighelp:
+            if self._sighelp and not self.view.match_selector(point, 'meta.function-call.arguments'):
+                self.view.hide_popup()
+            return
+        # Render on main thread.
+        sublime.set_timeout(lambda: self._show_sighelp_popup(new_sighelp, point))
 
-    def _show_sighelp_popup(self, content: str, point: int) -> None:
+    def _show_sighelp_popup(self, sighelp: SigHelp, point: int) -> None:
+        content = sighelp.render(self.view)
```
jwortmann: This puts more work onto the main thread. Could you keep the content = ... line before the set_timeout? The `render` method name is probably a bit misleading...

plugin/core/sessions.py
jwortmann: Yes, I think it's a good idea to pass the trigger_kind to this method.

plugin/documents.py
rchl: No, I wouldn't say this is ContentChange. The file content doesn't change in this case and the "cursor moving" refers to selection moving, not pressing the cursor key (which in this case doesn't move selection).

plugin/documents.py
rchl: The only reason for doing it that way was to avoid passing both sighelp and content but I can do that.

plugin/documents.py
```diff
@@ -703,14 +703,11 @@ def _on_signature_help(
...
                 self.view.hide_popup()
             return
-        # Render on main thread.
-        sublime.set_timeout(lambda: self._show_sighelp_popup(new_sighelp, point))
-
-    def _show_sighelp_popup(self, sighelp: SigHelp, point: int) -> None:
-        content = sighelp.render(self.view)
-        # TODO: There are a bunch of places in the code where we assume we have exclusive access to a popup. The reality
-        # is that there is really only one popup per view. Refactor everything that interacts with the popup to a common
-        # class.
```
rchl, line 726: Removed as I feel like this comment is not that relevant anymore.

plugin/documents.py
rchl: Sorry, you are right, I thought this is handling for selecting different signature in case there are multiple.

jwortmann: To clarify, this case is pressing <kbd>tab</kbd> (or <kbd>shift</kbd>+<kbd>tab</kbd>) when in a snippet, and then the caret jumps to the next tab stop (which is most likely the next parameter) in a snippet. Isn't this selection moving?

plugin/documents.py
rchl: Well, I guess it could still be relevant if the idea was to combine content's of different popups and show them all at once...

PR #2687, opened 2025-11-17
refactor: aggregate patterns matching same root and kind

rchl:
If server specifies multiple patterns with same base path and kind then collapse those into a single entry. Depending on watcher implementation that can be more efficient since the watcher will only set up one watcher for that directory instead of multiple.

For example in case of biome there are 4 patterns that can be bundled together:

```
{
  "registrations": [
    {
      "id": "biome_did_change_watched_files",
      "method": "workspace/didChangeWatchedFiles",
      "registerOptions": {
        "watchers": [
          {
            "globPattern": {
              "baseUri": {
                "name": "typescript-language-server",
                "uri": "file:///usr/local/workspace/github/typescript-language-server"
              },
              "pattern": "**/biome.{json,jsonc}"
            },
            "kind": 7
          },
          {
            "globPattern": {
              "baseUri": {
                "name": "typescript-language-server",
                "uri": "file:///usr/local/workspace/github/typescript-language-server"
              },
              "pattern": ".editorconfig"
            },
            "kind": 7
          },
          {
            "globPattern": {
              "baseUri": {
                "name": "typescript-language-server",
                "uri": "file:///usr/local/workspace/github/typescript-language-server"
              },
              "pattern": "**/.gitignore"
            },
            "kind": 7
          },
          {
            "globPattern": {
              "baseUri": {
                "name": "typescript-language-server",
                "uri": "file:///usr/local/workspace/github/typescript-language-server"
              },
              "pattern": "**/.ignore"
            },
            "kind": 7
          }
        ]
      }
    },
 ]
}
```

Ideally the watcher implementation would have smart logic to do something like that itself (and I'm planning to look into that) but this seems like a quick win anyway.

### DIFF
--- a/tests/test_file_watcher.py
+++ b/tests/test_file_watcher.py
@@ -4,6 +4,7 @@ from LSP.plugin import FileWatcher
 from LSP.plugin import FileWatcherEvent
 from LSP.plugin import FileWatcherEventType
 from LSP.plugin import FileWatcherProtocol
+from LSP.plugin import parse_uri
 from LSP.plugin.core.file_watcher import file_watcher_event_type_to_lsp_file_change_type
 from LSP.plugin.core.file_watcher import register_file_watcher_implementation
 from LSP.plugin.core.types import ClientConfig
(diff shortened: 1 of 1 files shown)


plugin/core/sessions.py
```diff
@@ -2236,20 +2237,24 @@ def register_file_system_watchers(self, registration_id: str, watchers: list[Fil
         if not self._watcher_impl:
             return
         self.unregister_file_system_watchers(registration_id)
-        file_watchers: list[FileWatcher] = []
+        # Aggregated list of patterns by base path and kind.
+        aggregated_watchers: dict[tuple[str, WatchKind], list[str]] = {}
         for config in watchers:
-            kind = lsp_watch_kind_to_file_watcher_event_types(config.get("kind") or DEFAULT_KIND)
+            kind = config.get("kind") or DEFAULT_KIND
```
jwortmann: Not changed in this PR, but `DEFAULT_KIND` is a very generic name, could we change that to `DEFAULT_WATCHKIND` or `DEFAULT_WATCH_KIND`?

predragnikolic:
Rafal can you please add a test(s :) ) for this?

predragnikolic (approved):

PR #2688, opened 2025-11-17
Limit some requests to a single session

### LINKED ISSUE #2534: Duplicate color provider references
**Describe the bug**

When using two or more language servers which have color provider functionality, there are situations when duplicate references are displayed. This is most recently visible when using [LSP-volar](https://github.com/sublimelsp/LSP-volar) and [LSP-sass](https://github.com/niksy/LSP-sass) for Vue SFC files.

**To Reproduce**

1. Create Vue SFC file (.vue) with following content
```vue
<template>
	<div></div>
</template>
<style lang="scss" scoped>
$color: #f00;
body {
	color: $color;
}
</style>
```
2. Follow [instructions](https://github.com/niksy/LSP-sass#using-with-vue-sfc) on how to use LSP-sass for Vue SFC files
3. Observe `$color` variable

**Expected behavior**

Only one color provider reference is displayed.

**Screenshots**

Actual:

Screenshot 2024-10-18 at 18 04 23

Expected:

Screenshot 2024-10-18 at 18 04 05

**Logs**

N/A

- OS: macOS 14.6.1
- Sublime Text version: 4180
- LSP version: 2.2.0
- Language servers used: LSP-volar, LSP-sass

**Additional context**

As mentioned in https://github.com/sublimelsp/repository/pull/120#issuecomment-2422619311, this is probably related to Sublime LSP. My initial reaction was that it’s related to how Volar packages language servers and since there are no hooks with which you can control those servers (at least I haven’t found them), there isn’t really much you can do except disable `colorProvider` in one of the servers, but that relates to all server functionalities. What you actually want is to disable color provider on SFC Sass style blocks.

jwortmann:
For some features (namely color boxes, inlay hints and semantic tokens) and if there are multiple servers running at the same time, it does not make sense to send requests to all sessions and then render these features for each session individually. For example this currently results in duplicated color boxes and inlay hints, and for semantic tokens it would result in the sessions competing each other to overwrite semantic highlighting regions. Instead, the best session should be determined for each of those features and then the request should only be sent to that session.

Fixes #2534 (*)

\* I need to mention that this solution does not work for the very first file that gets opened. This is because at that time the sessions are not yet initialized, so when the first session is ready it is the only session and all supported features are requested. Then when another session initializes, it might become the new "best session" for some of the features and therefore the requests are sent to that session as well. Not sure if there is any good workaround or if we should just leave it like that. So it's basically the status quo for the first file behavior but then it should work as expected after that.

Also note that in case of the color box example from the linked issue you still might want to disable the colorProvider in one of the configs to ensure that the "best session" is actually the right one (it looks like LSP-volar doesn't provide color boxes for all parts of the file).

### DIFF
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -436,7 +436,7 @@ class SessionBuffer:
         if userprefs().semantic_highlighting:
             self.semantic_tokens.needs_refresh = True
         else:
-            self._clear_semantic_tokens_async()
+            self.clear_semantic_tokens_async()
         for sv in self.session_views:
             sv.on_userprefs_changed_async()
 
@@ -485,6 +485,9 @@ class SessionBuffer:
             return
         phantoms = [lsp_color_to_phantom(view, color_info) for color_info in response]
         sublime.set_timeout(lambda: self._color_phantoms.update(phantoms))
+
+    def clear_color_boxes_async(self) -> None:
+        sublime.set_timeout(lambda: self._color_phantoms.update([]))
 
     # --- textDocument/documentLink ------------------------------------------------------------------------------------
 
@@ -783,7 +786,7 @@ class SessionBuffer:
     def get_semantic_tokens(self) -> list[SemanticToken]:
         return self.semantic_tokens.tokens
 
-    def _clear_semantic_tokens_async(self) -> None:
+    def clear_semantic_tokens_async(self) -> None:
         for sv in self.session_views:
             self._clear_semantic_token_regions(sv.view)
 
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -249,10 +249,24 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         if session.config.name not in self._session_views:
             self._session_views[session.config.name] = SessionView(self, session, self._uri)
             if buf := self.view.buffer():
-                text_change_listener = TextChangeListener.ids_to_listeners.get(buf.buffer_id)
-                if text_change_listener:
+                if text_change_listener := TextChangeListener.ids_to_listeners.get(buf.buffer_id):
                     text_change_listener.view_listeners.add(self)
             self.view.settings().set("lsp_active", True)
+            # Check whether this session is the new best session for color boxes, inlay hints, and semantic tokens. If
+            # that is the case, remove the color boxes, inlay hints or semantic tokens from the previously best session.
+            request_flags = self.get_request_flags(session)
+            if request_flags & RequestFlags.DOCUMENT_COLOR:
+                for sb in self.session_buffers_async('colorProvider'):
+                    if sb.session != session:
+                        sb.clear_color_boxes_async()
+            if request_flags & RequestFlags.INLAY_HINT:
+                for sb in self.session_buffers_async('inlayHintProvider'):
+                    if sb.session != session:
+                        sb.remove_all_inlay_hints()
+            if request_flags & RequestFlags.SEMANTIC_TOKENS:
+                for sb in self.session_buffers_async('semanticTokensProvider'):
+                    if sb.session != session:
+                        sb.clear_semantic_tokens_async()
 
     def on_session_shutdown_async(self, session: Session) -> None:
         if removed_session := self._session_views.pop(session.config.name, None):
@@ -358,11 +372,11 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
 
     def get_request_flags(self, session: Session) -> RequestFlags:
         request_flags = RequestFlags.NONE
-        if session == self.session_async('colorProvider'):
+        if session == self.session_async('colorProvider', 0):
             request_flags |= RequestFlags.DOCUMENT_COLOR
-        if session == self.session_async('inlayHintProvider'):
+        if session == self.session_async('inlayHintProvider', 0):
             request_flags |= RequestFlags.INLAY_HINT
-        if session == self.session_async('semanticTokensProvider'):
+        if session == self.session_async('semanticTokensProvider', 0):
             request_flags |= RequestFlags.SEMANTIC_TOKENS
         return request_flags
 


plugin/session_buffer.py
```diff
@@ -157,7 +158,7 @@ def __init__(self, session_view: SessionViewProtocol, buffer_id: int, uri: Docum
         self._dynamically_registered_commands: dict[str, list[str]] = {}
         self._supported_commands: set[str] = set()
         self._update_supported_commands()
-        self._check_did_open(view)
+        sublime.set_timeout_async(lambda: self._check_did_open(view))
```
rchl, line 162: I think that if this is required then at least some clarification comment is warranted since this already runs on async thread.

jwortmann: https://github.com/sublimelsp/LSP/pull/2688/commits/9c9e04d3f35fa79d5417909ac356bca92fe10c25

The SessionBuffer is created at
https://github.com/sublimelsp/LSP/blob/c48576f96fc42c59b79129b29b07d88044497474/plugin/session_view.py#L68

i.e. before the corresponding `__init__` method in SessionView has finished. To make this PR work we need to wait until the SessionView exists before running the requests.

rchl: Make sense but seems like similar problems could happen with `SessionView` since it's also doing a bunch of stuff like being registered in `SessionBuffer` and `Session` before it's fully initialized.

Not something to address here but it feels brittle and I remember we had similar issues before.

plugin/session_buffer.py
```diff
@@ -185,10 +187,14 @@ def _check_did_open(self, view: sublime.View) -> None:
             self.opened = True
             version = view.change_count()
             self._last_synced_version = version
-            self._do_color_boxes_async(view, version)
+            request_flags = self._get_request_flags(view)
+            if request_flags & RequestFlags.DOCUMENT_COLOR:
+                self._do_color_boxes_async(view, version)
```
rchl, line 192: While reviewing I've noticed that this line is unused:

https://github.com/sublimelsp/LSP/blob/c48576f96fc42c59b79129b29b07d88044497474/plugin/documents.py#L167

jwortmann: I removed that line

jwortmann:
I added the check for the best session to the inlay hint toggle too.

Probably the same needs also be done after inlay hint or semantic tokens refresh requests from the server. That would be at
https://github.com/sublimelsp/LSP/blob/c48576f96fc42c59b79129b29b07d88044497474/plugin/core/sessions.py#L2138-L2154

and at
https://github.com/sublimelsp/LSP/blob/c48576f96fc42c59b79129b29b07d88044497474/plugin/documents.py#L395-L400

predragnikolic:
So, if there are 3 language servers in one document that support definition/color-highlight  capability, should the request be send:
a) to the best language server
b) to all language servers

If a), then:
- how will we know what the best language server to handle the request is?
- what if other servers also contain a useful response for that request?

While working on Mir API, I draw the following conclusion:
to go with option b) and to send requests to all servers, and later combine the results in a way that makes sense, or strip down the conflicting parts in a later stage.

Although I cannot give real life examples, lets give one example for inlay-hints.
If there are two language servers that provide inlay hints for one file,
I would find it more useful to see results form both servers, than to see results from only one of them.

jwortmann:
> * how will we know what the best language server to handle the request is?

The current implementation decides based on which `"selector"` / `"priority_selector"` in the configs matches best:
https://github.com/sublimelsp/LSP/blob/4ffae29c97b002ef19def940c5219cce43bfd7d2/plugin/core/registry.py#L24-L33

> send requests to all servers, and later combine the results in a way that makes sense, or strip down the conflicting parts in a later stage.

How would you combine multiple color boxes or inlay hints? Currently they are just drawn as multiple phantoms at the same positions, which is bad. Even if you compare and then reject some parts of the server responses you still have to decide which one of the color boxes / inlay hints to keep, so we're back at step one to basically choose the "best session" between multiple responses for the same position. And in the case of semantic tokens it gets even more complicated because the regions might not overlap exactly and the complexity would just be way too much with the delta-requests and so on.

Note that selecting the best session and sending the request only to that session is what we already do for various features like document highlight (which is implemented directly in the listener class), goto definition, find references, and I do believe that it is the best strategy as well for the 3 features that are handled in this PR.

---

By the way, the limitation that I described in the opening comment should be fixed now with the latest commit:

> I need to mention that this solution does not work for the very first file that gets opened. This is because at that time the sessions are not yet initialized, so when the first session is ready it is the only session and all supported features are requested. Then when another session initializes, it might become the new "best session" for some of the features and therefore the requests are sent to that session as well. Not sure if there is any good workaround or if we should just leave it like that. So it's basically the status quo for the first file behavior but then it should work as expected after that.

jwortmann:
Can I merge this?

rchl:
> Can I merge this?

Yes please.

Just curious how "best" is determined in case when neither of the servers specifies priority selector? I guess it's the order of server registration during LSP initialization. If so then I wonder how stable is it (based on alphabetical order or seemingly random)?

That question doesn't block the merging though.

rchl (approved):

jwortmann:
According to https://docs.python.org/3/library/functions.html#max `max` returns the first one, and it looks like the sessions iterable is based on the `self._session_views` dict from the listener. And on Python 3.8 dicts should preserve the insertion order.
https://github.com/sublimelsp/LSP/blob/1094d05acc67047d9dab3b562b1b85d96cc41855/plugin/documents.py#L247-L250

This method is called from
https://github.com/sublimelsp/LSP/blob/1094d05acc67047d9dab3b562b1b85d96cc41855/plugin/core/windows.py#L193-L197
but here `self._sessions` is a `set`, which order is deterministic, but not based on insertion order. So I guess it eventually depends on the hash value of the session object.

---

tdlr; if my assumptions are correct, it should be stable for the same set of servers and server configs, but if you change some part of a config (like the server settings), the order might change.

rchl:
A bit late but just realized that in vue files both LSP-vue and LSP-typescript provide semantic tokens and those tokens don't overlap each other (typescript handles js/ts code and vue handles the html part). So this no longer works.

EDIT: Actually there is some overlap like on line 7 column 23-31 but still some highlighting is missing if only getting tokens from one server.

Screenshot 2025-11-22 at 23 56 32

jwortmann:
Hm... so should we revert the change for semantic tokens then? On the other hand, if there is some overlap and the token types and modifiers are not the same from both servers for the overlapping tokens, I assume that it could result in flickering color changes while typing. That would probably also look buggy.

niksy:
Somewhat related question: does this handle duplicate autocomplete suggestions?

For example, this is Vue SFC file with both LSP-vue and LSP-some-sass running:

Screenshot 2025-11-23 at 10 53 38

Properties like `color` and `colum-gap` are duplicated since they are coming from both CSS support in LSP-vue and extended CSS support with Sass specific completions in LSP-some-sass.

If I disable LSP-some-sass, I get only one instance of completions:

Screenshot 2025-11-23 at 10 54 12

How should this be handled? Is this something that each LSP server should provide as an option to disable or adjust?

If this should be posted as separate question or issue, sorry for sidetracking!

jwortmann:
> Somewhat related question: does this handle duplicate autocomplete suggestions?

No, auto-complete is not affected by this pull request. Different servers could provide distinct sets of completion items, and there is no problem with rendering them all in the popup, so I don't think it is necessary to limit the auto-complete request to one server. Maybe we could think about deduplicating items from multiple servers based on the `label` property (that is shown in the popup), but I'm not sure if that makes sense or would work well, because there could be multiple items with the same labels but different edits (like auto-import), or different kinds (for example a variable name that gets inserted directly and a function with the same name that automatically inserts the parentheses). Also there could still be completion items from other sources like snippets, sublime-completion files, or other plugins.

But you could manually disable auto-complete for one of the servers via
```json
"disabled_capabilities": {
    "completionProvider": true
}
```
in the settings file.

niksy:
> But you could manually disable auto-complete for one of the servers via
> 
> ```json
> "disabled_capabilities": {
>     "completionProvider": true
> }
> ```
> 
> in the settings file.

I’m aware of this, but disabling completions for LSP-vue will disable them for all SFC blocks and not just for `style` block, and disabling completions in LSP-some-sass will disable Sass specific completions.

> Maybe we could think about deduplicating items from multiple servers based on the label property (that is shown in the popup), but I'm not sure if that makes sense or would work well, because there could be multiple items with the same labels but different edits (like auto-import), or different kinds (for example a variable name that gets inserted directly and a function with the same name that automatically inserts the parentheses).

Yeah, this could work as an opt-in feature, but only if multiple variables are matched, e.g. label and kind?

But I don’t think that’s a large issue, just something I noticed, am aware why it’s happening and try to see if it can be adjusted.

rchl:
Think it would make sense to ask for such feature in the Vue server. It runs multiple servers for different parts of the file so I think it makes sense in that particular case. Probably would also help users of other editors.

niksy:
> Think it would make sense to ask for such feature in the Vue server. It runs multiple servers for different parts of the file so I think it makes sense in that particular case. Probably would also help users of other editors.

There is actually (undocummented) option in Some Sass language server intended just for this case 😄 

https://github.com/wkillerud/some-sass/discussions/364#discussioncomment-15059932

PR #2690, opened 2025-11-20
Don't expand range to cover all diagnostics for code action requests

rchl:
Don't expand `range` to cover all diagnostics touching the selection when making the `textDocument/codeAction` request. Just pass the actual selection.

Passing expanded range was implemented by me in https://github.com/sublimelsp/LSP/pull/1129 based on an issue I saw with the typescript server. It later [turned out](https://github.com/typescript-language-server/typescript-language-server/issues/885) that it was the responsibility of the server to expand the range to cover all passed diagnostics. This is now fixed in the typescript server and so removing it from the client.

### DIFF
--- a/tests/test_code_actions.py
+++ b/tests/test_code_actions.py
@@ -369,9 +369,9 @@ class CodeActionsTestCase(TextDocumentTestCase):
         )
         params = yield from self.await_message('textDocument/codeAction')
         self.assertEqual(params['range']['start']['line'], 1)
-        self.assertEqual(params['range']['start']['character'], 0)
+        self.assertEqual(params['range']['start']['character'], 1)
         self.assertEqual(params['range']['end']['line'], 1)
-        self.assertEqual(params['range']['end']['character'], 0)
+        self.assertEqual(params['range']['end']['character'], 1)
         self.assertEqual(len(params['context']['diagnostics']), 1)
 
     def test_applies_code_action_with_matching_document_version(self) -> Generator:


plugin/documents.py
```diff
@@ -272,41 +272,34 @@ def _diagnostics_async(
...
-        region: sublime.Region
-    ) -> tuple[list[tuple[SessionBufferProtocol, list[Diagnostic]]], sublime.Region]:
-        covering = sublime.Region(region.begin(), region.end())
+        self, region: sublime.Region
+    ) -> list[tuple[SessionBufferProtocol, list[Diagnostic]]]:
         result: list[tuple[SessionBufferProtocol, list[Diagnostic]]] = []
         for sb, diagnostics in self._diagnostics_async():
             intersections: list[Diagnostic] = []
             for diagnostic, candidate in diagnostics:
                 # Checking against points is inclusive unlike checking whether region intersects another region
```
predragnikolic, line 281: `diagnostics_intersecting_region_async` and `diagnostics_touching_point_async` are quiet similar,
but `diagnostics_intersecting_region_async` is missing this check:

```py
if diagnostic_severity(diagnostic) > max_diagnostic_severity_level:
   continue
```
Should we add the missing diagnostic_severity check to `diagnostics_intersecting_region_async`

Or should we remove the the check from `diagnostics_touching_point_async`? 

Although we have not noticed any issue so far, it does seem like the behavior can cause inconsistencies in theory.

predragnikolic: even if we want to take action I would avoid doing it in this PR, as the PR is fine.

rchl: No, that's a good point. Since the region will typically be empty now, we'll use a different code path (`diagnostics_touching_point_async` instead of `diagnostics_intersecting_region_async`) and we'll filter the diagnostics based on LSP settings which we don't want to do when getting code actions for example. I suppose we will want to add some explicit argument for choosing desired behavior.

rchl: Actually, it should work as excepted since when we don't pass `max_diagnostic_severity_level` to `diagnostics_touching_point_async` then it will return all diagnostics.

Then I wouldn't touch it here not to make PR harder to follow.

PR #2694, opened 2025-11-24
refactor: merge diagnostics getter functions

rchl:
Combine 3 different but very similar diagnostic getter functions (some internal, some public) into a single one.

Should have no runtime changes.

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -287,9 +287,7 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
 
     @override
     def get_diagnostics_async(
-        self,
-        location: sublime.Region | int,
-        max_diagnostic_severity_level: DiagnosticSeverity = DiagnosticSeverity.Hint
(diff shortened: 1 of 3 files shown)


plugin/documents.py
```diff
@@ -285,31 +285,24 @@ def _diagnostics_async(
...
-                result.append((sb, intersections))
-        return result
-
-    def diagnostics_touching_point_async(
-        self, pt: int, max_diagnostic_severity_level: int = DiagnosticSeverity.Hint
+    @override
+    def get_diagnostics_async(
+        self,
+        location: sublime.Region | int,
+        max_diagnostic_severity_level: DiagnosticSeverity = DiagnosticSeverity.Hint
```
jwortmann: Shouldn't this still be an `int`, because 0 is an allowed value and is not a DiagnosticSeverity. Then you also don't need the `cast` below. And comparing DiagnosticSeverity against int is fine, because DiagnosticSeverity is an IntEnum.

rchl: so there was a reason for this :)

PR #2695, opened 2025-11-28
Show WorkspaceEdit and TextEdit description in undo menu

jwortmann:
When ST runs a TextCommand, the command's `description` is used as a label in the undo/redo menu. If `description` isn't defined in the command, it used a label derived from the command name instead. So when a language server requests `workspace/applyEdit` or the user triggers a WorkspaceEdit via code actions or rename, currently LSP only shows a generic label in the menu:

before

---

[ApplyWorkspaceEditParams](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#applyWorkspaceEditParams) have an optional `label` field which is intended to be used for the undo stack.

This PR also adds support for [AnnotatedTextEdit](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#annotatedTextEdit) which allows different labels for different edits/URIs. In that case, the URI-specific label overrides the general label from ApplyWorkspaceEditParams.

Example from `workspace/applyEdit` server request:

after1

Example form code action (uses the code action title for the label):

after2

Example from rename (uses old and new name for the label):

after3

---

Question: should we additionally prepend "LSP: " to the label? I currently would tend not to do that, but I'd be interested in your opinions.

### DIFF
--- a/plugin/formatting.py
+++ b/plugin/formatting.py
@@ -203,7 +203,7 @@ class LspFormatDocumentRangeCommand(LspTextCommand):
             if session := self.best_session('documentRangeFormattingProvider.rangesSupport'):
                 request = text_document_ranges_formatting(self.view)
                 session.send_request(
-                    request, lambda response: apply_text_edits(self.view, response, label="Format Selections"))
+                    request, lambda response: apply_text_edits(self.view, response, label="Format Selection"))
 
 
 class LspFormatCommand(LspTextCommand):
@@ -218,11 +218,7 @@ class LspFormatCommand(LspTextCommand):
         return self.is_enabled(event, point)
 
     def description(self, **kwargs) -> str:
-        if self._range_formatting_available():
-            if has_single_nonempty_selection(self.view):
-                return "Format Selection"
-            return "Format Selections"
-        return "Format File"
+        return "Format Selection" if self._range_formatting_available() else "Format File"
 
     def run(self, edit: sublime.Edit, event: dict | None = None) -> None:
         command = 'lsp_format_document_range' if self._range_formatting_available() else 'lsp_format_document'


plugin/core/sessions.py
```diff
@@ -1933,36 +1937,44 @@ def _apply_code_action_async(
                                                                is_refactoring=is_refactoring))
         return promise
 
-    def apply_workspace_edit_async(self, edit: WorkspaceEdit, is_refactoring: bool = False) -> Promise[None]:
+    def apply_workspace_edit_async(
+        self, edit: WorkspaceEdit, label: str | None = None, is_refactoring: bool = False
+    ) -> Promise[None]:
```
jwortmann, line 1943: Should we make `label` and `is_refactoring` keyword-only arguments?

jwortmann: I made them keyword-only arguments in https://github.com/sublimelsp/LSP/pull/2695/commits/b7725b40cbd3a159e0ca51862c69d88267b0f5d1.

In case any plugin currently passes a value for `is_refactoring` as a positional argument, like `session.apply_workspace_edit_async(edit, True)`, then it needs to be updated.

plugin/rename.py
```diff
@@ -302,7 +303,12 @@ def _render_rename_panel(
                 'commands': [
                     [
                         'lsp_apply_workspace_edit',
-                        {'session_name': session_name, 'edit': workspace_edit, 'is_refactoring': True}
+                        {
+                            'session_name': session_name,
+                            'edit': workspace_edit,
+                            'description': description,
```
rchl: Looks like `LspApplyWorkspaceEditCommand` takes a `label` currently, not `description`. Which is a weird discrepancy also since `LspApplyDocumentEditCommand` takes `description`.

plugin/core/edit.py
```diff
@@ -9,13 +11,15 @@
...
-WorkspaceChanges = Dict[str, Tuple[List[Union[TextEdit, AnnotatedTextEdit]], Optional[int]]]
+WorkspaceChanges = Dict[str, Tuple[List[Union[TextEdit, AnnotatedTextEdit]], Optional[str], Optional[int]]]
 
 
-def parse_workspace_edit(workspace_edit: WorkspaceEdit) -> WorkspaceChanges:
+def parse_workspace_edit(workspace_edit: WorkspaceEdit, label: str | None = None) -> WorkspaceChanges:
     changes: WorkspaceChanges = {}
     document_changes = workspace_edit.get('documentChanges')
     if isinstance(document_changes, list):
+        change_annotations: dict[ChangeAnnotationIdentifier, ChangeAnnotation] = \
```
rchl: Do we need explicit typing here? Isn't inferred type enough?

jwortmann: Indeed can be inferred

plugin/rename.py
jwortmann: Thanks, I think I got confused from partly adjusting it after writing the initial code.
I changed it now to `label` everywhere (I think).

rchl:
> Question: should we additionally prepend "LSP: " to the label? I currently would tend not to do that, but I'd be interested in your opinions.

Not necessary, I think.

plugin/core/sessions.py
rchl: One could wonder if `label` should be a keyword argument...

At least with the current uses, it results in having duplicate info in the callers like `label=label` or similar. But then the same can be said about the `is_refactoring` argument.

(Linters typically only complain about boolean arguments (https://docs.astral.sh/ruff/rules/boolean-default-value-positional-argument/) but technically one could argue that it can also be a problem with other types of arguments. For this particular case, the callers pass nice variable names for both of them so this only adds some redundant info. I don't know what to think about it ;))

jwortmann: I don't have a strong preference. But if we make `label` a keyword-only argument now too, we can always change it later to allow as positional argument, without breaking anything.

jwortmann:
I added labels for changing the color format, inserting inlay hints and for the formatting commands too.

And also made a small adjustment for the format range(s) command label to not distinguish between "Selection"/"Selections" for single-region or multiple-region selection (because ST API and UI just calls both variants "Selection").

PR #2498, opened 2024-06-25
Add willRename and didRename fileOperations

predragnikolic:
closes #2199

Here is a example video:

https://github.com/sublimelsp/LSP/assets/22029477/441972ba-939c-40e0-899d-13629034a838

Notice the import changing in file `a.ts` when we rename the `b.ts` file.

[example.zip](https://github.com/user-attachments/files/15974460/example.zip)

PR #2696, opened 2025-11-30
Don't delay didOpen until next message loop iteration

rchl:
In https://github.com/sublimelsp/LSP/pull/2688/files#r2535127726 a `set_timeout_async` was introduced to ensure that didOpen notification is sent __after__ `SessionView` is initialized.

Thinking about it now (and after seeing it possibly introducing some errors in the https://github.com/sublimelsp/LSP/pull/2498 PR), I think its too risky to delay this request until next message loop invocation since in theory it allows other requests to creep in before it's called and make things out of order.

I think that a better solution is something like I did here where we trigger it from the same message loop iteration but only after `__init__` has completed.

I haven't thought too much about naming of the methods so those are likely not ideal but I just wanted to get out some solution for discussion. And also I'm not a big fan of such two-step initialization process normally due to the fact that it's not obvious that one needs to call extra methods when initializing and it's not enforced in any way. But of course we are not really gonna change this code often.

### DIFF
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -444,19 +444,15 @@ class SessionBuffer:
     def some_view(self) -> sublime.View | None:
         if not self.session_views:
             return None
-        view: sublime.View | None = None
         # Prefer active view if possible
         active_view = self.session.window.active_view()
         for sv in self.session_views:
             if sv.view == active_view and sv.view.is_valid():
-                view = active_view
-                break
-        else:
-            for sv in self.session_views:
-                if sv.view.is_valid():
-                    view = sv.view
-                    break
-        return view
+                return active_view
+        for sv in self.session_views:
+            if sv.view.is_valid():
+                return sv.view
+        return None
 
     def _if_view_unchanged(self, f: Callable[[sublime.View, Any], None], version: int) -> CallableWithOptionalArguments:
         """


rchl:
Also added 8fa8d749823fc8fcf9fcdcd65f18629d62d7a3a6 which catches some cases of using invalid view but it can still happen in long-running code in async thread.

plugin/session_buffer.py
```diff
@@ -443,13 +444,18 @@ def on_userprefs_changed_async(self) -> None:
...
         # Prefer active view if possible
         active_view = self.session.window.active_view()
         for sv in self.session_views:
             if sv.view == active_view:
-                return active_view
-        for sv in self.session_views:
-            return sv.view
+                view = active_view
+                break
+        if not view:
```
jwortmann:
```suggestion
        else:
```

I know that many people find this Python feature confusing, but I do like it and here it fits well.

Nevertheless I'll approve regardless if you apply this change or not (otherwise I'll try to sneak it in at some later time :) ).

plugin/session_buffer.py
```diff
@@ -443,13 +444,18 @@ def on_userprefs_changed_async(self) -> None:
...
-                return active_view
-        for sv in self.session_views:
-            return sv.view
+                view = active_view
+                break
+        else:
+            for sv in self.session_views:
+                view = sv.view
+                break
+        return view if view and view.is_valid() else None
```
jwortmann: Actually you could move the `is_valid()` check before assigning the `view` above. Then this function would allow to return another valid view if there is one.

jwortmann: And now to simplify you could return the view early instead of `break`... basically like it was initially, just with the additional `is_valid` check ;)

PR #2698, opened 2025-12-02
Fix duplicate requests on file open with dynamic registration

jwortmann:
I noticed that there are duplicate `textDocument/diagnostic` and `textDocument/codeLens` requests when a file gets opened and the providers for these features were registered dynamically. This happens because one time the requests are triggered by https://github.com/sublimelsp/LSP/blob/1554771471ea0c7077d910995d7a674889625637/plugin/session_buffer.py#L181

but then also another time when the capabilities are assigned to the SessionBuffer at https://github.com/sublimelsp/LSP/blob/1554771471ea0c7077d910995d7a674889625637/plugin/session_buffer.py#L274-L277

which gets triggered separately from https://github.com/sublimelsp/LSP/blob/1554771471ea0c7077d910995d7a674889625637/plugin/core/sessions.py#L1443-L1446

(the requests caused by that might even be sent before `didOpen`)

---

Not sure if there is a better solution than passing a boolean flag to these methods like I did in this PR, but this seems to work.

---

By the way, I think there is another, unrelated bug that `textDocument/diagnostic` doesn't get triggered after a `workspace/diagnostic/refresh` from the server. I will look into that later.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -224,13 +224,6 @@ class Manager(metaclass=ABCMeta):
         raise NotImplementedError()
 
     @abstractmethod
-    def sessions(self, view: sublime.View, capability: str | None = None) -> Generator[Session, None, None]:
-        """
-        Iterate over the sessions stored in this manager, applicable to the given view, with the given capability.
-        """
-        raise NotImplementedError()
-
-    @abstractmethod
     def get_session(self, config_name: str, file_path: str) -> Session | None:
         """
         Gets the session by name and file path.
@@ -720,6 +713,9 @@ class SessionBufferProtocol(Protocol):
     def get_semantic_tokens(self) -> list[Any]:
         ...
 
+    def evaluate_supported_custom_tokens(self, view: sublime.View) -> None:
+        ...
+
     def do_inlay_hints_async(self, view: sublime.View) -> None:
         ...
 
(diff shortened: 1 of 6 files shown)


plugin/core/sessions.py
rchl: Perhaps we should extract this marked code into a new method and call it from https://github.com/sublimelsp/LSP/blob/1554771471ea0c7077d910995d7a674889625637/plugin/session_view.py#L107-L108

Then it should happen after `didOpen` and we shouldn't need the `supress_requests` argument.

rchl: I suppose this could still result in code lens request being canceled and re-triggered but perhaps that's OK for now and in the future it could be made more robust by avoid requests based on last change_count?

jwortmann: > result in code lens request being canceled and re-triggered but perhaps that's OK for now and in the future it could be made more robust by avoid requests based on last change_count?

In that case we also need to add a `forced_update=True` parameter for code lens requests, which would be set when triggered by `workspace/codeLens/refresh`, otherwise we would run into the same problem as in #2699.

Also there is currently no tracking of pending code lens requests implemented, so if I move that block into `on_initialized` and just drop the `supress_requests` argument, then it would not cancel the previous request (currently only `codeLens/resolve` are cancelled on a new request). Frankly having such logic to prevent two open code lens requests for the same URI would be useful in general, but maybe we could do that refactoring at a later time.

rchl: feels a bit hacky (the supress_requests argument) but I don't have better solution for now

jwortmann: It should fix the bug (duplicate requests and before didOpen) at least. We can refactor and improve the code later if we find a better way.

PR #2700, opened 2025-12-02
Improve request cancellation handling

rchl:
Changes/fixes:
1. Rename `cancel_request` to `cancel_request_async`. Makes it clearer that cancellation has to be done on async thread.
2. On cancelling a request, mark `ActiveRequest` as canceled in SessionView. Fixes issue with cancelling same request twice from both `SessionView.on_before_remove` (on closing file) and `SessionBuffer._cancel_pending_requests_async` (on text change).
3. There was a `data.request.view.id() == self.view.id()` check when canceling requests in `SessionView`. I've removed it as `SessionView` should only have requests matching its view in the first place. Perhaps that wasn't the case in the past.
4. Ensure that progress reporting is stopped for requests that were canceled.
5. Ensure that error handler is called when canceling the request so that related request promise is resolved (with a `RequestCancelled` error).
6. Remove `ignore_response` argument as we are now calling error handler of the corresponding request immediately which fulfills the purpose that it was used for initially.

### DIFF
--- a/plugin/completion.py
+++ b/plugin/completion.py
@@ -255,9 +255,8 @@ class QueryCompletionsTask:
         self._cancel_pending_requests_async()
 
     def _cancel_pending_requests_async(self) -> None:
-        pending_requests = self._pending_completion_requests.copy()
-        self._pending_completion_requests.clear()
-        for request_id, weak_session in pending_requests.items():
+        # Iterate a copy of the dictionary since keys are popped on canceling.
+        for request_id, weak_session in self._pending_completion_requests.copy().items():
             if session := weak_session():
                 session.cancel_request_async(request_id)
 
--- a/plugin/core/active_request.py
+++ b/plugin/core/active_request.py
@@ -47,8 +47,7 @@ class ActiveRequest:
 
     def on_request_canceled_async(self) -> None:
         self.canceled = True
-        if self.progress:
-            self.progress = None
+        self.progress = None
 
     def _start_progress_reporter_async(
         self,


plugin/core/sessions.py
```diff
@@ -2480,10 +2480,11 @@ def send_request_task_2(self, request: Request) -> tuple[Promise, int]:
         request_id = self.send_request_async(request, resolver, lambda x: resolver(Error.from_lsp(x)))
         return (promise, request_id)
 
-    def cancel_request(self, request_id: int, ignore_response: bool = True) -> None:
+    def cancel_request_async(self, request_id: int, *, ignore_response: bool = True) -> None:
         self.send_notification(Notification("$/cancelRequest", {"id": request_id}))
-        if ignore_response and request_id in self._response_handlers:
-            request, _, _ = self._response_handlers[request_id]
+        request, _, _ = self._response_handlers[request_id]
```
jwortmann: This should better be a `self._response_handlers.get(request_id)`

The entry for the `request_id` gets pop'ed when the response arrives, so I think this line would be prone to a race condition. That is also the reason why I would keep it as part of the if-condition below.

rchl: I can re-add the check for request_id being present I guess. I've changed it with assumption that we should only cancel requests that are in progress and rather get notified with a crash if we don't do it right (which would likely mean there is a bug somewhere).

I don't think that a race condition should be possible if we assume that all modifications of `self._response_handlers` are done in the same (async) thread.

plugin/session_view.py
```diff
@@ -360,6 +361,9 @@ def on_request_started_async(self, request_id: int, request: Request) -> None:
     def on_request_finished_async(self, request_id: int) -> None:
         self._active_requests.pop(request_id, None)
 
+    def on_request_canceled_async(self, request_id: int) -> None:
+        self._active_requests.pop(request_id, None)
```
jwortmann: This doesn't really feel right to me. Here we pop the request, but the request is still open and when the response arrives, we try to pop the same request again. The response could either be an error response with "request cancelled" code, or the server is free to ignore the cancel and just send the regular response later. For example I see that there is some of the progress handling implemented in the ActiveRequest class, so what happens if progress is enabled for the particular request and the server continues to send progress notifications for it (not sure if this is even handled correctly at the moment)?

Another approach could be to add another field `cancelled` to the ActiveRequest class, set that to True when cancelled, and then skip those requests when triggered from `SessionView.on_before_remove`.

jwortmann: Since we only remove the response handlers when cacelling with `ignore_response=True` I assume that progress reports would still be handled as if no cancel happened (I haven't tested). But with such a `cancelled` field in ActiveRequest we could even check that in `ActiveRequest.update_progress_async` and then ignore those reports (it might need to be coupled to the `ignore_response` value too).

rchl: Sounds good. Will try that.

rchl: I'm not sure there is a need to couple this logic to `ignore_response`. The `ignore_response=True` is the default and the only case when we use `False` is when triggering completions. In that case we still want to get the callback to resolve the promise but we don't really care about that response anymore and wouldn't want the progress visible for it. In fact maybe it would be ideal to remove `ignore_response` argument and handle it differently for completions.

rchl: I've just realized something. In case we cancel a request, the original `on_result` and `on_error` handlers are never called which means that the corresponding promises are never resolved. We should probably call `on_error` immediately with a custom error code. It would fix the issue and also make `ignore_response` unnecessary.

rchl: I did a bit rushed fix since I'm going out but it should ensure that the promises are always resolved. It might in theory cause also some crashes if some handlers don't expect errors but that would be bug in the code.

plugin/session_view.py
```diff
@@ -360,6 +359,11 @@ def on_request_started_async(self, request_id: int, request: Request) -> None:
     def on_request_finished_async(self, request_id: int) -> None:
         self._active_requests.pop(request_id, None)
 
+    def on_request_canceled_async(self, request_id: int) -> None:
+        active_request = self._active_requests.get(request_id)
+        if active_request:
```
jwortmann:
```suggestion
        if active_request := self._active_requests.get(request_id):
```

plugin/symbols.py
```diff
@@ -366,17 +367,22 @@ def on_modified(self, text: str) -> None:
...
+        promises: list[Promise[list[sublime.ListInputItem]]] = [
+            session.send_request_task(Request.workspaceSymbol({"query": text}))
+            .then(partial(self._handle_response_async, session.config.name))
+            for session in self.command.sessions()
+        ]
+        Promise.all(promises).then(partial(self._on_all_responses, change_count))
 
     def _handle_response_async(
-        self, session_name: str, response: list[SymbolInformation] | list[WorkspaceSymbol] | None
+        self, session_name: str, response: list[SymbolInformation | WorkspaceSymbol] | Error | None
```
jwortmann: Not sure if it matters but the previous types `list[SymbolInformation] | list[WorkspaceSymbol]` instead of a possibly mixed list was what is defined in the specs. I don't know if it is on purpose, but afaik I have also seen mixed lists for other features in the specs typed as `(Foo | Bar)[]`, so maybe I would keep it as defined in the specs here.

rchl: reverted. this change was a byproduct of some experiments and not on purpse

plugin/core/active_request.py
```diff
@@ -44,6 +45,11 @@ def show() -> None:
 
             sublime.set_timeout_async(show, 200)
 
+    def on_request_canceled_async(self) -> None:
+        self.canceled = True
+        if self.progress:
```
jwortmann: Could save one line here by dropping this condition (run next line unconditionally).

plugin/completion.py
```diff
@@ -255,10 +255,11 @@ def cancel_async(self) -> None:
         self._cancel_pending_requests_async()
 
     def _cancel_pending_requests_async(self) -> None:
-        for request_id, weak_session in self._pending_completion_requests.items():
-            if session := weak_session():
-                session.cancel_request(request_id, False)
+        pending_requests = self._pending_completion_requests.copy()
         self._pending_completion_requests.clear()
```
jwortmann: I think this line is also not necessary anymore, because the pending requests should all be popped when the response handler is called (with an error response from cancel_request_async). Then you could even save another line by putting `self._pending_completion_requests.copy().items()` directly in the for-loop.

Issue #2697, opened 2025-11-30
Crash on view closing during applying of semantic tokens


rchl:
**Describe the bug**

It's possible to get an error like below. It reproduces fairly often when testing renaming of directories (with code from https://github.com/sublimelsp/LSP/pull/2498) when LSP opens new files, edits them and immediately closes.

```
Error handling None
Traceback (most recent call last):
  File "/Users/rafal/Library/Application Support/Sublime Text/Packages/LSP/plugin/core/sessions.py", line 2564, in on_payload
    handler(result)
    ~~~~~~~^^^^^^^^
  File "/Users/rafal/Library/Application Support/Sublime Text/Packages/LSP/plugin/session_buffer.py", line 696, in _on_semantic_tokens_viewport_async
    self._on_semantic_tokens_async(response)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/Users/rafal/Library/Application Support/Sublime Text/Packages/LSP/plugin/session_buffer.py", line 693, in _on_semantic_tokens_async
    self._draw_semantic_tokens_async()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/rafal/Library/Application Support/Sublime Text/Packages/LSP/plugin/session_buffer.py", line 754, in _draw_semantic_tokens_async
    if token_general_style["source_line"] != token_type_style["source_line"] or \
       ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
KeyError: 'source_line'
```

The problem is that `View` can become invalid at any point when running code in an async thread so the only reliable way to fix it is to never use any View API that might fail in async thread.

Moving all this logic to main thread would fix it but of course it's not ideal if the logic is expensive to run.

jwortmann:
> The problem is that `View` can become invalid at any point when running code in an async thread so the only reliable way to fix it is to never use any View API that might fail in async thread.
> 
> Moving all this logic to main thread would fix it but of course it's not ideal if the logic is expensive to run.

Another option would be to put all code that runs on the async thread and uses view API methods into a `try` block. But I assume this would apply to most features in LSP, not only semantic tokens. Even something like
```python
if view.is_valid():
    view.style_for_scope(...)
```
might fail if the view is still alive during the evaluation of the condition, but not anymore at the API call on next line.

[...]

PR #2704, opened 2025-12-03
Add caching for color scheme support of custom semantic token types

jwortmann:
As discussed in #2697. Not sure if it fixes that issue because there are some more view API calls like https://github.com/sublimelsp/LSP/blob/ae9e7a5ed725fe60e0ce4748b6f468611854916d/plugin/session_buffer.py#L737-L739
which could fail if the view got closed during the execution.

There might be more code like at https://github.com/sublimelsp/LSP/blob/ae9e7a5ed725fe60e0ce4748b6f468611854916d/plugin/session_buffer.py#L659-L660

that could potentially be cached to reduce ST API calls.

### DIFF
--- a/docs/src/customization.md
+++ b/docs/src/customization.md
@@ -103,7 +103,7 @@ If you use a custom color scheme, select `UI: Customize Color Scheme` from the C
 Furthermore, it is possible to adjust the colors for semantic tokens by applying a foreground color to the individual token types:
 
 | [Semantic Token Type](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#semanticTokenTypes) | scope | fallback scope |
-| ----- | ------------------ |
+| --- | --- | --- |
 | namespace | `meta.semantic-token.namespace` | `variable.other.namespace` |
 | type | `meta.semantic-token.type` | `storage.type` |
 | class | `meta.semantic-token.class` | `storage.type.class` |


docs/src/customization.md
```diff
@@ -79,6 +79,9 @@ The following tables give an overview of the scope names used by LSP.
 !!! note
     Semantic highlighting is disabled by default. To enable it, set `"semantic_highlighting": true` in your LSP user settings.
 
+!!! warning
+    There are several known limitations when semantic highlighting is used. In particular, there are visible artifacts on lines with semantic highlighting if the `"highlight_line"` setting is enabled, and italic and bold font styles are suppressed for regions with semantic highlighting.
```
rchl, line 83: You could mention it or not but also sometimes semantic tokens can make diagnostic highlights invisible. I know we have code that pre-initializes regions which should prevent that but it's still not always working as intended and I'm not sure we can fix it.

jwortmann:
I forgot to mention that what I do in this PR, to cache the supported custom token types per SessionBuffer is not really correct if there are cloned views and you would manually assign another color scheme to only one of the clones. But I think handling the caching in the SessionViews would add some more complexity to the code, because we would also need to move the `_draw_semantic_tokens_async` method or at least parts of that into the SessionView class. And also that would be such a niche use case that it shouldn't really matter.

docs/src/customization.md
jwortmann: Okay I didn't know that. But it's probably not so important to mention it here in the docs.

PR #2708, opened 2025-12-05
Make apply_text_edits return a Promise

rchl:
Make `apply_text_edits` return a promise so that it can be awaited in case we want to ensure that LSP notifications are processed before any further actions.

- It makes my last #2707 fix unnecessary since now the extra async timeout is part of the `apply_text_edits` itself.
- It fixes issues that I've observed when testing #2498 where on renaming files or directories, it would open new files, edit them and close immediately which would not allow didChange notifications to get processed which caused sync issues in the server.
- Refactoring in `formatting.py` is not strictly necessary but it cleans up code that no longer needs to call `set_timeout_async` explicitly

### DIFF
--- a/plugin/core/edit.py
+++ b/plugin/core/edit.py
@@ -57,10 +57,10 @@ def apply_text_edits(
     required_view_version: int | None = None
 ) -> Promise[sublime.View | None]:
     if not edits:
-        return Promise.resolve(None)
+        return Promise.resolve(view)
     if not view.is_valid():
         print('LSP: ignoring edits due to view not being open')
-        return Promise.resolve(view)
+        return Promise.resolve(None)
     view.run_command(
         'lsp_apply_document_edit',
         {


plugin/formatting.py
```diff
@@ -197,13 +195,17 @@ def run(self, edit: sublime.Edit, event: dict | None = None) -> None:
...
         elif self.view.has_non_empty_selection_region():
             if session := self.best_session('documentRangeFormattingProvider.rangesSupport'):
                 request = text_document_ranges_formatting(self.view)
-                session.send_request(
-                    request, lambda response: apply_text_edits(self.view, response, label="Format Selection"))
+                session.send_request_task(request).then(self._handle_response_async)
+
+    def _handle_response_async(self, response: FormatResponse) -> None:
+        if isinstance(response, Error):
+            sublime.status_message(f'Failed formatting - {response}')
```
predragnikolic:
```suggestion
            sublime.status_message(Formatting error: {response}')
```

This message format was the most common:
sublime.status_message(f'Hover error: {error_messages}')
sublime.status_message(f'Completion error: {error_messages}')

plugin/core/sessions.py
```diff
@@ -1995,18 +1995,25 @@ def _get_view_state_actions(self, uri: DocumentUri, auto_save: str) -> ViewState
...
-                self.window.focus_view(view)
-            self.window.run_command('close')
+            # Allow async thread to process save notifications before method returns.
+            promise = Promise(lambda resolve: sublime.set_timeout_async(lambda: resolve(None)))
+
+        def handle_close() -> None:
+            if should_close and not view.is_dirty():
+                if view != self.window.active_view():
+                    self.window.focus_view(view)
+                self.window.run_command('close')
```
predragnikolic:
```suggestion
                    view.close()
```

predragnikolic (approved):

plugin/core/sessions.py
rchl: Let's hope there was no specific reason why it was made like this (it's an old code).

plugin/core/sessions.py
```diff
@@ -1995,18 +1995,24 @@ def _get_view_state_actions(self, uri: DocumentUri, auto_save: str) -> ViewState
...
-            if view != self.window.active_view():
-                self.window.focus_view(view)
-            self.window.run_command('close')
+            # Allow async thread to process save notifications before method returns.
+            promise = Promise(lambda resolve: sublime.set_timeout_async(lambda: resolve(None)))
+
+        def handle_close() -> None:
+            if should_close and not view.is_dirty():
+                if view != self.window.active_view():
+                    view.close()
```
rchl:
```suggestion
                view.close()
```

jwortmann:
I wonder would it be better to return a `Promise[sublime.View]`, i.e. pass the view further in the promise chain? Then you don't need things like the `.then(lambda _: view)`.
Or return `Promise[sublime.View | None]` and remove the `_apply_text_edits` helper method while putting the `view.is_valid()` check directly into `apply_text_edits`.

rchl:
> I wonder would it be better to return a `Promise[sublime.View]`, i.e. pass the view further in the promise chain? Then you don't need things like the `.then(lambda _: view)`. Or return `Promise[sublime.View | None]` and remove the `_apply_text_edits` helper method while putting the `view.is_valid()` check directly into `apply_text_edits`.

Good idea.

Not sure if my solution is as clean as it could be but since I wanted to keep the `if view` check and corresponding print (and didn't want to change `apply_text_edits` to take optional view), I ended up with a nested function which are not my favorite from readability standpoint. Didn't want to make it a method since `Session` is a big class and it feels like I shouldn't pollute it with some helper functions like that.

plugin/core/edit.py
```diff
@@ -55,9 +55,12 @@ def apply_text_edits(
...
     process_placeholders: bool | None = False,
     required_view_version: int | None = None
-) -> Promise[None]:
+) -> Promise[sublime.View]:
     if not edits:
-        return Promise.resolve(None)
+        return Promise.resolve(view)
+    if not view.is_valid():
+        print('LSP: ignoring edits due to view not being open')
+        return Promise.resolve(view)
```
jwortmann: Let's use a `debug` instead of `print` here, okay?

Also what I was thinking about would be to return `Promise.resolve(None)` on this code path, because the view was closed anyway.

plugin/core/edit.py
```diff
@@ -54,9 +55,12 @@ def apply_text_edits(
...
     process_placeholders: bool | None = False,
     required_view_version: int | None = None
-) -> None:
+) -> Promise[sublime.View]:
     if not edits:
-        return
+        return Promise.resolve(view)
+    if not view.is_valid():
+        print('LSP: ignoring edits due to view not being open')
+        return Promise.resolve(view)
```
rchl: I suppose it's a bit weird to ignore edits but still return a view for something down the chain. But returning `None` could make usage annoying on the other hand.

plugin/core/edit.py
rchl: We might have a bit different opinions about debug vs. print in this case...

Normal people don't run with debug logs enabled. Heck, even I don't do that since it causes logs fatigue. That means that if anything triggers this case, pretty much no one will know about it.

Of course this can trigger both in legitimate cases due to a bug somewhere or after the view was closed on purpose. The former would be good to know, the latter we don't really care about. So there is argument for both ways.

As for returning `Promise.resolve(None)`, as I've alluded to in my other comment, I'm just not sure if it's worth dealing with if-checking the result in call sites...

rchl: OK, changed to `Promise.resolve(None)` since there is only one call site now that cares about the return value.

jwortmann (approved):

PR #2711, opened 2025-12-08
revert limiting semantic highlight to one session

rchl:
Revert semantic-highlighting specific part of #2688 since it breaks highlighting in Vue files.

### DIFF
--- a/plugin/semantic_highlighting.py
+++ b/plugin/semantic_highlighting.py
@@ -88,6 +88,7 @@ class LspShowScopeNameCommand(LspTextCommand):
                         if token.region.contains(point) and point < token.region.end():
                             token_modifiers = ', '.join(token.modifiers) if token.modifiers else '-'
                             info.append((token.type, token_modifiers, session.config.name))
+                            break
                     break
         return info
 


jwortmann:
If multiple sessions for semantic highlighting should really be allowed, then the show scope name command should probably check all `self.sessions(self.capability)` until it finds one that provides a semantic token for the given `point`, instead of using only one session in
https://github.com/sublimelsp/LSP/blob/9bfb68da38cbb71fa932924536395f1f8c3f676a/plugin/semantic_highlighting.py#L61

rchl:
I've changed it to show tokens from all servers:

Screenshot 2025-12-09 at 14 58 21

plugin/semantic_highlighting.py
```diff
@@ -97,6 +80,10 @@ def _render_with_plain_string_stackframes(
                 backtrace += '\n'
             backtrace += f'<div>{nums}{ctx}</div>'
 
+        semantic_tokens_html = ''
+        for info in semantic_info:
+            semantic_tokens_html += f'<div>Type: {info[0]}, Modifiers: {info[1]} <a>{info[2]}</a></div>'
```
jwortmann: I wonder why is there a link (`a` tag) for the session name, and without any link destination? I think it would be better to change this maybe using a dimmed color instead, like the syntax name which is displayed in the ST references popup.

rchl: Updated:

Screenshot 2025-12-10 at 20 24 10

While at it consolidated duplicated code.

plugin/semantic_highlighting.py
```diff
@@ -38,183 +63,120 @@ def run(self, _: sublime.Edit) -> None:
...
+        info: SemanticTokensInfo = []
+        for session in self.sessions('semanticTokensProvider'):
             for sv in session.session_views_async():
                 if self.view == sv.view:
-                    session_buffer = sv.session_buffer
+                    for token in sv.session_buffer.get_semantic_tokens():
+                        if token.region.contains(point) and point < token.region.end():
+                            token_modifiers = ', '.join(token.modifiers) if token.modifiers else '-'
+                            info.append((token.type, token_modifiers, session.config.name))
                     break
```
jwortmann, line 92: It looks like this break is on the wrong indentation level now. It's supposed to break as soon as token for the given point is found (inner `if`).
We don't have https://github.com/sublimelsp/LSP/blob/9bfb68da38cbb71fa932924536395f1f8c3f676a/plugin/core/sessions.py#L479
so there should never be more than one token applicable for the point.

jwortmann: Actually with this refactor it would be nice to break from both loops then. But afaik that's not directly possible in Python, so either this needs a bit more refactoring, or just break from the inner loop (there are likely many more tokens than SessionViews, and the outer if-check for the view should be cheap compared to the inner if-check).

rchl: added a break

jwortmann: OK that was simple 😆

PR #2713, opened 2025-12-10
Refactor plugin API methods for applicable views

### LINKED ISSUE #2503: `should_ignore` needs a counterpart `should_unignore`
Previously, LSP-Copilot had an open issue for how do we prevent a View/File from being managed by Github Copilot with disabling it on the entire syntax.

@jwortmann was kind enough to think this through and add the below function.

https://github.com/sublimelsp/LSP/blob/293f4a4340cca5ab1ad065643e4f20d9b270b2b1/plugin/core/sessions.py#L956-L965

However, now when we want to ignore a file, if that file is to ever be add back to the session, it requires asking the user to close and re-open the file or doing the below hack.

https://github.com/TerminalFi/LSP-copilot/blob/095fd059d1c6f7faaaedc75dc828ed83476d29f1/plugin/listeners.py#L61-L69

What this leads to is impacting other LSP sessions / plugins.  See https://discord.com/channels/280102180189634562/1261340378133434439/1261340581779603497 for more details

### Suggestion

Add another class func that can be used to trigger a view to be added to the LSP-* session

jwortmann:
Combine `selector` and `should_ignore` plugin API methods into a new `is_applicable`.

The previous methods are intended to still work with this PR, i.e. it should be backwards compatible (todo: still needs testing to verify).

This also adds a new TextCommand which can be used via
```python
view.run_command('lsp_check_applicable', {'session_name': 'LSP-foo'})
```
to retrigger the check for the given view and session. This is implemented as a command (instead of additional API method) to allow it to be called even if the session is not yet running (this needs it to be called e.g. from an independent listener implementation of course).

Closes #2503

### DIFF
--- a/plugin/core/registry.py
+++ b/plugin/core/registry.py
@@ -234,10 +234,14 @@ class LspCheckApplicableCommand(sublime_plugin.TextCommand):
         sublime.set_timeout_async(lambda: self._run_async(session_name))
 
     def _run_async(self, session_name: str) -> None:
-        if (wm := windows.lookup(self.view.window())) and (listener := windows.listener_for_view(self.view)):
+        if wm := windows.lookup(self.view.window()):
             config = wm.get_config_manager().get_config(session_name)
             if not config:
                 debug(f'Configuration with name {session_name} does not exist')
+                return
+            listener = windows.listener_for_view(self.view)
+            if not listener:
+                debug(f'No listener for view {self.view}')
                 return
             scheme, _ = parse_uri(uri_from_view(self.view))
             is_applicable = config.match_view(self.view, scheme)


plugin/core/configurations.py
```diff
@@ -38,6 +38,9 @@ def __init__(self, window: sublime.Window, global_configs: dict[str, ClientConfi
     def add_change_listener(self, listener: WindowConfigChangeListener) -> None:
         self._change_listeners.add(listener)
 
+    def get_config(self, name: str) -> ClientConfig | None:
+        return self.all.get(name)
```
rchl: We do typically use `config_name` nomenclature in other methods.

```suggestion
    def get_config(self, config_name: str) -> ClientConfig | None:
        return self.all.get(config_name)
```

jwortmann:
Language servers can dynamically register features only to certain files using a DocumentSelector, which is a list of [DocumentFilter](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#documentFilter), which in turn is a combination of language ID, scheme and glob pattern. We use that DocumentSelector to assing the server capabilities to the SessionBuffers, but the criterium which views the language server runs on is currently still limited to the `schemes` and `selector` setting from the config (settings file).

The new `is_applicable` method should make it possible/easier to match the files supported by the language server. For example a server could provide some features for a `Project.toml` (or `Cargo.toml`, etc.) configuration file, but should not run on any other TOML files. There might be awkward ways to achieve that using the previous API methods or by creating a special syntax only for such filenames and using a special scope name. But it should be much easier now using something like
```python
@classmethod
def is_applicable(cls, view: sublime.View, config: ClientConfig) -> bool:
    if (filename := view.file_name()) and os.path.basename(filename) == 'Project.toml':
        return True
    return super().is_applicable(view, config)
```

rchl:
I haven't given it too much thought but would it make sense to also (or only) provide a way to statically define those conditions in settings (to match the current approach)? It might not be as flexible as API-based solution but it might be enough if one can specify equivalent of `DocumentFilter` that way.

Something like:

```
document_filter: list[DocumentFilter]
```

but without `language`

The advantage of that? A clear visibility of the conditions triggering the server. 

EDIT: Copilot wants to read ignore patterns from `.copilotignore` file so that should also work for it if it would read those and store in config settings before starting server. And it would make its implementation simpler as it wouldn't need to match globs itself.

jwortmann:
I think providing an API method that can only modify the static config would be a step backwards because then there wouldn't be any way to react to changes in configuration files when the server is already running. The `is_applicable` just combines the existing methods while giving more flexibility (for instance the previous `should_ignore` could only be used as an additional condition when the static config already matches, but not to allow more files to be applicable) and with a better name.

We could think of adding something like the `DocumentSelector` to the static config independently of that API method and at a later time. Then we just need to modify the default implementation of `is_applicable` and the corresponding code for other configs that are not backed by a plugin. If I had started creating the LSP package, maybe I would have done it that way to keep the implementation close to what is defined in the LSP specs and to how language servers work. But in that case I had probably also included the `language` ID instead of our `selector`, and then map the language IDs to scope selectors internally. But I think that ship has sailed long ago. If we add a new config option `document_selector` now, to allow a list of language/scheme/pattern combinations, it might be confusing how it should behave in combination with the existing `selector` and `schemes` options though (should those simply be ignored if `document_selector` is defined?).

plugin/core/registry.py
```diff
@@ -226,6 +228,37 @@ def run_async() -> None:
...
+                if is_applicable and not session_view:
+                    listener.on_session_initialized_async(session)
+                elif not is_applicable and session_view:
+                    session.shutdown_session_view_async(session_view)
+            elif is_applicable:
+                wm.start_async(config, self.view)
+                if wm._new_session:
+                    wm._sessions.add(wm._new_session)
+                    listener.on_session_initialized_async(wm._new_session)
+                    wm._new_session = None
```
jwortmann, line 259: This part is pretty ugly, but the public `WindowManager.start_async` method is not actually sufficient to start a new session. That method sets an internal variable at https://github.com/sublimelsp/LSP/blob/d5aefde238d51ed3d6b3562b14380a6b5e6fe784/plugin/core/windows.py#L280 and it expects some more code to be run after that, in order to properly handle the start of a new session.

I won't touch the recursive https://github.com/sublimelsp/LSP/blob/d5aefde238d51ed3d6b3562b14380a6b5e6fe784/plugin/core/windows.py#L154 method because it is a riddle to me how that works...

Starting a new session when triggered from the `lsp_check_applicable` TextCommand seems to work correctly now though.

PR #2714, opened 2025-12-11
support dict variant for priority_selector

### LINKED ISSUE #2526: Multiple session and request handling
The new version of Volar (Vue Language Features) is expected to run in Vue files alongside a typescript server (LSP-typescript, for example). This seems to create issues with handling of some requests since some requests are supposed to go to LSP-volar and some to LSP-typescript. For example within a `script` tag like:

```vue
<script lang="ts">
export default {
  methods: {
    getSomething(key: string): HTMLElement {
      // ...
    }
  }
}
</script>
```

- "expand selection" request should go to LSP-volar
- "go to definition" should go to LSP-typescript

We have `prioritity_selector` which can force all requests go to one or the other but in this case it doesn't help because we need different behavior for different requests and scopes are not sufficient for determining which one to trigger.

Observing VSCode, it seems like triggering "go to definition" triggers requests both to Volar (which returns an empty list) and its internal Typescript server (note that it's not technically LSP-based in there).

But then I also see some issues with expand selection in VSCode that I don't see in ST. For example here it incorrectly expands selection to just `set` or `Timeout` in `setTimeout`:

https://github.com/user-attachments/assets/14374961-7651-4bee-9662-01ecaac96bbc

So not sure what is the best way for handling this issue.

rchl:
Want to get opinions about this solution.

Support `priority_selector` in a dict form where the keys are names of providers and values are selectors. There is also a special `*` key which acts as a fallback when no specific provider matches.

This allows to fix handling for `renameProvider` in LSP-vue where in `text.html.vue` scope (template block) LSP-vue should handle renaming while in `text.html.vue source.js|ts` LSP-typescript should handle renaming. The setting for this would look something like:

```jsonc
	"priority_selector": {
		"*": "text.html.vue, text.html.vue source.js, text.html.vue source.ts",
		"renameProvider": "text.html.vue",
	},
```

I'm making this a draft to get opinions before I do final touches and to also give me time to check handling for other providers in LSP-vue (want to make sure that all problematic cases can be fixed with this).

Fixes #2526


predragnikolic:
Rafal, I forgot the reason on why we need to redirect some request to vue and some to typescript. Can you maybe point me to some comment left by johnsoncodehk where he explained this? I know I read it somewhere but I cannot find it.

All I discovered this:
https://github.com/vuejs/language-tools/discussions/5456

> How do I migrate to version 3?
> Migrating to version 3 is quite simple, create a client handler for vue language server, when receiving the request, forward it to typescript language server and finally send the payload from typescript language server back to vue language server. We have an example for neovim users [here](https://github.com/vuejs/language-tools/wiki/Neovim).
>
> FAQ
> What's the request and response name I should handle/send to vue language server?
The request and response names are not configurable:
>
>Request(receive from vue lsp): tsserver/request
>Response(send to vue lsp): tsserver/response

rchl:
Those specific requests that are problematic are not related to `tsserver/request` and `tsserver/response`. Those go directly to respective servers and are not redirected.

I don't know if I can provide much of a answer for why we need to do this. The fact is that some requests need to be handled by different servers depending on where in the document we are. Perhaps this is something that could be solved by the vue language server. Ideally all requests would go to vue server and then vue server would decide whether to redirect it to typescript server or respond directly. I suppose I should ask that question in their repo.

For the record, VSCode seems to ask one server and if server responds with empty response then it will try the next one.

rchl:
`inlayHintProvider` case I can't fix with `priority_selector` since those are requested for whole document. Both LSP-vue and LSP-typescript provide different inlay hints:

Screenshot 2025-12-12 at 23 24 52

So here limiting inlay hints to one session does hurt the functionality...

PR #2716, opened 2025-12-16
Add prompt on triggering workspace edits during rename file

predragnikolic:
Follow up PR for #2498 

This PR introduces a prompt when renaming files:

https://github.com/user-attachments/assets/c9b72bd8-3547-498c-b76a-a321e12b2eb2

The prompt logic is taken from Rename Symbol, but extracted in a function  
`prompt_for_workspace_edits` that is now shared between `LSP: Rename Symbol` and `LSP: Rename File`. (NOTE, most of the logic was copy/pasted from the `plugin/rename.py` to `plugin/edit.py`)

Initially I did implement everything as one `LspRenamePathCommand`, but that ended in code that was hard to folow, so I split it in two commands that will execute in the following order:

1. `LspPromptRenamePathCommand`:
- responsible to send willRename request
- responsible to show the prompt if the server returned workspace edit
- responsible to invoke `LspRenamePathCommand`

2. `LspRenamePathCommand`:
- responsible to rename the path
- responsible to send didRename notification

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,1 +1,1 @@
-zensical==0.0.15
+zensical==0.0.17
--- a/tests/test_single_document.py
+++ b/tests/test_single_document.py
@@ -391,7 +391,7 @@ class SingleDocumentTestCase3(TextDocumentTestCase):
             ]
         })
         yield from self.await_message('textDocument/didChange')
-        # yield from self.await_message('textDocument/didSave')  # TODO why is this not sent?
+        yield from self.await_message('textDocument/didSave')
         yield from self.await_message('textDocument/didClose')
 
 


plugin/edit.py
```diff
@@ -140,3 +150,190 @@ def _sort_by_application_order(changes: Iterable[TextEditTuple]) -> list[TextEdi
...
+    label: str,
+    total_changes: int,
+    file_count: int,
+    on_accept_command: tuple[str, dict] | None = None
+) -> None:
+    def _get_relative_path(wm: WindowManager, file_path: str) -> str:
+        base_dir = wm.get_project_path(file_path)
+        return os.path.relpath(file_path, base_dir) if base_dir else file_path
+
+    wm = windows.lookup(session.window)
```
predragnikolic, line 220: The previous code had access to self.view

https://github.com/sublimelsp/LSP/pull/2716/changes#diff-1624146a688a4081a8a84dab444a153add3a434c7617c30b019ab7039110bb9aL246

I didn't want to pass in a view to access the window.
so I passed in the session as an argument and grabbed the window from the session

plugin/edit.py
```diff
@@ -140,3 +150,190 @@ def _sort_by_application_order(changes: Iterable[TextEditTuple]) -> list[TextEdi
...
+    panel = pm.ensure_workspace_edit_panel()
+    if not panel:
+        return
+    to_render: list[str] = []
+    reference_document: list[str] = []
+    header_lines = f"{total_changes} changes across {file_count} files - {label}\n"
+    to_render.append(header_lines)
+    reference_document.append(header_lines)
+    ROWCOL_PREFIX = " {:>4}:{:<4} {}"
+    base_dir = None
```
predragnikolic: the way how `base_dir` is set is a bit different in this PR.

Previously, `base_dir` was set by grabbing using the view, see:
https://github.com/sublimelsp/LSP/pull/2716/changes#diff-1624146a688a4081a8a84dab444a153add3a434c7617c30b019ab7039110bb9aL287

Now, base_dir is set when iterating through `changes_per_uri` as there we have a file path. (but I don't like the if in the for loop, I will see if I can do it a bit differently. If you have suggestions, please leave them)

jwortmann:
I would prefer to keep it as a single command if that is possible. But I haven't looked closely at the code yet though.

I like the refactoring/reuse of the rename panel to make it more general for applying any WorkspaceEdits. I had the idea of a "Refactor preview" feature before, which could also be used for example for code actions (perhaps with <kbd>Ctrl</kbd> + click instead of a regular mouse click), but I never ended up implementing that.

I created a PR #2717 to count the changes and changed files in a the WorkspaceEdit. This could be an alternative (or in addition) to this preview panel, to give at least some feedback to users after applying the rename path command. Something like
```python
.then(lambda change_count: window.status_message(f'Updated {change_count[0]} references to {new_path} in {change_count[1]} files')
```

Would a new setting whether to show the confirm/preview dialog perhaps be useful? I remember some user request in the past to suppress the confirmation dialog when doing a rename, and I kinda agree that limiting the number of dialog windows to a minimum is in general better UX. On the other hand the ST undo functionality only works per file and not for the entire WorkspaceEdit, so it might be good to force the confirmation dialog to prevent accidental changes that affect many files.

plugin/edit.py
```diff
@@ -2,17 +2,27 @@
...
+import os
 import re
 import sublime
 import sublime_plugin
+if TYPE_CHECKING:
+    from .core.sessions import Session
 
-
-TextEditTuple = Tuple[Tuple[int, int], Tuple[int, int], str]
+TextEditTuple = tuple[tuple[int, int], tuple[int, int], str]
```
rchl: This change will not work in Python 3.8 host. Using builtin types in "standalone" types like here is supported from 3.12, I believe.

plugin/edit.py
```diff
@@ -140,3 +150,190 @@ def _sort_by_application_order(changes: Iterable[TextEditTuple]) -> list[TextEdi
     # we use the index in the array as the key.
 
     return list(sorted(changes, key=operator.itemgetter(0)))
+
+
+def prompt_for_workspace_edits(
+    session: Session, response: WorkspaceEdit, on_accept_command: tuple[str, dict] | None = None,
+    label: str = "Apply workspace edits", yes_button=""
```
rchl: Probably no need to make `yes_button` customizable if it's always the same currently.

plugin/rename_file.py
```diff
@@ -69,26 +70,27 @@ def run(self, new_name: str, paths: list[str] | None = None) -> None:
...
             "newUri": filename_to_uri(new_path),
             "oldUri": filename_to_uri(old_path)
         }
+        rename_command = ("lsp_rename_path", {"old_path": old_path, "new_path": new_path})
         Promise.all(list(self.create_will_rename_requests_async(file_rename))) \
-            .then(lambda responses: self.handle_rename_async(responses)) \
-            .then(lambda _: self.rename_path(old_path, new_path)) \
-            .then(lambda success: self.notify_did_rename(file_rename) if success else None)
+            .then(lambda responses: self.handle_rename_async(responses, label, rename_command)) \
+            .then(lambda did_accept: self.window.run_command(*rename_command) if did_accept else None)
```
rchl: Also not a fan of this approach, it's kinda hard to follow.

My idea would probably be to make `prompt_for_workspace_edits` do only what its name suggests - ask. That way the flow would be clear and always the same - regardless if applying changes immediately or after showing the panel.

I would implement that by making it return either the resolved promise immediately, if edits are to be applied immediately, or storing it somewhere in a global and resolving when user performs given action from the edits panel. Clicking the buttons in the edits panel would invoke some new command dedicated to resolving those global promises by ID. That last part would be a bit unconventional but I think it should work nicely and make it clearer and easier to use by hiding that async nature behind a promise.

predragnikolic:
will mark as ready for review when I tweak the PR
(right now I see some things off that I done tonight :) but will try to correct them - just don't waste your time at the PR now)

plugin/core/sessions.py
```diff
@@ -1980,8 +1980,13 @@ def handle_view(
         return Promise.all(promises) \
             .then(lambda _: self._set_selected_sheets(selected_sheets)) \
             .then(lambda _: self._set_focused_sheet(active_sheet)) \
+            .then(lambda _: self.show_summary(summary)) \
```
predragnikolic: With this logic, each time workspave edits are applied
show_summary will show the message in the window status bar.

I needed to use a set_timeout of 300ms `sublime.set_timeout(lambda: self.window.status_message(message), 300)`, else the summary message will be overridden by the following text "Detect indentation...":

Image

plugin/rename_file.py
```diff
@@ -74,21 +75,29 @@ def run(self, new_name: str, paths: list[str] | None = None) -> None:
...
     @staticmethod
     def is_case_change(path_a: str, path_b: str) -> bool:
         return path_a.lower() == path_b.lower() and Path(path_a).stat().st_ino == Path(path_b).stat().st_ino
 
-    def run_async(self, old_path: str, new_path: str) -> None:
-        file_rename: FileRename = {
-            "newUri": filename_to_uri(new_path),
-            "oldUri": filename_to_uri(old_path)
-        }
+    def prompt_rename_async(self, file_rename: FileRename, label: str, rename_command: tuple[str, dict]) -> None:
```
predragnikolic: I refactored the code to only have one LspRenamePath command (instead of the previous approach LspRenamePath and LspPromptRenamePathCommand)

The command has two code paths depending on the bool value for the `prompt_workspace_edits` arg.

When `prompt_workspace_edits` is true,
I renamed `run_async` to `prompt_rename_async` to make it more obvious that in this code path 
it is expected to see the `prompt_for_workspace_edits` logic

predragnikolic:
> [jwortmann](https://github.com/jwortmann)
commented
[last month](https://github.com/sublimelsp/LSP/pull/2716#issuecomment-3665335270)
I would prefer to keep it as a single command if that is possible

Implemented it as a single command [6ccd0a9](https://github.com/sublimelsp/LSP/pull/2716/commits/6ccd0a9ea544f32e5a117eb5f8c2bbc5fb13542a)

predragnikolic:
> Would a new setting whether to show the confirm/preview dialog perhaps be useful?

We can revisit the PR if necessary -> https://github.com/sublimelsp/LSP/pull/1923 (if yes, we can do it in a separate PR after this)

plugin/rename_file.py
predragnikolic: I did change the logic to make the flow more obvious, 
albeit not as suggested by your comment, 
but I did want to do minimal code changes.

predragnikolic:
Pardon for the delay on my side

jwortmann:
Do you want to merge the other PR first and then retarget this one to the main branch?

Just to be clear, I think I'm in favor of the changes here (just haven't reviewed the code yet), but it should be okay to merge the PRs with improvements step by step and that would probably make it easier to follow and to review.

predragnikolic:
Ok, I will merge https://github.com/sublimelsp/LSP/pull/2498 soon
And retarget this PR to main

plugin/rename_file.py
rchl: Can I introduce a commit that implements my idea? It could be reverted or not based on your and @jwortmann's opinion.

plugin/core/sessions.py
rchl: This is something that would warrant a comment in the code since it's not obvious and nobody will remember your comment after some time :)

plugin/core/sessions.py
```diff
@@ -1980,8 +1980,13 @@ def handle_view(
...
             .then(lambda _: summary)
 
+    def show_summary(self, summary: WorkspaceEditSummary):
```
rchl: Missing return type annotation.

plugin/core/sessions.py
rchl: BTW. Feels weird that ST sets window-wide status for something that relates to currently open view.

predragnikolic: > BTW. Feels weird that ST sets window-wide status for something that relates to currently open view.

Maybe I miss read your comment,
I don't think this is correct "relates to currently open view.", 
The `promises` variable holds many edited views, 
so it doesn't feel that strange to use the window status message.

rchl: I meant the "Detect Indentation" status message.

plugin/rename_file.py
predragnikolic: Yes, please

rchl: With c843f755861c7136c490d0893f44cc8f49f922de `prompt_for_workspace_edits`'s only purpose is to get the answer whether the edits should be applied or not - it doesn't apply the changes itself anymore.

I think this flow is less magic and easier to follow.

tests/testfile2.txt
```diff
@@ -0,0 +1 @@
+TEST
```
predragnikolic: 🚮

predragnikolic: Probably committed by accident

plugin/edit.py
```diff
@@ -140,3 +153,182 @@ def _sort_by_application_order(changes: Iterable[TextEditTuple]) -> list[TextEdi
...
+    if choice == sublime.DialogResult.YES:
+        return Promise.resolve(True)
+    if choice == sublime.DialogResult.NO:
+        promise, resolve = Promise[bool].packaged_task()
+        _render_workspace_edit_panel(session, changes, label, total_changes, file_count, resolve)
+        return promise
+    return Promise.resolve(False)
+
+
+BUTTONS_TEMPLATE = """
```
rchl: Typically such constants would be at the top of the file.

plugin/core/sessions.py
```diff
@@ -1983,8 +1983,14 @@ def handle_view(
         return Promise.all(promises) \
             .then(lambda _: self._set_selected_sheets(selected_sheets)) \
             .then(lambda _: self._set_focused_sheet(active_sheet)) \
+            .then(lambda _: self.show_summary(summary)) \
```
jwortmann: My intention was that the summary can be used selectively in the function that calls `apply_workspace_edit_async`, which is why it is the return value of that method.

If you put it here, it will always print the summary to the status bar whenever a WorkspaceEdit gets applied, which happens not only for renaming symbols or files, but also for example when running code actions (possibly), or could be triggered by code lenses via `workspace/applyEdit`, etc.

I think for now it would be better to only use it for the rename and rename path commands.
Maybe we can consider it for code actions too, but I would probably not do it for `workspace/applyEdit`.

plugin/edit.py
```diff
@@ -140,3 +186,148 @@ def _sort_by_application_order(changes: Iterable[TextEditTuple]) -> list[TextEdi
...
+            if utf16_len > col:  # If col is in the middle of a character (emoji), don't advance to the next code point
+                idx -= 1
+            break
+        utf16_len += 1 if ord(c) < 65536 else 2
+    else:
+        idx += 1  # get_line function trims the trailing '\n'
+    return idx
+
+
+class LspConcludeWorkspaceEditPanelCommand(sublime_plugin.WindowCommand):
```
predragnikolic, line 326: I tried to find other names for "LspConcludeWorkspaceEditPanelCommand":
LspSettleWorkspaceEditPanelCommand
LspResolveWorkspaceEditPanelCommand
LspCommitWorkspaceEditPanelCommand
LspSubmitWorkspaceEditPanelCommand

Does any of the suggestions look like a better fit?

predragnikolic:
there is a small bug in the code.
when "Preview" is clicked in the modal,
if you press ESC or press the discard button in the panel, the file rename still happens, but it should not.

plugin/edit.py
rchl: I'm biased since I've named it but it seems fine to me as is ;)
Also it's an internal command so its naming is not that important.

rchl (approved):

rchl:
After this PR I think we can release new version since we've waited long enough with some good fixes.

predragnikolic:
I think the PR is fine. 
I wanted to wait for a couple of days to see if Janos also wants to take a look.

alexzel:
Hello, please let me know if there is a way to disable this LSP rename dialog when I do right click -> rename on a file name. I'd prefer to use the native ST rename panel.

predragnikolic:
Hi @alexzel, there is no way right now to disable the rename dialog.

We can:

1. reintroduce back the ST rename command in the sidebar. I did prepare a [branch](https://github.com/sublimelsp/LSP/tree/dont-override-st-rename-sidebar) with such change.

2. We can revisit the approach of adding a setting which will determine when to show the prompt modal https://github.com/sublimelsp/LSP/pull/1923

3. A combination of 1 and 2.

4. do some other approach suggested by someone else.

@alexzel feel free to also leave your feedback.

jwortmann:
I think for now you could also add
```json
{
    "disabled_capabilities": {
        "workspace.fileOperations.willRename": true
    },
}
```
into the configs for your language server(s), if you don't need that feature in general. Then the rename prompt would still open in the command palette instead of the bottom panel, but it would only do the file rename and won't do any content changes in that file or in other files. We could also think about adding a condition for that in https://github.com/sublimelsp/LSP/blob/5d246387f4c0d77cce45caa8fad5bf40b181f4fa/boot.py#L277-L281
so if there is no server available with that capability, then it would not replace the builtin command. It might be a little tricky because we would first retrieve the file URI from the command args, to check for applicable sessions with that capability, but I think it should be possible.

rchl:
I think another alternative would be to mimick ST's rename input? Are there specific benefits to using command palette?

predragnikolic:
@alexzel the next release of LSP will not override the default ST rename in the sidebar. 

It was addressed in this PR https://github.com/sublimelsp/LSP/pull/2764

Issue #2706, opened 2025-12-04
Support workspace.workspaceEdit.resourceOperations


rchl:
**Is your feature request related to a problem? Please describe.**

Add support for https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspaceEditClientCapabilities so that workspace edits can create/rename and delete files.

PR #2717, opened 2025-12-17
Return number of changes and changed files when applying WorkspaceEdit

jwortmann:
Implementing the idea I brought up in https://github.com/sublimelsp/LSP/pull/2498#issuecomment-3650892081.

Also fixes a bug that always returned the constant number of 3 changes per file instead of the actual changes number when doing a rename operation.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1964,7 +1964,7 @@ class Session(TransportCallbacks):
         auto_save = userprefs().refactoring_auto_save if is_refactoring else 'never'
         summary: WorkspaceEditSummary = {
             'total_changes': sum(len(value[0]) for value in changes.values()),
-            'updated_files': len(changes)
+            'edited_files': len(changes)
         }
         for uri, (edits, label, view_version) in changes.items():
             view_state_actions = self._get_view_state_actions(uri, auto_save)
--- a/plugin/core/edit.py
+++ b/plugin/core/edit.py
@@ -16,7 +16,7 @@ WorkspaceChanges = Dict[str, Tuple[List[Union[TextEdit, AnnotatedTextEdit]], Opt
 
 class WorkspaceEditSummary(TypedDict):
     total_changes: int
-    updated_files: int
+    edited_files: int
     created_files: NotRequired[int]
     renamed_files: NotRequired[int]
     deleted_files: NotRequired[int]


predragnikolic (approved):

plugin/core/sessions.py
```diff
@@ -1958,6 +1962,8 @@ def handle_view(
         selected_sheets = self.window.selected_sheets()
         promises: list[Promise[None]] = []
         auto_save = userprefs().refactoring_auto_save if is_refactoring else 'never'
+        total_changes = sum(len(value[0]) for value in changes.values())
+        file_count = len(changes)
```
predragnikolic: This is just an observation.

Once we implement file resource operations when applying workspace edits (https://github.com/sublimelsp/LSP/issues/2706), like CreateFile | RenameFile | DeleteFile:

> documentChanges?: (
		[TextDocumentEdit](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocumentEdit)[] |
		([TextDocumentEdit](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocumentEdit) | [CreateFile](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#createFile) | [RenameFile](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#renameFile) | [DeleteFile](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#deleteFile))[]
	);

I assume we will take into consideration created/renamed/deleted files when summarizing total_changes and file_count.

predragnikolic: Right now we display a message like: `Replace {total_changes} occurrences across {file_count} files?`.

Once we implement file resource operations, 
what message would be useful to show to user? Should we tweak it a bit to include the count of created/renamed/deleted files as well?

```
Updated 25 files 
Created  1 file 
Renamed  1 file 
 Delete  1 file
```

This is just a question to think about.

jwortmann: Yes, that seems like a good idea. In that case we would need to add 3 more components with the numbers of created/renamed/deleted files to the tuple returned from `apply_workspace_edit_async`. Or would it be better to use a dict instead of a tuple now? Like this:

```python
class WorkspaceEditSummary(TypedDict):
    total_changes: int
    file_count: int
```

Then the result type would be a `Promise[WorkspaceEditSummary]`. And later when resource operations are also supported, we could extend the WorkspaceEditSummary dict:

```python
class WorkspaceEditSummary(TypedDict):
    total_changes: int
    file_count: int
    created_files: NotRequired[int]
    renamed_files: NotRequired[int]
    deleted_files: NotRequired[int]
```
(Or alternatively without the `NotRequired`)

For the status bar, we could just append the file counts if there are created/renamed/deleted files. Something like
```python
message = f"Applied {summary['total_changes']} changes in {summary['file_count']} files"
if created_files := summary.get('created_files'):
    message += f", created {created_files} file"
    if created_files > 1:
        message += "s"
if renamed_files := summary.get('renamed_files'):
    message += f", renamed {renamed_files} file"
    if renamed_files > 1:
        message += "s"
if deleted_files := summary.get('deleted_files'):
    message += f", deleted {deleted_files} file"
    if deleted_files > 1:
        message += "s"
```

jwortmann: I implemented that idea in https://github.com/sublimelsp/LSP/pull/2717/changes/84d7dc562ab6fc26558dde412050fc35c9c6a60d.

I think it is a better approach than a tuple, if/when we add support for resource operations in the future.

PR #2720, opened 2026-01-13
Switch docs from "Material for MkDocs" to "Zensical" 

predragnikolic:
Material for MkDocs is entering [maintenance mode](https://github.com/squidfunk/mkdocs-material/issues/8523)

The "Material for MkDocs" maintainers are now developing Zensical.
https://squidfunk.github.io/mkdocs-material/blog/2025/11/05/zensical/

Current UI:
Screenshot 2026-01-13 at 14 35 19

The UI with this PR:
Screenshot 2026-01-13 at 14 34 56

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -6,12 +6,12 @@ on:
     paths:
       - ".github/workflows/docs.yml"
       - "docs/**"
-permissions:
-  contents: read
-  pages: write
-  id-token: write
 jobs:
   deploy:
+    permissions:
+      contents: read
+      pages: write
+      id-token: write
     environment:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}


rchl:
Possibly you can disable https://mise.jdx.dev/configuration/settings.html#idiomatic_version_file_enable_tools using ENV variable (to avoid creating another configuration file).

.github/workflows/docs.yml
```diff
@@ -6,15 +6,26 @@ on:
     paths:
       - ".github/workflows/docs.yml"
       - "docs/**"
+permissions:
```
predragnikolic: I copy pasted the GA from the zensical docs https://zensical.org/docs/publish-your-site/#with-github-actions

netlify.toml
```diff
@@ -1,7 +1,7 @@
 # https://docs.netlify.com/configure-builds/file-based-configuration
 
 [build]
-  environment = { PYTHON_VERSION = "3.8" }
+  environment = { PYTHON_VERSION = "3.13", MISE_OVERRIDE_TOOL_VERSIONS_FILENAMES = "none" }
```
predragnikolic: the root .python-version does seem to affect the python version on netlify.
Although I have set  PYTHON_VERSION = "3.13"
the .python-version file did cause mise to error.
Found a way to disable the tools version files by stumbling upon this comment on github https://github.com/jdx/mise/discussions/5245#discussioncomment-13321301

predragnikolic:
Thanks Rafal, saw your comment just now.
While MISE_OVERRIDE_TOOL_VERSIONS_FILENAMES does work -> https://mise.jdx.dev/configuration/settings.html#override_tool_versions_filenames

I will try your suggestion -> idiomatic_version_file_enable_tools 
as that is what the netlify logs also suggested. (I was just confused how and where to pass that setting)

netlify.toml
predragnikolic: I switched 
from using MISE_OVERRIDE_TOOL_VERSIONS_FILENAMES 
to MISE_IDIOMATIC_VERSION_FILE_ENABLE_TOOLS

.github/workflows/docs.yml
predragnikolic: That said, I am not sure why I needed this `permission:` section here.
I will try to read more about it.

docs/Makefile
```diff
@@ -3,13 +3,10 @@
...
 
 install:
-	pip install mkdocs mkdocs-material mkdocs-redirects
+	pip install zensical
 
 build:
-	mkdocs build
-
-deploy:
-	mkdocs gh-deploy --force
```
predragnikolic, line 12: Just a note, 
Zensical CLI doesn't have a gh-deploy command

> gh-deploy command: we plan to support many more deployment methods, so we're not implementing MkDocs' gh-deploy command. See our [publishing guide](https://zensical.org/docs/publish-your-site/) for alternatives. - copied from https://zensical.org/compatibility/cli/

.github/workflows/docs.yml
predragnikolic: It is not really needed, so I will remove it.
[37a09d2](https://github.com/sublimelsp/LSP/pull/2720/commits/37a09d25ff7ef6deb978178db1fee7a685c2d92d)

.github/workflows/docs.yml
```diff
@@ -6,10 +6,6 @@ on:
     paths:
       - ".github/workflows/docs.yml"
       - "docs/**"
-permissions:
-  contents: read
-  pages: write
-  id-token: write
```
rchl: I think those are needed (but could be set per job) - https://docs.github.com/en/pages/getting-started-with-github-pages/using-custom-workflows-with-github-pages

predragnikolic:
https://lsp.sublimetext.io/ is now using Zensical (which also means that the docs GA works)

PR #2722, opened 2026-01-14
Uncomment await_message for textDocument/didSave

### LINKED ISSUE #2461: Investigate why didSave notification is not sent if view closed immediately after saving
When a file is saved, a `textDocument/didSave` notifications should be sent to the server if it opts-in via TextDocumentSyncOptions.

But if the file gets closed immediately after saving, the notification is not sent (at least here on Windows, and apparently also on CI on all OS, but according to https://github.com/sublimelsp/LSP/pull/2438#issuecomment-2054138492 it seems to work in practice on Linux).

Key binding example:
```json
    {
        "keys": ["f1"],
        "command": "chain",
        "args": {
            "commands": [
                ["save", {"async": false}],
                ["close", {}]
            ]
        }
    },
```

I expect to see `textDocument/didSave` in the LSP server log panel before `textDocument/didClose`, but it's missing.

If this gets fixed, this line in the tests could be uncommented:
https://github.com/sublimelsp/LSP/blob/4c8da7820cff68901be808ea8bf2fcf14138d249/tests/test_single_document.py#L394

predragnikolic:
closes #2461

With the JSON language server,
I cannot reproduce #2461 anymore.
didSave is send before didClose, so I assume this has been resolved
```
:: [14:32:46.549]  -> LSP-json textDocument/didSave: {'textDocument': {'uri': 'file:///Users/predrag/Library/Application%20Support/Sublime%20Text/Packages/LSP/.release.json'}}
:: [14:32:46.553]  -> LSP-json textDocument/didClose: {'textDocument': {'uri': 'file:///Users/predrag/Library/Application%20Support/Sublime%20Text/Packages/LSP/.release.json'}}
```


predragnikolic:
Well, didSave is sent before didClose on my Mac :D 
but in the Github Action it seems like it is unreliable -> https://github.com/sublimelsp/LSP/actions/runs/20998731204/job/60362510002

jwortmann:
My guess why it originally didn't work was the following:
We use the `ViewEventListener.on_post_save_async` and `ViewEventListener.on_close` methods to trigger the save and close notifications. For testing we can use
```python
import sublime_plugin

class CloseListener(sublime_plugin.ViewEventListener):

        print('on_post_save_async')

    def on_close(self) -> None:
        print('on_close')
```

and when I run the key binding from #2461, most of the time I see
```
on_close
on_post_save_async
```
(in that order) in the console, but sometimes I see only
```
on_close
```

Commands are ran on the UI thread by ST, and so I assume the "save" and "close" actions are executed in the same rendering frame with no time in between. When the view gets closed, its listener is destroyed and it seems like that might cause the `on_post_save_async` method sometimes not even to be triggered. But even when it does trigger, we already use `on_close` to clear/destroy the SessionView (on the async thread via `set_timeout_async`)
https://github.com/sublimelsp/LSP/blob/a1574b9cd9fdc3a82a0c3b01b4df5cae37a43020/plugin/documents.py#L482-L486

And `on_post_save_async` calls a method on the SessionView and on the SessionBuffer to eventually trigger the `didSave` notification, so there might even be a race condition between sending the notification and destroying the SessionView.

---

But when I run the key binding from #2461 now, I can see that `didSave` is sent before `didClose` on Windows as well. And I think to remember that when I originally tested it, it would never send the `didSave`. So not sure what (partially) fixed it.

PR #2564, opened 2024-11-30
Fix dynamic registration and diagnostic streams

jwortmann:
Fix #2561

PR #2727, opened 2026-01-17
Fix diagnostic dynamic registration and rework Goto Diagnostic command

### LINKED ISSUE #2561: LSP doesn't handle dynamic registration correctly
I am using https://github.com/SofusA/roslyn-language-server for C# but there are no diagnostics. The reason seems to be that sublime doesn't request correct textDocument/diagnostic even after they are registered. I am attaching both the VSCode traces and sublime lsp logs below for comparison. Comparing the logs it can be seen that even though the DocumentCompilerSemantic was registered along with some other options, sublime doesn't request for them whereas VSCode does and it is this request that provides the diagnostics.

[sublime-logs.txt](https://github.com/user-attachments/files/17792528/sublime-logs.txt)
[trace.txt](https://github.com/user-attachments/files/17792529/trace.txt)

### LINKED ISSUE #2731: Rust analyzer breaking on file save since newest update
Ever since the most recent LSP update, the rust analyzer has been breaking and no longer displaying any warnings after saving the file (though errors manage to still show). All packages are up to date. Manually reverting to 2.7.0 manages to fix the problem, though I'm not well versed in the LSP and therefore have no clue why, but hopefully the attached logs will help. The main thing that sticks out to me is a sudden precense of "canceled by client", which doesn't appear in the previous version, though given it's only the main LSP server itself that's changed and not the analyzer, I'm not sure why that's the case

**Logs**
[log-2.7.0.txt](https://github.com/user-attachments/files/24797690/log-2.7.0.txt)
[log-2.8.0.txt](https://github.com/user-attachments/files/24797757/log-2.8.0.txt)

**Environment**
- OS: Windows
- Sublime Text version: 4200
- LSP version: Broken on Latest (4070-2.8.0), Working on 4070-2.7.0
- Language servers used: LSP-rust-analyzer

jwortmann:
Fixes #2561
Fixes #2731
Supersedes #2564 (which still had some bugs and TODOs)
Supersedes #2273

Instead of fixing dynamic registration of multiple providers for the same method (https://github.com/microsoft/language-server-protocol/issues/1069), this PR adds a special case for the registration and unregistration of diagnostic providers to handle them correctly. Diagnostic providers are currently the only LSP capability for which this is relevant. Multiple diagnostic streams should work correctly now, and also pull diagnostics and publishDiagnostics can work at the same time.
 
The DiagnosticStorage class and the diagnostics requests were fully rewritten to account for multiple diagnostic streams with different identifiers.

I have also rewritten the "Goto Diagnostic" command and combined that with "Goto Diagnostic in Project" into a single entry in the command palette and main menu. That is because it actually was the same command anyway and when you have the diagnostics quick panel (in the command palette) open for a particular file, the file selector is only a single <kbd>backspace</kbd> key press away, so I think having two different entries for that is quite useless. If the active view has some diagnostics in it, triggering the "Goto Diagnostic" command opens the quick panel with the active view preselected, and otherwise it opens the file selector first (unless there aren't any diagnostics in the window at all - then the command is disabled). Diagnostics are now sorted by location. And I think the code should be significantly cleaner and simpler now.

I haven't rigorously tested the PR yet, so possibly there might still be bugs that I have missed. Some servers which use dynamic registration for pull diagnostics are:
- Pyright (after initializing it registers two different diagnostic providers but with identical "identifier", then it immediately unregisters the first one)
- JETLS (Julia) (registers one diagnostic provider and also uses push diagnostics at the same time)
- rust-analyzer (I don't use Rust)
- [csharp-language-server](https://github.com/SofusA/csharp-language-server) (as mentioned in #2561; I don't use C#)

### DIFF
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -11,6 +11,7 @@ from ..protocol import FullDocumentDiagnosticReport
 from ..protocol import InlayHint
 from ..protocol import InlayHintParams
 from ..protocol import LSPErrorCodes
+from ..protocol import RelatedFullDocumentDiagnosticReport
 from ..protocol import SemanticTokensDeltaParams
 from ..protocol import SemanticTokensParams
 from ..protocol import SemanticTokensRangeParams
@@ -78,8 +79,14 @@ P = ParamSpec('P')
 
 
 def is_full_document_diagnostic_report(
-    diagnostic_report: DocumentDiagnosticReport | FullDocumentDiagnosticReport | UnchangedDocumentDiagnosticReport
+    diagnostic_report: FullDocumentDiagnosticReport | UnchangedDocumentDiagnosticReport
 ) -> TypeGuard[FullDocumentDiagnosticReport]:
+    return diagnostic_report['kind'] == DocumentDiagnosticReportKind.Full
+
+
+def is_related_full_document_diagnostic_report(
+    diagnostic_report: DocumentDiagnosticReport
+) -> TypeGuard[RelatedFullDocumentDiagnosticReport]:
     return diagnostic_report['kind'] == DocumentDiagnosticReportKind.Full
 
 
@@ -588,10 +595,10 @@ class SessionBuffer:
             # the diagnostics with the present resultId are still valid.
             self._document_diagnostic_pending_requests[identifier] = None
         self.session.diagnostics_result_ids[(self._last_known_uri, identifier)] = response.get('resultId')
-        if is_full_document_diagnostic_report(response):
+        if is_related_full_document_diagnostic_report(response):
             self.session.handle_diagnostics_async(self._last_known_uri, identifier, version, response['items'])
-        if 'relatedDocuments' in response:
-            for uri, diagnostic_report in response['relatedDocuments'].items():
+        if related_documents := response.get('relatedDocuments'):
+            for uri, diagnostic_report in related_documents.items():
                 uri = normalize_uri(uri)
                 self.session.diagnostics_result_ids[(uri, identifier)] = diagnostic_report.get('resultId')
                 if is_full_document_diagnostic_report(diagnostic_report):


plugin/diagnostics.py
```diff
@@ -1,14 +1,104 @@
...
+class DiagnosticsStorage:
+
+    def __init__(self) -> None:
+        self._providers: dict[str | None, DiagnosticOptions | DiagnosticRegistrationOptions] = {}
+        self._identifiers: set[DiagnosticsIdentifier] = set()
+        self._workspace_diagnostics_identifiers: set[DiagnosticsIdentifier] = set()
+        self._diagnostics: dict[DocumentUri, dict[DiagnosticsIdentifier, list[Diagnostic]]] = {}
+        self.token_identifier_map: dict[str, DiagnosticsIdentifier] = {}  # maps identifiers to partial result tokens
+
+    @lru_cache
```
jwortmann: This is for performance, so that it is not re-evaluated on each diagnostic request. But technically the cache must be cleared if the syntax for the view gets changed. But maybe it is good for now?

rchl: I have only skimmed through the changes so far so I don't really understand how things work yet but wouldn't cache need to be invalidated when new provider is registered (or unregistered)?

jwortmann: Yes, and that is already the case (see `_update_identifiers` method below).

rchl: Nice, didn't know about that method.

BTW. Is this something to worry about: https://docs.astral.sh/ruff/rules/cached-instance-method/ ?

plugin/goto.py
```diff
@@ -126,3 +145,175 @@ class LspSymbolImplementationCommand(LspGotoCommand):
...
+    def input_description(self) -> str:
+        return 'Goto Diagnostic'
+
+    def input(self, args: dict) -> sublime_plugin.CommandInputHandler | None:
+        view = self.window.active_view()
+        if not view:
+            return None
+        sessions = list(self.sessions())
+        if uri := args.get('uri'):
+            return DiagnosticUriInputHandler(self.window, view, sessions, uri)
```
rchl: My existing keybinding:

```json
    {
        "keys": [
            "f8"
        ],
        "command": "lsp_goto_diagnostic",
        "args": {
            "uri": "$view_uri"
        },
        "context": [
            {
                "key": "setting.lsp_active"
            }
        ]
    },
```

doesn't work anymore because `$view_uri` is not expanded anymore.

It might have been your intention in a way but IMO better to preserve backwards-compatibility.

rchl:
One thing from #2273 that we could add here (or later) is starting navigation from the closest diagnostic.
Always starting from the top can be disorienting.

EDIT: Actually diagnostics are sorted by session here. So implementing combined all-session sorting would also be good.

rchl:
The `plugin/core/diagnostics_storage.py` file is unused.

plugin/core/sessions.py
```diff
@@ -2074,27 +2079,39 @@ def session_views_by_visibility(self) -> tuple[set[SessionViewProtocol], set[Ses
...
-            if result_id is not None
+            {'uri': uri, 'value': result_id} for (uri, id_), result_id in self.diagnostics_result_ids.items()
+            if id_ == identifier and result_id is not None
         ]
         params: WorkspaceDiagnosticParams = {'previousResultIds': previous_result_ids}
-        if identifier := self.get_capability("diagnosticProvider.identifier"):
+        if identifier is not None:
             params['identifier'] = identifier
-        self.workspace_diagnostics_pending_response = self.send_request_async(
+        partial_result_token = _PARTIAL_RESULT_PROGRESS_PREFIX + str(self.request_id + 1)
```
rchl, line 2132: I don't see where `partial_result_token` name comes from and why it's prefixed with `_PARTIAL_RESULT_PROGRESS_PREFIX`. Based on how it's used, it feels to me like it's basically just a `request_id` and could be called that instead and just have a value of `self.request_id + 1`.

And if that's the case then `token_identifier_map`'s name could also be changed to something like `request_to_identifier_map`.

plugin/diagnostics.py
jwortmann: The identifiers cache is now cleared when the syntax or URI changes: https://github.com/sublimelsp/LSP/pull/2727/changes/fbb38a30c057af4dbe0f9493d798d4d4ad2ed1bb

---

> BTW. Is this something to worry about: https://docs.astral.sh/ruff/rules/cached-instance-method/ ?

> Using the functools.lru_cache and functools.cache decorators on methods can lead to memory leaks, as the global cache will retain a reference to the instance, preventing it from being garbage collected.

Well we need that method to be an instance method because the cache must be specific to the session. But it's not like we would create and delete sessions every second, so I assume this possible memory increase should be a non-issue.

jwortmann:
> EDIT: Actually diagnostics are sorted by session here. So implementing combined all-session sorting would also be good.

Yes that was my intention but I probably forgot to sort them again in case of multi-sessions.

*Edit:* Should be fixed with https://github.com/sublimelsp/LSP/pull/2727/changes/c9667bfbdcae7d0ddf71019694fdcf626e63b44c

---

~~I see that there is still a bug with the file previews for `res:` URIs. I will try to fix that.~~

*Edit:* I found no way to open a ressource file (`res` URI) as a transient sheet, but the bugs should be fixed now. The preview for `res` (and other non-file) URIs only works if the view is already open in the editor, which should typically be the case. And this use case of non-file URIs is probably quite rare.

plugin/diagnostics.py
jwortmann: We could probably make the `get_identifiers` a global function and then also pass the session or the session name as another argument. Not sure if that makes it more complicated though. I'll think about it.

rchl: Well, the ruff check is not about cache getting larger but about the whole `DiagnosticsStorage` not being garbage collected potentially. Though I guess it would still not prevent `Session` from being collected and ultimately get collected anyway...

rchl: Actually adding:

```py
    def __del__(self) -> None:
        print('DEL Storage')
```

to `DiagnosticsStorage` shows that it's not being collected on session being destroyed when lru-cache is used.

plugin/core/sessions.py
jwortmann: Yeah that's not immediately obvious. It is the same token that gets created in `send_request_async` which is called on the line after that.
https://github.com/sublimelsp/LSP/blob/ea0963d8ba474f47ba41fbc5999d0c84d933e56e/plugin/core/sessions.py#L2473-L2478

I agree that it is not very satisfying to do it like this, but I didn't find a better way. Do you have a better idea?

rchl: My point is not that it duplicates what is done there. My point is that the one you create doesn't really have anything to do with partial result tokens but still uses same format. My point is to just call it `request_id` and set to an int. I know it's still not ideal to duplicate part of what `send_request_async` is doing but that's another matter.

jwortmann: But the point of this map is that it maps the diagnostic identifiers to the partial result tokens that we use in our requests. So it must be the exact same value that we set for the "partialResultToken" field in the request.

This is used if the server uses progress notifications for the workspace diagnostics. In that case the notifications from the server only contain the partial result token, but we still have to identify the correct diagnostic stream for which this notification belongs to.

jwortmann: I guess we could also use the `request_id + 1` for the keys of this map and as the argument in the `_on_workspace_diagnostics_async` method, but that would be even less obvious and then we first need to convert the token back to the request_id when handling the progress notifications in `m___progress`.

Also that would not work if we later decide for example to change our "partialResultToken" format to for example UUIDs instead of the prefix + request number (although then we would need to refactor this line anyway to keep it in sync with the token generated in `send_request_async`). So I think using the token directly is the only thing that makes sense here.

rchl: Perhaps I have overlooked something since I didn't see the key coming from the server. I will look again and likely close this.

That said, I guess if we don't want to duplicate code then we could fetch that value from `Request` after request is dispatched since it mutates passed `Request`. But perhaps that would need some more type fixes or hacks.

plugin/core/windows.py
```diff
@@ -461,18 +460,16 @@ def update_diagnostics_panel_async(self) -> None:
...
-                    is_severity_included(max_severity), lambda _, diagnostic: format_diagnostic_for_panel(diagnostic)):
-                seen = path in contributions
-                contributions.setdefault(path, []).extend(contribution)
-                if not seen:
-                    contributions.move_to_end(path)
-        for path, contribution in contributions.items():
+            for uri, diagnostics in session.diagnostics.get_diagnostics(max_severity).items():
+                contributions.setdefault(uri, []).extend(diagnostics)
+        for uri, diagnostics in contributions.items():
+            _, path = parse_uri(uri)
```
rchl, line 474: I supposed this code is the reason why I'm seeing URIs with no diagnostics in the panel:

Image

plugin/session_buffer.py
```diff
@@ -543,56 +549,67 @@ def update_document_link(self, new_link: DocumentLink) -> None:
...
+            PendingDocumentDiagnosticRequest(version, request_id)
 
-    def _on_document_diagnostic_async(self, version: int, response: DocumentDiagnosticReport) -> None:
-        self._document_diagnostic_pending_request = None
-        self._if_view_unchanged(self._apply_document_diagnostic_async, version)(response)
+    def _on_document_diagnostic_async(
+        self, identifier: DiagnosticsIdentifier, version: int, response: DocumentDiagnosticReport
+    ) -> None:
+        self._document_diagnostic_pending_requests[identifier] = None
+        self._if_view_unchanged(self._apply_document_diagnostic_async, version)(identifier, response)
```
rchl: Looks like `self._diagnostics_versions` is never updated after receiving diagnostics which causes ignoring not to work:

<details>

<summary>My initial report of this problem (before I've investigated)</summary>

After document changes this happens within 1 second:

Screenshot 2026-01-17 at 22 13 29

Perhaps something to improve or just unfortunate timing (we requesting diagnostics after edit and server refreshing at the same time). Might also not be a new thing.

EDIT: Debugging this a bit.
3 document diagnostics requested in a row:
1. after document change - https://github.com/jwortmann/LSP/blob/e7e6f2957c0159bb0448f997668daa1665e6ca0c/plugin/session_buffer.py#L410
2. due to workspace diagnostic refresh - https://github.com/jwortmann/LSP/blob/068bdd5df6dfa9465ec707fb6feb8932b875b1bc/plugin/core/sessions.py#L2234
3. due to fetching code actions - https://github.com/jwortmann/LSP/blob/9bfb68da38cbb71fa932924536395f1f8c3f676a/plugin/code_actions.py#L146-L146

</details>

plugin/diagnostics.py
jwortmann: I added a global function with the lru-cache in https://github.com/sublimelsp/LSP/pull/2727/changes/59ed2baedd268c6b90ca4fc54eb4fac586aa0a58

plugin/goto.py
jwortmann: Added https://github.com/sublimelsp/LSP/pull/2727/changes/f3328bf19927d4979ad1f88ee5e0c98d6e9bfba1 to preserve backwards compatibility.

plugin/core/windows.py
jwortmann: Should be fixed with https://github.com/sublimelsp/LSP/pull/2727/changes/c9667bfbdcae7d0ddf71019694fdcf626e63b44c

plugin/session_buffer.py
jwortmann: Thanks for the testing.

The third request should be fixed with https://github.com/sublimelsp/LSP/pull/2727/changes/170eb0525c1d68fb6cb8d4ec53181119c02279b7 (could you check to confirm?)

I expect there to still be 2 requests because rust-analyzer abuses the `workspace/diagnostic/refresh` after every file content change. I'm quite confident we are doing the right thing by cancelling the pending request and sending a new one because of that. Note that it is not the intended way from the specs to use the refresh request after every change, so I would say this is the fault of rust-analyzer and they should fix or improve it on their side.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticWorkspaceClientCapabilities
```typescript
	/**
	 * Note that this event is global and will force the client to refresh all
	 * pulled diagnostics currently shown. It should be used with absolute care
	 * and is useful for situation where a server for example detects a project
	 * wide change that requires such a calculation.
	 */
```

plugin/core/open.py
```diff
@@ -124,6 +125,25 @@ def fullfill(resolve: ResolveFunc[sublime.View | None]) -> None:
     return promise
 
 
+def open_resource(window: sublime.Window, uri: DocumentUri, group: int = -1) -> Promise[sublime.View | None]:
```
rchl: Doesn't need to be a promise.

plugin/goto.py
```diff
@@ -1,16 +1,38 @@
...
 from ..protocol import LocationLink
+from .core.constants import DIAGNOSTIC_KINDS
+from .core.input_handlers import PreselectedListInputHandler
+from .core.paths import simple_project_path
+from .core.protocol import Point
 from .core.protocol import Request
 from .core.registry import get_position
 from .core.registry import LspTextCommand
+from .core.registry import LspWindowCommand
 from .core.sessions import Session, method_to_capability
```
rchl: Not a new issue but `method_to_capability` is in `.core.types`.

plugin/core/sessions.py
```diff
@@ -1858,6 +1867,22 @@ def handle_continuation(view: sublime.View | None) -> None:
...
+    ) -> Promise[sublime.View | None]:
+        result: PackagedTask[sublime.View | None] = Promise.packaged_task()
+
+        def handle_continuation(view: sublime.View | None) -> None:
+            if view and r:
+                sublime.set_timeout(functools.partial(center_selection, view, r))
+            sublime.set_timeout_async(lambda: result[1](view))
+
+        sublime.set_timeout(lambda: Promise.resolve(open_resource(self.window, uri, group)).then(handle_continuation))
+        return result[0]
```
rchl, line 1883: Can be simplified:

```suggestion
    def _open_res_uri_async(
        self,
        uri: DocumentUri,
        r: Range | None = None,
        group: int = -1
    ) -> Promise[sublime.View | None]:

        def continue_on_main_thread() -> None:
            view = open_resource(self.window, uri, group)
            if view and r:
                sublime.set_timeout(functools.partial(center_selection, view, r))
            sublime.set_timeout_async(lambda: result[1](view))

        result: PackagedTask[sublime.View | None] = Promise.packaged_task()
        sublime.set_timeout(continue_on_main_thread)
        return result[0]
```

BTW. I wonder if we should await until the file loads...

Since the files are loaded from memory then perhaps not.
And if loading still happens then extra timeouts might just happen to make it work without explicit wait.

plugin/core/open.py
```diff
@@ -124,6 +125,23 @@ def fullfill(resolve: ResolveFunc[sublime.View | None]) -> None:
...
+    """
+    prefix = 'res:/Packages/'
+    if not uri.startswith(prefix):
+        return None
+    if group != -1:
+        window.focus_group(group)
+    resource_path = uri[len(prefix):]
+    window.run_command('open_file', {'file': f'${{packages}}/{resource_path}'})
+    file = os.path.join(ST_PACKAGES_PATH, *resource_path.split('/'))
+    return _find_open_file(window, file, group)
```
rchl, line 142: Would `window.open_file()` not work instead? Why is this necessary?

plugin/core/sessions.py
```diff
@@ -1470,14 +1473,14 @@ def register_session_buffer_async(self, sb: SessionBufferProtocol) -> None:
         self._session_buffers.add(sb)
         for data in self._registrations.values():
             data.check_applicable(sb, suppress_requests=True)
-        if (uri := sb.get_uri()) and (diagnostics := self.diagnostics.diagnostics_by_document_uri(uri)):
+        if (uri := sb.get_uri()) and (diagnostics := self.diagnostics.get_diagnostics_for_uri(uri)):
             self._publish_diagnostics_to_session_buffer_async(sb, diagnostics)
 
     def _publish_diagnostics_to_session_buffer_async(
-        self, sb: SessionBufferProtocol, diagnostics: list[Diagnostic]
+        self, sb: SessionBufferProtocol, diagnostics: list[Diagnostic], version: int | None = None
```
rchl, line 1480: Perhaps remove default `None` value? Not a big deal though.

plugin/core/sessions.py
```diff
@@ -1371,8 +1374,8 @@ def __init__(self, manager: Manager, logger: Logger, workspace_folders: list[Wor
         self.state = ClientStates.STARTING
         self.capabilities = Capabilities()
         self.diagnostics = DiagnosticsStorage()
-        self.diagnostics_result_ids: dict[DocumentUri, str | None] = {}
-        self.workspace_diagnostics_pending_response: int | None = None
+        self.diagnostics_result_ids: dict[tuple[DocumentUri, DiagnosticsIdentifier], str | None] = {}
+        self.workspace_diagnostics_pending_responses: dict[DiagnosticsIdentifier, int | None] = {}
```
rchl, line 1378: Could we make value have type `int` instead of `int | None`? I guess it should work if we pop the key instead of resetting value to `None`?

plugin/core/sessions.py
rchl: I have an idea how to clean this up but will submit it later to avoid complicating this PR too much.

plugin/session_buffer.py
```diff
@@ -656,7 +677,8 @@ def present() -> None:
 
     def has_latest_diagnostics(self) -> bool:
         if view := self.some_view():
-            return self._diagnostics_version == view.change_count()
+            view_version = view.change_count()
+            return all(version == view_version for version in self._diagnostics_versions.values())
```
rchl: This is probably not correct (and wasn't before either) but if, as we've learned, the server would decide not to publish diagnostics after a change (because nothing has changed) then `has_latest_diagnostics` could return `False` and then the caller would not get any diagnostics.

Luckily I only know of TLS doing that and only in case the diagnostics are empty so in practice it wouldn't be a problem. But in theory...

plugin/diagnostics.py
```diff
@@ -1,14 +1,113 @@
...
+
+
+@lru_cache
+def get_diagnostics_identifiers(session: Session, view: sublime.View) -> set[DiagnosticsIdentifier]:
+    return session.diagnostics.get_identifiers(view)
+
+
+class DiagnosticsStorage:
+
+    def __init__(self) -> None:
```
rchl, line 42: Would it make sense to just mention in the doc block that this storage is per-session?

plugin/session_buffer.py
jwortmann: Yeah we should probably exclude the diagnostics pushed by the server via publishDiagnostics in this check.

The only place where this method is used is in
https://github.com/sublimelsp/LSP/blob/55d0ffab6b658d39d2fc21eea469fa53b956de01/plugin/documents.py#L285-L290

And this is all-or-nothing, i.e. if there are multiple diagnostic streams and if there is still a pending response from only one of the streams, then no diagnostics are returned at all.

But I think this is wrong alltogether because `sb.diagnostics` seems to be overwritten with each diagnostic response or notification from publishDiagnostics. It should probably use the `session.diagnostics.get_diagnostics_for_uri` instead.

plugin/core/sessions.py
jwortmann: The set of identifiers is unlikely to change over the lifetime of the session, so it might be better for performance to set values to `None` when there is no open request, instead of popping and inserting the keys into the dict over and over again (not sure if it matters in practice though).

plugin/core/open.py
jwortmann: `window.open_file()` can only open files that exist in the file system on disk, but not from .sublime-package archives.

plugin/core/sessions.py
jwortmann: I initially thought that the `open_file` command would be blocking anyway, but it turned out the `set_timeout` is needed for `center_selection` to work. I haven't tested with really big files though.

plugin/session_buffer.py
jwortmann: I've updated `has_latest_diagnostics` to exclude diagnostics from publishDiagnostics, but now a test fails because it would assume that diagnostics are ignored for the code action request when there isn't any new publishDiagnostics notification after a didChange.

Perhaps we need to delay the code action request after buffer change a bit in general, so that there is time for pull diagnostics responses and publishDiagnostics notifications to arrive, before the request is sent? Because the request contains the relevant diagnostics for possible quick fixes.

rchl: Delaying sounds good in general.

plugin/session_buffer.py
```diff
@@ -543,56 +554,72 @@ def update_document_link(self, new_link: DocumentLink) -> None:
...
-                    cast(SessionBuffer, sb)._apply_document_diagnostic_async(
-                        None, cast(DocumentDiagnosticReport, diagnostic_report))
-
-    def _on_document_diagnostic_error_async(self, version: int, error: ResponseError) -> None:
-        self._document_diagnostic_pending_request = None
+            self.session.handle_diagnostics(self._last_known_uri, identifier, version, response['items'])
+        for uri, diagnostic_report in cast(RelatedDocuments, response.get('relatedDocuments', {})).items():
+            self.session.diagnostics_result_ids[(uri, identifier)] = diagnostic_report.get('resultId')
+            if diagnostic_report['kind'] == DocumentDiagnosticReportKind.Full:
+                self.session.handle_diagnostics(uri, identifier, None, diagnostic_report['items'])
```
rchl: If we if-check the block with `if 'relatedDocuments' in response:` like before then we don't need to define extra type or do the cast.


```suggestion
        if 'relatedDocuments' in response:
            for uri, diagnostic_report in response['relatedDocuments'].items():
                self.session.diagnostics_result_ids[(uri, identifier)] = diagnostic_report.get('resultId')
                if diagnostic_report['kind'] == DocumentDiagnosticReportKind.Full:
                    self.session.handle_diagnostics(uri, identifier, None, diagnostic_report['items'])
```

plugin/session_buffer.py
jwortmann: I could imagine to delay code actions until one of the following conditions is fullfilled:
- if the server supports pull diagnostics, when all responses from pull diagnostics have arrived and are handled
- if the server doesn't support pull diagnostics, when a publishDiagnostics notification has arrived and is handled
- after a predefined time delay, for example 1 second after sending the didChange and diagnostic requests

But this should probably be done in a separate pull request.

plugin/session_buffer.py
jwortmann: I thought that saving one indentation level without the if-condition would be more elegant. But I can revert to the previous style if you prefer that.

To be honest I'm a bit surprised that Pyright is not smart enough to realize that in case of `{}` there is nothing to iterate, so the types for `uri` and `diagnostic_report` should be well defined and the cast wouldn't be necessary. I'm contemplating if this is a bug, or rather a missing feature in Pyright that is worth to report/suggest in their repository.

plugin/session_buffer.py
rchl: Sounds good to address it later.

plugin/session_buffer.py
rchl: Not using cast always wins in my book.
I'm not surprised that `{}` fallback doesn't pass type check since it's missing some required properties like `kind`. I suppose pyright doesn't know that dict is empty. It just infers the type as `dict[Unknown, Unknown]`.

plugin/core/sessions.py
jwortmann: That would be good. I tried to create the request object first and then retrieve the partialResultToken from it after the `send_request_async` line. But then I realized that it is not possible, because we already need that partial result token for the callback functions that are passed into `send_request_async`. I'm looking forward to see your approach.

rchl: I won't spoil it but I think it will be pretty clean :)

jwortmann:
I won't merge right away because I think I've discovered two more things that are not fully correct:

- In `SessionBuffer.on_diagnostics_async` the line at
    ```python
            def present() -> None:
                self._diagnostics_versions[None] = diagnostics_version
    ```
    is useless, because we now ignore the version for publishDiagnostics anyway, and also this method gets called as well after pull diagnostics arrive. So I think this line can simply be removed.

- when a diagnostic stream is unregistered by the server, we should either pop the corresponding identifier from the `SessionBuffer._diagnostics_versions` dict, or the `has_latest_diagnostics` method should only iterate over the currently available identifiers.

I'll take another look later.

plugin/core/sessions.py
```diff
@@ -2231,19 +2270,28 @@ def m_workspace_diagnostic_refresh(self, params: None, request_id: Any) -> None:
...
-        uri = params["uri"]
         reason = mgr.should_ignore_diagnostics(uri, self.config)
         if isinstance(reason, str):
             debug("ignoring unsuitable diagnostics for", uri, "reason:", reason)
             return
-        diagnostics = params["diagnostics"]
-        self.diagnostics.add_diagnostics_async(uri, diagnostics)
+        session_buffer = self.get_session_buffer_for_uri_async(uri)
+        if session_buffer and version is not None and version != session_buffer.last_synced_version:
+            # FIXME version > session_buffer.last_synced_version
```
predragnikolic: 👀

plugin/core/sessions.py
```diff
@@ -2252,12 +2300,15 @@ def m_client_registerCapability(self, params: RegistrationParams, request_id: An
...
             if self.config.is_disabled_capability(capability_path):
                 continue
-            debug(f"{self.config.name}: registering capability:", capability_path)
             options = registration.get("registerOptions")
             if not isinstance(options, dict):
                 options = {}
             options = self.config.filter_out_disabled_capabilities(capability_path, options)
             registration_id = registration["id"]
+            if capability_path == 'diagnosticProvider':
+                self.diagnostics.register_provider(registration_id, cast(DiagnosticOptions, options))
```
predragnikolic: should we put a debug statement to print when we register/unregister a diagnostic provider?

predragnikolic:
minor bug with goto next/prev diagnostics

Given this (incorrect) rust file:
```
fn main() {
    let x: Option<str> = None;
}
™
```

✅ If I slowly invoke the go to next/previous diagnostics,
The hover popup is position bellow that diagnostic, as expected.

❌ If I quickly invoke the go to next/previous diagnostics,
The hover popop might get position on the previous diagnostics location, 
instead of the current diagnostics postion.

See video: (note that from 0s-7s second I am navigating slowly. From 10s-21s, I am invoking the goto next command fast)

https://github.com/user-attachments/assets/aa52d58e-025f-4604-9722-d873dc9198eb


On the 12th second, 
the hover popup is positioned on the 2 line, 
but it should be positioned on the 4th line.

plugin/core/sessions.py
rchl: When reviewing, I was very close to saying instead that we should remove the existing debug logs. I don't see why would one need to see those. I feel like those logs might only be useful during developing of the dynamic registration functionality itself but now that it's done, could be removed.

jwortmann:
> minor bug with goto next/prev diagnostics

I don't think it is related to this pull request because there was nothing essentially changed for navigating diagnostics.

We use a 200 milliseconds delay for showing the popup:
https://github.com/sublimelsp/LSP/blob/531ac54e01cd75b6650ac6632ae56fa755baa741/plugin/core/registry.py#L290-L293

So if you repeatingly invoke the command with a shorter delay, it might get out of sync somehow. We could use `view.show_at_center(diag_pos, animate=False)`, remove the delay and then see if that fixes it, but I'd probably prefer to keep the scrolling animation.

---

*Edit:* I think I found a way to fix it: https://github.com/sublimelsp/LSP/pull/2727/changes/c784568b0d20a2ce2440ae0f84048f7ed59afc2a
(at least it didn't seem to happen anymore in my tests)

plugin/core/sessions.py
jwortmann: I've tried to debug this, but it turns out that it also happens on the main branch when typing fast (with short breaks in between).

It can be tested by adding
```python
        if version > self._last_synced_version:
            print(f"version {version}, last synced version {self._last_synced_version}")
```

before the `self._if_view_unchanged` line at
https://github.com/sublimelsp/LSP/blob/531ac54e01cd75b6650ac6632ae56fa755baa741/plugin/session_buffer.py#L572-L574

But I don't know why that happens. The `self._last_synced_version` is set to the same `version` that is passed further to the method that sends the diagnostics request:
https://github.com/sublimelsp/LSP/blob/531ac54e01cd75b6650ac6632ae56fa755baa741/plugin/session_buffer.py#L382-L392

So `version` should never be greater than `self._last_synced_version`... 😕

jwortmann: I still don't know why that happens, and we should probably keep that in mind and hopefully figure it out some time.
But I've realized that we can actually not ignore the diagnostics even if the version is outdated (i.e. document content changed in the meanwhile), because pull diagnostics allow to return a [UnchangedDocumentDiagnosticReport](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#unchangedDocumentDiagnosticReport), which means that the previous response is still valid. So if we would drop some of the responses because a new request will be made shortly anyway, we might end up with a wrong (older) result in the diagnostics store if the next response is an UnchangedDocumentDiagnosticReport. Therefore I decided to remove that check and I think it also needs to be removed from `SessionBuffer._on_document_diagnostic_async`.

jwortmann:
I pushed some fixes and changes for the workspace diagnostics requests. For workspace diagnostics it is possible that the server leaves the request open "forever" and then sends progress notifications with updates, basically as a replacement for publishDiagnostics notifications. If the server handles it that way, some of the changes probably won't have a relevant effect. Servers with workspace diagnostics support that I know are JETLS for Julia (since latest server release 2026-01-23; does send progress notification(s) but then also sends a response to the requests), and iirc the Luau (Roblox Lua) language server (I haven't tested that).

Previously, new workspace diagnostics requests were triggered by any buffer change, but with an additional 3 second debounce time. Now I decided to trigger new workspace diagnostics requests only on initialize / provider registration, and whenever a file gets saved. Workspace diagnostics are probably mostly used to keep diagnostics in other files updated, which are typically not visible at the moment, so I think it's not necessary to retrigger them on every buffer change. Without that we can save the additional debouncing work and also reduce the workload for the server. Another change I made is that workspace diagnostics for the current view are not ignored anymore (unless the diagnostics version is outdated). Although we do request diagnostics for the current view anyway via `textDocument/diagnostic`, I could imagine that a server might use a workspace diagnostics stream perhaps for some differnt kind of diagnostics than the regular `textDocument/diagnostic` responses. So with the implementation of different diagnostics streams, we might miss out some diagnostics if we would ignore workspace diagnostics reports for the active view. For example JETLS has multiple diagnostics sources, with basic diagnostics on typing via `textDocument/diagnostic`, and also a full analysis that runs only on save (although that one currently uses publishDiagnostics notifications, https://aviatesk.github.io/JETLS.jl/release/diagnostic/#diagnostic/source).

Overall the workspace diagnostics handling should be a bit simpler now. I think the pull request should be good now, in case anyone wants to re-review the additional changes.

plugin/session_buffer.py
jwortmann: I eventually made the type inference work properly and simplified these lines a bit in https://github.com/sublimelsp/LSP/pull/2727/changes/d48baa75f3049812d5ebcbc64c725415b9f4d104.

captain0xff:
Thank you for the fix :)
Looking forward for the next release.

rchl:
> Previously, new workspace diagnostics requests were triggered by any buffer change, but with an additional 3 second debounce time. Now I decided to trigger new workspace diagnostics requests only on initialize / provider registration, and whenever a file gets saved.

I've noticed this behavior (note that the code action request was triggered after save):

Screenshot 2026-01-31 at 15 10 59

ruff gets notified of changed `pyproject.toml` file and requests diagnostics refresh. We don't do that. You said we do that on save but just saving the file without doing any changes also doesn't trigger diagnostics. We need to actually make some change to trigger diagnostics.

Ideally the diagnostics get refreshed automatically on changing pyproject.toml file in this case.

jwortmann:
We should trigger document diagnostics after a `workspace/diagnostic/refresh`. Not sure why they don't show up here.

After a save only workspace diagnostics are requested (unless the request is still open or server doesn't support workspace diagnostics).

Could you show the full logs?

jwortmann:
> We should trigger document diagnostics after a `workspace/diagnostic/refresh`. Not sure why they don't show up here.

Could it be that no Python source file was ~~focused~~ visible? In that case diagnostics are requested the next time when activating the file.

jwortmann:
> but just saving the file without doing any changes also doesn't trigger diagnostics

That probably means that LSP-ruff doesn't support workspace diagnostics (alternatively it could be that it leaves the workspace diagnostics request open, but we need to see the full logs for that).

rchl:
Is the issue perhaps that we don't trigger workspace diagnostics if there were no changes in the file since last save?

https://github.com/sublimelsp/LSP/blob/290dd6cb063941eb0c192a6ad14dbed691bcc990/plugin/session_buffer.py#L436-L451

Here is the full log:

```
:: [17:42:11.038] --> LSP-ruff initialize (1): {'processId': 93862, 'clientInfo': {'name': 'Sublime Text LSP', 'version': '2.8.1'}, 'rootUri': 'file:///usr/local/workspace/lsp/lsp-python-types', 'rootPath': '/usr/local/workspace/lsp/lsp-python-types', 'workspaceFolders': [{'name': 'lsp-python-types', 'uri': 'file:///usr/local/workspace/lsp/lsp-python-types'}], 'capabilities': {'general': {'regularExpressions': {'engine': 'ECMAScript'}, 'markdown': {'parser': 'marko', 'version': '2.2.2'}}, 'textDocument': {'synchronization': {'dynamicRegistration': True, 'didSave': True, 'willSave': True, 'willSaveWaitUntil': True}, 'hover': {'dynamicRegistration': True, 'contentFormat': ['markdown', 'plaintext']}, 'completion': {'dynamicRegistration': True, 'completionItem': {'snippetSupport': True, 'deprecatedSupport': True, 'documentationFormat': ['markdown', 'plaintext'], 'tagSupport': {'valueSet': [1]}, 'resolveSupport': {'properties': ['detail', 'documentation', 'additionalTextEdits']}, 'insertReplaceSupport': True, 'insertTextModeSupport': {'valueSet': [2]}, 'labelDetailsSupport': True}, 'completionItemKind': {'valueSet': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25]}, 'insertTextMode': 2, 'completionList': {'itemDefaults': ['editRange', 'insertTextFormat', 'data']}}, 'signatureHelp': {'dynamicRegistration': True, 'contextSupport': True, 'signatureInformation': {'activeParameterSupport': True, 'documentationFormat': ['markdown', 'plaintext'], 'parameterInformation': {'labelOffsetSupport': True}}}, 'references': {'dynamicRegistration': True}, 'documentHighlight': {'dynamicRegistration': True}, 'documentSymbol': {'dynamicRegistration': True, 'hierarchicalDocumentSymbolSupport': True, 'symbolKind': {'valueSet': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26]}, 'tagSupport': {'valueSet': [1]}}, 'documentLink': {'dynamicRegistration': True, 'tooltipSupport': True}, 'formatting': {'dynamicRegistration': True}, 'rangeFormatting': {'dynamicRegistration': True, 'rangesSupport': True}, 'declaration': {'dynamicRegistration': True, 'linkSupport': True}, 'definition': {'dynamicRegistration': True, 'linkSupport': True}, 'typeDefinition': {'dynamicRegistration': True, 'linkSupport': True}, 'implementation': {'dynamicRegistration': True, 'linkSupport': True}, 'codeAction': {'dynamicRegistration': True, 'codeActionLiteralSupport': {'codeActionKind': {'valueSet': ['quickfix', 'refactor', 'refactor.extract', 'refactor.inline', 'refactor.rewrite', 'source.fixAll', 'source.organizeImports']}}, 'dataSupport': True, 'isPreferredSupport': True, 'resolveSupport': {'properties': ['edit']}}, 'rename': {'dynamicRegistration': True, 'prepareSupport': True, 'prepareSupportDefaultBehavior': 1}, 'colorProvider': {'dynamicRegistration': True}, 'publishDiagnostics': {'relatedInformation': True, 'tagSupport': {'valueSet': [1, 2]}, 'versionSupport': True, 'codeDescriptionSupport': True, 'dataSupport': True}, 'diagnostic': {'dynamicRegistration': True, 'relatedDocumentSupport': True}, 'selectionRange': {'dynamicRegistration': True}, 'foldingRange': {'dynamicRegistration': True, 'foldingRangeKind': {'valueSet': ['comment', 'imports', 'region']}}, 'codeLens': {'dynamicRegistration': True}, 'inlayHint': {'dynamicRegistration': True, 'resolveSupport': {'properties': ['textEdits', 'label.command']}}, 'semanticTokens': {'dynamicRegistration': True, 'requests': {'range': True, 'full': {'delta': True}}, 'tokenTypes': ['namespace', 'type', 'class', 'enum', 'interface', 'struct', 'typeParameter', 'parameter', 'variable', 'property', 'enumMember', 'event', 'function', 'method', 'macro', 'keyword', 'modifier', 'comment', 'string', 'number', 'regexp', 'operator', 'decorator', 'label'], 'tokenModifiers': ['declaration', 'definition', 'readonly', 'static', 'deprecated', 'abstract', 'async', 'modification', 'documentation', 'defaultLibrary'], 'formats': ['relative'], 'overlappingTokenSupport': False, 'multilineTokenSupport': True, 'augmentsSyntaxTokens': True}, 'callHierarchy': {'dynamicRegistration': True}, 'typeHierarchy': {'dynamicRegistration': True}}, 'workspace': {'applyEdit': True, 'didChangeConfiguration': {'dynamicRegistration': True}, 'executeCommand': {}, 'workspaceEdit': {'documentChanges': True, 'failureHandling': 'abort', 'changeAnnotationSupport': {'groupsOnLabel': False}}, 'workspaceFolders': True, 'symbol': {'dynamicRegistration': True, 'resolveSupport': {'properties': ['location.range']}, 'symbolKind': {'valueSet': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26]}, 'tagSupport': {'valueSet': [1]}}, 'configuration': True, 'codeLens': {'refreshSupport': True}, 'fileOperations': {'dynamicRegistration': True, 'willRename': True, 'didRename': True}, 'inlayHint': {'refreshSupport': True}, 'semanticTokens': {'refreshSupport': True}, 'diagnostics': {'refreshSupport': True}, 'didChangeWatchedFiles': {'dynamicRegistration': True, 'relativePatternSupport': True}}, 'window': {'showDocument': {'support': True}, 'showMessage': {'messageActionItem': {'additionalPropertiesSupport': True}}, 'workDoneProgress': True}}, 'initializationOptions': {'settings': {'codeAction': {'disableRuleComment': {'enable': True}, 'fixViolation': {'enable': True}}, 'configuration': {'lint': {'unfixable': ['F401']}}, 'configurationPreference': 'editorFirst', 'exclude': None, 'fixAll': True, 'format': {'preview': None}, 'lineLength': None, 'lint': {'enable': True, 'extendSelect': None, 'ignore': None, 'preview': None, 'select': None}, 'logLevel': 'error', 'organizeImports': True}}}
:: [17:42:11.119] <<< LSP-ruff (1) (duration: 81ms): {'capabilities': {'codeActionProvider': {'codeActionKinds': ['quickfix', 'source.fixAll.ruff', 'source.organizeImports.ruff', 'notebook.source.fixAll.ruff', 'notebook.source.organizeImports.ruff'], 'resolveProvider': True, 'workDoneProgress': True}, 'diagnosticProvider': {'identifier': 'Ruff', 'interFileDependencies': False, 'workDoneProgress': True, 'workspaceDiagnostics': False}, 'documentFormattingProvider': True, 'documentRangeFormattingProvider': True, 'executeCommandProvider': {'commands': ['ruff.applyFormat', 'ruff.applyAutofix', 'ruff.applyOrganizeImports', 'ruff.printDebugInformation'], 'workDoneProgress': False}, 'hoverProvider': True, 'notebookDocumentSync': {'notebookSelector': [{'cells': [{'language': 'python'}]}], 'save': False}, 'positionEncoding': 'utf-16', 'workspace': {'workspaceFolders': {'changeNotifications': True, 'supported': True}}, 'textDocumentSync': {'change': {'syncKind': 2}, 'didOpen': {}, 'didClose': {}}}, 'serverInfo': {'name': 'ruff', 'version': '0.14.14'}}
:: [17:42:11.120]  -> LSP-ruff initialized: {}
:: [17:42:11.125]  -> LSP-ruff textDocument/didOpen: <params with 252993 characters>
:: [17:42:11.128] --> LSP-ruff textDocument/diagnostic (2): {'textDocument': {'uri': 'file:///usr/local/workspace/lsp/lsp-python-types/generated/lsp_types.py'}, 'identifier': 'Ruff'}
:: [17:42:11.144] <-- LSP-ruff client/registerCapability (0): {'registrations': [{'id': 'ruff-server-watch', 'method': 'workspace/didChangeWatchedFiles', 'registerOptions': {'watchers': [{'globPattern': '**/.ruff.toml'}, {'globPattern': '**/ruff.toml'}, {'globPattern': '**/pyproject.toml'}]}}]}
:: [17:42:11.218] >>> LSP-ruff (0) (duration: 73ms): None
:: [17:42:11.221] <<< LSP-ruff (2) (duration: 92ms): {'items': [], 'kind': 'full'}
pyproject: [2026-01-31T16:42:16Z INFO  pyproject::server] Starting pyproject language server...
:: [17:42:16.167] --> pyproject initialize (1): {'processId': 93862, 'clientInfo': {'name': 'Sublime Text LSP', 'version': '2.8.1'}, 'rootUri': 'file:///usr/local/workspace/lsp/lsp-python-types', 'rootPath': '/usr/local/workspace/lsp/lsp-python-types', 'workspaceFolders': [{'name': 'lsp-python-types', 'uri': 'file:///usr/local/workspace/lsp/lsp-python-types'}], 'capabilities': {'general': {'regularExpressions': {'engine': 'ECMAScript'}, 'markdown': {'parser': 'marko', 'version': '2.2.2'}}, 'textDocument': {'synchronization': {'dynamicRegistration': True, 'didSave': True, 'willSave': True, 'willSaveWaitUntil': True}, 'hover': {'dynamicRegistration': True, 'contentFormat': ['markdown', 'plaintext']}, 'completion': {'dynamicRegistration': True, 'completionItem': {'snippetSupport': True, 'deprecatedSupport': True, 'documentationFormat': ['markdown', 'plaintext'], 'tagSupport': {'valueSet': [1]}, 'resolveSupport': {'properties': ['detail', 'documentation', 'additionalTextEdits']}, 'insertReplaceSupport': True, 'insertTextModeSupport': {'valueSet': [2]}, 'labelDetailsSupport': True}, 'completionItemKind': {'valueSet': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25]}, 'insertTextMode': 2, 'completionList': {'itemDefaults': ['editRange', 'insertTextFormat', 'data']}}, 'signatureHelp': {'dynamicRegistration': True, 'contextSupport': True, 'signatureInformation': {'activeParameterSupport': True, 'documentationFormat': ['markdown', 'plaintext'], 'parameterInformation': {'labelOffsetSupport': True}}}, 'references': {'dynamicRegistration': True}, 'documentHighlight': {'dynamicRegistration': True}, 'documentSymbol': {'dynamicRegistration': True, 'hierarchicalDocumentSymbolSupport': True, 'symbolKind': {'valueSet': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26]}, 'tagSupport': {'valueSet': [1]}}, 'documentLink': {'dynamicRegistration': True, 'tooltipSupport': True}, 'formatting': {'dynamicRegistration': True}, 'rangeFormatting': {'dynamicRegistration': True, 'rangesSupport': True}, 'declaration': {'dynamicRegistration': True, 'linkSupport': True}, 'definition': {'dynamicRegistration': True, 'linkSupport': True}, 'typeDefinition': {'dynamicRegistration': True, 'linkSupport': True}, 'implementation': {'dynamicRegistration': True, 'linkSupport': True}, 'codeAction': {'dynamicRegistration': True, 'codeActionLiteralSupport': {'codeActionKind': {'valueSet': ['quickfix', 'refactor', 'refactor.extract', 'refactor.inline', 'refactor.rewrite', 'source.fixAll', 'source.organizeImports']}}, 'dataSupport': True, 'isPreferredSupport': True, 'resolveSupport': {'properties': ['edit']}}, 'rename': {'dynamicRegistration': True, 'prepareSupport': True, 'prepareSupportDefaultBehavior': 1}, 'colorProvider': {'dynamicRegistration': True}, 'publishDiagnostics': {'relatedInformation': True, 'tagSupport': {'valueSet': [1, 2]}, 'versionSupport': True, 'codeDescriptionSupport': True, 'dataSupport': True}, 'diagnostic': {'dynamicRegistration': True, 'relatedDocumentSupport': True}, 'selectionRange': {'dynamicRegistration': True}, 'foldingRange': {'dynamicRegistration': True, 'foldingRangeKind': {'valueSet': ['comment', 'imports', 'region']}}, 'codeLens': {'dynamicRegistration': True}, 'inlayHint': {'dynamicRegistration': True, 'resolveSupport': {'properties': ['textEdits', 'label.command']}}, 'semanticTokens': {'dynamicRegistration': True, 'requests': {'range': True, 'full': {'delta': True}}, 'tokenTypes': ['namespace', 'type', 'class', 'enum', 'interface', 'struct', 'typeParameter', 'parameter', 'variable', 'property', 'enumMember', 'event', 'function', 'method', 'macro', 'keyword', 'modifier', 'comment', 'string', 'number', 'regexp', 'operator', 'decorator', 'label'], 'tokenModifiers': ['declaration', 'definition', 'readonly', 'static', 'deprecated', 'abstract', 'async', 'modification', 'documentation', 'defaultLibrary'], 'formats': ['relative'], 'overlappingTokenSupport': False, 'multilineTokenSupport': True, 'augmentsSyntaxTokens': True}, 'callHierarchy': {'dynamicRegistration': True}, 'typeHierarchy': {'dynamicRegistration': True}}, 'workspace': {'applyEdit': True, 'didChangeConfiguration': {'dynamicRegistration': True}, 'executeCommand': {}, 'workspaceEdit': {'documentChanges': True, 'failureHandling': 'abort', 'changeAnnotationSupport': {'groupsOnLabel': False}}, 'workspaceFolders': True, 'symbol': {'dynamicRegistration': True, 'resolveSupport': {'properties': ['location.range']}, 'symbolKind': {'valueSet': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26]}, 'tagSupport': {'valueSet': [1]}}, 'configuration': True, 'codeLens': {'refreshSupport': True}, 'fileOperations': {'dynamicRegistration': True, 'willRename': True, 'didRename': True}, 'inlayHint': {'refreshSupport': True}, 'semanticTokens': {'refreshSupport': True}, 'diagnostics': {'refreshSupport': True}, 'didChangeWatchedFiles': {'dynamicRegistration': True, 'relativePatternSupport': True}}, 'window': {'showDocument': {'support': True}, 'showMessage': {'messageActionItem': {'additionalPropertiesSupport': True}}, 'workDoneProgress': True}}, 'initializationOptions': {}}
:: [17:42:16.194] <<< pyproject (1) (duration: 27ms): {'capabilities': {'completionProvider': {'resolveProvider': False, 'triggerCharacters': ['[', '.', '=', '"', "'", ',']}, 'documentFormattingProvider': True, 'hoverProvider': True, 'textDocumentSync': {'change': {'syncKind': 2}, 'didOpen': {}, 'didClose': {}, 'save': {'includeText': False}}}, 'serverInfo': {'name': 'pyproject', 'version': '0.1.2'}}
:: [17:42:16.194]  -> pyproject initialized: {}
:: [17:42:16.216]  -> pyproject textDocument/didOpen: {'textDocument': {'uri': 'file:///usr/local/workspace/lsp/lsp-python-types/pyproject.toml', 'languageId': 'toml', 'version': 0, 'text': '[project]\nname = "lsp-types"\nversion = "1.0.0"\nlicense = "MIT"\nrequires-python = ">=3.8"\nclassifiers = [\n    "Programming Language :: Python :: 3",\n    "Operating System :: OS Independent",\n]\ndependencies = [\n    "typing-extensions",\n]\n\n[tool.pyright]\npythonVersion = "3.11"\nreportUnusedCallResult = "none"\n\n[[tool.pyright.executionEnvironments]]\nroot = "generated"\npythonVersion = "3.8"\n\n[tool.ruff]\nline-length = 120\n\n[tool.ruff.format]\nquote-style = "single"\n\n[tool.ruff.lint]\nextend-select = ["ALL"]\nignore = [\n    "COM812",  # https://docs.astral.sh/ruff/rules/missing-trailing-comma/\n    "C901",  # https://docs.astral.sh/ruff/rules/complex-structure/\n    "D100",  # https://docs.astral.sh/ruff/rules/undocumented-public-module/#undocumented-public-module-d100\n    "D101",  # https://docs.astral.sh/ruff/rules/undocumented-public-class/\n    "D102",  # https://docs.astral.sh/ruff/rules/undocumented-public-method/\n    "D103",  # https://docs.astral.sh/ruff/rules/undocumented-public-function/\n    "D104",  # https://docs.astral.sh/ruff/rules/undocumented-public-package/\n    "D203",  # https://docs.astral.sh/ruff/rules/missing-blank-line-after-summary/\n    "D205",  # https://docs.astral.sh/ruff/rules/missing-blank-line-after-summary/\n    "D212",  # https://docs.astral.sh/ruff/rules/multi-line-summary-first-line/\n    "D213",  # https://docs.astral.sh/ruff/rules/multi-line-summary-second-line/\n    "D400",  # https://docs.astral.sh/ruff/rules/missing-trailing-period/\n    "D415",  # https://docs.astral.sh/ruff/rules/missing-terminal-punctuation/\n    "I001",  # https://docs.astral.sh/ruff/rules/unsorted-imports/\n    "N801",  # https://docs.astral.sh/ruff/rules/invalid-class-name/\n    "PLR0911",  # https://docs.astral.sh/ruff/rules/too-many-return-statements/\n    "PLR0912",  # https://docs.astral.sh/ruff/rules/too-many-branches/\n    "S310",  # https://docs.astral.sh/ruff/rules/suspicious-url-open-usage/\n    "T201",  # https://docs.astral.sh/ruff/rules/print/\n    "TRY002",  # https://docs.astral.sh/ruff/rules/raise-vanilla-class/\n    "UP037"  # https://docs.astral.sh/ruff/rules/quoted-annotation/\n]\n\n[tool.ruff.lint.flake8-quotes]\ninline-quotes = "single"\n\n[tool.mypy]\n# ignore_missing_imports = true\n# check_untyped_defs = true\ndisallow_untyped_defs = true\nstrict_optional = true\n\n[tool.tox]\nlegacy_tox_ini = """\n[tox]\nenvlist = py3\nskipsdist = True\n\n[testenv]\n# mypy is disabled as it doesn\'t currently support cyclic definitions\n# https://github.com/python/mypy/issues/731\ndeps =\n    pyright==1.1.407\n    ruff==0.14.14\n    # mypy==0.971\ncommands =\n    pyright .\n    ruff check .\n    # mypy .\n"""\n'}}
:: [17:42:16.228] <-  pyproject window/logMessage: {'message': 'pyproject initialized', 'type': 3}
:: [17:42:16.243] <-  pyproject textDocument/publishDiagnostics: {'diagnostics': [], 'uri': 'file:///usr/local/workspace/lsp/lsp-python-types/pyproject.toml', 'version': 0}
:: [17:42:17.073]  -> pyproject textDocument/didSave: {'textDocument': {'uri': 'file:///usr/local/workspace/lsp/lsp-python-types/pyproject.toml'}}
:: [17:42:17.582]  -> LSP-ruff workspace/didChangeWatchedFiles: {'changes': [{'uri': 'file:///usr/local/workspace/lsp/lsp-python-types/pyproject.toml', 'type': <FileChangeType.Changed: 2>}]}
:: [17:42:17.588] <-- LSP-ruff workspace/diagnostic/refresh (1): None
:: [17:42:17.588] >>> LSP-ruff (1) (duration: 0ms): None
```

EDIT:

Actually...

> 'diagnosticProvider': {'identifier': 'Ruff', 'interFileDependencies': False, 'workDoneProgress': True, 'workspaceDiagnostics': False}

EDIT2:

But then the spec says:

> The [workspace/diagnostic/refresh](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic_refresh) request is sent from the server to the client. Servers can use it to ask clients to refresh all needed document and workspace diagnostics. This is useful if a server detects a project wide configuration change which requires a re-calculation of all diagnostics.

So we should still request non-workspace diagnostics?

jwortmann:
> So we should still request non-workspace diagnostics?

Yes, and in my experience so far we do that (already before this pull request). As I wrote above, I assume that no Python file was visible when you saved the `pyproject.toml`. Is that correct? In that case everything would work as expected: https://github.com/sublimelsp/LSP/pull/2727#issuecomment-3828695168 (we only need to refresh diagnostics right now for the files that are visible and can delay the request for other files to save unnecessary work)

rchl:
Yes, when saving `pyproject.toml` it was the one in front and I didn't use multiple columns.

But the diagnostics are not requested on activating the lsp_types.py file after that.

rchl:
Debugged it. It does trigger relevant methods but it does early return in https://github.com/sublimelsp/LSP/blob/4864ec4d4828730cdf550eae70f66d5ba48f63f7/plugin/session_buffer.py#L564-L565

jwortmann:
`forced_update` should be set in this case (iirc that was the reason for that parameter). If it isn't that looks like a new bug...

---

*Edit:* It looks like it's correct for the visible views at
https://github.com/sublimelsp/LSP/blob/eb34695a34b9a06c39a630b52a7048b6fde59b88/plugin/core/sessions.py#L2248
but missed to be set for the invisible views at
https://github.com/sublimelsp/LSP/blob/eb34695a34b9a06c39a630b52a7048b6fde59b88/plugin/documents.py#L407

So it was actually an older bug. But easy single line fix should do.

rchl:
I've realized now (after StabbarN reported it not working anymore for pyright) that workspace diagnostics now rely on `"diagnostics_mode": "workspace"` being set in client config. Is that really correct?

That initial purpose of that setting, even though it's called the way it's called, was really not related to LSP workspace diagnostics. I know the description was updated but I'm not sure it was the right move.

> Set to "workspace" (default is "open_files") to ignore diagnostics for files that are not within the project (window) folders. If project has no folders then this option has no effect and diagnostics are shown for all files. If the server supports pull diagnostics (diagnosticProvider), this setting also controls whether diagnostics are requested only for open files ("open_files"), or for all files in the project folders ("workspace").

jwortmann:
> I've realized now (after StabbarN reported it not working anymore for pyright) that workspace diagnostics now rely on `"diagnostics_mode": "workspace"` being set in client config. Is that really correct?

That is correct, but it has always been the case and it wasn't changed in this pull request. This config replaced the previous `hide_non_project_diagnostics` config. Iirc in the initial implementation I considered whether to keep that one and add another new config wether to clear diagnostics when a file gets closed (because with pull diagnostics that would be the responsibility of the client), but then decided to combine that into a single config option (ref. https://github.com/sublimelsp/LSP/pull/2225#discussion_r1166069590). There is no point in sending workspace diagnostic requests if diagnostics should be cleared / not shown in the panel for files that are not open (i.e. if set to "open_files"). The document diagnostic requests cover all files which are open anyway, and when a content change in the active file causes a change of diagnostics in a different file, then the server can (should) use the `relatedDocuments` property in the [diagnostic report](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#relatedFullDocumentDiagnosticReport) for that.

The only behavior change for workspace diagnostics in this pull request is that new workspace diagnostic requests (if none is open) are now only sent on file save, while before it was also re-triggered after buffer modifications. But that shouldn't be very relevant if the server follows the best practices and leaves the request open indefinitely, in which case there should be only a single request for the entire session, which gets sent after starting the server / provider registration (see e.g. https://github.com/microsoft/vscode-languageserver-node/issues/1669#issuecomment-3199531194). 

What is the reported issue with Pyright?

rchl:
> I have noticed that the LSP-pyright's `"python.analysis.diagnosticMode": "workspace",` setting doesn't work anymore. It's acting like `diagnosticMode` is set to `openFilesOnly` but I can see in LSP log that it's `workplace`. 

I have not verified the "anymore" part myself.

jwortmann:
By the way, Pyright doesn't even support workspace diagnostics:
```
:: [17:21:06.723] <-- LSP-pyright client/registerCapability (0): {'registrations': [{'id': 'a7608276-03f9-43bf-8fa9-88e8579b36c2', 'method': 'textDocument/diagnostic', 'registerOptions': {'interFileDependencies': True, 'workspaceDiagnostics': False, 'documentSelector': None, 'identifier': 'Pyright'}}]}
```
('workspaceDiagnostics': False)

The only two servers I currently know about, which support workspace diagnostics, are Luau server (Roblox Lua) and JETLS (Julia).

*Edit:* Ok I see now that Pyright sets `'workspaceDiagnostics': True` when `"python.analysis.diagnosticMode": "workspace"`

But still I think Pyright should either follow best practices that I linked above, or use `relatedDocuments` in the diagnostic reports.

rchl:
I would suggest that LSP-pyright always enables `"diagnostics_mode": "workspace"` so that pyright's own setting is functional but here is where the double-meaning of our option makes it problematic.

rchl:
> But still I think Pyright should either follow best practices that I linked above, or use `relatedDocuments` in the diagnostic reports.

I'm not sure I fully understand your suggestion but I'm also not sure if you have considered all options.

I'm thinking of just being able to see diagnostics for the whole project in the Diagnostics panel (or quick menu I guess).

rchl:
Another thing I've been noticing for a while. Previously the preview field for diagnostics quick panel had a bit more formatting / color:

Screenshot 2026-03-01 at 21 04 26

Now it's rather plain and looks kinda out of place (at least to me):

Screenshot 2026-03-01 at 21 05 17

Was that change on purpose?

I see why you would remove the text of the diagnostic since it's redundant but maybe the style could be improved a bit?

jwortmann:
> I'm thinking of just being able to see diagnostics for the whole project in the Diagnostics panel (or quick menu I guess).

I think it should do that already, it's just that pull diagnostics are only requested for the file you are currently typing in. If you want to see automatic updates for other files, the server can either use `relatedDocuments` (which seems to be the most obvious option exactly for that purpose), or it can send diagnostics either via publishDiagnostics notifications or via partial results progress for the workspace diagnostics request. The latter case requires `"diagnostics_mode": "workspace"` in the config. I don't remember exactly why I used `open_files` as the default, perhaps it was because at the time of implementing, there was basically no language server yet which supported workspace diagnostics (only the Luau server). We could consider to change that default value to `workspace`.

*Edit:* I think `open_files` was the default to preserve the previous behavior from the `hide_non_project_diagnostics` config.

---

> Was that change on purpose?

Yes, although I was experimenting with different things and wasn't so sure what exactly to put into the details pane at the bottom. The diagnostic kind is already shown as the colored letter, so having the color again at the bottom is kind of redundant too. It's quite subjective, but I'd rather prefer it without the colored background, because the padding is controlled by the theme (I'm not sure if it's even possible to eliminate the padding completely, or whether ST even forces some padding around the minihtml content there), so it doesn't look that nice with unpredictable horizontal/vertical paddings in my opinion.

rchl:
> > I'm thinking of just being able to see diagnostics for the whole project in the Diagnostics panel (or quick menu I guess).
> 
> I think it should do that already, it's just that pull diagnostics are only requested for the file you are currently typing in. If you want to see automatic updates for other files, the server can either use `relatedDocuments` (which seems to be the most obvious option exactly for that purpose),

I might have lost the context of the conversation already but my point was that one might want to see all project diagnostics initially on loading the project. For this we need to trigger workspace diagnostic request which we don't do unless `"diagnostics_mode": "workspace"` is set.

I don't think it would be typical for all diagnostics to be provided through `relatedDocuments` since that is meant for files that can be affected by changes in current file while there can be also diagnostics present in unrelated files.

jwortmann:
Iirc there were two other reasons for the behavior only to request workspace diagnostics if opt-in via `"diagnostics_mode": "workspace"`:

1. The initial implementation would request workspace diagnostics (if enabled) after typing, but with a longer debounce time. And I assume that workspace diagnostics are quite expensive to compute for the server, so it might be better to be opt-in, rather than run for every server. But now where new workspace diagnostics requests are only made on file save, this is probably not so important anymore.

2. With pull diagnostics, it is the responsibility of the client to clear diagnostics when appropriate (e.g. when file closes). Otherwise the diagnostics for many files that might have been temporarily opened some time ago would still show up in the diagnostics output panel in in "Goto Diagnostic". We have code to clear diagnostics when closing a file at https://github.com/sublimelsp/LSP/blob/eab1a177c0a4e2d8b9430e61fff5b0f5fb42259f/plugin/session_buffer.py#L272-L275
But this doesn't really work if there are also diagnostics reported for unopened files from workspace diagnostics requests. With `"diagnostics_mode": "workspace"` this isn't a problem, because in that case diagnostics for files not within the workspace folders are ignored, so they won't accumulate without limits.

PR #2728, opened 2026-01-17
Add docs for rename file

predragnikolic:
I will also add this video, once I get a link to the asset

[!Video](https://github.com/user-attachments/assets/573ebf9e-61b3-4db5-bef8-8e597126bead)

### DIFF
--- a/docs/src/features.md
+++ b/docs/src/features.md
@@ -130,7 +130,7 @@ Some language servers provide _global_ rename functionality as well. This packag
 
 Some language servers support updating imports on file rename.
 
-<video src="/videos/file-rename.mp4" controls></video>
+<video src="./videos/file-rename.mp4" controls></video>
 
 ## Code Actions
 
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -3,6 +3,7 @@ site_url: https://sublimelsp.github.io/LSP/
 repo_url: https://github.com/sublimelsp/LSP/
 edit_uri: edit/main/docs/src/
 docs_dir: 'src'
+use_directory_urls: false
 
 plugins:
   - search


predragnikolic:
Here is a direct link to the "File Rename" section in the preview link:
https://deploy-preview-2728--sublime-lsp.netlify.app/features/#rename-file

rchl:
> I will also add this video, once I get a link to the asset

It might be better to just include the video in the repo. Otherwise there is reliance on something separate from the repo itself which is maybe not ideal.

docs/src/features.md
```diff
@@ -126,6 +126,12 @@ Because a language server (usually) has an abstract syntax tree view of the file
 
 Some language servers provide _global_ rename functionality as well. This package will present a modal dialog to ask you to confirm to apply the changes if they span more than one file.
 
+## Rename File
+
+Some language servers support updating imports on file rename.
+
+<video src="/videos/file-rename.mp4" controls></video>
```
rchl: The url has to be relative, I believe (to work reliably, for example on github also).


```suggestion
<video src="videos/file-rename.mp4" controls></video>
```

predragnikolic: If I use a relative path the video src will be http(s)://.../features/videos/file-rename.mp4, (the video will point to the video file)

Keeping it as `/videos/file-ranme.mp4` should work,
the logo is also located as `/images/logo.svg` and it works -> https://lsp.sublimetext.io/images/logo.svg

rchl: Then it should probably start with `./videos/...` instead of `/videos`.

I know that `/videos` works for the documentation site but it's not reliable. If documentation would be at some subdirectory then it wouldn't. It also won't work when looking at md file at github.

predragnikolic: [207ef82](https://github.com/sublimelsp/LSP/pull/2728/commits/207ef82e214c679dfbccbe8daebf19f70fd666cd)

I wanted to embed the video like and be done with it
```
![](my video)
```
But that syntax is not supported.

Using a relative path with a raw HTML tag was a bit challenging. 

When `zensical build` is run, 
based on the `use_directory_urls` config value (by default it is true),
zensical will create folders for each md file (I didn't know that),
But because zensical cannot transform relative paths in raw HTML elements (like video/img src, https://github.com/zensical/zensical/issues/258)
and because a directory is created during build, 
the relative path is no longer correct. (As a workaround I could specify a relative like `../` instead of `./`, and the video will load, but then I discovered that I can just disable `use_directory_urls` that will keep the files on build on the same level as they were originally in the src.

rchl: That turned out much more involved than I thought.
Glad you found a solution but if it's gonna cause some issues then it would probably be better to just make sure that documentation site works.

predragnikolic: I did test a bit, 
This change didn't break existing links.

I noticed one broken link for the gif for https://lsp.sublimetext.io/features/#goto-symbol-in-project (but that was not caused by the changes in this PR)

PR #2732, opened 2026-01-22
fix outdated diagnostics

### LINKED ISSUE #2730: Inline decorations (diagnostics) not cleared if type fast
https://github.com/user-attachments/assets/bdec1c82-f452-49dd-ba15-115cdf3b8c00

In this screen record, I firstly typed `: ''` slowly, and it cleared the error decoration. Then I typed `:'.dts'` fast, the original error decoration is not cleared.

I'm not sure if this issue should be in `LSP-typescript` or `LSP`.

rchl:
Per @jwortmann's comment in https://github.com/sublimelsp/LSP/issues/2730#issuecomment-3783264277, rolled back the code a bit to allow `None` for diagnostics `version`. Then we convert that to `view.change_count()` later.

Also started a testsuite for diagnostics which tests this particular issue. Since the issue is very specific to timing and request order, it can look a bit magic.

Fixes #2730

### DIFF
--- a/tests/test_diagnostics.py
+++ b/tests/test_diagnostics.py
@@ -76,8 +76,7 @@ class DiagnosticsTestCase(TextDocumentTestCase):
         self.insert_characters('const x = 1')
         yield from self.await_message("textDocument/didChange")
         yield from self.await_client_notification(
-            "textDocument/publishDiagnostics",
-            create_test_diagnostics([('error', Point(0, 0), Point(0, 11))])
+            "textDocument/publishDiagnostics", create_test_diagnostics([('error', Point(0, 0), Point(0, 11))])
         )
         session_buffer = self.session.get_session_buffer_for_uri_async(TEST_FILE_URI)
         self.assertEqual(len(session_buffer.diagnostics), 1)


rchl:
Test fails on CI mac while it works perfectly fine on my mac :)

jwortmann:
I wonder how a proper fix for the underlying problem could look like. The diagnostics received from `textDocument/publishDiagnostics` would actually need to be applied to the previous document version, which I assume isn't really possible. This workaround here is fine if the diagnostics are cleared. But it could still result in "shifted" underlines for diagnostics that are on the same line that was modified in the meanwhile.

Maybe if a version mismatch is detected, filter out the diagnostics on the line with the caret position and apply only the rest? But perhaps it doesn't really matter much, because if there are diagnostics on the modified line, the server will probably send another `textDocument/publishDiagnostics` update soon. Also that might have unexpected results if the modification contains linebreaks...

rchl:
It really doesn't make sense to do any sophisticated handling for this corner case.
As you said, if the diagnostics don't match current document then server will send new ones soon enough.

jwortmann:
Then we should probably remove that check in case of publishDiagnostics alltogether? So basically ignore the optional `version` and use `None` instead of `get('version')`

rchl:
Hmm, good point.

plugin/core/sessions.py
```diff
@@ -1471,13 +1471,13 @@ def register_session_buffer_async(self, sb: SessionBufferProtocol) -> None:
         for data in self._registrations.values():
             data.check_applicable(sb, suppress_requests=True)
         if (uri := sb.get_uri()) and (diagnostics := self.diagnostics.diagnostics_by_document_uri(uri)):
-            self._publish_diagnostics_to_session_buffer_async(sb, diagnostics, sb.last_synced_version)
+            self._publish_diagnostics_to_session_buffer_async(sb, diagnostics)
```
rchl, line 1474: I was contemplating this change but in the end I think it doesn't really matter if we pass version here or not. I can't think of practical case where it would make a difference.

rchl:
I'm hoping this looks good now and will go ahead and merge this since it's a serious regression that should be addressed ASAP IMO.

PR #2733, opened 2026-01-22
Restrict package.json parser to http/https schemes

RinZ27:
I noticed that the `LspParseVscodePackageJson` command uses `urllib.request.urlopen` on a URL taken directly from the clipboard without validating the scheme. 

Since `urlopen` supports `file://`, this could allow unintended local file reads if a user is tricked into copying a malicious path (e.g., `file:///etc/passwd#package.json`). I've added a check to ensure only `http` and `https` schemes are allowed, which should be sufficient for the intended use case of fetching `package.json` from a remote source.

### DIFF
--- a/plugin/tooling.py
+++ b/plugin/tooling.py
@@ -127,10 +127,9 @@ class LspParseVscodePackageJson(sublime_plugin.ApplicationCommand):
     def run(self, base_package_name: str) -> None:
         # Download the contents of the URL pointing to the package.json file.
         base_url = sublime.get_clipboard()
-        try:
-            urllib.parse.urlparse(base_url)
-        except Exception:
-            sublime.error_message("The clipboard content must be a URL to a package.json file.")
+        parsed_url = urllib.parse.urlparse(base_url)
+        if parsed_url.scheme not in ("http", "https"):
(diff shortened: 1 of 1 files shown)


rchl:
I don't see how this could be exploited. The file would be read and then fail during json.loads that we do on the content.

RinZ27:
@rchl That's true for something like `/etc/passwd`, but if there's a sensitive JSON file on the local system (configs, credentials, etc.), `json.loads` will handle it just fine and dump the content into the new buffer. 

I think it's safer to just lock down the scheme to `http/https` anyway. It's a low-cost fix to prevent `urlopen` from accidentally touching the local filesystem or other schemes it doesn't need for this specific command.

predragnikolic:
Hello and thanks for effort of opening the PR. 
I took a look at the PR, but there is no really a need for this change.

This command is used only by LSP maintainers and I agree with what Rafal said in his comment.

RinZ27:
I understand your point about this being a maintainer-only tool. While I still think restricting the parser to network schemes is a solid safeguard against accidentally touching local files, I respect your decision on the project's risk priorities. I'll close this out to keep things clean. Thanks for the review anyway.

PR #2736, opened 2026-01-23
fix missing request duration in logs

rchl:
When matching request/response during duration logging, we have conflated server- and client-initiated requests.  Both server- and client-initiated requests can use their own request IDs so we need to treat those separately.

Logs from https://github.com/sublimelsp/LSP/issues/2731#issuecomment-3789916223 show the issue. Response to `textDocument/diagnostic (2)` doesn't have duration.


jwortmann:
Type annotations are not correct: `request_id: int`

Should be `int | str`

PR #2737, opened 2026-01-24
Fix broken links after changing use_directory_urls

predragnikolic:
I broke external links

When you visit 
https://github.com/supabase-community/postgres-language-server?tab=readme-ov-file#postgres-language-server

You will see this [Sublime Text](https://lsp.sublimetext.io/language_servers/#postgresql) link that points to 
https://lsp.sublimetext.io/language_servers/#postgresql (that link no longer works)
Screenshot 2026-01-24 at 17 10 03
(it would only work now if the link was https://lsp.sublimetext.io/language_servers#postgresql)

It is better to revert the change for use_directory_urls.

### DIFF
--- a/docs/src/features.md
+++ b/docs/src/features.md
@@ -130,7 +130,7 @@ Some language servers provide _global_ rename functionality as well. This packag
 
 Some language servers support updating imports on file rename.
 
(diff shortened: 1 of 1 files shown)


predragnikolic:
Thanks for the quick review

PR #2740, opened 2026-01-30
Refactor handling of progress notification for diagnostics

rchl:
Refactor to generalize handling of progress notifications for specific requests to not rely on internal details of the request handling.

(I did those changes some time ago and didn't properly retest yet)

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -2132,7 +2132,8 @@ class Session(TransportCallbacks):
             params['identifier'] = identifier
         self.workspace_diagnostics_pending_responses[identifier] = self.send_request_async(
             Request.workspaceDiagnostic(
-                params, on_partial_result=partial(self._on_workspace_diagnostics_partial_result_async, identifier)),
+                params,
+                on_partial_result=partial(self._on_workspace_diagnostics_async, identifier, reset_pending_response=False)),  # noqa: E501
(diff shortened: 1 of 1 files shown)


plugin/core/sessions.py
```diff
@@ -2129,24 +2130,21 @@ def _do_workspace_diagnostics_async(self, identifier: DiagnosticsIdentifier) ->
...
         if identifier is not None:
             params['identifier'] = identifier
-        partial_result_token = _PARTIAL_RESULT_PROGRESS_PREFIX + str(self.request_id + 1)
-        self.diagnostics.token_identifier_map[partial_result_token] = identifier
         self.workspace_diagnostics_pending_responses[identifier] = self.send_request_async(
-            Request.workspaceDiagnostic(params),
-            functools.partial(self._on_workspace_diagnostics_async, partial_result_token),
-            functools.partial(self._on_workspace_diagnostics_error_async, partial_result_token)
+            Request.workspaceDiagnostic(
+                params, on_partial_result=partial(self._on_workspace_diagnostics_partial_result_async, identifier)),
```
jwortmann: `partial` works with keyword arguments, so it should be possible to use `self._on_workspace_diagnostics_async` here too and drop the `self._on_workspace_diagnostics_partial_result_async` method.

jwortmann:
Nice idea to pass the callable directly. I've tested and it seems to work.

PR #2741, opened 2026-01-30
refactor(types): Type parameters and return value of Request

rchl:
Type params and return value of `Request` object. Typing `Request` ensures that all the `send_request` variants infer param/return types and enforce those for the handlers/chained promise. So more safety.

Also type `Notification` and `Response` but those don't benefit much from typing since there is no response to handle.

### DIFF
--- a/plugin/rename_file.py
+++ b/plugin/rename_file.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 from .core.edit import show_summary_message
+from .core.logging import debug
 from .core.open import open_file_uri
 from .core.promise import Promise
 from .core.protocol import Error, Notification, Request
@@ -129,7 +130,7 @@ class LspRenamePathCommand(LspWindowCommand):
         for response, weak_session in responses:
             if (session := weak_session()) and response:
                 if isinstance(response, Error):
-                    print(f'LSP: Error response during rename: {response}')
+                    debug(f'LSP: Error response during rename: {response}')
                     return
                 prompt_for_workspace_edits(session, response, label=label) \
                     .then(partial(self.on_prompt_for_workspace_edits_concluded, weak_session, response, label)) \


plugin/core/protocol.py
```diff
@@ -1,27 +1,37 @@
...
-from typing import Any, Callable, Generic, Iterable, Mapping, TypedDict, TypeVar, Union
+from typing import Any, Callable, Generic, Iterable, Mapping, TypedDict, TypeVar
 from typing_extensions import NotRequired
 import sublime
 
 INT_MAX = 2**31 - 1
 UINT_MAX = INT_MAX
 
+P = TypeVar('P', bound=None | bool | int | Uint | float | str | Mapping[str, Any] | Iterable[Any])
 R = TypeVar('R')
```
rchl: Those are defined in two places. Potentially could be imported from a single place but importing such generic names feels weird. Perhaps should be given more meaningful names and then imported? Or leave as is?

plugin/rename_file.py
```diff
@@ -112,17 +112,25 @@ def prompt_rename_async(self, file_rename: FileRename, label: str, rename_comman
...
+    def handle_rename_async(self, responses: list[tuple[WorkspaceEdit | None | Error, weakref.ref[Session]]],
                             label: str, rename_command_args: dict[str, Any]) -> None:
         for response, weak_session in responses:
             if (session := weak_session()) and response:
+                if isinstance(response, Error):
+                    print(f'LSP: Error response during rename: {response}')
+                    return
                 prompt_for_workspace_edits(session, response, label=label) \
                     .then(partial(self.on_prompt_for_workspace_edits_concluded, weak_session, response, label)) \
                     .then(lambda accepted: accepted and self.window.run_command('lsp_rename_path', rename_command_args))
```
rchl, line 137: Previously Error was not handled but in theory it can happen. I just interrupt rename if that happens but probably it's not the best way to do it...

plugin/core/protocol.py
```diff
@@ -1,27 +1,37 @@
...
+class JsonRpcPayload(TypedDict):
+    jsonrpc: str
+    error: NotRequired[ResponseError]
+    id: NotRequired[str | int]
+    method: NotRequired[str]
+    params: NotRequired[Any]
+    result: NotRequired[Any]
+
+
+class Request(Generic[P, R]):
```
rchl, line 38: We could have third generic for error result but it's the same in 99.9% of cases and it would not be possible to optionally provide it (at least until Python 3.13) so skipped it. Only `initialize` really used different error type but we haven't used anything from it so using standard ResponseError works for it too.

plugin/hierarchy.py
```diff
@@ -24,27 +24,26 @@
...
 class HierarchyDataProvider(TreeDataProvider):
 
     def __init__(
         self,
         weaksession: weakref.ref[Session],
-        request: Callable[..., Request],
-        request_handler: Callable[..., list[HierarchyItemWrapper]],
+        request: Callable[[Any], Request[Any, Any]],
+        request_handler: Callable[[Any], Any],
         root_elements: list[HierarchyItemWrapper]
```
rchl, line 47: Cheating with Any here since typing this properly is pretty much impossible without refactoring.

plugin/session_buffer.py
```diff
@@ -501,10 +500,8 @@ def _do_color_boxes_async(self, view: sublime.View, version: int) -> None:
             )
 
     def _on_color_boxes_async(self, view: sublime.View, response: list[ColorInformation]) -> None:
-        if response is None:  # Guard against spec violation from certain language servers
-            self._color_phantoms.update([])
-            return
-        phantoms = [lsp_color_to_phantom(view, color_info) for color_info in response]
+        # None-check guards against spec violation from certain language servers.
+        phantoms = [] if response is None else [lsp_color_to_phantom(view, color_info) for color_info in response]
```
jwortmann: I wonder which server is it and if we should really keep that workaround for a server bug in our code.
If so, could you add a `# pyright: ignore[...]` to silence the type checker (I haven't tested whether Pyright still reports a diagnostic for that after this change, but I assume so)?

plugin/core/views.py
```diff
@@ -358,27 +367,31 @@ def formatting_options(settings: sublime.Settings) -> dict[str, Any]:
     }
 
 
-def text_document_formatting(view: sublime.View) -> Request:
-    return Request("textDocument/formatting", {
+def text_document_formatting(view: sublime.View) -> Request[DocumentFormattingParams, list[TextEdit] | None]:
+    return Request.formatting({
         "textDocument": text_document_identifier(view),
         "options": formatting_options(view.settings())
-    }, view, progress=True)
```
jwortmann, line 365: I see this rather unrelated change to not provide work done progress tokens for formatting requests anymore. I don't have any opinions on that, but I wonder whether it was added because a server did report progress on formatting? Although formatting probably shouldn't take that long... so I guess it's okay to remove it.

plugin/session_buffer.py
rchl: https://github.com/sublimelsp/LSP/pull/2053#discussion_r972370161 (if github strips the anchor then you might need to copy paste the URL).

I think in that case the color boxes come from the css-language-service and I haven't checked if that was fixed. If not then it probably still affects LSP-vue too.

plugin/core/views.py
rchl: Actually it's still there, just moved to `Request.formatting` function.

rchl: BTW. The changes here were for consistency since other functions in view.py did use these Request helpers already.

plugin/session_buffer.py
```diff
@@ -501,10 +500,8 @@ def _do_color_boxes_async(self, view: sublime.View, version: int) -> None:
             )
 
     def _on_color_boxes_async(self, view: sublime.View, response: list[ColorInformation]) -> None:
-        if response is None:  # Guard against spec violation from certain language servers
-            self._color_phantoms.update([])
-            return
-        phantoms = [lsp_color_to_phantom(view, color_info) for color_info in response]
+        # None-check guards against spec violation from certain language servers.
+        phantoms = [] if response is None else [lsp_color_to_phantom(view, color_info) for color_info in response]  # pyright: ignore[reportUnnecessaryComparison]
```
rchl: using ruff which is smart enough not to complain about line length here here. flake8 does complain though. I suppose it's time to switch to ruff soon (after tailoring the rule set for our repo since it shows way too much now, even with my ignores)

plugin/core/views.py
jwortmann: Oh I see. Then it's fine and I misinterpreted the single-line diff here.

plugin/core/protocol.py
```diff
@@ -8,20 +8,30 @@
 INT_MAX = 2**31 - 1
 UINT_MAX = INT_MAX
 
+P = TypeVar('P', bound=Union[None, bool, int, Uint, float, str, Mapping[str, Any], Iterable[Any]])
 R = TypeVar('R')
```
rchl: I suppose `R` could rally have the same constrain as `P`

plugin/session_buffer.py
jwortmann: So has this bug been raised in the Volar repo? Apparently it's been years and it shouldn't be difficult to fix if that bug is still present, and if they don't care I'd say it's not the responsibility of clients to add workarounds for server bugs.
On the other hand if color boxes don't work with the Volar server wouldn't be a good result for Sublime users either...

rchl: Volar is deprecated so if the bug is still there, I'm not sure if someone will care to fix it. But AFAIR Volar uses MS's css-language-server for the style blocks so the bug would really be in https://github.com/microsoft/vscode-css-languageservice which is still used even by LSP-vue (successor of volar).

I guess I could check if the bug is still present but if someone is still using LSP-volar (with legacy projects) or LSP-css then it would still be an issue for those.

jwortmann: I have just tested with the LSP-css and it doesn't happen anymore, so I think you can remove the workaround.

rchl: I still see this with LSP-vue

Screenshot 2026-01-31 at 12 09 32

"empty" means None/null. I got it after removing the style block from vue file.

rchl: For reference, created https://github.com/volarjs/volar.js/issues/301

jwortmann: Okay. Then could you add the link to that issue in the comment at least?

plugin/core/protocol.py
```diff
@@ -2,26 +2,37 @@
...
 from typing import Any, Callable, Generic, Iterable, Mapping, TypedDict, TypeVar, Union
-from typing_extensions import NotRequired
+from typing_extensions import NotRequired, TypeAlias
 import sublime
 
 INT_MAX = 2**31 - 1
 UINT_MAX = INT_MAX
 
-R = TypeVar('R')
+LspPayload: TypeAlias = Union[None, bool, int, Uint, float, str, Mapping[str, Any], Iterable[Any]]
```
jwortmann: This should be exactly `LSPAny`, no?

plugin/core/protocol.py
```diff
@@ -2,26 +2,37 @@
...
 
 
-class Request(Generic[R]):
+class JsonRpcPayload(TypedDict):
+    jsonrpc: str
+    error: NotRequired[ResponseError]
+    id: NotRequired[str | int]
+    method: NotRequired[str]
+    params: NotRequired[Any]
+    result: NotRequired[Any]
```
jwortmann, line 25: I think it would be nicer to define
```python
JSONRpcPayload = Union[RequestMessage, ResponseMessage, NotificationMessage]
```
which are defined in the protocol (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#requestMessage) but for some reason are not in the generated types from https://github.com/microsoft/lsprotocol.

How about adding TypedDict definitions for these message types in our `plugin/core/protocol.py`?

plugin/core/protocol.py
```diff
@@ -30,148 +41,212 @@ def __init__(
...
+    def shutdown(cls) -> Request[None, None]:
         return Request("shutdown")
 
     def __repr__(self) -> str:
         return self.method + " " + str(self.params)
 
-    def to_payload(self, id: int) -> dict[str, Any]:
-        payload = {
+    def to_payload(self, request_id: int) -> JsonRpcPayload:
+        payload: JsonRpcPayload = {
```
jwortmann: This could be a `RequestMessage` if we add that type to `protocol.py`.

plugin/core/protocol.py
```diff
@@ -204,82 +279,81 @@ def from_exception(cls, ex: Exception) -> Error:
...
+    def exit(cls) -> Notification[None]:
         return Notification("exit")
 
     def __repr__(self) -> str:
         return self.method + " " + str(self.params)
 
-    def to_payload(self) -> dict[str, Any]:
-        payload = {
+    def to_payload(self) -> JsonRpcPayload:
+        payload: JsonRpcPayload = {
```
jwortmann: `NotificationMessage`

plugin/core/protocol.py
```diff
@@ -204,82 +279,81 @@ def from_exception(cls, ex: Exception) -> Error:
...
 
     __slots__ = ('request_id', 'result')
 
-    def __init__(self, request_id: Any, result: T) -> None:
+    def __init__(self, request_id: str | int, result: P) -> None:
         self.request_id = request_id
         self.result = result
 
-    def to_payload(self) -> dict[str, Any]:
+    def to_payload(self) -> JsonRpcPayload:
```
jwortmann: `ResponseMessage`

plugin/core/protocol.py
rchl: That triggers errors:

```
Type "InitializeParams" cannot be assigned to type variable "P@Request"
  Type "InitializeParams" is not assignable to upper bound "LSPAny" for type variable "P@Request"
    Type "InitializeParams" is not assignable to type "LSPAny"
      "InitializeParams" is not assignable to "Dict[str, LSPAny]"
      "InitializeParams" is not assignable to "List[LSPAny]"
      "InitializeParams" is not assignable to "str"
      "InitializeParams" is not assignable to "int"
      "InitializeParams" is not assignable to "float"
      "InitializeParams" is not assignable to "bool"
  ... (basedpyright)
```

We actually need `Mapping[str, Any], Iterable[Any]` here.

plugin/core/protocol.py
rchl: good idea!

plugin/core/protocol.py
rchl: More specifically we need `Iterable[Any]` because this works too: `Union[LSPAny, Iterable[Any]]`

rchl: I've asked Claude and it gave me the following (hopefully true) answer.

---

The issue is that **`TypedDict` is not considered compatible with `dict[str, T]`** by type checkers, even though at runtime they're both dictionaries. `TypedDict` is a **specific structured type** with known keys, while `dict[str, T]` is a **general mapping**.

## Solution 1: Cast to dict (Runtime)

```python
from typing import TypeAlias, TypedDict, cast

Uint = int

class InitializeParams(TypedDict):
    foo: int

LSPArray: TypeAlias = list[LSPAny]
LSPObject: TypeAlias = dict[str, LSPAny]
LSPAny: TypeAlias = LSPObject | LSPArray | str | int | Uint | float | bool | None

def f(p: LSPAny) -> None:
    pass

a: InitializeParams = {'foo': 1}

f(cast(dict, a))  # ✓ Works
```

## Solution 2: Include TypedDict Types in LSPAny

```python
from typing import TypeAlias, TypedDict, Mapping, Any

Uint = int

class InitializeParams(TypedDict):
    foo: int

LSPArray: TypeAlias = list[LSPAny]
LSPObject: TypeAlias = dict[str, LSPAny] | Mapping[str, Any]
LSPAny: TypeAlias = LSPObject | LSPArray | str | int | Uint | float | bool | None

def f(p: LSPAny) -> None:
    pass

a: InitializeParams = {'foo': 1}
f(a)  # ✓ Works
```

## Solution 3: Use Protocol or Mapping

```python
from typing import TypeAlias, TypedDict, Any
from collections.abc import Mapping

Uint = int

class InitializeParams(TypedDict):
    foo: int

LSPArray: TypeAlias = list[LSPAny]
LSPObject: TypeAlias = Mapping[str, Any]  # More permissive
LSPAny: TypeAlias = LSPObject | LSPArray | str | int | Uint | float | bool | None

def f(p: LSPAny) -> None:
    pass

a: InitializeParams = {'foo': 1}
f(a)  # ✓ Works
```

## Solution 4: Add type: ignore Comment

```python
from typing import TypeAlias, TypedDict

Uint = int

class InitializeParams(TypedDict):
    foo: int

LSPArray: TypeAlias = list[LSPAny]
LSPObject: TypeAlias = dict[str, LSPAny]
LSPAny: TypeAlias = LSPObject | LSPArray | str | int | Uint | float | bool | None

def f(p: LSPAny) -> None:
    pass

a: InitializeParams = {'foo': 1}

f(a)  # type: ignore[arg-type]
```

## Why This Happens

Type checkers like mypy and Pyright treat `TypedDict` as incompatible with `dict[str, T]` because:
- `TypedDict` has **fixed, known keys** with specific types
- `dict[str, T]` has **arbitrary string keys** all with the same type
- They're structurally different from a type-safety perspective

**Best Solution**: Use **Solution 3** with `Mapping[str, Any]` if you want maximum compatibility with various dict-like objects including TypedDicts.

rchl: Note that we can always update LSPAny in https://github.com/sublimelsp/lsp-python-types to be what we want if we think it makes sense.

jwortmann: I noticed that after https://github.com/sublimelsp/LSP/commit/5e19a9ed3e01ba58e41527c5120785fde361848a I would get a similar type incompatibility when using `session.execute_command`
https://github.com/sublimelsp/LSP-Tinymist/blob/4fe344f28e9a6fcae8cb1beda23c565b00612923/plugin.py#L267

So yes, I think making LSPAny more permissive for type checkers would be a good idea. I would also go with solution 3.

---

*Edit*: Although my use case that is linked above is the other way round compared to the example from Claude; the argument is typed LSPAny and the function input is `PreviewResult` (TypedDict). Would need to check if that works too.

rchl: It's a different issue.

plugin/core/protocol.py
```diff
@@ -1,27 +1,51 @@
...
+    error: NotRequired[ResponseError]
+
+
+class NotificationMessage(TypedDict):
+    jsonrpc: str
+    method: str
+    params: NotRequired[Any]
+
+
+JSONRpcPayload = Union[RequestMessage, ResponseMessage, NotificationMessage]
```
jwortmann: On second thought a better name might actually be `JSONRPCMessage`, in accordance with the message names from the specs (?)

plugin/hierarchy.py
```diff
@@ -223,9 +222,10 @@ class LspCallHierarchyCommand(LspHierarchyCommand):
 
     @classmethod
     def request(
-        cls, params: TextDocumentPositionParams, view: sublime.View
-    ) -> Request[list[CallHierarchyItem] | Error | None]:
-        return Request.prepareCallHierarchy(cast(CallHierarchyPrepareParams, params), view)
+        cls, params: CallHierarchyPrepareParams, view: sublime.View
+    ) -> Request[CallHierarchyPrepareParams, list[CallHierarchyItem] | None]:
+        print(params)
```
jwortmann: Leftover from debugging?

plugin/rename_file.py
```diff
@@ -112,17 +112,25 @@ def prompt_rename_async(self, file_rename: FileRename, label: str, rename_comman
...
+        self, weak_session: weakref.ref[Session], response: WorkspaceEdit | None | Error
+    ) -> tuple[WorkspaceEdit | None | Error, weakref.ref[Session]]:
+        return (response, weak_session)
+
+    def handle_rename_async(self, responses: list[tuple[WorkspaceEdit | None | Error, weakref.ref[Session]]],
                             label: str, rename_command_args: dict[str, Any]) -> None:
         for response, weak_session in responses:
             if (session := weak_session()) and response:
+                if isinstance(response, Error):
+                    print(f'LSP: Error response during rename: {response}')
```
jwortmann: I think I would be in favor to use the debug logging `debug` instead of `print` here, because afaik that is what is used in most other places too. If rename doesn't work, the error response will still be present in the server logs even if the `log_debug` setting is disabled.

jwortmann (approved):

PR #2743, opened 2026-01-31
Fix diagnostics request for not visible views after refresh

jwortmann:
Ensure that document diagnostics are requested after a `workspace/diagnostic/refresh` when a previously not visible view gets activated.

This should fix a bug discovered in https://github.com/sublimelsp/LSP/pull/2727#issuecomment-3828865080

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -404,7 +404,7 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
                 sb.do_code_lenses_async(self.view)
             if sb.document_diagnostic_needs_refresh:
                 sb.set_document_diagnostic_pending_refresh(needs_refresh=False)
-                sb.do_document_diagnostic_async(self.view, self.view.change_count())
+                sb.do_document_diagnostic_async(self.view, self.view.change_count(), forced_update=True)
             if sb.semantic_tokens.needs_refresh \
                     and (session_view := sb.session.session_view_for_view_async(self.view)) \
                     and session_view.get_request_flags() & RequestFlags.SEMANTIC_TOKENS:


rchl:
Just noticed a bug (don't have time to do proper issue).

Created an issue in JS file which created diagnostic from LSP-eslint.
Switched to terminal and ran `eslint --fix` which fixed the issue but after switching to ST the diagnostic is not removed.

It looks like we don't trigger diagnostic pull request because ST window is not in the foreground and we also don't trigger it when ST becomes focused.

It works fine when the file is fixed through terminal if ST window is in focus.

(I guess should be easily testable by just modifying file through another editor while ST is in background)

jwortmann:
There are some settings for that:
```jsonc
	// Whether to reload a file if it has changed on disk.
	"reload_file_on_change": true,

	// Always prompt before reloading a file, even if the file hasn't been
	// modified. If a file has unsaved changes, a prompt will always be shown.
	// Will only show if "reload_file_on_change" is true.
	"always_prompt_for_file_reload": false,

	// Whether to reload files when Sublime Text is not in focus.
	"reload_file_in_background": false,
```

It seems like you want to enable the `reload_file_in_background` setting? If you already have that enabled, I would guess that some event listener / TextChangeListener doesn't trigger in that case, which would probably be an upstream bug (I haven't tested that). But I don't think we have any check in the LSP code that would prevent diagnostic requests if the window is not in focus.

rchl:
The file changes on focus but no text change events are sent, as you've guessed. So ST bug.

rchl:
That said, there is [`on_reload` notification](https://www.sublimetext.com/docs/api_reference.html#sublime_plugin.TextChangeListener.on_reload) that triggers in this case and we even [have handler for it](https://github.com/sublimelsp/LSP/blob/f0a83a594c07b74b673c8b6e3af72dc9ca7fcbfa/plugin/session_buffer.py#L385-L391) so it's fixable from that handler.

jwortmann:
But that should already trigger document diagnostic request via `_on_after_change_async`:
https://github.com/sublimelsp/LSP/blob/f0a83a594c07b74b673c8b6e3af72dc9ca7fcbfa/plugin/session_buffer.py#L426

rchl:
Looks like, depending on the timing, sometimes diagnostics request triggers before reload event (probably from activation event) and then when reload comes later we ignore diagnostics request.

I guess `version` is already updated when initial diagnostics request triggers.

PR #2745, opened 2026-02-01
Fix dotted dict processing dots in second+ level keys

### LINKED ISSUE #2744: DottedDict processing dot when it shouldn't
**Describe the bug**

Our use of DottedDict is flawed. It's modeled after how VSCode handles its settings but we have not followed the same rules that VSCode uses which now causes compatibility issues.

For example LSP-yaml supports (currently undocumented) setting:

```
	"settings": {
		"files.associations": {
			".sublime-syntax": "yaml"
		},
	},
```

The dot in the `.sublime-syntax` key is processed by DottedDict breaking the structure of the setting.

Our problem is that we are processing dots at any nesting level of the dictionary while VSCode only processes the top-level keys (where top-level in this case would be the keys directly under the `settings` object).

Some example from VSCode settings:

```
    "editor.codeActionsOnSave": {
        "source.fixAll": "explicit"
    },
```

Here obviously the top level key is processed while the nested level key must not.

Though there are also VSCode settings like the one here:

```
    "[dockercompose]": {
        "editor.quickSuggestions": {
            "comments": false,
        },
    },
```

Where the "top-level" keys that get expanded are nested within `[dockercompose]`. This seems to be a special case for file-type-specific overrides but that does not apply in our case.

rchl:
Make `DottedDict` only process top-level keys as "dotted".

Some examples:

1. Same as before

```
DottedDict({
    "a.b": {
        "c": 1
    }
})
```

=>

```json
{
    "a": {
        "b": {
            "c": 1
        }
    }
}
```

2. Doesn't expand second level anymore

```
DottedDict({
    "a.b": {
        "c.d": 1
    }
})
```

=>

```json
{
    "a": {
        "b": {
            "c.d": 1
        }
    }
}
```

---

- `DottedDict.update(other)` also works like that - `other` only gets keys expanded at the first level.
- `DottedDict.set(path, other)` works as before - the whole `path` is expanded

(there is a bit of repetition in the code (`_merge()` and `set()` are pretty similar) and not sure about naming of `merge` since we sometimes call it "update".

Fixes #2744

### DIFF
--- a/tests/test_collections.py
+++ b/tests/test_collections.py
@@ -27,8 +27,8 @@ class DottedDictTests(TestCase):
 
     def test_does_not_expand_at_second_nesting_level(self) -> None:
         d = DottedDict({"editor.codeActionsOnSave": {"source.fixAll": "explicit"}})
+        self.verify(d, "editor.codeActionsOnSave", {"source.fixAll": "explicit"})
         self.assertIsNone(d.get("editor.codeActionsOnSave.source"))
-        self.assertIsNone(d.get("editor.codeActionsOnSave.source.fixAll"))
 
     def test_overwrite_int_with_dict(self) -> None:
         d = DottedDict({'foo.bar': 1})


jwortmann:
> Note that the second test shows that we cannot access individual nested keys with a dot in it anymore, but I guess that is fine.

Actually I think this could be solved as well by recursively testing the full key name first, before splitting parts from the right end (I haven't tried a concrete implementation).
https://github.com/sublimelsp/LSP/blob/dfd3a91cf5a4070e23a98f86800d26d706e51b1f/plugin/core/collections.py#L44-L51

Then if a language server asks for `"section": "editor.codeActionsOnSave.source.fixAll"` we could return the value `"explicit"`.

rchl:
Interesting point you are bringing up, I didn't think of "get" not working for those.

I'm curious how it works in VSCode when server asks. Only if there is an interoperability issues I would consider changing it.

rchl:
I see the same behavior in VSCode:

For a setting like:

```json
{
    "editor.codeActionsOnSave": {
        "source.fixAll.eslint": "always"
    },
}
```

here are relevant requests:

```

[Trace - 5:05:06 PM] Received request 'workspace/configuration - (2)'.
Params: {
    "items": [
        {
            "scopeUri": "file:///Users/rafal/Downloads/text.txt",
            "section": "editor.codeActionsOnSave"
        }
    ]
}


[Trace - 5:05:06 PM] Sending response 'workspace/configuration - (2)'. Processing request took 0ms
Result: [
    {
        "source.fixAll.eslint": "always"
    }
]


[Trace - 5:05:06 PM] Received request 'workspace/configuration - (3)'.
Params: {
    "items": [
        {
            "scopeUri": "file:///Users/rafal/Downloads/text.txt",
            "section": "editor.codeActionsOnSave.source"
        }
    ]
}


[Trace - 5:05:06 PM] Sending response 'workspace/configuration - (3)'. Processing request took 0ms
Result: [
    null
]


[Trace - 5:05:06 PM] Received request 'workspace/configuration - (4)'.
Params: {
    "items": [
        {
            "scopeUri": "file:///Users/rafal/Downloads/text.txt",
            "section": "editor.codeActionsOnSave.source.fixAll.eslint"
        }
    ]
}


[Trace - 5:05:06 PM] Sending response 'workspace/configuration - (4)'. Processing request took 0ms
Result: [
    null
]
```

jwortmann:
> ```
> [Trace - 5:05:06 PM] Received request 'workspace/configuration - (4)'.
> Params: {
>     "items": [
>         {
>             "scopeUri": "file:///Users/rafal/Downloads/text.txt",
>             "section": "editor.codeActionsOnSave.source.fixAll.eslint"
>         }
>     ]
> }
> ```

I assume you created this request just for testing, or is there a real server that uses a `section` like this?

Here is what I was thinking about in my comment above, how it might work:
(I have not tested if it really works, and probably it doesn't, but perhaps you get the idea what I had in my mind)

```python
    def get(self, path: str | None = None) -> Any:
        """
        Get a value from the dictionary.

        :param      path:  The path, e.g. foo.bar.baz, or None.

        :returns:   The value stored at the path, or None if it doesn't exist.
                    Note that this cannot distinguish between None values and
                    paths that don't exist. If the path is None, returns the
                    entire dictionary.
        """
        return self._d if path is None else self._get(self._d, path)

    @staticmethod
    def _get(d: dict[str, Any], path: str) -> Any:
        segments = path.split('.')
        for idx in range(len(segments), 0, -1):
            key = '.'.join(segments[:idx])
            if key in d:
                if rest := '.'.join(segments[idx:]):
                    val = d[key]
                    if isinstance(val, dict):
                        return DottedDict._get(val, rest)
                    return None
                return d[key]
        return None
```

rchl:
I've tested with a real server based on https://github.com/microsoft/vscode-extension-samples/tree/main/lsp-sample that I've just extended to ask for different sections.

I don't see any point in trying to make it work differently than how it works in VSCode.
Or, if having a more flexible DottedDict is useful than we can make it configurable how it works and have both behaviors.

That reminds me that I should verify that any server settings or custom plugin code currently doesn't rely on old behavior...

rchl:
Found two packages that would be affected. Those will be fixed.

tests/test_collections.py
```diff
@@ -25,6 +25,25 @@ def test_set_and_get(self) -> None:
         d.set("foo.bar.baz", {"some": "dict"})
         self.verify(d, "foo.bar.baz.some", "dict")
 
+    def test_does_not_expand_at_second_nesting_level(self) -> None:
+        d = DottedDict({"editor.codeActionsOnSave": {"source.fixAll": "explicit"}})
+        self.assertIsNone(d.get("editor.codeActionsOnSave.source"))
+        self.assertIsNone(d.get("editor.codeActionsOnSave.source.fixAll"))
```
jwortmann:
```suggestion
        self.verify(d, "editor.codeActionsOnSave", {"source.fixAll": "explicit"})
        self.assertIsNone(d.get("editor.codeActionsOnSave.source"))
```

I removed the second test because that is actually not really the desired/optimal behavior (even though it's the same behavior as VSCode).

PR #2747, opened 2026-02-03
Add support for `lsp_code_actions_on_format`

### LINKED ISSUE #2701: Run code actions on document format
With Sublime LSP you can configure `lsp_code_actions_on_save`, but I like to format the document (and run specific code actions) _without_ saving the file *and without triggering the code actions manually). Saving files triggers reloads of servers or file watchers and I do not always prefer this. Just trigger the `Format Document` which formats the document _and_ also sorts the imports, etc.

This [exact feature is also implemented in Zed which I really like ](https://zed.dev/docs/languages/javascript#eslint)(but of course I like Sublime Text more 😄  ).

I suggest to reuse the code already implemented for the `code_actions_on_save`, which I did in a POC: https://github.com/sublimelsp/LSP/compare/main...maerteijn:LSP:feature/lsp-actions-on-format 

Any chance this feature will be accepted? I'm happy to provide a PR (with the needed refactoring) if needed or help out in any way necessary.

mj026:
This PR allows the execution of (user defined) code actions before formatting when the `LSP: Format File` command is triggered by using a new setting `lsp_code_actions_on_format`. Functionality is the same (and now shared) with `lsp_code_actions_on_save`.

Implemented as described in issue https://github.com/sublimelsp/LSP/issues/2701.

Resolves #2701

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -23,7 +23,7 @@ from .save_command import SaveTask
 from abc import ABCMeta
 from abc import abstractmethod
 from functools import partial
-from typing import TYPE_CHECKING, Any, List, Tuple, Union
+from typing import TYPE_CHECKING, Any, List, Tuple, Union, final
 from typing import cast
 import sublime
 
@@ -219,9 +219,10 @@ def get_matching_on_save_kinds(
 
 
 class CodeActionTaskBase(SaveTask):
-    # Override and change to the specific setting in CodeActionOnSaveTask (lsp_code_actions_on_save)
-    # and CodeActionsOnFormatTask (lsp_code_actions_on_format)
-    SETTING_NAME: str = ""
+    """The base task that requests code actions from sessions and runs them."""
+
+    SETTING_NAME: str
+    """Override in your subclass to specific `lsp_code_actions_on_*` setting that should be read."""
 
     @classmethod
     def is_applicable(cls, view: sublime.View) -> bool:
@@ -267,23 +268,24 @@ class CodeActionTaskBase(SaveTask):
         Promise.all(tasks).then(lambda _: self._process_next_request(request_iterator))
 
 
+@final
 class CodeActionOnSaveTask(CodeActionTaskBase):
-    """
-    The main task that requests code actions from sessions and runs them.
+    """Request code actions from sessions before save and run them.
 
     The amount of time the task is allowed to run is defined by user-controlled setting. If the task
     runs longer, the native save will be triggered before waiting for results.
     """
+
     SETTING_NAME = "lsp_code_actions_on_save"
 
 
 LspSaveCommand.register_task(CodeActionOnSaveTask)
 
 
+@final
 class CodeActionsOnFormatTask(CodeActionTaskBase):
-    """
-    The main task that runs code actions on format.
-    """
+    """Run code actions on format."""
+
     SETTING_NAME = "lsp_code_actions_on_format"
 
 
--- a/plugin/save_command.py
+++ b/plugin/save_command.py
@@ -112,12 +112,10 @@ class LspTextCommandWithTasks(LspTextCommand):
         self._save_tasks_runner: SaveTasksRunner | None = None
 
     def on_before_tasks(self) -> None:
-        """Override this to execute code before the task handler starts"""
-        ...
+        """Override this to execute code before the task handler starts."""
 
     def on_tasks_completed(self, **kwargs: dict[str, Any]) -> None:
-        """Override this to execute code when all tasks are completed"""
-        ...
+        """Override this to execute code when all tasks are completed."""
 
     def _on_tasks_completed(self, **kwargs: dict[str, Any]) -> None:
         self._save_tasks_runner = None
--- a/plugin/formatting.py
+++ b/plugin/formatting.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 from ..protocol import TextDocumentSaveReason
 from ..protocol import TextEdit
+from .code_actions import CodeActionsOnFormatTask
 from .core.collections import DottedDict
 from .core.edit import apply_text_edits
 from .core.promise import Promise
@@ -16,7 +17,6 @@ from .core.views import text_document_formatting
 from .core.views import text_document_range_formatting
 from .core.views import text_document_ranges_formatting
 from .core.views import will_save_wait_until
-from .code_actions import CodeActionsOnFormatTask
 from .save_command import LspSaveCommand, SaveTask, LspTextCommandWithTasks
 from functools import partial
 from typing import Any, Callable, Iterator, List, Union


plugin/code_actions.py
```diff
@@ -270,9 +267,26 @@ def _handle_response_async(
...
 
 
+class CodeActionOnSaveTask(CodeActionTaskBase):
+    """
+    The main task that requests code actions from sessions and runs them.
+
+    The amount of time the task is allowed to run is defined by user-controlled setting. If the task
+    runs longer, the native save will be triggered before waiting for results.
+    """
+    CODE_ACTIONS = "lsp_code_actions_on_save"
```
rchl: Maybe `CODE_ACTIONS_SETTING_NAME`?

plugin/save_command.py
```diff
@@ -115,6 +111,15 @@ def __init__(self, view: sublime.View) -> None:
         super().__init__(view)
         self._save_tasks_runner: SaveTasksRunner | None = None
 
+    def _on_tasks_completed(self, kwargs: dict[str, Any]) -> None:
+        self._save_tasks_runner = None
+
```
rchl, line 123: Maybe also create `_on_before_tasks` and then the `LspSaveCommand.run` implementation could also be shared in the base and `SaveTasksRunner` could be a hidden implementation detail?

Also, since methods like `_on_tasks_completed` and `_on_before_tasks` are meant to be overridden now, I would remove the `_` prefix to make it look more public.

mj026:
Thank you for reviewing so fast! Will work on your suggestions the following day(s) and update the PR accordingly.

mj026:
Hi @rchl, I implemented all your suggestions: I like it now how `LspTextCommandWithTasks` became self containing regarding managing tasks  while `LspSaveCommand` and `LspFormatDocumentCommand` just contain specific implementation details for saving and formatting. Nice! 

Let me know if you need anything else for this PR to go through.

sublime-package.json
```diff
@@ -26,6 +26,16 @@
               },
               "markdownDescription": "A dictionary of code action identifiers that should be triggered on save.\n\nCode action identifiers are not officially standardized so refer to specific server's documentation on what is supported but `source.fixAll` is commonly used to apply fix-on-save code actions.\n\nThis option is also supported in syntax-specific settings and/or in the `\"settings\"` section of project files. Settings from all those places will be merged and more specific (syntax and project) settings will override less specific (from LSP or Sublime settings).\n\nOnly \"source.*\" actions are supported."
             },
+            "lsp_code_actions_on_format": {
```
rchl: Needs to be referenced here also: https://github.com/sublimelsp/LSP/blob/57388cae7d8d7487f8292d830093c83b48f15263/sublime-package.json#L474-L478

plugin/save_command.py
```diff
@@ -115,19 +111,34 @@ def __init__(self, view: sublime.View) -> None:
...
 
-    def run(self, edit: sublime.Edit, **kwargs: dict[str, Any]) -> None:
+    def on_before_tasks(self) -> None:
         if self._save_tasks_runner:
             self._save_tasks_runner.cancel()
-        sublime.set_timeout_async(self._trigger_on_pre_save_async)
-        self._save_tasks_runner = SaveTasksRunner(self, self._tasks, partial(self._on_tasks_completed, kwargs))
+
+    def on_tasks_completed(self, **kwargs: dict[str, Any]) -> None:
+        self._save_tasks_runner = None
```
rchl, line 121: I'm thinking that the implementation contract could be made even more obvious by having abstract methods in the base that would have to be implemented (and wouldn't have to call `super()` which can be error-prone).

I can push the changes later to avoid back and forth.

plugin/code_actions.py
jwortmann: I would probably name it just `SETTING_NAME`, but I guess it isn't too important.

plugin/core/types.py
```diff
@@ -272,6 +273,7 @@ def r(name: str, default: bool | int | str | list | dict) -> None:
         r("log_debug", False)
         r("log_max_size", 8 * 1024)
         r("lsp_code_actions_on_save", {})
+        r("lsp_code_actions_on_format", {})
```
jwortmann, line 275: These seem to be ordered alphabetically (also for the variable name above).

plugin/code_actions.py
rchl: I would be fine with that too.

It has to be a specifically structured setting so I was thinking of more specific name but it's not easy.

plugin/save_command.py
mj026: I get your point. I’ll see what I can think of to make this a bit more robust.

plugin/core/types.py
mj026: Ok, will fix this.

plugin/code_actions.py
mj026: I’ll change it into `SETTING_NAME`, feels like the most obvious name for it.

mj026: Changed it now to `SETTING_NAME`

sublime-package.json
mj026: Fixed[ the reference ](https://github.com/mj026/LSP/blob/7a339f89b55cb4de5097c3319f77d07631b000dd/sublime-package.json#L474-L479)now too 👍

plugin/core/types.py
mj026: [Fixed this now](https://github.com/mj026/LSP/blob/7a339f89b55cb4de5097c3319f77d07631b000dd/plugin/core/types.py#L229-L230)

plugin/save_command.py
mj026: [So this seems for me the most simple solution](https://github.com/mj026/LSP/blob/7a339f89b55cb4de5097c3319f77d07631b000dd/plugin/save_command.py#L102-L131) The `on_before_tasks` and `on_tasks_completed` are now optional. When you do not implement them or when you do _and forget_ `super()` nothing bad happens: The initialise / cleanup handling for `SaveTasksRunner` is now done outside these handlers:
```python
class LspTextCommandWithTasks(LspTextCommand):
    _tasks: list[type[SaveTask]] = []

    @classmethod
    def register_task(cls, task: type[SaveTask]) -> None:
        assert task not in cls._tasks
        cls._tasks.append(task)

    def __init__(self, view: sublime.View) -> None:
        super().__init__(view)
        self._save_tasks_runner: SaveTasksRunner | None = None

    def on_before_tasks(self) -> None:
        """Override this to execute code before the task handler starts"""
        ...

    def on_tasks_completed(self, **kwargs: dict[str, Any]) -> None:
        """Override this to execute code when all tasks are completed"""
        ...

    def _on_tasks_completed(self, **kwargs: dict[str, Any]) -> None:
        self._save_tasks_runner = None
        self.on_tasks_completed(**kwargs)

    def run(self, edit: sublime.Edit, **kwargs: dict[str, Any]) -> None:
        if self._save_tasks_runner:
            self._save_tasks_runner.cancel()
        self.on_before_tasks()
        self._save_tasks_runner = SaveTasksRunner(self, self._tasks, partial(self._on_tasks_completed, **kwargs))
        self._save_tasks_runner.run()
```

mj026:
Implemented the suggestions, cleaned up the commits and rebased the branch on upstream `main`.

plugin/save_command.py
rchl: perfect. I've only pushed some minor style changes

rchl:
One question on my mind is whether `lsp_format_on_save` functionality should trigger this too? It doesn't right now and I'm not sure it should...

rchl:
I have some refactoring in mind as with the new functionality some names no longer make sense (SaveTask, SaveTasksRunner). I'll refactor that after this is merge.

mj026:
> One question on my mind is whether `lsp_format_on_save` functionality should trigger this too? It doesn't right now and I'm not sure it should...

Good question and it crossed my mind: I’m slightly feeling it shouldn’t to make it not too complicated. But I can imagine this could be a useful thing. However, you can configure to run code actions on save if you want to do this. So how it is implemented now is the most flexible way for the end users.

mj026:
> I have some refactoring in mind as with the new functionality some names no longer make sense (SaveTask, SaveTasksRunner). I'll refactor that after this is merge.

I was thinking about this too but didn’t want to change too much in this PR. The advantage is that `SaveTask` and `SaveTasksRunner` are now hidden implementations so it probably is easy to do!

mj026:
> > One question on my mind is whether `lsp_format_on_save` functionality should trigger this too? It doesn't right now and I'm not sure it should...
> 
> Good question and it crossed my mind: I’m slightly feeling it shouldn’t to make it not too complicated. But I can imagine this could be a useful thing. However, you can configure to run code actions on save if you want to do this. So how it is implemented now is the most flexible way for the end users.

So I just gave the Zed editor another spin and checked there if it triggers the code actions configured with `code_actions_on_format` when `format_on_save` is enabled. And it does, which makes sense in a way too.

I'm happy to add this functionality to this PR if you think this would be the better choice. I'm also happy to do this in a separate pull request. I'm also perfectly fine to leave it as is it right now.

rchl:
> > > One question on my mind is whether `lsp_format_on_save` functionality should trigger this too? It doesn't right now and I'm not sure it should...
> > 
> > 
> > Good question and it crossed my mind: I’m slightly feeling it shouldn’t to make it not too complicated. But I can imagine this could be a useful thing. However, you can configure to run code actions on save if you want to do this. So how it is implemented now is the most flexible way for the end users.
> 
> So I just gave the Zed editor another spin and checked there if it triggers the code actions configured with `code_actions_on_format` when `format_on_save` is enabled. And it does, which makes sense in a way too.
> 
> I'm happy to add this functionality to this PR if you think this would be the better choice. I'm also happy to do this in a separate pull request. I'm also perfectly fine to leave it as is it right now.

I think it makes sense to match Zed's behavior since we are basically copying its feature. That said, it can be done in a follow up PR. I'll do my refactor first since I've already prepared it based on current code.

PR #2751, opened 2026-02-07
Refactor SaveTask -> LspTask

rchl:
After #2747 the `SaveTask` is also used during "format document" so its name was no longer accurate. Rename to `LspTask`.

Also rename `SaveTasksRunner` -> `TaskRunner` and move all task-related code to its own file instead of being in `save_command.py`

Also another notable change. Instead of `LspTextCommandWithTasks` having a `register_task` method, it now has a `tasks()` abstract property that has to be implemented in the subclass. The previous approach turned out to be messy and made it hard to follow where tasks were being registered.

@mj026

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,1 +1,1 @@
-zensical==0.0.18
+zensical==0.0.21
--- a/docs/src/customization.md
+++ b/docs/src/customization.md
@@ -208,7 +208,7 @@ For example, to add a custom rule for `Mariana` color scheme, select `UI: Custom
     "rules": [
         {
             "scope": "markup.unnecessary.lsp",
-            "foreground": "color(rgb(255, 255, 255) alpha(0.4))",
+            "foreground": "rgba(255, 255, 255, 0.4)",
             "background": "#00000101"
         }
     ]
--- a/plugin/core/constants.py
+++ b/plugin/core/constants.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 from ...protocol import CodeActionKind
 from ...protocol import CompletionItemKind
 from ...protocol import DiagnosticSeverity
+from ...protocol import DiagnosticTag
 from ...protocol import DocumentHighlightKind
 from ...protocol import LanguageKind
 from ...protocol import MessageType
@@ -214,6 +215,11 @@ SUBLIME_KIND_SCOPES: dict[SublimeKind, str] = {
     sublime.KIND_NAVIGATION: "entity.name.definition | entity.name.label | entity.name.section",
     sublime.KIND_MARKUP: "entity.other.attribute-name | entity.name.tag | meta.toc-list.id.html",
     sublime.KIND_VARIABLE: "entity.name.constant | constant.other | support.constant | variable.other | variable.parameter | variable.other.member | variable.other.readwrite.member"  # noqa: E501
+}
+
+DIAGNOSTIC_TAG_SCOPES: dict[DiagnosticTag, str] = {
+    DiagnosticTag.Unnecessary: 'markup.unnecessary.lsp',
+    DiagnosticTag.Deprecated: 'markup.deprecated.lsp'
 }
 
 DOCUMENT_HIGHLIGHT_KIND_SCOPES: dict[DocumentHighlightKind, str] = {
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -10,6 +10,7 @@ from ...protocol import Command
 from ...protocol import Diagnostic
 from ...protocol import DiagnosticRelatedInformation
 from ...protocol import DiagnosticSeverity
+from ...protocol import DiagnosticTag
 from ...protocol import DidChangeTextDocumentParams
 from ...protocol import DidCloseTextDocumentParams
 from ...protocol import DidOpenTextDocumentParams
@@ -82,7 +83,7 @@ class DiagnosticSeverityData:
 
     def __init__(self, severity: int) -> None:
         self.regions: list[sublime.Region] = []
-        self.regions_with_tag: dict[int, list[sublime.Region]] = {}
+        self.regions_with_tag: dict[DiagnosticTag, list[sublime.Region]] = {}
         self.annotations: list[str] = []
         _, _, self.scope, self.icon, _, _ = DIAGNOSTIC_SEVERITY[severity - 1]
         if userprefs().diagnostics_gutter_marker != "sign":
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 from ..protocol import Command
-from ..protocol import DiagnosticTag
 from ..protocol import DocumentUri
 from .core.active_request import ActiveRequest
+from .core.constants import DIAGNOSTIC_TAG_SCOPES
 from .core.constants import DOCUMENT_HIGHLIGHT_KIND_NAMES
 from .core.constants import HOVER_ENABLED_KEY
 from .core.constants import RegionKey
@@ -25,8 +25,6 @@ from weakref import WeakValueDictionary
 import html
 import itertools
 import sublime
-
-DIAGNOSTIC_TAG_VALUES: list[int] = [v for (k, v) in DiagnosticTag.__dict__.items() if not k.startswith('_')]
 
 
 class TagData:
@@ -296,12 +294,6 @@ class SessionView:
     def diagnostics_key(self, severity: int, multiline: bool) -> str:
         return "lsp{}d{}{}".format(self.session.config.name, "m" if multiline else "s", severity)
 
-    def diagnostics_tag_scope(self, tag: int) -> str | None:
-        for k, v in DiagnosticTag.__dict__.items():
-            if v == tag:
-                return f'markup.{k.lower()}.lsp'
-        return None
-
     def present_diagnostics_async(self, is_view_visible: bool) -> None:
         self._redraw_diagnostics_async()
         if listener := self.listener():
@@ -325,14 +317,14 @@ class SessionView:
     ) -> None:
         ICON_FLAGS = sublime.RegionFlags.HIDE_ON_MINIMAP | sublime.RegionFlags.DRAW_NO_FILL | sublime.RegionFlags.DRAW_NO_OUTLINE | sublime.RegionFlags.NO_UNDO  # noqa: E501
         key = self.diagnostics_key(severity, multiline)
-        tags = {tag: TagData(f'{key}_tags_{tag}') for tag in DIAGNOSTIC_TAG_VALUES}
+        tags = {tag: TagData(f'{key}_tags_{tag}') for tag in DIAGNOSTIC_TAG_SCOPES}
         data = self._session_buffer.diagnostics_data_per_severity.get((severity, multiline))
         if data and severity <= max_severity_level:
             non_tag_regions = data.regions
             for tag, regions in data.regions_with_tag.items():
-                tag_scope = self.diagnostics_tag_scope(tag)
+                tag_scope = DIAGNOSTIC_TAG_SCOPES[tag]
                 # Trick to only add tag regions if there is a corresponding color scheme scope defined.
-                if tag_scope and 'background' in self.view.style_for_scope(tag_scope):
+                if 'background' in self.view.style_for_scope(tag_scope):
                     tags[tag].regions = regions
                     tags[tag].scope = tag_scope
                 else:


mj026:
https://github.com/sublimelsp/LSP/blob/0db73ea4759eb2779b57a878c402089f567ab081/plugin/lsp_task.py#L105-L108

https://github.com/sublimelsp/LSP/blob/0db73ea4759eb2779b57a878c402089f567ab081/plugin/save_command.py#L21-L26

Much cleaner like this, I like it 👍 

PS, one suggestion: You might wanto to rename the private variable: `_save_tasks_runner` to `_tasks_runner` in the `LspTextCommandWithTasks` class:
https://github.com/sublimelsp/LSP/blob/0db73ea4759eb2779b57a878c402089f567ab081/plugin/lsp_task.py#L112

predragnikolic (commented):
I haven't tested it but the new code looks/reads nice.

plugin/save_command.py
```diff
@@ -1,154 +1,49 @@
...
     """
+
+    @property
+    @override
+    def tasks(self) -> list[type[LspTask]]:
+        return [
+            CodeActionOnSaveTask,
+            FormattingTask,
+            WillSaveWaitTask,
+        ]
```
predragnikolic, line 26: I do like that the order of tasks is clearly defined.
fix code actions -> format code -> will save

I ended up in a loop hole thinking about whether it makes sense to run code first or to format first... I did not come to a conclusion. The current order is fine.

rchl: Those can do incompatible changes but not sure which one should have the final say. I might be leaning a bit towards code actions being last after all...

mj026: I would say to keep it like  `code actions` -> `format code` with the following reasoning:

1) In most of the cases (let's say relative simple code actions like import sorting) I don't think it will introduce any problems one way or another: `code actions` -> `format code` or  `format code` -> `code_actions`. But I expect when a formatter is done, the code is 100% formatted according to the formatter rules. When you apply code actions _after_ formatting, it could happen the formatted code is not according the formatter rules which would be, in my opinion, very confusing. In the python world it is common to [adjust the formatter to the linting rules](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#custom-configuration) and not the other way around, so I'm slightly leaning to  `code actions` -> `format code`.

2) The current implementation in the Zed editor also run the code actions before the formatter: https://github.com/zed-industries/zed/blob/44364ebd0362e01395db04797d684d496fc5764a/crates/project/src/lsp_store.rs#L1652-L1653. I do not say the Zed editor should be our golden standard, but I'd say we should have very good reasons to deviate from it, at least with some practical examples where it is preferred to do it the other way around.

mj026:
@rchl  This PR is a perfect opportunity to fix a `CodeAction` / `CodeActions` inconsistency I (accidentally) introduced 😅  :
https://github.com/sublimelsp/LSP/blob/6241b580d90194c9f86ecda85ea0080f030136dc/plugin/code_actions.py#L220
https://github.com/sublimelsp/LSP/blob/6241b580d90194c9f86ecda85ea0080f030136dc/plugin/code_actions.py#L273
https://github.com/sublimelsp/LSP/blob/6241b580d90194c9f86ecda85ea0080f030136dc/plugin/code_actions.py#L284

(I would opt for ~~`CodeAction`~~ `CodeActions` everywhere as the setting is also named `code_actions` and you can define multiple code actions to be applied).

rchl:
> (I would opt for `CodeAction` everywhere

Based on the rest of the sentence I guess you meant `CodeActions` here?

plugin/save_command.py
jwortmann: Iirc something related to this was discussed before and we came to the conclusion that the current code action on save handling is very fragile and might lead to broken files (I think there was a bug report somewhere), because it just runs all code actions in the response, without purging changes and making new requests again inbetween. The current implementation can only work if all edits are single-line edits (or the added and removed lines cancel each other out exactly) and they don't intersect. Proper handling would require to make new requests again after each change.

Regarding code action & formatting ordering; in my opinion it seems to only makes sense to do code actions first and formatting last, since formatting should only affect whitespace, so there shouldn't be any new code actions available after formatting if the code fixes / import ordering was already done before that.

rchl: > Iirc something related to this was discussed before and we came to the conclusion that the current code action on save handling is very fragile and might lead to broken files (I think there was a bug report somewhere), because it just runs all code actions in the response, without purging changes and making new requests again inbetween.

I would say that any real issues were fixed (https://github.com/sublimelsp/LSP/pull/2674 and https://github.com/sublimelsp/LSP/pull/2707) and the remaining comment (see the one in the first PR) is fine since we are compatible with VSCode.

> Regarding code action & formatting ordering; in my opinion it seems to only makes sense to do code actions first and formatting last, since formatting should only affect whitespace, so there shouldn't be any new code actions available after formatting if the code fixes / import ordering was already done before that.

eslint's code actions can affect white space too, depending on configured rules. But in any case it looks like we have agreement that current logic is the way to go.

mj026:
> > (I would opt for `CodeAction` everywhere
> 
> Based on the rest of the sentence I guess you meant `CodeActions` here?

Now I did it again (corrected it) 🙃  YES, I mean `CodeActions`, the plural version.

mj026:
Another suggestion: Rename the `FormattingTask` to `FormatOnSaveTask` as this is task is implicitly bound to the `lsp_format_on_save` setting and makes this a bit more clear.

mj026:
Last but not least: To fully implement the Zed behaviour regarding `lsp_code_actions_on_format` we also need to trigger these code actions when `lsp_format_on_save` is enabled as discussed here: https://github.com/sublimelsp/LSP/pull/2747#issuecomment-3864122586.

When your refactoring is ready, it could be implemented like so:

```python
class LspSaveCommand(LspTextCommandWithTasks):
    ...

    @property
    @override
    def tasks(self) -> list[LspTask]:
        tasks = [
            CodeActionOnSaveTask,
            FormattingTask,
            WillSaveWaitTask,
        ]
        # Enable the CodeActionsOnFormatTask task when the FormattingTask is applicable
        if FormattingTask.is_applicable(view=self.view):
            tasks.insert(1, CodeActionsOnFormatTask)

        return tasks
```

If you think this approach is the way to go I could prepare a (separate) PR for it if eligible.

rchl:
> Last but not least: To fully implement the Zed behaviour regarding `lsp_code_actions_on_format` we also need to trigger these code actions when `lsp_format_on_save` is enabled as discussed here: [#2747 (comment)](https://github.com/sublimelsp/LSP/pull/2747#issuecomment-3864122586).
> 
> When your refactoring is ready, it could be implemented like so:
> 
> ```python
> class LspSaveCommand(LspTextCommandWithTasks):
>     ...
> 
>     @property
>     @override
>     def tasks(self) -> list[LspTask]:
>         tasks = [
>             CodeActionOnSaveTask,
>             FormattingTask,
>             WillSaveWaitTask,
>         ]
>         # Enable the CodeActionsOnFormatTask task when the FormattingTask is applicable
>         if FormattingTask.is_applicable(view=self.view):
>             tasks.insert(1, CodeActionsOnFormatTask)
> 
>         return tasks
> ```
> 
> If you think this approach is the way to go I could prepare a (separate) PR for it if eligible.

I suppose that makes sense...
Would mean that potentially code actions would be run both from `CodeActionOnSaveTask` and `CodeActionsOnFormatTask` but I'm not sure what kind of settings the user of `CodeActionsOnFormatTask` would typically have...

You can create PR for it.

predragnikolic:
If such PR is created later for `lsp_format_on_save`, we can also think about `lsp_format_on_paste`.

Note that the LSP spec also has:
- format on type (not implemented by LSP for ST - https://github.com/sublimelsp/LSP/pull/179)
- And format range - if we format a range, should we only include code actions for that formatted range when requesting code actions?

When thinking about running code actions on format on save, 
it would be good to think about these other things as well, and draw conclusions.

mj026:
> Would mean that potentially code actions would be run both from `CodeActionOnSaveTask` and `CodeActionsOnFormatTask` 

Yes indeed.

> but I'm not sure what kind of settings the user of `CodeActionsOnFormatTask` would typically have...

Well, me as the feature requester for this: I would run code actions `on_format` which are normally defined in `on_save`. When enabling `format_on_save` it would be nice to have these code actions applied without needing any extra configuration (add duplicate code actions via the `code_actions_on_save` setting).

I think there will be three situations "out in the field":
1. Configured `code_actions_on_format` without`format_on_save` (me and others who don't like actions / formatting + on save, we want to manually trigger this with the `LSP:Format Document` command)
2. Configured `code_actions_on_format` with`format_on_save` (some users)
3. Configured `code_actions_on_save` + `format_on_save` (most users)

I will make a PR for running the code actions defined in `code_actions_on_format` when `format_on_save` is enabled and discuss / see from there. I will go with the [implementation as described earlier](https://github.com/sublimelsp/LSP/pull/2751#issuecomment-3872568376) for now as it has the least implementation impact.

mj026:
> If such PR is created later for `lsp_format_on_save`, we can also think about `lsp_format_on_paste`.
> 
> Note that the LSP spec also has:
> 
>     * format on type (not implemented by LSP for ST - [Implement textDocument/onTypeFormatting #179](https://github.com/sublimelsp/LSP/pull/179))
> 
>     * And format range - if we format a range, should we only include code actions for that formatted range when requesting code actions?
> 
> 
> When thinking about running code actions on format on save, it would be good to think about these other things as well, and draw conclusions.

I think most of these imply the same thing: code formatting (and triggering of code actions) for a _selection of a document_ and not the _whole document_.

This differs from the current implementations for `format_on_save`, `code_actions_on_save` and `code_actions_on_format` as these are  only implemented_ for the whole document (save file and the `LSP: Format File` LSP action)
- https://github.com/sublimelsp/LSP/blob/736a78bc2921cf4827dcab9b6829808caa4d1ecd/plugin/save_command.py#L13-L26
- https://github.com/sublimelsp/LSP/blob/736a78bc2921cf4827dcab9b6829808caa4d1ecd/plugin/formatting.py#L112-L119

I think we could implement the `CodeActionsOnFormat` feature for document selections (`CodeActionsOnSave` for selections would not make sense as you can't partially save a file 🙃 )by using the same tasks paradigm for this class:
https://github.com/sublimelsp/LSP/blob/736a78bc2921cf4827dcab9b6829808caa4d1ecd/plugin/formatting.py#L184-L216

And then go on from there. But I'd say this is an extra feature which can follow _after this PR_ and the follow up [I proposed above](https://github.com/sublimelsp/LSP/pull/2751#issuecomment-3872568376).

plugin/save_command.py
predragnikolic: fwiw so the future direction of eslint is deprecating formatting rules https://eslint.org/blog/2023/10/deprecating-formatting-rules/

jwortmann: > I would say that any real issues were fixed (#2674 and #2707) and the remaining comment (see the one in the first PR) is fine since we are compatible with VSCode.

Yes, the main issues are fixed, but still if there are multiple code actions of the same kind it's just luck whether it works or would result in broken file content. Even if VSCode (which is certainly not bug-free) does the same. The code action response is just defined as `(Command | CodeAction)[]` and there is no guarantee in the specs that contained code actions are be compatible and sorted in such an order that allows applying them all after another.

However, it's debatable how much that matters, as long as it seems to work in practice. For example I would expect that a `source.organizeImports` will only return a single code action.

jwortmann:
> * And format range - if we format a range, should we only include code actions for that formatted range when requesting code actions?

That's actually a good observation, I would say that `"lsp_code_actions_on_format"` should also work for the `LSP: Format Selection` command. At least when I read the setting description, there doesn't seem to be any restriction that it would only work when formatting the full file.

The implementation for range formatting should certainly be possible, even without the need to do additional filtering of code actions in the response, because the code action request already contains the `range` in the [CodeActionParams](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeActionParams). So it would just need to use the range(s) of the selection(s) instead of the full document range at
https://github.com/sublimelsp/LSP/blob/b18e32bc30c589728ba818a0ddb88ca49494beb9/plugin/code_actions.py#L180

mj026:
> That's actually a good observation, I would say that `"lsp_code_actions_on_format"` should also work for the `LSP: Format Selection` command. 

I think this should be supported too and my guess for now is that it _shouldn’t be that hard to implement_ (famous last words 😉).

PR #2752, opened 2026-02-08
fix diagnostic tag regions not having correct key

rchl:
Since we didn't filter out keys starting with `_` we ended up picking some random key/value that happened to match `1` (for the "Unnecessary" tag) thus the custom color rules didn't match.

I'm sure this worked before and I don't see any changes to the logic so this might be only broken in 3.13 host where order of items differs a bit or because `_`-prefixed keys are different. Seems like the code was always wrong.

BTW. In this case iterating `DiagnosticsTag.__dict__` returned:

```
- __static_attributes__: ()
- _new_member_: <built-in method __new__ of type object at 0x104d2a970>
- _use_args_: True
- _member_names_: ['Unnecessary', 'Deprecated']
- _member_map_: {'Unnecessary': <DiagnosticTag.Unnecessary: 1>, 'Deprecated': <DiagnosticTag.Deprecated: 2>}
- _value2member_map_: {1: <DiagnosticTag.Unnecessary: 1>, 2: <DiagnosticTag.Deprecated: 2>}
- _hashable_values_: [1, 2]
- _unhashable_values_: []
- _unhashable_values_map_: {}
- _member_type_: <class 'int'>
- _value_repr_: <slot wrapper '__repr__' of 'int' objects>
- Unnecessary: 1
- Deprecated: 2
- __format__: <method '__format__' of 'int' objects>
- __new__: <function Enum.__new__ at 0x103371940>
- ```

And the `True` value from `_use_args_` has actually matched `1` that we were looking for.

### DIFF
--- a/plugin/core/constants.py
+++ b/plugin/core/constants.py
@@ -217,7 +217,7 @@ SUBLIME_KIND_SCOPES: dict[SublimeKind, str] = {
     sublime.KIND_VARIABLE: "entity.name.constant | constant.other | support.constant | variable.other | variable.parameter | variable.other.member | variable.other.readwrite.member"  # noqa: E501
 }
 
-DIAGNOSTIC_TAG_SCOPES: dict[int, str] = {
+DIAGNOSTIC_TAG_SCOPES: dict[DiagnosticTag, str] = {
     DiagnosticTag.Unnecessary: 'markup.unnecessary.lsp',
     DiagnosticTag.Deprecated: 'markup.deprecated.lsp'
 }
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -322,9 +322,9 @@ class SessionView:
         if data and severity <= max_severity_level:
             non_tag_regions = data.regions
             for tag, regions in data.regions_with_tag.items():
-                tag_scope = DIAGNOSTIC_TAG_SCOPES.get(tag)
+                tag_scope = DIAGNOSTIC_TAG_SCOPES[tag]
                 # Trick to only add tag regions if there is a corresponding color scheme scope defined.
-                if tag_scope and 'background' in self.view.style_for_scope(tag_scope):
+                if 'background' in self.view.style_for_scope(tag_scope):
                     tags[tag].regions = regions
                     tags[tag].scope = tag_scope
                 else:
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -10,6 +10,7 @@ from ...protocol import Command
 from ...protocol import Diagnostic
 from ...protocol import DiagnosticRelatedInformation
 from ...protocol import DiagnosticSeverity
+from ...protocol import DiagnosticTag
 from ...protocol import DidChangeTextDocumentParams
 from ...protocol import DidCloseTextDocumentParams
 from ...protocol import DidOpenTextDocumentParams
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by jwortmann** [comment]
Instead of using `DiagnosticTag.__dict__` and filtering out all kind of private stuff, can't we just use `DiagnosticTag.__members__`?

https://docs.python.org/3/howto/enum.html#iteration

And to be honest, instead of some complicated `diagnostics_tag_scope` method I would just define a dict
```python
DIAGNOSTIC_TAG_SCOPES: dict[DiagnosticTag, str] = {
    DiagnosticTag.Unnecessary: 'markup.unnecessary.lsp',
    DiagnosticTag.Deprecated: 'markup.deprecated.lsp'
}
```
and put that into `constants.py`. If the they ever add new diagnostic tag values into the LSP specs, we would need to update the docs and probably some of the code anyway, and then we could just manually add another line into that dict.

plugin/core/constants.py
```diff
@@ -216,6 +217,11 @@ class RegionKey(StrEnum):
     sublime.KIND_VARIABLE: "entity.name.constant | constant.other | support.constant | variable.other | variable.parameter | variable.other.member | variable.other.readwrite.member"  # noqa: E501
 }
 
+DIAGNOSTIC_TAG_SCOPES: dict[int, str] = {
```
jwortmann: I think if you modify
https://github.com/sublimelsp/LSP/blob/948fed7a9b52e79ae606db069c9e419e38a4abc8/plugin/core/views.py#L85
 to `dict[DiagnosticTag, list[sublime.Region]]`, you can put `dict[DiagnosticTag, str]` here.

plugin/session_view.py
```diff
@@ -325,12 +317,12 @@ def _draw_diagnostics(
...
         ICON_FLAGS = sublime.RegionFlags.HIDE_ON_MINIMAP | sublime.RegionFlags.DRAW_NO_FILL | sublime.RegionFlags.DRAW_NO_OUTLINE | sublime.RegionFlags.NO_UNDO  # noqa: E501
         key = self.diagnostics_key(severity, multiline)
-        tags = {tag: TagData(f'{key}_tags_{tag}') for tag in DIAGNOSTIC_TAG_VALUES}
+        tags = {tag: TagData(f'{key}_tags_{tag}') for tag in DIAGNOSTIC_TAG_SCOPES}
         data = self._session_buffer.diagnostics_data_per_severity.get((severity, multiline))
         if data and severity <= max_severity_level:
             non_tag_regions = data.regions
             for tag, regions in data.regions_with_tag.items():
-                tag_scope = self.diagnostics_tag_scope(tag)
+                tag_scope = DIAGNOSTIC_TAG_SCOPES.get(tag)
```
jwortmann: I think with proper typing (see previous comment) the `get` is not necessary and could just be `DIAGNOSTIC_TAG_SCOPES[tag]` (and the check in the line below can be simplified then).

rchl:
```
Argument of type "int" cannot be assigned to parameter "key" of type "DiagnosticTag" in function "__getitem__"
  "int" is not assignable to "DiagnosticTag" (basedpyright)
```

Could always cast somewhere but...

rchl: OK, will do more changes to make it work.

plugin/session_view.py
```diff
@@ -322,9 +322,9 @@ def _draw_diagnostics(
         if data and severity <= max_severity_level:
             non_tag_regions = data.regions
             for tag, regions in data.regions_with_tag.items():
-                tag_scope = DIAGNOSTIC_TAG_SCOPES.get(tag)
+                tag_scope = DIAGNOSTIC_TAG_SCOPES[tag]
                 # Trick to only add tag regions if there is a corresponding color scheme scope defined.
-                if tag_scope and 'background' in self.view.style_for_scope(tag_scope):
+                if 'background' in self.view.style_for_scope(tag_scope):
```
jwortmann, line 327: By the way. We should probably do some caching for this, like I did a while ago for semantic token color scheme rules in https://github.com/sublimelsp/LSP/commit/f496608d9eae849df49397c639f0626809039215.
So that the ST API is not called everytime when diagnostics with tags are redrawn.

The `evaluate_semantic_tokens_color_scheme_support` function could be generalized to evaluate all rules from color schemes that we use throughout the code. That function automatically updates when the color scheme changes.

Probably that's for another PR though.

PR #2753, opened 2026-02-08
chore: use ruff for linting

rchl:
Switch from flake8 to ruff. Much faster, supports much more rules.

Ignore all errors that we currently have in the project. We can fix those when we see fit (ideally rule per commit) or leave them ignored.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,1 +1,1 @@
-zensical==0.0.18
+zensical==0.0.21
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -18,13 +18,12 @@ from .core.views import first_selection_region
 from .core.views import format_code_actions_for_quick_panel
 from .core.views import kind_contains_other_kind
 from .core.views import text_document_code_action_params
-from .save_command import LspSaveCommand
-from .save_command import SaveTask
+from .lsp_task import LspTask
 from abc import ABCMeta
 from abc import abstractmethod
 from functools import partial
-from typing import TYPE_CHECKING, Any, List, Tuple, Union, final
-from typing import cast
+from typing import cast, TYPE_CHECKING, Any, List, Tuple, Union, final
+from typing_extensions import override
 import sublime
 
 if TYPE_CHECKING:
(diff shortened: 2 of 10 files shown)


pyproject.toml
```diff
@@ -1,57 +1,264 @@
 [tool.pyright]
 pythonVersion = "3.8"
+reportAny = "none"
+reportExplicitAny = "none"
+reportImplicitOverride = "none"
 reportIncompatibleMethodOverride = "none"
 reportMissingModuleSource = "none"
+reportUnannotatedClassAttribute = "none"
+reportUnusedCallResult = "none"
 stubPath = "./stubs"
```
rchl, line 11: Some of those "report*" settings are specific to basedpyright. pyright complains a bit about unknown settings but it doesn't affect the exit code.

mj026:
Not that I have anything to say about it 😉  but I would be a proponent to have `ruff` in this project!  👍

pyproject.toml
```diff
@@ -1,57 +1,264 @@
...
 
-[tool.ruff.format]
-quote-style = "preserve"
-indent-style = "space"
-# Respect magic trailing commas.
-skip-magic-trailing-comma = false
-# Automatically detect the appropriate line ending.
-line-ending = "auto"
+[tool.mypy]
+python_version = "3.9"
```
rchl, line 246: mypy doesn't support 3.8 target anymore so picked closest that it does

pyproject.toml
```diff
@@ -1,57 +1,264 @@
...
+    "D404",  # https://docs.astral.sh/ruff/rules/docstring-starts-with-this/
+    "D412",  # https://docs.astral.sh/ruff/rules/blank-lines-between-header-and-content/
+    "D413",  # https://docs.astral.sh/ruff/rules/missing-blank-line-after-last-section/
     "D415",  # https://docs.astral.sh/ruff/rules/missing-terminal-punctuation/
+    "DOC201",  # https://docs.astral.sh/ruff/rules/docstring-missing-returns/
+    "DOC402",  # https://docs.astral.sh/ruff/rules/docstring-missing-yields/
+    "DOC501",  # https://docs.astral.sh/ruff/rules/docstring-missing-exception/
+    "DTZ005",  # https://docs.astral.sh/ruff/rules/call-datetime-now-without-tzinfo/
+    "E241",  # https://docs.astral.sh/ruff/rules/multiple-spaces-after-comma/
+    "E501",  # https://docs.astral.sh/ruff/rules/line-too-long/
```
jwortmann: There are some `noqa: E501` coments in the LSP code base like
https://github.com/sublimelsp/LSP/blob/6a25bd1e9e401e1cbbd994117a7013c90cd9f880/plugin/core/views.py#L68

I wonder if that is still relevant and works for Ruff as well? So should this rule either not be in the ignore list, or should all these comments be removed?

Also this ignore list seems pretty long, does Ruff report lint errors for all of those rules for some code in the LSP package...?

rchl: Ruff supports noqa comments. In this case I definitively prefer this rule to be ignored per line than globally so will remove it.

Those are reported as errors, yes, (at least in CLI, in the editor I see them as warnings) and for every ignore there is some code to fix in our code base.

PR #2739, opened 2026-01-27
Introduce LspPlugin API

rchl:
~~"Introduce new plugin API" is a little over the top. I just wanted to jump-start it with one new method to allow for progressive enhancement and get early feedback.~~

The new `handle_update_or_installation_async` method combines `needs_update_or_installation` and `install_or_update`. Those were split because the code sets `installing...` status text if `needs_update_or_installation` returns true but that meant that various if checks had to be duplicated across those two and generally made the logic harder to read. With this one method, there is now a callable function passed that can be used to start the progress (perhaps unnecessary since user could use `configuration.set_view_status` but then that would likely lead to inconsistent status messages across packages).

The new method takes a dict with params. This is to allow adding new params without breaking API compatibility.

Most class methods are passed `PluginContext` now with `configuration` instance now to allow packages to get resolved configuration. Some packages currently read settings with `sublime.load_settings` but that has the problem of not considering project overrides.

Resolves #2039
Resolves #2491

PR #2755, opened 2026-02-09
feat: add API decorators

rchl:
~~Add `APIDecorator` class that can be used to decorate `Session`/`AbstractPlugin` to handle server requests and notifications.~~

Add `request_handler` and `notification_handler` decorators for setting up handlers for server-initiated requests/notifications.

This is variant of the decorator included in `lsp_utils` that I want to retire.

Decorator allows simplifying handlers by not having to use awkward `method2attr` transformation on method names and not dealing with `request_id`. All the request handlers have to do is to return a promise resolved with the result.

To show how it works, converted existing `Session.m_window_showMessageRequest`:

Before:

```py
   def m_window_showMessageRequest(self, params: ShowMessageRequestParams, request_id: int | str) -> None:
        """handles the window/showMessageRequest request"""
        self.call_manager('handle_message_request', self, params, request_id)
```

After:

```py
    @request_handler('window/showMessageRequest')
    def on_show_message_request(self, params: ShowMessageRequestParams) -> Promise[MessageActionItem | None]:
        if mgr := self.manager():
            return mgr.handle_message_request(self.config.name, params)
        return Promise.resolve(None)
```

(the `call_manager` approach was also hacky so refactored that to normal public methods)

~~I'm not yet finished with the changes. Some more code can be converted in `Session` to use this decorator.~~ DONE

Open to any suggestions regarding naming and file name.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -2214,20 +2214,20 @@ class Session(APIHandler, TransportCallbacks['dict[str, Any]']):
     @request_handler('workspace/codeLens/refresh')
     def on_workspace_code_lens_refresh(self, _: None) -> Promise[None]:
 
-        def continue_after_return() -> None:
+        def continue_after_response() -> None:
             visible_session_views, not_visible_session_views = self.session_views_by_visibility()
             for sv in visible_session_views:
                 sv.session_buffer.do_code_lenses_async(sv.view)
             for sv in not_visible_session_views:
                 sv.session_buffer.set_code_lenses_pending_refresh()
 
-        sublime.set_timeout_async(continue_after_return)
+        sublime.set_timeout_async(continue_after_response)
         return Promise.resolve(None)
 
     @request_handler('workspace/semanticTokens/refresh')
     def on_workspace_semantic_tokens_refresh(self, _: None) -> Promise[None]:
 
-        def continue_after_return() -> None:
+        def continue_after_response() -> None:
             visible_session_views, not_visible_session_views = self.session_views_by_visibility()
             for sv in visible_session_views:
                 if sv.get_request_flags() & RequestFlags.SEMANTIC_TOKENS:
@@ -2237,13 +2237,13 @@ class Session(APIHandler, TransportCallbacks['dict[str, Any]']):
             for sv in not_visible_session_views:
                 sv.session_buffer.set_semantic_tokens_pending_refresh()
 
-        sublime.set_timeout_async(continue_after_return)
+        sublime.set_timeout_async(continue_after_response)
         return Promise.resolve(None)
 
     @request_handler('workspace/inlayHint/refresh')
     def on_workspace_inlay_hint_refresh(self, _: None) -> Promise[None]:
 
-        def continue_after_return() -> None:
+        def continue_after_response() -> None:
             visible_session_views, not_visible_session_views = self.session_views_by_visibility()
             for sv in visible_session_views:
                 if sv.get_request_flags() & RequestFlags.INLAY_HINT:
@@ -2253,20 +2253,20 @@ class Session(APIHandler, TransportCallbacks['dict[str, Any]']):
             for sv in not_visible_session_views:
                 sv.session_buffer.set_inlay_hints_pending_refresh()
 
-        sublime.set_timeout_async(continue_after_return)
+        sublime.set_timeout_async(continue_after_response)
         return Promise.resolve(None)
 
     @request_handler('workspace/diagnostic/refresh')
     def on_workspace_diagnostic_refresh(self, _: None) -> Promise[None]:
 
-        def continue_after_return() -> None:
+        def continue_after_response() -> None:
             visible_session_views, not_visible_session_views = self.session_views_by_visibility()
             for sv in visible_session_views:
                 sv.session_buffer.do_document_diagnostic_async(sv.view, sv.view.change_count(), forced_update=True)
             for sv in not_visible_session_views:
                 sv.session_buffer.set_document_diagnostic_pending_refresh()
 
-        sublime.set_timeout_async(continue_after_return)
+        sublime.set_timeout_async(continue_after_response)
         return Promise.resolve(None)
 
     @notification_handler('textDocument/publishDiagnostics')
@@ -2328,11 +2328,11 @@ class Session(APIHandler, TransportCallbacks['dict[str, Any]']):
                 capability_options = cast('DidChangeWatchedFilesRegistrationOptions', options)
                 self.register_file_system_watchers(registration_id, capability_options['watchers'])
 
-        def continue_after_return() -> None:
+        def continue_after_response() -> None:
             if new_workspace_diagnostics_provider:
                 self.do_workspace_diagnostics_async()
 
-        sublime.set_timeout_async(continue_after_return)
+        sublime.set_timeout_async(continue_after_response)
         return Promise.resolve(None)
 
     @request_handler('client/unregisterCapability')


plugin/core/message_request_handler.py
```diff
@@ -48,15 +44,11 @@ def show(self) -> None:
...
             wrapper_class='notification',
-            on_navigate=self._send_user_choice,
-            on_hide=self._send_user_choice)
+            on_navigate=lambda href: self._send_user_choice(resolve, href),
+            on_hide=lambda href: self._send_user_choice(resolve, href))
+        return promise
 
-    def _send_user_choice(self, href: int = -1) -> None:
-        if self.request_sent:
-            return
```
rchl, line 56: Perhaps this `request_sent` flag should not be removed. Don't remember how the UI for this works.

plugin/core/api_decorator.py
```diff
@@ -0,0 +1,78 @@
...
+# P represents the parameters *after* the 'self' argument
+P = TypeVar('P', bound=LSPAny)
+R = TypeVar('R', bound=LSPAny)
+
+
+class APIDecorator:
+    """Decorate class methods to handle server initiated requests and notifications.
+
+    1. Ensure class is decorated with `APIDecorator.initialize`.
+    2. Add `APIDecorator.request('...')` and/or `APIDecorator.notification('...')` decorators on chosen class methods.
```
jwortmann: I haven't yet tried to understand how it works, but would it be possible to not require the `APIDecorator.initialize` for users of these decorators? Like could we put that already on the `LspPlugin` / `AbstractPlugin` class, so that the handler decorators work out of the box?

In 2. the description doesn't match the decorator names (I think the `notification_handler` and `request_handler` names are fine). I would write "on chosen methods" (without the "class"), because it's not necessarily a class method.

Also is it possible to use the decorators without the class name prefix, i.e. could we export `notification_handler` and `request_handler` directly, instead of `APIDecorator`? I think something like
```python
    @request_handler('window/showMessageRequest')
    def on_window_show_message_request(self, params: ShowMessageRequestParams) -> Promise[MessageActionItem | None]:
```
would look a bit nicer.

rchl: > Also is it possible to use the decorators without the class name prefix, i.e. could we export `notification_handler` and `request_handler` directly, instead of `APIDecorator`? I think something like

In lsp_utils those were loose functions and in here I was going back and forth about that. I've decided to go with class because that makes it more clear that the `initialize` decorator has to be used also. That said...

> I haven't yet tried to understand how it works, but would it be possible to not require the `APIDecorator.initialize` for users of these decorators? Like could we put that already on the `LspPlugin` / `AbstractPlugin` class, so that the handler decorators work out of the box?

That should be doable since `AbstractPlugin` (and `Session` too of course) has a concrete `__init__` implementation. In that case it will make sense to just have loose decorators for requests and notifications instead of the class approach.

BTW. The way this works is:
1. method decorator sets a marker (arbitrary unique key that I've chosen) on the function just to mark those as Notification or Request handlers. The method name is stored in that marker.
2. For the case of Request decorator also wraps the handler so that:
   - we intercept the `request_id` and not expose it to the original handler
   - handle result from the original function that now is expected to return a Promise. This result is then returned in [`Session.on_payload`](https://github.com/sublimelsp/LSP/blob/ee79cfb430999f1a195f97f4ac50d8b842d10e1b/plugin/core/sessions.py#L2628-L2650) and the promise result is special-cased (old-style handlers return `None` always) to send payload back to the server.
3. The class decorator wraps `__init__` and once the class is instantiated it looks for all methods that have our marker and using `method2attr` assigns handler function on the class with new, transformed name (the original name is left present). We can only assign new methods once the class is instantiated so that's why we need separate decorator for the class.

plugin/core/sessions.py
```diff
@@ -2222,10 +2227,10 @@ def m_workspace_semanticTokens_refresh(self, params: None, request_id: int | str
                 sv.session_buffer.set_semantic_tokens_pending_refresh()
         for sv in not_visible_session_views:
             sv.session_buffer.set_semantic_tokens_pending_refresh()
+        return Promise.resolve(None)
```
rchl, line 2241: Here and elsewhere we'll now trigger requests before sending a response. Could this be an issue for some servers? Not sure. We can change it by delaying method's logic until "next tick" but not sure if we should unless we see any issues.

plugin/core/api_decorator.py
rchl: Refactored to just `request_handler` and `notification_handler` functions.

plugin/core/sessions.py
rchl: I think this is now actually more correct. In the spec, in the documentation for [workspace/codeLens/refresh](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeLens_refresh) we can see the response documentation:

> Response:
>
>    result: void
>    error: code and message set in case an exception happens during the ‘workspace/codeLens/refresh’ request

Which means that we should actually trigger our requests before responding and respond with error if something failed (which is unlikely case I guess but that's another matter).

jwortmann: This is just the generic text that is written on all requests in the specs, because sending an error response is always possible. Note that `workspace/codeLens/refresh` is the server request and not our `textDocument/codeLens` requests. So assuming that it is no typo, the only way I see how an error could happen here would be if the server sends params that are not `null`, which would be a spec violation. Probably it would be easier if it was defined as a notification instead of a request. If an error happens during one of our `textDocument/codeLens` requests, we should still not send an error here. But I think actually sending the response first, before our requests, is the correct thing to do, for the following reason. This request is triggered on changes in a configuration file. Let's say the server is still progressing our code lens requests and there happens another change in the configuration file. Currently, we would cancel the open requests (I think) and send new ones. But in case the server blocks new `workspace/codeLens/refresh` requests when there is still an old one open, this would not work and everything would be delayed until all requests have finished, or in the worst case the new trigger from the second change could even be skipped entirely. So I think it might be better to delay our request via `set_timeout_async` (or maybe don't use the decorator for handling the refresh requests, to save that ST API call). But maybe also it doesn't really matter in practice...

plugin/core/message_request_handler.py
rchl: Tested it now and this was required since `on_hide` always triggers after `on_navigate` (due to `hide_popup()`),

plugin/api.py
```diff
@@ -0,0 +1,98 @@
...
+from typing import Any, Callable, TypeVar, TYPE_CHECKING
+from typing_extensions import ParamSpec
+
+if TYPE_CHECKING:
+    from .promise import Promise
+
+__all__ = [
+    'initialize_api_decorators',
+    'notification_handler',
+    'request_handler',
```
predragnikolic, line 15: What do you think about these names?on_notification
on_request

jwortmann: I think the `_handler` names are better. See the examples in the docstrings, where something like `on_some_notification` is already a commonly used convention for the actual handler methods.

rchl: Yeah, it's making the function become a handler. Decorator is not a handler itself.

jwortmann: But I think I would probably use just `api_decorators` instead of `initialiaze_api_decorators` for the class decorator.

---

Also maybe we should rename the file to just `plugin/api.py` (see #2739) so that the decorators are in the same file as the plugin API class. That makes it easier to find them if someone wants to read the docstrings.

plugin/core/api_decorator.py
```diff
@@ -0,0 +1,98 @@
...
+    @wraps(original_init)
+    def init_wrapper(self: T, *args: Params.args, **kwargs: Params.kwargs) -> None:
+        original_init(self, *args, **kwargs)
+        for attr in dir(self):
+            if (func := getattr(self, attr)) and callable(func) and hasattr(func, HANDLER_MARKER):
+                # Set method with transformed name on the class instance.
+                setattr(self, method2attr(getattr(func, HANDLER_MARKER)), func)
+
+    _class.__init__ = init_wrapper
+    return _class
```
rchl: This could technically just be a class that has to be inherited. It being a decorator might be overengineering.

plugin/api.py
rchl: Sure, renamed now.

plugin/core/sessions.py
```diff
@@ -2172,83 +2174,103 @@ def _on_workspace_diagnostics_error_async(self, identifier: DiagnosticsIdentifie
...
+
+    @request_handler('workspace/applyEdit')
+    def on_workspace_apply_edit(self, params: ApplyWorkspaceEditParams) -> Promise[ApplyWorkspaceEditResult]:
+        return self.apply_workspace_edit_async(params.get('edit', {}), label=params.get('label')) \
+            .then(lambda _: {"applied": True})
+
+    @request_handler('workspace/codeLens/refresh')
+    def on_workspace_code_lens_refresh(self, _: None) -> Promise[None]:
+
+        def continue_after_return() -> None:
```
jwortmann: Should we name this `continue_after_response`? Maybe that would make the purpose a bit clearer.

plugin/api.py
```diff
@@ -0,0 +1,90 @@
...
+__all__ = [
+    'APIHandler',
+    'notification_handler',
+    'request_handler',
+]
+
+HANDLER_MARKER = '__HANDLER_MARKER'
+
+# P represents the parameters *after* the 'self' argument
+P = TypeVar('P', bound=LSPAny)
```
jwortmann, line 21: I think the type annotations could be improved by using `Concatenate` and/or `ParamSpec('P', bound=LSPAny)`, compare https://github.com/sublimelsp/LSP/blob/8c77738df6cdff50bf60784f29a271f8e84b033b/plugin/documents.py#L81-L93

But I'd have to checkout the code and try if that actually works, so we could do that later.

plugin/core/sessions.py
```diff
@@ -2305,12 +2327,16 @@ def m_client_registerCapability(self, params: RegistrationParams, request_id: in
...
                 self.register_file_system_watchers(registration_id, capability_options['watchers'])
-        self.send_response(Response(request_id, None))
-        if new_workspace_diagnostics_provider:
-            self.do_workspace_diagnostics_async()
 
-    def m_client_unregisterCapability(self, params: UnregistrationParams, request_id: int | str) -> None:
-        """handles the client/unregisterCapability request"""
+        def continue_after_return() -> None:
+            if new_workspace_diagnostics_provider:
+                self.do_workspace_diagnostics_async()
```
jwortmann, line 2333: By the way (not special to this PR):

I've noticed that we should probably also trigger document diagnostics after provider registration. Currently, after starting a server that uses dynamic registration for pull diagnostics, they are not triggered for the first file (because it is already opened when the registration happens). So we don't get any diagnostics for that file until the buffer is modified (unless the server supports workspace diagnostics and sends a diagnostics report for that file in the workspace diagnostics response / progress reports).

jwortmann (approved):

plugin/api.py
rchl: I've contemplated this when implementing but I think it's not appropriate here since it would allow any numbers of arguments while we need to have exactly one.

Unless I don't fully understand something.

PR #2757, opened 2026-02-11
Add caching for diagnostic annotation colors

jwortmann:


jwortmann:
I saw that there is a bit of duplication, but I didn't want to do a big refactoring here. I would probably even inline the `format_diagnostics_for_annotation` function (or move into DiagnosticsAnnotationsView as a staticmethod)  because that is the only place where it's used.

PR #2761, opened 2026-02-13
Finalize `code_actions_on_format`

mj026:
This PR is a follow up of https://github.com/sublimelsp/LSP/pull/2747 and https://github.com/sublimelsp/LSP/pull/2751

An extra task `CodeActionsFormatOnSaveTask` is added to trigger the code actions defined in `code_actions_on_format` when `format_on_save` is enabled. I also added the `code_actions_on_format` setting to the documentation.

[There was some discussion](https://github.com/sublimelsp/LSP/pull/2751#issuecomment-3874254634) if these code actions should also be triggered when utilising the `LSP: Format Selection` command.  

The consensus was that we should (and even take `lsp_format_on_paste` in account)  so I investigated and tested some scenario's with this and came to the conclusion that I think we should not do this, at least for now. 

The code actions that make sense to configure (and are currently allowed) in `lsp_code_actions_on_save` and `lsp_code_actions_on_format` are `source.*` code actions. For example, `source.fixAll`,`source.organizeImports`, `source.sort.json` etc. 

This is also enforced in the current implementation: 

https://github.com/sublimelsp/LSP/blob/83bc39c4a8afb0ba6a2fc8ccdba1f7bfbbd716b5/plugin/code_actions.py#L231-L238

[According to the specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/), `source.*` actions are **applied to the whole file**:

https://github.com/sublimelsp/LSP/blob/83bc39c4a8afb0ba6a2fc8ccdba1f7bfbbd716b5/protocol/__init__.py#L470-L484

And this is documented as such in Sublime LSP:

https://github.com/sublimelsp/LSP/blob/83bc39c4a8afb0ba6a2fc8ccdba1f7bfbbd716b5/sublime-package.json#L27

When asking the LSP for a code action while sending a specific range, it will return a response with a task to format the whole file without any range data:

`:: [17:18:39.508] --> ruff textDocument/codeAction (21):`
```json
{
  "textDocument": {
    "uri": "file:///my/django/project/manage.py"
  },
  "range": {
    "start": { "line": 3, "character": 0 },
    "end": { "line": 5, "character": 0 }
  },
  "context": {
    "diagnostics": [],
    "triggerKind": 2,
    "only": ["source.organizeImports.ruff"]
  }
}
```
`:: [17:18:39.510] <<< ruff (21) (duration: 2ms):`
```json
[
  {
    "data": "file:///my/django/project/manage.py",
    "kind": "source.organizeImports.ruff",
    "title": "Ruff: Organize imports"
  }
]
```

The result is that _other parts of your document are changed_ when formatting a specific section, parts which are not inside the current selection. I find this very counter intuitive and not something I would expect. 

Client side filtering will result in complex situations and messy code so I did not consider it.

I also checked this specific scenario in the Zed editor (which I "stole" the idea from) and there the code actions **are actually applied**, outside the current selection. I think it shouldn't and therefore I did not include this behaviour.

I'm happy to see / read your thoughts about this 😅

### DIFF
--- a/tests/test_code_actions.py
+++ b/tests/test_code_actions.py
@@ -306,10 +306,7 @@ class CodeActionsOnFormatTestCase(CodeActionsTestCaseBase):
         self.assertEqual(self.view.is_dirty(), False)
 
 
-class CodeActionsOnFormatOnSaveTaskTestCase(unittest.TestCase):
-    def setUp(self):
-        self.view = sublime.active_window().active_view()
-
+class CodeActionsOnFormatOnSaveTaskTestCase(TextDocumentTestCase):
     def test_code_actions_format_on_save_task_enabled__unset(self):
         self.view.settings().set('lsp_code_actions_on_format', {})
         self.view.settings().set("lsp_format_on_save", False)


plugin/code_actions.py
```diff
@@ -287,6 +287,20 @@ class CodeActionsOnFormatTask(CodeActionsTaskBase):
...
+@final
+class CodeActionsFormatOnSaveTask(CodeActionsTaskBase):
+    """Run code actions on format when format_on_save is enabled."""
+
+    SETTING_NAME = "lsp_code_actions_on_format"
+
+    @classmethod
+    @override
+    def is_applicable(cls, view: sublime.View) -> bool:
+        format_on_save_enabled = bool(view.settings().get('lsp_format_on_save', False))
```
jwortmann: Does this work with/without project-specific overrides?
Compare to
https://github.com/sublimelsp/LSP/blob/8c77738df6cdff50bf60784f29a271f8e84b033b/plugin/formatting.py#L90-L91
which checks both values from the view's setting and from the global userprefs.

plugin/code_actions.py
```diff
@@ -287,6 +287,20 @@ class CodeActionsOnFormatTask(CodeActionsTaskBase):
...
+        code_actions_on_format_defined = bool(cls._get_code_actions(view))
+        return bool(view.window() and format_on_save_enabled and code_actions_on_format_defined)
```
jwortmann: I would rewrite this method to return early if possible and you could also make use of the implementation in the base class.

Something like
```python
if not super().is_applicable(view):
    return False
format_on_save_enabled = ...
return format_on_save_enabled
```

plugin/save_command.py
```diff
@@ -1,5 +1,5 @@
 from __future__ import annotations
-from .code_actions import CodeActionsOnSaveTask
+from .code_actions import CodeActionsFormatOnSaveTask, CodeActionsOnSaveTask
```
jwortmann: By convention in this code base each import should be on a single line (except for imports from the typing and typing_extensions modules).

tests/test_code_actions.py
```diff
@@ -1,6 +1,6 @@
 from __future__ import annotations
 from copy import deepcopy
-from LSP.plugin.code_actions import get_matching_on_save_kinds
+from LSP.plugin.code_actions import get_matching_on_save_kinds, CodeActionsFormatOnSaveTask
```
jwortmann: ... probably not so important in the test files, but ideally here too on separate lines.

jwortmann (commented):
Should we add another sentence in the setting description that the code actions configured in "lsp_code_actions_on_format" are also run on save if "lsp_format_on_save" is enabled?

---

> `source.*` actions are applied to the whole file

Right, I forgot about that. I agree that they should not be triggered then when formatting a selection.

mj026:
> Should we add another sentence in the setting description that the code actions configured in "lsp_code_actions_on_format" are also run on save if "lsp_format_on_save" is enabled?

Yes, I'll add explanation of this behaviour to the description.

> > `source.*` actions are applied to the whole file
> 
> Right, I forgot about that. I agree that they should not be triggered then when formatting a selection.

👍

plugin/code_actions.py
mj026: Yes, it seems like `view.settings()` already merges default/user/project settings. I've been testing this feature all the time with a project override without issues and this is supposed to be the expected behaviour as [confirmed on the Sublime Forum](https://forum.sublimetext.com/t/api-and-project-settings-hierarchy/30764/3).

plugin/code_actions.py
mj026: I changed the implementation as requested. ~~I left out the specific `userprefs()` call as [it doesn't seems to be explicitly needed](https://github.com/sublimelsp/LSP/pull/2761#discussion_r2807334191).~~ I check for the `userprefs()` setting now.

plugin/code_actions.py
jwortmann: I'm aware that ST settings from *Preferences.sublime-settings* are accessible in the view's settings, but userprefs() are the settings from *LSP.sublime-settings*. I believe we need to check those if set to `true` by the user.

Perhaps we should refactor that into a small function for reading LSP settings including project overrides, and then use that function here and at other places where we read such settings.

mj026: Ah ok, now I understand _LSP.sublime-settings_ are not automatically taken into account. I added a new commit to check for the `userprefs()` setting too (and added this to the tests)

tests/test_code_actions.py
```diff
@@ -1,12 +1,14 @@
...
 from copy import deepcopy
 from LSP.plugin.code_actions import get_matching_on_save_kinds
+from LSP.plugin.code_actions import CodeActionsFormatOnSaveTask
 from LSP.plugin.core.constants import RegionKey
 from LSP.plugin.core.protocol import Point
 from LSP.plugin.core.url import filename_to_uri
 from LSP.plugin.core.views import entire_content
 from LSP.plugin.core.views import kind_contains_other_kind
 from LSP.plugin.core.views import versioned_text_document_identifier
+from LSP.plugin.core.settings import userprefs
```
rchl: `from LSP.plugin.core.views import versioned_text_document_identifier` not sorted properly.

Also `from LSP.plugin.code_actions import CodeActionsFormatOnSaveTask` since we tend to use ST sorting which disregards case.

tests/test_code_actions.py
```diff
@@ -111,6 +113,8 @@ def doCleanups(self) -> Generator:
         yield from self.await_clear_view_and_save()
         yield from super().doCleanups()
 
+
+class CodeActionsOnSaveTestCase(CodeActionsTestCaseBae):
```
rchl:
```suggestion
class CodeActionsOnSaveTestCase(CodeActionsTestCaseBae):
```

tests/test_code_actions.py
mj026: Fixed in latest version of the branch.

Would it be an idea to enforce isort rules via `ruff`? I'm now sort of guessing what it should be and I’ve noticed it's not consistent throughout the codebase.

These rules:
```toml
[tool.ruff.lint.isort]
case-sensitive = false
from-first = true
force-single-line = true
no-sections = true
order-by-type = false
```
seem to match your sorting rules pretty closely (including the one import per line, case insensitive alphabetic order etc) 

Then we can let tox and the CI runner complain about it and have the ruff formatter do the job for us, see the diff: https://github.com/sublimelsp/LSP/compare/main...mj026:LSP:chore/enforce-ruff-isort-rules?expand=1 where I applied these rules to the whole repository.

Also a perfect use-case to enable `lsp_code_actions_on_format` with the `source.organizeImports.ruff` action 😃

plugin/code_actions.py
```diff
@@ -287,6 +287,22 @@ class CodeActionsOnFormatTask(CodeActionsTaskBase):
...
+    SETTING_NAME = "lsp_code_actions_on_format"
+
+    @classmethod
+    @override
+    def is_applicable(cls, view: sublime.View) -> bool:
+        if not super().is_applicable(view):
+            # When there are no code actions configured we can return immediately
+            return False
+        format_on_save_enabled = bool(view.settings().get('lsp_format_on_save', userprefs().lsp_format_on_save))
+        return format_on_save_enabled
```
jwortmann: You don't really need that variable here, my code was just an example if it takes multiple lines.
Also the `bool` seems unnecessary because the setting is already a bool. If the type cannot be inferred correctly (Any), I would vote to either use `cast` or a `pyright: ignore[...]` comment here. (Maybe we can figure out later how to properly type view settings for the setting names from LSP.)

Also this behaves a bit different to the other place in https://github.com/sublimelsp/LSP/blob/8c77738df6cdff50bf60784f29a271f8e84b033b/plugin/formatting.py#L90-L91

If someone puts `"lsp_format_on_save": 42` in project-specific settings, that code ignores it and uses the default / global setting, while here it would be considered `true`. I think I'd prefer ignoring non-bool values, but in any case I'd say it should be consistent accross different places in the code, because otherwise it can lead to difficult to find bugs or issue reports later.

---

(Also imo the comment 2 lines above is not really needed)

jwortmann:
> Yes, I'll add explanation of this behaviour to the description.

Actually I meant the LSP.sublime-settings file, because that's what most people probably read. Maybe add a sentence there too?

---

By the way. If someone puts for example
```json
  "lsp_code_actions_on_format": {
    "source.fixAll": true,
    "source.organizeImports": true,
  },
  "lsp_code_actions_on_save": {
    "source.organizeImports": true,
  },
```
And also has "lsp_format_on_save" enabled, do we try to run the organizeImports code action twice when saving a file? Or could/should we skip the kinds that are already defined in "lsp_code_actions_on_save" when it gets triggered on save due to format-on-save?

plugin/code_actions.py
mj026: I changed the implementation now to :
```python
    @classmethod
    @override
    def is_applicable(cls, view: sublime.View) -> bool:
        if not super().is_applicable(view):
            return False
        view_format_on_save = view.settings().get('lsp_format_on_save', None)
        return view_format_on_save if isinstance(view_format_on_save, bool) else userprefs().lsp_format_on_save
```

This:
- is consistent with the implementation in `FormattingOnSaveTask`
- will respect a boolean setting for `lsp_format_on_save` and ignore any other value
- returns the `userprefs().lsp_format_on_save` value when `lsp_format_on_save` is not a boolean

mj026:
> Actually I meant the LSP.sublime-settings file, because that's what most people probably read. Maybe add a sentence there too?

Ok, I'll adjust the ` LSP.sublime-settings` as well, ~~will follow soon(ish)~~ 
**EDIT** updated the `LSP.sublime-settings` settings file now.

> 
> By the way. If someone puts for example
> 
> ```json
>   "lsp_code_actions_on_format": {
>     "source.fixAll": true,
>     "source.organizeImports": true,
>   },
>   "lsp_code_actions_on_save": {
>     "source.organizeImports": true,
>   },
> ```
> 
> And also has "lsp_format_on_save" enabled, do we try to run the organizeImports code action twice when saving a file? 

With the current implementation **yes**. The actions defined in `lsp_code_actions_on_save` will run before the actions in `lsp_code_actions_on_format` as defined in the tasks method:
```python
def tasks(self) -> list[type[LspTask]]:
    return [
        CodeActionsOnSaveTask,
        CodeActionsFormatOnSaveTask,
        FormattingOnSaveTask,
        WillSaveWaitTask,
    ]
```

> Or could/should we skip the kinds that are already defined in "lsp_code_actions_on_save" when it gets triggered on save due to format-on-save?

Good and relevant question (and also asked by @rchl earlier) and I decided until now to not do it as it might complicate things a bit too much. 

But we could skip the code actions which are configured via `lsp_code_actions_on_save` indeed. ~~I leave it up to you to decide.~~ This seems like a good solution to prevent unneeded duplicate code actions. ~~I'll add this functionality the following day(s).~~ 
**EDIT** I added this now in commit https://github.com/sublimelsp/LSP/pull/2761/changes/dc455726958591a25950a2b627af6ef82feaed61

plugin/code_actions.py
```diff
@@ -287,6 +287,33 @@ class CodeActionsOnFormatTask(CodeActionsTaskBase):
     SETTING_NAME = "lsp_code_actions_on_format"
 
 
+@final
+class CodeActionsFormatOnSaveTask(CodeActionsTaskBase):
```
jwortmann: I think this class could inherit from `CodeActionsOnFormatTask`, because it's a special kind of that task, which only runs on save and with the additional logic to filter out duplicated code action kinds. Then you can remove `SETTING_NAME` in this class.

And then we could also get rid of the `@final` decorator for `CodeActionsOnFormatTask` and for this class, which seems like a mistake to me here anyway.
`@final` is useful to prevent overrides, for example in
https://github.com/sublimelsp/LSP/blob/8c77738df6cdff50bf60784f29a271f8e84b033b/plugin/core/input_handlers.py#L69-L75
where `list_items` is normally something that you need to override in the implementation, but not in this case where it is already implemented internally and must not be changed in any subclass.
https://github.com/sublimelsp/LSP/blob/8c77738df6cdff50bf60784f29a271f8e84b033b/plugin/core/input_handlers.py#L52-L53

I don't see any reason why it would be useful on this task classes here.

jwortmann: Small nitpick: I would probably also rename to `CodeActionsOnFormatOnSaveTask` to be consistent with the other naming conventions.

*Edit:* I see now the other one is named `FormattingOnSaveTask`. Should we rename that to `FormatOnSaveTask`? The setting is also named "lsp_format_on_save" and not "lsp_formatting_on_save"...

tests/test_code_actions.py
rchl: If we are gonna have to reformat so much, we might as well just re-think the sorting rules. I'm not attached current ones personally and could go with whatever as long as it's enforced and auto-fixed.

The current one-symbol-per-import can look a bit silly, that's why we weren't consistent with it when it comes to `typing` module.

plugin/code_actions.py
rchl: Not having `@final` can make pyright/basedpyright complain if the class has class properties (or sets properties in `__init__`) since it then assumes that the class can be subclassed which then could lead to different types for existing properties.

I would say that it makes sense to have `@final` for classes that are not to be subclassed.

rchl:
@mj026 there should be no need for you to force-push changes. This just makes it harder to review incrementally and at the end the whole PR is squashed anyway.

plugin/code_actions.py
jwortmann: > Not having `@final` can make pyright/basedpyright complain if the class has class properties (or sets properties in `__init__`) since it then assumes that the class can be subclassed which then can lead to different types for existing properties.

That seems weird, because we use class properties all the time, for example in
https://github.com/sublimelsp/LSP/blob/8c77738df6cdff50bf60784f29a271f8e84b033b/plugin/core/registry.py#L103-L115

where they should explicitly be overriden in subclasses.

> I would say that it makes sense to have `@final` for classes that are not to be subclassed.

But in this case it does make sense to have a class hierarchy and allow subclasses, no? So that the subclass can reuse the `SETTING_NAME` from the superclass...

rchl: It's not that it's not allowed to have those class properties but not having explicit type can cause warning:

> Type annotation for attribute `capability` is required because this class is not decorated with `@final` (basedpyright)

This is the [reportUnannotatedClassAttribute](https://docs.basedpyright.com/v1.38.0/configuration/config-files/#reportUnannotatedClassAttribute) setting that is currently disabled in the project. Also it's specific to basedpyright.

rchl: > But in this case it does make sense to have a class hierarchy and allow subclasses, no? So that the subclass can reuse the `SETTING_NAME` from the superclass...

In that case it might make more sense to just add type annotation I suppose. But if the class is not currently subclassed by anything and the class is meant to be internal then we might just as well add `@final`

jwortmann: basedpyright's default rules are kinda ridiculous anyway because it marks almost everything as a warning.

My opinion on this is that we should use `@final` only if there is a particular reason that a class or a method must not be overridden, and don't just throw it on everything that currently doesn't have a subclass (yet).

Also in this case there seems already to be an explicit type annotation for that class property (not sure if that gets propagated or should get propagated to the subclass automatically):
https://github.com/sublimelsp/LSP/blob/8c77738df6cdff50bf60784f29a271f8e84b033b/plugin/code_actions.py#L223-L224

mj026:
> @mj026 there should be no need for you to force-push changes. This just makes it harder to review incrementally and at the end the whole PR is squashed anyway

@rchl Ok, sorry for messing it up, I will not force push anymore when not needed.

> Edit: I see now the other one is named FormattingOnSaveTask. Should we rename that to FormatOnSaveTask? The setting is also named "lsp_format_on_save" and not "lsp_formatting_on_save"...

> So that the subclass can reuse the SETTING_NAME from the superclass...

@jwortmann I renamed `FormattingOnSaveTask` to `FormatOnSaveTask` and `CodeActionsFormatOnSaveTask` to `CodeActionsOnFormatOnSaveTask`. I also subclassed from `CodeActionsOnFormatTask` now so it inherits the`SETTING_NAME` attribute.

> But in this case it does make sense to have a class hierarchy and allow subclasses, no? So that the subclass can reuse the SETTING_NAME from the superclass...
> I would say that it makes sense to have @final for classes that are not to be subclassed.

I have removed `@final` from the `CodeActionsOnFormatTask` as `CodeActionsOnFormatOnSaveTask` now inherits from it. I explicitly type annotated the `SETTING_NAME` everywere now so `pyright` will not complain about the `reportUnannotatedClassAttribute` rule.

tests/test_code_actions.py
mj026: Ok, not something for this PR anyway, but I would be very happy if I could just let `ruff` handle the import order stuff so you guys do not have to spend time reviewing PR's with import inconsistencies 🥲

rchl: I've tried quickly without `force-single-line = true` and I think I would prefer that. You can submit a PR for that and maybe everyone will agree :)

plugin/code_actions.py
mj026: It doesn't get propagated to the subclass, so I explicitly re-annotated the property in the subclasses now. See also my comment below https://github.com/sublimelsp/LSP/pull/2761#issuecomment-3908517906

tests/test_code_actions.py
```diff
@@ -93,13 +95,13 @@ def diagnostic_to_lsp(diagnostic: tuple[str, Range]) -> dict:
     }
 
 
-class CodeActionsOnSaveTestCase(TextDocumentTestCase):
-
+class CodeActionsTestCaseBae(TextDocumentTestCase):
```
jwortmann: Typo? (Bae -> Base)

LSP.sublime-settings
```diff
@@ -24,7 +24,8 @@
   "show_code_actions": "annotation",
 
   // A dictionary of code action identifiers that should be triggered on format or
-  // on save.
+  // on save. Code actions defined in "lsp_code_actions_on_format" will also be triggered
+  // when "lsp_format_on_save" is enabled.
```
jwortmann:
```suggestion
  // on save if "lsp_format_on_save" is enabled.
```

tests/test_code_actions.py
mj026: Yeah, sorry, small dyslexia sometimes... fixed now.

plugin/code_actions.py
```diff
@@ -277,14 +277,38 @@ class CodeActionsOnSaveTask(CodeActionsTaskBase):
...
+    @classmethod
+    @override
+    def get_code_actions(cls, view: sublime.View) -> dict[str, bool]:
+        code_actions_on_format = super().get_code_actions(view)
+        code_action_on_save = CodeActionsOnSaveTask.get_code_actions(view)
+        # Prevent triggering of duplicate code actions
+        return {
+            action: enabled
+            for action, enabled in code_actions_on_format.items()
+            if action not in code_action_on_save
```
rchl, line 302: I suppose this is a bit naive because keys could be subsets of other keys in which case those would also be redundant but it can be tweaked later if needed.

tests/test_code_actions.py
```diff
@@ -277,6 +273,76 @@ def test_format_with_fixall_code_action(self) -> Generator:
...
+        # Response is fixed (fixAll added ";") and formatted (removed leading space)
+        self.assertEqual(entire_content(self.view), 'const x = 1;')
+        # Document should be saved
+        self.assertEqual(self.view.is_dirty(), False)
+
+
+class CodeActionsOnFormatOnSaveTaskTestCase(unittest.TestCase):
+    def setUp(self):
+        self.view = sublime.active_window().active_view()
+
```
rchl: If you need a view then just use `TextDocumentTestCase`.
```suggestion
class CodeActionsOnFormatOnSaveTaskTestCase(TextDocumentTestCase):
```

plugin/code_actions.py
mj026: Could you give an example of keys from a subset while being redundant in this case?

rchl: I'm thinking `source.fixAll.eslint` is a subset of `source.fixAll`.

mj026:
PS, I'd like to thank you @rchl and @jwortmann for accepting this feature and for all your time spend (late in the evening) on reviewing the PR's I submitted + all the suggestions and critical remarks on how to implement it / what to accept, highly appreciated!

I came across two smaller issues regarding this feature which I will submit some small future PR's to fix those. But the "main thing" is finished and I'm a happy that it is there! 🙏

jwortmann:
Thanks for your contributions :)

PR #2762, opened 2026-02-16
Disable project specific lsp server when it is disabled for the current session

mj026:
This PR fixes an issue where the "LSP server has crashed" dialog pops up forever when I have a sublime project with "project only" LSP servers, even when I press cancel. 

Example configuration:

```json
{
 	"folders":
	[
		{
			"path": "."
		}
	],
	"settings":
	{
		"LSP": {
			"ruff": {
				"enabled": true,
				"command": ["docker/bin/lsp", ".venv/bin/ruff", "server"],
				"selector": "source.python"
			},
			"zuban": {
				"enabled": true,
				"command": ["docker/bin/lsp", ".venv/bin/zuban", "server"],
				"selector": "source.python"
			}
		}
	}
}
```

In this case I run the LSP servers inside a docker container, so it's easy to let them "crash" by stopping the container. The "LSP server has crashed" dialog pops up:

image

When pressing cancel, the dialog will reappear **forever** because for project settings, the `_disabled_for_session` method is not consulted:

https://github.com/sublimelsp/LSP/blob/8c77738df6cdff50bf60784f29a271f8e84b033b/plugin/core/configurations.py#L74-L94

With the proposed changes, `_disabled_for_session` will be taken into account so "project only" LSP servers will also be disabled during the session when pressing "Cancel". ~~An informational message will be printed to the console~~ (has been rejected):
image

It is then possible to fire up your container (or do other stuff like installing dependencies, nodejs etc, whatever is needed to start up the LSP server) and then start the LSP via `Tools` -> `LSP` -> `Enable Language Server ...` without restarting Sublime Text.

### DIFF
--- a/plugin/core/configurations.py
+++ b/plugin/core/configurations.py
@@ -80,7 +80,6 @@ class WindowConfigManager:
             else:
                 overrides = {}
             if name in self._disabled_for_session:
-                printf(f"current session has '{name}' disabled")
                 overrides["enabled"] = False
             updated_config = ClientConfig.from_config(config, overrides)
             self.all[name] = updated_config
@@ -90,7 +89,6 @@ class WindowConfigManager:
                 continue
             debug("loading project-only configuration", name)
             if name in self._disabled_for_session:
-                printf(f"current session has '{name}' disabled")
                 config["enabled"] = False
             try:
                 updated_config = ClientConfig.from_dict(name, config)


plugin/core/configurations.py
```diff
@@ -84,12 +84,15 @@ def _reload_configs(self, updated_config_name: str | None = None, notify_listene
             updated_config = ClientConfig.from_config(config, overrides)
             self.all[name] = updated_config
             updated_configs.append(updated_config)
-        for name, c in project_settings.items():
+        for name, config in project_settings.items():
             if updated_config_name and updated_config_name != name:
                 continue
             debug("loading project-only configuration", name)
+            if name in self._disabled_for_session:
+                printf(f"current session has '{name}' disabled")
```
rchl: The other case doesn't print any messages so it seems inconsistent to print here.

mj026: So I could add printing to the other case (which I would prefer so users can actually see via logging what is going on) or remove printing here, what would you like?

rchl: I can see how someone can get confused by forgetting that server crashed and then not understanding why things are not working. So I would be OK with that but I'd wait for @jwortmann's opinion who is more "strict" on logging. :)
Though it would only trigger on a crash and not on trying to start server so not sure how much it would help...

We also have `debug()` logging which depends on a setting being enabled but I guess it doesn't help if most users don't have it enabled.

mj026: I added logging to the "other case" too, so now it is at least consistent.  I'll wait for @jwortmann's opinion regarding this, I can remove it again if desired.

jwortmann: To be honest I would prefer using `debug()`. I guess it's reasonable to turn on debug logging in the settings if something doesn't work as expected and also the popup dialog clearly states how to re-enable the config
https://github.com/sublimelsp/LSP/blob/8c77738df6cdff50bf60784f29a271f8e84b033b/plugin/core/windows.py#L389
so if someone ignores that and still complains about things not working, then I'd say it's their own fault.

Also this would get re-printed everytime any of the configs (including any other config) gets enabled or disabled, so I'm not even sure I would add debug logging for that, at least not for global configs. Imo better would be to add a single `debug()` or maybe `printf()` statement when the error occurs, right before https://github.com/sublimelsp/LSP/blob/8c77738df6cdff50bf60784f29a271f8e84b033b/plugin/core/windows.py#L393

mj026: Well, right before `restart = sublime.ok_cancel_dialog(msg, "Restart")` would IMO not make sense as all the info available _at that moment_ is already shown inside the popup. 

I would like to see why my project LSP is not running and with this debug info I can see that it is disabled for the current session.  But I can already know by reading the popup info so this might be not needed. I'll remove it in a follow up commit.

jwortmann (approved):

PR #2763, opened 2026-02-17
refactor: setup imports formatting

rchl:
Per suggestion in https://github.com/sublimelsp/LSP/pull/2761#discussion_r2809156640, enable import sorting.

Did not include `force-single-line = true` even though it would match our formatting closer because it makes import list longer which is already long enough in many places.

To me it doesn't matter that much how those are formatted as long as I don't have to waste time sorting those myself so I'm open to changing the rules in any way.

### DIFF
--- a/boot.py
+++ b/boot.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-# Please keep this list sorted (Edit -> Sort Lines)
 from .plugin.code_actions import LspCodeActionsCommand
 from .plugin.code_actions import LspRefactorCommand
 from .plugin.code_actions import LspSourceActionCommand


jwortmann:
To be honest, I think I would prefer it with the `force-single-line = true`.

I don't really mind multiple imports on a single line, but the very long import lists like in `sessions.py` or `views.py` just get converted to something like
```python
from ...protocol import (
    CodeAction,
    CodeActionContext,
    CodeActionKind,
    ...
)
```
which doesn't save lines and looks too busy / inconsistent in my opinion when mixed with other imports on single lines before and after that.

mj026:
Great this is now added! 🥳

PR #2764, opened 2026-02-18
Don't replace builtin rename_path command

jwortmann:
It seems like people are not happy about LSP replacing the builtin `rename_path` command when using "Rename…" from the sidebar context menu.

Instead, let's add a separate "LSP: Rename…" entry for now, which is only visible when there is a server with that capability.

### DIFF
--- a/Side Bar.sublime-menu
+++ b/Side Bar.sublime-menu
@@ -1,0 +1,4 @@
+[
+    { "caption": "-", "id": "lsp" },
+    { "caption": "LSP: Rename…", "command": "lsp_rename_path", "args": {"paths": []} }
+]
--- a/boot.py
+++ b/boot.py
@@ -274,13 +274,6 @@ class Listener(sublime_plugin.EventListener):
                     tup[1](None)
                     break
 
-    def on_window_command(
-        self, window: sublime.Window, command_name: str, args: dict[str, Any]
-    ) -> tuple[str, dict[str, Any]] | None:
-        if command_name == "rename_path":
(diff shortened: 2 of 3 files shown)


Side Bar.sublime-menu
```diff
@@ -0,0 +1,4 @@
+[
+    { "caption": "-", "id": "lsp" },
+    { "caption": "LSP: Rename…", "command": "lsp_rename_path", "args": {"paths": []} }
```
rchl, line 3: The caption feels too generic - you can't know if it's for renaming symbols or files unless you see the "rename symbol" also.

rchl: Oh, it's a sidebar menu. NVM.

rchl:
Just noticed two entries in side menu:

Screenshot 2026-02-21 at 11 35 20

Have not looked into the code at all but perhaps it's because I'm running multiple sessions in active file (though if this depends on active file then it shouldn't).

Screenshot 2026-02-21 at 11 38 04

jwortmann:
I don't know how that could happen. Perhaps you saved the `boot.py` file which would trigger a reload of the command, and then maybe ST also kept the old one in memory. But even then it shouldn't matter because there is only one entry in the menu file...

Do you have a reproducible way to make it happen?

rchl:
I've figured it out. It's a local issue.

I've used Claude to add documentation for ClientConfig and it created a git workspace which is seen by ST as an extra copy of LSP in a way. Even gitignoring doesn't help that - ST still picks up extra sublime menu from it.

PR #2765, opened 2026-02-18
Expose unknown settings in ClientConfig and support project overrides

rchl:
An idea to handle https://github.com/sublimelsp/LSP/pull/2739#discussion_r2733324680

Store the whole dict (either from `sublime.Settings` or project overrides) in `ClientConfig` and expose access to it through dot-access (`.foo`). The object will also include all keys that we are already special casing in `ClientConfig` but that shouldn't be a problem because processed variants will shadow those non-processed ones.

This allows accessing any unknown root properties from `LSP-*.sublime-settings` that we normally don't expose from `ClientConfig`, while considering project overrides. Currently plugins use `sublime.load_settings` to access those which means they don't get the benefit of supporting project overrides by default.

The type of a setting gotten that way is Any since we don't know the schema of unknown settings. We could in theory add a generic that could specify schema for unknown settings but I don't think anyone will bother creating and maintaining it (keeping in sync with `sublime-package.json`).

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -1063,9 +1063,9 @@ class ClientConfig:
     def is_disabled_capability(self, capability_path: str) -> bool:
         """Return `True` if the given capability has been disabled in the config.
 
-        Walks :attr:`disabled_capabilities` along `capability_path` (a string such as `"hoverProvider"`). A
-        capability is considered disabled when the value at that path is `True`, or an empty dict (leaf node with no
-        sub-keys).
+        Walks :attr:`disabled_capabilities` along `capability_path` (a string such as `"hoverProvider"` or dotted string
+        like `textDocumentSync.didOpen`). A capability is considered disabled when the value at that path is `True`, or
+        an empty dict (leaf node with no sub-keys).
 
         :param capability_path: Dotted capability path to check.
         """


plugin/core/types.py
```diff
@@ -769,6 +772,25 @@ def __init__(self,
...
+            return self._all_settings[name]
+        raise AttributeError(name)
+
+    def __getitem__(self, key: str) -> Any:
+        """Get property through subscription access (`['foo']`)."""
+        if key in self.__dict__:
+            return self.__dict__[key]
+        if key in self._all_settings:
+            return self._all_settings[key]
+        raise KeyError(key)
```
jwortmann: I wonder should'n it be sufficient to define only one of these ways? Python is not JavaScript :)
Or alternatively instead of these magic methods, we could use a regular method
```python
def get_setting(self, key: str) -> Any:
    return self._all_settings.get(key)
```
That would be a more explicit way to access the settings from the .sublime-settings file, although maybe less elegant.

The dot-access would not work for setting names that also exist as properties of the ClientConfig class (which some of them could be considered private) and this could lead to hard to find bugs, for example if someone defines a custom setting "status_key"...

rchl: > wonder should'n it be sufficient to define only one of these ways? Python is not JavaScript :)

I thought it would be nice to provide dot access to those properties since then from the user perspective it  almost looks like ClientConfig has 1:1 mapping to the original settings. But of course some property names aren't compatible with dot notation so then we need bracket access also.

> Or alternatively instead of these magic methods, we could use a regular method

It's an option but potentially more confusing since how do you then make it clear to the user why those are separated from other options?
Also even in that case I would filter out keys that are already handled by ClientConfig since we don't want user to have access to the non-processed settings (not using DottedDict for example).

> The dot-access would not work for setting names that also exist as properties of the ClientConfig class (which some of them could be considered private) and this could lead to hard to find bugs, for example if someone defines a custom setting "status_key"...

I would guess that would be caught during development of the package.

rchl: I see you are not convinced but from my perspective it's the easiest to understand and use for the user.

I can add some documentation if that helps.

rchl: Added docs.
I wouldn't go as far but AI is really good at doing it so I've let it.

plugin/core/types.py
```diff
@@ -725,6 +725,16 @@ def __init__(
 
 
 class ClientConfig:
+    """Represents the configuration for an LSP client (language server).
```
rchl: One could wonder whether we should call it a configuration for LSP client / Language server / LSP plugin...

plugin/core/types.py
jwortmann: My concern is more that this defines two different ways to do exactly the same thing, which somewhat goes against Python's general rules. In my opinion we should decide for one of them. Otherwise some packages might use dot-access and other packages dict-like access, or even mix both, which would just create inconsistency. I guess I would probably remove the `__getitem__`; the dict-like access would be fine for a class like `DottedDict`, but feels a bit wrong to me here because `ClientConfig` is just a regular class that doesn't really behave like a dict. And if we keep the dot-access to access the settings from the `.sublime-settings` file, I would probably also rename the private property `self.status_key` to `self._status_key` (not sure if there is any other which should be private and not read by packages).

---

> But of course some property names aren't compatible with dot notation so then we need bracket access also.

Property names which are not compatible would be properties with the same name that are already defined on the class, which mostly should match the actually setting names (maybe rename `self.init_options` -> `self.initializationOptions` to match? - btw why is this named `"initializationOptions"` in the settings and not `"initialization_options"`?), and it would also not work for setting names that have a space or other non-ascii character in them, but I have never seen such a setting name and that would go against a strong convention for ST setting names, so that shouldn't really matter.

jwortmann: Can we deprecate `"initializationOptions"` setting and use `"initialization_options"` instead, while keeping backwards compatibility for a while?

rchl: > Property names which are not compatible would be properties with the same name that are already defined on the class

I think you are missing one important case - properties with dashes in them. Or dots but that's probably a lot less likely.

jwortmann: > I think you are missing one important case - properties with dashes in them. Or dots but that's probably a lot less likely.

Okay I have rarely seen settings with dots like `"mdpopups.debug"`, but that's also pretty much against common ST convention, so I would say if a package uses such setting, they should either rename it or fallback to current way of loading via `sublime.load_settings` API.

I believe I have never ever seen a setting name with a dash in it.

Can you point me to any example where dash or dot is used in an LSP helper package?

plugin/core/types.py
jwortmann: I assume Claude got confused by our weird naming convention `ClientConfig` instead of `ServerConfig`. Better don't let it see the `"clients"` key in LSP.sublime-settings, which should have been named `"servers"` in my opinion...

So yes, the docstring could actually just use
```
Represents the configuration for a language server.
```

plugin/core/types.py
rchl: Made a little script to find all non-standard root properties in `LSP-*.sublime-settings` that my ST could see:

```
LSP-Bicep.sublime-settings ['dotnet_executable']
LSP-clangd.sublime-settings ['binary', 'system_binary']
LSP-jdtls.sublime-settings ['test.filterStacktrace', 'version']
LSP-julia.sublime-settings ['julia_executable_path', 'repl_env_variables']
LSP-lemminx.sublime-settings ['java_vmargs', 'server_binary', 'server_binary_args', 'server_version']
LSP-ltex-ls.sublime-settings ['allow_untested', 'version']
LSP-lua.sublime-settings ['locale', 'server_version']
LSP-PowerShellEditorServices.sublime-settings ['powershell_exe', 'version']
LSP-pylsp.sublime-settings ['python_binary']
LSP-TexLab.sublime-settings ['lsp_server_download_url']
LSP-clangd.sublime-settings ['binary', 'system_binary']
LSP-lemminx.sublime-settings ['java_vmargs', 'server_binary', 'server_binary_args', 'server_version']
LSP-ltex-ls.sublime-settings ['allow_untested', 'version']
LSP-lua.sublime-settings ['locale', 'server_version']
LSP-pylsp.sublime-settings ['python_binary']
```

So there is one case (`test.filterStacktrace`) that would be affected but it can be changed I guess. I will update the code to follow your suggestion then.

plugin/core/types.py
```diff
@@ -756,24 +800,50 @@ def __init__(
...
+        return self.initialization_options
+
+    def __getattr__(self, name: str, /) -> Any:
+        """Get property through attribute access (`.foo`) for properties that don't exist natively."""
+        if name in self._all_settings:
+            return self._all_settings[name]
+        raise AttributeError(name)
+
+    def __contains__(self, key: str) -> bool:
+        return key in self._all_settings
```
rchl: Not sure this should also return true for native class properties. It doesn't by default for classes but maybe it's a bit inconsistent with supporting non-native properties.

plugin/core/types.py
rchl: > Can we deprecate `"initializationOptions"` setting and use `"initialization_options"` instead, while keeping backwards compatibility for a while?

I've misunderstood so I've instead deprecated `init_options` but that's probably fine to do also for consistency.

rchl: I don't want to do the deprecation of `initializationOptions` in this PR at least. Something for a separate PR.

plugin/core/types.py
jwortmann: I would remove this function too. This is an interface for list-like or dict-like classes.

If you develop a helper package, you already know if some setting name is defined or not, because you are the one who creates the setting names.

rchl: OK

plugin/core/types.py
```diff
@@ -735,16 +746,49 @@ def __init__(
...
         disabled_capabilities: DottedDict | None = None,
         file_watcher: FileWatcherConfig | None = None,
         semantic_tokens: dict[str, str] | None = None,
         diagnostics_mode: str = "open_files",
-        path_maps: list[PathMap] | None = None
+        path_maps: list[PathMap] | None = None,
+        all_settings: dict[str, Any] | None = None
     ) -> None:
+        """
+        :param name: Unique identifier for this language server client.
```
jwortmann:
```suggestion
        :param name: Unique identifier for this language server.
```

plugin/core/types.py
```diff
@@ -735,16 +746,49 @@ def __init__(
...
+        :param selector: Sublime Text scope selector that determines which views this client is active for (e.g.
```
jwortmann:
```suggestion
        :param selector: Sublime Text scope selector that determines which views this server is active for (e.g.
```

plugin/core/types.py
```diff
@@ -735,16 +746,49 @@ def __init__(
...
+            `"source.python"`).
+        :param priority_selector: Selector used when multiple clients match the same view; the highest-scoring client
```
jwortmann:
```suggestion
        :param priority_selector: Selector used when multiple servers match the same view; the highest-scoring server
```

plugin/core/types.py
```diff
@@ -735,16 +746,49 @@ def __init__(
...
+            takes precedence. Falls back to `selector` when not provided.
+        :param schemes: URI schemes this client handles (e.g. `["file", "untitled"]`). Defaults to `["file"]`.
```
jwortmann:
```suggestion
        :param schemes: URI schemes this client handles (e.g. `["file", "buffer"]`). Defaults to `["file"]`.
```

plugin/core/types.py
```diff
@@ -735,16 +746,49 @@ def __init__(
...
+        :param command: Command and arguments used to launch the language server process.
+        :param tcp_port: Port for TCP transport. `None` uses stdio. `0` picks a free port. Negative values cause LSP to
+            host a TCP server (the language server connects to LSP rather than the other way around); `-1` picks any
+            free port, and `-N` binds to port `N`.
+        :param auto_complete_selector: Scope selector that restricts when auto-complete suggestions are shown. `None`
+             means that the default Sublime Text's value is used.
```
jwortmann:
```suggestion
             means that the value from the Sublime Text setting of the same name is used.
```

plugin/core/types.py
```diff
@@ -735,16 +746,49 @@ def __init__(
...
+        :param enabled: Whether this client is enabled.
```
jwortmann:
```suggestion
        :param enabled: Whether this server is enabled.
```

plugin/core/types.py
```diff
@@ -735,16 +746,49 @@ def __init__(
...
+        :param initialization_options: `initializationOptions` sent to the server during the LSP `initialize` handshake.
+        :param settings: Server-specific settings sent via `workspace/didChangeConfiguration`.
```
jwortmann:
```suggestion
        :param settings: Server-specific settings sent via `workspace/didChangeConfiguration` and for
            `workspace/configuration` requests.
```

plugin/core/types.py
```diff
@@ -756,24 +800,47 @@ def __init__(
...
+        raise AttributeError(name)
 
     @classmethod
     def from_sublime_settings(cls, name: str, s: sublime.Settings, file: str) -> ClientConfig:
+        """Create a ClientConfig from a Sublime Text `Settings` object.
+
+        Plugin-defined defaults are read from `file` (a resource path to the plugin's `.sublime-settings` file) and user
+        overrides are layered on top from `s`.
+
+        :param name: Unique client name.
```
jwortmann:
```suggestion
        :param name: Unique server name.
```

plugin/core/types.py
```diff
@@ -791,19 +858,25 @@ def from_sublime_settings(cls, name: str, s: sublime.Settings, file: str) -> Cli
...
-            path_maps=PathMap.parse(s.get("path_maps"))
+            path_maps=PathMap.parse(s.get("path_maps")),
+            all_settings=s.to_dict()
         )
 
     @classmethod
     def from_dict(cls, name: str, d: dict[str, Any]) -> ClientConfig:
+        """Create a ClientConfig from a plain dictionary.
+
+        :param name: Unique client name.
```
jwortmann:
```suggestion
        :param name: Unique server name.
```

plugin/core/types.py
```diff
@@ -894,16 +986,36 @@ def resolve_transport_config(self, variables: dict[str, str]) -> TransportConfig
         return TransportConfig(self.name, command, tcp_port, env, listener_socket)
 
     def set_view_status(self, view: sublime.View, message: str) -> None:
+        """Update the view status bar entry for this client.
```
jwortmann:
```suggestion
        """Update the view status bar entry for this server.
```

plugin/core/types.py
```diff
@@ -894,16 +986,36 @@ def resolve_transport_config(self, variables: dict[str, str]) -> TransportConfig
...
         if sublime.load_settings("LSP.sublime-settings").get("show_view_status"):
             status = f"{self.name} ({message})" if message else self.name
-            view.set_status(self.status_key, status)
+            view.set_status(self._status_key, status)
         else:
             self.erase_view_status(view)
 
     def erase_view_status(self, view: sublime.View) -> None:
-        view.erase_status(self.status_key)
+        """Remove this client's entry from the view's status bar.
```
jwortmann:
```suggestion
        """Remove this server's entry from the view's status bar.
```

plugin/core/types.py
```diff
@@ -894,16 +986,36 @@ def resolve_transport_config(self, variables: dict[str, str]) -> TransportConfig
...
+
+        :param view: The view whose status bar entry should be cleared.
+        """
+        view.erase_status(self._status_key)
 
     def match_view(self, view: sublime.View, scheme: str) -> bool:
+        """Return `True` if this client should be active for the given view.
```
jwortmann:
```suggestion
        """Return `True` if this server should be active for the given view.
```

plugin/core/types.py
```diff
@@ -931,6 +1060,14 @@ def map_server_uri_to_client_path(self, uri: DocumentUri) -> str:
         return path
 
     def is_disabled_capability(self, capability_path: str) -> bool:
+        """Return `True` if the given capability has been disabled in the config.
+
+        Walks :attr:`disabled_capabilities` along `capability_path` (a dotted string such as `"textDocument.hover"`). A
```
jwortmann: That `textDocument.hover` looks like some hallucination. It would acutally be `hoverProvider`, no?

plugin/core/types.py
```diff
@@ -756,24 +801,47 @@ def __init__(
...
+        self._status_key = f"lsp_{self.name}"
         self.semantic_tokens = semantic_tokens
         self.diagnostics_mode = diagnostics_mode
+        # For accessing configuration keys not explicitly handled above. Accessable through dunder methods below.
+        self._all_settings = all_settings or {}
+
+    @property
+    @deprecated('Use initialization_options instead')
+    def init_options(self) -> DottedDict:
+        return self.initialization_options
```
jwortmann, line 820: This should work as long as nobody tries to assign an entirely new DottedDict object.

I think you recently wrote that some helper package modifies the `configuration.init_options` in `on_pre_start`. I assume that it only modifies the existing DottedDict and doesn't create a new one.

rchl: I don't see anything overwriting it. Only using .set or .update

PR #2767, opened 2026-02-20
Remove deprecated properties of ClientConfig

rchl:
Remove:
- binary_args
- languages
- document_selector
- syntaxes

`languages` is used by LSP-leo and LSP-anakin so will update those.

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -752,8 +752,7 @@ class ClientConfig:
             self.schemes: list[str] = schemes
         else:
             self.schemes = ["file"]
-        if isinstance(command, list):
-            self.command = command
+        self.command = command
         self.tcp_port = tcp_port
         self.auto_complete_selector = auto_complete_selector
         self.enabled = enabled


plugin/core/types.py
```diff
@@ -725,25 +725,26 @@ def __init__(
...
+        init_options: DottedDict | None = None,
+        settings: DottedDict | None = None,
+        env: dict[str, str | list[str]] | None = None,
+        experimental_capabilities: dict[str, Any] | None = None,
+        disabled_capabilities: DottedDict | None = None,
+        file_watcher: FileWatcherConfig | None = None,
+        semantic_tokens: dict[str, str] | None = None,
+        diagnostics_mode: str = "open_files",
+        path_maps: list[PathMap] | None = None
+    ) -> None:
```
rchl, line 747: Removed default, non literal, values since such code can lead to bugs.

```py
class Foo:
    def __init__(self, test: list[str] = []) -> None:
        test.append('a')
        print(test)


f1 = Foo()
f2 = Foo()
```

produces:

```
['a']
['a', 'a']
```

plugin/core/types.py
```diff
@@ -753,18 +754,15 @@ def __init__(self,
             self.schemes = ["file"]
         if isinstance(command, list):
             self.command = command
-        else:
-            assert isinstance(binary_args, list)
-            self.command = binary_args
```
jwortmann: This means that `self.command` is possibly undefined now (if the setting value for it isn't a list, or if the default value from the constructor is used), which I assume would raise an attribute error later.
Probably best to just to remove that check above.

rchl: It was the same before because binary_args defaulted to None.

jwortmann: But it looks like that would raise an AssertionError, which is much easier to debug.
To me it seems like the `isinstance(command, list)` is just useless.

PR #2770, opened 2026-02-20
refactor: Move AbstractPlugin to plugin.api

rchl:
- Move `AbstractPlugin`, `_plugins`, `register_plugin` and `unregister_plugin` to `plugin/api.py`.
- Rename global, internal `_plugins` to `g_plugins` to use better naming convention.

This is groundwork for the new plugin API changes. It:
- makes it easier to handle things (I want to make `register_plugin` and `unregister_plugin` part of LspPlugin class later so that those don't need to be imported separately)
- keeps related things together
- makes `sessions.py` slightly less bloated

This is basically just a move. No deprecated APIs were removed in the process.

Compatibility with existing plugins should be preserved since those moved symbols were exported from `LSP.plugin` already.

### DIFF
--- a/plugin/api.py
+++ b/plugin/api.py
@@ -356,9 +356,9 @@ class AbstractPlugin(APIHandler, ABC):
                      workspace_folders: list[WorkspaceFolder], configuration: ClientConfig) -> str | None:
         """
         Callback invoked just before the language server subprocess is started. This is the place to do last-minute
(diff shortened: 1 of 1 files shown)


rchl:
This is just a simple move but since it's hard to tell due to move, I've prepared this commit that moves the code back to its original place - c382ebf74ee7130a0323dd939641b438e2cc40. It's just to show the actual changes that were made which is mostly just imports and the `_plugins` variable name. Should make it easy to see what has changed.

PR #2771, opened 2026-02-21
Create single LSP status per view

rchl:
Had this idea for a while to combine statuses of each server.

The benefits of that are:
- takes less space if we remove "LSP-" prefix from each server
- status can be sorted by name which means the order is consistent
- there are possibilities to customize formatting per liking (not implemented)

Before:

Screenshot 2026-02-21 at 23 10 13

After:

Screenshot 2026-02-21 at 23 09 45

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -738,12 +738,7 @@ class DefaultViewStatusHandler(ViewStatusHandler):
 
     @override
     def on_view_status_changed(self, config_name: str, view: sublime.View, status: str | None) -> None:
-        status_key = f"lsp_{config_name}"
-        if status is not None and sublime.load_settings("LSP.sublime-settings").get("show_view_status"):
-            status = f"{config_name} ({status})" if status else config_name
-            view.set_status(status_key, status)
-        else:
-            view.erase_status(status_key)
+        pass
 
 
 default_status_view_handler = DefaultViewStatusHandler()


plugin/core/windows.py
```diff
@@ -515,6 +519,27 @@ def on_configs_changed(self, configs: list[ClientConfig]) -> None:
         config_names = [config.name for config in configs]
         sublime.set_timeout_async(lambda: self.restart_sessions_async(config_names))
 
+    # --- Implements ViewStatusHandler ---------------------------------------------------------------------------------
+
+    @override
+    def on_view_status_changed(self, config_name: str, view: sublime.View, status: str | None) -> None:
+        view_id = view.id()
+        if view_id not in self._view_statuses:
+            self._view_statuses[view_id] = {}
```
jwortmann: Could be simplified

```suggestion
        self._view_statuses.setdefault(view_id, {})
```

plugin/core/windows.py
```diff
@@ -515,6 +519,27 @@ def on_configs_changed(self, configs: list[ClientConfig]) -> None:
...
+        if status is None:
```
jwortmann, line 528: I think this doesn't handle empty string correctly, which is currently used in `Session.set_config_status_async` to clear the status message.

rchl: Empty string should just remove the part in parenthesis, not erase the status completely.

jwortmann: Oh right, I see now empty string is handled properly in the else-branch below.

rchl:
MIxed two PRs together so had to clean up and force-push.

plugin/core/types.py
```diff
@@ -725,6 +735,21 @@ def __init__(
         self.listener_socket = listener_socket
 
 
+class DefaultViewStatusHandler(ViewStatusHandler):
+
+    @override
+    def on_view_status_changed(self, config_name: str, view: sublime.View, status: str | None) -> None:
+        status_key = f"lsp_{config_name}"
+        if status is not None and sublime.load_settings("LSP.sublime-settings").get("show_view_status"):
```
jwortmann: It seems like this was already part of the previous code, but is there any reason for reading the settings file manually, rather than using `userprefs().show_view_status`?

jwortmann: By the way, is this `DefaultViewStatusHandler` even actually used? As far as I see, the handler gets set in `start_async`, which should cover all status bar printing.

If the code in this method is actually never run, I would probably replace it with just `pass`.

rchl: > It seems like this was already part of the previous code, but is there any reason for reading the settings file manually, rather than using `userprefs().show_view_status`?

hackaround circular dependency

jwortmann (approved):

PR #2773, opened 2026-02-22
Move SettingsRegistration from g_plugins to ClientConfigs

rchl:
I've extracted this refactor from the "New plugin API" PR to make it easier to review the other one.

This change doesn't have any impact on AbstractPlugin and current code but it will be needed for LspPlugin because it doesn't implement `configuration()` anymore and we can't just get `sublime.Settings` whenever we want.

Now that I'm writing this, I'm wondering if I could have changed `get_plugin` and use it instead but it's almost never better to cheat with globals like that so maybe this is more correct anyway.

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -761,6 +761,7 @@ class ClientConfig:
 
     def __init__(
         self,
+        *,
         name: str,
         selector: str,
         priority_selector: str | None = None,
@@ -823,7 +824,7 @@ class ClientConfig:
             self.schemes: list[str] = schemes
         else:
             self.schemes = ["file"]
-        self.command = command
+        self.command = command or []
         self.tcp_port = tcp_port
         self.auto_complete_selector = auto_complete_selector
         self._enabled = enabled
--- a/tests/test_mocks.py
+++ b/tests/test_mocks.py
@@ -9,7 +9,7 @@ from typing import Any
 from typing import Callable
 
 TEST_CONFIG = ClientConfig(name="test", command=[], selector="text.plain", tcp_port=None)
-DISABLED_CONFIG = ClientConfig("test", command=[], selector="text.plain", tcp_port=None, enabled=False)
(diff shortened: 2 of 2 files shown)


plugin/core/configurations.py
```diff
@@ -46,7 +49,7 @@ def get_config(self, config_name: str) -> ClientConfig | None:
     def get_configs(self) -> list[ClientConfig]:
         return sorted(self.all.values(), key=lambda config: config.name)
 
-    def match_view(self, view: sublime.View, include_disabled: bool = False) -> Generator[ClientConfig, None, None]:
+    def match_view(self, view: sublime.View) -> Generator[ClientConfig, None, None]:
```
rchl, line 49: `include_disabled` was just not used anywhere.

rchl:
Actually maybe the instance of SettingsRegistration should just be a property of `ClientConfig`. I think that would make sense so will try that.

rchl:
Hmm, not entirely happy with the current state... It's not that consistent that for some `ClientConfigs` (plugin-based) `Settings` are updated from `ClientConfig` and for other (LSP.sublime-settings based) from `ClientConfigs._set_enabled`.

rchl:
I don't have good idea how this could be improved further so feel free to review in spare time.

plugin/core/types.py
```diff
@@ -816,10 +824,10 @@ def __init__(
             self.schemes: list[str] = schemes
         else:
             self.schemes = ["file"]
-        self.command = command
+        self.command = command or []
```
rchl, line 827: In practice this was always the case since all 3 factory functions always passed `[]` as a fallback to __init__. Having it explicit here makes resolved type better.

jwortmann (approved):
It is a bit difficult to follow what happens with these partials and self-references, but `g_plugins` looks cleaner now, so I guess we can go with it.

PR #2774, opened 2026-02-22
fix tox config

rchl:
I've accidentally made tox a no-op when migrating its config to pyproject.toml. We actually need to use legacy format for it to work.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,22 +13,22 @@ reportUnusedCallResult = "none"
 stubPath = "./stubs"
 
 [tool.tox]
-legacy_tox_ini = """
-[tox]
-env_list = py3
+env_list = ["py3"]
 skipsdist = true
 
-[testenv]
-deps =
-    mypy==1.18.2
-    pyright==1.1.408
-    ruff==0.15.0
-    orjson==3.11.4
-commands =
-    mypy stubs
-    pyright plugin stubs
-    ruff check
-"""
+[tool.tox.env_run_base]
+deps = [
+    "mypy==1.18.2",
+    "pyright==1.1.408",
+    "ruff==0.15.0",
+    "orjson==3.11.4",
+]
+commands = [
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -13,22 +13,22 @@ reportUnusedCallResult = "none"
 stubPath = "./stubs"
 
 [tool.tox]
-env_list = ["py3"]
+legacy_tox_ini = """
```
predragnikolic: Ok, this is what I saw in their docs. https://tox.wiki/en/4.14.1/config.html#pyproject-toml

When I saw the previous PR I assumed you found a more modern way to configure tox in pyproject.toml

jwortmann: Native TOML config should be possible: https://tox.wiki/en/4.44.0/reference/config.html#pyproject-toml-native

This seems to work:
```diff
diff --git a/pyproject.toml b/pyproject.toml
index 8e55db39..7965d426 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ stubPath = "./stubs"
 env_list = ["py3"]
 skipsdist = true

-[tool.tox.testenv]
+[tool.tox.env.py3]
 deps = [
     "mypy==1.18.2",
     "pyright==1.1.408",
```

jwortmann: Or alternatively
```diff
diff --git a/pyproject.toml b/pyproject.toml
index 8e55db39..ddbfe7a7 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ stubPath = "./stubs"
 env_list = ["py3"]
 skipsdist = true

-[tool.tox.testenv]
+[tool.tox.env_run_base]
 deps = [
     "mypy==1.18.2",
     "pyright==1.1.408",
```

rchl: I had it like that initially and after testing I've changed to one that matches original config as I thought it doesn't matter. ;)

PR #2775, opened 2026-02-22
Minor code improvements

jwortmann:
Just a few small improvements that I found.


plugin/documents.py
```diff
@@ -870,7 +868,7 @@ def _do_highlights_async(self) -> None:
             return
         point = region.b
         if session := self.session_async("documentHighlightProvider", point):
-            params = cast(DocumentHighlightParams, text_document_position_params(self.view, point))
+            params: DocumentHighlightParams = {**text_document_position_params(self.view, point)}
```
rchl, line 871: `{**text_document_position_params(self.view, point)}` could be inlined in the Request line to save a line

jwortmann: True, but then it's not so clear why this odd syntax `{**...}` is used. I will leave it like this for now.

PR #2776, opened 2026-02-22
Implement Document on Type Formatting

rchl:
Add support for [Document on Type Formatting Request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_onTypeFormatting).

For this functionality it's essential that we ignore edits triggered by undo/redo/paste since in that case formatting shouldn't trigger. Of course ST doesn't tell us explicitly how the edit was triggered so I'm listening for those specific commands and annotating the next TextChange with that command name. This should be pretty reliable given that we only consider change to be triggered by that command if it happened "right after" the command (within next tick).

LSP-basedpyright implements this functionality to automatically add `f` in front a string when typing `{`. It advertises this capability:

```json
    "documentOnTypeFormattingProvider": {
      "firstTriggerCharacter": "{"
    },
```

Some thoughts that I had while implementing this:
- When ST auto-inserts matching brace then we don't trigger the request since the last character is not `{`. Spec mentions that the reported `ch` should not necessarily be the last character for cases exactly like this but implementing such logic without some help from ST would rely on heuristics so not sure we should do it. Not handling it makes the feature a lot less useful though as ST often does match brace depending on where in the string we insert it.
- Should we apply text edits without version even if document version has changed since the request was triggered (note that LSP-basedpyright returns TextEdits without version)? It should be safe in this particular case of inserting `f` but might not be safe in general. We currently ignore the edit if document has changed.
- I've tried this feature with the default debouncing we do before notifying didChange and the experience was not good - it took too long to insert an edit (and it was too easy to drop the edit if something was typed in the meantime). So in this change, if we detect we should trigger a request, we purge the changes and trigger request immediately instead of waiting for debounce.

### DIFF
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -611,9 +611,12 @@ class SessionBuffer:
         if not mgr or mgr.should_ignore_diagnostics(self._last_known_uri, self.session.config):
             return
         if version < view.change_count():
+            # If the document content changed in the meanwhile, new diagnostic requests will automatically be triggered
+            # from _on_after_change_async after the didChange notification.
             return
         for identifier in get_diagnostics_identifiers(self.session, view):
             self._do_document_diagnostic_async(view, identifier, version, forced_update=forced_update)
+        self.document_diagnostic_needs_refresh = False
 
     def _do_document_diagnostic_async(
         self, view: sublime.View, identifier: DiagnosticsIdentifier, version: int, *, forced_update: bool = False
@@ -669,8 +672,8 @@ class SessionBuffer:
                     DOCUMENT_DIAGNOSTICS_RETRIGGER_DELAY
                 )
 
-    def set_document_diagnostic_pending_refresh(self, needs_refresh: bool = True) -> None:
-        self.document_diagnostic_needs_refresh = needs_refresh
+    def set_document_diagnostic_pending_refresh(self) -> None:
+        self.document_diagnostic_needs_refresh = True
 
     # --- textDocument/publishDiagnostics ------------------------------------------------------------------------------
 
@@ -813,6 +816,7 @@ class SessionBuffer:
             }, view)
             self.semantic_tokens.pending_response = self.session.send_request_async(
                 request, self._on_semantic_tokens_async, self._on_semantic_tokens_error_async)
+        self.semantic_tokens.needs_refresh = False
 
     def _on_semantic_tokens_async(self, response: SemanticTokens | None) -> None:
         self.semantic_tokens.pending_response = None
@@ -905,8 +909,8 @@ class SessionBuffer:
         for region_key in self.semantic_tokens.active_region_keys:
             view.erase_regions(f"lsp_semantic_{session_name}_{region_key}")
 
-    def set_semantic_tokens_pending_refresh(self, needs_refresh: bool = True) -> None:
-        self.semantic_tokens.needs_refresh = needs_refresh
+    def set_semantic_tokens_pending_refresh(self) -> None:
+        self.semantic_tokens.needs_refresh = True
 
     def get_semantic_tokens(self) -> list[SemanticToken]:
         return self.semantic_tokens.tokens
@@ -931,6 +935,7 @@ class SessionBuffer:
             "range": entire_content_range(view)
         }
         self.session.send_request_async(Request.inlayHint(params, view), self._on_inlay_hints_async)
+        self.inlay_hints_needs_refresh = False
 
     def _on_inlay_hints_async(self, response: list[InlayHint] | None) -> None:
         if response:
@@ -945,8 +950,8 @@ class SessionBuffer:
     def present_inlay_hints(self, phantoms: list[sublime.Phantom]) -> None:
         self._inlay_hints_phantom_set.update(phantoms)
 
-    def set_inlay_hints_pending_refresh(self, needs_refresh: bool = True) -> None:
-        self.inlay_hints_needs_refresh = needs_refresh
+    def set_inlay_hints_pending_refresh(self) -> None:
+        self.inlay_hints_needs_refresh = True
 
     def remove_inlay_hint_phantom(self, phantom_uuid: str) -> None:
         new_phantoms = list(filter(
@@ -973,6 +978,7 @@ class SessionBuffer:
                 break
         request = Request('textDocument/codeLens', {'textDocument': text_document_identifier(view)}, view)
         self.session.send_request_async(request, partial(self._on_code_lenses_async, view))
+        self.code_lenses_needs_refresh = False
 
     def _on_code_lenses_async(self, view: sublime.View, response: list[CodeLens] | None) -> None:
         self._code_lenses.handle_response_async(response or [])
@@ -1009,8 +1015,8 @@ class SessionBuffer:
         for sv in self.session_views:
             sv.handle_code_lenses_async(supported_code_lenses)
 
-    def set_code_lenses_pending_refresh(self, needs_refresh: bool = True) -> None:
-        self.code_lenses_needs_refresh = needs_refresh
+    def set_code_lenses_pending_refresh(self) -> None:
+        self.code_lenses_needs_refresh = True
 
     # ------------------------------------------------------------------------------------------------------------------
 
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -765,7 +765,7 @@ class SessionBufferProtocol(Protocol):
     def do_semantic_tokens_async(self, view: sublime.View) -> None:
         ...
 
-    def set_semantic_tokens_pending_refresh(self, needs_refresh: bool = ...) -> None:
+    def set_semantic_tokens_pending_refresh(self) -> None:
         ...
 
     def get_semantic_tokens(self) -> list[SemanticToken]:
@@ -777,7 +777,7 @@ class SessionBufferProtocol(Protocol):
     def do_inlay_hints_async(self, view: sublime.View) -> None:
         ...
 
-    def set_inlay_hints_pending_refresh(self, needs_refresh: bool = ...) -> None:
+    def set_inlay_hints_pending_refresh(self) -> None:
         ...
 
     def remove_inlay_hint_phantom(self, phantom_uuid: str) -> None:
@@ -789,13 +789,13 @@ class SessionBufferProtocol(Protocol):
     def do_document_diagnostic_async(self, view: sublime.View, version: int, *, forced_update: bool = ...) -> None:
         ...
 
-    def set_document_diagnostic_pending_refresh(self, needs_refresh: bool = ...) -> None:
+    def set_document_diagnostic_pending_refresh(self) -> None:
         ...
 
     def do_code_lenses_async(self, view: sublime.View) -> None:
         ...
 
-    def set_code_lenses_pending_refresh(self, needs_refresh: bool = True) -> None:
+    def set_code_lenses_pending_refresh(self) -> None:
         ...
 
 
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -431,20 +431,16 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
             self._do_code_actions_async()
         for sb in self.session_buffers_async():
             if sb.code_lenses_needs_refresh:
-                sb.set_code_lenses_pending_refresh(needs_refresh=False)
                 sb.do_code_lenses_async(self.view)
             if sb.document_diagnostic_needs_refresh:
-                sb.set_document_diagnostic_pending_refresh(needs_refresh=False)
                 sb.do_document_diagnostic_async(self.view, self.view.change_count(), forced_update=True)
             if sb.semantic_tokens.needs_refresh \
                     and (session_view := sb.session.session_view_for_view_async(self.view)) \
                     and session_view.get_request_flags() & RequestFlags.SEMANTIC_TOKENS:
-                sb.set_semantic_tokens_pending_refresh(needs_refresh=False)
                 sb.do_semantic_tokens_async(self.view)
             if sb.inlay_hints_needs_refresh \
                     and (session_view := sb.session.session_view_for_view_async(self.view)) \
                     and session_view.get_request_flags() & RequestFlags.INLAY_HINT:
-                sb.set_inlay_hints_pending_refresh(needs_refresh=False)
                 sb.do_inlay_hints_async(self.view)
 
     @requires_session


plugin/core/views.py
```diff
@@ -288,7 +290,7 @@ def render_text_change(change: sublime.TextChange) -> TextDocumentContentChangeE
 
 
 def did_change_text_document_params(
-    view: sublime.View, version: int, changes: Iterable[sublime.TextChange] | None = None
+    view: sublime.View, version: int, changes: list[sublime.TextChange] | None = None
```
rchl, line 294: Changed to `list` because it's actually a `list` (as ST API documentation states). Also `Iterable` was problematic since it doesn't support `len()` and we had to anyway convert it to `list` to be able to do the check.

plugin/session_buffer.py
```diff
@@ -382,6 +396,31 @@ def _cancel_pending_requests_async(self) -> None:
...
+    def _get_on_type_formatting_params_async(
+        self, view: sublime.View, changes: list[sublime.TextChange]
+    ) -> DocumentOnTypeFormattingParams | None:
+        selection = first_selection_region(view)
+        if len(changes) == 0 or selection is None:
+            return None
+        capability: DocumentOnTypeFormattingOptions | None = self.get_capability('documentOnTypeFormattingProvider')
+        last_char = changes[-1].str[-1] if changes[-1].str else None
+        if capability and last_char and (
+            last_char == capability['firstTriggerCharacter'] or last_char in capability.get('moreTriggerCharacter', [])
```
jwortmann: I wonder what is the idea behind having such a weird definition in the specs:
```typescript
export interface DocumentOnTypeFormattingOptions {
	/**
	 * A character on which formatting should be triggered, like `{`.
	 */
	firstTriggerCharacter: string;

	/**
	 * More trigger characters.
	 */
	moreTriggerCharacter?: string[];
}
```

Can't we just collect them like
```python
on_type_formatting_trigger_chars = set((capability['firstTriggerCharacter'], *capability.get('moreTriggerCharacter', [])))
```

and store that as a variable on the SessionBuffer? Then we don't need a complicated check like this one here.

predragnikolic: https://github.com/microsoft/language-server-protocol/issues/206

jwortmann:
> * Should we apply text edits without version even if document version has changed since the request be triggered (note that LSP-basedpyright returns TextEdits without version)?

The response is of type `TextEdit[]`, and `TextEdit` never has a version. So the only thing we can do is to track the version from when the request was sent, and hope that the response arrives and is handled before the next change. It looks like this is what you already do here.

What do you think about making `SessionBuffer.purge_changes_async` return its `version`, and use that in the subsequent onTypeFormatting request? That could eliminated a rare case where the document version changed between sending didChange and the onTypeFormatting request (because it runs in `on_text_changed_async` on the async thread), in which case the server would calculate the edits for the version that is not the current one anymore, but we miss that version being outdated already.

plugin/session_buffer.py
rchl: If we'd want to store then we then need to make sure to set it in case it's registered dynamically.
Is it really worth to complicate it just to have simpler if check?

rchl:
> What do you think about making `SessionBuffer.purge_changes_async` return its `version`, and use that in the subsequent onTypeFormatting request? That could eliminated a rare case where the document version changed between sending didChange

I've just changed the argument to existing `change_count` that was in scope which should cover this case.
Not sure it's necessary to change `purge_changes_async` for this use case.

plugin/session_buffer.py
rchl: well, OK, maybe it's fine

plugin/session_buffer.py
```diff
@@ -563,6 +566,12 @@ def _update_supported_commands(self) -> None:
         self._supported_commands = set(self.session.get_capability('executeCommandProvider.commands') or [])
         self._supported_commands.update(itertools.chain.from_iterable(self._dynamically_registered_commands.values()))
 
+    def _update_on_type_formatting_trigger_characters(self) -> None:
+        capability: DocumentOnTypeFormattingOptions | None = self.get_capability('documentOnTypeFormattingProvider')
+        if capability:
+            self._on_type_formatting_trigger_characters = set(capability['firstTriggerCharacter'],
+                                                              *capability.get('moreTriggerCharacter', []))
```
jwortmann: I assume it doesn't work this way because the `set()` constructor expects an iterable, like in my code example from above.

jwortmann: Or you could use literal set syntax if you prefer that though:
```python
{capability['firstTriggerCharacter'], *capability.get('moreTriggerCharacter', [])}
```

rchl: when breaking into multiple lines that new syntax looks like a dict so not a fan in this case

rchl:
> * When ST auto-inserts matching brace then we don't trigger the request since the last character is not `{`. Spec mentions that the reported `ch` should not necessarily be the last character for cases exactly like this but implementing such logic without some help from ST would rely on heuristics so not sure we should do it. Not handling it makes the feature a lot less useful though as ST often does match brace depending on where in the string we insert it.

I feel like this should be implemented actually. The whole point of this feature (at least in case of basedpyright) is kinda lost without it.

VSCode handles it, Zed doesn't.

jwortmann:
I agree that it would be very useful, and the specs exlicitly mention the automatic brace completion, so I think we should try to support this.

You could also check the penultimate character position via `last_str[-2]` if `last_str` is long enough.

Maybe something like

```python
if last_str := changes[-1].str:
    if (ch := last_str[-1]) in self._on_type_formatting_trigger_characters \
            or len(last_str) >= 2 and (ch := last_str[-2]) in self._on_type_formatting_trigger_characters:
        return {
            **text_document_position_params(view, selection.a),
            'options': formatting_options(view.settings()),
            'ch': ch,
        }
```

I haven't tested whether that works or if there is a way to simplify that more. Ideally we want to keep this `_get_on_type_formatting_params_async` method as efficient as possible, because unlike most other features it is not debounced and is called on every key press.

rchl:
I would actually try to make it a bit more specific and less error-prone by only checking second-to-last character if the last two characters match a specific pair ([], (), '', "" etc). But yes, I think it's doable.

jwortmann:
> I would actually try to make it a bit more specific and less error-prone by only checking second-to-last character if the last two characters match a specific pair ([], (), '', "" etc). But yes, I think it's doable.

Yes, sounds like a useful idea.

Another aspect that could be used is that such matched chars are always (I believe) inserted via `insert_snippet` command by ST. Maybe we could make use of that, since you already listen for some special commands and pass that as the `action` argument.

rchl:
> Another aspect that could be used is that such matched chars are always (I believe) inserted via `insert_snippet` command by ST. Maybe we could make use of that, since you already listen for some special commands and pass that as the `action` argument.

That's true but it might be overkill in this case and even less functional because there can be cases where user just inserts `{}` manually fast (in places where auto-matching doesn't trigger).

jwortmann:
I would suggest that we add a global setting `"format_on_type": true | false`, which should be `false` by default, because I expect it to be quite controversial if the line where you are typing is automatically formatted, or if there are suddenly things like the f-string added whenever you type a `{` in a string. So this should probably be opt-in.

I know that we can still always disable features via `disabled_capabilities`, but this is not very intuitive for users and you either need to have in-depth LSP knowledge to know the correct provider name, or do some research first on our docs website and in the LSP specs, to find out where and how exactly to turn it off. And you would need to disable it for every server config individiually.

---

By the way, similarly I am thinking about adding a new setting `"show_signature_help": true | false`, which is `true` by default. Most other editors seems to have a setting for that (e.g. `"editor.parameterHints.enabled"` in VSCode), and every few months someone on the forum creates a new topic and asks how to disable it. I think it would be good to have settings for all features which probably not everybody wants to use, that easily allow to disable them (we already have that for most things: `"show_inlay_hints"`, `"semantic_highlighting"`, `"show_code_actions": ""`, `"show_diagnostics_severity_level": 0`, `"document_highlight_style": ""`, etc).

rchl:
The counterargument for adding a setting is that in case of basedpyright it provides a setting for enabling this functionality and it's not enabled by default. So I would say that it would be more confusing for user to find this basedpyright setting, enable it and see that nothing is happening.

If we're adding a setting then I would be more for having it enabled by default. I believe that it would be better for the user to notice the functionality and look for how to disable it if it's not wanted than to never notice it and potentially miss out on something useful. (That said, it wouldn't apply to basedpyright anyway since its setting is disabled by default).

jwortmann:
> The counterargument for adding a setting is that in case of basedpyright it provides a setting for enabling this functionality and it's not enabled by default. So I would say that it would be more confusing for user to find this basedpyright setting, enable it and see that nothing is happening.

A lot of servers provide settings to enable/disable various features like inlay hints or semantic highlighting on the servers side. So that is already the case for those features, and I think it is reasonable for a new user who freshly installed LSP to go through all settings in the LSP settings file and configure it to their liking. I would take the fact that it's disabled in basedpyright by default also as a hint that this feature can be controversal (personally I think I will not use it, because I don't like the editor interfering too much in what I typed) and should rather be opt-in. Note that we have `"lsp_format_on_paste"` also disabled by default.

I'm planning a moderate revision of the features page of the docs website, and I could add an info box there about that format-on-type has to be enabled in the user settings.

That said, in case everyone other than me would be in favor of having it enabled by default, I would give in and say having a setting for it that is enabled by default is still better than having no setting at all.

P.S: I was contemplating whether it should be `"format_on_type"` or rather `"lsp_format_on_type"` and also be supported in project-specific settings. But I came to the conclusion that imo the regular `"format_on_type"` is better, because the project-specific settings are most useful for projects with certain formatting or code style rules, typically enforced via some configuration file that is commited to the repository. While this format-on-type feature is more of a small quality-of-life utility provided by the editor, and not really specific to project configurations. In retrospective I would even say `"lsp_format_on_paste"` should also better have been just a non-project-specific `"format_on_paste"`, because it's more of a user-specific utility rather than project-specific (if the project is configured with formatting rules and you want to enforce that, the `"lsp_format_on_save"` would cover the formatting for the entire file when saving anyway).

plugin/core/views.py
```diff
@@ -69,6 +70,7 @@
 import sublime_plugin
 import tempfile
 
+ChangeEventAction = Literal['paste', 'redo', 'undo', 'type']
```
predragnikolic: Should 'cut' be in the list?

rchl:
@predragnikolic would you like to add something? :)

predragnikolic:
Tomorrow in the evening I will try to find some time to have a look at this.

plugin/core/views.py
rchl: yes!

plugin/core/constants.py
```diff
@@ -40,6 +40,8 @@
 - on Linux: ~/.cache/sublime-text
 """
 MARKO_MD_PARSER_VERSION: str | None = marko_version
+# Character sets that are commonly auto-paired by ST.
+AUTO_PAIR_ITEMS = ('{}', '()', '[]', "''", '""')
```
predragnikolic: ` is sometimes (but not commonly) auto-paired

plugin/core/types.py
```diff
@@ -285,14 +286,17 @@ def r(name: str, default: bool | int | str | list | dict) -> None:
...
             setattr(self, name, val if isinstance(val, default.__class__) else default)
 
+        r("completion_insert_mode", 'insert')
         r("diagnostics_additional_delay_auto_complete_ms", 0)
         r("diagnostics_delay_ms", 0)
         r("diagnostics_gutter_marker", "dot")
         r("diagnostics_panel_include_severity_level", 4)
         r("disabled_capabilities", [])
         r("document_highlight_style", "underline")
+        r("format_on_type", True)
```
predragnikolic: `false` would be safer for the following reason:

While the `basepyright` server gets the DX right, add an `f` when a `{` is typed in a string...

What will happen when some server gets the DX wrong? (and adds annoying or wrong text edits during typing)
Imagine a user typing and he sees unexpected text edits as he types?  
Will the user know that he needs to turn off `format_on_type`?

By making this an opt-in feature, users are more likely to associate any unexpected behavior with the specific setting they enabled.

LSP.sublime-settings
```diff
@@ -15,6 +15,10 @@
   // "settings" section of project files.
   "lsp_format_on_save": false,
 
+  // Enable on-type-formatting functionality that allows language server
```
predragnikolic:
```suggestion
  // Enable on-type-formatting functionality that allows a language server
```

sublime-package.json
```diff
@@ -422,6 +422,11 @@
               "default": 2000,
               "deprecationMessage": "Use the \"on_save_task_timeout_ms\" setting instead."
             },
+            "format_on_type": {
+              "type": "boolean",
+              "default": true,
+              "markdownDescription": "Enable on-type-formatting functionality that allows language server to format parts of the document during typing."
```
predragnikolic:
```suggestion
              "markdownDescription": "Enable on-type-formatting functionality that allows a language server to format parts of the document during typing."
```

plugin/session_buffer.py
```diff
@@ -349,27 +363,33 @@ def should_notify_did_save(self) -> tuple[bool, bool]:
...
+        if self._pending_changes is None:
+            self._pending_changes = PendingChanges(change_count, changes)
+            purge = True
+        elif self._pending_changes.version < change_count:
+            self._pending_changes.update(change_count, changes)
+            purge = True
+        if purge:
+            self._cancel_pending_requests_async()
+            if userprefs().format_on_type and action == 'type' and \
+                    (params := self._get_on_type_formatting_params_async(view, last_change)):
```
predragnikolic, line 386: textDocument/onTypeFormatting can be fired on `redo`

Steps to reproduce 
1. open any `.py` file
1. type `a = ""`
1. type `{`, the state will be `a = "{}"`
1. as expected textDocument/onTypeFormatting is called, the state will be `a = f"{}"`
1. press undo, the state will be `a = "{}"`
1. press undo again, the state will be `a = ""`
2. when I press redo

I expect the state to be:
`a = "{}"`
But the state is: 
`a = f"{}"`

https://github.com/user-attachments/assets/af479efb-83df-4a2b-9bf8-e9cfcf37e86b

predragnikolic: if I understood correctly onTypeFormatting should only be fired when the user types.

It should note be called when: 'cut', 'paste', 'redo', 'undo'

predragnikolic: Additionally, should format on type not be called when a view is reloaded or reverted (on_reload, on_revert api calls https://www.sublimetext.com/docs/api_reference.html)?

plugin/core/types.py
rchl: It's 2 against 1 so I guess I have to give up.

In that case, to avoid users not understanding why the feature is not working after enabling, I guess we will have to rely on extra notes in settings.

plugin/session_buffer.py
rchl: > textDocument/onTypeFormatting can be fired on `redo`
> 
> Steps to reproduce
> 
>     1. open any `.py` file
> 
>     2. type `a = ""`
> 
>     3. type `{`, the state will be `a = "{}"`
> 
>     4. as expected textDocument/onTypeFormatting is called, the state will be `a = f"{}"`
> 
>     5. press undo, the state will be `a = "{}"`
> 
>     6. press undo again, the state will be `a = ""`
> 
>     7. when I press redo
> 
> 
> I expect the state to be: `a = "{}"` But the state is: `a = f"{}"`
> Clipboard-20260228-213721-496.mp4

Hmm, it doesn't reproduce here.

Would you be able to log what happens after redo by adding print in:

```py
    def on_text_changed_async(
        self, view: sublime.View, change_count: int, changes: list[sublime.TextChange], action: ChangeEventAction
    ) -> None:
        print(action, changes)
```

rchl: > Additionally, should format on type not be called when a view is reloaded or reverted (on_reload, on_revert api calls https://www.sublimetext.com/docs/api_reference.html)?

It won't now because we don't get text change events in this case (`on_text_changed_async` is not called).

rchl: Perhaps when logging also enable command logging with `sublime.log_commands()`.

plugin/core/types.py
predragnikolic: by setting it to false, we are decreasing the discoverability.

we can increase the discoverability for this feature by adding docs for it, like -> https://lsp.sublimetext.io/features/#formatting and provide basepyright 

We usually put features at the top of the release notes, that helps a bit.

plugin/session_buffer.py
predragnikolic: Screenshot 2026-02-28 at 23 19 05

looks like the command is `redo_or_repeat`

predragnikolic: https://docs.sublimetext.io/
Screenshot 2026-02-28 at 23 20 25

Ah it seems like there are multiple variants to `undo` and `redo`

rchl: I suppose I should catch more commands:

OSX:

```
	{ "keys": ["super+z"], "command": "undo" },
	{ "keys": ["super+shift+z"], "command": "redo" },
	{ "keys": ["super+y"], "command": "redo_or_repeat" },
	{ "keys": ["super+u"], "command": "soft_undo" },
	{ "keys": ["super+shift+u"], "command": "soft_redo" },

	{ "keys": ["super+x"], "command": "cut" },
	{ "keys": ["super+c"], "command": "copy" },
	{ "keys": ["super+v"], "command": "paste" },
	{ "keys": ["super+shift+v"], "command": "paste_and_indent" },
	{ "keys": ["super+k", "super+v"], "command": "paste_from_history" },
	{ "keys": ["super+option+v"], "command": "paste_from_history" },
```

predragnikolic:
Would be nice if docs were added for this feature. (does not have to be in this PR)

plugin/session_buffer.py
rchl: I'm not able to handle `paste_from_history`. We get notified about it in `on_text_command` but at that point it just opens the popup menu. Selecting an item to paste triggers `paste` command but in that specific case we are actually not notified about it in `on_text_command`.

predragnikolic: Maybe listen on on_window_command for the paste_from_history, maybe it is a window command.

Edit: And this is the moment I realized that my suggestion is wrong, and that I should prepare for sleep 🙂

plugin/session_buffer.py
```diff
@@ -717,6 +737,42 @@ def has_latest_diagnostics(self) -> bool:
...
+        self, view: sublime.View, last_change: sublime.TextChange
+    ) -> DocumentOnTypeFormattingParams | None:
+        if not self._on_type_formatting_triggers:
+            return None
+        if not (self._get_request_flags(view) & RequestFlags.ON_TYPE_FORMATTING):
+            return None
+        selection = first_selection_region(view)
+        if selection is None:
+            return None
+        if trigger := next((t for t in self._on_type_formatting_triggers if last_change.str.endswith(t)), None):
```
jwortmann: This creates a new generator object on every key press, just to immediately take the first element. I believe it would be more efficient to just use a regular for-loop instead.

rchl: Not sure what you mean by always taking the first element. It takes one that matches end of the change.

I can change to a loop but it won't look as concise.

As for performance, I would also guess that generators are slower but I have no data to prove one way or another.

EDIT: Never mind, it looks clean enough :)

plugin/core/constants.py
rchl: Actually I've just realized that we don't need to have pairs for identical characters in the list because in that case the auto-added character will still match the trigger. We only need those that have opposing characters.

plugin/session_buffer.py
jwortmann: I mean `next` uses the first item of the generator, and then the generator is not used anymore. That is a pretty useless use of generators :)

plugin/session_buffer.py
rchl: I would say this is not fixable right now in a clean and generic way.

Perhaps `paste_from_history` already rewrites the comment internally or something.

Personally I didn't even know this feature existed until now. :)

predragnikolic: Spent some time to understand why `paste_from_history` doesn't work.

Lets see `paste_from_history` implementation:

```py
class PasteFromHistoryCommand(sublime_plugin.TextCommand):
    def run(self, edit):
        if self.view.settings().get('is_widget'):
            return

        # provide paste choices
        paste_list = g_clipboard_history.get()
        keys = [x[0] for x in paste_list]
        self.view.show_popup_menu(keys, lambda choice_index: self.paste_choice(choice_index))

    def is_enabled(self):
        return not g_clipboard_history.empty()

    def paste_choice(self, choice_index):
        if choice_index == -1:
            return
        # use normal paste command
        text = g_clipboard_history.get()[choice_index][1]

        # rotate to top
        g_clipboard_history.push_text(text)

        sublime.set_clipboard(text)
        self.view.run_command("paste")

```

The only difference between `paste_from_history` and all other commands, 
** is that other commands don't invoke other commands**. 

---

`paste_from_history` is the only command that invokes a `paste` command -> (see last line in the implementation):

Why does this matter? 

set_last_edit_action` relies on `sublime.set_timeout`.

When the 'cut' command is called 
`set_last_edit_action` is set to `cut`,
and `sublime.set_timeout` will correctly reset the `type` after the `cut` command was executed. 

When the 'paste_from_history' command is executed 
`set_last_edit_action` is set to `paste`,
the command `paste` is executed, 
but it looks like `sublime.set_timeout(self._reset_last_edit_action)` is called after the `paste` command thus it resets `_last_edit_action` to `type`.


```diff
    def set_last_edit_action(self, action: ChangeEventAction) -> None:
        self._last_edit_action = action
        # ST should have already scheduled text_change event internally so resetting it from a timeout ensures it's
        # reset after the event was triggered and also in case the change event didn't trigger.
-        sublime.set_timeout(self._reset_last_edit_action) # resets to type to early if the commands invokes a different command
+        sublime.set_timeout(self._reset_last_edit_action, 200)
```

predragnikolic: Second note `on_text_command` will only catch commands that were invoked via keyboard shortcuts, command palette, menu item.

Commands that are invoked programmatically, like `self.view.run_command("paste")` will not be picked up by on_text_command`. (which is the expected ST behavior)

rchl: Are you suggesting to do that change? That's just asking for trouble. I have on purpose made it so that it resets itself after one "tick".

The main issue is that the `paste` command doesn't trigger `on_text_command` in this specific case. Perhaps it's some safety measure to avoid infinite loops.

Also I would assume that to cover this case the delay would have to be longer than the amount of time the user takes to select an item from the popup. Which is not very reasonable to do.

EDIT: I see about `on_text_command`. Still, I think it's the best we can do for now. False triggers could in theory happen but at the same time the chances are quite low.

predragnikolic: I think we are fine with not addressing `paste_from_history`.

predragnikolic (approved):

PR #2784, opened 2026-02-28
Fix some more ruff issues

rchl:


plugin/core/open.py
```diff
@@ -160,12 +160,11 @@ def open_in_browser(uri: str) -> None:
...
 
 
-def open_externally(uri: str, take_focus: bool) -> bool:
+def open_externally(uri: str) -> bool:
     """
     A blocking function that invokes the OS's "open with default extension"
     """
     try:
-        # TODO: handle take_focus
         if ST_PLATFORM == "windows":
```
rchl, line 168: Spec says that the `takeFocus` parameter is not relevant for external URIs.

jwortmann (approved):

PR #2785, opened 2026-02-28
Fix pull diagnostics after reload/revert

rchl:
Fixed the bug reported initially in https://github.com/sublimelsp/LSP/pull/2743#issuecomment-3959101654

When file is changed externally and either ST window is not active or that file is not the active one then on activating the window/file, we often didn't request updated diagnostics.

Depending on timing, sometimes (often) diagnostic request triggered from activation event before reload event and then when reload came later we ignored diagnostic request as version was already updated when initial diagnostics request triggered.

Took a bit of a hammer approach and just forced diagnostic request from reload/revert handler. That case should not happen frequently so I don't think we need to optimize this.

### DIFF
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -414,8 +414,6 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         session_views = self.session_views_async()
         if not session_views:
             return
-        if userprefs().show_code_actions and self._code_actions_for_selection_needs_refresh:
-            self._do_code_actions_for_selection_async()
         for sb in self.session_buffers_async():
             if sb.code_lenses_needs_refresh:
                 sb.do_code_lenses_async(self.view)
@@ -429,6 +427,8 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
                     and (session_view := sb.session.session_view_for_view_async(self.view)) \
                     and session_view.get_request_flags() & RequestFlags.INLAY_HINT:
                 sb.do_inlay_hints_async(self.view)
+        if userprefs().show_code_actions and self._code_actions_for_selection_needs_refresh:
+            self._do_code_actions_for_selection_async()
 
     @requires_session
     def on_selection_modified_async(self) -> None:


rchl:
Thinking about it more, it is buggy that we send request from `on_activated_async` with latest view version but we don't send `didChange` because changes have not arrived yet. Maybe this should be made more robust...

rchl:
I think my description of what's happening might not be accurate. I'm checking exactly.

rchl:
That's accurate actually, `on_activated_async` triggers before `on_reload_async` but during `on_activated_async` the `change_count` is already updated to version that includes latest changes.

jwortmann:
What triggers the `on_activated_async`? ST shouldn't automatically trigger that before `on_reload_async` when a file change happens from external program (for non-active views). At least I cannot reproduce that on Windows.

Also we only trigger the diagnostic requests from `on_activated_async` if the server previously did send a `workspace/diagnostic/refresh`, and furthermore only for views that were not visible.

jwortmann:
I tested both cases now:

1. External modification for a file that is not visible:
  - I only get `on_reload_async`
  - the buffer may have `document_diagnostic_needs_refresh = True` set from a previous `workspace/diagnostic/refresh`, but this doesn't trigger any requests because the view is still not active

2. External modification for a file that is visible
  - I get both `on_activated_async` and then `on_reload_async`
  - since the view was already visible, there cannot be any unsent requests from previous `workspace/diagnostic/refresh`, because for visible views (buffers) we send the requests immediately

So in both cases I get only a single request triggered due to `on_reload_async`, and in both cases this is sent after `didChange`

jwortmann:
But a different thing I noticed is that we only reset the `SessionBuffer.document_diagnostic_needs_refresh` flag from `on_activated_async` at
https://github.com/sublimelsp/LSP/blob/42ff4ee47826a920b2acbeb6cfeebc8669f080bc/plugin/documents.py#L420

I think we should actually move that line into the `do_document_diagnostic_async` method (and also do the same for resetting the inlay hint / code lens / semantic tokens need refresh flags). Because we don't need to send another request when the view gets activated, if the request was already sent before due to `on_reload_async` (I didn't realize this case before).

rchl:
1. Trigger diagnostic
2. Unfocus ST window
3. Modify file externally
4. Focus ST and the view with diagnostic
5. 
```
on_activated_async
:: [21:36:48.677] --> LSP-ruff textDocument/diagnostic (6): {'textDocument': {'uri': '...'}, 'identifier': 'Ruff'}
on_reload_async
:: [21:36:48.702]  -> LSP-ruff textDocument/didChange: <params with 20143 characters>
:: [21:36:48.712] <<< LSP-ruff (6) (duration: 35ms): ...
```

The diagnostic request is triggered from within `on_activated_async` but indirectly through `_do_code_actions_async()`

jwortmann:
So what does trigger the `textDocument/diagnostic (6)` after `on_activated_async` in your case? I don't see that on my side.

rchl:
The diagnostic request is triggered from within `on_activated_async` but indirectly through `_do_code_actions_async()`

jwortmann:
Ah okay I was testing with Pyright which doesn't support code actions.

rchl:
https://github.com/sublimelsp/LSP/pull/2108 is the one where I've added that code actions request.

plugin/session_buffer.py
```diff
@@ -84,6 +85,12 @@
 P = ParamSpec('P')
 
 
+class RequestStrategy(Enum):
+    DEFAULT = 1
+    FORCE = 2
```
jwortmann: This is only relevant for the diagnostic requests, so it should probably be named `FORCE_DIAGNOSTICS`.

jwortmann:
I think we should reconsider the exact conditions when to trigger the code action request. Currently it is triggered on selection change (okay), on diagnostic updated if the view is visible, and always on activated event. I think if the view is not visible we could set a refresh flag for code actions when diagnostics are updated (unless it is a [UnchangedDocumentDiagnosticReport](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#unchangedDocumentDiagnosticReport) from pull diagnostics), and then check that flag in `on_activated_async`. Then we would only do the request if needed (i.e. if diagnostics changed in the meanwhile). And furthermore, in that case the code action request from `on_activated_async` should be delayed if there is also the diagnostic refresh flag set for the buffer. In that case the code action request should wait until the new diagnostics response(s) arrive.

rchl:
Not sure if I've missed something since the implementation seemed simpler than what you've suggested.

rchl:
> And furthermore, in that case the code action request from on_activated_async should be delayed if there is also the diagnostic refresh flag set for the buffer. In that case the code action request should wait until the new diagnostics response(s) arrive.

If I'm to do that maybe I need a bit of help here.
I'm thinking that if I do that and then diagnostic request returns `UnchangedDocumentDiagnosticReport` then we'll potentially miss out on code actions update (previous diagnostic update might still have been relevant for code actions).

jwortmann:
I did a bit of refactoring/simplifaction for the "needs refresh" variables in #2787, which should make the handling a bit easier.

Instead of adding a new `self._code_actions_for_selection_needs_refresh` variable, that refactor should make it possible to just add another `RequestFlags.CODE_ACTION = 32` member, and then set that bit in `SessionBuffer.on_diagnostics_async` and check for it in `on_activated_async`. Although that is (and should be) specific to the SessionBuffer.

When new diagnostics arrive, we actually only need to request code actions from that particular session. I think currently we request new code actions from all sessions. That might also need a bit of refactoring if we want to do it more efficiently.

Yeah, special handling for `UnchangedDocumentDiagnosticReport` might be a bit tricky, so I would not do that for now.

jwortmann:
> Not sure if I've missed something since the implementation seemed simpler than what you've suggested.

Looks right and basically what I had in mind. I haven't tested, but I think that it should cover all possible cases. Either the view is visible, then we shouldn't do any diagnostics triggered by `on_activated_async`. Or the view is in the background, but in that case ST doesn't trigger `on_activated_async`.

jwortmann (approved):

plugin/documents.py
```diff
@@ -425,6 +427,8 @@ def on_activated_async(self) -> None:
                     and (session_view := sb.session.session_view_for_view_async(self.view)) \
                     and session_view.get_request_flags() & RequestFlags.INLAY_HINT:
                 sb.do_inlay_hints_async(self.view)
+        if userprefs().show_code_actions and self._code_actions_for_selection_needs_refresh:
```
jwortmann, line 430: Actually I think we could add `and not sb.document_diagnostic_needs_refresh` here, I guess?

This would be to delay the code action request if there is also a new diagnostic request scheduled anyway. In that case code actions would be triggered automatically when the diagnostic response arrives.

rchl: That's the part that I've asked about in my previous comment. What if we get "unchanged diagnostics" response?

jwortmann: Oh now I understand. Well I think it should be possible to handle that case by moving the 2 lines from here into a new method
```python
def maybe_do_code_actions(self) -> None:
    if userprefs().show_code_actions and self._code_actions_for_selection_needs_refresh \
            and not sb.document_diagnostic_needs_refresh: 
        self._do_code_actions_for_selection_async()
```
and then call this method directly from the SessionBuffer via `listener().maybe_do_code_actions()` when "unchanged diagnostics" response arrives. (And also still call it from `on_activated_async`)

Or am I missing something?

rchl: Can we not do that for now?
I don't think there is even a direct way to access listener from SessionBuffer. I don't want to have to do a bunch of refactors or exposing listener.

jwortmann: Yes we can refactor and improve that later.

PR #2788, opened 2026-03-01
Modernize popup style

### LINKED ISSUE #2635: Popup background color should match the diagnostic result
Currently only the content inside popup is marked as red in case of errors / warning / info

**Current Behaviour**

Image

**Expected Behaviour**

Image

rchl:
Took the style from @predragnikolic (thanks!), tweaked slightly and applied to original stylesheet.

I think the flatter style and darker colors look more modern.

Some examples:

| Before | After |
| ------- | -------- |
| Screenshot 2026-03-01 at 19 39 06  | Screenshot 2026-03-01 at 19 29 28 |
| Screenshot 2026-03-01 at 19 40 24  | Screenshot 2026-03-01 at 19 08 29 |
| Screenshot 2026-03-01 at 19 41 03 | Screenshot 2026-03-01 at 19 08 13 |
| Screenshot 2026-03-01 at 19 43 19 | Screenshot 2026-03-01 at 19 42 37 |

Fixes #2635

### DIFF
--- a/popups.css
+++ b/popups.css
@@ -98,7 +98,6 @@
 .code-actions {
     background-color: color(var(--foreground) alpha(0.1));
 }
-.actions,
 .signature-help-intro {
     font-size: var(--font-size-sm);
 }


jwortmann (changes requested):
I agree that popup style could use an overhaul, but these changes cause various issues on my side:

- Why doesn't the content use a proportional font (system) anymore?
- The background of the popup content has the exact same color as the background in the view now, which is quite bad in my opinion. ST by default uses a slightly different color for popup background for a good reason.
- There shouldn't be any border around the whole popup (which even uses "fake" rounded corners here), because there might already be a border from the theme.
- "information" and "hint" severities are not blue anymore? Now they can't easily be distinguished from regular hover content.
- In general, code blocks can't be distinguished from regular content anymore...
- Admonitions look a bit cramped against the side border, without the previous margin.

| Before | After |
| ------- | ------ |
| hover1a | hover1 |
| hover2a | hover2 |
| hover3a | hover3 |

rchl:
> * Why doesn't the content use a proportional font (system) anymore?

To be fair I'm not against proportional font but the default size always looked weird to me. Why is it so tiny?

Screenshot 2026-03-01 at 21 10 34

Perhaps we should keep system font but just reset font size to 1rem (mdpopups makes it smaller by default).

> * "information" and "hint" severities are not blue anymore? Now they can't easily be distinguished from regular hover content.

Was not intentional.

rchl:
What's the markdown syntax for those admonitions so that I can debug this?

predragnikolic:
> [!TIP]
> Example

```
> [!TIP]
> Example
```


https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

https://facelessuser.github.io/sublime-markdown-popups/about/changelog/#500
> NEW: Use default admonition classes that align with GitHub callouts: note, tip, warning, caution, and important. Legacy classes that do not align are still available: info, success, error.

rchl:
Why do those not render for me? Is it a mdpopups in 3.13 thing?

Screenshot 2026-03-01 at 21 34 02

EDIT: It is a 3.13 thing. It works in 3.8. The new commonmark parser from mdpopups doesn't support those. It should in theory support markdown specific flavor of those but not sure.

EDIT2: Yes, the GH-flavor is supported:

Screenshot 2026-03-01 at 22 14 11

Maybe LSP-julia should try to rewrite those.

popups.css
```diff
@@ -1,22 +1,36 @@
...
+html,
+body {
+    --mdpopups-bg: color(var(--background) lightness(- 3%));
     font-family: system;
 }
-.lsp_popup--spacer {
-    margin-top: 0.5rem;
+.lsp_popup {
+    border-radius: 4px;
+    border: 1px solid color(var(--foreground) alpha(0.1));
```
rchl: Border vs. no border.

Screenshot 2026-03-01 at 21 40 53

Screenshot 2026-03-01 at 21 41 15

I would still say that it looks better with the fake border. Ideally we'd only get border from the theme but we can't control how it looks.

popups.css
```diff
@@ -1,22 +1,37 @@
...
 .lsp_popup h1,
 .lsp_popup h2,
 .lsp_popup h3,
 .lsp_popup h4,
 .lsp_popup h5,
-.lsp_popup h6 {
+.lsp_popup h6,
+.lsp_popup .admonition {
     font-size: 1rem;
+    margin: 0.5rem;
```
rchl: We should actually flip this logic around and instead of adding horizontal margin to various elements, add explicit padding in containers.

Once we agree on the styling, I could try to refactor this. It will require changes in code then.

rchl:
> * In general, code blocks can't be distinguished from regular content anymore...

I think we could make it so that the top-level code blocks don't have background while we don't touch the ones that are nested in hover info etc.

popups.css
rchl: But this extra border certainly looks bad in the theme you are using.

For my own sake, do you know how to edit theme to add a border like the one you have?

jwortmann: For square border:
```jsonc
{
    "rules": [
        {
            "class": "popup_control html_popup",
            "content_margin": 1,
            "layer1.opacity": 1.0,
            "layer1.tint": "color(var(--background) lightness(+ 20%))"
        },
    ]
}
```

For border with rounded corners you need to have some PNG textures prepared by the theme, but as a workaround you can re-use some shipped textures from the default theme (that gives a 6px corner radius):
```jsonc
{
    "rules": [
        {
            "class": "popup_control html_popup",
            "content_margin": 3,
            "layer0.texture": "Theme - Default/common/radio_button_back.png",
            "layer0.opacity": 1.0,
            "layer0.tint": "color(var(--background) blend(white 95%))",
            "layer0.inner_margin": 6,
            "layer1.texture": "Theme - Default/common/radio_button_border.png",
            "layer1.opacity": 1.0,
            "layer1.tint": "color(var(--background) lightness(+ 20%))",
            "layer1.inner_margin": 6
        },
        {
            "class": "popup_control html_popup",
            "parents": [{"class": "title_bar", "attributes": ["file_light"]}],
            "layer0.tint": "color(var(--background) blend(black 95%))",
            "layer1.tint": "color(var(--background) lightness(- 20%))"
        },
    ]
}
```

Here I used some `layer1.tint` that is applicable to dark themes; for light themes you probably want to decrease the lightness instead.

---

Note that ~~`color(var(--background) blend(var(--foreground) 94%))`~~ `layer0.tint` is the default background color for popups that is applied by ST (i.e. default `background-color` CSS property for `html` tag I assume), and we should not modify that color for the LSP popups.

jwortmann: > I would still say that it looks better with the fake border. Ideally we'd only get border from the theme but we can't control how it looks.

Whether there should be a border or not is up to the theme, and I'm strongly against interfering with that from LSP popups.

While in general the popups look better with than without border imo, in your case the reason why it looks really bad without border is because you modified the background color so that it is exactly the same or almost the same as the regular background from the view.

And we should also not change that just for consistency with all other popups (like the builtin definitions popup for files that don't have any language server running, or popups from other plugins).

rchl: Note that mdpopups by default [overrides](https://github.com/facelessuser/sublime-markdown-popups/blob/28882011f653096c7bc3bf990c317b9d657b03d1/mdpopups/css/default.css#L48-L71) background color of the content to `color(var(--background) blend(black 95%))` for light theme and `color(var(--background) blend(white 95%))` for dark theme.

jwortmann: > Note that mdpopups by default [overrides](https://github.com/facelessuser/sublime-markdown-popups/blob/28882011f653096c7bc3bf990c317b9d657b03d1/mdpopups/css/default.css#L48-L71) background color

Interesting, I wasn't aware of that and it seems like those are the actual colors used by ST defaults, in contrary to what I falsely wrote above. I have now fixed those colors in my comment above. Not sure how I initially gotten to the `color(var(--background) blend(var(--foreground) 94%))`, probably I experimented with different combinations of background and foreground colors until the results matched for some common color schemes. It shouldn't make a visible difference though, unless the foreground color is very unusual (like red or so).

Furthermore, I forgot that there is another rule required for "real" rounded corners, because you need to override the default popup shadow texture with a different texture that has rounded corners too. Unfortunately there is no such texture shipped with the default themes. You can see the corresponding theme rule in my theme at https://github.com/jwortmann/brackets-theme/blob/599cb1f7be6e50f397e7cb4dcd34ee01552b7778/Brackets.sublime-theme#L2127-L2135 and could copy the texture file from that repository (although in that case the corner radius wouldn't match exactly, and I haven't confirmed whether it would work properly if combined with the theme rules from above).

predragnikolic:
What do you think about this style for alerts?

different
Screenshot 2026-03-02 at 20 18 17

current
Screenshot 2026-03-02 at 20 18 59


for the different look the css is:
```
.lsp_popup blockquote.alert {
    border-radius: 0;
    border-width: 0;
    border-left-width: 4px;
}
.lsp_popup blockquote .alert-title {
    margin-left: 0rem;
    font-size: 1rem;
}
.lsp_popup blockquote.alert-tip .alert-title {
    color: var(--mdpopups-admon-tip-accent);
    background-color: transparent;
}
.lsp_popup blockquote.alert-warning .alert-title {
    color: var(--mdpopups-admon-warning-accent);
    background-color: transparent;
}
.lsp_popup blockquote.alert-caution .alert-title {
    color: var(--mdpopups-admon-caution-accent);
    background-color: transparent;
}
.lsp_popup blockquote.alert-note .alert-title {
    color: var(--mdpopups-admon-note-accent);
    background-color: transparent;
}
.lsp_popup blockquote.alert-important .alert-title {
    color: var(--mdpopups-admon-important-accent);
    background-color: transparent;
}
```

popups.css
rchl: Can you check how the latest code (with bg and border override removed) looks like with error diagnostics?

I wonder if you will see a gap between the border and the diagnostic bg.

rchl:
Looks nice.
Not sure how likely are we gonna see those GFM-style admonitions in the wild :)

popups.css
jwortmann: Yes there is a small gap if the theme adds additional margin due to rounded corners, but there is nothing we can do about that. I think having no additional margin around the diagnostics now looks better than before.

But I'm not really sure about this flat style without the darkened background color and the border for code blocks, that is used by mdpopups defaults. I find it more difficult to read now, if there are several code blocks interspersed with paragraphs of regular text.

hover

For example in this screenshot it is impossible to tell, that the code under "Examples" is actually two separate code blocks (compare it with the other screenshot I posted earlier).
Here the `isempty(fp::AbstractPath)` line doesn't belong to the examples anymore (the function hover in Julia just combines the different docstrings from all function instances, so after the Examples block begins the docstring of a different instance (method) of the `isempty` function).

rchl: Reverted all the overrides. Now only adding padding in some root containers.

Most likely still missed something. If you find something broken please post the MarkupContent contents or just the hover response.

It's now more back to how it was before but still with some flatter, more modern elements.

---

Screenshot 2026-03-03 at 22 36 22

---

Screenshot 2026-03-03 at 22 36 31

---

Screenshot 2026-03-03 at 22 36 39

---

Screenshot 2026-03-03 at 22 36 44

---

Screenshot 2026-03-03 at 22 50 36

plugin/completion.py
```diff
@@ -314,9 +315,10 @@ def _handle_resolve_response_async(self, language_map: MarkdownLangMap | None, i
             documentation = self._format_documentation(markdown, None)
         minihtml_content = ""
         if detail:
-            minihtml_content += f"<div class='highlight'>{detail}</div>"
+            minihtml_content += wrap_html(detail, cls='c')
         if documentation:
-            minihtml_content += documentation
+            cls = 'c' if not detail else 'c border-top'
+            minihtml_content += wrap_html(documentation, cls=cls)
```
rchl: Image

jwortmann:
The signature help popup looks bad now without any margin/padding:

sighelp

The font size for documentation text in the signature help popup doesn't match the font size in the regular popup

---

How did you manage to increase the font size for text in the regular popup? I can't see any corresponding rule in the diff.
I actually like the smaller font size from mdpopups defaults and would like to reset that in my user `mdpopups.css`.

popups.css
```diff
@@ -82,9 +112,9 @@
     padding: 0.5rem;
 }
 pre.related_info {
-    border-top: 1px solid color(var(--background) alpha(0.25));
-    margin-top: 0.7rem;
-    padding-top: 0.7rem;
+    border-top: 1px solid color(var(--background) alpha(0));
```
jwortmann: I think this could be simplified to `border-top: 1px solid transparent;`

Not sure if transparent border was intended here?

rchl:
> The signature help popup looks bad now without any margin/padding:
> sighelp
>
> The font size for documentation text in the signature help popup doesn't match the font size in the regular popup

Good to know. I've overlooked that. Needs a `c` container like other cases.

> How did you manage to increase the font size for text in the regular popup? I can't see any corresponding rule in the diff. I actually like the smaller font size from mdpopups defaults and would like to reset that in my user `mdpopups.css`.

https://github.com/sublimelsp/LSP/blob/c24ad29750bf962cfdb50592dc0b82873d8fa637/popups.css#L10-L16

and

https://github.com/sublimelsp/LSP/blob/c24ad29750bf962cfdb50592dc0b82873d8fa637/popups.css#L59-L60

We can continue discussing this and see if it ends up in final changes.

jwortmann:
Could we use a more descriptive class name for `.c`? Like `.container` or `.content` or whatever it stands for.

popups.css
rchl: Updated in 938a31e3d9b762e499612ac695ff174c0052a782

This is used when diagnostic has `relatedInformation`.

I've updated margin of `hr` which can affect something else negatively but we'll see...

Screenshot 2026-03-04 at 19 35 12

jwortmann: Okay that part looks good to me.

One different thing I thought about before, is whether we could couple code actions more to their diagnostics.
Because we only show the quickfix code actions in the hover anyway ("refactor" code actions are in the context menu and main menu), so in theory every code action should have a corresponding diagnostic. And [CodeAction](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeAction) has a (optional) `diagnostics` property:
```ts
    /**
	 * The diagnostics that this code action resolves.
	 */
	diagnostics?: [Diagnostic](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic)[];
```

So maybe it could be useful to display the code action directly as part of the diagnostic(s) it belongs to, similar to the `relatedInformation`. If `diagnostics` isn't provided, we could still show it separately, as it is now.

This would require some changes in the source code. The implementation would also need to consider the "show_code_actions_in_hover" setting. And probably we don't need to show the server name in the code action in that case, because it is already shown after the diagnostic message.

predragnikolic:
> [!TIP] 
> it is possible to specify `"mdpopups.debug": 3` in Preferences.sublime-settings
and mdpopups will log in the ST the whole popup HTML + CSS 

Noticed one small bug.
When you open a package.json file and hover a field that has no docs,
a empty popup is displayed, but the popup should not be visible:

Screenshot 2026-03-04 at 22 23 57

```
mdpopups: =====HTML OUTPUT=====
mdpopups: 
<body><div class="c"><p>Sets the colors used within the text area.</p>
<div class="c--spacer"></div></div></body>
```

plugin/core/css.py
```diff
@@ -15,6 +15,10 @@ def __init__(self) -> None:
         self.annotations = sublime.load_resource("Packages/LSP/annotations.css")
         self.annotations_classname = "lsp_annotation"
 
+    @property
+    def popups(self) -> str:
+        return sublime.load_resource("Packages/LSP/popups.css")
```
rchl: for faster iteration - will not be merged

plugin/core/signature_help.py
rchl, line 1: | Before | After |
|:-------|:------|
| Image | Image |

This is obviously almost the same but worth noting that I'm also applying smaller text to the lower (signature documentation) code block. Previously the code blocks would never follow font-size of the container (the content under the line is a single container).

We should probably do further updates to the styles of this one. For now I have focused on refactoring to adapt to the way those containers are laid out so that we don't get extra paddings anywhere.

plugin/hover.py
```diff
@@ -267,7 +268,9 @@ def hover_content(self) -> str:
         for hover, language_map in self._hover_responses:
             content = (hover.get('contents') or '') if isinstance(hover, dict) else ''
             allowed_formats = FORMAT_MARKED_STRING | FORMAT_MARKUP_CONTENT
-            contents.append(minihtml(self.view, content, allowed_formats, language_map))
+            parsed = minihtml(self.view, content, allowed_formats, language_map)
+            if parsed:
+                contents.append(html_wrapper(parsed))
         return '<hr>'.join(contents)
```
rchl: Have to test this with multiple servers or just fake it because it looks like it will have extra padding.

Will probably have to switch to using `border-top` or `hr` without margins.

predragnikolic:
```php
<?php

$hello = 21;
```

Screenshot 2026-03-05 at 19 21 48

if you focus on the inline code, you will notice (because the font is smaller in the inline code) that there is more padding on top than from the bottom.

We can fix this by either reverting the font-size from 1.05->rem 0.9.
here is how it looks with 0.9
Screenshot 2026-03-05 at 19 26 32

rchl:
Can you paste the payload for that hover?

I have my testing playground utilizing lsp-sample server from ms.

predragnikolic:
```
:: [19:53:49.682] <<< LSP-intelephense (57) (duration: 1ms): {'contents': {'kind': 'markdown', 'value': '_@var_ `int $hello`'}, 'range': {'start': {'line': 2, 'character': 0}, 'end': {'line': 2, 'character': 6}}}
```

On linux it is more noticeable:
image

rchl:
> We can fix this by either reverting the font-size from 1.05->rem 0.9.
> here is how it looks with 0.9

This is pretty generic contents. Are you suggesting to change the font-size of hover contents back to 0.9?

predragnikolic:
> Are you suggesting to change the font-size of hover contents back to 0.9?

It is possible to also increase the font-size of the inline code to 1.05rem? (That might also do it)

rchl:
I'm not really sure what you seeing.
You said:

> Screenshot 2026-03-05 at 19 21 48
> 
> if you focus on the inline code, you will notice (because the font is smaller in the inline code) that there is more padding on top than from the bottom.

You are talking about the padding in the whole popup container, not about padding within the inline highlight itself, right?

In that case I don't think that's the case. It might look like so because the bigger text (non-code) uses only lowercase letters.

Here is similar content with and without inline code:

Screenshot 2026-03-05 at 21 05 53

Screenshot 2026-03-05 at 21 06 04

But seems to have the same top padding.

EDIT: I've looked more closely at your linux screenshot and the issue with the inline code highlight:

> ```
> :: [19:53:49.682] <<< LSP-intelephense (57) (duration: 1ms): {'contents': {'kind': 'markdown', 'value': '_@var_ `int $hello`'}, 'range': {'start': {'line': 2, 'character': 0}, 'end': {'line': 2, 'character': 6}}}
> ```
> 
> On linux it is more noticeable: image

So I guess you meant that?

predragnikolic:
The padding from the top seems a bit bigger (I marked it with red in the screenshot)

Screenshot_20260305-215740

rchl:
You can try making `line-height` bigger. I think it's 1.1rem by default.

```
.lsp_popup {
    line-height: 1.2rem;
}
```

Not saying it would be good solution but just to identity the problem.

That said, since font rendering varies between platforms, it's not necessarily a good idea to implement workarounds for stuff like this.

Though we can't rule non-ideal styles or minihtml quirks.

predragnikolic:
Should we automatically set the user set font in the popup?

```
        font = globalprefs().get('font_face', 'system')
        rule = f"""
        html, body {{
              --mdpopups-font-mono: {font};
               font-family: {font};
        }}"""
        return sublime.load_resource("Packages/LSP/popups.css") + rule
```

rchl:
> Should we automatically set the user set font in the popup?
> 
> ```
>         font = globalprefs().get('font_face', 'system')
>         rule = f"""
>         html, body {{
>               --mdpopups-font-mono: {font};
>                font-family: {font};
>         }}"""
>         return sublime.load_resource("Packages/LSP/popups.css") + rule
> ```

Not sure if popup style does necessarily need to be consistent with the style of the code but not saying I'm against it. Would need to be tried I guess.

But implementation wise not setting the `font-family` should be equivalent. Currently we are setting `font-family: system` in LSP but mdpopups doesn't itself set `font-family` apart for some cases:

```
html {
  ...
  --mdpopups-font-mono: "sf mono", Consolas, "Liberation Mono", Menlo, Courier, monospace;
  ...
}
.mdpopups .highlight,
.mdpopups code,
.mdpopups var,
.mdpopups tt {
  font-family: var(--mdpopups-font-mono);
}
```

So all we would need to do is to remove our `font-family` override and then potentially also override `--mdpopups-font-mono` as you've suggested. Note that the font name needs to be quoted to work with fonts that have spaces in the name.

I've tried this without any further tweaks with the font I'm using:

Screenshot 2026-03-08 at 21 40 33

Compare without it:

Screenshot 2026-03-08 at 21 43 33

(the highlighted block in this particular case probably looks better with default mdpopups mono font)

rchl:
Switched to user font in latest commit.

Not really liking it in the current form. Maybe if we play with font-size (make it smaller) it would be better...

## Before

Screenshot 2026-03-11 at 09 21 24

## After

Screenshot 2026-03-11 at 09 21 33

jwortmann:
> Switched to user font in latest commit.

Didn't we already discuss before to keep the proportional font? :thinking:

The built-in "show_definitions" popup from ST also uses a proportional font, as well as other editors like VSCode do in their popups.

If you really want to do such drastic changes, then it should be opt-in with a user setting. But actually users can already do that if they want, via `mdpopups.css` (but who would even want to do that?). Or you could try to convince facelessuser to use a monospace font in the mdpopups defaults, although in that case I would vote to opt-out of that for LSP popups...

---

By the way, I would consider the fact that the source code lines in the hover content from screenshot above are not actually Markdown code blocks (and therefore don't use the code font) as a bug in Pyright. They seem to just copy the raw docstring content without converting it to Markdown.

predragnikolic:
I am fine with not having the user font by default in the popup:

> If you really want to do such drastic changes, ... users can already do that if they want, via mdpopups.css

I do think that having a mono font by default in the popup makes the popup text more readable (for me). But as Janos said, I can already do that in mdpopups so for me it is fine to remove the logic for other users.

predragnikolic:
I did test this PR (before the font-size change) with all sorts of ls (elm, intelephense, jdtls, typescript, JSON. css, docker, astro...), and I did not notice any issues.

jwortmann:
P.S. regarding the font-face.
I would even say that the diagnostics should also use a proportional font (so remove the `font-family: var(--mdpopups-font-mono);` for the `.diagnostics` class). The 3.18 version of the specs allow diagnostic messages as Markdown:
```ts
	/**
	 * The diagnostic's message.
	 *
	 * @since 3.18.0 - support for MarkupContent. This is guarded by the client
	 * capability `textDocument.diagnostic.markupMessageSupport`.
	 */
	message: string | [MarkupContent](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#markupContentInnerDefinition);
```
and it seems all we have to do to support this would be to set that capability in the initialize request. Then parts of the message that are marked as inline code via ` `` ` in Markdown would automatically be rendered with the monospace font by mdpopups. Although I don't know whether there are servers which support that already.

In general I think we should use monospace font for code (inline code and code blocks) and the proportional font for regular text content.

Btw I see that we're missing a few options for pull diagnostics that we actually support. I can make a PR to fix that.

jwortmann:
Update to my last comment:

I made some experiments with adding `markupMessageSupport` on a local branch. A server which supports it already is [Pyrefly](https://lsp.sublimetext.io/language_servers/#pyrefly). The code changes would result in some type errors at the moment, because our LSP types are not updated to the 3.18 specs yet (also the upstream repo isn't updated yet, I think).

We can use https://github.com/sublimelsp/LSP/blob/06a1b8d279bdace894add56c6f9863ee53f99f14/plugin/core/views.py#L490-L497
which is a wrapper for `mdpopups.md2html`, to convert the diagnostic message from Markdown to minihtml.

But I saw that unfortunately mdpopups hardcodes the background color for code blocks and inline code as a `<span style="background-color: ...">` tag into the HTML. As far as I can tell, there is no way to pass CSS into `mdpopups.md2html`. I think ideally we want to reduce the opacity of the background color for the code spans, so that it would still use the color from the diagnostic severity, just with a different shading. Currently it doesn't look so well for the code spans.

For the diagnostic annotations and Goto Diagnostic panel / output panel I think we could just take the raw message, including the backticks. I don't think that servers will put really complicated Markdown e.g. with fenced code blocks into the message.

hover

---

goto_diagnostic

popups.css
```diff
@@ -1,71 +1,112 @@
...
+    background-color: transparent;
+}
+.mdpopups blockquote.alert-important .alert-title {
+    color: var(--mdpopups-admon-important-accent);
+    background-color: transparent;
+}
+.lsp_popup .wrapper {
+    font-size: var(--font-size);
     padding: 0.5rem;
-    white-space: pre-wrap;
```
rchl: Have to look into why there was `white-space: pre-wrap` but only for the error diagnostics. If this is better then it should be set for all diagnostic types. For now it's removed.

rchl: I've added it in the release commit 79a2128840af25a27dec14aeb4be23790efd5459 which feels like a mistake.

rchl:
> But I saw that unfortunately mdpopups hardcodes the background color for code blocks and inline code as a `<span style="background-color: ...">` tag into the HTML.

I've seen quite a few raw html dumps when working on this task and I don't think I've ever seen hardcoded `background-color` from mdpopups.

jwortmann:
> I've seen quite a few raw html dumps when working on this task and I don't think I've ever seen hardcoded `background-color` from mdpopups.

You're right, I was mistaken. It actually only hardcodes the text color:
```
<p>Parameter <code class="highlight"><span style="color: #f6e5cb;">edit</span></code> is unused</p>
```

I just saw the hex-color value, but apparently didn't pay close attention.

We don't have any rule for the background color in `popups.css` for the `code` tag or `.highlight` class. So I assume it must come from mdpopups default CSS, and we should be able to override that.

I will try to tweak the style (make the background half-transparent) and then see how it looks.

popups.css
rchl: You are saying that code action should always reference existing diagnostics but this is in practice not really enforced. Also it would be our job to match diagnostics attached to code action against actual diagnostics which could be a bit tricky? And what if we don't find matching diagnostic? We'll show code action as we do now?

This change raises a bunch of questions and potential issues so I certainly wouldn't make it part of this PR.

rchl:
> We don't have any rule for the background color in `popups.css` for the `code` tag or `.highlight` class. So I assume it must come from mdpopups default CSS, and we should be able to override that.

Comes from something like this:

```css
html.light {
  --mdpopups-bg: color(var(--background) blend(black 95%));
  /* ... */
  --mdpopups-hl-bg: color(var(--background) blend(black 98%));
  background-color: var(--mdpopups-bg);
}
html.dark {
  --mdpopups-bg: color(var(--background) blend(white 95%));
  /* ... */
  --mdpopups-hl-bg: color(var(--background) blend(white 98%));
  background-color: var(--mdpopups-bg);
}
/* ... */
.mdpopups .highlight,
.mdpopups code {
  border: 1px solid var(--mdpopups-hl-border);
  border-radius: 0.25rem;
  font-size: 0.9rem;
  background-color: var(--mdpopups-hl-bg);
}
```

jwortmann:
Yes, I saw it too.

Update:

```css
.diagnostics code.highlight {
    border-color: color(var(--background) alpha(0.25));
    background-color: color(var(--background) alpha(0.25));
}
```



---

```css
.diagnostics code.highlight {
    border-color: transparent;
    background-color: transparent;
}
```



---

And the same message from Pyright (no MarkupContent support for diagnostics):



I don't find it less readable than the diagnostics with monospace font that we currently use.

*Edit:* Here is a diagnostic from Pyright with more code in it:



And with monospace font:



In this case I'm not really sure what is better. The monospace font is probably more readable for messages with such large amounts of code, but on the other hand the proportional font is more compact, which could result in fewer linebreaks.

I guess we could also use monospace font if the message is a plain string (Pyright), and the proportional font (excluding code markup) if the message is MarkupContent. But that would be a bit inconsistent...

rchl:
I would try to override the `--mdpopups-hl-bg` / `--mdpopups-hl-border` instead of `background-color` / `border` specifically. Potentially even try to find styles for those that work in any context.

plugin/hover.py
```diff
@@ -269,8 +270,10 @@ def hover_content(self) -> str:
         for hover, language_map in self._hover_responses:
             content = (hover.get('contents') or '') if isinstance(hover, dict) else ''
             allowed_formats = FORMAT_MARKED_STRING | FORMAT_MARKUP_CONTENT
-            contents.append(minihtml(self.view, content, allowed_formats, language_map))
-        return '<hr>'.join(contents)
+            parsed = minihtml(self.view, content, allowed_formats, language_map)
+            if parsed:
```
jwortmann:
```suggestion
            if parsed := minihtml(self.view, content, allowed_formats, language_map):
```

popups.css
```diff
@@ -1,71 +1,112 @@
...
+html,
+body {
     font-family: system;
 }
-.lsp_popup--spacer {
-    margin-top: 0.5rem;
+.lsp_popup {
+    --font-size: 1.05rem;
+    --font-size-sm: 1rem;
+    --font-size-xs: .9rem;
```
jwortmann: If you want to override the mdpopups defaults, could you maybe set the regular font size to 1rem? That would be consistent with ST built-in popups (I think), and 1.05rem is such a weird number.

And was it intended for inline code to still use the smaller font? In my opinion that looks a bit too small compared to the text font now. Maybe it looks okay already if `--font-size` is set to 1rem.

I would say diagnostics messages should also use the regular font size then, because currently they still use the smaller font size.

Image

plugin/core/signature_help.py
```diff
@@ -100,18 +106,11 @@ def select_signature(self, forward: bool) -> None:
...
-        fmt = '<p><div style="font-size: 0.9rem"><b>{}</b> of <b>{}</b> overloads ' + \
-              '(use <kbd>↑</kbd> <kbd>↓</kbd> to navigate, press <kbd>Esc</kbd> to hide):</div></p>'
-        return fmt.format(
-            self._active_signature_index + 1,
-            len(self._signatures),
-        )
-
-    def _render_label(self, signature: SignatureInformation) -> list[str]:
+        return f'<p><b>{self._active_signature_index + 1}</b> of <b>{len(self._signatures)}</b> overloads ' + \
+               '(use <kbd>↑</kbd> <kbd>↓</kbd> to navigate, press <kbd>Esc</kbd> to hide)</p>'
```
jwortmann: In my opinion this header line should use the `--font-size-xs`, because it is not part of the content, and is less important.

popups.css
```diff
@@ -1,71 +1,112 @@
...
+.lsp_popup .font-size-xs pre.highlight {
+    font-size: var(--font-size-xs)
+}
+.color-muted {
+    color: color(var(--foreground) alpha(0.6));
+}
+.errors,
+.warnings,
+.info,
+.hints {
```
jwortmann: While we're at it, could we make this consistent with the [DiagnosticSeverity](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticSeverity), and not use mixed singular/plural?

`.error`
`.warning`
`.information`
`.hint`

rchl: Addressed in https://github.com/sublimelsp/LSP/pull/2809

plugin/core/signature_help.py
rchl: e405f6bbef8353958d9f10980577cffa715e64b9

popups.css
rchl: Following is ST's default definition popup. It uses 1.05rem font-size for `p`:

```html
<body id=show-definitions>
    <style>
        body {
            font-family: system;
        }

        body {
            white-space: nowrap;
        }

        h1 {
            font-size: 1.1rem;
            font-weight: bold;
            margin: 0 0 0.5em 0;
        }

        h1 code {
            font-size: 0.9em;
            font-weight: normal;
            color: color(var(--foreground) a(0.7));
        }

        p {
            font-size: 1.05rem;
            margin: 0;
        }

        .icon {
            text-decoration: none;
            font-size: 1em;
        }

        a {
            line-height: 1.3;
        }

        span.syntax {
            color: color(var(--foreground) a(0.5));
            padding-left: 10px;
            font-style: italic;
            font-size: 0.9em;
        }

        span.kind {
            font-weight: bold;
            font-style: italic;
            width: 1.5rem;
            display: inline-block;
            text-align: center;
            font-family: system;
            line-height: 1.3;
            border-radius: 2px;
            position: relative;
            top: 1px;
            margin-right: 6px;
        }

        span.kind_ambiguous {
            display: none;
        }

        span.kind_keyword {
            background-color: color(var(--pinkish) a(0.2));
            color: var(--pinkish);
        }

        span.kind_type {
            background-color: color(var(--purplish) a(0.2));
            color: var(--purplish);
        }

        span.kind_function {
            background-color: color(var(--redish) a(0.2));
            color: var(--redish);
        }

        span.kind_namespace {
            background-color: color(var(--bluish) a(0.2));
            color: var(--bluish);
        }

        span.kind_navigation {
            background-color: color(var(--yellowish) a(0.2));
            color: var(--yellowish);
        }

        span.kind_markup {
            background-color: color(var(--orangish) a(0.2));
            color: var(--orangish);
        }

        span.kind_variable {
            background-color: color(var(--cyanish) a(0.2));
            color: var(--cyanish);
        }

        span.kind_snippet {
            background-color: color(var(--greenish) a(0.2));
            color: var(--greenish);
        }
    </style>


    <h1>References&nbsp;to&nbsp;<code>_render_intro</code></h1>
    <p><a href='subl:open_symbol_definition  {&quot;path&quot;:&quot;/usr/local/workspace/sublime-packages/LSP/plugin/core/signature_help.py:82:32&quot;,&quot;hide_popup&quot;:true,&quot;add_jump_point&quot;:3966,&quot;focus_view&quot;:73}' title="Cmd+Click to Open Tab to Right of View">LSP/plugin/core/signature_help.py:82</a>&nbsp;<a class="icon" href='subl:open_symbol_definition  {&quot;path&quot;:&quot;/usr/local/workspace/sublime-packages/LSP/plugin/core/signature_help.py:82:32&quot;,&quot;hide_popup&quot;:true,&quot;new_tab&quot;:true,&quot;add_jump_point&quot;:3966,&quot;focus_view&quot;:73}' title="Open Tab to Right">◨</a>&nbsp;<span class="syntax">Python</span></p>

</body>
```

rchl: d94382e7a8811fd466e1030caec6a552ca789b26 - I think now code blocks and diagnostics should use same font-size as normal text.

rchl: Also, using 1.05rem actually fixes [my complain](https://github.com/sublimelsp/LSP/pull/2788#issuecomment-3980944430) about font-size being unusually small by default so given that ST also does this, I think it's a good default.

## 1.0rem

Screenshot 2026-03-16 at 23 26 04

## 1.05rem

Screenshot 2026-03-16 at 23 25 53

jwortmann: Hm, okay I didn't realize that the show-definitions popup also uses 1.05rem. However, the "show scope name" popup doesn't adjust the font size for `p` though, so I assume it's 1rem there:
```html
            <body id=show-scope>
                <style>
                    h1 {
                        font-size: 1.1rem;
                        font-weight: 500;
                        margin: 0 0 0.5em 0;
                        font-family: system;
                    }
                    p {
                        margin-top: 0;
                    }
                    a {
                        font-weight: normal;
                        font-style: italic;
                        padding-left: 1em;
                        font-size: 1.0rem;
                    }
                    span.nums {
                        display: inline-block;
                        text-align: right;
                        width: %dem;
                        color: color(var(--foreground) a(0.8))
                    }
                    span.context {
                        padding-left: 0.5em;
                    }
                </style>
                <h1>Scope Name <a href="c:%s">Copy</a></h1>
                <p>%s</p>
                <h1>Context Backtrace</h1>
                %s
            </body>
```

---

I don't know why it looks so tiny in your screenshots. The font in those screenshots looks even smaller compared to what I see here with the default from mdpopups (0.9rem) that we currently use.

This is how it looks with the latest commit from this PR on my side:

hover

Mabye it's just a font thing? I believe the default `system` font on Windows is Segoe UI, perhaps the system fonts on Mac and Linux are just significantly smaller? https://tonsky.me/blog/font-size/

jwortmann:
Just saw that with the latest changes, the font color in the diagnostics is now the color from the DiagnosticSeverity:

hover

I assume that was an unintended change (?)

rchl:
This is weird. ST has internal `error` / `warning` classes that modify foreground color. I don't see that documented in https://www.sublimetext.com/docs/minihtml.html

Check with:

```
view.show_popup('<span class="error">Test</span>')
```

EDIT:

I've looked inside the binary and found those 3 stylesheet that I guess are internal styles used in different contexts:

## One

```css
html {
    background-color: color(var(--background) blend(#fff 95%));
    color: var(--foreground);
}

html.light {
    background-color: color(var(--background) blend(#000 95%));
}

a {
    color: var(--bluish);
}

.error,
.deleted {
    color: var(--redish);
}

.warning,
.modified {
    color: var(--orangish);
}

.success,
.inserted {
    color: var(--greenish);
}
```

---

### Two

```css
html {
    background-color: #00000000;
    color: var(--foreground);
}

a {
    color: var(--bluish);
}

.error,
.deleted {
    background-color: color(var(--redish) blend(var(--background) 30%));
    color: var(--foreground);
}

.warning,
.modified {
    background-color: color(var(--orangish) blend(var(--background) 30%));
    color: var(--foreground);
}

.success,
.inserted {
    background-color: color(var(--greenish) blend(var(--background) 30%));
    color: var(--foreground);
}
```

---

## Three

```css
html {
    background-color: color(var(--background));
    color: var(--foreground);
}

html.light {
    background-color: color(var(--background));
}

a {
    color: var(--bluish);
}

.error,
.deleted {
    color: var(--redish);
}

.warning,
.modified {
    color: var(--orangish);
}

.success,
.inserted {
    color: var(--greenish);
}
```

jwortmann:
Oh right, I forgot about those. They are used for showing build errors as phantoms.

Well, I think then it might not be the best idea to also use the same class names in LSP too. Maybe that was even the reason why our class names were named differently. Although I think it should be safe to just override these styles in our own CSS, like you just did.

popups.css
jwortmann: We can fix the OS dependent font size like this:

```diff
diff --git a/plugin/core/views.py b/plugin/core/views.py
index f159d322..7c6ee1b0 100644
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -42,6 +42,7 @@ from ...protocol import WillSaveTextDocumentParams
 from .constants import CODE_ACTION_KINDS
 from .constants import MARKO_MD_PARSER_VERSION
 from .constants import ST_CACHE_PATH
+from .constants import ST_PLATFORM
 from .constants import ST_STORAGE_PATH
 from .constants import SUBLIME_KIND_SCOPES
 from .constants import SublimeKind
@@ -488,6 +489,7 @@ def show_lsp_popup(
         flags=flags,
         location=location,
         wrapper_class=wrapper_class,
+        template_vars={'is_windows': ST_PLATFORM == 'windows'},
         max_width=int(view.em_width() * float(userprefs().popup_max_characters_width)),
         max_height=int(view.line_height() * float(userprefs().popup_max_characters_height)),
         on_navigate=on_navigate,
@@ -506,7 +508,13 @@ def update_lsp_popup(
     css = css if css is not None else lsp_css().popups
     wrapper_class = wrapper_class if wrapper_class is not None else lsp_css().popups_classname
     body_wrapper = f'<body id="{body_id}">{{}}</body>' if body_id else '<body>{}</body>'
-    mdpopups.update_popup(view, body_wrapper.format(contents), css=css, md=md, wrapper_class=wrapper_class)
+    mdpopups.update_popup(
+        view,
+        body_wrapper.format(contents),
+        css=css,
+        md=md,
+        wrapper_class=wrapper_class,
+        template_vars={'is_windows': ST_PLATFORM == 'windows'})
 
 
 FORMAT_STRING = 0x1
diff --git a/popups.css b/popups.css
index 4fdae062..9224a971 100644
--- a/popups.css
+++ b/popups.css
@@ -3,9 +3,15 @@ body {
     font-family: system;
 }
 .lsp_popup {
+{% if plugin.is_windows %}
+    --font-size: 1rem;
+    --font-size-sm: 0.9rem;
+    --font-size-xs: 0.9rem;
+{% else %}
     --font-size: 1.05rem;
     --font-size-sm: 1rem;
-    --font-size-xs: .9rem;
+    --font-size-xs: 0.9rem;
+{% endif %}
 }
 .mdpopups hr {
     margin-top: 0.5rem;
diff --git a/stubs/mdpopups.pyi b/stubs/mdpopups.pyi
index defa52f0..b5bbfe8b 100644
--- a/stubs/mdpopups.pyi
+++ b/stubs/mdpopups.pyi
@@ -31,7 +31,7 @@ def update_popup(
     md: bool = True,
     css=None,  # type: Optional[str]
     wrapper_class=None,  # type: Optional[str]
-    template_vars=None,  # type: Optional[str]
+    template_vars=None,  # type: Optional[dict]
     nl2br: bool = True,
     allow_code_wrap: bool = False
 ) -> None: ...
```

jwortmann: I think the signature help popup needs a bit of further tweaking. The signature code block doesn't seem to use the same font size as code blocks in the hover popup. This is with the changes from the diff I pasted above:

Signature help:

sighelp

Hover:

hover

popups.css
predragnikolic: The original question was if the popup should have a fake border and radius.

The answer was to not have a fake border. 

It is the responsibility of a ST theme to provide do that, like Janos showed here -> https://github.com/sublimelsp/LSP/pull/2788#discussion_r2870003882

I will resolve this thread now.

popups.css
rchl: You can just push your changes to this branch.

rchl: Removed font-size-sm from signature help (the top part). The bottom part (documenation) still has `font-size-xs` set. Not sure why anymore. We can remove it or change to `font-size-sm`.

Screenshot 2026-03-17 at 21 28 43


EDIT: The bottom part follows `main` branch where it's also xs (though code block doesn't follow that size there):

Screenshot 2026-03-17 at 21 31 30

jwortmann: Okay, I pushed these changes for Windows in https://github.com/sublimelsp/LSP/pull/2788/changes/c5cfd20624bfe8aad7341e74b2bcd4263effda53.

Can you confirm that you still see the desired font size on Mac?

---

Yeah I'm not sure if there are really 3 different font sized needed. I think I would use the smaller font size for the signature help header (the "1 of 3 overloads" line), and probably also for the symbol action links at the bottom of the popup (if enabled), because those are not part of the content. And for everything else I would use the regular font size. So in that case there would only be a `--font-size` and a `--font-size-sm` variable.

I think due to using rem units, which is a relative unit, we still see the differences from mdpopups defaults. For example the code blocks have a little bit larger font than the text, even though both use the `--font-size` variable.

rchl: > I think due to using rem units, which is a relative unit, we still see the differences from mdpopups defaults. For example the code blocks have a little bit larger font than the text, even though both use the `--font-size` variable.

We are using rem though, not em. rem is relative to the root (html) so given how native/mdpopups and our styles are constructed, it should still be unaffected by what mdpopups does.

mdpopups sets font-size in `.mdpopups` and we set it in `.lsp_popup`. Nothing seems to explicitly set it for root element (`html`).

```
<div class="mdpopups">
    <div class="lsp_popup">
    ...
    </div>
</div>
```

It's still possible that some of the overrides are not working properly. Do you have specific example where this issue can be seen? Maybe try to make a significant change to our `--font-size` variable to make the issue more pronounced for investigation?

jwortmann: > It's still possible that some of the overrides are not working properly. Do you have specific example where this issue can be seen? Maybe try to make a significant change to our `--font-size` variable to make the issue more pronounced for investigation?

I have set `--font-size: 2rem` for the following screenshot.

hover

The inline code font is smaller than the font for code blocks. Maybe that is not a bad thing, because it looks fine when `--font-size` is set to 1rem. I actually like it that the inline code is a little bit smaller than the code block font here:

hover2

jwortmann: I have set the small font for inline code explicitly in https://github.com/sublimelsp/LSP/pull/2788/changes/4a13c44210e37b922851ef415507d49ab3573f79

jwortmann:
> Yeah I'm not sure if there are really 3 different font sized needed. I think I would use the smaller font size for the signature help header (the "1 of 3 overloads" line), and probably also for the symbol action links at the bottom of the popup (if enabled), because those are not part of the content. And for everything else I would use the regular font size. So in that case there would only be a --font-size and a --font-size-sm variable.

Any comments/opinions on that? I pushed a change to implement that in https://github.com/sublimelsp/LSP/pull/2788/changes/27776be965771cf36aa81f95d06492a5708a7488 and https://github.com/sublimelsp/LSP/pull/2788/changes/f6e5724798990861b8430a018c2d012a97cff4b7. We can revert it if you want.

Symbol action links now use the smaller font:
hover

I set the small font for Mac/Linux to 0.95rem, which is between the previous `--font-size-sm: 1rem` and `--font-size-xs: 0.9rem`. Feel free to adjust it if you think a different value fits better.

---

We could also remove the underline from those links:

hover2

What do you think?

popups.css
jwortmann: By the way, what I meant is that in the 2nd screenshot the code block font appears significantly larger than the text font. But both scale witht the `--font-size` variable. So my guess would be that there is still some influence from mdpopups defaults. I could be wrong though if it is just caused by different relative sizes of the used fonts.

> For example the code blocks have a little bit larger font than the text, even though both use the --font-size variable.

rchl:
> We could also remove the underline from those links:
> hover2
> 
> What do you think?

I'm up for it in cases when link clearly looks like a link without the underline.

popups.css
rchl: The different font size comes from the fact that it's a different font family. Even though computed font-size is exactly the same, rendered size is different. I guess different fonts can distribute within available space differently.

jwortmann:
> I'm up for it in cases when link clearly looks like a link without the underline.

So is this such a case?

Could you please check whether font sizes still look good on Mac with the latest changes? For example with the signature help popup.

Also should the font sizes on Linux follow the sizes from Mac (which should be the case now), or the sizes from Windows?

Feel free to do some adjustments if needed. Otherwise I think it looks good from my side now.

rchl:
> > I'm up for it in cases when link clearly looks like a link without the underline.
> 
> So is this such a case?

Yes

> Could you please check whether font sizes still look good on Mac with the latest changes? For example with the signature help popup.

Normal text is almost the same as on main. Perhaps a pixel bigger. I guess that's fine.

> Also should the font sizes on Linux follow the sizes from Mac (which should be the case now), or the sizes from Windows?

@predragnikolic can probably check

> Feel free to do some adjustments if needed. Otherwise I think it looks good from my side now.

jwortmann:
> Normal text is almost the same as on main. Perhaps a pixel bigger. I guess that's fine.

It should still use the 1.05rem on Mac that you set before. Can you check that the `{% else %}` in the CSS actually works and those sizes are applied on Mac?

rchl:
> > Normal text is almost the same as on main. Perhaps a pixel bigger. I guess that's fine.
> 
> It should still use the 1.05rem on Mac that you set before. Can you check that the `{% else %}` in the CSS actually works and those sizes are applied on Mac?

All is applied correctly. I guess my change wasn't that significant.

popups.css
```diff
@@ -1,40 +1,95 @@
...
-}
-.lsp_popup hr {
+.lsp_popup {
+{% if plugin.is_windows %}
+    --font-size: 1rem;
+    --font-size-sm: 0.9rem;
+{% else %}
+    --font-size: 1.05rem;
+    --font-size-sm: 0.95rem;
+{% endif %}
```
predragnikolic: > >Also should the font sizes on Linux follow the sizes from Mac (which should be the case now), or the sizes from Windows?

> @predragnikolic can probably check

The change looks fine.

I did check today, I assume `--font-size-sm: 0.95rem` was used because xs is 0.9rem and the normal size is 1rem... so 0.95 was the value in between those values.

It is fine. It is even fine if everything is set to 1rem, or 1.05rem or 0.9rem or 0.95rem.
In practice I don't notice a huge difference.

popups.css
```diff
@@ -1,40 +1,95 @@
...
+    font-size: var(--font-size);
+}
+.lsp_popup code.highlight {
+    font-size: var(--font-size-sm);
+}
+.lsp_popup .font-size-sm,
+.lsp_popup .font-size-sm div.highlight,
+.lsp_popup .font-size-sm pre.highlight {
+    font-size: var(--font-size-sm);
+}
```
rchl: `font-size-sm` class is only used once in signature help "intro" so perhaps it's better to just remove it and instead use a class like `signature-help-intro`. Also that place doesn't contain nested `.highlight` blocks so those rules are not that relevant.

popups.css
predragnikolic: but it would be really nice to get rid of the `plugin.is_window` check and use one value:
- 1rem or 1.05rem for --font-size
- 0.9rem or 0.95rem for --font-size-sm

If we don't cannot come to a common conclusion for those values, 
I would suggest even removing code entirely and let mdpopups decide.

popups.css
rchl: Or instead of `.signature-help-intro` and `.actions` selectors, which just set font-size-sm, we could use `.font-size-sm` class in the HTML.

I guess it can make a difference if someone wants to override styles. In the former cases each of those places can be overridden separately, in latter case one can only override style for `.font-size-sm` itself (or `--font-size-sm` variable in general).

popups.css
rchl: I'm OK with using the windows defaults on Mac.

predragnikolic: Windows defaults on Linux also look ok

popups.css
jwortmann: > You are saying that code action should always reference existing diagnostics but this is in practice not really enforced. Also it would be our job to match diagnostics attached to code action against actual diagnostics which could be a bit tricky? And what if we don't find matching diagnostic? We'll show code action as we do now?
> 
> This change raises a bunch of questions and potential issues so I certainly wouldn't make it part of this PR.

I missed this comment somehow and only saw it now.

Yes, I'd like to try that out and potentially make another PR for that after this one is merged.

For code actions that are shown as annotations in the view, or when hovering the lightbulb icon in the gutter, only "quickfix" actions are shown. This includes code actions which don't have the optional `kind` property set, because for historic reasons code actions without kind are considered to be quick fixes (I would have put a reference here, but I can't find it right now, although I think to remember that `@dbaeumer` wrote that in some comment in the specs repo).

https://github.com/sublimelsp/LSP/blob/833541709b40223a26b0cf3f7cf1c64a722eb185/plugin/code_actions.py#L120-L127

I was under the impression that we would do the same for the hover popup, but it seems that filtering is not applied there. I think maybe we should do that, and in fact only show the quickfix code actions that resolve a diagnostic which is visible in the popup. If there are multiple code actions (for one or multiple diagnostics) we could show them all, one line per code action, instead of a single link "choose (3 available)" that opens the command palette.

With the current behavior there can be popups containing only a code action even when you hover over empty space:
code_action_hover
In the case from this screenshot, the code action range spans the entire block from `@testset` to an `end` below (outside of the screenshot). This might be a bit of an exotic case, because the same action is also shown as a code lens, but since the author of that language server uses the Zed editor, and Zed doesn't support code lenses, they added it also as a code action. But I think something like this looks and behaves a bit strange.

I would probably keep the refactor actions out of the view (can be triggered manually via context menu or main menu), because I don't want to see some suggestion like "extract this code into separate function" or similar at every occasion when I hover something.

Personally I had code actions disabled in the popup via `"show_code_actions_in_hover": false` setting, but I think I would enable them if shown as part of the diagnostics.

popups.css
jwortmann: Sounds good. I would say pick the solution you prefer.

rchl:
Did a bunch of simplifications in ed86e912fa3af6c498ad1e6cb02ddacd0ce69e70 to mainly use `.wrapper` class in more places which gives us default padding and handles margin collapsing bug (just in case).

This should introduce no practical changes ideally.

popups.css
rchl: The action links are a bit small though on Mac... Two screenshots showing Mac and Windows font-size in use.

Screenshot 2026-03-18 at 21 11 17
Screenshot 2026-03-18 at 21 10 51

But I guess it's fine?

rchl:
> 
> 
> And the same message from Pyright (no MarkupContent support for diagnostics):
> 
> 
> I don't find it less readable than the diagnostics with monospace font that we currently use.
> 
> _Edit:_ Here is a diagnostic from Pyright with more code in it:
> 
> 
> And with monospace font:
> 
> 
> In this case I'm not really sure what is better. The monospace font is probably more readable for messages with such large amounts of code, but on the other hand the proportional font is more compact, which could result in fewer linebreaks.
> 
> I guess we could also use monospace font if the message is a plain string (Pyright), and the proportional font (excluding code markup) if the message is MarkupContent. But that would be a bit inconsistent...

I can't quite get used to proportional font for diagnostic texts.

Should we change back to monospace?

predragnikolic:
> Should we change back to monospace?

Don't ask me (I have set my whole os font to be monospaced 😅 I just find it easier to read)

jwortmann:
I like the proportional font.

PR #2789, opened 2026-03-02
`code_actions_on_format` improve action redundancy detection

mj026:
This PR is a follow up of #2747 and #2761 and  targeted at this remark https://github.com/sublimelsp/LSP/pull/2761#discussion_r2814184733 regarding action redundancy:

- ~~Commit 4ea88b402b01f01283b15ea2c14e6cb7f2243e0f
  Filter out disabled code actions, for example `"source.organizeImports.ruff": false`. These are also filtered out later in "the chain" but I do not see why we shouldn't do it right away. Now done in `get_matching_on_save_kinds`~~ We can't do this as we should ask the server which actions are supported and act accordingly.
EDIT: Doesn't make any sense so this has been reverted.

- ~~Commit 71434eccaa45c6008cc2f80146071621eb71b4c
  Checks if a "parent action" (so `source.organizeImports` when the specified action is `source.organizeImports.ruff`)  is already present in `code_actions_on_save` so we can consider it to be redundant. Bonus feature: Because we filter out disabled actions now (see first point), we do not skip actions defined in`code_actions_on_format` when the same actions is set to `false` in `code_actions_on_save`.~~
  
~~I now implemented this with a regex, alternative implementation could be like this (I like the regex version more)~~: ~~Implemented now by reusing the existing `kind_contains_other_kind` function.~~

EDIT: Now the `CodeActionsOnFormatOnSaveTask` task is independent from other tasks, the implementation is totally different.

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -233,7 +233,7 @@ class CodeActionsTaskBase(LspTask):
     @classmethod
     @override
     def is_applicable(cls, view: sublime.View) -> bool:
-        return bool(view.window()) and bool(cls.get_code_actions(view))
+        return bool(view.window()) and bool(cls.get_code_action_kinds(view))
 
     @classmethod
     def format_on_save_enabled(cls, view: sublime.View) -> bool:
@@ -241,20 +241,20 @@ class CodeActionsTaskBase(LspTask):
         return view_format_on_save if isinstance(view_format_on_save, bool) else userprefs().lsp_format_on_save
 
     @classmethod
-    def get_code_actions(cls, view: sublime.View) -> dict[str, bool]:
-        view_code_actions = cast('dict[str, bool]', view.settings().get(cls.SETTING_NAME) or {})
-        code_actions = getattr(userprefs(), cls.SETTING_NAME, {}).copy()
-        code_actions.update(view_code_actions)
+    def get_code_action_kinds(cls, view: sublime.View) -> dict[str, bool]:
+        view_code_action_kinds = cast('dict[str, bool]', view.settings().get(cls.SETTING_NAME) or {})
+        code_action_kinds = getattr(userprefs(), cls.SETTING_NAME, {}).copy()
+        code_action_kinds.update(view_code_action_kinds)
         return {
-            action: enabled for action, enabled in code_actions.items() if action.startswith(CodeActionKind.Source)
+            kind: enabled for kind, enabled in code_action_kinds.items() if kind.startswith(CodeActionKind.Source)
         }
 
     @override
     def run_async(self) -> None:
         super().run_async()
         view = self._task_runner.view
-        code_actions = self.get_code_actions(view)
-        request_iterator = actions_manager.request_on_save_or_format_async(view, code_actions)
+        code_action_kinds = self.get_code_action_kinds(view)
+        request_iterator = actions_manager.request_on_save_or_format_async(view, code_action_kinds)
         self._process_next_request(request_iterator)
 
     def _process_next_request(self, request_iterator: Iterator[Promise[CodeActionsByConfigName]]) -> None:
@@ -313,15 +313,15 @@ class CodeActionsOnFormatOnSaveTask(CodeActionsOnFormatTask):
 
     @classmethod
     @override
-    def get_code_actions(cls, view: sublime.View) -> dict[str, bool]:
-        code_actions_on_save = CodeActionsOnSaveTask.get_code_actions(view)
-        code_actions_on_format = super().get_code_actions(view)
-        # Merge the code actions which should be run. A duplicate code action will be enabled when
+    def get_code_action_kinds(cls, view: sublime.View) -> dict[str, bool]:
+        code_action_kinds_on_save = CodeActionsOnSaveTask.get_code_action_kinds(view)
+        code_action_kinds_on_format = super().get_code_action_kinds(view)
+        # Merge the code action kinds which should be run. A duplicate code action kind will be enabled when
         # it's enabled in one of the two available settings.
-        actions = set(code_actions_on_save.keys()).union(code_actions_on_format.keys())
+        action_kinds = set(code_action_kinds_on_save.keys()).union(code_action_kinds_on_format.keys())
         return {
-            action: code_actions_on_save.get(action, False) or code_actions_on_format.get(action, False)
-            for action in actions
+            kind: code_action_kinds_on_save.get(kind, False) or code_action_kinds_on_format.get(kind, False)
+            for kind in action_kinds
         }
 
     @classmethod
--- a/tests/test_code_actions.py
+++ b/tests/test_code_actions.py
@@ -371,7 +371,7 @@ class CodeActionsOnFormatOnSaveTaskTestCase(TextDocumentTestCase):
         # Actions defined in both settings are merged. When a duplicate action is found it will be True (enabled)
         # when enabled in lsp_code_actions_on_save or lsp_code_actions_on_format
         self.assertEqual(
-            CodeActionsOnFormatOnSaveTask.get_code_actions(view=self.view),
+            CodeActionsOnFormatOnSaveTask.get_code_action_kinds(view=self.view),
             {"source.fixAll": True, "source.organizeImports": True, "source.sort.json": False},
         )
 


plugin/code_actions.py
```diff
@@ -301,16 +302,24 @@ class CodeActionsOnFormatTask(CodeActionsTaskBase):
 class CodeActionsOnFormatOnSaveTask(CodeActionsOnFormatTask):
     """Run code actions on format when format_on_save is enabled."""
 
+    @staticmethod
+    def action_is_redundant(action: str, actions: dict[str, bool]) -> bool:
+        # Checks if the specified 'action' (source.fixAll.eslint) or the parent (source.fixAll,  when applicable)
+        # is already present in 'actions'
+        if (match_parent_action := re.match(r"([^.]+\.[^.]+(?=\.))", action)):
+            return action in actions or match_parent_action[0] in actions
+        return action in actions
```
jwortmann: I think there is already a function at https://github.com/sublimelsp/LSP/blob/cdce162bba15368f8f9fa1ce19a8d1cc0008e753/plugin/core/views.py#L873-L883
which you could use here. Then this function would become
```python
    @staticmethod
    def action_is_redundant(action: str, actions: dict[str, bool]) -> bool:
        return any(kind_contains_other_kind(kind, action) for kind in actions)
```
(This could be inlined in the code below. Also I haven't confirmed if that actually works or if my thinking is wrong, so it should better be tested first.)

plugin/code_actions.py
```diff
@@ -243,7 +244,9 @@ def get_code_actions(cls, view: sublime.View) -> dict[str, bool]:
         code_actions = getattr(userprefs(), cls.SETTING_NAME, {}).copy()
         code_actions.update(view_code_actions)
         return {
-            key: value for key, value in code_actions.items() if key.startswith(CodeActionKind.Source)
+            action: enabled
+            for action, enabled in code_actions.items()
+            if enabled and action.startswith(CodeActionKind.Source)
```
rchl: If you are now returning only enabled actions then what's the point of returning a dict?

But in fact we have to return disabled actions or else cases like this won't be handled properly:

```
{
  "source.fixAll": true,
  "source.fixAll.eslint": false
}
```

(at least I believe so, have not looked through the code again)

mj026: I’m not sure I understand why this usecase wouldn’t be handled properly? (`source.fixAll` will be applied and `source.fixAll.eslint` will not, which is what I expect?)

rchl: When we are asking server for `source.fixAll` and returned action is actually `source.fixAll.eslint` then how are we gonna know that the latter is to be ignored if we have already filtered out the false values?

The latter is a subset of the former.

rchl: Well, I suppose that matching happens before actually requesting - when making requests we are asking specifically for those that server advertizes supporting. But still, that matching happens later, after the code quoted here.

mj026: Ah ok, I understand now. Then filtering out the disabled actions _up front_ makes this more complicated instead of easier and factually wrong. I'll look at it again and will adjust it (I'll probably remove this part completely).

plugin/code_actions.py
mj026: I think you're right and this makes the implementation even more simple. I'll adjust this.

plugin/code_actions.py
mj026: I reverted the changes regarding filtering out disabled actions in commit 106bbcf9352df9b07dd03a4e41e1ff528e1161ba as it is indeed wrong to do so.

plugin/code_actions.py
mj026: Now implemented like you suggested and with manual testing it works as expected (the added unit test also confirmes this).

plugin/code_actions.py
```diff
@@ -303,12 +303,12 @@ class CodeActionsOnFormatOnSaveTask(CodeActionsOnFormatTask):
...
-        # Prevent triggering of duplicate code actions
+        code_actions_on_save = CodeActionsOnSaveTask.get_code_actions(view)
+        # Prevent triggering of redundant and duplicate code actions
         return {
-            action: enabled
-            for action, enabled in code_actions_on_format.items()
-            if action not in code_action_on_save
+            format_action: enabled
+            for format_action, enabled in code_actions_on_format.items()
+            if not any(kind_contains_other_kind(save_action, format_action) for save_action in code_actions_on_save)
```
jwortmann: Do we need to do this check in "both directions" in some sense?

Let's say the server advertises support for `source.fixAll.eslint` code action kind.
For example, this scenario should already be covered by this PR:
```json
{
    "lsp_code_actions_on_format": {
        "source.fixAll.eslint": true
    },
    "lsp_code_actions_on_save": {
        "source.fixAll": true
    }
}
```
Here `source.fixAll.eslint` is already included in `source.fixAll`, so we only send a single request for `source.fixAll.eslint` due to the on-save actions.

But consider
```json
{
    "lsp_code_actions_on_format": {
        "source.fixAll": true
    },
    "lsp_code_actions_on_save": {
        "source.fixAll.eslint": true
    }
}
```
Here `source.fixAll` is not included in `source.fixAll.eslint`, so it is not filtered out, and as a result one request is sent due to on-save actions for `source.fixAll.eslint`, and after that another request with the same kind again is sent due to format-on-save.

But this is a bit tricky to fix. Because consider if the server advertises support for `source.fixAll.eslint` and `source.fixAll.meow`. In this case we cannot simply filter out the `source.fixAll` from the "lsp_code_actions_on_format" setting, because then we wouldn't send a request for `source.fixAll.meow`.

jwortmann: I think we need to keep track of the kinds that are actually used for the code-actions-on-save request and store that somewhere. Then before sending the code-actions-on-format request, we would just remove the exact kinds that were already sent in the previous request (and if the resulting kinds is empty, don't send another request at all). That should work even without the need to compare the user settings.

mj026: Because we now split it up in different (almost) self containing tasks (based on `CodeActionsTaskBase`) there isn't a clean way how to store it (I think) without breaking this (almost) independent task paradigm (which I really like btw) with some state / session management elsewhere.

But what if we solve it like so: When `lsp_format_on_save` is enabled, we do not run the `CodeActionsOnSaveTask` (so we check this in the `is_applicable` method for the `lsp_format_on_save` setting). Then we collect **all code actions** defined in `lsp_format_on_save` **and** `lsp_format_on_format`  in the `CodeActionsOnFormatOnSaveTask` task. 

So in this scenario, we have all code actions _in one place_ (in the`get_code_actions` method of `CodeActionsOnFormatOnSaveTask`. We now can do additional filtering (keep track) in the part of the codebase where the actual requests are send (as you suggested above) because now this is handled in the same task.

jwortmann: Yeah that should work too, and I like this idea.

mj026: Ok! 👍 I'll try to implement this in the following days and will post the progress here.

mj026: So I implemented this idea and added new commits to this PR:
- `CodeActionsOnSaveTask` 
    Only runs _on save_ and when `format_on_save` is **disabled** (and when `lsp_code_actions_on_save` is set)
    
- `CodeActionsOnFormatTask` 
     Only runs _on format_ (and when `lsp_code_actions_on_format` is set)

- `CodeActionsOnFormatOnSaveTask` 
    Only runs _on save_ and when `format_on_save` is **enabled** (and when `lsp_code_actions_on_save` and/or `lsp_code_actions_on_format`is set, these settings are now merged to produce a single list of code actions).

As the code actions from both `lsp_code_actions_on_save` and `lsp_code_actions_on_format` are now merged in the `CodeActionsOnFormatOnSaveTask`, we don't need to keep track of which code action is already run as no duplicates can exist. This single list of code actions (in a single task) will be filtered with `get_matching_kinds` (I renamed `get_matching_on_save_kinds`). 

I didn't need to bring in any code for duplicate detection as there cannot be any and `get_matching_kinds` is already checking for code actions which are available in the current session and also checks if code actions are enabled / disabled.

I'd say this is now a pretty clean solution. All tasks are now self containing and the implementation is simple and the outcome is what I expect it would be.

 Let's see what you guys think ;)

plugin/code_actions.py
```diff
@@ -297,27 +307,23 @@ class CodeActionsOnFormatTask(CodeActionsTaskBase):
...
+        code_actions_on_save = CodeActionsOnSaveTask.get_code_actions(view)
         code_actions_on_format = super().get_code_actions(view)
-        code_action_on_save = CodeActionsOnSaveTask.get_code_actions(view)
-        # Prevent triggering of duplicate code actions
-        return {
-            action: enabled
-            for action, enabled in code_actions_on_format.items()
-            if action not in code_action_on_save
-        }
+        return {**code_actions_on_save, **code_actions_on_format}
```
jwortmann: This simply overrides values from `code_actions_on_save` with values from `code_actions_on_format` for identical keys. I think we would expect a OR operation here (?)

Consider
```json
{
    "lsp_code_actions_on_format": {
        "source.fixAll.eslint": false
    },
    "lsp_code_actions_on_save": {
        "source.fixAll.eslint": true
    }
}
```

As far as I can tell, this would result in no code actions when saving a file and format_on_save is enabled. But it should actually run the code action, right?

mj026: Makes sense. I will adjust it to merge the code action dictionaries with `OR` for identical actions.

plugin/code_actions.py
```diff
@@ -199,8 +199,8 @@ def get_session_kinds(sb: SessionBufferProtocol) -> list[CodeActionKind]:
     return sb.get_capability('codeActionProvider.codeActionKinds') or []
 
 
-def get_matching_on_save_kinds(
-    user_actions: dict[str, bool], session_kinds: list[CodeActionKind]
+def get_matching_kinds(
+    code_actions: dict[str, bool], session_kinds: list[CodeActionKind]
```
rchl: fits one line now

plugin/code_actions.py
```diff
@@ -297,27 +307,29 @@ class CodeActionsOnFormatTask(CodeActionsTaskBase):
...
 
     @classmethod
     @override
     def get_code_actions(cls, view: sublime.View) -> dict[str, bool]:
+        code_actions_on_save = CodeActionsOnSaveTask.get_code_actions(view)
         code_actions_on_format = super().get_code_actions(view)
-        code_action_on_save = CodeActionsOnSaveTask.get_code_actions(view)
-        # Prevent triggering of duplicate code actions
+        # Merge the code actions which should be run. A dupliacte code action will be enabled when
+        # it's enabled in one of the two available settings
```
rchl:
```suggestion
        # Merge the code actions which should be run. A duplicate code action will be enabled when
        # it's enabled in one of the two available settings.
```

plugin/code_actions.py
```diff
@@ -297,27 +307,29 @@ class CodeActionsOnFormatTask(CodeActionsTaskBase):
...
+        actions = set(code_actions_on_save).union(code_actions_on_format)
```
rchl: Seeing this I was wondering what's the behavior of converting dict to set. I think it would be clearer to be explicit:


```suggestion
        actions = set(code_actions_on_save.keys()).union(code_actions_on_format.keys())
```

plugin/code_actions.py
```diff
@@ -297,27 +305,29 @@ class CodeActionsOnFormatTask(CodeActionsTaskBase):
 
 @final
 class CodeActionsOnFormatOnSaveTask(CodeActionsOnFormatTask):
-    """Run code actions on format when format_on_save is enabled."""
+    """
+    Run code actions on format when format_on_save is enabled.
+
+    Code actions defined in both 'lsp_code_actions_on_save' and 'lsp_code_actions_on_format' will be run.
```
jwortmann:
```suggestion
    Code actions enabled in either 'lsp_code_actions_on_save' or 'lsp_code_actions_on_format' will be run.
```

plugin/code_actions.py
```diff
@@ -237,13 +235,18 @@ class CodeActionsTaskBase(LspTask):
...
+        return view_format_on_save if isinstance(view_format_on_save, bool) else userprefs().lsp_format_on_save
+
     @classmethod
     def get_code_actions(cls, view: sublime.View) -> dict[str, bool]:
         view_code_actions = cast('dict[str, bool]', view.settings().get(cls.SETTING_NAME) or {})
         code_actions = getattr(userprefs(), cls.SETTING_NAME, {}).copy()
         code_actions.update(view_code_actions)
         return {
-            key: value for key, value in code_actions.items() if key.startswith(CodeActionKind.Source)
+            action: enabled for action, enabled in code_actions.items() if action.startswith(CodeActionKind.Source)
```
jwortmann: If we want to be really accurate with the terminology, this line would be
```python
kind: enabled for kind, enabled in code_action_kinds.items() if kind.startswith(CodeActionKind.Source)
```
and the method could be named `get_code_action_kinds`.

A [code action](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeAction) is actually what gets returned from the server, and the strings like `"source.fixAll"` would be the code action [kind](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeActionKind).

But I would be okay to leave it like it is now, if you don't want to do these renames here and in the other methods.

mj026: Let's do it proper while we are working on it. I'll implement your suggestions in a minute!

mj026:
🎉Thank you @rchl and @jwortmann!

PR #2790, opened 2026-03-02
Include view settings in project files schema

rchl:
Make it so that LSP-json recognizes `lsp_*` settings in project files.

Also updated `Preferences.json` schema to include `lsp_code_actions_on_format` as it was forgotten.

### DIFF
--- a/sublime-package.json
+++ b/sublime-package.json
@@ -875,18 +875,20 @@
           "/Preferences.sublime-settings"
         ],
         "schema": {
-          "lsp_format_on_paste": {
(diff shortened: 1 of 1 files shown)


sublime-package.json
```diff
@@ -836,16 +875,17 @@
           "/Preferences.sublime-settings"
         ],
         "schema": {
-          "properties": {
```
jwortmann, line 839: Was it intended to remove this line?

rchl: no, good catch

jwortmann (approved):

PR #2792, opened 2026-03-05
Docs: rewrite feature page

jwortmann:
I have rewritten the "Features" page and combined that with the respective sections from the "Customization" page. I think that is better than having the information about a feature split into two different locations.

I added some images instead of the GIF links, and tried to remove filler sentences that carry no real information like

> The auto-complete functionality of Sublime Text is well-known to any user.

There are also new sections for features that were not yet included on the features page (semantic highlighting, call hierarchy, goto diagnostic).

Testing how it looks - open to suggestions if you have ideas for improvements.

### DIFF
--- a/docs/src/features.md
+++ b/docs/src/features.md
@@ -16,10 +16,10 @@ Almost all capabilities can also be bound to a key with a key binding.
 ![Auto Complete](./images/auto_complete.png)
 
 The LSP package enhances the auto-complete functionality of Sublime Text with results provided by the language server.
-If available, you can click the **More** link or use the default key binding <kbd>F12</kbd> to show an additional documentation popup with detailed information about the highlighted completion item.
+If available, you can click the **More** link or use the default key binding <kbd>F12</kbd> to show an additional documentation popup with detailed information about the highlighted item.
 
 Some language servers provide two different modes for inserting a completion item when the caret is in the middle of a word, in which case **Replace** or **Insert** is shown at the bottom of the auto-completion popup.
-The default insertion mode can be configured with the `"completion_insert_mode"` setting, and the opposite mode can be used by selecting a completion item with the key binding <kbd>Alt</kbd><kbd>Enter</kbd>.
+The default insertion mode can be configured with the `"completion_insert_mode"` setting, and the opposite mode can be used by confirming a completion item with the key binding <kbd>Alt</kbd><kbd>Enter</kbd>.
 
 [Example GIF for "Replace" mode](https://user-images.githubusercontent.com/22029477/189607770-1a8018f6-1fd1-40de-b6d9-be1f657dfc0d.gif)
 
@@ -28,7 +28,7 @@ The default insertion mode can be configured with the `"completion_insert_mode"`
 
 ![Signature Help](./images/signature_help.png)
 
-The signature help popup appears when typing the arguments of a function invocation.
+The signature help popup appears when typing the arguments of a function call.
 It highlights the name of the current parameter and often presents additional type information and documentation of the function and parameters.
 If multiple overloads of the function exist, you can switch between them using the up and down arrow keys.
 
@@ -66,7 +66,7 @@ See the [mdpopups documentation](http://facelessuser.github.io/sublime-markdown-
 
 LSP can also highlight the word or range for which a hover popup is shown, if the `"hover_highlight_style"` setting is enabled.
 In that case you can use the scope `markup.highlight.hover` in a color scheme rule to control the highlighting color.
-If `"hover_highlight_style"` is set to `"background"`, the highlighting color can be controlled using the "background" property in the color scheme rule.
+If the setting is set to `"background"`, the highlighting color can be controlled using the "background" property in the color scheme rule.
 
 
 ## Highlights
@@ -76,12 +76,12 @@ If `"hover_highlight_style"` is set to `"background"`, the highlighting color ca
 When you select a word, Sublime Text highlights other occurrences of that word in the file (controlled by the `"match_selection"` setting).
 LSP has a similar capability to highlight the identifier name that is currently under the caret.
 It is enhanced in the sense that the highlighted locations are restricted to only the relevant part of the file, according to the scoping rules of the language.
-Furthermore it can distinguish between read and write access to a variable and may highlight them with different colors.
+Furthermore it can distinguish between read and write access of a variable and may highlight them with different colors.
 
 The highlighting color can be adjusted with color scheme rules for the following scopes:
 
-| DocumentHighlightKind | scope | description |
-| --------------------- | ----- | ----------- |
+| [Document Highlight Kind](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#documentHighlightKind) | scope | description |
+| --- | --- | --- |
 | Text | `markup.highlight.text` | A textual occurrence |
 | Read | `markup.highlight.read` | Read-access of a symbol |
 | Write | `markup.highlight.write` | Write-access of a symbol |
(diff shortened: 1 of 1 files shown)


docs/src/features.md
```diff
@@ -1,181 +1,365 @@
...
+The highlighting color can be adjusted with color scheme rules for the following scopes:
+
+| DocumentHighlightKind | scope | description |
+| --------------------- | ----- | ----------- |
+| Text | `markup.highlight.text` | A textual occurrence |
+| Read | `markup.highlight.read` | Read-access of a symbol |
+| Write | `markup.highlight.write` | Write-access of a symbol |
+
+If `"document_highlight_style"` is set to `"background"` in the LSP settings, the highlighting color can be controlled using the "background" property in the color scheme rule.
+
```
predragnikolic, line 90:
```jsonc
{
    "rules": [
        {
            "scope": "markup.highlight",
            "background": "#ffffff16"
        }
    ]
}
```

Maybe provide an example?

docs/src/features.md
```diff
@@ -1,181 +1,365 @@
...
+| Hint | `markup.info.hint` | stippled underline |
 
-Sublime Text subtly highlights words that equal the one under your caret.
+Diagnostics also optionally include the following scopes:
 
-This package does that as well, if your language server has that capability. It is enhanced in the sense that a language server can also annotate a region as being _written to_, to being _read from_. This is what stock Sublime Text cannot determine.
+| [Diagnostic Tag](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#diagnosticTag) | scope | description |
+| --- | --- | --- |
+| Unnecessary | `markup.unnecessary` | Unused or unnecessary code |
+| Deprecated | `markup.deprecated` | Deprecated or obsolete code |
```
predragnikolic, line 233: We kind of have a not so obvious feature that people like hidden behind the `show_diagnostics_annotations_severity_level` - it allows us to display diagnostics in the annotations

Image

docs/src/features.md
```diff
@@ -1,181 +1,365 @@
...
-[Example GIF 1](https://user-images.githubusercontent.com/6579999/128552021-d9058c65-d6f6-48f5-b7aa-652eafe23247.gif)
+| [Diagnostic Severity](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#diagnosticSeverity) | scope | drawn as |
+| --- | --- | --- |
+| Error | `markup.error` | squiggly underline |
+| Warning | `markup.warning` | squiggly underline |
+| Information | `markup.info` | stippled underline |
+| Hint | `markup.info.hint` | stippled underline |
 
-Sublime Text subtly highlights words that equal the one under your caret.
+Diagnostics also optionally include the following scopes:
```
predragnikolic, line 228: should we mention that it is possible to assign keybindings to go to the next/prev diagnotics?

predragnikolic: ah now I see that we have https://deploy-preview-2792--sublime-lsp.netlify.app/features/#goto-diagnostic

Maybe we can mention the keybindings there?

docs/src/features.md
```diff
@@ -1,181 +1,365 @@
...
+}
+```
 
-Goto Symbol can be accessed by clicking on Goto > Goto Symbol. A common key binding for this is <kbd>ctrl</kbd><kbd>R</kbd>. Sublime Text will show a Quick Panel where you can select a symbol from the current buffer. This package provides a replacement if your language server has this capability. Each symbol's type is accurately described, and the start and end positions of each symbol are clearly visible.
 
-## Goto Symbol In Project
+## Code Actions
 
-[Example GIF 1](https://camo.githubusercontent.com/e7a6dd90838b0a8dab8c08d7846c86bfe0e271e7f213e362ba1b07df2d90c156/68747470733a2f2f692e696d6775722e636f6d2f385830584e69322e676966)
+Code actions is an umbrella term for "Quick Fixes" and "Refactorings".
```
predragnikolic, line 254: maybe mention code_actions_on_save code_actions_on_format?

docs/src/features.md
```diff
@@ -1,181 +1,365 @@
...
+<video src="/videos/file-rename.mp4" controls></video>
 
-Inlay hints are short textual annotations that show parameter names and type hints for servers that support that feature.
 
-![inlay-hints](./images/inlay-hints.png)
+## Expand Selection
 
-Inlay hints are disabled by default and can be enabled with the `"show_inlay_hints": true` setting through `Preferences: LSP Settings`.
+Sublime Text has the built-in functionality to expand the current selection, with the default key binding <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>A</kbd>.
+A language server may also have this capability and is in a better position to decide what a "smart" Expand Selection should do.
```
predragnikolic, line 365:
```  
  // Expand Selection (a replacement for ST's "Expand Selection")
    // {
    //     "keys": ["primary+shift+a"],
    //     "command": "lsp_expand_selection",
    //     "args": {"fallback": false},
    //     "context": [{"key": "lsp.session_with_capability", "operand": "selectionRangeProvider"}]
    // },
```
Because it is not obivous on how to do the oposite action,
maybe we should also add document that we can shrink a selection someone could do:
```
    {
        "keys": ["UNBOUND"],
        "command": "soft_undo",
        "context": [{"key": "lsp.session_with_capability", "operand": "selectionRangeProvider"}]
    },
```

rchl: soft_undo is by default assigned to cmd+u (on mac but I assume also on other systems) and doesn't need a custom binding. Especially with `context` :)

docs/src/images/auto_complete.png
rchl, line 1: do you have a consistent setup that can be used to make screenshots in the future? :)

I guess it's just stock profile basically

docs/src/features.md
jwortmann: But we don't have any default key bindings for those commands. The commands are listed on the "Keyboard shortcuts" page, so I don't think we need to repeat them here.

docs/src/features.md
jwortmann: I added a sentence and screenshot for that.

docs/src/images/auto_complete.png
jwortmann: I used the Default Dark theme and Mariana color scheme, and I have kept some adjustments for the LSP settings that I prefer over the defaults, like `"document_highlight_style": "background"` and `"show_symbol_action_links": false`. And I used some adjustments for the style in the signature help popup, that I'm considering to propose as new defaults and scope adjustments in another PR (maybe soon). That would look more like the default style in VSCode. Currently the `entity.name.function` is mostly used for punctuations, which doesn't really match regular syntax highlighting, so I would change that maybe.

The screenshots have a width of 756 pixel, because that seems to be size that is used to scale down images by the Zensical website (at least on full HD display).

And the font is a custom variant of [Recursive Code](https://github.com/arrowtype/recursive-code-config), which I use for a few years now.

If you have ideas for more screenshots that would be useful, I could make them, so that we keep them looking consistent.

docs/src/features.md
jwortmann: There are already two other color scheme rule examples a bit lower on the page, and here the colors are up to the user, so I think I would not add another example here.

docs/src/features.md
jwortmann: Included in https://github.com/sublimelsp/LSP/pull/2792/changes/6fdaae78205b633e91e1c6f0e15cdd9729bcfcf8

docs/src/features.md
predragnikolic: True.
I guess what I wanted to emphasize is the, I did not know that I can use soft_undo to shrink a selection after I expanded it. 

I am fine if we don't document it, but I do that that such info could help someone

predragnikolic: Btw. I'm fine if it is not documented(maybe it would been better to just mention that info in the key bindings file).
I already found this PR useful. This is just a minor thing, so I will resolve it.

jwortmann:
I think this looks good now.

PR #2794, opened 2026-03-07
Adjust scopes and default style in signature help popup

jwortmann:
This is a suggestion to update the default styles in the signature help popup. The new style only colors the active parameter and uses bold font by default (instead of bold + underline). I believe this is similar to the style that is used in VSCode.

With these changes it looks like in the screenshot at https://lsp.sublimetext.io/features/#signature-help

This should be mostly backwards compatible to previous user modifications (in the color scheme) for the active parameter style. And now it is also possible to use italic font style for the active parameter.

---

A reason why I'd like to do this change is because the previous scope `entity.name.function.sighelp` was used for everything else than the parameters, which typically consists of parentheses, comma between the parameters and possibly return type of the function. The actual function name might not even appear in the signature help line (e.g. with Pyright), so that scope wasn't very good and resulted in unusual colors.
(Also in my opinion the new style looks better in general and makes it easier to see the active parameter.)

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -2111,6 +2111,7 @@ class Session(APIHandler, TransportCallbacks['dict[str, Any]']):
         session_view.shutdown_async()
 
     def _handle_shutdown_result(self, _: Any) -> None:
+        self._progress.clear()
         self.exit()
 
     def on_transport_close(self, exit_code: int, exception: Exception | None) -> None:


plugin/core/signature_help.py
```diff
@@ -183,16 +184,21 @@ def _parameter(self, content: str, active: bool) -> str:
...
+                italic=self._style['active_parameter_italic'],
+                underline=self._style['active_parameter_underline']
             )
         return _wrap_with_color(content, self._style['inactive_parameter_color'])
 
 
-def _wrap_with_color(content: str, color: str, bold: bool = False, underline: bool = False) -> str:
+def _wrap_with_color(
+    content: str, color: str, *, bold: bool = False, italic: bool = False, underline: bool = False
+) -> str:
```
rchl: Can you change it the same way as I did in 8d495193609f02880c7d75bbb90dfc13b5578f1f ?

`*_with_style` seems more appropriate since it's not only setting color and then `color` should also be a keyword argument.

jwortmann: https://github.com/sublimelsp/LSP/pull/2794/changes/4bafc5d117decc291161659da2570b37d47868d5

PR #2796, opened 2026-03-08
Fix false-positive PackageDev linter warnings

deathaxe:
This commit adds a settings hints file for user and project-specific settings to tell PackageDev about their existence and validity.

Without this setting PackageDev marks all of them invalid by underlining them, in Preferences and project-specific `"settings"` keys.

_Despite LSP-json proving superior code intelligence features, numerous packages still lack proper subliem-package.json files to provide required information. In such cases PackageDev still helps to fill the gaps._

### DIFF
--- a/Preferences.sublime-settings-hints
+++ b/Preferences.sublime-settings-hints
@@ -41,6 +41,6 @@
   // existing configurations. The keys of this dictionary are free-form. They
   // give a humany-friendly name to the server configuration. They are shown
   // in the bottom-left corner in the status bar once attached to a view
-  // (unless you have `"show_view_status"` set to `false`).
+  // (unless you have "show_view_status" set to `false`).
   "LSP": {}
 }


rchl:
If it would be just something additional that wouldn't affect LSP-json then it would probably be fine but as is it overrides LSP-json tooltips which I don't like.

Before:

Screenshot 2026-03-10 at 18 22 15

After:

Screenshot 2026-03-10 at 18 24 13

The content looks broken/out of place also. Especially when hovering the `LSP` key.

Screenshot 2026-03-10 at 18 26 17

The documentation should be adjusted for the use case and not just copied from somewhere else. Which then creates the issue of adding more maintenance...

deathaxe:
The issue of fighting popups basically exists everywhere, both PackageDev and LSP-json offer something. That's not unique nor related with this PR. I've always seen LSP-json winning.

Unfortunatelly it is not easy to detect, when LSP-json offers something.

The tooltip's look would be the same as for the `"clients"` setting in LSP.sublime-settings

rchl:
> Unfortunatelly it is not easy to detect, when LSP-json offers something.

What about PackageDev detecting that LSP is running by using `view.settings().get('lsp_active')`?
It doesn't tell whether LSP-json is running but it's close enough and most likely case when having json file open.

rchl:
Actually with the `"settings.tooltip": false` PackageDev setting (which I always had enabled anyway) those don't override LSP-json so I'm fine with that.

That said tooltip still shows some "corrupted" messages. Here is a tooltip for the LSP key:

Screenshot 2026-03-21 at 20 26 19

Those `source.*` lines shouldn't be there.

deathaxe:
Those rendering issues are more or less unrelated to this PR, but more a limitation of how PackageDev scans comments and attaches them to settings.

Checking the setting to disable tooltips automatically seems reasonable to avoid conflicts. PackageDev is intended to jump in only, if no json schema is available, anyway, as it can't provide the quality of service LSP-json can.

deathaxe:
Unfortunately, the solely precense of a running LSP-json session doesn't tell anything about whether a proper json schema is available, literly causing PackageDev's features to be disabled and all related features relying on `view.settings().get('lsp_active')` being disabled.

So finally that's probably not close enough.

rchl (approved):

PR #2797, opened 2026-03-08
Create dataclasses with slots

deathaxe:
Follow-up of fae96235780432a221f97a328e8c8f4adc51fe50

This commit configures recently introduced data classes to use slots again.

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -416,7 +416,7 @@ class Settings:
             return [None] * 4  # default styling
 
 
-@dataclass
+@dataclass(slots=True)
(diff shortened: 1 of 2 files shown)


predragnikolic:
For python 3.8 the build failed with
> TypeError: dataclass() got an unexpected keyword argument 'slots'

deathaxe:
OK, an ST4202+ change then.

predragnikolic:
Note, we will probably do what you proposed once we start supporting only ST4202+

predragnikolic:
We will revisit this, when the time for this comes

PR #2799, opened 2026-03-09
Clear all progress reports when session ends

### LINKED ISSUE #2798: Bottom status bar is stuck after disable LSP server.
**Describe the bug**

Calling "LSP: Disable Language Server in Project" on JETLS while it's analyzing the project prevents the bottom status bar from refreshing.

**To Reproduce**

1. Enable JETLS on any Julia project.
2. Run "LSP: Disable Language Server in Project" on JETLS while it's analyzing the project.

**Expected behavior**

I expect the bottom status bar to be usable after a language server is disabled. Specifically, the part that displays the line number and the number of characters selected. I'm not 100% sure if this a JETLS bug, but I'd expect LSP to be robust to these kinds of things.

- OS: MacOS
- Sublime Text version: 4200
- LSP version: 4070-2.8.2
- Language servers used: JETLS

jwortmann:
Fixes #2798

Who had the idea to make erasing the status messages for progress reports to rely on the garbage collector, lol.

On the other hand this resulted in this fix being a single-line change :-)

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -2111,6 +2111,7 @@ class Session(APIHandler, TransportCallbacks['dict[str, Any]']):
         session_view.shutdown_async()
 
     def _handle_shutdown_result(self, _: Any) -> None:
+        self._progress.clear()
         self.exit()
 
     def on_transport_close(self, exit_code: int, exception: Exception | None) -> None:


plugin/core/sessions.py
```diff
@@ -2111,6 +2111,7 @@ def shutdown_session_view_async(self, session_view: SessionViewProtocol) -> None
         session_view.shutdown_async()
 
     def _handle_shutdown_result(self, _: Any) -> None:
+        self._progress.clear()
```
rchl, line 2114: We clear a bunch of Session attributes in `end_async` so it's maybe a bit unconventional to put it here?

rchl: Also, having to do this might indicate that there is some bigger issue that prevents Session from being GCed. I don't think that those progress dicts would do that themselves.

jwortmann: I initially had it in `end_async`, but then I realized that a server might still send the "end" progress reports when it receives the shutdown request. That's why I think we should wait for the response first.

rchl: It looks like the `request_handler` / `notification_handler` decorators keep the Session alive.

jwortmann: Not sure if the version in the bug report #2798 is accurate, but if they are on the latest release version, it should not have these handler decorators yet.

Anyway, I think we should still clean up progress explicitly, because otherwise it might happen again for some other reason in the future.

I would even say having that erase_status in `__del__` is really bad practice, because the garbage collector might run only sporadic, or only when there is high memory usage for example. According to an answer on stackoverflow, GC runs immediately (https://stackoverflow.com/a/1481512), but I wouldn't rely on that and perhaps the Python devs decide to change that behavior in the future. So I think I had put that status cleanup into a separate method, and then call that method explicitly, if I had written that code. But I don't want to do such refactoring here.

rchl: I agree we should not rely on GC, especially for something that is reflected in the UI. I have fixed issues like that before by explicitly removing SessionView, I think.

EdsterG: > Not sure if the version in the bug report https://github.com/sublimelsp/LSP/issues/2798 is accurate, but if they are on the latest release version, it should have these handler decorators yet.

Doubled checked the reported versions and they are correct.

rchl: Identified two reasons for GC not working in #2801

PR #2801, opened 2026-03-09
Fix issue with Session not being garbage collected

rchl:
Sessions were not being garbage collected for two reasons:
- In connection with `request_handler` / `notification_handler` decorators, we've stored reference to methods on `self`. Since methods have hard reference to `self` that created a cycle. Store only mapping from handler name to `m_*` handler and use that for calling the right function when `m_*` method is called.
- The `get_diagnostics_identifiers` function uses `lru_cache` decorator which stores reference to `Session`. We need to clear the cache when session is destroyed. ~~This is not a proper fix since it clears the whole cache instead of only caches related to ended session but I don't have time now to investigate it further.~~ Addressed by switching to manual caching.

For testing one can add

```py
    def __del__(self):
        print('DEL Session')
```

in `Session` to verify that session is destroyed on last file in the window being closed.

### DIFF
--- a/plugin/diagnostics.py
+++ b/plugin/diagnostics.py
@@ -48,7 +48,7 @@ class DiagnosticsStorage:
         self._identifiers_cache[view_id] = identifiers
         return identifiers
 
-    def clear_identifier_cache_for_view(self, view: sublime.View) -> None:
+    def clear_identifiers_cache_for_view(self, view: sublime.View) -> None:
         self._identifiers_cache.pop(view.id(), None)
 
     @property
--- a/plugin/documents.py
+++ b/plugin/documents.py
@@ -253,7 +253,7 @@ class DocumentSyncListener(sublime_plugin.ViewEventListener, AbstractViewListene
         self._cleanup()
         self._setup()
         for session in self.sessions_async():
-            session.diagnostics.clear_identifier_cache_for_view(self.view)
+            session.diagnostics.clear_identifiers_cache_for_view(self.view)
         # But this has to run on the async thread again
         sublime.set_timeout_async(self.on_activated_async)
 


rchl:
Switched to manual caching for diagnostic identifiers.

plugin/documents.py
```diff
@@ -253,7 +252,6 @@ def _reset(self) -> None:
         # Have to do this on the main thread, since __init__ and __del__ are invoked on the main thread too
         self._cleanup()
         self._setup()
-        get_diagnostics_identifiers.cache_clear()
```
jwortmann, line 256: `_reset` is called when either the syntax or the URI of the view changes:
https://github.com/sublimelsp/LSP/blob/e74ce6bb5ba2fa1e0ac2032625c8edefc8dd5c19/plugin/documents.py#L1128-L1129

In that case the DocumentSelector might not match anymore for the diagnostic provider(s). That is the reason why we need to clear the cache here.

For the manual cache, you could also just pop the value for this particular view from the cache instead. That would even be more efficient than clearing the entire cache.

plugin/diagnostics.py
```diff
@@ -46,12 +35,16 @@ def __init__(self) -> None:
...
-            diagnostic_options.get('identifier') for diagnostic_options in self._providers.values()
-            if DocumentSelector_(diagnostic_options.get('documentSelector') or []).matches(view)
-        )
+        view_id = view.id()
+        if view_id not in self._identifiers_cache:
+            self._identifiers_cache[view_id] = set(
+                diagnostic_options.get('identifier') for diagnostic_options in self._providers.values()
+                if DocumentSelector_(diagnostic_options.get('documentSelector') or []).matches(view)
+            )
+        return self._identifiers_cache[view_id]
```
jwortmann: I think writing this in a slightly different way could make it a little bit more efficient by using only a single dict-access:

```suggestion
        if (identifiers := self._identifiers_cache.get(view_id)) is not None:
            return identifiers
        identifiers = set(
            diagnostic_options.get('identifier') for diagnostic_options in self._providers.values()
            if DocumentSelector_(diagnostic_options.get('documentSelector') or []).matches(view)
        )
        self._identifiers_cache[view_id] = identifiers
        return identifiers
```

plugin/diagnostics.py
```diff
@@ -46,12 +35,21 @@ def __init__(self) -> None:
...
+        if (identifiers := self._identifiers_cache.get(view_id)) is not None:
+            return identifiers
+        identifiers = set(
             diagnostic_options.get('identifier') for diagnostic_options in self._providers.values()
             if DocumentSelector_(diagnostic_options.get('documentSelector') or []).matches(view)
         )
+        self._identifiers_cache[view_id] = identifiers
+        return identifiers
+
+    def clear_identifier_cache_for_view(self, view: sublime.View) -> None:
```
jwortmann: Could maybe use `identifiers` instead of `identifier` in the name for consistency.

PR #2807, opened 2026-03-13
Use scheme color for lightbulb icon in popup

jwortmann:
Previously I used the orange color from the Mariana color scheme for the lightbulb icon for code actions in the hover popup. This means that the color isn't accurate for other color schemes and in that case doesn't match the color from the lightbulb gutter icon.

We can actually use a mdpopups function to dynamically tint the icon, based on the color scheme. This requires the newest version of mdpopups (v4.3.5 for Py3.8 or v5.1.1 for Py3.13). You can run *Package Control: Satisfy Libraries* to force an update if you still have a previous version.

---

Example with Monokai:

Before:

before

After:

after

---

If we decide to place the code action directly into the diagnostic box (see https://github.com/sublimelsp/LSP/pull/2788#discussion_r2885770451, perhaps I will look into that later) we could probably remove the lightbulb icon in the hover popup. But for now we could fix the icon color.


plugin/hover.py
```diff
@@ -96,10 +96,12 @@ def code_actions_content(actions_by_config: list[CodeActionsByConfigName], light
             text = actions[0].get('title', 'code action')
         href = "{}:{}".format('code-actions', config_name)
         link = make_link(href, text)
-        lightbulb_html = '<span class="lightbulb"><img src="res://Packages/LSP/icons/lightbulb_colored.png"></span>' \
-            if lightbulb else ''
         formatted.append(
-            f'<div class="actions">{lightbulb_html}{link} <span class="color-muted">{config_name}</span></div>')
+            f'''
+            <div class="code-actions">
```
rchl, line 101: It's a single code action so `code-action` class might be more appropriate.

jwortmann: In the example from the screenshot yes, but if there are multiple actions it shows as "choose (3 available)" on a single line, which triggers the quick panel in that case.

jwortmann: If you agree to refactor displaying the code actions as part of the diagnostic boxes directly (like the relatedInformation), then I think we could show all available code actions separately and reconsider the CSS class name, if it's still needed then.

rchl: Have not given that change a lot of thought.
I wonder if it would then look inconsistent when code action doesn't have related diagnostic (in case of manually triggering "source actions" for example).

jwortmann: We only show "quickfix" code actions in the popup, and I think in theory all quickfix actions should have a corresponding diagnostic. 

When manually triggered from the context menu, they are shown in the quick panel.

PR #2809, opened 2026-03-16
Refactor diagnostic css classes

rchl:
Address https://github.com/sublimelsp/LSP/pull/2788#discussion_r2941235440. I don't want to make that other PR too busy so making this change separately.

- Change names of diagnostics CSS classes for consistency.
- Replace `DIAGNOSTIC_SEVERITY` (list of tuples) with `DIAGNOSTIC_STYLES` (dict keyed by `DiagnosticSeverity`). Accessing properties by index was not very readable and same for having to subtract 1 from severity when indexing `DIAGNOSTIC_SEVERITY`.

### DIFF
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -324,8 +324,6 @@ class SessionView:
         key = self.diagnostics_key(severity, multiline)
         tags = {tag: TagData(f'{key}_tags_{tag}') for tag in DIAGNOSTIC_TAG_SCOPES}
         data = self._session_buffer.diagnostics_data_per_severity.get((severity, multiline))
-        region_scope = DIAGNOSTIC_STYLES[severity].region_scope
-        icon = diagnostic_icon(severity)
         if data and severity <= max_severity_level:
             non_tag_regions = data.regions
             for tag, regions in data.regions_with_tag.items():
@@ -336,6 +334,8 @@ class SessionView:
                     tags[tag].scope = tag_scope
                 else:
                     non_tag_regions.extend(regions)
+            region_scope = DIAGNOSTIC_STYLES[severity].region_scope
+            icon = diagnostic_icon(severity)
             self.view.add_regions(f"{key}_icon", non_tag_regions, region_scope, icon, DIAGNOSTIC_ICON_FLAGS)
             self.view.add_regions(f"{key}_underline", non_tag_regions, region_scope, "", flags)
         else:


plugin/core/views.py
```diff
@@ -76,34 +77,78 @@
...
+            return "" if self.severity == DiagnosticSeverity.Hint else userprefs().diagnostics_gutter_marker
 
 
 class InvalidUriSchemeException(Exception):
     def __init__(self, uri: str) -> None:
-        self.uri = uri
-
-    def __str__(self) -> str:
-        return f"invalid URI scheme: {self.uri}"
+        super().__init__(f"invalid URI scheme: {uri}")
```
rchl, line 139: Drive-by change. Was unnecessarily complicated.

plugin/core/views.py
```diff
@@ -76,34 +77,78 @@
...
+    @property
+    def scope(self) -> str:
+        return DIAGNOSTIC_STYLES[self.severity].region_scope
+
+    @property
+    def icon(self) -> str:
+        if userprefs().diagnostics_gutter_marker == "sign":
+            return DIAGNOSTIC_STYLES[self.severity].icon_resource
+        else:
+            return "" if self.severity == DiagnosticSeverity.Hint else userprefs().diagnostics_gutter_marker
```
rchl: I'm not sure if we really need to expose those on `DiagnosticSeverityData`.

Or at least `scope` we could easily read from `DIAGNOSTIC_STYLES`.

`icon` needs to be re-evaluated currently so it needs to exist somewhere.

plugin/session_view.py
```diff
@@ -306,9 +307,10 @@ def _redraw_diagnostics_async(self) -> None:
         flags = userprefs().diagnostics_highlight_style_flags()  # for single lines
         multiline_flags = None if userprefs().show_multiline_diagnostics_highlights else sublime.RegionFlags.DRAW_NO_FILL | sublime.RegionFlags.DRAW_NO_OUTLINE | sublime.RegionFlags.NO_UNDO  # noqa: E501
         level = userprefs().show_diagnostics_severity_level
-        for sev in reversed(range(1, len(DIAGNOSTIC_SEVERITY) + 1)):
-            self._draw_diagnostics(sev, level, flags[sev - 1] or DIAGNOSTIC_SEVERITY[sev - 1][4], multiline=False)
-            self._draw_diagnostics(sev, level, multiline_flags or DIAGNOSTIC_SEVERITY[sev - 1][5], multiline=True)
+        for sev in reversed(DIAGNOSTIC_STYLES.keys()):
+            style = DIAGNOSTIC_STYLES[sev]
```
jwortmann: Can be simplified:

```suggestion
        for sev, style in reversed(DIAGNOSTIC_STYLES.items()):
```

plugin/core/views.py
```diff
@@ -698,10 +743,8 @@ def document_color_params(view: sublime.View) -> DocumentColorParams:
     return {"textDocument": text_document_identifier(view)}
 
 
-def format_severity(severity: int) -> str:
-    if 1 <= severity <= len(DIAGNOSTIC_SEVERITY):
-        return DIAGNOSTIC_SEVERITY[severity - 1][0]
-    return "???"
+def format_severity(severity: DiagnosticSeverity) -> str:
+    return DIAGNOSTIC_STYLES[severity].kind if severity in DIAGNOSTIC_STYLES else "???"
```
jwortmann: If `severity` is typed as `DiagnosticSeverity`, then we shouldn't need the `in` check.
I think this is pretty old code with the "???" and that fallback should actually be unnecessary anyway.

Also this function seems to be used only in a single place a few lines below, so you could just inline that below as
```python
DIAGNOSTIC_STYLES[diagnostic_severity(diagnostic)].kind
```

plugin/core/views.py
rchl: ab2ad818e665b87129c9b5956ada618f493ecf3a and 9c9974ab2d5e949045d9b6d4251d173a3ab4daab

jwortmann (approved):

PR #2818, opened 2026-03-20
Show code actions with diagnostics in hover popup

jwortmann:
This is what I had in mind. Let me know what you think.

Example:

hover

If there are multiple code actions, all of them will be shown on individual lines. But only "quickfix" code actions are requested now for the hover popup. For "refactor" actions you must use the context menu or main menu.

If there are multiple diagnostics at the hover point, only code actions which can safely be matched to a diagnostic (via `diagnostics` property in the code action) are shown. If a code action applies to multiple diagnostics, then the same code action is shown in each of the diagnostics that it resolves.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeAction
```ts
	/**
	 * The diagnostics that this code action resolves.
	 */
	diagnostics?: Diagnostic[];
```

---

The code action (or command) is now encoded into the link URI directly (using a special `code-action` URI scheme), which makes additional logic to match a selected index to available code actions unnecessary.

---

I'm a bit unsure about the lightbulb icon in the popup, for now I have used it only for code actions that have the `isPreferred` property set. We can reconsider that and could also remove it entirely.

---

There was some very basic caching for code actions in the hover popup in the previous implementation (in CodeActionsManager), and that isn't applied anymore. Instead, if there are any diagnostics at the hover point, we now always make a new code action request (in addition to the hover request). I could probably add back the caching, but I'm not sure if it is really needed for the manual mouse hover, and also we currently don't cache the hover response for the point.

---

I have only tested with the basedpyright server, so maybe some more testing could be useful.

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -930,7 +930,7 @@ def format_diagnostic_for_html(
         content += '<hr>' + _html_element("div", info, escape=False)
     for code_action in sorted(code_actions, key=lambda a: a.get('isPreferred', False), reverse=True):
         icon = lightbulb_html(lightbulb_color, code_action.get('isPreferred', False))
-        code_action_uri = encode_code_action_uri(code_action, config.name)
+        code_action_uri = encode_code_action_uri(config.name, code_action)
         content += '<hr>' + icon + make_link(code_action_uri, code_action['title'], tooltip='Run Code Action')
     severity_class = DIAGNOSTIC_STYLES[diagnostic_severity(diagnostic)].css_class
     return html_wrapper(content, class_name=severity_class)
--- a/plugin/core/url.py
+++ b/plugin/core/url.py
@@ -103,7 +103,7 @@ def _uppercase_driveletter(match: Any) -> str:
     return f"{match.group(1).upper()}:"
 
 
-def encode_code_action_uri(action: Command | CodeAction, session_name: str) -> URI:
+def encode_code_action_uri(session_name: str, action: Command | CodeAction) -> URI:
     return f'{CODE_ACTION_SCHEME}:{session_name}/{urlsafe_b64encode(json.dumps(action).encode()).decode()}'
 
 


plugin/hover.py
```diff
@@ -305,9 +295,6 @@ def _show_hover(self, listener: AbstractViewListener, point: int, only_diagnosti
         elif link_content:
             contents += html_wrapper(link_content)
 
-        _test_contents.clear()
-        _test_contents.append(contents)  # for testing only
```
jwortmann, line 309: I think we should design the test differently instead of needing to keep a global variable that is only used for testing...

rchl:
> I'm a bit unsure about the lightbulb icon in the popup, for now I have used it only for code actions that have the `isPreferred` property set. We can reconsider that and could also remove it entirely.

I think it looks weird to have it only for some items. Especially since the icon doesn't really match the meaning (not sure why bulb would represent "preferred"). Or does it match behavior in vscode for example?

Screenshot 2026-03-21 at 14 04 47

I guess the options are to either remove for all or add for all? Maybe having it for all would look too busy?

rchl:
It's more of a style refactor issue but I wonder whether we should only add the horizontal separator between adjacent diagnostics of same severity - in the screenshot between two "blue" diagnostics".
The horizontal lines can look busy and there seems to be missing separation between same-severity diagnostics

Screenshot 2026-03-21 at 14 13 40

jwortmann:
> Especially since the icon doesn't really match the meaning (not sure why bulb would represent "preferred"). Or does it match behavior in vscode for example?

I think in VSCode the lightbulb just means "quick fix available". But they don't show it in the popup or gutter, but instead as a floating icon that can be clicked to open a popup menu: https://code.visualstudio.com/docs/typescript/typescript-tutorial#_quick-fixes

What kind of icon would you associate with "preferred"? The set of available icons to choose from is at https://primer.style/octicons/ (the error/warning/info/lighbulb icons are all from this icon set so that they look consistently).

---

I've pushed a commit https://github.com/sublimelsp/LSP/pull/2818/changes/3cb3bbadfeff30f69e726612025eb4e2a5766547 to show the icon to the right side of the link, so that all code action links are left-aligned. Maybe that looks better? We could also remove the lightbulb entirely, but how do we show the `isPreferred` information then?

If you hover with the mouse over the lighbulb in the popup, it shows a "Preferred Quick Fix" tooltip. So I think that is an intuitive way to learn what it means (although that is not fully consistent with the lightbulb icon in the gutter - so yes we can try out different things and discuss what would be the best solution).

jwortmann:
> It's more of a style refactor issue but I wonder whether we should only add the horizontal separator between adjacent diagnostics of same severity - in the screenshot between two "blue" diagnostics".
> The horizontal lines can look busy and there seems to be missing separation between same-severity diagnostics

I think you could add a 1 pixel `border-bottom` to the severity classes at https://github.com/sublimelsp/LSP/blob/32c3f147a1b16957f3c969df8931809ce216610a/popups.css#L79-L84

and then test how it looks. Otherwise adding `<hr>` between diagnostics of the same severity could be a bit complicated for the implementation, because at the moment I just sort by severity and then do a `''.join()`.

Btw, I have also tested to remove the `<hr>` between multiple code actions for a diagnostic (like in your first screenshot), but then the links look too cramped together in my opinion and you could easily misclick a code action link if they don't have a visible separator line.

rchl:
> What kind of icon would you associate with "preferred"? The set of available icons to choose from is at https://primer.style/octicons/ (the error/warning/info/lighbulb icons are all from this icon set so that they look consistently).

Could we create our own variation of the bulb icon - for example a bulb icon with a small star in the corner?

I think it would be good to have icon for all code actions but then having completely different icon for one of them would look out of place IMO. So that's where my idea comes in.

EDIT: Or a bulb icon with filled in bulb (turned on).

rchl:
> I think you could add a 1 pixel `border-bottom` to the severity classes at and then test how it looks. Otherwise adding `<hr>` between diagnostics of the same severity could be a bit complicated for the implementation, because at the moment I just sort by severity and then do a `''.join()`.

I think the first option would not look that good:

Screenshot 2026-03-21 at 15 29 05

Here is an example with no separators at all but with more prominent spacing that kinda makes it more clear where the separation is:

Screenshot 2026-03-21 at 15 34 51

(we don't have to address it in this PR)

EDIT: Quickly hacked (paddings are not correct) variant with separator but I've just noticed that those last two are actually not the same severity. One is hint and one is information. So logic based on severity wouldn't really help here:

Screenshot 2026-03-21 at 15 46 22

jwortmann:
> Could we create our own variation of the bulb icon - for example a bulb icon with a small star in the corner?

I'll try that out later perhaps.

> I think the first option would not look that good:

Ok could you try with ~~`padding-bottom: 1px`~~ `margin-bottom: 1px` istead? Maybe that looks more subtle. I prefer to keep the `<hr>` between related information and between code actions.

rchl:
> > I think the first option would not look that good:
> 
> Ok could you try with ~`padding-bottom: 1px`~ `margin-bottom: 1px` istead? Maybe that looks more subtle. I prefer to keep the `<hr>` between related information and between code actions.

To me this looks weird. I prefer our flat look without margin.

Screenshot 2026-03-21 at 17 14 06

jwortmann:
I have added a modified version of the lightbulb icon for preferred code actions, and the other code actions use the regular lightbulb now. Could you test that with a server that sets the `isPreferred` property, and tell whether it looks good now?

lightbulb_hd

rchl:
> I have added a modified version of the lightbulb icon for preferred code actions, and the other code actions use the regular lightbulb now. Could you test that with a server that sets the `isPreferred` property, and tell whether it looks good now?

Screenshot 2026-03-21 at 19 51 29

If I'd to comment I'd say that it's a bit hard to tell what it means. It covers most of the bulb and of course the icon is quite small so it's probably hard to make it better.

What about filling the bulb itself with the star icon? :)

jwortmann:
On Windows the icon is slightly bigger (13 pixel) than in your screenshot (11 pixel), so it looks a bit better here. I think I would prefer the current version with the star in the corner.

hover1

hover2

---

It seems that the image is always positioned at the baseline of the font (see the following screenshot). I wonder whether there is some CSS trick to move the image a few more pixel to the bottom, below the font baseline. Then we could increase the image size a bit:
https://github.com/sublimelsp/LSP/blob/bad408925ec0ad5ab066d4b149e2002796196f3b/popups.css#L107-L110

But that is probably not possible in minihtml, I guess?

hover3

rchl:
> It seems that the image is always positioned at the baseline of the font (see the following screenshot). I wonder whether there is some CSS trick to move the image a few more pixel to the bottom, below the font baseline. Then we could increase the image size a bit:

I've pushed something to that effect with 1.1rem size. 1.0 would potentially work too.

Not sure how portable that is in different font sizes...

Also, I quite like the icon with star inside the bulb, to be honest.

## 1rem

Screenshot 2026-03-22 at 11 20 34

## 1.1rem

Screenshot 2026-03-22 at 11 20 51

plugin/core/views.py
```diff
@@ -870,28 +879,71 @@ def _html_element(tag: str, content: str, *, class_name: str | None = None, esca
...
         'lsp_copy_text', {'text': copy_text}
     )}'>⧉</a>"""
     if related_infos := diagnostic.get("relatedInformation"):
         info = "<br>".join(_format_diagnostic_related_info(config, info, base_dir) for info in related_infos)
-        html += '<hr>' + _html_element("div", info, escape=False)
+        content += '<hr>' + _html_element("div", info, escape=False)
+    for code_action in sorted(code_actions, key=lambda a: a.get('isPreferred', False), reverse=True):
+        icon = lightbulb_html(lightbulb_color, code_action.get('isPreferred', False))
+        code_action_uri = encode_code_action_uri(code_action, config.name)
+        content += '<hr>' + icon + make_link(code_action_uri, code_action['title'], tooltip='Run Code Action')
```
rchl, line 934: Would find this less error prone than adding a trailing space in icon's html:

```suggestion
        content += f'<hr>{icon} ' + make_link(code_action_uri, code_action['title'], tooltip='Run Code Action')
```

rchl:
Two cases - hovering diagnostics and hovering gutter icon:

Screenshot 2026-03-22 at 11 45 34

Screenshot 2026-03-22 at 11 45 39

---

- Is it expected that code actions don't show up in latter?
- Why is the related information link rendered differently in both cases?

plugin/hover.py
```diff
@@ -336,34 +325,24 @@ def _on_images_resolved(self, original_contents: str, contents: str) -> None:
...
+        scheme = parse_uri(href)[0]
+        if scheme == 'subl':
             pass
-        elif href.startswith("file:"):
+        elif scheme == 'file':
             if window := self.view.window():
                 open_file_uri(window, href)
-        elif href.startswith('code-actions:'):
-            self.view.run_command("lsp_selection_set", {"regions": [(point, point)]})
-            _, config_name = href.split(":")
```
rchl, line 347: I guess we don't set selection anymore. Not useful?

jwortmann:
I've set the size to 1rem, because 1.1rem looks too huge on Windows. I initially used 0.9rem because that would match the size from the gutter icon exactly (on Windows). But 1rem looks good on Windows too.

---

> * Is it expected that code actions don't show up in latter?

It still follows the `"show_code_actions"` setting, like before. If we want to change that behavior and always include code actions when hovering the gutter icon, we should probably convert that into a boolean setting `"show_code_action_annotations": true/false`.

We could then also consider to include "refactor" code actions in the annotations. Also maybe instead of "choose (3 available)" we could show all code actions directly in the annotation, because ST only shows the first line of the annotation anyway (with additional linebreak icon if there are multiple lines), and if there are more code actions they would get revealed on mouse hover. But maybe that would be for another PR.

Btw, if there are multiple diagnostics on the line from a single server, then the code actions are only shown when hovering the gutter icon if the code actions include the `diagnostics` property.

---

> * Why is the related information link rendered differently in both cases?

Should be fixed now.

plugin/core/views.py
jwortmann: I've replaced the space by CSS padding now. Looks better with a bit more padding for the bigger icon with the star.

plugin/hover.py
jwortmann: I didn't see the purpose of that, and since there was no source comment explaing why it was needed, I removed it. It kind of felt like a bug to me that the selection would change when running a code action.

plugin/hover.py
```diff
@@ -336,34 +325,24 @@ def _on_images_resolved(self, original_contents: str, contents: str) -> None:
...
-                        items,
-                        lambda i: self.handle_code_action_select(config_name, actions, i),
-                        selected_index=selected_index,
-                        placeholder="Code actions")
-            else:
-                self.handle_code_action_select(config_name, actions, 0)
+        elif scheme == CODE_ACTION_SCHEME:
+            session_name, action = decode_code_action_uri(href)
+            if session := self.session_by_name(session_name):
+                sublime.set_timeout_async(lambda: session.run_code_action_async(action, progress=True, view=self.view))
```
rchl, line 317: The popup should close on running a code action.

One reason is that it just feel right and another is that popup doesn't update after running so we'd be showing stale code action (for example after applying ignore comment).

plugin/documents.py
```diff
@@ -569,37 +568,31 @@ def on_hover(self, point: int, hover_zone: int) -> None:
...
-                    content += format_diagnostic_for_html(config, diagnostic)
-                content += '</div>'
-        if content:
+        if userprefs().diagnostics_gutter_marker:
+            diagnostics = self.get_diagnostics_async(self.view.line(point), userprefs().show_diagnostics_severity_level)
+            code_actions = dict(self._code_actions_for_selection) \
+                if self._lightbulb_line == self.view.rowcol(point)[0] else {}
+            base_dir = self._manager.get_project_path(filename) \
+                if self._manager and (filename := self.view.file_name()) else None
+            content = format_diagnostics_for_html(diagnostics, code_actions, self.lightbulb_color, base_dir)
```
rchl: `self._code_actions_for_selection` contains diagnostics from the line where selection is which is not necessarily the line corresponding to hovered gutter position. This means that we only show code actions if selection is on the same line.

I think we should request ~diagnostics~ code actions for the whole line for this purpose.

plugin/documents.py
```diff
@@ -569,37 +568,31 @@ def on_hover(self, point: int, hover_zone: int) -> None:
...
             show_lsp_popup(
                 self.view,
                 content,
                 flags=sublime.PopupFlags.HIDE_ON_MOUSE_MOVE_AWAY,
                 location=point,
-                on_navigate=lambda href: self._on_navigate(href, point))
+                on_navigate=self._on_navigate)
```
rchl: Looks like we are triggering the popup even if there are no diagnostics to show.
I'm surprised we don't see any tiny popup content but it's better to explicitly check for that anyway.

plugin/documents.py
```diff
@@ -569,37 +568,31 @@ def on_hover(self, point: int, hover_zone: int) -> None:
...
+    def _on_navigate(self, href: str) -> None:
+        scheme = parse_uri(href)[0]
+        if scheme == 'file':
+            if window := self.view.window():
+                open_file_uri(window, href)
+        elif scheme == CODE_ACTION_SCHEME:
+            session_name, action = decode_code_action_uri(href)
+            if session := self.session_by_name(session_name):
+                sublime.set_timeout_async(lambda: session.run_code_action_async(action, progress=True, view=self.view))
+        elif scheme.lower() in ("http", "https") or scheme == '' and href.startswith('www.'):
```
rchl, line 616: Should close the popup on applying code action.

plugin/hover.py
```diff
@@ -216,17 +198,37 @@ def _on_resolved_link(
     def _on_all_document_links_resolved(
         self, listener: AbstractViewListener, point: int, links: list[DocumentLink | None]
     ) -> None:
-        self._document_links = list(filter(None, links))
-        self.show_hover(listener, point, only_diagnostics=False)
+        if document_links := list(filter(None, links)):
+            self._document_links = document_links
+            self.show_hover(listener, point, only_diagnostics=False)
```
rchl, line 200: We should always overwrite `self._document_links`. Otherwise in theory we could use stale result if result with resolved links is followed by one without resolved link (however unlikely that scenario might be).

EDIT: Never mind, I see that we re-initialize in `run`.

plugin/hover.py
```diff
@@ -216,17 +198,37 @@ def _on_resolved_link(
...
+            # "Quickfix" actions were requested.
+            code_actions = [action for action in response if not action.get('disabled', False)]
+        else:
+            # If there are multiple diagnostics for the hover region, we can only use those code actions which include
+            # the "diagnostics" property, because we need to match each code action to its corresponding diagnostics.
+            code_actions = [
+                action for action in response
+                if is_code_action_with_diagnostics(action) and not action.get('disabled', False)
+            ]
+        return config_name, code_actions  # pyright: ignore[reportReturnType]
```
rchl: No need to ignore type - better to add explicit `list[Command | CodeAction]` type inside the first `if`.

plugin/hover.py
```diff
@@ -216,17 +198,37 @@ def _on_resolved_link(
...
         listener: AbstractViewListener,
         point: int,
-        responses: list[CodeActionsByConfigName]
+        responses: list[tuple[str, list[Command | CodeAction]]]
     ) -> None:
-        self._actions_by_config = responses
-        self.show_hover(listener, point, only_diagnostics=False)
+        if actions := {config_name: code_actions for config_name, code_actions in responses if code_actions}:
+            self._actions_by_config = actions
+            self.show_hover(listener, point, only_diagnostics=False)
```
rchl, line 210: Here also would be better to always overwrite instance properties to avoid any potential surprises. I'm not bothering to think if this is a realistic problem - I just say that based on the fact that TextCommands are not re-initialized for each invocation so always better to be safe.

EDIT: Never mind, I see that we re-initialize in `run`.

plugin/documents.py
jwortmann: Good catch. Actually I do see a popup, but it is very tiny (only 1px). I will fix it. I think I accidentally removed such a check when I rewrote and significantly simplified this method.

plugin/documents.py
jwortmann: https://github.com/sublimelsp/LSP/pull/2818/changes/21804a4f4ed2575c291eab76b6749a2f4d3df49e

plugin/hover.py
jwortmann: https://github.com/sublimelsp/LSP/pull/2818/changes/21804a4f4ed2575c291eab76b6749a2f4d3df49e

plugin/documents.py
jwortmann: Fixed in https://github.com/sublimelsp/LSP/pull/2818/changes/000066908f4c33528349b5c116d0f2e20e2a9a05

plugin/session_buffer.py
```diff
@@ -958,6 +964,30 @@ def remove_inlay_hint_phantom(self, phantom_uuid: str) -> None:
...
+            'triggerKind': trigger_kind
+        }
+        if kinds:
+            context['only'] = kinds
+        params: CodeActionParams = {
+            'textDocument': text_document_identifier(view),
+            'range': region_to_range(view, region),
+            'context': context
+        }
+        request: Request[CodeActionParams, list[Command | CodeAction] | None] = Request.codeAction(params, view)
```
rchl: No need to type the variable

```suggestion
        request = Request.codeAction(params, view)
```

Also might consider skipping the variable though that's a matter of preference I guess.

plugin/documents.py
jwortmann: > `self._code_actions_for_selection` contains diagnostics from the line where selection is which is not necessarily the line corresponding to hovered gutter position. This means that we only show code actions if selection is on the same line.

Absolutely. This preserves the current behavior on main.

> I think we should request diagnostics for the whole line for this purpose.

We could, but then we should reconsider/update the `show_code_actions` setting to reflect that (see https://github.com/sublimelsp/LSP/pull/2818#issuecomment-4106307097).

plugin/hover.py
jwortmann: https://github.com/sublimelsp/LSP/pull/2818/changes/28c33691a5eca51c73d335323f4b6feda1ff1bfa

plugin/documents.py
rchl: > We could, but then we should reconsider/update the `show_code_actions` setting to reflect that (see [#2818 (comment)](https://github.com/sublimelsp/LSP/pull/2818#issuecomment-4106307097)).

I'm not sure I see the connection with `show_code_actions`... Why couldn't we just request code actions for the whole line and keep the setting as is?

That said, yes, it's an old issue and you can choose not to address it here.

(I of course meant "code actions", not diagnostics (corrected my original comment))

plugin/hover.py
```diff
@@ -216,17 +198,37 @@ def _on_resolved_link(
...
+            # If there is only a single diagnostic from this server, all enabled code actions can be shown, because only
+            # "Quickfix" actions were requested.
+            code_actions = [action for action in response if not action.get('disabled', False)]
+        else:
+            # If there are multiple diagnostics for the hover region, we can only use those code actions which include
+            # the "diagnostics" property, because we need to match each code action to its corresponding diagnostics.
+            code_actions: list[Command | CodeAction] = [
+                action for action in response
+                if is_code_action_with_diagnostics(action) and not action.get('disabled', False)
+            ]
```
rchl: This does make pyright happy but it's really the first assignment that makes it unsure about the type so it would be more natural to add it there.

```suggestion
        if isinstance(response, Error) or not response:
            code_actions: list[Command | CodeAction] = []
        elif diagnostics_count == 1:
            # If there is only a single diagnostic from this server, all enabled code actions can be shown, because only
            # "Quickfix" actions were requested.
            code_actions = [action for action in response if not action.get('disabled', False)]
        else:
            # If there are multiple diagnostics for the hover region, we can only use those code actions which include
            # the "diagnostics" property, because we need to match each code action to its corresponding diagnostics.
            code_actions = [
                action for action in response
                if is_code_action_with_diagnostics(action) and not action.get('disabled', False)
            ]
```

plugin/documents.py
jwortmann: I guess we could always request code actions for the entire line when hovering a gutter icon. I'll think about it.

I think the current behavior is partly for historic reasons. First, we only had the code action annotations on the righthand side, and the lightbulb icon, but no popup when hovering over the lightbulb. Then I added the popup with code actions when hovering over the lightbulb. I'm not quite sure, but I believe the diagnostics when hovering a diagnostic gutter icon were added even later into that popup. But still the code actions were only in the popup for the line with the lightbulb icon, which kind of makes sense, I would say, because in theory code actions don't need to be coupled to diagnostics, and previously we displayed them independently of diagnostics.

plugin/hover.py
jwortmann: Interesting, I didn't know that would work. I thought the problem would be that `is_code_action_with_diagnostics` makes the type in that case a `list[CodeAction]`, and that is not compatible with `list[Command | CodeAction]` for some reason. But I guess we can use your suggestion if it works.

plugin/documents.py
jwortmann: https://github.com/sublimelsp/LSP/pull/2818/changes/9fa1c0e2b40bba8af686bbbe34f0830d899bb58e

Issue #2815, opened 2026-03-20
OnTypeFormatting misfires in LSP-yaml


rchl:
1. Install LSP-yaml
2. Create `docker-compose.yaml`:
```yaml
services:
  foo:
    environment:
      - foo=bar
```
3. Place the cursor somewhere in the middle of `foo=bar` and run "duplicate line"

https://github.com/user-attachments/assets/0b724d35-62b0-4fe6-94c2-0d9a135da40d

## Request

DidChange

```json
{
  "contentChanges": [
    {
      "range": {
        "end": {
          "character": 0,
          "line": 4
        },
        "start": {
          "character": 0,
          "line": 4
        }
      },
      "rangeLength": 0,
      "text": "      - foo=bar\n"
    }
  ],
  "textDocument": {
    "uri": "file:///Users/rafal/Downloads/docker-compose.yaml",
    "version": 71
  }
}
```

OnTypeFormatting

```json
{
  "ch": "\n",
  "options": {
    "insertFinalNewline": true,
    "insertSpaces": true,
    "tabSize": 2,
    "trimFinalNewlines": true,
    "trimTrailingWhitespace": false
  },
  "position": {
    "character": 11,
    "line": 4
  },
  "textDocument": {
    "uri": "file:///Users/rafal/Downloads/docker-compose.yaml"
  }
}
```

## Response

```json
[
  {
    "newText": "- ",
    "range": {
      "end": {
        "character": 11,
        "line": 4
      },
      "start": {
        "character": 11,
        "line": 4
      }
    }
  }
]
```

I haven't investigated yet but:
1. the position we are sending is likely not correct (the position of the cursor is not relevant in this scenario). But I doubt that a correct position will help here - server will likely still insert another `-`.
2. the request should likely not even trigger in the first place in this scenario

rchl:
"Funny" thing is that it fails to work when it should (when pressing enter at the end of the line) because then newline is not the last character due to auto-indent...

PR #2819, opened 2026-03-20
Trigger OnTypeFormatting on newline insertion

rchl:
Detect insertion of a newline in change action logic.

When enter is pressed, this typically triggers `insert` command with `{ characters: '\n' }` arguments and results in ST inserting a newline with trailing spaces (due to auto-indentation logic). So if that case is detected, strip trailing spaces before trying to match the trigger.

This allows us to support automatic `-` insertion in lists in LSP-yaml (test with example in #2815).

Let me know if the logic is getting too crazy.

### DIFF
--- a/plugin/session_buffer.py
+++ b/plugin/session_buffer.py
@@ -765,14 +765,14 @@ class SessionBuffer:
         if not self._on_type_formatting_triggers or \
                 not (self._get_request_flags(view) & RequestFlags.ON_TYPE_FORMATTING):
             return None
-        if action == ChangeEventAction.INSERT_NEWLINE:
-            if '\n' in self._on_type_formatting_triggers and text.rstrip(' ').endswith('\n'):
-                return self._create_on_type_formatting_params_async(view, '\n')
-            return None
         if action == ChangeEventAction.TYPE:
             for trigger in self._on_type_formatting_triggers:
                 if text.endswith(trigger):
                     return self._create_on_type_formatting_params_async(view, trigger)
+            return None
+        if action == ChangeEventAction.INSERT_NEWLINE:
+            if '\n' in self._on_type_formatting_triggers and text.rstrip(' ').endswith('\n'):
+                return self._create_on_type_formatting_params_async(view, '\n')
             return None
         return None
 


plugin/session_buffer.py
```diff
@@ -764,7 +765,8 @@ def _get_on_type_formatting_params_async(
         if selection is None:
             return None
         for trigger in self._on_type_formatting_triggers:
-            if text.endswith(trigger):
+            if (action == ChangeEventAction.TYPE and text.endswith(trigger) or
+                action == ChangeEventAction.INSERT_NEWLINE and text.rstrip(' ').endswith(trigger)):
```
predragnikolic: Can we get away without introducing `INSERT_NEWLINE`?

When we type a new line, the `ChangeEventAction` will be `TYPE`,
the only thing that we need to change in the code is to add `rstrip(' ')`.

I have not tried what I suggested, but it seems like `INSERT_NEWLINE` might not be needed. `INSERT_NEWLINE` (I treat `INSERT_NEWLINE` as a specific subtype of `TYPE`)

rchl: We don't want to strip trailing spaces when action is `TYPE`. Imagine how it would break the case where server has a space trigger.

Or even when it has a different trigger - we would be matching against not the last character but the one before spaces when typing space.

rchl:
@jwortmann @predragnikolic I'd like this integrated since it fixes one of major use cases of OnTypeFormatting.
I know it adds a bit of overhead but mainly only on pressing enter so it's fairly localized.

plugin/session_buffer.py
```diff
@@ -760,22 +760,30 @@ def _update_on_type_formatting_triggers(self) -> None:
             self._on_type_formatting_triggers = ()
 
     def _get_on_type_formatting_params_async(
-        self, view: sublime.View, text: str
+        self, view: sublime.View, action: ChangeEventAction, text: str
     ) -> DocumentOnTypeFormattingParams | None:
-        if not self._on_type_formatting_triggers:
+        if not self._on_type_formatting_triggers or \
+                action not in {ChangeEventAction.TYPE, ChangeEventAction.INSERT_NEWLINE} or \
```
jwortmann: This line should be removed because you do the same tests in the code below already. You should rewrite the first "if" below into 2 nested "if"s then:
```python
if action == ChangeEventAction.INSERT_NEWLINE:
    if ...
```

rchl: I don't understand the second part of your suggestion.

rchl: Oh, I see, to avoid checking `action` twice.

jwortmann:
I have not tried on-type-formatting yet.

But I noticed that some key bindings for newline don't use the `insert` command directly:
* <kbd>Ctrl</kbd>+<kbd>Enter</kbd>: `run_macro_file {"file": "res://Packages/Default/Add Line.sublime-macro"}`
* <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>Enter</kbd>: `run_macro_file {"file": "res://Packages/Default/Add Line Before.sublime-macro"}`
* when caret inside `{}`: `run_macro_file {"file": "res://Packages/Default/Add Line in Braces.sublime-macro"}`

(this is the output from `sublime.log_commands(True)`; I haven't tested if the `on_text_command` API sees the same)

Have you tested if it works in these occasions too? Should it trigger in these cases?

Some servers like Tinymist (I think rust-analyzer too) also use a custom request `experimental/onEnter` for this use case. I don't think that Tinymist supports on-type-formatting, but I wonder whether that could interfere with it. But I guess it should be okay.

rchl:
> Have you tested if it works in these occasions too? Should it trigger in these cases?

I know it won't work even without testing (but I've tested to be sure).

It should ideally trigger but I think it's fine if it does not. I would be more concerned if it triggered when it shouldn't have.

Detecting those cases where macro is run would technically be doable by whitelisting those specific macros but it feels a bit hacky to me maybe. One thing is detecting specific built in commands and another is matching specific paths to macro files (still built-in though).

rchl:
Actually it wouldn't be even that easy (or even feasible) to support some of those since `Add Line Before` for example has this code:

```
[
	{"command": "move_to", "args": {"to": "hardbol"}},
	{"command": "insert", "args": {"characters": "\n"}},
	{"command": "move", "args": {"by": "lines", "forward": false, "ignore_auto_complete": true}},
	{"command": "reindent", "args": {"force_indent": false}}
]
```

which means that the last edit won't contain the newline necessarily so we wouldn't be able to match it.

One could even argue that some of those are not related to typing (which this API is supposed to trigger on only). But don't discuss this with me because I can see how this can be taken in any direction. ;)

PR #2827, opened 2026-03-24
Always request workspace diagnostics if supported by server

jwortmann:
Always request workspace diagnostics, independently of the `"diagnostics_mode"` setting.

If we want to do this change, the code at https://github.com/sublimelsp/LSP/blob/eab1a177c0a4e2d8b9430e61fff5b0f5fb42259f/plugin/session_buffer.py#L272-L275

probably doesn't make sense anymore in that form. But also I think we can't just clear diagnostics unconditionally on closing a file, because that would forfeit the purpose of workspace diagnostics (to have diagnostics for unopened files). So we might need to consider how to handle the problem 2. from https://github.com/sublimelsp/LSP/pull/2727#issuecomment-4119397484. Maybe another check whether the file is outside of the workspace folders would do.


jwortmann:
Pull diagnostics are cleared now for files outside of the workspace folders (or for all files if there are no workspace folders), when the file gets closed. I think that should be the expected behavior.

I noticed that the current code with `self.has_capability("diagnosticProvider")` doesn't work anyway, because that always returns False since the pull diagnostics with multiple providers were fixed (because there is special handling for diagnostic provider registration).

And the `diagnostics_mode` config setting is now basically reverted to the predecessor setting `hide_non_project_diagnostics`. Not sure if we also want to revert back to the old name, or leave it like that. But it's probably rarely used anyway.

docs/src/client_configuration.md
```diff
@@ -33,7 +33,7 @@ Below is an example configuration for the [Phpactor](https://phpactor.readthedoc
...
-| diagnostics_mode | Set to `"workspace"` (default is `"open_files"`) to ignore diagnostics for files that are not within the project (window) folders. If project has no folders then this option has no effect and diagnostics are shown for all files. If the server supports _pull diagnostics_ (`diagnosticProvider`), this setting also controls whether diagnostics are requested only for open files (`"open_files"`), or for all files in the project folders (`"workspace"`). |
+| diagnostics_mode | Set to `"workspace"` (default is `"open_files"`) to ignore diagnostics for files that are not within the project (window) folders. If project has no folders then this option has no effect and diagnostics are shown for all files. |
```
rchl, line 36: I have some issue with this configuration's name and values.

```
"open_files" - All diagnostics reported from the server are shown.
"workspace" - If there are project folders (folders in the side bar),
    diagnostics for files not within those folders are ignored.
```

"open_files" seem to lost its meaning. It doesn't have to do anything with open files - server might return diagnostics for the whole workspace or just open files.

"workspace" description seems fine but the name of it is not very descriptive (for example could be called "limited_to_workspace" instead).

If we are OK with just completely changing naming around and break backwards compatibility then we could probably come up with better naming. Are we OK with that?

For what it's worth, no servers seem to use it in their default settings.
Of course people might be using it already and would be affected but probably not that many and if it breaks for them nothing really bad will happen. People might just notice and figure out how to fix it.

jwortmann: I wonder if we shouldn't just remove this config alltogether, and simply decide for one of the two behaviors, and without it being configurable. I wonder how other editors behave and if the have a config option for this (I assume they don't).

Or we could make it a user setting, because this is not really something specific to the server, but just personal preference. So I'd say it doesn't really make sense to have this as a server-specific config.

rchl: I would be fine with user setting that defaults to ignoring diagnostics outside of workspace.

I suppose the setting should then be a `hide_non_project_diagnostics` boolean like before (or some variation of it).

jwortmann: If we want to keep it as part of the server config, we could probably just rename "open_files" to anything else, because that is currently the default value. And that option isn't even referenced anywhere in the source code. The only place left where `diagnostics_mode` is checked is in
https://github.com/sublimelsp/LSP/blob/72796d3b1a774127ef60b60ee22b83cdb0303fca/plugin/core/windows.py#L369-L374

---

Another question is what to do about code actions with files outside of the workspace folders, if someone has `"diagnostics_mode": "workspace"` configured. Because currently, we completely ignore the diagnostics (instead of just not drawing them in the view):

For pull diagnostics, no request is made:
https://github.com/sublimelsp/LSP/blob/72796d3b1a774127ef60b60ee22b83cdb0303fca/plugin/session_buffer.py#L620-L623

For publishDiagnostics:
https://github.com/sublimelsp/LSP/blob/72796d3b1a774127ef60b60ee22b83cdb0303fca/plugin/core/sessions.py#L1911-L1914

So it's probably useless to automatically request code actions (for current caret position) for such views.

rchl: > Another question is what to do about code actions with files outside of the workspace folders

your suggestion makes sense to me

rchl (approved):

docs/src/client_configuration.md
jwortmann: Yeah but I realized that we actually don't request code actions if there aren't any diagnostics (e.g. if ignored due to outside of workspace folders if configured like that). I have added that logic recently when refactoring the code actions. So it should be fine already.

Perhaps it could be optimized to return early from the function when the file is outside of workspace folders. But that's probably not so important.

PR #2849, opened 2026-04-05
Update LSP types

jwortmann:


jwortmann:
I've added `DiagnosticClientCapabilities.markupMessageSupport` manually (see https://github.com/sublimelsp/lsp-python-types/pull/34#issuecomment-4189204935).

jwortmann:
Lint fails at the moment, but types are correct because we don't declare `markupMessageSupport` in the initialize params, so the diagnostics message from the server can only be a string.

I could add some `pyright: ignore` comments, but it would also be fixed automatically if merging #2845 (I will revert the last commit therein after lsp types are updated). So if we could ignore the lint CI fail for now, I wouldn't need to resolve any merge conflicts there.

rchl:
> Lint fails at the moment, but types are correct because we don't declare `markupMessageSupport` in the initialize params, so the diagnostics message from the server can only be a string.
> 
> I could add some `pyright: ignore` comments, but it would also be fixed automatically if merging #2845 (I will revert the last commit therein after lsp types are updated). So if we could ignore the lint CI fail for now, I wouldn't need to resolve any merge conflicts there.

I'd suggest removing changes related to MarkupContent and only adding them together with the implementation. Don't think it's that important to introduce LSP changes only through dedicated commit.

jwortmann:
> I'd suggest removing changes related to MarkupContent and only adding them together with the implementation. Don't think it's that important to introduce LSP changes only through dedicated commit.

Okay, done in #2845

PR #2859, opened 2026-04-10
Fix some more ruff issue

rchl:


plugin/core/configurations.py
```diff
@@ -127,13 +130,14 @@ def record_crash(self, config_name: str, exit_code: int, exception: Exception |
...
         return crash_count < RETRY_MAX_COUNT
 
     def _reenable_disabled_for_session(self, config_name: str) -> bool:
         try:
             self._disabled_for_session.remove(config_name)
-            return True
         except KeyError:
             return False
+        else:
+            return True
```
predragnikolic: On first glance, the previous code was more readable

rchl: For some probably is (especially coming from other languages). For hardcore python people it's probably just fine.

I don't care either way as long as it's enforced and I don't have to think about it and the less ignore rules we have the better.

predragnikolic: I got used to seeing `for else`.

> For hardcore python people it's probably just fine.

Yes this is the first time I see `try except else`.

Already got used to it now.

But functionally, both do the same thing, but the previous approach uses less code. (One keyword less) 😅 And the old code is more natural to write, I will take the time to think when I would use try else in practice.

predragnikolic: Found one example:
```py
try:
    cur.exec(begin_tran)
    cur.exec(insert_data_to_stg)
    cur.exec(merge_data_from_stg)
    cur.exec(commit_tran)
except:
    cur.exec(rollback_tran)
    notify_fail()
else:
    # this must go here because we cannot rollback a transaction that has already committed if this function raises an exception 
    notify_success()
```
I would just write it as 
```py
try:
    cur.exec(begin_tran)
    cur.exec(insert_data_to_stg)
    cur.exec(merge_data_from_stg)
    cur.exec(commit_tran)
    notify_success()
except:
    cur.exec(rollback_tran)
    notify_fail()
```

predragnikolic: https://www.reddit.com/r/Python/comments/y4lndz/try_except_else_question_when_do_you_use_else/

predragnikolic: I learned something new.
I don't mind the change now

rchl: Yeah, python documentation also has an example that shows that:

```py
for arg in sys.argv[1:]:
    try:
        f = open(arg, 'r')
    except OSError:
        print('cannot open', arg)
    else:
        print(arg, 'has', len(f.readlines()), 'lines')
        f.close()
```

https://docs.python.org/3/tutorial/errors.html

jwortmann: Maybe it would be simpler to just remove the `else`, so that the `return` would be after the try/except block.

That would be similar to the convention that we used for if-blocks, like

```python
if condition:
    # some code
    return True
return False
```

instead of

```python
if condition:
    # some code
    return True
else:
    return False
```

plugin/core/tree_view.py
```diff
@@ -74,19 +77,20 @@ def html(self, sheet_name: str, indent_level: int) -> str:
         kind_class_name = KIND_CLASS_NAMES.get(self.kind[0], 'kind kind_ambiguous')
         icon_html = '<span class="{}" title="{}">{}</span>'.format(
             kind_class_name, self.kind[2], self.kind[1] or '&nbsp;')
+        escaped_tooltip = html.escape(self.tooltip)
```
jwortmann, line 80: Probably not too relevant, but previously this was only evaluated when needed, in two of the four branches below.

plugin/core/configurations.py
rchl: Yeah, could do that in this and some other cases. But it wouldn't be equivalent in this case: https://github.com/sublimelsp/LSP/blob/d12e60a4eacb548fce46ab870a02885a3d3ae3ca/plugin/core/sessions.py#L2053-L2062

plugin/core/tree_view.py
rchl: Yes, was aware but decided to lean on readability vs. not doing unnecessary work since this is not really performance sensitive code.

plugin/core/configurations.py
rchl: I refactored the tricky case to get rid of this syntax in all cases.

jwortmann (approved):

Issue #2800, opened 2026-03-09
Needed refactorings


jwortmann:
This is a list of refactorings that I think would be useful, in order to make the behavior of LSP more correct, more efficient, and would make it possible to implement certain new functionalities. Please comment if you have opinions about these, or better or more suggestions.

---

### ~~Move logic for code action requests into SessionBuffer class~~ implemented in #2835

One way how code action requests are triggered is whenever new diagnostics arrive (from `publishDiagnostics` notifications, from pull diagnostic responses, or from workspace diagnostics partial result progress reports):
https://github.com/sublimelsp/LSP/blob/294a1af081fef9b0b25578e8cf5d0de8ea729424/plugin/documents.py#L343-L349

Currently, code action requests for all sessions are created at the same time in
https://github.com/sublimelsp/LSP/blob/294a1af081fef9b0b25578e8cf5d0de8ea729424/plugin/code_actions.py#L149-L161

Therefore, in a multi-session environment, a new code action request is sent again to a particular session, even when the diagnostics from *another* session arrive. This results in many useless and repeated code action requests for multi-session setups.

Instead, when diagnostics arrive, the code action request should only be triggered for that particular session. So the logic for code action requests, or at least some parts of that, should be moved into the SessionBuffer, and then it can be triggered directly from `SessionBuffer.on_diagnostics_async`. When code actions are triggered due to selection change, or due to code actions on save/format, we then still need to iterate all SessionBuffers for the individual requests.

Then the https://github.com/sublimelsp/LSP/blob/294a1af081fef9b0b25578e8cf5d0de8ea729424/plugin/documents.py#L217
flag could be moved into the SessionBuffer, as part of the `SessionBuffer.pending_refreshes` bitflag.

---

### ~~Move server configurations into its own settings file~~ implemented in #2862

Currently, server configurations are stored under a "clients" key in the main `LSP.sublime-settings` file.
There is a change listener for that settings file and previously all servers were restarted whenever anything in that settings file changed. In #2448 I implemented that LSP keeps a hash value derived from the "clients" setting in memory, and when a change to the settings file is detected, it is compared with the new hash value and servers are only restarted if something in the "clients" did actually change. This allowed general LSP settings to be modified without automatically restarting all servers.

One idea is to move the server configurations into a new settings file, then we could use a separate change listener for that file and the current hash comparison for "clients" wouldn't be necessary. The individual server configs could then be top level keys in that settings file. Possible names for the new file could be
[...]

Issue #2243, opened 2023-04-30
All servers should share Workspace Configuration


rchl:
**Is your feature request related to a problem? Please describe.**

In VSCode (and maybe other editors) all servers have access to the whole editor configuration which means that one server can access configuration of another server and which is why server `settings` are normally scoped with a prefix of the server name or similar (for example `json.*` for `LSP-json` or `javascript.*` or `typescript.*` for `LSP-typescript`  etc.).

That means that one server can reuse settings of another server. For example Volar being a sort of a "meta server" can read settings of the LSP-html, LSP-typescript, LSP-css, etc...

**Describe the solution you'd like**

Opening server-specific settings like `Preferences: LSP-json Settings` should expose `settings` object as the root object on the left and a global `LSP-server-settings.sublime-settings` on the right.

Question 1: Where to expose `initializationSettings`, `selector`, `command` and other settings in that case? In VSCode the extension can contribute extra settings that are then exposed.

Question 2: How to specify which settings should be sent to the extension on client-initialized `workspace/didChangeConfiguration`? In VSCode the extension does that with some extra code.

Question 3: Since settings of LSP-typescript would then apply for both .js, .ts and .vue files, one might want to specify different `javascript.*` settings for vue files so ideally we'd also support language-id specific overrides. In VSCode possible with something like:
```js
    "[python]": {
        "editor.formatOnType": true
    },
```

**Additional context**

All in all this doesn't seem very feasible to do in ST but I've reported as a formality as this is how settings are handled in VSCode.

rchl:
I suppose this would solve https://github.com/sublimelsp/LSP/issues/2044 as we wouldn't have to restart the server knowing that only workspace configuration has changed.

Also I remember we've discussed that quite a bit but not sure where that was. Maybe in some PR. I remember @jwortmann has been experimenting with separating workspace settings.

jwortmann:
> In VSCode (and maybe other editors) all servers have access to the whole editor configuration which means that one server can access configuration of another server and which is why server `settings` are normally scoped with a prefix of the server name or similar

The difference in VSCode is, that it uses only a single settings file for everything, i.e. global editor settings, language specific settings, and settings from extensions all in one file. This is why the prefix like `json.*` in the setting names exists, so that there is no risk of naming conflicts. The fact that language servers can access the whole editor configuration is a byproduct of that, and not the cause.

[...]

PR #2862, opened 2026-04-12
Add support for workspace/didChangeConfiguration

jwortmann:
This PR implements the second point in #2800 

* Removed all remaining "default_clients".
* Introduces a new settings file `LanguageServers.sublime-settings` for the manual configurations. This allows for a cleaner separation of LSP settings and server configs, so we can attach different change listeners to these files.
* The "clients" in `LSP.sublime-settings` still work, but I've marked it as deprecated in the JSON schema. If there is a configuration with the same name also in `LanguageServers.sublime-settings`, then that will take precedence.
* Adds support for `workspace/didChangeConfiguration` to notify the server when settings were changed. This is only implemented for configs from the new `LanguageServers.sublime-settings` file and for external configs (from helper packages), but not for the "clients" in the LSP settings. Servers can opt-in to change notifications via dynamic registration (see https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration). This allows to dynamically change settings while the server is running, without forcing a restart of the server (restart can take a long time for example if the server is implemented in a JIT-compiled language, or if project indexing is slow).
* `workspace/didChangeConfiguration` can be tested for example with JETLS. (That server shows the updated settings after a new configuration request in a `window/showMessage` notification, which we log to the console.)
* If a server doesn't support (or opts into) `workspace/didChangeConfiguration`, nothing happens until the server initiates a new `workspace/configuration` request by itself (which might never happen), or until the server is manually restarted via *LSP: Restart Server*. This could be unexpected to users, so I wonder if we should handle that case also in LSP directly. This happens for example with Pyright.

Is the change with the new config file too intrusive? IMO it's nice to have a clean separation between LSP settings and server configs, and I always disliked that server configurations were named "clients" in the settings file.


plugin/core/settings.py
jwortmann: By the way, is this existing line a bug? How/why does it work? The `config` variable should be a dict, so shouldn't this throw a runtime error when trying to use dot-access? Souldn't it use `config['enabled'] = is_enabled` instead?

predragnikolic: It would not work on a dict,
It does work with the new charges introduced in
https://github.com/sublimelsp/LSP/commit/ead1085da645c6ad71e2efa9a3104b276502e605
And
https://github.com/sublimelsp/LSP/pull/2765/changes

predragnikolic: But after re-reading your message, I need to try it out, because your reason is valid

rchl:
I did not look at the changes yet but I have a question.

Some packages like for example LSP-pyright implement a custom configuration setting like `pyright.dev_environment` that is not really a server configuration but rather a package configuration. Server doesn't know anything about it and we should actually restart server when it changes.

Does this change affect that? Will we need to update those packages and move the setting to the root of the settings, for example?

plugin/core/settings.py
rchl: Looks like a bug. The dict case need to be handled differently. The case above uses `ClientConfig` so that one is fine.

jwortmann:
> Some packages like for example LSP-pyright implement a custom configuration setting like `pyright.dev_environment` that is not really a server configuration but rather a package configuration.

I know, and I believe I have written somewhere before that it is a bad idea and helper packages shouldn't mix custom package settings with server settings. So I would consider that as a bug in LSP-pyright. In LSP-julia I have ensured that custom package settings don't mix with server settings: https://github.com/sublimelsp/LSP-julia/blob/4751ae10ab015451292c11a8b1e9cc65594c8fac/LSP-julia.sublime-settings#L2-L9

> Server doesn't know anything about it and we should actually restart server when it changes.

With this PR, servers are restarted if the ClientConfig does not equal (`==`) the old config anymore. But the `_all_settings` are currently excluded from that due to starting with an underscore (see `__eq__` implementation). I don't know if there was a reason for that, but we could consider to change that property name if servers should be restarted on custom package settings change. So this should be a trivial change. But we might want to discuss if that should actually happen. These custom settings might be unrelated to the running server, so it might be desired to keep the server running. If the server should be restarted on a change, I guess the helper package could also do it manually via `lsp_restart_server` command. Although that would require another change listener to be implemented also in the helper package, so I'm not saying that I'm particular against restarting servers on package setting change.

> Does this change affect that? Will we need to update those packages and move the setting to the root of the settings, for example?

Servers that opt-in to didChangeConfiguration notifications would get a notification and then should pull the configuration. If there is no change to the settings that are known by the server, in theory nothing should happen.

plugin/core/settings.py
```diff
@@ -101,21 +112,38 @@ def update_configs(self) -> None:
             self._notify_userprefs_listener()
             return
         self._clients_hash = clients_hash
-        clients = DottedDict(read_dict_setting(settings_obj, "default_clients", {}))
-        clients.update(clients_dict)
+        clients = DottedDict(clients_dict)
+        clients.update(sublime.load_settings(SERVER_CONFIGS_FILENAME).to_dict())
```
rchl: Can we use `_configs_registration.settings.to_dict()`?

rchl:
> > Some packages like for example LSP-pyright implement a custom configuration setting like `pyright.dev_environment` that is not really a server configuration but rather a package configuration.
> 
> I know, and I believe I have written somewhere before that it is a bad idea and helper packages shouldn't mix custom package settings with server settings. So I would consider that as a bug in LSP-pyright. In LSP-julia I have ensured that custom package settings don't mix with server settings: https://github.com/sublimelsp/LSP-julia/blob/4751ae10ab015451292c11a8b1e9cc65594c8fac/LSP-julia.sublime-settings#L2-L9

It was required to do that that way before because otherwise you wouldn't be able to do project-specific override.

> With this PR, servers are restarted if the ClientConfig does not equal (`==`) the old config anymore. But the `_all_settings` are currently excluded from that due to starting with an underscore (see `__eq__` implementation). I don't know if there was a reason for that, but we could consider to change that property name if servers should be restarted on custom package settings change. So this should be a trivial change. But we might want to discuss if that should actually happen. These custom settings might be unrelated to the running server, so it might be desired to keep the server running.

I would think that restarting on a change of those would be more often correct given what those settings are used for currently (setting server path or toggling server manage). In your lsp-julia example one would require restart, the other perhaps not?

The `_all_settings` feature is not currently used by anything, I believe, so we should still be able to make breaking changes. And as I said in my newly opened discussion, I think I would want to do something differently there to be able to type things better.

> Servers that opt-in to didChangeConfiguration notifications would get a notification and then should pull the configuration. If there is no change to the settings that are known by the server, in theory nothing should happen.

That's understandable but I'm obviously talking about cases like the one from LSP-pyright in which case we should do something.

Can/should we handle those cases gracefully somehow?
Perhaps opt-in mechanism so that packages have a chance to migrate?
I know that this is not ideal and requires more work for little gain but breaking user's config is not great either.

plugin/core/settings.py
jwortmann: https://github.com/sublimelsp/LSP/pull/2862/changes/cddf97eaf911da905c5d2270fe1e5053d30528fa

predragnikolic:
related issues: 
https://github.com/sublimelsp/LSP/issues/2044
https://github.com/sublimelsp/LSP/issues/2243

plugin/core/types.py
```diff
@@ -1116,10 +1126,7 @@ def __repr__(self) -> str:
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, ClientConfig):
             return False
-        for k, v in self.__dict__.items():
-            if not k.startswith("_") and v != getattr(other, k):
-                return False
-        return True
+        return self.name == other.name and self._all_settings == other._all_settings
```
rchl, line 1132: Should we actually have a dedicated method for comparing two configs? I feel like overriding the `__eq__` implementation for that purpose is not right. Technically `__eq__` should answer the question whether two objects are identical but we're using it more for "did settings that cause server restart change".

Also it's hard to search for places that rely on this check to verify if we didn't break something by changing this implementation.

jwortmann:
I have noticed one strange bug with this PR that makes helper packages often not work correctly.
Sometimes the change listener for the settings files from helper packages don't work. Sometimes I can set `"enabled": false` and then the server is shutdown as expected, but when I set `"enabled": true` again, no more change event gets triggered. But I don't know yet what causes this or how I broke it in this PR.

I really wonder what the purpose of this code in SettingsRegistration is:
https://github.com/sublimelsp/LSP/blob/b2cd51d90859d6a41949e681b84e8dba0aec2de7/plugin/core/types.py#L179-L188

When helper plugins are registered, we create a ClientConfig for each of them. This included a SettingsRegistration object, which is stored as part of the ClientConfig. But when it is garbage collected, the change listener is removed, so new changes in the sublime-settings file don't trigger the callback function anymore. And when something in the settings file changes, we have to create a new ClientConfig and replace the old one that was stored in the `ClientConfigs.all` and `ClientConfigs.external` dicts, which in turn triggers the old ones (including the SettingsRegistration) to be garbage collected. In theory I think it should work, because a new ClientConfig object with new SettingsRegistration should be created before replacing the old one:
https://github.com/sublimelsp/LSP/blob/b2cd51d90859d6a41949e681b84e8dba0aec2de7/plugin/core/settings.py#L83-L85

I don't think I would have written such complicated code with all these self references and weak references that rely on things not being garbage collected for change listeners to work. Why not just attach a single change listener function when a plugin gets registered, and keep that around forever (until plugin unregistered)?

In general when writing the code in this PR I saw so much really weird and complicated code with all kind of self references and custom change listener abstractions in abstract base classes which even result in reference cycles between objects. And all these abstract base class interfaces make "Goto Definition" not work in the LSP code base, because Pyright always jumps to the abstract base class instead of to the implementation (I noticed that same annoyance many times before with the SessionBufferProtocol and SessionViewProtocol classes).

For example the following chain of functions is called when the "clients" in LSP.sublime-settings gets modified:

| File | Method |
| ---- | ------ |
| settings.py | `_on_sublime_settings_changed` |
| settings.py | `ClientConfigs.update_configs` |
| settings.py | `ClientConfigs._notify_clients_listener` |
| windows.py  | `WindowRegistry.on_client_config_updated` |
| configurations.py | `WindowConfigManager.update` |
| configurations.py | `WindowConfigManager._reload_configs` |
| windows.py | `WindowManager.on_configs_changed` |
| windows.py | `WindowManager.restart_session_async` |

This jumps from `settings.py` to `windows.py` to `configurations.py` back to `windows.py`...

---

*Edit:* And another point why I think relying for things trigger on garbage collection is bad:
Imagine a future Python version doesn't trigger garbage collector immediately. Then if there is a change to the settings file of a helper package, the old change listener from the SettingsRegistration object in the old ClientConfig is still around, so if there is another change to that settings file, we would now get 2 distinct change events, i.e. our functions would also trigger twice, which again creates 2 new ClientConfig objects with corresponding change listeners. So I'd say that entire SettingsRegistration class is super dangerous...

plugin/core/types.py
jwortmann: Yeah maybe. I also had the same thoughts before. But I don't think there is any other code that would compare two ClientConfigs, and this already had some custom implementation (which excludes properties starting with `_` from being considered when compared for equality, which for example already includes the `"enabled"`, so I guess we could do the same with `"settings"`). There is also no comment in the source code that explains the exact purpose of the current custom `__eq__` implementation. So I thought we could just use `__eq__` when checking for "server-restart-equality"...

rchl:
> I really wonder what the purpose of this code in SettingsRegistration is:

It might be wrong but also `self.settings.add_on_change` is potentially a bit magic (at least to me). I guess a case where there are two listeners using same key is not really supported because calling `clear_on_change` from one of those places will then clear both listeners (I think). Maybe the issue has something to do with that behavior?

> I don't think I would have written such complicated code with all these self references and weak references that rely on things not being garbage collected for change listeners to work. Why not just attach a single change listener function when a plugin gets registered, and keep that around forever (until plugin unregistered)?

Makes sense in theory but I would have to dig into the code to understand things better.

> In general when writing the code in this PR I saw so much really weird and complicated code with all kind of self references and custom change listener abstractions in abstract base classes which even result in reference cycles between objects. And all these abstract base class interfaces make "Goto Definition" not work in the LSP code base, because Pyright always jumps to the abstract base class instead of to the implementation (I noticed that same annoyance many times before with the SessionBufferProtocol and SessionViewProtocol classes).

Yes, but there is "Go to implementation" command to jump from abstract class to implementations.

> _Edit:_ And another point why I think relying for things trigger on garbage collection is bad:

That specific case of SettingsRegistration didn't rely on GC originally. It was done like that to workaround an issue with ST freezing - https://github.com/sublimelsp/LSP/pull/2822

jwortmann:
> I have noticed one strange bug with this PR that makes helper packages often not work correctly.

Okay, I figured out that this is unrelated to this PR and also happens on main (can be tested by manually setting `"enabled": false`, then save, and then set back to true in a config). This bug was introduced with https://github.com/sublimelsp/LSP/commit/e31ee1e933d6bd445057f752307d45f3b7a605ba. I think by using deepcopy for the SettingsRegistration object in https://github.com/sublimelsp/LSP/blob/b2cd51d90859d6a41949e681b84e8dba0aec2de7/plugin/core/types.py#L985 we lose the reference to the old SettingsRegistration object when the ClientConfig that is passed as an argument gets garbage collected. The deepcopy copies only the `self.settings` and `self.settings_path` properties (I think). And then the new copy doesn't have any change listener function attached.

rchl:
> Okay, I figured out that this is unrelated to this PR and also happens on main (can be tested by manually setting `"enabled": false`, then save, and then set back to true in a config). This bug was introduced with [e31ee1e](https://github.com/sublimelsp/LSP/commit/e31ee1e933d6bd445057f752307d45f3b7a605ba). I think by using deepcopy for the SettingsRegistration object in

Good catch. I didn't intend to copy non-plain objects like that and I didn't do it in `from_sublime_settings` method.

Created https://github.com/sublimelsp/LSP/pull/2867

plugin/core/settings.py
jwortmann: I fixed it in this PR (https://github.com/sublimelsp/LSP/pull/2862/changes/6ab5347b305ce6a1e15bbff392d971e939608109)

pyproject.toml
```diff
@@ -144,6 +144,7 @@ ignore = [
     "SIM110",  # https://docs.astral.sh/ruff/rules/reimplemented-builtin/
     "SLF001",  # https://docs.astral.sh/ruff/rules/private-member-access/
     "T20",  # https://docs.astral.sh/ruff/rules/#flake8-print-t20
+    "TD001",  # https://docs.astral.sh/ruff/rules/invalid-todo-tag/
```
rchl, line 144: You could have just used TODO: Fix... :)

It's easier to search just for "TODO" when searching for pending issues than seraching both for TODO and FIXME.

LanguageServers.sublime-settings
```diff
@@ -0,0 +1,73 @@
...
+  //   "command": ["pyls"],
+  //
+  //   // This selector filters which views apply to this configuration, in
+  //   // the same way that a selector for a build system filters which build
+  //   // system applies to which view. To determine the base scope of your
+  //   // view, click on Tools > Developer > Show Scope Name. The first line
+  //   // in the popup is the base scope.
+  //   "selector": "source.python",
+  //
+  //   # ------ From this point onward, each key-value is **optional** -------
```
rchl: I suppose the intention is that one will copy parts of this config to user side and modify. That's why there are those double `//` markers?

In that case we should replace `#` with `//` since those won't be valid after uncommenting.

Also first line can do without double `//` as it doesn't have to be included when copying.

```suggestion
  // Each new server configuration must have the following structure.
  // "SERVER_NAME": // some human-friendly name
  // {
  //   // ----------- The following three settings are **required** -----------
  //
  //   // Enable or disable this client configuration.
  //   "enabled": true,
  //
  //   // The command line required to run the server.
  //   "command": ["pyls"],
  //
  //   // This selector filters which views apply to this configuration, in
  //   // the same way that a selector for a build system filters which build
  //   // system applies to which view. To determine the base scope of your
  //   // view, click on Tools > Developer > Show Scope Name. The first line
  //   // in the popup is the base scope.
  //   "selector": "source.python",
  //
  //   // ------ From this point onward, each key-value is **optional** -------
```

sublime-package.json
rchl, line 1: Maybe lets add a snippet to `ClientConfig` to make the process a bit easier?

```diff
diff --git a/sublime-package.json b/sublime-package.json
index 4641a93..8c6a36a 100644
--- a/sublime-package.json
+++ b/sublime-package.json
@@ -325,6 +325,16 @@
             "ClientConfig": {
               "type": "object",
               "additionalProperties": false,
+              "defaultSnippets": [
+                {
+                  "body": {
+                    "command": ["$0"],
+                    "selector": "source.php",
+                    "enabled": true,
+
+                  }
+                }
+              ],
               "markdownDescription": "The configuration that informs LSP on how to start your language server and to which views it should attach the language server process",
               "properties": {
                 "command": {
```

LanguageServers.sublime-settings
```diff
@@ -0,0 +1,73 @@
+{
+  // // Each new server configuration must have the following structure.
+  // "SERVER_NAME": // some human-friendly name
```
rchl, line 6: Maybe add a link to language_servers page in docs?

plugin/core/types.py
```diff
@@ -655,6 +659,9 @@ def text_sync_kind(self) -> TextDocumentSyncKind:
         value: TextDocumentSyncKind = self.get("textDocumentSync.change.syncKind")
         return value if isinstance(value, int) else TextDocumentSyncKind.None_
 
+    def should_notify_did_change_configuration(self) -> bool:
+        return 'workspace.didChangeConfiguration' in self
```
rchl, line 663: Is this looking for `workspace/didChangeConfiguration` in **server capabilities**?
If so then I don't think it will ever find it because it's a client capability - https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeConfiguration

EDIT: I've missed your comment:

> * dds support for `workspace/didChangeConfiguration` to notify the server when settings were changed. This is only implemented for configs from the new `LanguageServers.sublime-settings` file and for external configs (from helper packages), but not for the "clients" in the LSP settings. Servers can opt-in to change notifications via dynamic registration (see https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration). This allows to dynamically change settings while the server is running, without forcing a restart of the server (restart can take a long time for example if the server is implemented in a JIT-compiled language, or if project indexing is slow).

But what the heck is going on with that?

Is this the part of the spec that refers to that?

> This pull model replaces the old push model were the client signaled configuration change via an event. If the server still needs to react to configuration changes (since the server caches the result of workspace/configuration requests) the server should register for an empty configuration change using the following registration pattern:
> 
> connection.client.register(DidChangeConfigurationNotification.type, undefined);

First of all, why did someone put a snippet from node-language-server implementation instead of LSP message and how am I supposed to know what capability does it represent? :)

Also, it might be correct to only send configuration changes if server registers for it using mechanism above but in practice I feel like that might actually give worse results. In the sense that servers won't receive configuration changes even though they might be perfectly capable of handling them? Though I might be missing something... For example ruff won't receive updated settings because it doesn't register for them using that mechanism. But is it in practice handling didChangeConfiguration regardless? I guess we never knew because we always restarted the server but I always thought that server will be able to process such push configuration changes.

rchl: I've checked ruff + vscode and it also doesn't send didChangeConfiguration but it does restart the server instead. Assuming that your changes are correct per spec, shouldn't we also restart the server instead of doing nothing?

EDIT: And I see now that you've also covered that already with:

> If a server doesn't support (or opts into) workspace/didChangeConfiguration, nothing happens until the server initiates a new workspace/configuration request by itself (which might never happen), or until the server is manually restarted via LSP: Restart Server. This could be unexpected to users, so I wonder if we should handle that case also in LSP directly. This happens for example with Pyright.

Then I would say yes, we should restart the server.

LanguageServers.sublime-settings
predragnikolic, line 6: Is SERVER_NAME the configuration name 
or is it the PackageName?

pyproject.toml
jwortmann: I often use TODO for things that would be "nice to have", like additional features, but which are probably not so important right now. And FIXME is for rather critical things that might need more attention.

I also use a custom Python syntax extending the default one, but which adds special scopes to such comments:
```yaml
%YAML 1.2
---
name: Python
scope: source.python
version: 2

extends: Packages/Python/Python.sublime-syntax

file_extensions:
  - py

contexts:

  type-hint:
    - meta_prepend: true
    - match: (?i:(TODO))\b[^\n]*
      scope: meta.codetag.todo.python
      captures:
        1: entity.name.label.todo.python
      pop: true
    - match: (?i:(FIXME))\b[^\n]*
      scope: meta.codetag.fixme.python
      captures:
        1: entity.name.label.fixme.python
      pop: true
    - match: https?://\S*
      scope: markup.underline.link.python
      pop: true
```

So having both TODO and FIXME allows to use different colors for them (although I actually use red for both of them and just made TODO blue here for this screenshot):

todo

LanguageServers.sublime-settings
jwortmann: > I suppose the intention is that one will copy parts of this config to user side and modify. That's why there are those double `//` markers?

I assume so. I just copied that over from the other file, without any modifications (except for removing the preamble comment in front of the "clients" in the original setting).

LanguageServers.sublime-settings
jwortmann: What do you mean with PackageName here? This is the settings file for the manually configured servers (previously "clients" in LSP.sublime-settings).

plugin/core/types.py
jwortmann: I considered that, but I would now tend to not force server restart if the server doesn't opt-in to change notifications.

The specs are pretty clear on that and the link that I have put as a comment in the source code is from 2019, so there should have been enough time for servers to update to the newer pull model now. Although I see no real reason why a server would **not** register for that notification - but then why is that guarded behind a server opt-in anyway? Maybe old clients would always send the notification on settings change, but then include either all settings, or the setting which were changed (that would require additional diff work in the client). So maybe that was the reason to require server registering for that notification now, so that it can properly handle the "empty" notification.

Maybe I'll reconsider, but I think it's more the server responsibility to do what's intended in the specs.

rchl: Notifying servers that don't register for notification might be debatable but I don't understand why would we not restart the server in that case.

If user is changing settings then he obviously wants the settings to get applied. Not trigger another command to restart it. And the message saying that server night need to be restarted is almost guaranteed to be overlooked.

LanguageServers.sublime-settings
jwortmann: Link added in https://github.com/sublimelsp/LSP/pull/2862/changes/dd09bbe106c0a21a6d740380fd92242614cb3243.
Also added back `//` for the first line because now everything that is not code has double-comments, so it's more consistent.

sublime-package.json
jwortmann: Hm, not sure if that is so useful, the description (comments) on the left side of the settings window is already very detailed and the filled in "body" name and "source.php" selector seem quite arbitrary.

plugin/core/types.py
jwortmann: Server might only have limited settings and could use `workspace/configuration` each time before doing relevant actions like formatting, if it doesn't want to cache settings.

In general, the `workspace/configuration` pull model gives the responsibility for pulling relevant settings to the server. How that is done in detail, is up to the server. It might pull settings on demand before responding to relevant requests, or pull settings periodically, or after change notifications if it wants to receive them. This PR ensures that we always have the latest settings from the settings file.

Maybe this should be something to discuss in the specs repo and/or issue tracker of the respective servers? I doubt that automatic restart of servers on settings change is something that would be intended in the LSP specs.

sublime-package.json
rchl: easier to trigger snippet and change then retype everything or paste that commented out out code.
selector value could be a $1 placeholder.

plugin/core/types.py
rchl: Frankly I don't like this approach.

Regardless if it's per spec or not, it will just make the experience worse for the majority, if not 99%, of servers.

When changing settings I want them to apply and I know that most popular servers won't refetch settings so will be running with outdated settings.

I think that something went wrong when designing that part of the spec and it's  probably un-fixable now but it's not good.

If initially there was only push model and then the pull model was introduced but without server capability then how is client supposed to know that it's not supposed to send didChangeConfiguration on changes anymore?

jwortmann: > Regardless if it's per spec or not, it will just make the experience worse for the majority, if not 99%, of servers.

Why do you think most servers don't support it? In my experience most servers support it and do register for that method. I don't know where you got the 99% number from.

* I have tested JETLS before and it does register `workspace/didChangeConfiguration`. And I just tested a few more servers:

* LSP-julia does register:
```
:: [21:36:58.931] <-- julia client/registerCapability (fd1777b2-cec4-4e46-808d-b95838c3006a): {'registrations': [{'id': 'aaf0e886-face-433b-ba7d-e8fde2a6817d', 'method': 'workspace/didChangeConfiguration'}]}
```

* LST-Tinymist does register:
```
:: [21:33:17.206] <-- LSP-Tinymist client/registerCapability (2): {'registrations': [{'id': 'config', 'method': 'workspace/didChangeConfiguration'}]}
```

* LSP-TexLab does register:
```
:: [21:39:43.082] <-- LSP-TexLab client/registerCapability (2): {'registrations': [{'id': 'pull-config', 'method': 'workspace/didChangeConfiguration'}]}
```

So maybe it's just unlucky for your case that Ruff doesn't (I haven't tested, because I don't use it). But it should be trivial to add support for it on the server side, so maybe it would even be "fixed" quickly if you open an issue in the server repo and ask for it?

rchl: pyright, ruff, typescript, json, eslint are the main that I use.

We are not making anything worse by restarting servers that don't opt-in.

jwortmann: > If initially there was only push model and then the pull model was introduced but without server capability then how is client supposed to know that it's not supposed to send didChangeConfiguration on changes anymore?

I would have to find how it was supposed to work with the old push model, because it seems that was quite long ago and the pull model probably already existed when I started contributing to this repository. But I'd guess that clients would either send a notification with the settings that actually changed, or maybe just with all settings from the user config. That is why the params have this `{"settings": LSPAny}` type.

Perhaps we could do that in the `else` case, i.e. if the server doesn't register for `workspace/didChangeConfiguration`. Then it should still work for servers that only support the old push model.

*Edit:* Although the question would then be *what* settings structure exactly to include. Because the `section` used by servers for the `workspace/configuration` requests can look quite different. Some servers leave section empty, some use a single item with prefix like `"jetls"`, and some use separate items with dotted section paths. So if we put just the entire "settings" dict from the settings file it might not be what all servers expect there. I guess that is the reason why the old push model was deprecated.

rchl: We send the same data that we are sending in the initial didChangeNotification

jwortmann: I found https://github.com/microsoft/language-server-protocol/issues/1888#issuecomment-1916443267. So it seems that the old push model is still supported.

We fall back to the old push model now if server doesn't register for the pull model: https://github.com/sublimelsp/LSP/pull/2862/changes/a01eab559f08158de6826f2d9368bd60e1bcddaa

I think that should fix it for your servers that don't want to update to the specs change from ~2018/2019.

pyproject.toml
rchl: You kinda admitted with this that both are equal :)

Also I wouldn't agree that it's something that needs to be fixed ASAP. Not anymore at least (after the the overlooked bug was fixed).

rchl (approved):

jwortmann:
Is everyone okay with the new settings file? This is technically not a breaking change, because the old "clients" still work, but we should probably add one or two sentences in the release notes about this.

Also, as soon as this PR is merged, the description for manual server configurations on the docs website won't match to what is currently possible in the latest released version of the package. That could be a bit confusing for new users, until the next release is made.

rchl:
> Is everyone okay with the new settings file? This is technically not a breaking change, because the old "clients" still work, but we should probably add one or two sentences in the release notes about this.

You can initialize that update message as part of this PR.

> Also, as soon as this PR is merged, the description for manual server configurations on the docs website won't match to what is currently possible in the latest released version of the package. That could be a bit confusing for new users, until the next release is made.

We can release new version soon. I don't think we have very risky stuff that we need to marinate on.

jwortmann:
Added a few notes in https://github.com/sublimelsp/LSP/pull/2862/changes/90bfba2dda3a0e7ef691cc35df62d58a25dd121b. Feel free to modify if you spot typos or have better formulations or if I missed something.

rchl (approved):

PR #2864, opened 2026-04-13
Introduce LspPlugin API (2)

rchl:
Implementing different solution inspired by @predragnikolic's comments from #2739.

It's a bit hacky initial implementation for starting the discussion.

This just handles static methods before plugin is started.

We can go further and also refactor non-static methods into `self` methods in a listener style.

### DIFF
--- a/plugin/api.py
+++ b/plugin/api.py
@@ -1,25 +1,12 @@
 from __future__ import annotations
 
-from ..protocol import ConfigurationItem
-from ..protocol import DocumentUri
-from ..protocol import ExecuteCommandParams
 from ..protocol import LSPAny
 from .core.constants import ST_STORAGE_PATH
 from .core.logging import exception_log
-from .core.promise import Promise
-from .core.protocol import ClientNotification
-from .core.protocol import ClientRequest
-from .core.protocol import ClientResponse
-from .core.protocol import Notification
-from .core.protocol import Request
 from .core.protocol import Response
-from .core.protocol import ServerNotification
-from .core.protocol import ServerResponse
 from .core.settings import client_configs
-from .core.types import ClientConfig
 from .core.types import method2attr
 from .core.url import parse_uri
-from .core.views import MarkdownLangMap
 from .core.views import uri_from_view
 from abc import ABC
 from abc import abstractmethod
@@ -39,9 +26,16 @@ import sublime
 if TYPE_CHECKING:
     from ..protocol import ConfigurationItem
     from ..protocol import DocumentUri
+    from ..protocol import ExecuteCommandParams
     from .core.collections import DottedDict
+    from .core.promise import Promise
+    from .core.protocol import ClientNotification
+    from .core.protocol import ClientRequest
+    from .core.protocol import ClientResponse
     from .core.protocol import Notification
     from .core.protocol import Request
+    from .core.protocol import ServerNotification
+    from .core.protocol import ServerResponse
     from .core.sessions import Session
     from .core.sessions import SessionBufferProtocol
     from .core.sessions import SessionViewProtocol
@@ -64,7 +58,7 @@ COMMAND_HANDLER_MARKER = '__COMMAND_HANDLER_MARKER'
 # P represents the parameters *after* the 'self' argument
 P = TypeVar('P', bound=LSPAny)
 R = TypeVar('R', bound=LSPAny)
-CommandHandler = Callable[[Any, ExecuteCommandParams], Promise[None]]
+CommandHandler = Callable[[Any, 'list[LSPAny] | None'], 'Promise[None]']
 
 
 g_plugins: dict[str, type[AbstractPlugin | LspPlugin]] = {}
@@ -252,16 +246,16 @@ def command_handler(command_name: str) -> Callable[[CommandHandler], CommandHand
     Decorator to mark a method as a handler for a specific server command.
 
     When the language server sends a `workspace/executeCommand` request with the given
-    command name, the decorated method is called with the full command payload.
+    command name, the decorated method is called with the command's `arguments` list (or `None` if absent).
 
     Usage:
         ```py
-        @command_handler('editor.action.triggerSuggest')
-        def on_trigger_suggest(self, params: ExecuteCommandParams) -> Promise[None]:
+        @command_handler('editor.action.showReferences')
+        def on_show_references(self, arguments: list[LSPAny] | None) -> Promise[None]:
             ...
         ```
 
-    :param      command_name:   The command name as advertised by the server (e.g., 'editor.action.triggerSuggest').
+    :param      command_name:   The command name as advertised by the server (e.g., 'editor.action.showReferences').
     :returns:   A decorator that registers the function as a command handler.
     """
 
--- a/plugin/api.py
+++ b/plugin/api.py
@@ -412,12 +412,21 @@ class LspPlugin(APIHandler):
     @classmethod
     def on_before_start_async(cls, context: OnBeforeStartContext) -> None:
         """
-        Update or install the server binary if this plugin manages one. Called before the server is started.
+        Called just before the language server process is started.
 
-        This method runs on a worker thread. Perform any blocking I/O (e.g. downloading a binary,
-        running ``npm install``) directly here without spawning additional threads.
+        Override to perform any preparation needed before startup - for example installing or updating server binaries,
+        resolving the working directory, or injecting extra template variables into `context.variables`.
 
-        :param      context:    The plugin context.
+        This method runs on a worker thread so perform any blocking I/O (e.g. downloading a binary, running
+        `npm install`) directly here without spawning additional threads.
+
+        Mutations to `context.working_directory` and `context.variables` are picked up and used when launching the
+        server process.
+
+        Raise `PluginStartError` with a message to abort startup and display a user-visible status message.
+
+        :param      context:    The startup context. `context.configuration`, `context.variables` and
+                                `context.working_directory` can be mutated to influence how the server is launched.
         """
         pass
 
--- a/plugin/api.py
+++ b/plugin/api.py
@@ -267,7 +267,7 @@ def command_handler(command_name: str) -> Callable[[CommandHandler], CommandHand
 
 
 @dataclass
-class IsApplicableContext:
+class ContextIsApplicable:
     configuration: ClientConfig
     """The resolved `ClientConfig` for this session."""
     view: sublime.View
@@ -277,7 +277,7 @@ class IsApplicableContext:
 
 
 @dataclass
-class OnBeforeStartContext:
+class ContextOnBeforeStart:
     configuration: ClientConfig
     """The resolved `ClientConfig` for this session."""
     variables: dict[str, str]
@@ -290,7 +290,7 @@ class OnBeforeStartContext:
 
 
 @dataclass
-class OnStartContext:
+class ContextOnStart:
     transport: TransportWrapper
 
 
@@ -383,7 +383,7 @@ class LspPlugin(APIHandler):
         unregister_plugin_impl(cls)
 
     @classmethod
-    def is_applicable(cls, context: IsApplicableContext) -> bool:
+    def is_applicable(cls, context: ContextIsApplicable) -> bool:
         """
         Determine whether the server should run on the view given by `context.view`.
 
@@ -404,7 +404,7 @@ class LspPlugin(APIHandler):
         return False
 
     @classmethod
-    def on_before_start_async(cls, context: OnBeforeStartContext) -> None:
+    def on_before_start_async(cls, context: ContextOnBeforeStart) -> None:
         """
         Called just before the language server process is started.
 
@@ -446,7 +446,7 @@ class LspPlugin(APIHandler):
         super().__init__()
         self.weaksession: ref[Session] = weaksession
 
-    def on_start_async(self, context: OnStartContext) -> None:
+    def on_start_async(self, context: ContextOnStart) -> None:
         """
         Called after the transport is established but before the LSP ``initialize`` request is sent.
 
--- a/plugin/api.py
+++ b/plugin/api.py
@@ -433,6 +433,10 @@ class LspPlugin(APIHandler):
         """
         super().__init__()
         self.weaksession: ref[Session] = weaksession
+        self.execute_commands: dict[str, Callable[[ExecuteCommandParams], None]]
+
+    def on_start_async(self) -> None:
+        pass
 
     def on_workspace_configuration(self, params: ConfigurationItem, configuration: Any) -> Any:
         """
@@ -445,7 +449,7 @@ class LspPlugin(APIHandler):
         """
         return configuration
 
-    def on_execute_command_async(self, command: ExecuteCommandParams) -> Promise[None] | None:
+    def register_command(self, command: str, handler: Callable[[ExecuteCommandParams], None]) -> None:
         """
         Intercept a command that is about to be sent to the language server.
 
@@ -453,7 +457,7 @@ class LspPlugin(APIHandler):
 
         :returns: Promise if *YOU* will handle this command plugin-side, None otherwise.
         """
-        pass
+        self.execute_commands[command] = handler
 
     def on_pre_send_request_async(self, request: ClientRequest, view: sublime.View | None) -> None:
         """
--- a/plugin/api.py
+++ b/plugin/api.py
@@ -186,7 +186,6 @@ class APIHandler:
                 self.handler_attr_map[method2attr(getattr(value, HANDLER_MARKER))] = name
             elif hasattr(value, COMMAND_HANDLER_MARKER):
                 self.command_handler_map[getattr(value, COMMAND_HANDLER_MARKER)] = name
-        print(f'self.command_handler_map: {self.command_handler_map}')
 
 
 def notification_handler(method: str) -> Callable[[Callable[[Any, P], None]], Callable[[Any, P], None]]:
--- a/plugin/api.py
+++ b/plugin/api.py
@@ -434,11 +434,12 @@ class LspPlugin(APIHandler):
         super().__init__()
         self.weaksession: ref[Session] = weaksession
         self.execute_commands: dict[str, Callable[[ExecuteCommandParams], Promise[Any]]]
+        self.workspace_configuration_handler: Callable[[ConfigurationItem, Any], Any]
 
     def on_start_async(self) -> None:
         pass
 
-    def on_workspace_configuration(self, params: ConfigurationItem, configuration: Any) -> Any:
+    def on_workspace_configuration(self, handler: Callable[[ConfigurationItem, Any], Any]) -> None:
         """
         Override to augment configuration returned for the workspace/configuration request.
 
@@ -447,7 +448,7 @@ class LspPlugin(APIHandler):
 
         :returns: The resolved configuration for given params.
         """
-        return configuration
+        self.workspace_configuration_handler = handler
 
     def register_command(self, command: str, handler: Callable[[ExecuteCommandParams], Promise[Any]]) -> None:
         """
--- a/plugin/api.py
+++ b/plugin/api.py
@@ -433,7 +433,7 @@ class LspPlugin(APIHandler):
         """
         super().__init__()
         self.weaksession: ref[Session] = weaksession
-        self.execute_commands: dict[str, Callable[[ExecuteCommandParams], None]]
+        self.execute_commands: dict[str, Callable[[ExecuteCommandParams], Promise[Any]]]
 
     def on_start_async(self) -> None:
         pass
@@ -449,7 +449,7 @@ class LspPlugin(APIHandler):
         """
         return configuration
 
-    def register_command(self, command: str, handler: Callable[[ExecuteCommandParams], None]) -> None:
+    def register_command(self, command: str, handler: Callable[[ExecuteCommandParams], Promise[Any]]) -> None:
         """
         Intercept a command that is about to be sent to the language server.
 
--- a/plugin/api.py
+++ b/plugin/api.py
@@ -415,7 +415,6 @@ class LspPlugin(APIHandler):
         super().__init__()
         self.weaksession: ref[Session] = weaksession
         self.execute_commands: dict[str, Callable[[ExecuteCommandParams], Promise[Any]]]
-        self.workspace_configuration_handler: Callable[[ConfigurationItem, Any], Any]
 
     def on_start_async(self, context: OnStartContext) -> None:
         """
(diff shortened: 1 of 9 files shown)


rchl:
Example how LSP-ruff would look like with this API:

```py
@final
class RuffLsp(LspPlugin):
    uv_venv_manager: UvVenvManager | None = None

    @classmethod
    @override
    def is_applicable(cls, context: IsApplicableContext) -> bool:
        return bool(
            super().is_applicable(context)
            # REPL views (https://github.com/sublimelsp/LSP-pyright/issues/343)
            and not context.view.settings().get("repl")
        )

    @classmethod
    @override
    def on_before_start_async(cls, context: BeforeStartContext) -> None:
        if context.configuration.server_binary == 'auto':
            if not cls.uv_venv_manager:
                cls.uv_venv_manager = UvVenvManager(cls.name, 'pyproject.toml', cls.plugin_storage_path.parent)
            if cls.uv_venv_manager.needs_install_or_update():
                cls.uv_venv_manager.install()
            path = context.configuration.env.get('PATH', '')
            context.configuration.env['PATH'] = f'{cls.uv_venv_manager.venv_bin_path}{pathsep}{path}'
            context.variables['server_binary'] = str(cls.uv_venv_manager.venv_bin_path / 'ruff')
        else:
            context.variables['server_binary'] = context.configuration.server_binary
```

vs. in the original `LspPlugin`:

```py
@final
class RuffLsp(LspPlugin):
    uv_venv_manager: UvVenvManager | None = None

    @classmethod
    @override
    def is_applicable(cls, context: PluginContext) -> bool:
        return bool(
            super().is_applicable(context)
            # REPL views (https://github.com/sublimelsp/LSP-pyright/issues/343)
            and not context.view.settings().get("repl")
        )

    @classmethod
    @override
    def install_async(cls, context: PluginContext) -> None:
        if context.configuration.server_binary == 'auto':
            if not cls.uv_venv_manager:
                cls.uv_venv_manager = UvVenvManager(cls.name, 'pyproject.toml', cls.plugin_storage_path.parent)
            if cls.uv_venv_manager.needs_install_or_update():
                cls.uv_venv_manager.install()
            path = context.configuration.env.get('PATH', '')
            context.configuration.env['PATH'] = f'{cls.uv_venv_manager.venv_bin_path}{pathsep}{path}'

    @classmethod
    @override
    def additional_variables(cls, context: PluginContext) -> dict[str, str]:
        if context.configuration.server_binary == 'auto':
            if cls.uv_venv_manager:
                return {
                    'server_binary': str(cls.uv_venv_manager.venv_bin_path / 'ruff')
                }
        return {
            'server_binary': context.configuration.server_binary
        }
```

Note that `LSP-ruff` is one of the simplest. More complex plugins would benefit more.

plugin/api.py
```diff
@@ -477,6 +433,10 @@ def __init__(self, weaksession: ref[Session]) -> None:
         """
         super().__init__()
         self.weaksession: ref[Session] = weaksession
+        self.execute_commands: dict[str, Callable[[ExecuteCommandParams], None]]
+
+    def on_start_async(self) -> None:
+        pass
```
rchl: It's not strictly necessary because one can override `__init__` but overriding it is annoying as it requires importing types that you might not need otherwise.

And also who likes overriding `__init__` and remembering to call `super()`?

predragnikolic:
We could remove super().is_applicable(context) in is_applicable

When code actions on format were done, we did the same. https://github.com/sublimelsp/LSP/pull/2747#discussion_r2765510982

before:
```py
    @classmethod
    @override
    def is_applicable(cls, context: IsApplicableContext) -> bool:
        return bool(
            super().is_applicable(context)
            # REPL views (https://github.com/sublimelsp/LSP-pyright/issues/343)
            and not context.view.settings().get("repl")
        )
```

after:
```py
    @classmethod
    @override
    def is_applicable(cls, context: IsApplicableContext) -> bool:
        # REPL views (https://github.com/sublimelsp/LSP-pyright/issues/343)
        return not context.view.settings().get("repl")
```

predragnikolic:
can we remove classmethod  on_before_start_async and use self?

```
    @override
    def on_before_start_async(self, context: BeforeStartContext) -> None:
        if context.configuration.server_binary == 'auto':
            if not self.uv_venv_manager:
                self.uv_venv_manager = UvVenvManager(self.name, 'pyproject.toml', self.plugin_storage_path.parent)
            if self.uv_venv_manager.needs_install_or_update():
                self.uv_venv_manager.install()
            path = context.configuration.env.get('PATH', '')
            context.configuration.env['PATH'] = f'{self.uv_venv_manager.venv_bin_path}{pathsep}{path}'
            context.variables['server_binary'] = str(self.uv_venv_manager.venv_bin_path / 'ruff')
        else:
            context.variables['server_binary'] = context.configuration.server_binary
```

ANSWER: 
https://github.com/sublimelsp/LSP/pull/2864#issuecomment-4243718873

predragnikolic:
the _async suffix

the _async suffix turned out to be helpful in LSP,
to know on what thread we are.

Do you think that we can get away without the _async suffix in the LspPlugin API?

Reasons to have the suffix:
It can give more context 
The ST API also has a similar convention (ST API has on_new and on_new_async)

Reasons to not have it:
on_before_start looks a bit nicer and shorter than on_before_start_async (:

rchl:
> can we remove classmethod on_before_start_async and use self?

The plugin instance would have to be instantiated before it's called then.
Currently we do that only after initialize request completes (Session and Transport is already created). I don't think we want to initialize plugin instances without those initially.

rchl:
> We could remove super().is_applicable(context) in is_applicable

The base implementation in LspPlugin has logic that we want to run: https://github.com/sublimelsp/LSP/blob/231f9a3aef943d66cfddc3731cfda89c7dc2a273/plugin/api.py#L362-L380

and we wouldn't want all packages repeat it as it would be error-prone

rchl:
> Do you think that we can get away without the _async suffix in the LspPlugin API?

I think it's an useful reminder that we are on async thread. But if all methods are basically called on async thread then maybe it would be enough to document it. Not sure.

plugin/core/sessions.py
```diff
@@ -1319,8 +1320,10 @@ def execute_command(
         """Run a command from any thread. Your .then() continuations will run in Sublime's worker thread."""
         if self._plugin:
             if isinstance(self._plugin, LspPlugin):
-                if promise := self._plugin.on_execute_command(command):
-                    return promise.then(lambda _: None)
+                command_name = command['command']
+                if handler := self._plugin.execute_commands.get(command_name):
+                    handler(command)
+                    return Promise.resolve(None)
```
rchl: It would be more correct to still have the handler return a promise.

rchl:
I think this is better approach than the original one.

If someone has something against it then please speak up now and not after I invest a ton more time into it. :)

This approach is perhaps less friendly to beginners since it's more like coding and less like just putting blocks in place but it does allow for more flexibility, is more future proof and allows to avoid a ton of redundant code.

plugin/api.py
```diff
@@ -244,15 +244,28 @@ def wrapper(self: Any, params: P, request_id: int) -> Promise[Response[Any]]:
...
     configuration: ClientConfig
     """The resolved `ClientConfig` for this session."""
     view: sublime.View
+    """Template variables to be substituted in `command`, `env`, and `initialization_options`."""
+    workspace_folders: list[WorkspaceFolder]
+    """The workspace folders active for this session."""
+
+
+@dataclass
+class BeforeStartContext:
```
jwortmann: This name is a bit inconsistent with the `IsApplicableContext`. To be consistent, shouldn't it be `OnBeforeStartContext`?

plugin/api.py
jwortmann: It feels like this name could be a bit misleading, because "on_start" sounds like this is done on server starting and before the initialize request. But this method is actually called after the initialze response. So maybe it would be better to rename as `on_after_initialize`?

jwortmann: In general, I wonder if we shouldn't wait for `__init__` to be after the initialize request, but create the plugin instance even before that request?

That would allow to use `on_pre_send_request_async` to modify some of the initialize params (not sure if that could be useful for some plugins that might provide custom implementations with enhanced features - e.g. iirc Visual Studio uses some additional SymbolKinds that are not defined in the specs. Not sure if it checks the client capabilities for that.)

I assume the idea behind calling `__init__` after the initialize response was that the server might send an error response on initialize. But how realistic is that in practice? I guess if it's configured incorrectly and cannot start, we already get an exception before when trying to establish the connection?

plugin/api.py
```diff
@@ -477,8 +433,13 @@ def __init__(self, weaksession: ref[Session]) -> None:
...
         super().__init__()
         self.weaksession: ref[Session] = weaksession
+        self.execute_commands: dict[str, Callable[[ExecuteCommandParams], Promise[Any]]]
+        self.workspace_configuration_handler: Callable[[ConfigurationItem, Any], Any]
 
-    def on_workspace_configuration(self, params: ConfigurationItem, configuration: Any) -> Any:
+    def on_start_async(self) -> None:
+        pass
+
+    def on_workspace_configuration(self, handler: Callable[[ConfigurationItem, Any], Any]) -> None:
```
jwortmann: This looks pretty confusing, does this handler function replace the implementation for `workspace/configuration` that we have in LSP? Or what does this function do now? Because the return type is `None` now.

plugin/api.py
```diff
@@ -487,17 +448,17 @@ def on_workspace_configuration(self, params: ConfigurationItem, configuration: A
 
         :returns: The resolved configuration for given params.
         """
-        return configuration
+        self.workspace_configuration_handler = handler
 
-    def on_execute_command(self, command: ExecuteCommandParams) -> Promise[None] | None:
+    def register_command(self, command: str, handler: Callable[[ExecuteCommandParams], Promise[Any]]) -> None:
```
jwortmann: I kind of like this strategy to register for commands in an explicit way.

Alternatively we could think about using a decorator, similar to the `@request_handler` / `@notification_handler`.
Then we would have something like
```python
@command_handler('tinymist.runCodeLens')
def on_code_lens(self, arguments: LSPAny) -> None:
    # ...
```

Do we also need the ability to unregister commands?

Keep in mind that in VSCode it works this way that extension can explicitly register to handle commands, because afaik there is not real distinction between LSP commands (from `workspace/executeCommand`) and from built-in VSCode commands (something like `editor.action.triggerSuggest`). That is why servers often use some commands like "showReferences" in code lenses, even though the server doesn't handle those itself via `workspace/executeCommand`.

plugin/api.py
rchl: It's to modify the configuration before it's returned to the server.

Now that you said it, perhaps we should have more generic hook for hooking any client response? Though then I'd want it to be type-friendly also which could be tricky.

plugin/api.py
jwortmann: Btw, and now where there isn't any more unified `PluginContext` for various methods, I still somehow think that it would be easier to pass the view and ClientConfig directly to `is_applicable`...

If we in the future for example support to specify a DocumentSelector (instead or in addition to our "selector" from ST) in the server config, then the ClientConfig class will need to be updated anyway. And I don't know what else you imagine could possibly be needed to pass into `is_applicable` as a new parameter.

plugin/api.py
rchl: Don't hate the idea of decorator. I guess the downside could be that you couldn't bind some extra arguments to the handler (just thinking aloud, I don't remember if it's something that would be beneficial for any existing case).

Also, we don't have to pick just one - we can have both decorators and this listener approach. In that case it would probably also make sense to add corresponding listener approach for notifications and requests.

Don't think we would need unregistering. But good thing about that approach is that we could add "unregister_command" later if needed without breaking anything.

plugin/api.py
jwortmann: Yeah, iirc we talked about that some time before. I think the only method that we currently miss in the API is a `on_pre_send_response`.

I think I would prefer having such a generic method, over this special one for on_workspace_configuration.

---

Although there are not many different server request methods I think. There are the "refresh" requests from diagnostics / inlay hints / code lenses, but they only return `null` in the response, so they shouldn't be very relevant for plugins.

To be useful, it would probably require to also include the original params from the server request into that `on_pre_send_response` (if we want to add such a method).

plugin/api.py
rchl: We can change it. I've touched on that in https://github.com/sublimelsp/LSP/pull/2739#discussion_r2990321766

So then we could have those hooks:
- (static) on_before_start_async()
- ~~(static) on_before_initialize_async()~~ - remove
- on_start_async() - before initialize
- on_after_initialize_async() - after initialize

plugin/api.py
jwortmann: > Also, we don't have to pick just one - we can have both decorators and this listener approach. In that case it would probably also make sense to add corresponding listener approach for notifications and requests.

We already have that for notifications and requests. It is the `m_eslint_somerequest` methods defined in the plugin. To my understanding the decorators were introduces to hide that as an "internal" thing and allow a more convenient implementation / naming in the plugin.

In the case of command handlers I would maybe do the same. So only expose this decorator, and then do the registration in some internal method `self._register_command` (or probably not even needed, because it can be done in the decorator implementation itself). What do you think?

Although then it would be a bit asymmetrical if we also want to add an "unregister_command" method later.

plugin/api.py
rchl: With the old API there were cases where I wanted to add extra arguments but couldn't.
I'd rather have consistent convention with passing just a single context that can be extended in non-breaking way in the future without breaking compatibility.

And having dedicated contexts for each method is also on purpose as otherwise we could end up with an issue if we want to extend one but not the other.

plugin/api.py
rchl: > Although there are not many different server request methods I think.

https://github.com/sublimelsp/LSP/blob/999e07277b56c3d7238f7ade42352a858f8f90e2/plugin/core/protocol.py#L1277-L1293

> To be useful, it would probably require to also include the original params from the server request into that `on_pre_send_response` (if we want to add such a method).

I could do:

```py
def on_pre_send_response(self, request: ClientResponse, result: Any) -> Promise[Any]:
```

but the result wouldn't be typed properly here so here is also where it would be beneficial to have listener-style registration instead. Then there is no problems with types because one can just manually chose some.

```py
self.register_on_pre_send_response('workspace/configuration', self._handle_workspace_configuration)

# ...

def _handle_workspace_configuration(self, request: ClientResponse, result: List[LSPAny]) -> List[LSPAny]:
```

(not the best example because `workspace/configuration` result is not that strictly typed but the point is there)

plugin/api.py
rchl: `m_eslint_somerequest` doesn't quite compare to a registration-style where one can pick the method name to be anything:

```py
self.register_request_handler('eslint/somerequest', self._handle_eslint_request)
```

And actually I should make the `m_eslint_somerequest` style not work with `LspPlugin` since the idea was to phase it out with `AbstractPlugin`.

rwols:
> If someone has something against it then please speak up now and not after I invest a ton more time into it. :)

I haven’t read the whole thread. Neither have I read all of the 100+ comments on the older LspPlugin proposal. At the risk of coming across like an ignorant fool, I would like to propose we also incorporate async as a first class citizen in this new API.

I would like to make the transport handling work with asyncio. Once that’s implemented we can introduce async in the LspPlugin as well. What do you think?

rchl:
Don't think this was brought up anywhere in those PRs but IMO that's completely out of scope.

Introducing async would probably mean months of work and refactoring the whole code base.

Also after many discussions around async on Discord I have impression that it's not really ready to be more widely used. I think it would be beneficial for you @jwortmann to also have the same context and see the same discussions but not sure how to do that if are refusing to touch Discord. ;)

If you can make transport handling work with asyncio then feel free to try and then we'll see what is required to expand it into more of the code base but I don't think it should be a blocking issue for adding LspPlugin.

Possibly we could even introduce async equivalents of existing functions (`on_before_start_async` -> `async on_before_start`) later to make transition possible without introducing new plugin class again.

plugin/api.py
jwortmann: > And actually I should make the `m_eslint_somerequest` style not work with `LspPlugin` since the idea was to phase it out with `AbstractPlugin`.

+1

> Don't hate the idea of decorator. I guess the downside could be that you couldn't bind some extra arguments to the handler

I don't really understand what exactly do you mean with "extra arguments" here? Like as a `partial` function like this?
```python
self.register_command('tinymist.runCodeLens', partial(self.on_code_lens, static_arg1, static_arg2))
```
Not sure what would be the purpose of that, but perhaps I'm misunderstanding the intention.

Also the signature of `handler: Callable[[ExecuteCommandParams], Promise[Any]]` seems wrong, it shouldn't actually return a promise anymore, no?

I think I would prefer to only have the decorator style, without exposing a `self.register_command` method in the plugin. Then every method with the decorator gets registered automatically, like the request and notification handlers.

rwols:
Agreed, let’s first see how things pan out with the transport handling.

plugin/api.py
rchl: > Also the signature of `handler: Callable[[ExecuteCommandParams], Promise[Any]]` seems wrong, it shouldn't actually return a promise anymore, no?

I think that ideally it should still be a promise so that the code that triggers execute_command doesn't proceed until handling of the command is done. In most cases we'll probably handle the command synchronously but it's not a guarantee.

plugin/api.py
rchl: Added in cdf24c3b35b60cc848c7839f39bb1908a63ee003

- It includes `method`, `params` and `response` so that it's possible to infer type based on checking `method`.
- The `response['result']` can be modified directly, nothing is returned from the handler
- The handler should ideally return a Promise but I think for consistency with other methods and because there are few methods here, it shouldn't really be needed.

plugin/api.py
rchl: Added command decorator: ca4a2d60093e22be2284e77734a6a0a9acdd698e

As for whether to return promise or nothing I found https://github.com/sublimelsp/LSP-Deno/blob/396e44f0cf64659044ad3369c40039d621e7bad4/plugin.py#L37-L49 that only calls the done_callback after handling the command asynchronously.

plugin/api.py
```diff
@@ -243,20 +247,59 @@ def wrapper(self: Any, params: P, request_id: int) -> Promise[Response[Any]]:
...
+        setattr(func, COMMAND_HANDLER_MARKER, command_name)
+        return func
+
+    return decorator
+
+
 @dataclass
-class PluginContext:
-    """Plugin context information passed to various `LspPlugin` classmethods."""
+class IsApplicableContext:
```
rchl, line 276: Maybe `Context` should be the prefix for all contexts for more consistency and so that it can't be confused with a "condition" of sort...

plugin/api.py
jwortmann: Nice. Makes sense to keep the promise in order to retain current behavior. This `done_callback` from current API is also used to resolve a promise, which triggers the next `then` if there is one chained to `Session.execute_command`. This is the case e.g. for code lenses, which use the `lsp_execute` command
https://github.com/sublimelsp/LSP/blob/cf67841783fb63298133f5c6e92d3341ee35cd67/plugin/session_view.py#L443
and that command prints a message to the status bar when the promise gets resolved:
https://github.com/sublimelsp/LSP/blob/cf67841783fb63298133f5c6e92d3341ee35cd67/plugin/execute_command.py#L45-L54

In my opinion this status message is kind of useless, and more of a debug message that shouldn't really be visible to users. Maybe we could remove that status message for the default implementation?

---

I think it would be good to let the decorator also add the registered command names to the `_supported_commands` in the SessionBuffer:
https://github.com/sublimelsp/LSP/blob/cf67841783fb63298133f5c6e92d3341ee35cd67/plugin/session_buffer.py#L1036-L1041

Currently, if the session has a plugin, all code lenses are drawn in the view, even if the command is neither handled by the server, nor implemented in the plugin. But with the command_handler decorator we could filter the commands that are actually implemented in the plugin. That would ensure that only code lenses which can actually be executed are visible.

So the code at
https://github.com/sublimelsp/LSP/blob/cf67841783fb63298133f5c6e92d3341ee35cd67/plugin/session_buffer.py#L551-L555
would also need to add the command names which are registerd via that decorator. And then you could remove the if-condition from above.

plugin/api.py
```diff
@@ -243,20 +247,59 @@ def wrapper(self: Any, params: P, request_id: int) -> Promise[Response[Any]]:
...
+    """
+    Decorator to mark a method as a handler for a specific server command.
+
+    When the language server sends a `workspace/executeCommand` request with the given
+    command name, the decorated method is called with the full command payload.
+
+    Usage:
+        ```py
+        @command_handler('editor.action.triggerSuggest')
+        def on_trigger_suggest(self, params: ExecuteCommandParams) -> Promise[None]:
```
jwortmann, line 260: I would expect that only the `params.get('arguments')` are passed to this method, because the command name is already known as specified in the decorator call. Then the generic signature of that handler would be `(self, arguments: list[LSPAny] | None) -> Promise[None]`.

rchl: https://github.com/sublimelsp/LSP/pull/2864/commits/904ccb0d97a25dc78eae290e87f20a31153b5d2c

At the same time noticed that the `plugin/api.py` has a ton of duplicate imports. I'd wish ruff would catch that. Also it doesn't always suggest to move import to type-checking block for some reason.

plugin/api.py
rchl: > In my opinion this status message is kind of useless, and more of a debug message that shouldn't really be visible to users. Maybe we could remove that status message for the default implementation?

Personally I never see any status messages unless I know they are gonna show up there and looking specifically for them. So I don't mind removing such messages.

427483a5dc99fb8f419256bb725e914082dfdfb3

plugin/api.py
rchl: Implemented my suggested methods above and initializing plugin before `initialize` now - 088f5653f14bebb1fc33967de28995e0ae0bc502

plugin/api.py
rchl: Also renamed contexts so that the `"Context"` is the prefix.

plugin/api.py
rchl: > I think it would be good to let the decorator also add the registered command names to the `_supported_commands` in the SessionBuffer:

78c75c76a51c5bdcf31d4076e333dd1709369276

plugin/session_buffer.py
```diff
@@ -1034,14 +1037,9 @@ def resolve_visible_code_lenses_async(self, view: sublime.View) -> None:
         Promise.all(promises).then(lambda _: self._on_visible_code_lenses_resolved_async())
 
     def _filter_supported_code_lenses(self) -> list[ResolvedCodeLens]:
-        code_lenses = self._code_lenses.code_lenses_with_command()
-        if self.session.uses_plugin():
-            # TODO: should plugins announce the commands that they can handle, so we can filter out the unsupported
-            # commands here as well?
-            return code_lenses
         supported_code_lenses: list[ResolvedCodeLens] = []
         # Filter out CodeLenses with commands that are not handled directly by the language server
```
jwortmann:
```suggestion
        # Filter out CodeLenses with commands that are not handled by the language server or plugin
```

plugin/api.py
jwortmann: The names sound a bit weird now, but it allows nicer import sorting, so I guess it's okay.

jwortmann:
Merge into the other PR and continue there?

PR #2866, opened 2026-04-13
Test both TCP modes

rwols:
Add a test that checks that "tcp server mode" works. Server meaning that this plugin acts as the TCP server and the langserver connects as TCP client.

### DIFF
--- a/tests/server.py
+++ b/tests/server.py
@@ -21,18 +21,16 @@ resumes (since response to request will arrive after requested notification).
 from __future__ import annotations
 
 from argparse import ArgumentParser
-import argparse
-from enum import IntEnum
-import enum
 from typing import Any
 from typing import Awaitable
 from typing import Callable
 from typing import Dict
 from typing import Iterable
 from typing import List
-from typing import Literal
 from typing import Union
+import argparse
 import asyncio
+import enum
 import json
 import os
 import sys
@@ -48,7 +46,7 @@ PayloadLike = Union[List[StringDict], StringDict, None]
 ENCODING = "utf-8"
 
 
-class ErrorCode(IntEnum):
+class ErrorCode(enum.IntEnum):
     # Defined by JSON RPC
     ParseError = -32700
     InvalidRequest = -32600
@@ -480,12 +478,9 @@ def _win32_stdio(loop: asyncio.AbstractEventLoop) -> tuple[asyncio.StreamReader,
 # END: https://stackoverflow.com/a/52702646/990142
 
 
-class Mode(enum.Enum):
+class Mode(enum.StrEnum):
     server = "server"
     client = "client"
-
-    def __str__(self) -> str:
-        return self.value
 
 
 async def main(tcp_port: int | None = None, mode: Mode = Mode.server) -> bool:
@@ -530,7 +525,7 @@ if __name__ == '__main__':
     parser.add_argument("-v", "--version", action="store_true", help="print version and exit")
     parser.add_argument("-p", "--tcp-port", type=int)
     parser.add_argument("--mode", type=Mode, default=Mode.server)
-    args: CmdLineArgs = parser.parse_args(namespace=CmdLineArgs())
+    args = parser.parse_args(namespace=CmdLineArgs())
     if args.version:
         print(__package__, __version__)
         sys.exit(0)
--- a/tests/server.py
+++ b/tests/server.py
@@ -1,8 +1,7 @@
 """
 A simple test server for integration tests.
 
-Only understands stdio.
-Uses the asyncio module and mypy types, so you'll need a modern Python.
+Can do JSON-RPC with stdio or TCP sockets as the transport.
 
 To make this server reply to requests, send the $test/setResponse notification.
 
@@ -18,7 +17,6 @@ with expected notification method in params['method'] and params in params['para
 Tests can await this request to make sure that they receive notification before code
 resumes (since response to request will arrive after requested notification).
 
-TODO: Untested on Windows.
 """
 from __future__ import annotations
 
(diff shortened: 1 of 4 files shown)


tests/server.py
```diff
@@ -479,24 +482,36 @@ def do_blocking_drain() -> None:
 # END: https://stackoverflow.com/a/52702646/990142
 
 
-async def main(tcp_port: int | None = None) -> bool:
+async def main(tcp_port: int | None = None, mode: str | None = None) -> bool:
```
rchl, line 485: Can you set type for `mode` to `Literal['client', 'server'] | None` so that it self-documents allowed values?

tests/server.py
```diff
@@ -479,24 +482,36 @@ def do_blocking_drain() -> None:
...
-                session = Session(reader, writer)
-                self.received_shutdown = await session.run_forever()
+            callback = ClientConnectedCallback()
+            server = await asyncio.start_server(callback, port=tcp_port)
+            # NOTE: This is deliberately wrong -- we should stop serving once the exit notification is received. But,
+            # it's good to have this botched logic here to make sure that servers shutdown in the integration tests.
+            await server.serve_forever()
+            return callback.received_shutdown
+
+        if mode is not None and mode == "client":
```
rchl, line 507:
```suggestion
        if mode == "client":
```

tests/server.py
```diff
@@ -506,6 +521,7 @@ async def __call__(self, reader: asyncio.StreamReader, writer: asyncio.StreamWri
     parser = ArgumentParser(prog=__package__, description=__doc__)
     parser.add_argument("-v", "--version", action="store_true", help="print version and exit")
     parser.add_argument("-p", "--tcp-port", type=int)
+    parser.add_argument("--mode", help="one of 'client' or 'server'", default="server")
```
rchl, line 524: Can validate the argument early and then also doesn't really need a redundant help:

```suggestion
    parser.add_argument("--mode", help="one of 'client' or 'server'", default="server", choices=["client", "server"])
```

tests/setup.py
```diff
@@ -44,23 +41,42 @@ def result(self) -> Any:
...
+def make_stdio_test_config(name: str, init_options: dict[str, Any]) -> ClientConfig:
+    """Start the fake language server in STDIO mode."""
+    config = ClientConfig(
+        name=name,
         command=["python3", join("$packages", "LSP", "tests", "server.py")],
         selector="text.plain",
         enabled=True,
     )
+    config.initialization_options.assign(init_options)
+    return config
```
rchl, line 53: Why not pass init_options to the constructor and save lines?

```suggestion
    return ClientConfig(
        name=name,
        command=["python3", join("$packages", "LSP", "tests", "server.py")],
        selector="text.plain",
        initialization_options=DottedDict(init_options),
        enabled=True,
    )
```

Same below.

tests/setup.py
```diff
@@ -44,23 +41,42 @@ def result(self) -> Any:
         return self.__result
 
 
-def make_stdio_test_config() -> ClientConfig:
-    return ClientConfig(
-        name="TEST",
+def make_stdio_test_config(name: str, init_options: dict[str, Any]) -> ClientConfig:
```
rchl, line 44: Maybe make `init_options` optional since it's not always needed?

```suggestion
def make_stdio_test_config(name: str, init_options: dict[str, Any] | None = None) -> ClientConfig:
```

tests/setup.py
```diff
@@ -44,23 +41,42 @@ def result(self) -> Any:
...
+    """Start the fake language server in STDIO mode."""
```
rchl, line 45: Nitpicking but the function doesn't really start the server. It just creates a config for starting a server.

tests/server.py
```diff
@@ -479,8 +485,22 @@ def do_blocking_drain() -> None:
...
 
 
-async def main(tcp_port: int | None = None) -> bool:
-    if tcp_port is not None:
+class Mode(enum.Enum):
+    server = "server"
+    client = "client"
+
+    def __str__(self) -> str:
+        return self.value
```
rchl: There exists a `StrEnum` which shouldn't require `__str__` implementation..

tests/server.py
```diff
@@ -493,28 +513,37 @@ async def __call__(self, reader: asyncio.StreamReader, writer: asyncio.StreamWri
...
+        version: bool = False
+        tcp_port: int | None = None
+        mode: Mode = Mode.server
+
     parser = ArgumentParser(prog=__package__, description=__doc__)
     parser.add_argument("-v", "--version", action="store_true", help="print version and exit")
     parser.add_argument("-p", "--tcp-port", type=int)
-    args = parser.parse_args()
+    parser.add_argument("--mode", type=Mode, default=Mode.server)
+    args: CmdLineArgs = parser.parse_args(namespace=CmdLineArgs())
```
rchl: No need for explicitly typing the `args`.

```suggestion
    args = parser.parse_args(namespace=CmdLineArgs())
```

rchl (commented):
Can you resolve conflicts and fix import linting issues?

rchl (approved):
StrEnum is not available in 3.8 and our pyright config specifies 3.8 runtime so `enum.StrEnum` is an error.

That said, CI doesn't check tests/* so it's only visible in editor when looking at the file.

That's fine for now but ideally we'd move server.py into its own directory and apply different runtime version to it.

rchl:
Note that I've tried that but defining `executionEnvironments` and overriding `pythonVersion` doesn't seem to work.

Previously I've reported https://github.com/microsoft/pyright/issues/11298 which is kind of an opposite issue but I think it's the same problem.

So doesn't seem like we can fix that for now - it's fine to merge as is.

rwols:
Yeah. It's just a fake test server for integration tests. Not a big deal to see an error in that file to me. I welcome the effort to make the environment for that file more precise.

PR #2875, opened 2026-04-18
Include diagnostic code when copying diagnostic text

rchl:
We have not been copying diagnostic code when copying diagnostic text to clipboard which was inconsistent with text in hover popup and also annoying if one wants to lookup the code.

Before:

```
Could not assign item in TypedDict
  "Literal['jsonc']" is not assignable to "LanguageKind" (basedpyright)
```

After:

```
Could not assign item in TypedDict
  "Literal['jsonc']" is not assignable to "LanguageKind" (basedpyright[reportGeneralTypeIssues])
```

The formatting doesn't exactly match style in the hover popup but it's because we don't have colors in plain text so we can't just include `source` without parenthesis.

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -917,12 +917,12 @@ def format_diagnostic_for_html(
     content = _format_diagnostic_message(view, message)
     code = diagnostic.get("code")
     source = diagnostic.get("source")
-    copy_text = raw_message.replace(' ', ' ')
+    copy_text = f'{raw_message} '.replace(' ', ' ')
     if source or code is not None:
         meta_info = ""
         if source:
             meta_info += text2html(source)
-            copy_text += f' ({source})' if code is None else f' {source}'
+            copy_text += f'({source})' if code is None else source
         if code is not None:
             if code_description := diagnostic.get("codeDescription"):
                 href = code_description["href"]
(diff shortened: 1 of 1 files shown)


rchl:
Feel free to suggest improvements to the formatting.

plugin/core/views.py
```diff
@@ -917,18 +917,21 @@ def format_diagnostic_for_html(
     content = _format_diagnostic_message(view, message)
     code = diagnostic.get("code")
     source = diagnostic.get("source")
+    copy_text = raw_message.replace(' ', ' ')
     if source or code is not None:
         meta_info = ""
         if source:
+            copy_text += f' ({source})'
             meta_info += text2html(source)
```
rchl:
```suggestion
            meta_info += text2html(source)
            copy_text += f' ({source})'
```

rchl:
I think a better formatting would be:

```
Could not assign item in TypedDict
  "Literal['jsonc']" is not assignable to "LanguageKind" basedpyright(reportGeneralTypeIssues - https://docs.basedpyright.com/v1.39.2/configuration/config-files/#reportGeneralTypeIssues)
```

when there is `code` and when there isn't then:

```
Could not assign item in TypedDict
  "Literal['jsonc']" is not assignable to "LanguageKind" (basedpyright)
```

jwortmann:
I would probably exclude the URL (codeDescription). You can click it in the popup to open the website, so I wouldn't find it useful to copy that too.

Maybe could use formatting like
```
(basedpyright[reportGeneralTypeIssues])
```
if there is `code`, and otherwise
```
(basedpyright)
```

jwortmann (approved):

PR #2883, opened 2026-04-24
Adjust font size of code lens annotations

jwortmann:
Code lenses used "small" font size, which is fine for phantoms, but looks bad in annotations if there are multiple code lenses on a single line, because in that case ST adds a linebreak symbol at the end of the annotation, and then the annotation height doesn't match (see screenshot). Also it looks inconsistent with code action and diagnostic annotations, so let's use regular font size for code lens annotations too.

Before (2x zoom):

before

After:

after

(ignore slightly different annotation colors in the screenshots, which happened because I switched color scheme while the annotation was visible)

### DIFF
--- a/plugin/session_view.py
+++ b/plugin/session_view.py
@@ -407,7 +407,10 @@ class SessionView:
             key = self._code_lens_key()
             regions = [self._code_lens_region(code_lens) for code_lens in self._code_lenses]
             flags = sublime.RegionFlags.NO_UNDO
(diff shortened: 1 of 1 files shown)


rchl:
On mac (looks better to me):

## Before

Screenshot 2026-04-25 at 16 09 44

## After

Screenshot 2026-04-25 at 16 10 32

rchl (approved):

Issue #2625, opened 2025-09-02
Settings can get mangled due to `sublime.expand_variables` use


rchl:
**Describe the bug**

Backslashes in server settings can get mangled due to use of `sublime.expand_variables` which replaces double backslashes with a single one. See https://github.com/sublimelsp/LSP-astro/pull/122

PR #2884, opened 2026-04-27
Add support for SnippetTextEdit

jwortmann:
Supersedes #2426

---

Example:

https://github.com/user-attachments/assets/0dd307a9-fbeb-48d0-9553-31f72976327a

```java
public class Main {
    public class SuperType {
        public SuperType(String name) {}
    }
    public class SubType extends SuperType {
    }
}
```

Here the inserted SnippetTextEdit has 3 tab stops, two of them with a placeholder text and the last one moving the caret to the end of the inserted text (actually that one is implicitly added, because the snippet text doesn't contain `${0}`).

It can be tested with LSP-jdtls, and that is the only server I found which already supports SnippetTextEdit at the moment.
This requires additional setup:

Under https://github.com/sublimelsp/LSP/blob/9105fb55f50a247b4e9ca537dccdf98257ca9367/plugin/core/sessions.py#L528
you need to temporarily add the line
```
"resourceOperations": ["create", "rename", "delete"],  # pyright: ignore[reportAssignmentType]
```
because otherwise jdtls doesn't want to emit edits in the SnippetTextEdit format.

Also you need to create an override for the `LSP-jdtls/modules/jdtls.py` file, and adjust the `initializationOptions`, because LSP-jdtls ignores the initialization options from its own settings file. Then you need to add the line `"snippetEditSupport": True,` into the `"extendedClientCapabilities"` at https://github.com/sublimelsp/LSP-jdtls/blob/3866f45806d452d81617830062ac7fbd43f82d8b/modules/jdtls.py#L178-L202 and also change some of the other values to `True` (in case of the code example from above it is `"generateConstructorsPromptSupport": True`).
 
*Edit:* Oh and I forgot to mention that LSP-jdtls is quite broken in general. To make it work with a recent JDK version you need to change the downloaded server version in `LSP-jdtls.sublime-settings`, like
```json
{
    "enabled": true,
    "version": "1.58.0",
    "settings": {
        "java.jdt.ls.java.home": "C:\\Program Files\\Java\\jdk-26.0.1"
    },
}
```

---

I think we should probably also remove the custom implementation for `process_placeholders` in https://github.com/sublimelsp/LSP/blob/9105fb55f50a247b4e9ca537dccdf98257ca9367/plugin/edit.py#L99 that is specific to rust-analyzer, because SnippetTextEdit is now part of the official (3.18) LSP specs, so rust-analyzer should just switch to that instead (also I believe there are some bugs and inefficiencies in that implementation). But I haven't removed it in this PR yet.


plugin/core/sessions.py
```diff
@@ -1835,8 +1838,10 @@ def on_workspace_configuration(self, params: ConfigurationParams) -> Promise[lis
 
     @request_handler('workspace/applyEdit')
     def on_workspace_apply_edit(self, params: ApplyWorkspaceEditParams) -> Promise[ApplyWorkspaceEditResult]:
-        return self.apply_workspace_edit_async(params.get('edit', {}), label=params.get('label')) \
-            .then(lambda _: {"applied": True})
+        is_refactoring = params.get('metadata', {}).get('isRefactoring', False)
```
jwortmann, line 1869: This is a "drive-by" change to add support for [WorkspaceEditMetadata](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#workspaceEditMetadata) from the LSP 3.18 specs.

plugin/edit.py
```diff
@@ -96,6 +99,87 @@ def run(
...
+
+    def _get_region(self, edit: TextEdit | AnnotatedTextEdit | SnippetTextEdit) -> tuple[int, sublime.Region]:
+        range_ = edit['range']
+        start = parse_lsp_position(range_['start'])
+        end = parse_lsp_position(range_['end'])
+        region = sublime.Region(
+            self.view.text_point_utf16(*start, clamp_column=True),
+            self.view.text_point_utf16(*end, clamp_column=True)
+        )
+        return start[0], region
```
rchl, line 165: There is function for that - https://github.com/sublimelsp/LSP/blob/7bc9670cde17c1267b704895d90fe5bcdb7764d9/plugin/core/views.py#L202-L202

plugin/edit.py
```diff
@@ -260,7 +344,9 @@ def _get_relative_path(wm: WindowManager, file_path: str) -> str:
             reference_document.append(original_line)
             if scheme == "file" and line_content:
                 end_row, end_col_utf16 = parse_lsp_position(edit['range']['end'])
-                new_text_rows = edit['newText'].split('\n')
+                new_text = sublime.expand_variables(edit['snippet']['value'], {}) if is_snippet_text_edit(edit) else \
```
rchl, line 347: Can you explain why are we calling `expand_variables` on a snippet?
I would expect snippet to contain tab stops, not variables to expand.

jwortmann: `expand_variables` (with empty variables) automatically eliminates the placeholders. So this converts the SnippetTextEdit into the 'newText' from a regular TextEdit, which is needed if we don't want to insert it as a snippet (e.g. if not focused view, etc.).

plugin/edit.py
jwortmann: That one is less efficient because it converts to custom Point objects first, and it would unnecessarily access the `start` twice from the range dict (because we also need `start` separately).

I used range_to_region initially, but then decided to align with the current implementation in LspApplyDocumentEditCommand:
https://github.com/sublimelsp/LSP/blob/9a3c3944df6795434104c15cd2c660f415c305e1/plugin/edit.py#L143-L146
and
https://github.com/sublimelsp/LSP/blob/9a3c3944df6795434104c15cd2c660f415c305e1/plugin/edit.py#L158-L161

plugin/edit.py
jwortmann: It's basically a built-in and more powerful variant of what you re-invented with custom regex and string replacement at https://github.com/sublimelsp/LSP/blob/9a3c3944df6795434104c15cd2c660f415c305e1/plugin/edit.py#L174-L176

```
>>> sublime.expand_variables('Hello, ${1:name}!$0', {})
'Hello, name!'
```

rchl: Also more buggy: https://github.com/sublimelsp/LSP/issues/2625

rchl (approved):

plugin/edit.py
jwortmann: I think the only places where two consecutive backslashes are used in source code would be in LaTeX and also in strings with Windows paths. But I imagine that would be quite rare to happen in WorkspaceEdits / SnippetTextEdits. So I would say we should keep that for now, and if it turns out that this bug regularly happens in practice, we could maybe think about implementing a custom patched version of `sublime.expand_variables`.

PR #2887, opened 2026-04-29
disable plugins on repl views

rchl:
This check is repeated in a [bunch of packages](https://github.com/search?q=org%3Asublimelsp+%22settings%28%29.get%28%5C%22repl%5C%22%29%22&type=code). Lets simplify.

### DIFF
--- a/plugin/api.py
+++ b/plugin/api.py
@@ -441,7 +441,9 @@ class LspPlugin(APIHandler):
         """
         if (syntax := context.view.syntax()) and (selector := context.configuration.selector.strip()):
             scheme, _ = parse_uri(uri_from_view(context.view))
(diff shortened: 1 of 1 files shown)


jwortmann:
That condition as it is now would disable LSP for all non-REPL views....

jwortmann:
I think there should at least be a comment with the package name that adds this setting.

rchl:
> with the package name

hmm?

jwortmann (approved):

jwortmann:
@rchl After looking at it again, maybe a better place for this condition would be in
https://github.com/sublimelsp/LSP/blob/def3cf814b275c17f710306612138592efe7b764/plugin/documents.py#L116-L131

Then it would be more consistent because it would also apply to manual configs.

Do you want to do a follow up PR to move that condition there?

PR #2890, opened 2026-04-30
Add support to open URI schemes with server-provided content

jwortmann:
LSP specs: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#workspace_textDocumentContent

It looks like support for this was [recently added](https://github.com/dotnet/roslyn/pull/83119) into `roslyn-language-server` for the `roslyn-source-generated` URI scheme, but as far as I can tell it hasn't landed in a release yet. Or at least the language server installed with
```
dotnet tool install --global roslyn-language-server --prerelease
```
doesn't seem to register anything for that method. Therefore untested and draft PR for now.

*Edit:* According to https://www.nuget.org/packages/roslyn-language-server/#versions-body-tab the last release was 3 months ago, but the functionality was added only 3 weeks ago. So we'll have to wait for a new release first, if we want to test it with that server.

---

Sidenote:

Not sure whether this URI scheme (or in general server-provided content for custom URI schemes) is also relevant for links in hover popups, but perhaps we should refactor the code from
https://github.com/sublimelsp/LSP/blob/17b2e6776304cffc12472957706e9c94cbe611d4/plugin/hover.py#L314-L345
for links in hover popups to call `Session.open_uri_async` instead, in order to reduce the code duplication. Then some parts which are currently specific to links in hover popups would need to be moved into `Session.open_uri_async`, and it would also need to pass the `view` as an argument. Although there could be multiple sessions running for the view, with the hover content merged from the different hoverProviders, so it's not clear which session to pick, or how to delegate the links to the correct session. Maybe the URI opening could alternatively refactored into a free function, possibly with the session(s) as another argument.
But we could look into that later and do it in a separate PR. Maybe we could also encode the target for DocumentLinks directly into the URI (instead of using `quick-panel:DocumentLink` URI), similar to the encoded code actions in https://github.com/sublimelsp/LSP/blob/17b2e6776304cffc12472957706e9c94cbe611d4/plugin/core/url.py#L113-L114
or even handle that differently (maybe just take the first valid response instead of listing the links in a quick panel if there are links from multiple sessions).


jwortmann:
So it looks like a new version of the roslyn server was published, and I can see in the initialize response that it registers the `roslyn-source-generated` scheme for that method.

Unfortunately I have no idea how to make the server emit URIs with that scheme. Perhaps anyone who knows more about C# could help me with a simple example how to trigger this?

plugin/core/sessions.py
```diff
@@ -1558,14 +1572,45 @@ def continue_on_main_thread() -> None:
...
+        return view
+
+    def _on_text_document_content_async(
+        self, response: TextDocumentContentResult | Error, uri: DocumentUri, flags: sublime.NewFileFlags, group: int
+    ) -> Promise[sublime.View | None]:
+        if isinstance(response, Error):
+            return Promise.resolve(None)
+        title = urlparse(uri).path.split('/')[-1]
+        content = response['text'].replace('\r', '')
+        syntax = ''
```
jwortmann, line 1595: According to https://github.com/microsoft/language-server-protocol/pull/1994#issuecomment-2282703368 the client is responsible to determine the language/syntax for the buffer. The only information that we have are the URI and the content, but I see no generic way to determine it just from the URI if it is for example something like `generated-source:aca9a9c1-a993-4ae4-b5c0-1864392b4630`.

Another comment at https://github.com/microsoft/language-server-protocol/pull/1994#issuecomment-2284076913 suggests that the editor extension (in VSCode) sets the language ID for a given URI scheme. So we could add this as a new config key that maps URI schemes to syntax names (`Window.new_file` takes the syntax name as a string). It seems that using `scope:source.xxx` also works, but is somewhat buggy, because the syntax name in the bottom right corner is bugged then (https://github.com/sublimehq/sublime_text/issues/4449). But perhaps we could use [`sublime.find_syntax_by_scope`](https://www.sublimetext.com/docs/api_reference.html#sublime.find_syntax_by_scope) API for that. So maybe it should be a URI scheme -> scope mapping in the config instead.

I still think that it is a mistake in the spec design to do it that way, because the language server should easily know the proper language ID for the document content that itself has generated. So it doesn't make much sense to me why it was designed how it is, but frankly we can't do anything about that.

plugin/core/sessions.py
```diff
@@ -2027,6 +2072,21 @@ def _refresh_diagnostics(self) -> None:
...
 
+    @request_handler('workspace/textDocumentContent/refresh')
+    def on_workspace_text_document_content_refresh(self, params: TextDocumentContentRefreshParams) -> Promise[None]:
+        sublime.set_timeout_async(lambda: self._refresh_text_document_content_async(params['uri']))
+        return Promise.resolve(None)
+
+    def _refresh_text_document_content_async(self, uri: DocumentUri) -> None:
+        for view in self.window.views():
+            try:
+                if uri_from_view(view) == uri:
```
rchl, line 2087: Should we use `get_session_buffer_for_uri_async` instead?

It should be more efficient but would also behave differently with clones. With the current solution we'd trigger request for each clone, with `get_session_buffer_for_uri_async` we would only trigger for one view.

plugin/core/sessions.py
```diff
@@ -1558,14 +1572,45 @@ def continue_on_main_thread() -> None:
...
+        new_content = response['text'].replace('\r', '')
+        content_region = entire_content_region(view)
+        if new_content == view.substr(content_region):
+            return
+        selection_region = first_selection_region(view)
+        selection = view.sel()
+        selection.add(content_region)
+        with mutable(view):
+            view.run_command('insert', {'characters': new_content})
+        if selection_region is not None and selection_region.begin() < view.size():
```
rchl, line 1606: I wonder if this will work reliably from an async thread. I'm a little concerned about timing and potentially view becoming read only before insert has finished.

plugin/core/sessions.py
jwortmann: `get_session_buffer_for_uri_async` only works for SessionBuffers. We don't know whether the server should also attach to the new view with the generated content, and currently it probably won't because for now we don't set the syntax, so it will default to plain text syntax.

plugin/core/sessions.py
jwortmann: This is the same logic just copied from the panel logging function, so I assume that it works.

jwortmann: Or do you think the insert could take longer if there is a lot content for the generated view?
If it turns out that this is the case, we probably would need to add some delay into the `mutable` decorator.

rchl: In other places it runs on the main thread.
The solution would be to just run it from main thread.

jwortmann: Ok, I can try to put it on the main thread.

jwortmann: https://github.com/sublimelsp/LSP/pull/2890/changes/cb1f1f31cfe76ef6c87756566c9916efec37f6aa

rchl (approved):
Not testing this (I can't get dotnet to work on mac) but looks good to me.

jwortmann:
I have tried to test this using the repo from the link in https://github.com/neovim/neovim/pull/39592#issuecomment-4391171751, but I have no idea how to make the server create URIs with that `roslyn-source-generated` scheme.

But I think if anything from this PR doesn't work, we will find out sooner or later.

---

As mentioned above we might want to also add a config option (or some other strategy) to let LSP know the appropriate syntax for given URI schemes, but we can probably add that at some later time, when we have a practical use case for it or when more servers add support for this feature.

PR #2142, opened 2022-12-15
Fix "folder_exclude_patterns" handling with relative project path

rchl:
Handle relative project path.

PR #2891, opened 2026-05-02
Fix some cases in ST pattern to glob conversion

### LINKED ISSUE #2112: Ability to exclude paths from `lsp_goto_diagnostic`
When I invoke `lsp_goto_diagnostic` I want to see diagnostics of my own code only, but currently it also shows all the diagnostics from external libraries within the project.

It seems `folder_exclude_patterns` in the sublime-project doesn't effect this feature either.

I want to be able to exclude a bunch of paths from diagnostics checking and listing.

rchl:
Align `sublime_pattern_to_glob` to match the behavior of ST implementation. In https://github.com/sublimehq/sublime_text/issues/6906 I've raised some inconsistencies in ST documentation so I'm aligning with those findings. I'm assuming that the documentation will be clarified rather than implementation updated.

Resolves #2112

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -395,7 +395,10 @@ class WindowManager(Manager, WindowConfigChangeListener, ViewStatusHandler):
             return "matches a pattern in binary_file_patterns"
         if matches_pattern(path, settings.get("file_exclude_patterns")):
             return "matches a pattern in file_exclude_patterns"
-        patterns = [sublime_pattern_to_glob(pattern, True) for pattern in settings.get("folder_exclude_patterns") or []]
+        patterns = [
+            sublime_pattern_to_glob(pattern, is_directory_pattern=True)
+            for pattern in settings.get("folder_exclude_patterns") or []
+        ]
         if matches_pattern(path, patterns):
             return "matches a pattern in folder_exclude_patterns"
         if self._workspace.includes_excluded_path(path):
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -397,7 +397,7 @@ class WindowManager(Manager, WindowConfigChangeListener, ViewStatusHandler):
             return "matches a pattern in file_exclude_patterns"
         patterns = [
             sublime_pattern_to_glob(pattern, is_directory_pattern=True)
-            for pattern in settings.get("folder_exclude_patterns") or []
+            for pattern in settings.get("folder_exclude_patterns")
         ]
         if matches_pattern(path, patterns):
             return "matches a pattern in folder_exclude_patterns"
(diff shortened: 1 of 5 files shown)


plugin/core/windows.py
```diff
@@ -395,7 +395,10 @@ def should_ignore_diagnostics(self, uri: DocumentUri, configuration: ClientConfi
             return "matches a pattern in binary_file_patterns"
         if matches_pattern(path, settings.get("file_exclude_patterns")):
             return "matches a pattern in file_exclude_patterns"
-        patterns = [sublime_pattern_to_glob(pattern, True) for pattern in settings.get("folder_exclude_patterns") or []]
+        patterns = [
+            sublime_pattern_to_glob(pattern, is_directory_pattern=True)
+            for pattern in settings.get("folder_exclude_patterns") or []
```
jwortmann: I wonder what was the purpose of the `or []`?

That setting should always exist due to having default value in `Preferences.sublime-settings`.

jwortmann: Also in general I wonder why we do this Sublime-to-glob conversion only for `folder_exclude_patterns`. Shouldn't it also apply to the `binary_file_patterns` and `file_exclude_patterns`?

rchl: I've started looking at this due to this old PR that needs finishing - https://github.com/sublimelsp/LSP/pull/2142

I will do any semi-related fixes separately (in that PR most like).

There is a lot of weird stuff going on. We match patterns using `matches_pattern` which uses `fnmatch.fnmatch` which is neither fully compatible with ST's own patterns nor with globs. So it all seems a bit broken...

rchl: > Also in general I wonder why we do this Sublime-to-glob conversion only for `folder_exclude_patterns`. Shouldn't it also apply to the `binary_file_patterns` and `file_exclude_patterns`?

It was your change BTW: https://github.com/sublimelsp/LSP/pull/2113 :)

jwortmann: I really wonder why I put the `or []` there, while for project data I used `get('folder_exclude_patterns', [])` instead... for project data it makes sense to have a fallback value because it must be explicitly set in project settings file.

But I guess it can't be important, because otherwise I had probably put a comment there.

---

Interestingly I did some considerations about `binary_file_patterns` and `file_exclude_patterns` in that linked PR before:

> "binary_file_patterns" can probably be ignored, because there are no lines or columns in binary files, so language servers can't send any diagnostics for them.

> Yeah, "file_exclude_patterns" could need the same treatment like in the change here, but then "file_include_patterns" (and "folder_include_patterns") should ideally also be taken into account.

jwortmann (approved):

PR #2896, opened 2026-05-11
Return correct result for workspace/applyEdit request

jwortmann:
When handling a `workspace/applyEdit` request from the server, the client must return a [ApplyWorkspaceEditResult](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#applyWorkspaceEditResult).

Currently, we unconditionally return `"applied": true`, even when the WorkspaceEdit cannot be applied correctly (e.g. due to failed to open view for URI, or view version doesn't match a given version in the WorkspaceEdit):
https://github.com/sublimelsp/LSP/blob/ced62c991cd4bc28cd633b26414eae1f3cd6ca43/plugin/core/sessions.py#L1883

Apart from that, we declare
https://github.com/sublimelsp/LSP/blob/ced62c991cd4bc28cd633b26414eae1f3cd6ca43/plugin/core/sessions.py#L540
for handling WorkspaceEdits, which means that (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#failureHandlingKind)
```ts
	/**
	 * Applying the workspace change is simply aborted if one of the changes
	 * provided fails. All operations executed before the failing operation
	 * stay executed.
	 */
	export const Abort: FailureHandlingKind = 'abort';
```

But in reality, we simply continue with further changes even if one of the change failed:
https://github.com/sublimelsp/LSP/blob/ced62c991cd4bc28cd633b26414eae1f3cd6ca43/plugin/core/sessions.py#L1674-L1680

So the actual behavior isn't what we declared in the client capabilities as our failure handling strategy. Also I think this async implementation (via `Promise.all`) cannot work if support for resource operations (`create` / `rename`) should be implemented later at some point (#2706), because we cannot run a TextDocumentEdit before a file for the corresponding URI has been created or renamed. This means that if the [WorkspaceEdit](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspaceEdit) contains `documentChanges`, we have to apply them sequentially.

This PR should fix that, but only for WorkspaceEdits from the `workspace/applyEdit` request. Other WorkspaceEdits, for example from code actions or from a `textDocument/rename` request still use the previous (current) implementation. For those it would require some more refactorings, because they currently return a custom WorkspaceEditSummary which is used to display the number of changes and affected files in the UI (in case of rename).

---

Also this PR still ignores the `needsConfirmation` flag from the [ChangeAnnotation](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#changeAnnotation) of [AnnotatedTextEdit](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#annotatedTextEdit):
```ts
	/**
	 * A flag which indicates that user confirmation is needed
	 * before applying the change.
	 */
	needsConfirmation?: boolean;
```
If we want to support that, for example with a [`sublime.ok_cancel_dialog`](https://www.sublimetext.com/docs/api_reference.html#sublime.ok_cancel_dialog), then probably more considerations will be required to make the code wait for the user choice.

### DIFF
--- a/plugin/core/edit.py
+++ b/plugin/core/edit.py
@@ -1,8 +1,12 @@
 from __future__ import annotations
 
 from ...protocol import AnnotatedTextEdit
+from ...protocol import CreateFile
+from ...protocol import DeleteFile
 from ...protocol import Position
+from ...protocol import RenameFile
 from ...protocol import SnippetTextEdit
+from ...protocol import TextDocumentEdit
 from ...protocol import TextEdit
 from ...protocol import WorkspaceEdit
 from .logging import debug
@@ -30,7 +34,24 @@ class WorkspaceEditSummary(TypedDict):
     deleted_files: NotRequired[int]
 
 
-# TODO: change return type to TypeIs[SnippetTextEdit] when properly supported by Pyright
+def is_text_document_edit(
+    document_change: TextDocumentEdit | CreateFile | RenameFile | DeleteFile
+) -> TypeGuard[TextDocumentEdit]:
+    return 'edits' in document_change
+
+
+def is_create_file(document_change: TextDocumentEdit | CreateFile | RenameFile | DeleteFile) -> TypeGuard[CreateFile]:
+    return document_change.get('kind') == 'create'
+
+
+def is_rename_file(document_change: TextDocumentEdit | CreateFile | RenameFile | DeleteFile) -> TypeGuard[RenameFile]:
+    return document_change.get('kind') == 'rename'
+
+
+def is_delete_file(document_change: TextDocumentEdit | CreateFile | RenameFile | DeleteFile) -> TypeGuard[DeleteFile]:
+    return document_change.get('kind') == 'rename'
+
+
 def is_snippet_text_edit(edit: TextEdit | AnnotatedTextEdit | SnippetTextEdit) -> TypeGuard[SnippetTextEdit]:
     return 'snippet' in edit
 
@@ -92,11 +113,8 @@ def apply_text_edits(
                 'required_view_version': required_view_version,
             }
         )
-    else:
-        view.run_command(
-            'lsp_apply_text_document_edit',
-            {'edits': edits, 'version': required_view_version, 'label': label}
-        )
+    elif required_view_version is None or required_view_version == view.change_count():
+        view.run_command('lsp_apply_text_document_edit', {'edits': edits, 'label': label})
     # Resolving from the next message loop iteration guarantees that the edits have already been applied in the main
     # thread, and that we've received view changes in the asynchronous thread.
     return Promise(lambda resolve: sublime.set_timeout_async(lambda: resolve(view if view.is_valid() else None)))
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 from ...protocol import AnnotatedTextEdit
 from ...protocol import ApplyWorkspaceEditParams
 from ...protocol import ApplyWorkspaceEditResult
+from ...protocol import ChangeAnnotation
+from ...protocol import ChangeAnnotationIdentifier
 from ...protocol import ClientCapabilities
 from ...protocol import CodeAction
 from ...protocol import CodeActionKind
@@ -11,6 +13,8 @@ from ...protocol import Command
 from ...protocol import CompletionItemKind
 from ...protocol import CompletionItemTag
 from ...protocol import ConfigurationParams
+from ...protocol import CreateFile
+from ...protocol import DeleteFile
 from ...protocol import Diagnostic
 from ...protocol import DiagnosticOptions
 from ...protocol import DiagnosticServerCancellationData
@@ -48,6 +52,7 @@ from ...protocol import ProgressToken
 from ...protocol import PublishDiagnosticsParams
 from ...protocol import Range
 from ...protocol import RegistrationParams
+from ...protocol import RenameFile
 from ...protocol import SemanticTokenModifiers
 from ...protocol import SemanticTokenTypes
 from ...protocol import ShowDocumentParams
@@ -59,6 +64,7 @@ from ...protocol import SnippetTextEdit
 from ...protocol import SymbolKind
 from ...protocol import SymbolTag
 from ...protocol import TextDocumentClientCapabilities
+from ...protocol import TextDocumentEdit
 from ...protocol import TextDocumentSyncKind
 from ...protocol import TextEdit
 from ...protocol import TokenFormat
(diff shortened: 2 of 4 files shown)


plugin/core/edit.py
```diff
@@ -30,7 +34,24 @@ class WorkspaceEditSummary(TypedDict):
     deleted_files: NotRequired[int]
 
 
-# TODO: change return type to TypeIs[SnippetTextEdit] when properly supported by Pyright
```
jwortmann, line 33: I've read in some Pyright issue that (unlike TypeGuard) TypeIs is not supported for TypedDicts, because TypedDicts are "structural types" (not sure what that means exactly). So it will probably never be possible to use TypeIs for the LSP types, at least when using Pyright.

plugin/core/sessions.py
```diff
@@ -1635,6 +1645,109 @@ def _apply_code_action_async(
...
+                return Promise.resolve({
+                    'applied': False,
+                    'failureReason': failure_reason,
+                    'failedChange': index
+                })
+            return self._apply_document_changes_recursive_async(
+                document_changes, change_annotations, index + 1, label, auto_save)
+
+        try:
+            document_change = document_changes.pop(0)
```
jwortmann, line 1711: Note that this seems to cause the `documentChanges` of a WorkspaceEdit from a `workspace/applyEdit` request being shown as an empty array in the LSP log panel. But I would consider this as a bug in our logging functionality (I opened related issue #2897).

jwortmann:
I've tested with different code actions from LSP-julia, which don't contain a WorkspaceEdit but trigger server commands via `workspace/executeCommand`, which in turn trigger the `workspace/applyEdit` requests from the server.

(Although those WorkspaceEdits only contain a single TextDocumentEdit, and also obviously they are all applied successfully. I don't know how we could test a WorkspaceEdit that does fail in practice - but perhaps we can write a test for that.)

plugin/edit.py
```diff
@@ -109,14 +109,10 @@ def run(
...
         edit: sublime.Edit,
         edits: list[TextEdit | AnnotatedTextEdit | SnippetTextEdit],
-        version: int | None = None,
         label: str | None = None
     ) -> None:
         if not edits:
             return
-        if version is not None and version != self.view.change_count():
-            debug('ignoring edit due to non-matching document version')
-            return
```
jwortmann, line 119: I think it would be nice to keep this check here (because it is executed on the UI thread) to ensure that there can't be any race condition when applying this TextDocumentEdit while typing at the same time. But I don't think there is any good way to signalize from the command back to the calling method that the edit has failed, and we need that for stopping the execution and returning back the `failureReason`. Therefore I moved this check into the functions which call this command.

tests/test_server_requests.py
```diff
@@ -117,6 +118,91 @@ def test_m_workspace_applyEdit_with_nontrivial_promises(self) -> Generator:
                 self.assertEqual(view.substr(sublime.Region(0, view.size())), expected[i])
                 view.close()
 
+    @unittest.skip("TODO: fix to not automatically create a new view for nonexistent file URI")
```
jwortmann: I think this test should work after the change from #2898, and then this line can be removed.
(Perhaps need to adjust the URI for non-Windows OS or make this test Windows-specific)

plugin/core/sessions.py
```diff
@@ -1877,10 +1990,15 @@ def on_workspace_configuration(self, params: ConfigurationParams) -> Promise[lis
 
     @request_handler('workspace/applyEdit')
     def on_workspace_apply_edit(self, params: ApplyWorkspaceEditParams) -> Promise[ApplyWorkspaceEditResult]:
-        is_refactoring = params.get('metadata', {}).get('isRefactoring', False)
-        return self.apply_workspace_edit_async(
-            params['edit'], label=params.get('label'), is_refactoring=is_refactoring
-        ).then(lambda _: {'applied': True})
+        edit = params['edit']
+        label = params.get('label')
+        is_refactoring = metadata.get('isRefactoring', False) if (metadata := params.get('metadata')) else False
```
rwols, line 1995: These type of statements involving the `:=` operator where the value is used before the assignment are getting into "too clever" territory :) but I'll allow it.

Something more plain and dumb is more clear to read, IMO:

```py
metadata = params.get('metadata')
is_refactoring = metadata is not None and metadata.get('isRefactoring, False)
```

.

or even

```py
is_refactoring = params.get('metadata', {}).get('isRefactoring', False)
```

plugin/edit.py
rwols: I have some ideas for this once the `feat/asyncio` branch is merged in `main`. But they're still a bit  fuzzy.

plugin/core/sessions.py
rwols: Yeah, we can log it from the transports level probably. Although, the problem with logging the plain JSON is that you can't easily hide large payloads _while still printing which request/notification_ it is (and the request ID). Maybe there's a way to deal with that.

rwols (approved):

plugin/core/sessions.py
jwortmann: I initially used the
```python
is_refactoring = params.get('metadata', {}).get('isRefactoring', False)
```
when I added it recently, but the change I did here with the if-condition is a micro optimization that doesn't require to create an empty dict in most cases and then do the membership test (which involves calculating the hash value, I believe) for `'isRefactoring'`. So the new code should be a little bit more efficient. In general I quite like the `:=` operator and think it often results in more readable code. But okay, I'll try not to push it too far in the future :-)

plugin/edit.py
jwortmann: I hope with asyncio we can also rewrite the recursive function as a simple for-loop which `await`s each document change.

jwortmann:
As a sidenote, I think we had an unspoken convention over the past time that for non-trivial pull requests, once it is approved, the PR author is the one who merges it. I don't think I had anything particular to add here, but in general the author probably knows best when it is fully ready, and sometimes you want to look over it again later, or still have something in your mind that you want to add. I think then it's a bit easier to handle and you don't always need to mark every PR as a draft PR until it gets merged eventually.

plugin/edit.py
rwols: Indeed. But for now I kept the recursion in my branch.

plugin/core/edit.py
```diff
@@ -30,7 +34,24 @@ class WorkspaceEditSummary(TypedDict):
...
+def is_create_file(document_change: TextDocumentEdit | CreateFile | RenameFile | DeleteFile) -> TypeGuard[CreateFile]:
+    return document_change.get('kind') == 'create'
+
+
+def is_rename_file(document_change: TextDocumentEdit | CreateFile | RenameFile | DeleteFile) -> TypeGuard[RenameFile]:
+    return document_change.get('kind') == 'rename'
+
+
+def is_delete_file(document_change: TextDocumentEdit | CreateFile | RenameFile | DeleteFile) -> TypeGuard[DeleteFile]:
+    return document_change.get('kind') == 'rename'
```
jwortmann, line 52: Copy/paste mistake from above, that I see now.
Will do a fixup PR -> #2900

rwols:
> As a sidenote, I think we had an unspoken convention over the past time that for non-trivial pull requests, once it is approved, the PR author is the one who merges it.

OK, I'll keep this in mind. In this particular case, I was going to make changes related to the document changes, so I wanted to make sure this was merged before I got to that.

PR #2905, opened 2026-05-16
Refactor workspace edit handling

jwortmann:
Refactor `Session.apply_workspace_edit_async` which may be used by code actions and the rename feature, so that the changes in the WorkspaceEdit are also applied in a sequential way (and that the WorkspaceEdit is aborted as soon as one change fails). This should make it possible for us to add support for resource operations.

### DIFF
--- a/plugin/core/edit.py
+++ b/plugin/core/edit.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from ...protocol import AnnotatedTextEdit
+from ...protocol import ApplyWorkspaceEditResult
 from ...protocol import CreateFile
 from ...protocol import DeleteFile
 from ...protocol import Position
@@ -120,7 +121,14 @@ def apply_text_edits(
     return Promise(lambda resolve: sublime.set_timeout_async(lambda: resolve(view if view.is_valid() else None)))
 
 
-def show_summary_message(window: sublime.Window, summary: WorkspaceEditSummary) -> None:
-    message = f"Applied {summary['total_changes']} changes in {summary['edited_files']} files"
+def show_summary_message(
+    window: sublime.Window, result: ApplyWorkspaceEditResult, summary: WorkspaceEditSummary
+) -> None:
+    if result['applied']:
+        message = f"Applied {summary['total_changes']} changes in {summary['edited_files']} files"
+    else:
+        message = "Error while applying WorkspaceEdit"
+        if failure_reason := result.get('failureReason'):
+            message += f": {failure_reason}"
     # a 300ms timeout prevents "Detect indentation: ..." status message from overriding the summary status message
     sublime.set_timeout(lambda: window.status_message(message), 300)
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -97,13 +97,10 @@ from .constants import MARKO_MD_PARSER_VERSION
 from .constants import RequestFlags
 from .constants import SEMANTIC_TOKENS_MAP
 from .constants import SUPPORTED_DIAGNOSTIC_TAGS
-from .edit import apply_text_edits
 from .edit import is_create_file
 from .edit import is_delete_file
 from .edit import is_rename_file
 from .edit import is_text_document_edit
-from .edit import parse_workspace_edit
-from .edit import WorkspaceChanges
 from .edit import WorkspaceEditSummary
 from .file_watcher import DEFAULT_WATCH_KIND
 from .file_watcher import file_watcher_event_type_to_lsp_file_change_type
@@ -161,6 +158,7 @@ from abc import abstractmethod
 from enum import IntFlag
 from functools import lru_cache
 from functools import partial
+from operator import itemgetter
 from typing import Any
 from typing import Callable
 from typing import cast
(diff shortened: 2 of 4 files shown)


plugin/core/edit.py
```diff
@@ -120,7 +121,14 @@ def apply_text_edits(
...
-    message = f"Applied {summary['total_changes']} changes in {summary['edited_files']} files"
+def show_summary_message(
+    window: sublime.Window, result: ApplyWorkspaceEditResult, summary: WorkspaceEditSummary
+) -> None:
+    if result['applied']:
+        message = f"Applied {summary['total_changes']} changes in {summary['edited_files']} files"
+    else:
+        message = "Error while applying WorkspaceEdit"
+        if failure_reason := result.get('failureReason'):
+            message += f": {failure_reason}"
```
rchl, line 132: Potentially has a long error message here that would be hard to read in few seconds that it's shown.

I would opt in for adding it to the console also.

plugin/core/sessions.py
```diff
@@ -1698,6 +1696,7 @@ def apply_text_document_edit(
 
         def _continue(failure_reason: str | None) -> Promise[ApplyWorkspaceEditResult]:
             if failure_reason:
+                debug(f'Error while applying WorkspaceEdit: {failure_reason}')
```
rchl: If my earlier suggestion is applied then we shouldn't need this.

I've said something to that effect before but `debug` seems like something only useful during feature development because nobody should want to run with it enabled normally. Which means that pretty much no one will see this during normal use.

And if it's only useful during implementation than probably also doesn't need to be included in the final code.

plugin/core/sessions.py
```diff
@@ -1749,50 +1748,41 @@ def _continue(failure_reason: str | None) -> Promise[ApplyWorkspaceEditResult]:
...
+            if is_text_document_edit(document_change):
+                summary['total_changes'] += len(document_change['edits'])
+                summary['edited_files'] += 1
+            elif is_create_file(document_change):
+                summary['created_files'] += 1
+            elif is_rename_file(document_change):
+                summary['renamed_files'] += 1
+            elif is_delete_file(document_change):
+                summary['deleted_files'] += 1
+        return self.apply_document_changes_async(
```
rchl, line 1782: Is it not ideal that counts are determined before we know if those succedded or not? Or should we assume that counts don't matter if the operation failed anyway?

plugin/core/sessions.py
```diff
@@ -1749,50 +1748,41 @@ def _continue(failure_reason: str | None) -> Promise[ApplyWorkspaceEditResult]:
...
-
-        active_sheet = self.window.active_sheet()
-        selected_sheets = self.window.selected_sheets()
-        promises: list[Promise[None]] = []
-        auto_save = userprefs().refactoring_auto_save if is_refactoring else 'never'
+        if not (document_changes := edit.get('documentChanges', [])):
+            document_changes: list[TextDocumentEdit | CreateFile | RenameFile | DeleteFile] = [
+                cast('TextDocumentEdit', {'textDocument': {'uri': uri, 'version': None}, 'edits': edits})
+                for uri, edits in edit.get('changes', {}).items()
+            ]
```
rchl: Alternative syntax that IMO would look cleaner (due to no need for manual type annotation):
```suggestion
        document_changes = edit.get('documentChanges', [])
        if not document_changes:
            document_changes.extend([
                cast('TextDocumentEdit', {'textDocument': {'uri': uri, 'version': None}, 'edits': edits})
                for uri, edits in edit.get('changes', {}).items()
            ])
```

rchl (approved):

plugin/core/edit.py
jwortmann: The same message automatically gets logged to the console, but only if debug logging is enabled. I think that should be okay?

plugin/core/sessions.py
jwortmann: Hm, but then we could also remove "log_debug" from the user settings file (i.e. make it only a "hidden" setting), if it's not supposed to be used by end users...

plugin/core/sessions.py
jwortmann: Yeah, currently counts are only shown if everything succeeds. I thought about that, but I believe providing the actual counts would need to pass all these as variables through the recursive `Session._apply_document_changes_recursive_async` calls, which is not great. Maybe we can improve it later when the asyncio PR is merged (that should allow to simplify that logic significantly, because we don't need the recursive method anymore).

plugin/core/sessions.py
rchl: It is useful sometimes to enable it temporarily. For example when wanting to see what command is used to start the server. But it's not useful to keep it enabled all the time which makes it not really suitable for cases like this one. You want to see this one when it happens without having to try to reproduce it again.

jwortmann: Or maybe another idea would be to replace "log_debug" by a `"log_severity": "off" | "error" | "warning" | "info" | "log" | "debug"` setting, that provides more granularity about what is printed to the console. Then it could also be used for the `window/showMessage` and `window/logMessage` notifications, that we also print to the console (iirc `window/logMessage` only if "log_debug" is enabled, but I'm not sure right now).

And I think it would be nice to add a timestamp to the console messages too, similar to the timestamps in the LSP log panel.

rchl: I think it's a good idea but not necessarily relevant to this case.

In this case we are showing the message to the user always (through status field) so I would say that the  console message should also be logged always since its purpose is not to log something extra but just allow the user to read the details that he/she will likely not be able to read in 2 seconds.

rchl: And since the message could potentially be long, maybe best to just show `Error while applying WorkspaceEdit - see console for more details` in the status field.

rchl (approved):

PR #2906, opened 2026-05-16
Allow setting root ClientConfig settings

rchl:
I've realized that the approach using `__getattr__` fails horribly when trying to set the value (which I would like to do in some cases). Adding `__setattr__` is not really feasible or at least it would make things a lot more complicated...

Just expose `all_settings` directly.

I'm not convinced about the name. Perhaps should be `root_settings` or something else?

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -551,6 +551,7 @@ _METHOD_TO_CAPABILITY_EXCEPTIONS: dict[str, tuple[str, str | None]] = {
 }
 
 
+@deprecated('Remove once lsp_utils drops it from the code')
 def method2attr(method: str) -> str:
     # window/messageRequest -> m_window_messageRequest
     # $/progress -> m___progress
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -871,6 +871,10 @@ class ClientConfig:
         raise AttributeError(name)
 
     @property
+    def root_settings(self) -> Any:
+        return self._all_settings
+
+    @property
     @deprecated('Use initialization_options instead')
     def init_options(self) -> DottedDict:
         return self.initialization_options
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -852,14 +852,23 @@ class ClientConfig:
         self.diagnostics_mode = diagnostics_mode
         # Transformed mapping that uses tuples instead of lists for mdpopups.
         self.resolved_markdown_language_map: MarkdownLangMap | None = None
-        self.all_settings = all_settings or {}
-        # Exclude 'settings' because it shouldn't be considered when ClientConfig is checked for equality
-        if 'settings' in self.all_settings:
-            del self.all_settings['settings']
         self._markdown_language_map = markdown_language_map
         # For accessing configuration keys not explicitly handled above. Accessable through dunder methods below.
         self._settings_registration = settings_registration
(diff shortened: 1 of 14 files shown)


jwortmann:
Plugin-specific settings are not known or used by LSP, so I wonder what is the intention to make them writeable? Shouldn't they be read-only?

rchl:
What you mean?

Settings like `command` are writable so any other custom ones should also be. In my case I have introduced `server_path` in some plugins which user can override manually but in some cases it makes sense to set those dynamically also.

jwortmann:
For example I would expect your `server_path` to be useful only within `on_pre_start_async`. Why would you want to mutate that value on the ClientConfig? I would see accessing the plugin-specific custom settings from ClientConfig only as a convenience functionality, but not for storing additional things as a mutatable variable. If you need that later in another method (where?), you could use an instance variable for the plugin class instead (e.g. set in `on_initialized_async`).

`command` is writeable, because it is used by LSP itself.

rchl:
I'm not saying it can't be done differently but at the moment I have a case where in `on_pre_start_async`, I potentially [modify it first](https://github.com/sublimelsp/LSP-biome/blob/02a1f8ded12d826913f9ae25df7df7b782888fcc/plugin.py#L26-L42) and then [pass the whole `context` to lsp_utils](https://github.com/sublimelsp/lsp_utils/blob/ef0d404b2b98356584cd0ec40ccc9a5af24d0c24/lsp_utils/_node/node_manager.py#L61-L65) which has different logic depending on its value.

rchl:
So with my use case, the value is not modified for the sake of LSP but another "downstream" code. I don't feel that being able to modify custom root properties is inherently wrong.

That said, I'm not sure I like the `all_settings` name so I'm open to suggestions there.

rchl:
So regardless of what we do, I'll keep existing code for now for backward compatibility. Don't want to break anything.

But I've added a "normal" `root_settings` getter which I think is a better solution overall.

plugin/core/types.py
```diff
@@ -870,6 +870,10 @@ def __getattr__(self, name: str, /) -> Any:
             return self._all_settings[name]
         raise AttributeError(name)
 
+    @property
+    def root_settings(self) -> Any:
```
jwortmann: This could be typed a bit more strictly as `dict[str, Any]`, right?

jwortmann:
I guess by returning a reference to the original dict, which is mutable of course, this `root_settings` property would still make it possible to mutate the settings values on the ClientConfig object itself, despite being a `@property`...

rchl:
> I guess by returning a reference to the original dict, which is mutable of course, this `root_settings` property would still make it possible to mutate the settings values on the ClientConfig object itself, despite being a `@property`...

I do want it mutable though.

jwortmann (approved):
I think this is bad practice, and it could easily lead to weird and hard to track bugs if the setting properties become mutable from withing the plugin class. But I guess I don't have to use it and therefore I'm not really affected by this, so if you really want to have this, go on.

rchl:
I just have to disagree with you on that. I don't see much of a difference between those being mutable vs. `settings` or `initialization_settings` (used by the language server) or `command` or `selector` (used by LSP). All of those reflect user settings and I don't see how those being mutable could cause any more or new issues than the other ones being mutable.

Issue #2499, opened 2024-05-08
Create markdown file code action does not do anything


kapitanluffy:
Pressing the code action does not do anything

![image](https://github.com/sublimelsp/LSP-marksman/assets/485799/acc8b5bb-57b0-4db3-b9fc-1c9492728385)

For more context, seems like sublime's LSP implementation does not support `create` yet

```
:: [02:51:52.785] <<< marksman (23) (duration: 0ms): [{'title': 'Create `jjj.md`', 'kind': 'quickfix', 'isPreferred': False, 'edit': {'documentChanges': [{'kind': 'create', 'uri': 'file:///d%3A/Repo/ST-AutoSetSyntax/jjj.md'}]}}]
```
`'kind': 'create'` is unsupported atm. https://github.com/sublimelsp/LSP/blob/68c70316595685d2dc40f06e02b8d891f4bdab1a/plugin/core/edit.py#L20

rchl:
Duplicate of #2706

PR #2907, opened 2026-05-17
Add support for resource operations in WorkspaceEdit

### LINKED ISSUE #2706: Support workspace.workspaceEdit.resourceOperations

jwortmann:
Resolves #2706

I haven't tested in practice, but provided that tests are passing, I think it should be working.

Iirc the PHP server wanted to use resource operations, so feel free to test with that (or with another server if you know one that uses resource operations).

*Edit:* According to https://github.com/sublimelsp/LSP-intelephense/issues/112, testing with the PHP server would require a licensed version of that server. But it looks like there is also a code action in rust-analyzer that wants to use this: https://github.com/sublimelsp/LSP-rust-analyzer/issues/119

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -53,6 +53,7 @@ from ...protocol import PublishDiagnosticsParams
 from ...protocol import Range
 from ...protocol import RegistrationParams
 from ...protocol import RenameFile
+from ...protocol import ResourceOperationKind
 from ...protocol import SemanticTokenModifiers
 from ...protocol import SemanticTokenTypes
 from ...protocol import ShowDocumentParams
@@ -546,6 +547,11 @@ def get_initialize_params(
         "applyEdit": True,
         "workspaceEdit": {
             "documentChanges": True,
+            "resourceOperations": [
+                ResourceOperationKind.Create,
+                ResourceOperationKind.Rename,
+                ResourceOperationKind.Delete
+            ],
             "failureHandling": FailureHandlingKind.Abort,
             "normalizesLineEndings": True,
             "changeAnnotationSupport": {
@@ -1695,6 +1701,43 @@ class Session(APIHandler, TransportCallbacks):
                 return promise.then(lambda _: self._set_view_state(view_state_actions, view))  # pyright: ignore[reportReturnType]
             return promise
 
+        def create_file(path: str) -> Promise[str | None]:
+            # TODO: We cannot really distinguish whether the target path should be a file or a folder, because files
+            # don't necessarily have a file extension. Should we create a folder instead, if there is no file extension?
+            try:
+                open(path, 'x').close()
+            except FileExistsError as ex:
+                return Promise.resolve(str(ex))
+            except OSError as ex:
+                return Promise.resolve(str(ex))
+            return Promise.resolve(None)
+
+        def rename_file(old_path: str, new_path: str) -> Promise[str | None]:
+            view = self.window.find_open_file(old_path) if os.path.isfile(old_path) else None
+            try:
+                os.rename(old_path, new_path)
+            except FileExistsError as ex:
+                return Promise.resolve(str(ex))
+            except IsADirectoryError as ex:
+                return Promise.resolve(str(ex))
+            except NotADirectoryError as ex:
+                return Promise.resolve(str(ex))
+            except OSError as ex:
+                return Promise.resolve(str(ex))
+            if view:
+                view.retarget(new_path)
+            return Promise.resolve(None)
+
+        def delete_file(path: str) -> Promise[None]:
+            # The delete_file command moves the given files into the recycle bin
+            self.window.run_command('delete_file', {'files': [path], 'prompt': False})
+            return Promise(lambda resolve: sublime.set_timeout_async(lambda: resolve(None)))
+
+        def delete_folder(path: str) -> Promise[None]:
+            # The delete_folder command moves the given folders into the recycle bin
+            self.window.run_command('delete_folder', {'dirs': [path], 'prompt': False})
+            return Promise(lambda resolve: sublime.set_timeout_async(lambda: resolve(None)))
+
         def _continue(failure_reason: str | None) -> Promise[ApplyWorkspaceEditResult]:
             if failure_reason:
                 printf(f'Error while applying WorkspaceEdit: {failure_reason}')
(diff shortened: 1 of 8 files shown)


plugin/core/sessions.py
```diff
@@ -1695,6 +1701,43 @@ def apply_text_document_edit(
                 return promise.then(lambda _: self._set_view_state(view_state_actions, view))  # pyright: ignore[reportReturnType]
             return promise
 
+        def create_file(path: str) -> Promise[str | None]:
+            # TODO: We cannot really distinguish whether the target path should be a file or a folder, because files
+            # don't necessarily have a file extension. Should we create a folder instead, if there is no file extension?
```
rchl: I kinda feel like servers should never need to create just folders.
I know that spec mentions folders in ResourceOperationKind but it also mentions that WorkspaceEdits related to documents and to TextDocumentIdentifiers which folders don't have...

Perhaps the only case when I would create folder is if the path ends with `os.path.sep` but not sure how feasible that would be.

predragnikolic:
There is also a markdown server that supports creating files via a code action https://github.com/sublimelsp/LSP/issues/2499#issue-2373798433

I  try to test with intelephense(pro) this weekend.

predragnikolic:
When we create/rename/delete files,
do we need to send the following requests:
workspace/willCreateFiles
workspace/willRenameFiles
workspace/willDeleteFiles

And then notifications: 
workspace/didCreateFiles
workspace/didRenameFiles
workspace/didDeleteFiles

jwortmann:
> When we create/rename/delete files, do we need to send the following requests: workspace/willCreateFiles workspace/willRenameFiles workspace/willDeleteFiles
> 
> And then notifications: workspace/didCreateFiles workspace/didRenameFiles workspace/didDeleteFiles

I wondered the same, but I would assume that we only need to send these request/notifications if the create/rename/delete is caused by a user action. Otherwise maybe this could even lead to an infinite loop (?)
Because the resource operations were already triggered from a WorkspaceEdit, so I don't think that we would need to do these additional requests, which in turn might result in another WorkspaceEdit.

I think what we need to do in case of a rename resource operation is, that we need to send a `didClose` with the old URI, and then a `didOpen` notification with the new URI, for the renamed files that were open in the editor. But I this should happen automatically now, because in https://github.com/sublimelsp/LSP/pull/2907/changes/81474da68d1d1149a86d9271ac1eaa70608ec16f I added that the `lsp_uri` setting gets updated, which should then trigger a reset of the DocumentSyncListener for that view, and that should automatically send these notifications. But I haven't tested that.

plugin/core/sessions.py
```diff
@@ -1678,6 +1685,42 @@ def apply_text_document_edit(
...
+            try:
+                Path(old_path).rename(new_path)
+            except FileExistsError as ex:
+                return Promise.resolve(str(ex))
+            except IsADirectoryError as ex:
+                return Promise.resolve(str(ex))
+            except NotADirectoryError as ex:
+                return Promise.resolve(str(ex))
+            except OSError as ex:
+                return Promise.resolve(str(ex))
```
rchl: I'd prefer this pattern:


```suggestion
            except (FileExistsError, IsADirectoryError, NotADirectoryError, OSError) as ex:
                return Promise.resolve(str(ex))
```

plugin/core/sessions.py
```diff
@@ -1678,6 +1685,42 @@ def apply_text_document_edit(
...
+        def create_file(path: str) -> Promise[str | None]:
+            try:
+                Path(path).open('x', encoding='utf-8').close()
+            except FileExistsError as ex:
+                return Promise.resolve(str(ex))
+            except OSError as ex:
+                return Promise.resolve(str(ex))
+            return Promise.resolve(None)
+
+        def rename_file(old_path: str, new_path: str) -> Promise[str | None]:
```
rchl, line 1767: I wonder if it would be better to make those a standalone functions as those don't need anything from Session (window can be passed). Maybe `core/edit.py`?

rchl: Well, I've noticed now that those are nested functions. Then maybe not the best idea (unless we could move the parent functions elsewhere themselves).

rchl:
> there is also a code action in rust-analyzer that wants to use this: [sublimelsp/LSP-rust-analyzer#119](https://github.com/sublimelsp/LSP-rust-analyzer/issues/119)

Worked nicely with that use case

plugin/core/sessions.py
```diff
@@ -1678,6 +1685,36 @@ def apply_text_document_edit(
                 return promise.then(lambda _: self._set_view_state(view_state_actions, view))  # pyright: ignore[reportReturnType]
             return promise
 
+        def create_file(path: str) -> Promise[str | None]:
+            try:
+                Path(path).open('x', encoding='utf-8').close()
+            except FileExistsError as ex:
+                return Promise.resolve(str(ex))
+            except OSError as ex:
+                return Promise.resolve(str(ex))
```
rchl:
```suggestion
            except (FileExistsError, OSError) as ex:
                return Promise.resolve(str(ex))
```

rchl (approved):

plugin/core/sessions.py
jwortmann: Probably it's best to wait until the asyncio PR is merged, and then I believe we can refactor this whole function into a simple for-loop. Then we can decide whether to move these helper functions to somewhere else.

predragnikolic:
Tested with intelephense
✅ Tested successfully ,renaming the symbol will also rename the file as it can be seen in the video

https://github.com/user-attachments/assets/0e661cdf-5d15-413b-bbf1-206aad1746df


But there seem to be a UI bug lag


https://github.com/user-attachments/assets/120cc153-8f88-461d-876a-d2b312317110

Pay attention to the blue modified file indicator in the tab
Screenshot 2026-05-24 at 22 27 06
it takes a bit longer to hide... it can be perceived as lagging and there is some UI slight UI flickering 


When I rename the symbol:
if I screen record the ST window app:
  the blue modified file indicator in the tab is hidden right after the file renamed successfully
else
  the blue modified file indicator in the tab stays visible way longer after the file renamed successfully

I do expect the that the blue indicator is hidden right after the file has been rename.

predragnikolic:
I tested with LSP-marksman as well.

When I click the create file code action,
the fill will be created.
but after we create the file we need to send the workspace/didCreateFiles,
else the server will still show the diagnostic error that the file doesn't exist ("Link to non-existent file"), as it can be seen in the video. 

https://github.com/user-attachments/assets/e816ced0-25ce-4abc-9abb-011221ff20f2

If we send the notification, the server will remove that diagnostic.

(I hit the same error when I playing around with resources operations in https://github.com/sublimelsp/LSP/tree/resource-operations-wip - I will delete this branch, I just pushed the changes so you can take a look)

jwortmann:
Thanks for testing!

Ok I will add the notifications.

---

Regarding the blue modified file indicator, I assume that is out of our control and depends on ST itself (?)

But I can also not really tell from the video whether you manually save the file after the rename... or otherwise I would expect that the file is still in a modified state after the rename (because the file content on disk was copied into the other file location, but there was no save after the content change), and therefore the blue indicator shouldn't go away. But I might be wrong because I have not tested it.

predragnikolic:
> But I can also not really tell from the video whether you manually save the file after the rename...

I forgot to mention, I am not the one who is pressing save. 
The file is saved automatically because I have set:
```
 "refactoring_auto_save": "always"
```

jwortmann:
> I forgot to mention, I am not the one who is pressing save.
> The file is saved automatically because I have set:

Oh, that makes sense. But yeah, I think this is out of our control and depends on ST (perhaps it takes a bit for ST to react due to the `view.retarget` that we do when renaming the file).

predragnikolic:
I will tinker a bit with the code.

I have one assumption now...
UI glitches like this usually happen when code is executed on the async thread. 
I will try to move some code to be run on the main thread and see if that would address it.

jwortmann:
I have implemented the `workspace/didCreateFiles` / `workspace/didRenameFiles` / `workspace/didDeleteFiles` notifications in https://github.com/sublimelsp/LSP/pull/2907/changes/d20b41f70966bd0fbfca10b998d0156d24505d31, and I have successfully tested it with LSP-marksman using your example from above.

*Edit:* I have not implemented these notifications for the MockManager class from https://github.com/sublimelsp/LSP/blob/7313cec2d2284f1deb0ecebd36ec0f869fd769e7/tests/test_session.py#L30
so I assume they won't be sent during the unit tests.

jwortmann:
By the way, what happens if someone has also installed LSP-file-watcher-chokidar - will there be additional notifications triggered by that when a file in the workspace gets created or deleted?

*Edit:* I guess I can answer myself; according to the readme https://github.com/sublimelsp/LSP-file-watcher-chokidar#usage the file watcher seems only be relevant for `workspace/didChangeWatchedFiles` notifications. So it should be fine and there should not be any duplicate notifications.

*Edit 2:* The description in the specs at https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#workspace_didChangeWatchedFiles for `workspace/didChangeWatchedFiles` reads like this notification is a replacement for the individual `didCreateFiles` / `didRenameFiles` / `didDeleteFiles` notifications (?). Although that is not explicitly mentioned, and I wonder why the description is so vague about that. But I guess having support for all those notification types in LSP is fine.

jwortmann:
> I will tinker a bit with the code.
> 
> I have one assumption now... UI glitches like this usually happen when code is executed on the async thread. I will try to move some code to be run on the main thread and see if that would address it.

Do you still want to test this, or should we merge this PR and if we find some improvements we can do that later?

predragnikolic:
> or should we merge this PR and if we find some improvements we can do that later?

Merge would be fine. I will try to revisit the issue that I reported, once I dedicate more time to LSP.

Janos, can you just confirm that this part of the spec is present in the PR - "execute the operations in the order in which they are provided"

> Since version 3.13.0 a workspace edit can contain resource operations (create, delete or rename files and folders) as well. If resource operations are present clients need to execute the operations in the order in which they are provided. So a workspace edit for example can consist of the following two changes: (1) create file a.txt and (2) a text document edit which insert text into file a.txt. An invalid sequence (e.g. (1) delete file a.txt and (2) insert text into file a.txt) will cause failure of the operation. How the client recovers from the failure is described by the client capability: workspace.workspaceEdit.failureHandling

jwortmann:
> Janos, can you just confirm that this part of the spec is present in the PR - "execute the operations in the order in which they are provided"

Yes, we should handle it correctly with this implementation. This is the why it is implemented as such an ugly recursive function, because we need to chain the next Promise after each operation (or stop as soon as there is a failure).

rchl (approved):

huyby:
Thanks, works like a charm using lsp intelephense ❤️

PR #2913, opened 2026-05-17
show "Follow link" in hover popup for unknown links

rchl:
The text shown when hovering document links was not ideal for unknown schemas. In JSON we've shown "Open in browser" for links with `json-schema` schema.

Screenshot 2026-05-17 at 22 33 20

Show "Open in browser" only for URIs with http/https schema. For other show "Follow link".

If we don't have `target` then we don't know schema until resolve and we'll also show "Follow link".

### DIFF
--- a/plugin/core/open.py
+++ b/plugin/core/open.py
@@ -51,7 +51,6 @@ def lsp_range_from_uri_fragment(fragment: str) -> Range | None:
 def open_file_uri(
     window: sublime.Window, uri: DocumentUri, flags: sublime.NewFileFlags = sublime.NewFileFlags.NONE, group: int = -1
 ) -> Promise[sublime.View | None]:
-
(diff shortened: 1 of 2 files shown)


jwortmann:
IIRC the code at
https://github.com/sublimelsp/LSP/blob/608929e3fbd358430d0abf36c92dcb2cb946e297/plugin/hover.py#L344-L345
was added later, and at the time when this label was implemented it would always fall back to open unknown links in the browser, that is why the original label was like that.

The fix looks good, but should we maybe adjust the label to `Open Link` instead of `Follow Link`?

jwortmann (approved):

rchl:
> The fix looks good, but should we maybe adjust the label to `Open Link` instead of `Follow Link`?

In that example that I gave, `Follow link` makes more sense IMO since it's a link to a different position in the same file. But `Open link` would work too I guess.

PR #2914, opened 2026-05-17
Fix on_sheet_opened called twice from uri_handler

rchl:
This is a bit rushed fix so might need to have another, more thorough, look.

Problem:

When following the [migration guide](https://lsp.sublimetext.io/migrating_to_lsp_plugin/#13-replace-on_open_uri_async-with-uri_handler) and using `session.open_scratch_buffer`, the `_on_sheet_opened` was called [two](https://github.com/sublimelsp/LSP/blob/9c689ccbfbdf62f29b44a57fbdcf409752c7558b/plugin/core/sessions.py#L1479-L1479) [times](https://github.com/sublimelsp/LSP/blob/9c689ccbfbdf62f29b44a57fbdcf409752c7558b/plugin/core/sessions.py#L1572-L1572). Once with the `uri` provided by the handler and then overridden with the original `uri`.

Remove the `_on_sheet_opened` call chained to the handler - if handler uses `session.open_scratch_buffer` then it is not needed. When handler doesn't use `session.open_scratch_buffer` then this might be problematic to not call it though...

(Another issue is that when following the migration guide the types are not validating because `session.open_scratch_buffer` returns `View` while `uri_handler` wants `sheet` so there is extra `then` needed to convert the return value.)

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1476,7 +1476,7 @@ class Session(APIHandler, TransportCallbacks):
             if isinstance(self._plugin, LspPlugin):
                 scheme, _ = parse_uri(uri)
                 if handler := self._plugin.get_uri_handler(scheme):
-                    return handler(uri, flags).then(lambda sheet: self._on_sheet_opened(sheet, uri, r))
+                    return handler(uri, flags).then(lambda sheet: sheet.view() if sheet else None)
             else:
                 return self._open_uri_with_plugin_async(self._plugin, uri, r, flags, group)
         return None
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1567,21 +1567,20 @@ class Session(APIHandler, TransportCallbacks):
             view.set_name(title)
             view.run_command("append", {"characters": content})
             view.set_read_only(True)
-            self._on_sheet_opened(view.sheet(), uri, r).then(lambda _: resolve(view))
+            self._on_sheet_opened(view.sheet(), uri, r)
+            resolve(view)
 
         sublime.set_timeout(continue_on_main_thread)
         return promise
 
-    def _on_sheet_opened(
-        self, sheet: sublime.Sheet | None, uri: DocumentUri, r: Range | None
-    ) -> Promise[sublime.View | None]:
+    def _on_sheet_opened(self, sheet: sublime.Sheet | None, uri: DocumentUri, r: Range | None) -> sublime.View | None:
         if sheet and (view := sheet.view()):
             uri_no_fragment = urldefrag(uri).url
             view.settings().set('lsp_uri', uri_no_fragment)
             if r:
                 center_selection(view, r)
-            return Promise.resolve(view)
-        return Promise.resolve(None)
+            return view
+        return None
 
     def open_location_async(
         self,
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1476,7 +1476,7 @@ class Session(APIHandler, TransportCallbacks):
             if isinstance(self._plugin, LspPlugin):
                 scheme, _ = parse_uri(uri)
                 if handler := self._plugin.get_uri_handler(scheme):
-                    return handler(uri, flags).then(lambda sheet: sheet.view() if sheet else None)
+                    return handler(uri, flags).then(lambda sheet: self._on_sheet_opened(sheet, uri, r))
             else:
                 return self._open_uri_with_plugin_async(self._plugin, uri, r, flags, group)
         return None
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1523,7 +1523,7 @@ class Session(APIHandler, TransportCallbacks):
         # I cannot type-hint an unpacked tuple
         pair: PackagedTask[tuple[str, str, str]] = Promise.packaged_task()
         # It'd be nice to have automatic tuple unpacking continuations
-        callback = lambda a, b, c: pair[1]((a or 'Untitled', b, c))  # noqa: E731
+        callback = lambda a, b, c: pair[1]((a or 'untitled', b, c))  # noqa: E731
         if plugin.on_open_uri_async(uri, callback):
             result: PackagedTask[sublime.View | None] = Promise.packaged_task()
             pair[0].then(lambda tup: self.open_scratch_buffer(*tup, uri, r, flags, group)).then(result[1])
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1549,8 +1549,8 @@ class Session(APIHandler, TransportCallbacks):
         title: str,
         content: str,
         syntax: str,
-        uri: DocumentUri,
-        r: Range | None,
+        uri: DocumentUri | None = None,
+        r: Range | None = None,
         flags: sublime.NewFileFlags = sublime.NewFileFlags.NONE,
         group: int = -1,
     ) -> Promise[sublime.View]:
@@ -1573,10 +1573,13 @@ class Session(APIHandler, TransportCallbacks):
         sublime.set_timeout(continue_on_main_thread)
         return promise
 
-    def _on_sheet_opened(self, sheet: sublime.Sheet | None, uri: DocumentUri, r: Range | None) -> sublime.View | None:
+    def _on_sheet_opened(
+        self, sheet: sublime.Sheet | None, uri: DocumentUri | None, r: Range | None
+    ) -> sublime.View | None:
         if sheet and (view := sheet.view()):
-            uri_no_fragment = urldefrag(uri).url
-            view.settings().set('lsp_uri', uri_no_fragment)
+            if uri:
+                uri_no_fragment = urldefrag(uri).url
+                view.settings().set('lsp_uri', uri_no_fragment)
             if r:
                 center_selection(view, r)
             return view
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -1535,9 +1535,9 @@ class Session(APIHandler, TransportCallbacks):
         group: int,
     ) -> Promise[sublime.View | None] | None:
         # I cannot type-hint an unpacked tuple
-        pair: PackagedTask[tuple[str | None, str, str]] = Promise.packaged_task()
+        pair: PackagedTask[tuple[str, str, str]] = Promise.packaged_task()
         # It'd be nice to have automatic tuple unpacking continuations
-        callback = lambda a, b, c: pair[1]((a, b, c))  # noqa: E731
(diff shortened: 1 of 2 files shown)


jwortmann:
Isn't the right solution to just remove it from `open_scratch_buffer` at https://github.com/sublimelsp/LSP/blob/9c689ccbfbdf62f29b44a57fbdcf409752c7558b/plugin/core/sessions.py#L1572-L1572

Then it would get called a single time in all cases, no?

rchl:
> Isn't the right solution to just remove it from `open_scratch_buffer` at
> 
> https://github.com/sublimelsp/LSP/blob/9c689ccbfbdf62f29b44a57fbdcf409752c7558b/plugin/core/sessions.py#L1572-L1572
> 
> Then it would get called a single time in all cases, no?

No, due to fact that then this code will call `_on_sheet_opened` with the original `uri` and `r` which might not be the same as the ones that the `uri_handler` passed to `open_scratch_buffer`.

The example case is: uri handler gets an URI with a fragment. The fragment declares position in the file so it's meant to be stripped from the uri before its set as `lsp_uri`. This code then overwrites it to be the full uri with the fragment.

But then I also think it's a bit rushed fix and ideally the `_on_sheet_opened` would always get called after the handler. But then there would need to be be some way of telling whether it was already called.

rchl:
> But then I also think it's a bit rushed fix and ideally the `_on_sheet_opened` would always get called after the handler. But then there would need to be be some way of telling whether it was already called.

But perhaps that's not that important because if you are opening a View then likely you want to use `session.open_scratch_buffer`. If not a View then `_on_sheet_opened` is not relevant since it only works for Views.

jwortmann:
I think at https://github.com/sublimelsp/LSP/blob/828a94d756ac4fc72acb0f1bee816a04e26f744e/plugin/core/sessions.py#L1465 we should just remove the fragment from the `uri` that is passed into `_on_sheet_opened`. Possibly we might also want to normalize the URI at https://github.com/sublimelsp/LSP/blob/828a94d756ac4fc72acb0f1bee816a04e26f744e/plugin/core/sessions.py#L1567 but I think that shouldn't really be relevant for URIs with custom schemes (or perhaps it would even be bad if we send back slightly different URIs to the server, so probably normalizing wouldn't be a good idea and is only relevant for `file:` URIs).

And I think using the original `r` is still the correct thing, because it might come from a [`Location`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#location) and then that would/should take precedence over some row/column that might be derived from the URI fragment. If `r` is None, the position in the view can still be controlled from the custom URI handler.

jwortmann:
Actually setting `lsp_uri` to the view returned from the handler might not always be good; for example LSP-Tinymist uses some really weird URIs to open other files either in the editor or in an external viewer (iirc the only place where I have seen these URIs are when hovering over links to image files).

https://github.com/sublimelsp/LSP-Tinymist/blob/5173ee56e777e3587f4823ef82107905e7445cd5/plugin.py#L247-L260

rchl:
> I think at
> 
> https://github.com/sublimelsp/LSP/blob/828a94d756ac4fc72acb0f1bee816a04e26f744e/plugin/core/sessions.py#L1465
> we should just remove the fragment from the `uri` that is passed into `_on_sheet_opened`.

That would work for my use case.

LSP-yaml supports inline hints that when clicked open schema for the corresponding yaml file. In that case it triggers opening of URI like:
```
json-schema://www.schemastore.org/github-workflow.json#https%3A%2F%2Fwww.schemastore.org%2Fgithub-workflow.json
```

where the fragment is the URL to download the schema from.

In that case we want to remove the fragment and use `json-schema://www.schemastore.org/github-workflow.json` URI for the opened file. Then document links within this file will look like:

```
json-schema://www.schemastore.org/github-workflow.json#412,14
```

and LSP will handle those automatically to support jumping to specified location.

But if that wouldn't work for `LSP-Tinymist` then we should just allow setting URI manually which this change allows.

jwortmann:
In the Tinymist use case the server should better just use regular `file:` URIs, and since it seems to be only used for image files, allowing to set `lsp_uri` manually isn't really needed for that use case.

I think I would prefer setting the URI without fragment from `_on_sheet_opened`. That would ensure that the URI is set even when not using `Session.open_scratch_buffer`.

rchl:
> I think I would prefer setting the URI without fragment from `_on_sheet_opened`. That would ensure that the URI is set even when not using `Session.open_scratch_buffer`.

But then there is the second issue of it being called twice if handler uses `session.open_scratch_buffer`. Which shouldn't be a big issue in practice but still weird.

Or if we remove `_on_sheet_opened` from `session.open_scratch_buffer` then it will be less useful if someone wants to use it from somewhere else.

jwortmann:
> Or if we remove `_on_sheet_opened` from `session.open_scratch_buffer` then it will be less useful if someone wants to use it from somewhere else.

The purpose for `_on_sheet_opened` is to be used when a URI with custom scheme gets opened. So I'd say that's fine and I would not expect to call that when `Session.open_scratch_buffer` is used from somewhere else. Random scratch buffers will automatically get the `buffer` URI from the ViewListener.

rchl:
> > Or if we remove `_on_sheet_opened` from `session.open_scratch_buffer` then it will be less useful if someone wants to use it from somewhere else.
> 
> The purpose for `_on_sheet_opened` is to be used when a URI with custom scheme gets opened. So I'd say that's fine and I would not expect to call that when `Session.open_scratch_buffer` is used from somewhere else. It will automatically get the `buffer` URI from the ViewListener.

Not always, LSP-json wants to handle custom `json-schema` scheme files opened by LSP-yaml:
https://github.com/sublimelsp/LSP-json/blob/2efca03a275a4b6c1e3ee5dd657f445276c1d76f/LSP-json.sublime-settings#L11-L12

That said, I've realized that just returning `Promise(None)` from the handler will avoid calling `_on_sheet_opened` again which is good enough solution for me.

We could try to make other discussed changes but not sure if needed and that we've arrived at perfect solution.

jwortmann:
I would even modify `open_scratch_buffer` such that `title` must be provided (not None), remove `uri` argument, and return type would be `Promise[sublime.View]`.

Then the example in the URI handler for the docs would be
```python
return session.open_scratch_buffer(...).then(lambda view: view.sheet())
```

plugin/core/sessions.py
```diff
@@ -1578,7 +1579,8 @@ def _on_sheet_opened(
         self, sheet: sublime.Sheet | None, uri: DocumentUri, r: Range | None
     ) -> Promise[sublime.View | None]:
         if sheet and (view := sheet.view()):
-            view.settings().set('lsp_uri', uri)  # Preserve original URI given by the language server
+            uri_no_fragment = urldefrag(uri).url
+            view.settings().set('lsp_uri', uri_no_fragment)
```
rchl, line 1564: So I still did this change as I think it makes sense in general. Will allow to simplify the `LSP-yaml` handler slightly.

jwortmann:
Still doesn't feel optimal to me. I would expect `open_scratch_buffer` to be possible to call without proving the URI, and then URI default to the `buffer:123` URIs that we use.

rchl:
> I would even modify `open_scratch_buffer` such that `title` must be provided (not None), remove `uri` argument, and return type would be `Promise[sublime.View]`.
> 
> Then the example in the URI handler for the docs would be
> 
> ```python
> return session.open_scratch_buffer(...).then(lambda view: view.sheet())
> ```

Made title required and changed return type.
But I'm still hesitant to remove `uri` and `_on_sheet_opened` from it because:
a) there might be use cases that want the `lsp_uri` to be something more custom.
b) `session.open_scratch_buffer` will be less flexible if someone wants to use it from some other place

plugin/core/sessions.py
```diff
@@ -1545,21 +1546,18 @@ def _open_uri_with_plugin_async(
 
     def open_scratch_buffer(
         self,
-        title: str | None,
+        title: str,
         content: str,
         syntax: str,
         uri: DocumentUri,
```
rchl: Let's use thread (easier to follow)

> Still doesn't feel optimal to me. I would expect open_scratch_buffer to be possible to call without proving the URI, and then URI default to the buffer:123 URIs that we use.

We can make `uri` optional but I would still argue for keeping `_on_sheet_opened` in here.

plugin/core/sessions.py
```diff
@@ -1520,9 +1521,9 @@ def _open_uri_with_plugin_async(
         group: int,
     ) -> Promise[sublime.View | None] | None:
         # I cannot type-hint an unpacked tuple
-        pair: PackagedTask[tuple[str | None, str, str]] = Promise.packaged_task()
+        pair: PackagedTask[tuple[str, str, str]] = Promise.packaged_task()
         # It'd be nice to have automatic tuple unpacking continuations
-        callback = lambda a, b, c: pair[1]((a, b, c))  # noqa: E731
+        callback = lambda a, b, c: pair[1]((a or 'Untitled', b, c))  # noqa: E731
```
jwortmann: I think ST uses `untitled` by default (lowercased).

jwortmann:
To be honest, I still don't like it. We would also need to update the docstring: "Return a Promise resolved with the opened sheet if you opened the URI as an editor tab. Unless you used this particular function, then return Promise[None] for some internal reasons that we don't tell you here ¯\\_(ツ)_/¯"

The reason for having the sheet/view returned is that the function that called `Session.open_uri_async` might further need to work with the returned view. For example for changes in WorkspaceEdit to a document with custom URI scheme.

plugin/core/sessions.py
```diff
@@ -1520,32 +1521,27 @@ def _open_uri_with_plugin_async(
...
         # It'd be nice to have automatic tuple unpacking continuations
-        callback = lambda a, b, c: pair[1]((a, b, c))  # noqa: E731
+        callback = lambda a, b, c: resolve((a or 'untitled', b, c))  # noqa: E731
         if plugin.on_open_uri_async(uri, callback):
-            result: PackagedTask[sublime.View | None] = Promise.packaged_task()
-            pair[0].then(lambda tup: self.open_scratch_buffer(*tup, uri, r, flags, group)).then(result[1])
-            return result[0]
+            return promise.then(lambda tup: self.open_scratch_buffer(*tup, flags, group)) \
+                .then(lambda view: self._on_sheet_opened(view.sheet(), uri, r))
         return None
```
rchl, line 1533: Two PackagedTask's were unnecessary - one is enough.

jwortmann (approved):

plugin/core/sessions.py
predragnikolic: If a server provides an uri, 

and we set a different uri here.

can that have some negative side effects, when we later send that uri back to the server?

PR #2920, opened 2026-05-20
docs: add instructions to install nixd for Nix

maximsmol:
Also see https://github.com/nix-community/nixd/pull/825

### DIFF
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -493,19 +493,7 @@ Follow installation instructions on [LSP-nimlangserver](https://github.com/subli
 
 ## Nix
 
-1. Install [nixd](https://github.com/nix-community/nixd/blob/main/nixd/docs/editor-setup.md#installation---get-a-working-executable)
-2. Open `Preferences > Package Settings > LSP > Server Configurations` (or run "Preferences: LSP Server Configurations" in the command palette) and add the `"nixd"` configuration:
-
-
-    ```json
-    {
-      "nixd": {
-        "enabled": true,
-        "command": ["nixd"],
-        "selector": "source.nix"
-      }
-    }
-    ```
+Follow installation instructions on [nixd](https://github.com/nix-community/nixd).
 
 ## OCaml/Reason
 
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -493,7 +493,7 @@ Follow installation instructions on [LSP-nimlangserver](https://github.com/subli
 
 ## Nix
 
-Follow installation instructions on [nixd](https://github.com/nix-community/nixd).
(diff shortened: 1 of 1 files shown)


rchl:
Duplicating instructions is not the best. Ideally upstream would link to our instructions as that guarantees that it will be up to date. Alternatively we could link to instructions upstream but duplicating instructions is not ideal.

maximsmol:
It looks like Sublime LSP docs already link out in a bunch of places [(e.g. for C/C++).](https://lsp.sublimetext.io/language_servers/#cc) so I'll follow that precedent

docs/src/language_servers.md
```diff
@@ -491,6 +491,10 @@ Follow installation instructions on [LSP-marksman](https://github.com/sublimelsp
 
 Follow installation instructions on [LSP-nimlangserver](https://github.com/sublimelsp/LSP-nimlangserver).
 
+## Nix
+
+Follow installation instructions on [nixd](https://github.com/nix-community/nixd).
```
rchl: Should the link be more specific, like https://github.com/nix-community/nixd/blob/main/nixd/docs/editor-setup.md#sublime-text-4 ?

maximsmol: I was following the convention on the other entries - they all link to the overal project (again see C/C++, C#, CSS, etc.)

Personally I think it's probably better to give a specific link (and also to mention that you need to install the binary itself). Let me know if you want me to do that rather than the way it is for the other entries.

rchl: The other cases you've mentioned link to our packages which DO actually have installation/setup instructions on that specific page (in the root readme).

Here at the very least I'd link to the root setup page - https://github.com/nix-community/nixd/blob/main/nixd/docs/editor-setup.md

maximsmol: OK made the change. I decided against linking to the sublime-specific heading because you have to install the executable before adding the LSP config. The executable install instructions are on the same page.

docs/src/language_servers.md
```diff
@@ -491,6 +491,10 @@ Follow installation instructions on [LSP-marksman](https://github.com/sublimelsp
 
 Follow installation instructions on [LSP-nimlangserver](https://github.com/sublimelsp/LSP-nimlangserver).
 
+## Nix
+
+Follow installation instructions on [nixd](https://github.com/nix-community/nixd/blob/main/nixd/docs/editor-setup.md).
```
rchl: It would be better to follow good accessibility practices and make the link's text describe the contents.

```suggestion
Follow [installation instructions on nixd](https://github.com/nix-community/nixd/blob/main/nixd/docs/editor-setup.md).
```

rchl (approved):

PR #2923, opened 2026-05-22
refactor: group functions logically in views.py

rchl:
Suggestion to refactor the views.py behemoth into a module with functions grouped into logical parts.


jwortmann:
I think I would prefer not to add another folder level. Actually I would even prefer to remove the `core` folder and have a flat structure for the source code files (e.g. I really dislike that there is a `plugin/edit.py` file, and then also another `plugin/core/edit.py`, which I find totally confusing and I have no clue why this functionality is split into two files), but I guess others might disagree to that and such a restructuring would probably cause many breaking changes.

The `views.py` file seems to be the place for various utility functions (I also wonder why it was named `views.py` instead of `utils.py`), so if we want to clean it up, we could perhaps move some of the content into other the other existing files which are related to the specific functionalities. Perhaps the file was named `views.py` because some of the functions take a `sublime.View` as an argument? If we want to revise some of these utility functions (for #2910), I would suggest the convention that all functions which take a `sublime.View` as an argument, should always have that as the *first* argument. This is already the case for many, but not all functions, for example it should be `position_to_offset(view: sublime.View, position: Position)` instead of the existing `position_to_offset(position: Position, view: sublime.View)`.

rchl:
OK. We shouldn't do any big refactors while the asyncio PR is in progress though (this would be fine because asyncio task doesn't touch `views.py`).

rchl:
I know this is unreviewable and not that useful in ST but it does group related stuff together so it would be easier to move and manage code maybe.

rchl:
I've added grouping using region comments instead. Unfortunately it doesn't do much in ST but maybe in the future it would do this:

Screenshot 2026-05-23 at 16 07 26

predragnikolic:
Will this PR cause conflicts with the async PR? If yes, it would be nice to delay this one, to avoid causing git conflicts for Raoul.

(I am afk, but maybe his PR didn't touch this views.py file)

Update: thanks and sorry for missing the messages above

rchl:
> this would be fine because asyncio task doesn't touch `views.py`

PR #2925, opened 2026-05-23
Fix lsp_code_lens command executing last code lens on cancel

jwortmann:
When `Window.show_quick_panel` is cancelled with <kbd>Esc</kbd>, it returns the index `-1` to the callback function. This would accidentally cause the last code lens in the list to be executed, because indexing a list with `-1` returns the last list item in Python.

### DIFF
--- a/plugin/code_lens.py
+++ b/plugin/code_lens.py
@@ -171,10 +171,7 @@ class LspCodeLensCommand(LspTextCommand):
     def on_select(self, commands: list[tuple[str, Command]], index: int) -> None:
         if index == -1:
             return
-        try:
-            session_name, command = commands[index]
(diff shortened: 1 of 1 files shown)


plugin/code_lens.py
rchl: The try-except block seems unnecessary also (it pretended to handle the -1 case before, I guess).

rchl (approved):

plugin/code_lens.py
jwortmann: Yeah, I think should be safe to remove: https://github.com/sublimelsp/LSP/pull/2925/changes/a8750c9c1821c14cf4c78cc72d912e2f8f26468e

PR #2929, opened 2026-05-24
Add link to Civet LSP

edemaine:
[Civet](https://civet.dev) is a language that compiles to TypeScript. We just tested Civet's LSP in Sublime; see [these demo screens](https://github.com/DanielXMoore/Civet/pull/2111). So thought it might help to add to this list.

### DIFF
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -60,6 +60,10 @@ Follow installation instructions on [LSP-clangd](https://github.com/sublimelsp/L
 
 Follow installation instructions on [LSP-OmniSharp](https://github.com/sublimelsp/LSP-OmniSharp).
 
+## Civet
+
+Follow installation instructions in the [Civet Sublime Text package](https://github.com/DanielXMoore/Civet/tree/main/lsp/sublime).
+
 ## Clojure
 
 1. Download [clojure-lsp](https://clojure-lsp.io/installation/).


rchl:
Note that instructions mention deprecated way of configuring clients using global LSP preferences. There is now a dedicated file for those - https://lsp.sublimetext.io/client_configuration/

edemaine:
Sorry, I'm new to Sublime. What are the recommended instructions? Or if you can point me to an example, I can adapt.

docs/src/language_servers.md
```diff
@@ -60,6 +60,10 @@ Follow installation instructions on [LSP-clangd](https://github.com/sublimelsp/L
 
 Follow installation instructions on [LSP-OmniSharp](https://github.com/sublimelsp/LSP-OmniSharp).
 
+## Civet
+
+Follow installation instructions in the [Civet Sublime Text package](https://github.com/DanielXMoore/Civet/tree/main/lsp/sublime).
```
rchl, line 65: I would say that the `Civet Sublime Text package` phrasing is a bit unfortunate because it's not providing any package. In ST packages are things that you can install directly in the editor using its Package Control package manager.

rchl:
> Sorry, I'm new to Sublime. What are the recommended instructions? Or if you can point me to an example, I can adapt.

I did link to instructions.
Specifically:

> After you have installed a language server, LSP needs to be configured to enable communication between LSP and that server for suitable filetypes. Your server configurations file is stored at Packages/User/LanguageServers.sublime-settings and can be opened via "Preferences > Package Settings > LSP > Server Configurations" from the menu or with the Preferences: LSP Server Configurations entry from the Command Palette.

So client settings no longer go into `Package Settings → LSP → Settings` like the instructions say.
And in that file there no longer is a root `clients` key. The server's name is now the root key.

docs/src/language_servers.md
edemaine: https://github.com/DanielXMoore/Civet/tree/main/lsp/sublime is providing a package in the sense that it's installable into the `Packages` directory of Sublime. We haven't gotten to Package Control listing yet because it's apparently not compatible with our monorepo layout...

Are you objecting to the link title or the description in our README?

rchl: It's not a correct package even. You can't ship it and expect that your configuration will be merged with what is already present in LSP.sublime-settings. In fact I'm not even sure what will happen but I guess it will either overwrite the original LSP.sublime-settings or do nothing. Both of which would be unexpected.

You should either just have instructions on how to add the server configuration to `Preferences: LSP Server Configurations` manually (probably easiest for now) or create a dedicated package with its own configuration (you have tons of examples in https://github.com/sublimelsp) but those are not just configurations but also include a bit of plugin code.

edemaine: Yes, the Preferences documentation are updated according to your comments in a PR linked below (under review).

This package does do something: it provides syntax highlighting.

rchl: > This package does do something: it provides syntax highlighting.

But then it most likely overrides the whole `clients` object so that all servers that the user has defined won't work anymore.

edemaine: 1. That's recommended configuration (`.example`), not part of the package. 

2. It'll be fixed once https://github.com/DanielXMoore/Civet/pull/2116 is merged. 

Unless I missed something?

rchl: OK, I understand everything now and the changes here look fine.

I've missed before that the package provides syntax highlighting because there is no actual syntax file in there. I see that the process to get things built is a bit complex but I understand that it's just the state of things for now.

rchl (approved):

edemaine:
Sorry it was confusing. Thanks for all the quick feedback!

PR #2930, opened 2026-05-25
Remove (seemingly incorrect) workaround for old markdown parser

### LINKED ISSUE #2928: Issue with rendering paragraphs
Initially posted in mdpopups repo: https://github.com/facelessuser/sublime-markdown-popups/issues/163

After author's reply looking in LSP code it turned that the problem is with:

https://github.com/sublimelsp/LSP/blob/7981a132ce889cfa0967372805b6909483a3c064/plugin/core/views.py#L618

Commenting that regex substitution, shows paragraphs correctly:

Image

rchl:
Discussion and context in #2928

Fixes #2928

### DIFF
--- a/plugin/core/views.py
+++ b/plugin/core/views.py
@@ -614,8 +614,6 @@ def minihtml(
                 }
             }
         ]
-        # Workaround CommonMark deficiency: two spaces followed by a newline should result in a new paragraph.
-        result = re.sub('(\\S)  \n', '\\1\n\n', result)  # noqa: RUF039
     if isinstance(language_id_map, dict):
         frontmatter["language_map"] = language_id_map
     return mdpopups.md2html(view, mdpopups.format_frontmatter(frontmatter) + result)


rchl:
Well, it would regress clangd so maybe we should instead make it more specific.

### Before

Screenshot 2026-05-25 at 18 04 07

### After

Screenshot 2026-05-25 at 18 04 18

### Payload

```json
{
  "contents": {
    "kind": "markdown",
    "value": "### instance-method `assign`\n\nprovided by `<vector>`\n\n---\n→ `void`\n\nParameters:\n\n- `_InputIterator __first`\n- `_InputIterator __last`\n\n---\n```cpp\n// In vector<char>\npublic: template <\n    class _InputIterator,\n    __enable_if_t<\n        __has_exactly_input_iterator_category<_InputIterator>::value &&\n            is_constructible<value_type, typename iterator_traits<\n                                             _InputIterator>::reference>::value,\n        int> = 0>\nvoid assign(_InputIterator __first, _InputIterator __last)\n```"
  },
  "range": {
    "end": {
      "character": 12,
      "line": 17
    },
    "start": {
      "character": 6,
      "line": 17
    }
  }
}
```

rchl:
We can add a workaround inside LSP-clangd itself. Probably something like `result = re.sub('[ ]{2,}\n-', '\n\n-', result)`

jwortmann (approved):

rwols:
Why does the code flow not use the CommonMark parser from mdpopups?

jwortmann:
> Why does the code flow not use the CommonMark parser from mdpopups?

It does:
https://github.com/sublimelsp/LSP/blob/5561bfa5de3a3c5f0244de7069ffe85e7b917031/plugin/core/views.py#L601-L602

But only if supported by the mdpopups version (i.e. on ST dev build).

rwols:
So: it should be auto-fixed when MARKO_MD_PARSER_VERSION becomes true-ish for the majority of users. Then we can remove the workaround.

I will revert this because it breaks my hover info.

rchl:
Wait. I said that I will add this workaround to LSP-clangd instead, where it belongs.

rchl:
And please follow our convention and only push any meaningful changes to main through a PR.

rwols:
Not all users use LSP-* helper packages... so that would only help for a subset of the users.

rwols:
> And please follow our convention and only push any meaningful changes to main through a PR.

OK, I'll keep it in mind.

rchl:
> Not all users use LSP-* helper packages... so that would only help for a subset of the users.

I think this is a pretty weak argument. We don't have numbers of course but why would one not uses dedicated package for it given that it provides all the extra convenience like documented settings and extra features? And also that's the documented way to set up clangd according to [our instructions](https://lsp.sublimetext.io/language_servers/#cc).

It doesn't even use managed server by default so those who have problem with that shouldn't complain either.

I think that if someone is in the minority that is not using that package should also not mind this minor rendering inconvenience.

I don't want to create a conflict about this now but I really think you should revert your revert. It's an issue in clangd so we shouldn't create incorrect and non-compliant workarounds that break rendering in other servers.

PR #2932, opened 2026-05-27
fix project overrides getting lost on saving global configs

rchl:
I've discovered a bug with saving  LSP packages preferences.

After making any change in `settings` and saving, we'd send `workspace/didChangeConfiguration` to the server without including project overrides.

To reproduce:

1. Set in project settings:
```jsonc
{
    "settings": {
        "LSP": {
            "LSP-pyright": {
                "settings": {
                    "pyright.dev_environment": "sublime_text"
                }
            }
        }
    }
}
```
2. Open `Preferences: LSP-pyright Settings`
3. Change (add/remove) some other key in `settings`
4. Save

We'd send updated configuration without `pyright.dev_environment`.

The issue was that we've evaluated whether config has changed too early (in `ClientConfigs`). We have to let `WindowConfigManager` evaluate that after merging project overrides and then notify session about changes from within `WindowManager`.

### DIFF
--- a/plugin/core/configurations.py
+++ b/plugin/core/configurations.py
@@ -99,6 +99,7 @@ class WindowConfigManager:
                 self.all[name] = updated_config
                 configs_recreated.append(updated_config)
             elif stored_config.settings != updated_config.settings:
+                stored_config.settings = updated_config.settings
                 configs_with_changed_settings.append(updated_config)
         # project_settings no longer contains config_names that were handled above.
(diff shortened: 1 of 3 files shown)


jwortmann:
I think there is still another bug that if you modify a server setting in project settings, then the server is restarted (but that also happens on main, so it's not a new bug).

jwortmann (approved):

rchl:
I'll have a look at it next.

PR #2935, opened 2026-05-28
Only request documentLink after changes if underline enabled

jwortmann:
I though about how LSP handles `textDocument/documentLink` request, and I think we can make it more efficient than the current implementation. Currently, this request is made after each change while typing. This is necessary if links should be underlined in the editor.

But the main interaction with links is via the hover popup which shows a "Open Link" (or "Open in Browser") link at the bottom of the popup, if applicable. But the mouse hover probably happens significantly less often than the (debounced) changes from typing, and that means if the user has disabled the underline for links in the LSP settings, we do a lot of unnecessary `textDocument/documentLink` requests. In that case it would be sufficient to only do that request on demand when the hover popup is requested. I have implemented that in this PR, including result caching if there are multiple hover actions without a buffer change in between. Also the `DocumentLink`s are only resolved (if necessary) when the link in the popup is clicked. If the underline style is used, requests are still made after each change, like before.

Currently there are 3 different options for the "link_highlight_style" setting, one of it is "disabled" which disabled the DocumentLink functionality completely. I think with this refactoring the "disabled" option isn't needed anymore, because if the underline style isn't used, the requests are only done on demand, so it should have any performance impact. In case someone really wants to disable any DocumentLink functionality, they of course still can do that via "disabled_capabilities" in the server config.

I think the on-demand-request strategy for DocumentLink is also a good reason to not use the underline style by default, so I changed that default value (probably worth to mention that in the change log, if we want to do this change).

There is only small drawback, but I think it's only an extremely niche use case; there is a `lsp.link_available` context for key bindings that would check whether there is a link on the given point (caret position). This was implemented for a special user request to use the same key binding for opening a DocumentLink and for the "Goto Definition" feature. So you could use both of these commands with the same key trigger, but prioritize `lsp_open_link` while using the mentioned "context" to fall back to "Goto Definition" if there is no link at the given point. But if we don't trigger the DocumentLink request after each change, we can't tell that anymore. So this only works like before if the "link_highlight_style" setting is set to "underline" now.

---

I've also rewritten the `lsp_open_link` command to always fetch the latest links from the server via `textDocument/documentLink` request, which is needed now in case of the underline style being disabled.

And another change from this PR is that the hover popup only uses the result from one documentLinkProvider session now (in case of multi-session setup), which means that if you click on "Open Link" and there are multiple session with that capability, we don't show the quick panel anymore to select one of the results. I assume that if there are multiple sessions, the link targets would be the same anyway, so I think we can save that additional step and the quick panel wasn't really useful.

### DIFF
--- a/LSP.sublime-settings
+++ b/LSP.sublime-settings
@@ -213,16 +213,12 @@
   // customize the color.
   "hover_highlight_style": "",
 
-  // Highlight style of links to internal or external resources, like another text
-  // document or a web site. Link navigation is implemented via the popup on mouse hover.
+  // Highlight style of links to internal or external resources, for example another
+  // file or a website. Link navigation is implemented via the popup on mouse hover.
   // Valid values are:
-  // "underline"
-  // "none" - disables special highlighting, but still allows to navigate links in the
-  //          hover popup
-  // "disabled" - disables highlighting and the hover popup for links
-  // Note that depending on the syntax and color scheme, some internet URLs may still be
-  // underlined via regular syntax highlighting.
-  "link_highlight_style": "underline",
+  // "underline" - draw links with underline decoration
+  // "" - don't draw links with underline decoration
+  "link_highlight_style": "",
 
   // Enable semantic highlighting in addition to standard syntax highlighting.
   // Note: Must be supported by the language server and also requires a special rule in
--- a/sublime-package.json
+++ b/sublime-package.json
@@ -653,11 +653,12 @@
             "link_highlight_style": {
               "enum": [
                 "underline",
-                "none",
-                "disabled"
+                "none", // For backwards-compatibility, same as ""
+                "disabled", // For backwards-compatibility, same as ""
+                ""
               ],
-              "default": "underline",
-              "markdownDescription": "Highlight style of links to internal or external resources, like another text document or a web site."
+              "default": "",
+              "markdownDescription": "Highlight style of links to internal or external resources, for example another file or a website."
             },
             "semantic_highlighting": {
               "type": "boolean",
--- a/Main.sublime-menu
+++ b/Main.sublime-menu
@@ -213,7 +213,7 @@
         "command": "lsp_symbol_references"
       },
       {
-        "caption": "LSP: Follow Link",
+        "caption": "LSP: Open Link",
         "command": "lsp_open_link"
       }
     ]
--- a/plugin/api.py
+++ b/plugin/api.py
@@ -225,7 +225,7 @@ def notification_handler(method: str) -> Callable[[Callable[[Any, P], None]], Ca
 
 def request_handler(
     method: str
-) -> Callable[[Callable[[Any, P], RequestHandlerResponse]], Callable[[Any, P, int], Promise[Response[R]]]]:
+) -> Callable[[Callable[[Any, P], RequestHandlerResponse]], Callable[[Any, P, int], Promise[Response[R]]]]:  # pyright: ignore[reportInvalidTypeVarUse]
     """
     Decorator to mark a method as a handler for a specific LSP request.
 
--- a/plugin/core/url.py
+++ b/plugin/core/url.py
@@ -20,11 +20,13 @@ import sys
 if TYPE_CHECKING:
     from ...protocol import CodeAction
     from ...protocol import Command
+    from ...protocol import DocumentLink
     from ...protocol import DocumentUri
     from ...protocol import URI
     import sublime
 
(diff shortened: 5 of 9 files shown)


sublime-package.json
```diff
@@ -653,11 +653,12 @@
             "link_highlight_style": {
               "enum": [
                 "underline",
-                "none",
-                "disabled"
+                "none", // For backwards-compatibility, same as ""
+                "disabled", // For backwards-compatibility, same as ""
```
rwols, line 657: Are comments actually allowed in this file? I guess so.

rwols (approved):

rchl:
I've noticed that we have this menu item that says "Follow link" while we've changed the text to "Open link" in 271724d3175b03a754ead54ae3fb90097f003e54

Screenshot 2026-05-30 at 20 12 18

I've also noticed that it doesn't work the same as the "Open link" in the popup. At least for custom `json-schema` scheme. The menu item tries to open that link in the browser on main while on this branch it forwards to the plugin.

The link in the popup has some additional logic that tries to parse line:column which actually is expected in this case.

That is not really a new issue though (even if this branch changes behavior). But at least one thing that could be done is rename the menu item text to "Open link"

rchl:
Though one thing that is not great is that on this branch the "Follow link" menu item is always active (at least with default setting) and usually does nothing.

EDIT: Even with underline enabled it still is always active.

jwortmann:
I've updated the menu label and added back `is_enabled` which uses the cached result if the underline style is enabled. But we still run a request when the command is run, to be sure that we always have the latest result. If underline isn't used, the menu item is always enabled (because there aren't any cached results).

> I've also noticed that it doesn't work the same as the "Open link" in the popup. At least for custom `json-schema` scheme. The menu item tries to open that link in the browser on main while on this branch it forwards to the plugin.

Hm, that is strange, I think I used more or less the same logic from the hover popup.
```python
    def _open_uri_async(self, session: Session, uri: URI) -> None:
        scheme = parse_uri(uri)[0]
        if scheme == 'file':
            if window := self.view.window():
                open_file_uri(window, uri)
        elif scheme.lower() in {'http', 'https'} or (not scheme and uri.startswith('www.')):
            open_in_browser(uri)
        else:
            session.open_uri_async(uri)
```
For unknown URI scheme it should fall back to `session.open_uri_async` but it seems that only evaluates the line:col from fragment for `file` URIs. I think we should move this part
https://github.com/sublimelsp/LSP/blob/b5d3208d0a53c30305e2df2ba178d3137693279d/plugin/hover.py#L347-L354
also into `session.open_uri_async` if we want to do that for custom schemes too and not only for `file` URIs. But I think it should actually go through the plugin handler anyway?

*Edit:* Oh, I see the problem is that we don't know from the hover popup which session to use to resolve the custom URI scheme (currently it seems to just try all sessions).

sublime-package.json
jwortmann: There already are a few comments in this file, like https://github.com/sublimelsp/LSP/blob/b5d3208d0a53c30305e2df2ba178d3137693279d/sublime-package.json#L289
So it seems to be allowed.

rchl:
> also into `session.open_uri_async` if we want to do that for custom schemes too and not only for `file` URIs. But I think it should actually go through the plugin handler anyway?

It seems like a nice thing that LSP can handle links like

Screenshot 2026-05-30 at 21 40 09

to jump to a different location and only fallback to the plugin otherwise.
Otherwise we would have to replicate that logic in plugin command handlers.
So I think we should move that logic to where it can be used in more cases.

jwortmann:
> So I think we should move that logic to where it can be used in more cases.

I agree, the `#line:col` fragment convention seems to be useful to work for all URIs, and not only `file` URIs.
But I think we can do that refactoring separately, because that behavior wasn't changed in this PR.

LSP.sublime-settings
```diff
@@ -213,16 +213,12 @@
...
-  // "underline"
-  // "none" - disables special highlighting, but still allows to navigate links in the
-  //          hover popup
-  // "disabled" - disables highlighting and the hover popup for links
-  // Note that depending on the syntax and color scheme, some internet URLs may still be
-  // underlined via regular syntax highlighting.
-  "link_highlight_style": "underline",
+  // "underline" - draw links with underline decoration
+  // "" - don't draw links with underline decoration
+  "link_highlight_style": "",
```
rchl, line 221: I will probably keep this on by default so not sure I agree with changing the default but I don't have strong enough opinion to fight for it.

rchl (approved):

PR #2936, opened 2026-05-28
fix ClientConfig equality check

rchl:
In a3bd1df the ClientConfig equality check was simplified to checking `_all_settings` but that's not correct in case someone changes `config.initialization_options` (or any other explicitly exposed property) after config was created because `_all_settings` wouldn't get updated in that case.

In practice this was maybe not a problem (or not always a problem) because before checking for equality the new config would be re-created from latest sublime settings so both `_all_settings` and `initialization_options` would change in sync.

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -768,15 +768,18 @@ class ClientConfig:
     """
 
     CONFIG_KEYS: Final[set[str]] = {
+        'auto_complete_selector',
         'command',
         'diagnostics_mode',
         'disabled_capabilities',
         'enabled',
         'env',
         'experimental_capabilities',
+        'file_watcher',
         'initialization_options',
         'markdown_language_map',
         'priority_selector',
+        'semantic_tokens',
         'selector',
         'settings',
         'tcp_port',
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -27,6 +27,7 @@ from typing import Any
 from typing import Callable
 from typing import cast
 from typing import Dict
+from typing import Final
 from typing import Generator
 from typing import Iterable
 from typing import List
@@ -766,6 +767,22 @@ class ClientConfig:
     file) are accessible through attribute access (`.foo`).
     """
 
+    CONFIG_KEYS: Final[set[str]] = {
+        'command',
+        'diagnostics_mode',
+        'disabled_capabilities',
+        'enabled',
+        'env',
+        'experimental_capabilities',
+        'initialization_options',
+        'markdown_language_map',
+        'priority_selector',
+        'selector',
+        'settings',
+        'tcp_port',
+    }
+    """Server configuration keys that we have dedicated handling for."""
+
     def __init__(
         self,
         *,
@@ -853,13 +870,13 @@ class ClientConfig:
         # Transformed mapping that uses tuples instead of lists for mdpopups.
         self.resolved_markdown_language_map: MarkdownLangMap | None = None
         self.markdown_language_map = markdown_language_map  # use the setter to populate resolved_markdown_language_map
-        # For accessing configuration keys not explicitly handled above. Accessable through dunder methods below.
         self._settings_registration = settings_registration
         if isinstance(all_settings, dict):
             self._all_settings = all_settings
-            # Exclude 'settings' because it shouldn't be considered when ClientConfig is checked for equality
-            if 'settings' in self._all_settings:
-                del self._all_settings['settings']
+            # Only retain server configuration keys that we don't have dedicated properties for.
+            for key in self.CONFIG_KEYS:
+                if key in self._all_settings:
+                    del self._all_settings[key]
         else:
             self._all_settings = {}
         self._view_status_handler = default_status_view_handler
@@ -872,6 +889,7 @@ class ClientConfig:
 
     @property
     def root_settings(self) -> dict[str, Any]:
+        """Provides access to configuration keys that are not explicitly exposed on ClientConfig."""
         return self._all_settings
 
     @property
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -1185,9 +1185,13 @@ class ClientConfig:
         return "{}({})".format(self.__class__.__name__, ", ".join(items))
 
     def __eq__(self, other: object) -> bool:
-        if not isinstance(other, ClientConfig):
+        if not isinstance(other, ClientConfig) or self.name != other.name or self._all_settings != other._all_settings:
             return False
-        return self.name == other.name and self._all_settings == other._all_settings
+        for k, v in self.__dict__.items():
+            # "settings" are not considered when checking for equality
+            if not k.startswith("_") and k != 'settings' and v != getattr(other, k):
+                return False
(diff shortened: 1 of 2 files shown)


plugin/core/types.py
```diff
@@ -1185,9 +1185,13 @@ def __repr__(self) -> str:
         return "{}({})".format(self.__class__.__name__, ", ".join(items))
 
     def __eq__(self, other: object) -> bool:
-        if not isinstance(other, ClientConfig):
+        if not isinstance(other, ClientConfig) or self.name != other.name or self._all_settings != other._all_settings:
             return False
-        return self.name == other.name and self._all_settings == other._all_settings
+        for k, v in self.__dict__.items():
+            # "settings" are not considered when checking for equality
+            if not k.startswith("_") and k != 'settings' and v != getattr(other, k):
```
predragnikolic, line 1213: is this `not k.startswith("_")` necessary?
Do we have configuration keys that start with `_`?

rchl: `_all_settings`

rchl: There is also `_enabled`.

To be fair, it would be OK to remove this condition currently (and then we wouldn't need `self._all_settings != other._all_settings` above) but the idea is that properties starting with `_` are internal and we shouldn't care about those.

rchl: Actually there is even more - `_settings_registration`, `_view_status_handler`, `_markdown_language_map`. So no, I wouldn't remove it, the idea makes sense.

plugin/core/types.py
```diff
@@ -1185,9 +1185,13 @@ def __repr__(self) -> str:
         return "{}({})".format(self.__class__.__name__, ", ".join(items))
 
     def __eq__(self, other: object) -> bool:
-        if not isinstance(other, ClientConfig):
+        if not isinstance(other, ClientConfig) or self.name != other.name or self._all_settings != other._all_settings:
```
jwortmann: The first part `name` of this addition seems redundant if it's also checked below, and the the second part `_all_settings` feels wrong if we want to define `__eq__` as a "dynamic equality". I'd say we should either decide for "dynamic" equality, i.e. later changes to the fields are considered (the change from this PR), or for "static" equality, which considers only the config from the JSON (sublime-settings) file. But with this change, ClientConfigs are only equal if they satisfy both conditions, which doesn't feel right to me.

jwortmann:
By the way, is there any good use case for allowing to modify `config.initialization_options`? If not, in my opinion this should not be / have been exposed this way as a mutable property. For example, LSP-jdtls for some reason overwrites and completely ignores the user's "initialization_options" from the `.sublime-settings` file, which did cost me some time debugging while trying to make that server work. https://github.com/sublimelsp/LSP-jdtls/blob/2e675dc8a3a9442171777323d0a42a2e5ee24475/modules/jdtls.py#L177

rchl:
> By the way, is there any good use case for allowing to modify `config.initialization_options`? If not, in my opinion this should not be / have been exposed this way as a mutable property. For example, LSP-jdtls for some reason overwrites and completely ignores the user's "initialization_options"

There are good reasons to mutate `initialization_options` (or other keys) dynamically. One could ask a question whether it should be done by mutating `ClientConfig` or by overriding request but I guess it doesn't really make a big difference for this case.

For example LSP-typescript [enables plugins dynamically](https://github.com/sublimelsp/LSP-typescript/blob/c8947e7759a01473c4f0352fec1b95b900786cdc/plugin.py#L129-L129) based on plugins it finds in the plugin host.

If LSP-jdtls does something weird then its up to it to do better.

plugin/core/types.py
rchl: Removed redundant name check.

As for `_all_settings` it's really intended to only contain properties that are not already explicitly exposed on `ClientConfig`.

rchl: I admit that it's a bit confusing now how it works - the `_all_settings` contains the original settings but those can drift out of sync with explicitly exposed `ClientConfig` properties that have the same name. The getter implementation ensured that those shadowed properties can't be accessed on `_all_settings` but since I've transitioned to `root_settings`, that's no longer the case.

I'm not sure what would be the best way forward. Current way works only that it can be confusing when it comes to the value of those shadowed keys. Perhaps we should explicitly remove those from `_all_settings` at least?

jwortmann: `_all_settings` contains entire JSON dict from the settings file, except for the "settings" key:
https://github.com/sublimelsp/LSP/blob/de4034e647ac030b222fe2f7cd0487e9a892012c/plugin/core/types.py#L971
https://github.com/sublimelsp/LSP/blob/de4034e647ac030b222fe2f7cd0487e9a892012c/plugin/core/types.py#L859-L862

So it compares static "command" / "initialization_options", etc. from the settings file too. So IMO that check should be removed, if we also want to compare the same things with dynamic values.

Btw, that last 2 lines (and comment) aren't necessary anymore after removing that check.

jwortmann: An alternative would be to keep that check, but remove all known keys from `_all_settings`. Then config would still be restarted when custom settings change.

rchl: Yes, that was my thinking too.

Though we don't really have programmatically defined list of root properties of sublime-settings config. For example ClientConfig has `name` property but it should still be allow on `all_settings` because it's not treated in any special way in sublime-settings.

rchl: 508b4572709b40021893fb411ff36d247f322b56
plus rewording:
0cff53be8f46397af8895a813b95321f5b619f86
16f70cf6988d14753fa40b7c439667711ff8b2d3

jwortmann: I didn't see your previous comment until page reload. Ok, I would probably at least remove `command` and `initialization_options` from `_all_settings`.

I think the two lines I used above could be simplified to `self._all_settings.pop('settings', None)`.

rchl: And now you also didn't see my change before posting. :) We are too fast.

plugin/core/types.py
```diff
@@ -768,15 +768,18 @@ class ClientConfig:
...
         'diagnostics_mode',
         'disabled_capabilities',
         'enabled',
         'env',
         'experimental_capabilities',
+        'file_watcher',
         'initialization_options',
         'markdown_language_map',
         'priority_selector',
+        'semantic_tokens',
```
rchl, line 782: Missed those 3 because those are not documented in https://lsp.sublimetext.io/client_configuration/

jwortmann (approved):

jwortmann:
I think we have broke the `"enabled": true/false` toggle in configs.

rchl:
Will look. Do you understand why?

rchl:
I see, because it's `_enabled`.

jwortmann:
Yes, so we either need to keep it in `_all_settings`, i.e. remove from `CONFIG_KEYS`, or it needs special handling later in `__eq__`.

rchl:
https://github.com/sublimelsp/LSP/pull/2940 - decided to use CONFIG_KEYS as a source of truth (and exclude `settings`).

PR #2937, opened 2026-05-28
Always notify correctly on config changes including in projects

rchl:
This should properly detect whether only "settings" were updated (which shouldn't restart server) or also other settings (that should result in server restart).

Should handle changes in package, global and project settings.

Fixes case reported in https://github.com/sublimelsp/LSP/pull/2932#issuecomment-4563807717:

> I think there is still another bug that if you modify a server setting in project settings, then the server is restarted (but that also happens on main, so it's not a new bug).

Should also correctly handle cases when project-only server config is removed (should shut down server) but this is probably a rather rare use case.

### DIFF
--- a/plugin/core/configurations.py
+++ b/plugin/core/configurations.py
@@ -12,7 +12,6 @@ from abc import abstractmethod
 from collections import deque
 from datetime import datetime
 from datetime import timedelta
-from typing import Any
 from typing import Generator
 from typing import Literal
 from typing import TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


rchl:
The failures are due to https://github.com/sublimelsp/LSP/pull/2936.

plugin/core/configurations.py
```diff
@@ -24,6 +24,9 @@
 RETRY_COUNT_TIMEDELTA = timedelta(minutes=3)
 
 
+ConfigChangeType = Literal['added', 'removed', 'root_changed', 'settings_changed', 'unchanged']
```
rchl, line 27: Hard to come up with good names for changes in this PR...

rchl:
Depending on how you feel about this and whether you have time to review @jwortmann we'll either include it in pending release or not. Fine either way. I just don't want to delay release more due to https://github.com/sublimelsp/LSP-copilot/issues/288#issuecomment-4573770300 for example.

jwortmann (approved):
I haven't reviewed all the changes line by line, but I have tested with a server and modifying server settings in the project settings file works correctly now.

PR #2939, opened 2026-05-29
disable tests that fail on latest github mac runners

rchl:
I think github runners update caused those tests to fail but I don't want to spend time investigating now.


rchl:
macos tests also passed in https://github.com/sublimelsp/LSP/pull/2937 without this fix but it seems that in some cases the tests run much longer (like for the commit here) and then those specific `FileWatcherDynamicTests` fail.

jwortmann (approved):

PR #2940, opened 2026-05-29
fix config equality check not considering "enabled"

rchl:
Fixes https://github.com/sublimelsp/LSP/pull/2936#issuecomment-4577979968

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -1208,9 +1208,9 @@ class ClientConfig:
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, ClientConfig) or self._all_settings != other._all_settings:
             return False
-        for k, v in self.__dict__.items():
+        for k in self.CONFIG_KEYS:
             # "settings" are not considered when checking for equality
-            if not k.startswith("_") and k != 'settings' and v != getattr(other, k):
+            if k != 'settings' and getattr(self, k) != getattr(other, k):
                 return False
         return True
 
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -808,7 +808,7 @@ class ClientConfig:
         markdown_language_map: MarkdownLangMapJson | None = None,
         path_maps: list[PathMap] | None = None,
         settings_registration: SettingsRegistration | None = None,
-        all_settings: dict[str, Any] | None = None
+        custom_config_keys: dict[str, Any] | None = None
     ) -> None:
         """
         :param name: Unique identifier for this language server.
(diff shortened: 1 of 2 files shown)


jwortmann (approved):
This should work.

But we may want to rename `_all_settings` to something like `_custom_config_keys` or `_custom_plugin_settings` now, just for clarity.

But there is one strange thing that I don't quite understand at the moment. It looks like before I made the change in https://github.com/sublimelsp/LSP/commit/a3bd1df392c366eff9b473e81a1d08f4f69084d6, the "enabled" field was also not checked for equality. Maybe there was another related change in that commit that made starting/stopping configs on "enabled" toggle work before and after, but I don't see it right now.

rchl:
> But there is one strange thing that I don't quite understand at the moment. It looks like before I made the change in [a3bd1df](https://github.com/sublimelsp/LSP/commit/a3bd1df392c366eff9b473e81a1d08f4f69084d6), the "enabled" field was also not checked for equality. Maybe there was another related change in that commit that made starting/stopping configs on "enabled" toggle work before and after, but I don't see it right now.

Because `enabled` was in `_all_settings`, which was checked for equality?

EDIT: Oh, you said **before** your change. Yeah...

plugin/core/types.py
```diff
@@ -1206,13 +1206,12 @@ def __repr__(self) -> str:
...
-        for k, v in self.__dict__.items():
+        return (
+            isinstance(other, ClientConfig)
+            and self._custom_config_keys == other._custom_config_keys
             # "settings" are not considered when checking for equality
-            if not k.startswith("_") and k != 'settings' and v != getattr(other, k):
-                return False
-        return True
+            and all(k == 'settings' or getattr(self, k) == getattr(other, k) for k in self.CONFIG_KEYS)
+        )
```
rchl, line 1214: Perhaps this way of writing reads better:

```py
        return (
            isinstance(other, ClientConfig)
            and self._custom_config_keys == other._custom_config_keys
            # "settings" are not considered when checking for equality
            and all(k == 'settings' or getattr(self, k) == getattr(other, k) for k in self.CONFIG_KEYS)
        )
```

jwortmann:
> Because `enabled` was in `_all_settings`, which was checked for equality?

Yes, but I mean before that commit. The `_all_settings` check was introduced in that commit, but before that `_enabled` would still be excluded from the equality check due to starting with `_`.

diff

plugin/core/types.py
jwortmann: Okay for me.

You should probably make a hotfix release with this fix (or re-tag before Package Control sees the release).

rchl: https://github.com/sublimelsp/LSP/releases/tag/4070-2.12.1

PR #2941, opened 2026-05-30
Deprecate process_placeholders argument for apply_text_edits function

jwortmann:
Users of this function (LSP-rust-analyzer ?) can convert the TextEdit with placeholder(s) into a SnippetTextEdit instead, which is part of the LSP 3.18 specs. It's pretty simple; just copy the `newText` into `snippet['value']` and also add `snippet['kind'] = "snippet"` - for an example see https://github.com/sublimelsp/LSP-Tinymist/commit/70e6a50771398972a5dc3f2fe57eac9b5035c6a4

Then we can remove the `LspApplyDocumentEditCommand`.

### DIFF
--- a/plugin/core/edit.py
+++ b/plugin/core/edit.py
@@ -11,6 +11,7 @@ from ...protocol import TextDocumentEdit
 from ...protocol import TextEdit
 from ...protocol import WorkspaceEdit
 from .logging import debug
+from .logging import printf
 from .promise import Promise
 from .protocol import UINT_MAX
 from typing import Dict
@@ -100,11 +101,14 @@ def apply_text_edits(
     if not edits:
         return Promise.resolve(view)
     if not view.is_valid():
-        print('LSP: ignoring edits due to view not being open')
+        printf('ignoring edits due to view not being open')
         return Promise.resolve(None)
     if process_placeholders:
-        # TODO: remove rust-analyzer specific handling for placeholders in TextEdit, because SnippetTextEdit is now part
-        # of the LSP specs.
+        # Deprecated because SnippetTextEdit is now part of the LSP 3.18 specs.
+        printf(
+            'The "process_placeholders" argument for the apply_text_edits function is deprecated.',
+            'Convert the TextEdit into a SnippetTextEdit instead.'
+        )
         view.run_command(
             'lsp_apply_document_edit',
             {
--- a/plugin/edit.py
+++ b/plugin/edit.py
@@ -19,6 +19,7 @@ from typing import Generator
 from typing import Iterable
 from typing import Tuple
 from typing import TYPE_CHECKING
+from typing_extensions import deprecated
 import operator
 import os
 import re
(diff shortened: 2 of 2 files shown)


rchl (approved):

rchl:
Note that the spec-compliant implementation doesn't work as well because it only processes topmost edit as a snippet. In rust-analyzer case the edits are for example:

```json
[
  {
    "range": {
      "start": {
        "line": 0,
        "character": 3
      },
      "end": {
        "line": 0,
        "character": 6
      }
    },
    "snippet": {
      "kind": "snippet",
      "value": "main"
    }
  },
  {
    "range": {
      "start": {
        "line": 1,
        "character": 13
      },
      "end": {
        "line": 1,
        "character": 29
      }
    },
    "snippet": {
      "kind": "snippet",
      "value": "\"Hello, world!\""
    }
  },
  {
    "range": {
      "start": {
        "line": 4,
        "character": 0
      },
      "end": {
        "line": 6,
        "character": 1
      }
    },
    "snippet": {
      "kind": "snippet",
      "value": "fn two$0() {\n    println!(\"Hello, world1!\");\n}"
    }
  }
]
```

in practice only the last one is a snippet but I guess that doesn't really matter that other are also snippets for this issue.

The effect of that is that the cursor location is not updated the way it should be.

rchl:
Spec says:

> For the active file, only one snippet can specify a cursor position. In case there are multiple snippets defining a cursor position for a given URI, it is up to the client to decide the end position of the cursor.

So it doesn't say to only process the topmost one.

I've even worked around RA saying that all edits are snippets (it uses `insertTextFormat` for that) and ended up with just a single snippet but it's still not processed:

```json
[
  {
    "range": {
      "start": {
        "line": 0,
        "character": 3
      },
      "end": {
        "line": 0,
        "character": 6
      }
    },
    "newText": "main",
    "insertTextFormat": 2
  },
  {
    "range": {
      "start": {
        "line": 1,
        "character": 13
      },
      "end": {
        "line": 1,
        "character": 29
      }
    },
    "newText": "\"Hello, world!\"",
    "insertTextFormat": 2
  },
  {
    "range": {
      "start": {
        "line": 4,
        "character": 0
      },
      "end": {
        "line": 6,
        "character": 1
      }
    },
    "snippet": {
      "kind": "snippet",
      "value": "fn tw$0o() {\n    println!(\"Hello, world1!\");\n}"
    }
  }
]
```

jwortmann:
I'll think about how we could handle that. Currently we only process the "last" (i.e. topmost) edit as snippet because that is an easy way to ensure that there is only one snippet. Also we need to check that it still works correctly if we first insert a snippet via `insert_snippet` command, and after that do additional edits to the view (for other TextEdits if the snippet is not the topmost edit).

rwols (approved):

jwortmann:
I created PR #2944, could you check whether that fixes the case from rust-analyzer?

PR #2945, opened 2026-05-31
type-check all LSP packages

rchl:
Workflow that checks out all LSP packages (known to me) + LSP + dependencies and runs basedpyright against those. The choice of type checking rules is a bit arbitrary - we want as many as possible to catch any breaking changes but we don't want to report package's own type issues as errors.

Utilizes script from https://github.com/sublimelsp/lsp_maintainers/blob/main/scripts/checkout-repos.py to check out and run type-checking. I'm using that repository since it has database of some third-party packages that are not in our main lsp repository channel.

A couple of third-party packages are ignored since those are either somewhat broken or I can't get them fixed at the moment.

### DIFF
--- a/.github/workflows/test-community-packages.yaml
+++ b/.github/workflows/test-community-packages.yaml
@@ -30,10 +30,10 @@ jobs:
           ln -s ${{ github.workspace }}/LSP lsp_maintainers/repositories/LSP
 
       - name: Setup uv
-        uses: astral-sh/setup-uv@v8.1.0
+        uses: astral-sh/setup-uv@v8.2.0
 
       - name: Sync venv and run basedpyright
         run: |
           uv sync
-          uv run basedpyright
+          uv run pyright
         working-directory: lsp_maintainers/repositories
--- a/.github/workflows/test-community-packages.yaml
+++ b/.github/workflows/test-community-packages.yaml
@@ -15,7 +15,7 @@ jobs:
         with:
           path: LSP
 
-      - name: Check out lsp_maintainers (feat/test)
+      - name: Check out lsp_maintainers
         uses: actions/checkout@v6
         with:
           repository: sublimelsp/lsp_maintainers
--- a/.github/workflows/test-community-packages.yaml
+++ b/.github/workflows/test-community-packages.yaml
@@ -27,6 +27,7 @@ jobs:
 
       - name: Use checkout of LSP from PR
         run: |
+          rm -rf ${{ github.workspace }}/LSP/tests
           ln -s ${{ github.workspace }}/LSP lsp_maintainers/repositories/LSP
 
       - name: Setup uv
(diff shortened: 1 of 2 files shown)


.github/workflows/test-community-packages.yaml
```diff
@@ -0,0 +1,39 @@
...
+        run: |
+          ln -s ${{ github.workspace }}/LSP lsp_maintainers/repositories/LSP
+
+      - name: Setup uv
+        uses: astral-sh/setup-uv@v8.1.0
+
+      - name: Sync venv and run basedpyright
+        run: |
+          uv sync
+          uv run basedpyright
```
jwortmann, line 40: Is there any advantage over running pyright for this check? We already use pyright for type checking LSP, so it might be good to use the same tool here.

rwols:
If this runs from a feature branch, is there a way to check if a feature branch with the same branch name exists, and if so, check out the feature branch, instead of the main branch?

.github/workflows/test-community-packages.yaml
rchl: No, reason, switched to pyright.

rchl:
> If this runs from a feature branch, is there a way to check if a feature branch with the same branch name exists, and if so, check out the feature branch, instead of the main branch?

Implemented that.

rchl:
I've noticed that it only checks the root python files within packages. Which is better than nothing but some packages have main files in sub-directories.

I want the behavior that checks the root files + all files that are referenced from those but apparently that's not easy to achieve.

I don't want to test all files within packages as directories like `tests` or `scripts` are not relevant to us and those WILL trigger many type issues...

.github/workflows/test-community-packages.yaml
rchl: Actually basedpyright provides better (colored) output in CI.
Otherwise the type checking is 99% the same so I would opt to use it instead.

rchl:
I've updated the logic so that the whole package is tested.

Using git archive to filter the files (according to `.gitattrbiutes`) to only ones that are included in published package.

rwols (approved):

predragnikolic:
This is just a question, but I noticed a lot of nested LSP folders:
/Users/runner/work/LSP/LSP/lsp_maintainers/repositories/LSP/uv.lock

The first LSP is probably the LSP package, the second LSP/LSP is probably the extracted zip... 
But I find this confusing path confusing 
LSP/LSP/lsp_maintainers/repositories/LSP/

Can you please just explain the folder structure? (when/if you have time)
Why do we see this many LSP folders? (Like in the path provided above)

rchl:
It's not really up to me mostly. It's actions/checkout that decides most of it. The base is always /Users/runner/work/LSP/ I guess, since the repo is called that. Then I check out LSP and lsp_maintainers in separate sub-directories.

If you look at the workflow you don't see that mess so it's not that confusing from that perspective.

PR #2947, opened 2026-06-07
Rename: protocol.Error -> protocol.ErrorResponseException

rwols:
This does trigger the [lint rule N818](https://docs.astral.sh/ruff/rules/error-suffix-on-exception-name/). I've added an exception for it in the pyproject.toml file.


rchl:
I don't want to focus too much on it given that the main PR is not done yet but you should probably consider compatibility with existing packages which this change would "destroy".

rwols:
You're right, this change is going to be bad for compatibility with existing packages... closing.

PR #2949, opened 2026-06-07
chore: fix ruff warning fallible-context-manager (RUF075)

rwols:


rchl (approved):
This is not a warning yet in the version of ruff that is declared in pyproject.toml. So ideally we'd bump the version also but then you'll also get a bunch of other errors.

rchl:
Bumped ruff, fixed same issue elsewhere and ignored https://docs.astral.sh/ruff/rules/too-many-statements-in-try-clause/ (to be looked at later potentially)

PR #2951, opened 2026-06-07
Fix errors on re-enabling disabled LSP package

rchl:
Fix issue mentioned in https://github.com/sublimelsp/LSP-basedpyright/issues/154 where disabling and then enabling LSP-* package triggers weird errors. The same issue can trigger on package updates.

The issue was that on disabling the plugin, the active settings listener from `SettingsRegistration` would trigger and [immediately re-register the external plugin](https://github.com/sublimelsp/LSP/blob/a04972b38024149f6c1873175a0f11047d3c8643/plugin/core/settings.py#L98). Fix by explicitly unregistering the listener when unregistering a plugin.

I think I've finally came up with a clean fix for the weird `SettingsRegistration` listener handling. Now `SettingsRegistration` is owned entirely by `ClientConfigs` where both registration and un-registration happens and it's not passed to `ClientConfig`s. `ClientConfig` only receives the new `SettingsStore` with `sublime.Settings` and settings path - all that it needs.

### DIFF
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -190,11 +190,7 @@ class SettingsRegistration:
 
     def __init__(self, settings: sublime.Settings, on_change: Callable[[], None]) -> None:
         self.settings = settings
-
-        def on_change_handler() -> None:
-            on_change()
-
-        self.settings.add_on_change("LSP", on_change_handler)
+        self.settings.add_on_change("LSP", on_change)
 
     def unregister(self) -> None:
         self.settings.clear_on_change("LSP")
--- a/plugin/core/types.py
+++ b/plugin/core/types.py
@@ -199,12 +199,6 @@ class SettingsRegistration:
     def unregister(self) -> None:
         self.settings.clear_on_change("LSP")
 
-    def __del__(self) -> None:
-        # FIXME: Relying on reference counts and garbage collection timings for the change listener handling is a very
-        # bad idea, because instances of this class are dynamically passed between other object instances, and it can
-        # easily cause bugs like https://github.com/sublimelsp/LSP/pull/2867
-        self.settings.clear_on_change("LSP")
-
 
 class DebouncerNonThreadSafe:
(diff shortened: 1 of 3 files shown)


plugin/core/types.py
rchl: I have not removed `__del__` implementation because there is [one place](https://github.com/sublimelsp/LSP/blob/2242c6870aa41ec17917d057978ac3972ba9ca30/plugin/documents.py#L220-L220) that doesn't explicitly unregister the listener.

I could try to make things more explicit there but I'm a bit afraid of touching it.

rchl: Changed `DocumentSyncListener` to explicitly unregister and removed `__del__`.

Also added cc083663bf0ad706e56d961ea962d221379ae92c which might not strictly be needed but it's good for ensuring that cleanup happens while `DocumentSyncListener` is still alive rather than at some arbitrary time when GC triggers.

jwortmann (approved):

rwols:
It's an improvement! But I still see a discrepancy between "LSP: Enable/Disable language server globally" and "Package Control: Enable/Disable Package":

* The "enable/disable language server globally" command seems to work the way I'd expect: the session is detached and stopped for all tabs. And when enabling it, the session is attached again. No closing, and then reopening, of a tab, is necessary.
* The "enable/disable package" command: you can disable something like LSP-ruff, and it detaches the session from the tabs. But when then re-enabling it, I have to first close the tab, and then reopen it. Closing and reopening tabs is necessary for all open tabs.

rchl:
I would guess some state is not cleaned/reset properly in `DocumentSyncListener`.
But it fixes what it claims to fix - the errors. :)

rchl:
Unfortunately this triggers https://github.com/sublimehq/sublime_text/issues/6868 sometimes when re-saving `LSP/boot.py`. I will try to reproduce this consistently and find a fix or workaround.

PR #2954, opened 2026-06-13
add focus state to minihtml tree view

rchl:
The idea: implement focus handling in tree view (used by call hierarchy) so that it's possible to navigate quickly using keyboard. I kinda want this functionality for another use case that doesn't exist yet (navigating symbol list) but it would be nice to have for call hierarchy also.

Problem with implementing keyboard navigation for call hierarchy is that clicking a caller focuses another view which is interfering with being able to navigate quickly. This is not a solved problem and it means that supporting keyboard navigation in it might not be feasible (though I haven't thought hard enough about it yet). That said, supporting focus state would be useful anyway.

I've included some keybindings and a command for navigating using keyboard for testing but those are not meant to be integrated in their current form, most likely. This PR mostly just concerns adding focus state. The keyboard navigation is what it allows to build on top of that but this PR is not really meant to add that, it's here just to be able to test the focus state.

https://github.com/user-attachments/assets/eec1ece7-821b-4a65-a018-94ccfd6fb08a

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -244,41 +244,41 @@
         "args": {"async": true},
         "context": [{"key": "lsp.session_with_capability", "operand": "textDocumentSync.willSave | textDocumentSync.willSaveWaitUntil | codeActionProvider.codeActionKinds | documentFormattingProvider | documentRangeFormattingProvider"}]
     },
-    // Navigate in Tree Views
-    {
-        "keys": ["left"],
-        "command": "lsp_handle_tree_view_action",
-        "args": {"action": "move_left"},
-        "context": [{"key": "lsp.tree_view_sheet_has_focus"}]
-    },
-    {
-        "keys": ["right"],
-        "command": "lsp_handle_tree_view_action",
-        "args": {"action": "move_right"},
-        "context": [{"key": "lsp.tree_view_sheet_has_focus"}]
-    },
-    {
-        "keys": ["up"],
-        "command": "lsp_handle_tree_view_action",
-        "args": {"action": "move_up"},
-        "context": [{"key": "lsp.tree_view_sheet_has_focus"}]
-    },
-    {
-        "keys": ["down"],
-        "command": "lsp_handle_tree_view_action",
-        "args": {"action": "move_down"},
-        "context": [{"key": "lsp.tree_view_sheet_has_focus"}]
-    },
-    {
-        "keys": ["enter"],
-        "command": "lsp_handle_tree_view_action",
-        "args": {"action": "activate"},
-        "context": [{"key": "lsp.tree_view_sheet_has_focus"}]
-    },
-    {
-        "keys": ["escape"],
-        "command": "lsp_handle_tree_view_action",
-        "args": {"action": "close"},
-        "context": [{"key": "lsp.tree_view_sheet_has_focus"}]
-    },
+    // Navigate in Tree Views (pending keybinding support in Sheets - https://github.com/sublimehq/sublime_text/issues/5377)
+    // {
+    //     "keys": ["left"],
+    //     "command": "lsp_handle_tree_view_action",
+    //     "args": {"action": "move_left"},
+    //     "context": [{"key": "lsp.tree_view_sheet_has_focus"}]
+    // },
+    // {
+    //     "keys": ["right"],
+    //     "command": "lsp_handle_tree_view_action",
+    //     "args": {"action": "move_right"},
+    //     "context": [{"key": "lsp.tree_view_sheet_has_focus"}]
+    // },
+    // {
+    //     "keys": ["up"],
+    //     "command": "lsp_handle_tree_view_action",
+    //     "args": {"action": "move_up"},
+    //     "context": [{"key": "lsp.tree_view_sheet_has_focus"}]
+    // },
+    // {
+    //     "keys": ["down"],
+    //     "command": "lsp_handle_tree_view_action",
+    //     "args": {"action": "move_down"},
+    //     "context": [{"key": "lsp.tree_view_sheet_has_focus"}]
+    // },
+    // {
+    //     "keys": ["enter"],
+    //     "command": "lsp_handle_tree_view_action",
+    //     "args": {"action": "activate"},
+    //     "context": [{"key": "lsp.tree_view_sheet_has_focus"}]
+    // },
+    // {
+    //     "keys": ["escape"],
+    //     "command": "lsp_handle_tree_view_action",
+    //     "args": {"action": "close"},
+    //     "context": [{"key": "lsp.tree_view_sheet_has_focus"}]
+    // },
 ]
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -270,6 +270,12 @@
         "context": [{"key": "lsp.tree_view_sheet_has_focus"}]
     },
     {
+        "keys": ["enter"],
+        "command": "lsp_handle_tree_view_action",
+        "args": {"action": "activate"},
+        "context": [{"key": "lsp.tree_view_sheet_has_focus"}]
+    },
+    {
         "keys": ["escape"],
         "command": "lsp_handle_tree_view_action",
         "args": {"action": "close"},
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -1,4 +1,43 @@
 [
+    // TODO: Only for testing.
+    {
+        "keys": [
+            "left"
+        ],
+        "command": "lsp_move_focus",
+        "args": {
+            "direction": "left"
+        },
+    },
+    {
+        "keys": [
+            "right"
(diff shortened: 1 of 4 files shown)


rwols (approved):

rchl:
(forgot to add test keybindings)

jwortmann:
It's not really clear from the description and code comments which parts of this PR are intended to be included in LSP, but a few things that I saw from a quick look over the code are:
- instead of using `is_enabled`, I think LSP should better provide a context `lsp.tree_view_sheet_has_focus`, which I think would be possible from https://github.com/sublimelsp/LSP/blob/8b579d895e0528ee985ce69a8d5d8d988b7529d1/boot.py#L226
- that context should be enabled for all instances of https://github.com/sublimelsp/LSP/blob/8b579d895e0528ee985ce69a8d5d8d988b7529d1/plugin/core/tree_view.py#L124-L125 and not only for "Call Hierarchy" (we also have a "Type Hierarchy" implementation)
- I would probably not add the "use ... to move around" line, because there also doesn't exist such a hint in the ST sidebar. But that's not too important to me, because space isn't really a limiting factor in the TreeViewSheet.

rchl:
The idea was to not include key bindings and the command for navigating.
That said, I could make it so that navigating using keyboard doesn't "click" the item but just activates it and then make enter trigger "click". Then including those would be fine.

rchl:
> * instead of using `is_enabled`, I think LSP should better provide a context `lsp.tree_view_sheet_has_focus`, which I think would be possible from

I thought about it but couldn't think of pros and cons of either.
I thought that for key bindings it doesn't matter and for potential menu items, it would even be more correct to use "is_enabled".

jwortmann:
> I thought that for key bindings it doesn't matter and for potential menu items, it would even be more correct to use "is_enabled".

Yes, but we don't have any menu item for up/down/left/right (LspMoveFocusCommand) and hopefully also won't add that in the future.

Adding key bindings for such essential keys up/down/left/right without any context restriction would feel too dangerous from my side. I doubt that something like that would be accepted if this were a new package for https://github.com/sublimehq/package_control_channel


> That said, I could make it so that navigating using keyboard doesn't "click" the item but just activates it and then make enter trigger "click". Then including those would be fine.

I haven't tested the feature, but that seems like a better idea. That is also the behavior in the sidebar.

rchl:
> Adding key bindings for such essential keys up/down/left/right without any context restriction would feel too dangerous from my side. I doubt that something like that would be accepted if this were a new package for [sublimehq/package_control_channel](https://github.com/sublimehq/package_control_channel?rgh-link-date=2026-06-17T17%3A44%3A15Z)

I am not sure but I think is_enabled and context have essentially equivalent behavior so I don't think that is_enabled is more dangerous. As I've realized in https://github.com/sublimelsp/LSP-json/issues/231#issuecomment-4595712017, a context doesn't have any advantage if the command is not available for some reason - ST still wouldn't ignore it.

That said, I'm not against changing to context, if only for better transparency.

rchl:
> * instead of using `is_enabled`, I think LSP should better provide a context `lsp.tree_view_sheet_has_focus`, which I think would be possible from

1. HtmlSheets don't have views so `ViewListener` doesn't trigger for those.
2. `is_enabled` is indeed not working as expected - keybinding is not ignored when it returns false.

jwortmann:
>     1. HtmlSheets don't have views so `ViewListener` doesn't trigger for those.

EventListener should be triggered when checking context for key bindings executing a WindowCommand, I assume. But I haven't tested that.

>     2. `is_enabled` is indeed not working as expected - keybinding is not ignored when it returns false.

That was what I assumed. Because the key binding from the sublime-keymap file would just override the corresponding bindings from the Default package, if not restricted by a context.

rchl:
> EventListener should be triggered when checking context for key bindings executing a WindowCommand, I assume. But I haven't tested that.

No, `on_query_context` in `ViewListener` has a `view` argument so it triggers in context of a view (which is not present here).

HtmlSheets are half baked... I don't see any way to ensure that key binding can only trigger in their context.

rchl:
https://github.com/sublimehq/sublime_text/issues/5377#issuecomment-4703190437

rchl:
Pushed my keybinding updates which DON'T WORK given the above. I will have to remove those from the PR but maybe ST will wake up one day and make it possible...

Default.sublime-keymap
```diff
@@ -244,4 +244,41 @@
...
+        "command": "lsp_handle_tree_view_action",
+        "args": {"action": "activate"},
+        "context": [{"key": "lsp.tree_view_sheet_has_focus"}]
+    },
+    {
+        "keys": ["escape"],
+        "command": "lsp_handle_tree_view_action",
+        "args": {"action": "close"},
+        "context": [{"key": "lsp.tree_view_sheet_has_focus"}]
+    },
```
rchl: Since those use context that is never set, we could keep those in the code, waiting for better times?

plugin/core/tree_view.py
```diff
@@ -58,13 +61,12 @@ def __init__(
...
-        self.command_url = command_url
-        """ A HTML embeddable URL for a command that should be executed when the tree item label is clicked.
-        Use the sublime.command_url function to generate this URL. """
+        self.action_command = action_command
+        """ The command and its arguments to be executed when the tree item label is clicked. """
         self.collapsible_state = TreeItemCollapsibleState.COLLAPSED
         self.id = str(uuid.uuid4())
 
-    def html(self, sheet_name: str, indent_level: int) -> str:
+    def html(self, sheet_name: str, indent_level: int, is_active: bool = False) -> str:
```
jwortmann: Maybe `has_focus` is better?

plugin/core/tree_view.py
```diff
@@ -261,17 +330,33 @@ def _update_contents(self) -> None:
                 color: color(var(--foreground) alpha(0.6));
                 padding-left: 0.5rem;
             }}
+            .active {{
+                background-color: color(var(--accent));
```
jwortmann: I think something like `color(var(--accent) alpha(0.3))` would be better because otherwise the text foreground might not have sufficient contrast to the accent background color.

Default.sublime-keymap
jwortmann: I would probably comment them out for now. It seems that ST considers the context as "false" if there is no plugin that handles this context key. If we want to keep the key bindings, maybe we should already add a handler like
```python
def on_query_context(self, view: sublime.View, key: str, operator: sublime.QueryOperator, operand: str, match_all: bool) → bool | None:
    if key == 'lsp.tree_view_sheet_has_focus':
        return False
    return None
```

rchl: Initially both commented out and added `on_query_context` but then removed `on_query_context` because I don't think that it would work this way when ST fixes things. I think we would need a separate `SheetListener` or something.

jwortmann: They could make it work for EventListener and just pass `view = None` into this method. But yes, this suggestion was more for the case if key bindings are not commented out. Now it's not necessary and probably better for performance to not implement this method.

rchl: That would be a breaking change type-wise. And not only type wise since existing code could start throwing.

jwortmann: Okay true, but then instead they could pass a view that is not valid. I believe that should work and wouldn't really be a breaking change. Although there might be existing plugins which perhaps would not handle that case correctly. But getting a view with `view.is_valid() == False` already happens when you open an image file using `window.open_file("path/to/example.png")`, so that would not be an unprecedented practice.

Default.sublime-keymap
```diff
@@ -244,4 +244,41 @@
         "args": {"async": true},
         "context": [{"key": "lsp.session_with_capability", "operand": "textDocumentSync.willSave | textDocumentSync.willSaveWaitUntil | codeActionProvider.codeActionKinds | documentFormattingProvider | documentRangeFormattingProvider"}]
     },
+    // Navigate in Tree Views (pending keybinding support in Sheets - https://github.com/sublimehq/sublime_text/issues/5377)
```
jwortmann, line 247: By the way. Even better than this would be if ST adds support for a sidebar API that allows plugins to create additional sidebar "tabs", like in Sublime Merge. That would probably require some more implementation effort, so my hopes aren't too high, but it is currently the 4th-most upvoted issue in the issue tracker. This entire TreeViewSheet implementation is just a workaround for the missing sidebar API functionality, and I think the UI/UX could be nicer if that was possible (it would save half of the editor window space, which is currently occupied by the TreeViewSheet). Perhaps we should formulate a desired API design for our use case and post that in the issue tracker, e.g. file and folder icons should be replaced with the ST kind icons, and that we need callbacks for expanding an item (for sending LSP request and dynamically update child nodes, unless cached result is already available) and for selecting an item (open file at given symbol location).

jwortmann (approved):

PR #2955, opened 2026-06-14
Allow pylsp and pyrefly to work with this repository

rwols:
This adds a few lines of configuration so that:

* pylsp does not complain about line length being >79 chars
* pyrefly works

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -150,3 +150,12 @@ ignore = [
     "TRY003",  # https://docs.astral.sh/ruff/rules/raise-vanilla-args/
     "TRY301",  # https://docs.astral.sh/ruff/rules/raise-within-try/
 ]
+
+[tool.pyrefly]
+python-version = "3.8.0"
+site-package-path = ["./stubs"]
+infer-with-first-use = false
+
+[tool.pyrefly.errors]
+bad-override = "ignore"
+missing-source = "ignore"
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,0 @@
-[pycodestyle]
-max_line_length = 120
-max_doc_length = 120
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,3 @@
+[pycodestyle]
+max_line_length = 120
+max_doc_length = 120


setup.cfg
```diff
@@ -0,0 +1,3 @@
+[pycodestyle]
+max_line_length = 120
+max_doc_length = 120
```
rchl: I don't see the point of adding that configuration if we are not actively using pycodestyle in the project. We are using ruff which is a much faster alternative that implements the same rules.

Also, if you really want to set those, you can do that either globally in LSP-pylsp config or in project settings, like

```
    "settings": {
        "pylsp.plugins.pycodestyle.maxLineLength": 120
     }
```

pyproject.toml
```diff
@@ -150,3 +150,12 @@ ignore = [
...
 ]
+
+[tool.pyrefly]
+python-version = "3.8.0"
+site-package-path = ["./stubs"]
+infer-with-first-use = false
+
+[tool.pyrefly.errors]
+bad-override = "ignore"
+missing-source = "ignore"
```
rchl, line 161: I would be fine with introducing those as a means of evaluating different type checkers (there is also `ty`). I'm not super confident that pyright has a bright future.

rchl: That said, please adapt the configuration so that there is not a ton of errors when running `pyrefly check`. Ignoring errors in tests and third_party would be a start.

setup.cfg
rwols: This is just to remove some basic friction when browsing the code with LSP-pylsp enabled… I don’t think we should engage in all of the CI strictness.

pyproject.toml
rwols: This is just to remove some initial friction when browsing the repository with a pyrefly config enabled. But I could add some disabled checks. I actually haven’t ran pyrefly in lint mode from the CLI for this. Nor do I think we should engage further in adding this to CI.

setup.cfg
rchl: Who needs this legacy config?

rchl (approved):

PR #2966, opened 2026-06-20
fix: server starts in wrong dir

skarasov:
Fix mentioned in sublimelsp/LSP-ty#60.

For `LSP-ty` and `LSP-ruff`. If open folder from ST menu or start subl . in file explorer - servers starts in working/project dir:
`LSP: starting  ['C:\\Users\\...\\LSP-ty\\v0.0.50\\ty.exe', 'server'] in D:\working_dir`

If open file (eg. double click in file explorer or Open file from ST menu) - servers starts in installation dir of ST:
`LSP: starting ['C:\\Users\\...\\LSP-ty\\v0.0.50\\ty.exe', 'server'] in C:\Program Files\Sublime Text`
The server is expected to run in the directory of the file being edited for full functionality.
If the server is started in the wrong directory, not all functions will work (eg. `rename` or `completeFunctionParentheses`).

Maybe the same fix is ​​needed for plugin/tooling.py?

### DIFF
--- a/plugin/core/windows.py
+++ b/plugin/core/windows.py
@@ -66,6 +66,7 @@ from weakref import ref
 from weakref import WeakSet
 import functools
 import json
+import os
 import sublime
 import threading
 
@@ -295,7 +296,11 @@ class WindowManager(Manager, WindowConfigChangeListener, ViewStatusHandler):
             workspace_folders = sorted_workspace_folders(self._workspace.folders, file_path)
             plugin_class = get_plugin(config.name)
             variables = extract_variables(self._window)
-            cwd = workspace_folders[0].path if workspace_folders else None
+            cwd = (
+                workspace_folders[0].path
+                if workspace_folders
+                else (os.path.dirname(file_path) if file_path else None)
+            )
             context = OnPreStartContext(config, variables, initiating_view, cwd, workspace_folders)
             if plugin_class:
                 if issubclass(plugin_class, LspPlugin):


rchl:
As I said, server should not really be concerned about the working directory. Nothing in the LSP spec says that it has any meaning.

If server relies on that then it feels like a random server quirk.

That said, I don't see anything wrong with taking your fix in but note that it's not really gonna fix this reliably because working directory is only decided when the server starts. If you then open file from a completely different directory then it will attach to the existing server instance and you'll have the same issue there. 

I think that it would be more correct to report this to ruff/ty and let them figure out the logic for handling that in LSP mode. They could infer directory based on the file that makes the request, for example.

jwortmann:
Plugin can set working directory if desired, so shouldn't this be handled by LSP-ty / LSP-ruff? See for example https://github.com/sublimelsp/LSP-julia/blob/6575c619fe9604845ddbd0ef0b542e7943b662d5/plugin.py#L260-L268

If the single Python file is part of a Python project, wouldn't ty/ruff expect the working directory to be the directory that contains `pyproject.toml`, and not the file directory?

skarasov:
Of course, the server must respect the working directory. You should run different servers for different working directories (as is done in Zed). How do you format different projects with different settings if you only have one?

skarasov:
> Plugin can set working directory if desired, so shouldn't this be handled by LSP-ty / LSP-ruff? See for example https://github.com/sublimelsp/LSP-julia/blob/6575c619fe9604845ddbd0ef0b542e7943b662d5/plugin.py#L260-L268
> 
> If the single Python file is part of a Python project, wouldn't ty/ruff expect the working directory to be the directory that contains `pyproject.toml`, and not the file directory?

Correct, but if file is just a script (eg. shell script, not part of a project)?

jwortmann:
File URIs used in LSP have absolute paths, so working directory shouldn't really matter, unless the server relies on some project configuration file like `Project.toml` / `pyproject.toml` / `Cargo.toml` etc.
And if there are multiple such files it might require server/language-specific logic to decide which one to pick.

Server instances are started for individual windows, because we use sidebar folders as the workspace folders.

skarasov:
>File URIs used in LSP have absolute paths, so working directory shouldn't really matter

However, the current implementation explicitly sets the cwd to None (when opening only a file, not a directory), which results in the server receiving an incorrect dir.
Both the `ty` and `ruff` do not work correctly if the server is started in the wrong directory as decribed in first post.

rchl:
LSP-ty/ruff could override working directory based on presence of pyproject.toml (for example). neovim does that too, I believe. But again, this feels a bit like a hack since it won't handle cases when then file from another project is opened in the same window. Unless LSP would handle those as two separate server instances which would be a bigger change in LSP itself.

skarasov:
Moved to sublimelsp/LSP-ty#65 and sublimelsp/LSP-ruff#162

PR #2967, opened 2026-06-23
Fix return default on non-existing key

### LINKED ISSUE #218: Latest release broken
This code was added in 1.21.1, but throws an error on my system:

```
Unable to initialize language server for LSP-rust-analyzer
  File "/Users/lina/Library/Application Support/Sublime Text/Installed Packages/LSP.sublime-package/plugin/core/windows.py", line 303, in start_async
    plugin_class.on_pre_start_async(context)
  File "/Users/lina/Library/Application Support/Sublime Text/Installed Packages/LSP-rust-analyzer.sublime-package/plugin.py", line 142, in on_pre_start_async
    context.configuration.initialization_options.update(legacy_settings)
  File "/Users/lina/Library/Application Support/Sublime Text/Installed Packages/LSP.sublime-package/plugin/core/collections.py", line 153, in update
    for key, value in d.items():
AttributeError: 'NoneType' object has no attribute 'items'
```

I guess it only works if some settings component existed that does not for me?

asl97:
Fixes https://github.com/sublimelsp/LSP-rust-analyzer/issues/218

Feel free to edit as require for style/lint.

I left the `current = current.get(key)` alone as I am not sure if changing that to `current = current[key]` would be considered as `unrelated changes`.

This keep the return None if the value is actually an None value instead of replacing with default as per previous function, pre-https://github.com/sublimelsp/LSP/commit/7981a132ce889cfa0967372805b6909483a3c064

### DIFF
--- a/plugin/core/collections.py
+++ b/plugin/core/collections.py
@@ -55,8 +55,8 @@ class DottedDict:
         current: Any = self._d
         keys = path.split('.')
         for key in keys:
-            if isinstance(current, dict):
-                current = current.get(key)
+            if isinstance(current, dict) and key in current:
+                current = current[key]
             else:
                 return default
         return current


jwortmann:
Makes sense. I'd say the suggested `current[key]` change instead of `current.get(key)` would be preferred now.

asl97:
Replaced `current.get(key)` with `current[key]` as recommend.

jwortmann (approved):

jwortmann:
Thanks

PR #2972, opened 2026-06-30
chore: ignore LSP-vale-ls when testing packages

rchl:
~~We don't use that type anywhere but not having proper type results in an error when running the `test-community-packages.yaml` workflow because LSP-copilot [bundles its own](https://github.com/sublimelsp/LSP-copilot/blob/843a9006f440a2fadc15e8ec9105ed9770dca993/plugin/types.py#L36-L39) (matching our new one) type for Layout.~~

Just ignore one package whose maintainer is inactive.

### DIFF
--- a/.github/workflows/test-community-packages.yaml
+++ b/.github/workflows/test-community-packages.yaml
@@ -22,7 +22,13 @@ jobs:
           path: lsp_maintainers
 
       - name: Check out community package repositories
-        run: python3 scripts/checkout-repos.py --preferred-branch ${{ github.head_ref }} --exclude LSP --exclude LSP-pyvoice --exclude WolframLanguage
+        run: |
+          python3 scripts/checkout-repos.py \
+            --preferred-branch ${{ github.head_ref }} \
+            --exclude LSP \
+            --exclude LSP-pyvoice \
+            --exclude LSP-VHDL-ls \
+            --exclude WolframLanguage
         working-directory: lsp_maintainers
 
       - name: Use checkout of LSP from PR
--- a/.github/workflows/test-community-packages.yaml
+++ b/.github/workflows/test-community-packages.yaml
@@ -28,7 +28,6 @@ jobs:
             --exclude LSP \
             --exclude LSP-pyvoice \
             --exclude LSP-vale-ls \
-            --exclude LSP-VHDL-ls \
             --exclude WolframLanguage
         working-directory: lsp_maintainers
 
--- a/.github/workflows/test-community-packages.yaml
+++ b/.github/workflows/test-community-packages.yaml
@@ -27,6 +27,7 @@ jobs:
             --preferred-branch ${{ github.head_ref }} \
             --exclude LSP \
             --exclude LSP-pyvoice \
+            --exclude LSP-vale-ls \
             --exclude LSP-VHDL-ls \
             --exclude WolframLanguage
         working-directory: lsp_maintainers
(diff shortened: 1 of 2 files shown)


.github/workflows/test-community-packages.yaml
```diff
@@ -22,7 +22,13 @@ jobs:
           path: lsp_maintainers
 
       - name: Check out community package repositories
-        run: python3 scripts/checkout-repos.py --preferred-branch ${{ github.head_ref }} --exclude LSP --exclude LSP-pyvoice --exclude WolframLanguage
+        run: |
+          python3 scripts/checkout-repos.py \
+            --preferred-branch ${{ github.head_ref }} \
+            --exclude LSP \
+            --exclude LSP-pyvoice \
+            --exclude LSP-VHDL-ls \
```
rchl: Ignored `LSP-VHDL-ls` because I can't be bothered checking packages that authors [don't bother maintaining](https://github.com/martinbarez/LSP-VHDL-ls/pull/2/changes).

stubs/sublime.pyi
```diff
@@ -1,7 +1,13 @@
...
 from enum import IntEnum, IntFlag
 from typing import Any, Callable, Iterable, Iterator, Literal, Optional, Reversible, Sequence
-from typing_extensions import deprecated
+from typing_extensions import TypedDict, deprecated
+
+
+class Layout(TypedDict):
+    cells: list[list[int]]
+    cols: list[float]
+    rows: list[float]
```
jwortmann: IMO this shouldn't go into this file, because it's not part of the `sublime` module. You could create a new stub file for custom types like `stub/custom.pyi`. Maybe also prefix custom types with an underscore.

I don't really understand why it is needed? Because this type doesn't seem to exist in reality: https://www.sublimetext.com/docs/api_reference.html#sublime.Window.layout

Iirc all types in this stub file are derived directly from the original `sublime.py` module, so you can be sure that they do actually exist, so for example when using the types via auto-complete you can be sure that they won't cause any runtime errors.

jwortmann (changes requested):

stubs/sublime.pyi
rchl: Why it's needed? Because `window.set_layout()` takes `dict[str, Any]` (or `dict[str, Value]`, don't remember) but LSP-copilot passes an object that has its Layout type (like the one here). So that creates type issue as those don't match. If our stubs match LSP-copilot then those two will match.

But I agree with your other points that it's not ideal this way.

jwortmann: Isn't the `StLayout` type from LSP-copilot a subtype of `dict[str, Any]`, so there shouldn't be a type issue in the described case? But maybe I misremember how the type inference with TypedDict works.

I'm not sure even whether adding the custom type just to satisfy the type checker would really be worth it. Because currently our stub files directly correspond to the definitions in the original modules, and they are easily portable, for example to be used in other ST packages. If we add a custom type definitions and corresponding imports, that doesn't work anymore so easily (would require to copy the other file too).

So actually I'd prefer if either LSP-copilot would align with our stub files, or this problem gets solved somehow with a configuration rule for the type checker (e.g. exclude that rule for LSP-copilot if that is possible, or add `type: ignore` comments on the lines where it is used in LSP-copilot).

rchl:
```
Argument of type "StLayout" cannot be assigned to parameter "layout" of type "dict[str, Any]" in function "set_layout"
  "StLayout" is not assignable to "dict[str, Any]" (basedpyright[reportArgumentType])
```

rchl: LSP-copilot could align with our types but that would sacrifice the strictness of "layout" object.

I don't immediately see a perfect solution for that but I'll think about it.

rchl: Actually ST's 3.14 types even have `dict[str, Value]` which is even more annoying to deal with.

jwortmann (approved):

PR #2981, opened 2026-07-06
chore: add integration test for workspace/textDocumentContent

rwols:
This teaches our fake server in tests/server.py to execute a server->client request when receiving a command. With this, we can set up an integration test that makes the fake server do a `window/showDocument` request when receiving a `workspace/executeCommand` request. In turn, that will then cause the client to make a `workspace/textDocumentContent` request.

So the request flow is:

1. Client->Server: `workspace/executeCommand`
2. Server->Client: `window/showDocument`
3. Client->Server: `workspace/textDocumentContent`
4. server resolves `workspace/textDocumentContent`
5. client resolves `window/showDocument`
6. server resolves `workspace/executeCommand`

I also ported along the `$test/getAndClearUnusedMockResponses` method, but that remains unused for now.

### DIFF
--- a/tests/server.py
+++ b/tests/server.py
@@ -18,7 +18,6 @@ Tests can await this request to make sure that they receive notification before 
 resumes (since response to request will arrive after requested notification).
 
 To make the server do a request when running a Command, send the $test/setupCommandAction request.
-
 """
 from __future__ import annotations
 
--- a/tests/server.py
+++ b/tests/server.py
@@ -17,6 +17,8 @@ with expected notification method in params['method'] and params in params['para
 Tests can await this request to make sure that they receive notification before code
 resumes (since response to request will arrive after requested notification).
 
+To make the server do a request when running a Command, send the $test/setupCommandAction request.
+
 """
 from __future__ import annotations
 
@@ -174,6 +176,7 @@ class Session:
         self._responses: list[tuple[str, PayloadLike]] = []
         self._received: dict[str, PayloadLike] = {}
         self._received_cv = asyncio.Condition()
+        self._command_actions: dict[str, PayloadLike] = {}
 
         self._install_handlers()
 
@@ -323,6 +326,7 @@ class Session:
 
     def _install_handlers(self) -> None:
         self._on_request("initialize", self._initialize)
+        self._on_request("workspace/executeCommand", self._execute_command)
         self._on_request("shutdown", self._shutdown)
         self._on_notification("exit", self._on_exit)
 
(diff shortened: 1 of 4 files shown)


tests/setup.py
```diff
@@ -259,6 +260,37 @@ def error_handler(params: Any) -> None:
         self.session.send_request(Request("$test/setResponses", payload), handler, error_handler)
         yield from self.await_promise(promise)
 
+    def mock_command_action(self, command_name: str, action: LSPObject) -> Generator:
```
rchl: You can probably use `ServerRequest` type for `action`.
If not then create a more generic type that takes `method` and `params` keys.

tests/server.py
```diff
@@ -17,6 +17,8 @@
 Tests can await this request to make sure that they receive notification before code
 resumes (since response to request will arrive after requested notification).
 
+To make the server do a request when running a Command, send the $test/setupCommandAction request.
+
 """
```
rchl, line 21:
```suggestion
To make the server do a request when running a Command, send the $test/setupCommandAction request.
"""
```

tests/server.py
```diff
@@ -347,6 +353,42 @@ async def _set_responses(self, params: PayloadLike) -> PayloadLike:
...
+            raise Error(ErrorCode.InvalidParams, "expected params to be a dictionary")
+        command_name = params.get("command")
+        if command_name is None:
+            raise Error(ErrorCode.InvalidParams, 'expected "command" key')
+        action = self._command_actions.pop(command_name, None)
+        if isinstance(action, dict):
+            request_method = action.get("method")
+            if not isinstance(request_method, str):
+                raise Error(ErrorCode.InvalidParams, 'expected action to have a "method" key')
+            return await self.request(request_method, action.get("params"))
```
rchl, line 373: Shouldn't we await the request but return the mocked response still, instead of that response's result?

rwols: Nah, that gives the test no interesting information. The params for the Server->Client request are known in the test. It's more interesting to return what the client returned as response to that Server->Client request.

rwols: You can see in the new test that it checks what the `@request_handler('window/showDocument')` handler in `Session` returned:

```python
self.assertEqual(promise.result(), {"success": True})
```

tests/setup.py
```diff
@@ -259,6 +261,35 @@ def error_handler(params: Any) -> None:
         self.session.send_request(Request("$test/setResponses", payload), handler, error_handler)
         yield from self.await_promise(promise)
 
+    def mock_command_action(self, command_name: str, action: ServerRequest) -> Generator:
```
rchl: I wouldn't call it "mocking". That would be more when we would force some response without going through the server. Maybe `set_command_response_action` would be more consistent with other server methods. Or just `set_command_response`.

PR #2984, opened 2026-07-08
Add optional argument severity_level for diagnostics navigation

### LINKED ISSUE #2983: `severity_level` for `lsp_prev_diagnostic` and `lsp_next_diagnostic`
`lsp_goto_diagnostic` support `severity_level` to filter based on the error level, can `lsp_prev_diagnostic` and `lsp_next_diagnostic` support it too?

Also a new argument `select`, that will select the error instead of moving the cursor before would be nice (we have the region info, so it should be easy)

jwortmann:
Closes #2983

Align the behavior of the jump to next/previous diagnostic commands with the "Goto Diagnostic" quick panel (only jumps to visible diagnostics by default).

### DIFF
--- a/docs/src/keyboard_shortcuts.md
+++ b/docs/src/keyboard_shortcuts.md
@@ -66,15 +66,15 @@ Below is a list of supported commands and the corresponding keyboard shortcut (i
 | Format Selection | unbound | `lsp_format_document_range`
 | Goto Declaration | unbound | `lsp_symbol_declaration`
 | Goto Definition | unbound<br>suggested: <kbd>F12</kbd> | `lsp_symbol_definition`
-| Goto Diagnostic | unbound<br>suggested: <kbd>F8</kbd> | `lsp_goto_diagnostic`<br>Supports optional argument `"severity_level"` for the severity level of included diagnostics: 1 (error), 2 (warning), 3 (information), 4 (hint). Uses the `"show_diagnostics_severity_level"` setting if not provided.
 | Goto Implementation | unbound | `lsp_symbol_implementation`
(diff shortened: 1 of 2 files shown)


docs/src/keyboard_shortcuts.md
```diff
@@ -73,8 +73,8 @@ Below is a list of supported commands and the corresponding keyboard shortcut (i
 | Goto Type Definition | unbound | `lsp_symbol_type_definition`
 | Hover Popup | unbound | `lsp_hover`
 | Insert/Replace Completions | <kbd>Alt</kbd> <kbd>Enter</kbd> | `lsp_commit_completion_with_opposite_insert_mode`
-| Next Diagnostic | unbound | `lsp_next_diagnostic`
-| Previous Diagnostic | unbound | `lsp_prev_diagnostic`
+| Next Diagnostic | unbound | `lsp_next_diagnostic`<br>Supports optional argument `"severity_level"`
```
predragnikolic, line 74: Sould we copy/paste the the description as well?


> "severity_level"` for the severity level of included diagnostics: 1 (error), 2 (warning), 3 (information), 4 (hint). Uses the `"show_diagnostics_severity_level"` setting if not provided.

jwortmann: I thought about that, but wanted to avoid multiple linebreaks because there isn't much space in the table and maybe it's not necessary to repeat all that info again. Or maybe we should add `(see "Goto Diagnostic")` instead?

predragnikolic: Maybe we can move the lsp_next/prev_diagnostic row closer to the "goto diagnostics" row? That way it might be likely that they will spot the explanation for the severity level.

jwortmann: The table was ordered alphabeticall, but that's probably not so important, so I moved them closer together now.

predragnikolic (approved):
Tested ✔️ 
and the code is nice

PR #2987, opened 2026-07-13
feat: show progress when requesting code actions manually

rwols:
Some servers take quite some time to return code actions (JETLS for example takes ~2 seconds for me).

When requesting such code actions from the command palette, it can be a little confusing to understand what is happening, as there is no feedback given when the code action results take ~2 seconds to compute.

This change shows the usual progress notification in the status bar when it's taking more than 200ms.

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -528,7 +528,7 @@ class LspMenuActionCommand(LspWindowCommand, ABC):
         if not view:
             return
         if (region := self._get_region(event)) is not None:
-            actions_manager.request_for_region_async(view, region, [], MENU_ACTIONS_KINDS, manual=True, progress=True)
+            actions_manager.request_for_region_async(view, region, [], MENU_ACTIONS_KINDS, manual=True)
 
 
 class LspRefactorCommand(LspMenuActionCommand):
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -98,7 +98,9 @@ class CodeActionsManager:
         region: sublime.Region,
         session_buffer_diagnostics: list[tuple[SessionBufferProtocol, list[Diagnostic]]],
         only_kinds: list[str | CodeActionKind] | None = None,
+        *,
         manual: bool = False,
+        progress: bool = False,
     ) -> Promise[list[CodeActionsByConfigName]]:
         """
         Requests code actions with provided diagnostics and specified region. If there are
@@ -129,7 +131,7 @@ class CodeActionsManager:
                     diagnostics = diags
                     break
             params = text_document_code_action_params(view, region, diagnostics, only_kinds, manual)
-            return Request.codeAction(params, view)
+            return Request.codeAction(params, view, progress=progress)
 
         def response_filter(sb: SessionBufferProtocol, actions: list[CodeActionOrCommand]) -> list[CodeActionOrCommand]:
             # Filter out non "quickfix" code actions unless "only_kinds" is provided.
@@ -397,9 +399,9 @@ class LspCodeActionsCommand(LspTextCommand):
         if not listener:
             return
         session_buffer_diagnostics = listener.get_diagnostics_async(region)
-        actions_manager \
-            .request_for_region_async(view, region, session_buffer_diagnostics, only_kinds, manual=True) \
-            .then(lambda actions: sublime.set_timeout(lambda: self._handle_code_actions(actions)))
+        actions_manager.request_for_region_async(
+            view, region, session_buffer_diagnostics, only_kinds, manual=True, progress=True
+        ).then(lambda actions: sublime.set_timeout(lambda: self._handle_code_actions(actions)))
 
     def _handle_code_actions(self, response: list[CodeActionsByConfigName], run_first: bool = False) -> None:
         # Flatten response to a list of (config_name, code_action) tuples.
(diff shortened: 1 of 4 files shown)


jwortmann:
There is nothing wrong with this PR, but I just wonder do you know any server which does progress reporting based on our client-initiated progress via `workDoneToken`?

JETLS takes around 5ms for code action requests on my side, and I think what you're seeing is compile time latency, but that happens only on the very first request (for each request type). Maybe it's not worth to trigger the
https://github.com/sublimelsp/LSP/blob/4e3828b9dc829977365f58454c7a34c94583a409/plugin/core/active_request.py#L48
on every code action request just for that, unless there is a server which actually uses the progress reporting for code action requests? Or otherwise should we maybe just activate work done progress on all requests by default, because as mentioned compilation with JETLS happens for every request type. I see that currently it is only active by default on document symbols, workspace symbols, formatting and prepare call/type hierarchy. All of those requests are triggered only on demand via user action, but code action requests can also happen automatically on typing or cursor position change. So perhaps for efficiency we should limit that only to requests which must be triggered manually?

Or perhaps better to only activate progress if the request was triggered manually (via command palette or context menu or key binding), but not when triggered automatically?

rwols:
> There is nothing wrong with this PR, but I just wonder do you know any server which does progress reporting based on our client-initiated progress via `workDoneToken`?

- Off the top of my head, I remember pyright supporting this: https://github.com/microsoft/pyright/issues/2545.
- clojure-lsp supposedly handles `workDoneToken` only for the `initialize` request: https://github.com/clojure-lsp/clojure-lsp/commit/531e508f8532628a5ab41812a704ca5fec8be923. It can take a long time to parse class paths and jars, apparently.
- gopls handles `workDoneToken` only for the `workspace/executeCommand` request: https://github.com/golang/go/issues/40527. [Here's the exact line in the source code](https://github.com/golang/tools/blob/b0dc4aefbf2ea447be555074f0bec76ccade8e2d/gopls/internal/server/command.go#L418). And [here's the function that starts the progress reporter, falling back to server-initiated progress if there's no token](https://github.com/golang/tools/blob/b0dc4aefbf2ea447be555074f0bec76ccade8e2d/gopls/internal/progress/progress.go#L84). I'm not experienced enough with the Go language to know which commands gopls supports.

There doesn't seem to be much more... The trend seems to be that most language server authors opt for server-initiated progress (which I find less elegant, IMO, as client-initiated progress has the ability to restrict to a specific view/buffer, compared to server-initiated progress which always reports "for the whole window").

Whether the server supports `workDoneToken` or not, when setting `progress=True` in the `Request` we show some kind of hint in the status bar regardless.

> JETLS takes around 5ms for code action requests on my side, and I think what you're seeing is compile time latency, but that happens only on the very first request (for each request type). Maybe it's not worth to trigger the `sublime.set_timeout_async(show, 200)` on every code action request just for that, unless there is a server which actually uses the progress reporting for code action requests?

The JETLS doesn't seem to cache this compilation for me. It always takes a long time to resolve the code action request (and other requests as well). Perhaps there's a setting I'm missing?

Regardless, there's other servers out there that are also not quite that rapid, for instance jdtls.

> Or otherwise should we maybe just activate work done progress on all requests by default, because as mentioned compilation with JETLS happens for every request type. I see that currently it is only active by default on document symbols, workspace symbols, formatting and prepare call/type hierarchy. All of those requests are triggered only on demand via user action, but code action requests can also happen automatically on typing or cursor position change. So perhaps for efficiency we should limit that only to requests which must be triggered manually?
> Or perhaps better to only activate progress if the request was triggered manually (via command palette or context menu or key binding), but not when triggered automatically?

Yes, I totally agree that:

- "background requests" (like code lenses, pull diagnostics, semantic tokens) should not display progress.
- "foreground", or user-triggered, or manually-triggered requests should display progress.

Code actions are also requested in the background. The background code actions don't display progress. The background code actions go through the `SessionBuffer.request_code_actions_async` path, which sets `progress=False`.

The foreground code actions go through the `CodeActionsManager`, which does set `progress=True`.

(I'm not sure why `CodeActionsManager` isn't used in `SessionBuffer` and related classes).

jwortmann:
> Code actions are also requested in the background. The background code actions don't display progress. The background code actions go through the `SessionBuffer.request_code_actions_async` path, which sets `progress=False`.

Oh, okay I think I missed that when I took a quick look. Then it should be fine to have progress active in that case.

> (I'm not sure why `CodeActionsManager` isn't used in `SessionBuffer` and related classes).

I think CodeActionsManager always requests code actions from all available sessions. Which is useful when triggered manually. But iirc code actions are also updated automatically when new diagnostics arrive, and in that case we only want to send the request to that particular session, which is done in the method in SessionBuffer.

plugin/code_actions.py
```diff
@@ -526,7 +528,7 @@ def _request_menu_actions_async(self, event: dict | None) -> None:
         if not view:
             return
         if (region := self._get_region(event)) is not None:
-            actions_manager.request_for_region_async(view, region, [], MENU_ACTIONS_KINDS, True)
+            actions_manager.request_for_region_async(view, region, [], MENU_ACTIONS_KINDS, manual=True, progress=True)
```
jwortmann: Do we want progress active here too? I would say this request is also kind of in the background, because it happens already when you hover over the "Edit" dropdown menu. The implementation is a bit hacky with an invisible dummy item
https://github.com/sublimelsp/LSP/blob/4e3828b9dc829977365f58454c7a34c94583a409/Main.sublime-menu#L26-L29
that triggers the request. Then when you move the mouse over "LSP: Source Action" or "LSP: Refactor" the responses should already have been arrived, so that the available actions are displayed immediately.

So in this case I would set `progress=False`, because users probably don't need to see progress notification in the status bar just from hovering over "Edit".

jwortmann (approved):

PR #2990, opened 2026-07-15
docs: update LspPlugin migration documentation

rwols:


rchl:
This is a migration guide for migrating from AbstractPlugin to LspPlugin. We shouldn't just overwrite it with asyncio version since the original LspPlugin implementation was already released and some plugins were already converted. We should instead add another section for migrating from non-asyncio to asyncio LspPlugin that details the changes between those two variants.

You could refactor this page to have two sections like:
- Migrate from AbstractPlugin to LspPlugin
- Migrate from LspPlugin (non-asyncio) to LspPlugin (asyncio)

Then existing LspPlugin implementations would just follow the second section while AbstractPlugin implementations would follow both, in order.

EDIT: Not saying this will be an ideal approach but it would be the most correct with least amount of maintenance required. More sophisticated approach could be to have a dynamic page that allows the user to select the target and modify the page accordingly but this would probably take too much effort.

docs/src/migrating_to_lsp_plugin.md
```diff
@@ -14,22 +14,23 @@
 | `name()` | Removed - derived automatically from the package name and exposed as a `name` property |
 | `configuration()` | Removed - settings file located automatically |
 | `storage_path()` | `ST_STORAGE_PATH` constant or `plugin_storage_path` class attribute |
-| `needs_update_or_installation()` + `install_or_update()` + `can_start()` + `on_pre_start()` + `additional_variables()` | `on_pre_start_async(context)` |
+| `needs_update_or_installation()` + `install_or_update()` + `can_start()` + `on_pre_start()` + `additional_variables()` | `(async) on_pre_start(context)` |
```
rchl: When you mention `(async)`, I thik it would make sense to just write `async ...` to make it clearer. So just `async on_pre_start(context)` in this case.

docs/src/migrating_to_lsp_plugin.md
```diff
@@ -167,9 +168,12 @@ from LSP.plugin import OnPreStartContext
 from LSP.plugin import PluginStartError
 
 @classmethod
-def on_pre_start_async(cls, context: OnPreStartContext) -> None:
+async def on_pre_start(cls, context: OnPreStartContext) -> None:
     if not server_binary().exists():
-        download_server(server_binary())
+        await download_server(server_zipfile())
+        await run_in_threadpool(unzip_server_zipfile)
```
rchl: When using `run_in_threadpool` in the example, also include the import line for it.

PR #2991, opened 2026-07-15
chore: enhance logging functions

rwols:
I'm porting some self-contained bits from the `feat/asyncio` branch so that PR becomes a bit smaller.

### DIFF
--- a/plugin/core/logging.py
+++ b/plugin/core/logging.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 from .constants import ST_PACKAGES_PATH
 from typing import Any
 import inspect
-import threading
 import traceback
 
 log_debug = False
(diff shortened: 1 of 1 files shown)


plugin/core/logging.py
```diff
@@ -19,23 +20,33 @@ def debug(*args: Any) -> None:
...
         return
     previous_frame = current_frame.f_back
     file_name, line_number, function_name, _, _ = inspect.getframeinfo(previous_frame)  # type: ignore
-    file_name = file_name[len(ST_PACKAGES_PATH) + len("/LSP/"):]
-    debug(f"TRACE {function_name:<32} {file_name}:{line_number}")
+    file_name = file_name[len(ST_PACKAGES_PATH) + len("/LSP/") :]
+    debug(f"TRACE {threading.current_thread().name:<16} {function_name:<32} {file_name}:{line_number}")
+    if print_callstack:
+        for frame in traceback.extract_stack():
+            debug(f"TRACE {frame.filename}:{frame.lineno} in {frame.name}")
```
rchl: The stack trace should already be available in inspected frames so we shouldn't need to call `traceback.extract_stack()`. I had some old code in stash like this to iterate frames:

```py
def trace(depth: int = 1) -> None:
    current_frame = inspect.currentframe()
    if current_frame is None:
        debug("TRACE (unknown frame)")
        return
    output: list[str] = []
    while depth > 0:
        depth -= 1
        if current_frame := current_frame.f_back:
            file_name, line_number, function_name, _, _ = inspect.getframeinfo(current_frame)
            file_name = file_name[len(ST_PACKAGES_PATH) + len("/LSP/"):]
            output.append(f'  {function_name:<32} {file_name}:{line_number}')
        else:
            break
    print("TRACE\n" + '\n'.join(output))
```

Don't need to use it exactly but would still prefer some variation of that so that we iterate already available frames.

rchl (approved):

PR #2993, opened 2026-07-16
Ensure selection start is retained when range formatting

### LINKED ISSUE #2986: `lsp_format_document_range` selection is broken after formating
**Describe the bug**

Before the shortcut
Image

After `lsp_format_document_range`

Image

**To Reproduce**

```python
def f():
    records = self.env['documents.document'].search_read(domain, search_panel_fields)
    alias_tag_data = {}
    if not self.env.user.share:
        alias_tag_ids = {alias_tag_id for rec in records for alias_tag_id in rec['alias_tag_ids']}
        alias_tag_data = {
            alias_tag['id']: {
                'id': alias_tag.id,
                'color': alias_tag.color,
                'display_name': alias_tag.display_name
            } for alias_tag in self.env['documents.tag'].browse(alias_tag_ids)
        }
```

Select from `records` to the end of the function, and lsp format with ruff

jwortmann:
Fixes #2986 

Workaround for what I would call a bug in the Ruff formatter, which modifies content outside of the range to format.

### DIFF
--- a/plugin/formatting.py
+++ b/plugin/formatting.py
@@ -207,21 +207,30 @@ class LspFormatDocumentRangeCommand(LspTextCommand):
             listener.purge_changes_async()
         if has_single_nonempty_selection(self.view):
             session = self.best_session(self.capability)
-            selection = first_selection_region(self.view)
-            if session and selection is not None:
-                request = text_document_range_formatting(self.view, selection)
-                session.send_request_task(request).then(self._handle_response_async)
+            selection_region = first_selection_region(self.view)
+            if session and selection_region is not None:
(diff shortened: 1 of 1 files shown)


rwols:
What effect does this have on other formatters? Have you test-driven a few?

jwortmann:
No, I have only tested with Ruff. But unless I missed anything important, I think it should be pretty safe, given that the selection is only modified here if its start point does not match the expected start point anymore.

rwols:
Let me try with clangd, jdtls, and Roslyn first

rwols (approved):

PR #2994, opened 2026-07-17
chore: make promise adhere to the typing.Awaitable protocol

rwols:
I'm porting some self-contained bits from the feat/asyncio branch so that PR becomes a bit smaller.


rchl:
I think this code shouldn't be separate from asyncio task because it's not really usable without the rest of the asyncio changes.

PR #3000, opened 2026-07-24
Add Svelte language identifier

dontwanttothink:
Adding this language identifier is necessary for Biome to work with Svelte files correctly.

See: https://github.com/sublimelsp/LSP-biome/issues/92


predragnikolic (approved):

jwortmann:
Please put that entry in alphabetical order into that dict.

PR #3003, opened 2026-07-28
docs: section in migration guide about async LspPlugin

rwols:
note: I accidentally force-pushed changes that included no changes, for some reason that made GitHub drop it.


docs/src/migrating_to_lsp_plugin.md
```diff
@@ -470,3 +470,240 @@ from LSP.plugin import request_handler
...
+
+The `on_pre_start_async` method of LspPlugin is supposed to download and unzip a language server binary, or it may have to run NPM, uv, or some other package manager tool for installing the language server files. The functions needed for these operations were all blocking:
+
+- Downloading was usually done with the [`urllib.request.urlretrieve`](https://docs.python.org/3/library/urllib.request.html#urllib.request.urlretrieve). function.
+- Unzipping was usually done with the [`zipfile`](https://docs.python.org/3/library/zipfile.html) module. Or perhaps [`tarfile`](https://docs.python.org/3/library/tarfile.html).
+- Running a tool was usually done with the [`subprocess`](https://docs.python.org/3/library/subprocess.html) module.
+
+There exist asynchronous counterparts to these functions, but they're not trivial to replace. Here is a non-exhaustive replacement list:
+
+- Downloading a file asynchronously can be done with the [httpx](https://www.python-httpx.org) third-party package. Package Control can download this library.
```
rchl: I'm not sure if this suggestion is accurate but I would write something more useful like that.

```suggestion
- Downloading a file asynchronously can be done with the [httpx](https://www.python-httpx.org) library. Add `httpx` to package dependencies to use it.
```

rchl (commented):
This structure doesn't really work with Zensical. You can create a preview and check (I guess you need to mark as ready for review). The TOC on the right only sees the first header.

Not sure off-hand how this should be done instead, would have to experiment.

EDIT: Probably easiest to have one first-level header and then separate second-level headers for each of the two migrations.

docs/src/migrating_to_lsp_plugin.md
rwols: httpx isn't even available. I replaced the suggestion with aiosonic.

PR #3004, opened 2026-08-02
Make LspTextCommand inherit from sublime_aio.ViewCommand

rwols:
This PR makes LspTextCommand inherit from sublime_aio.ViewCommand. The benefit is that this ViewCommand has special handling for `async def run()` methods. So, you can write directly such a method, without having to use the somewhat cumbersome pattern of defining an `async def _run()` method and invoking that one, via `run_coroutine`, in the regular `run()` method. So:

```py
# OLD
FooBarCommand(LspTextcommand):
    def run(self, edit: sublime.Edit, foo: str, bar: str) -> None:
        run_coroutine(self._run(foo, bar))

    async def _run(self, foo: str, bar: str) -> None:
        ...
```

```py
# NEW
FooBarCommand(LspTextCommand):
    async def run(self, foo: str, bar: str) -> None:
        ...
```

_However_, I see TypeErrors being thrown for old LspTextCommand-derived classes that don't have an async `run()` method. This PR is blocked on that: https://github.com/packagecontrol/sublime_aio/issues/34

### DIFF
--- a/plugin/code_actions.py
+++ b/plugin/code_actions.py
@@ -353,9 +353,8 @@ class LspCodeActionsCommand(LspTextCommand):
             return self.is_enabled(event, point)
         return True
 
-    def run(
+    async def run(
         self,
-        edit: sublime.Edit,
         event: dict | None = None,
         only_kinds: list[str | CodeActionKind] | None = None,
         code_actions_by_config: list[CodeActionsByConfigName] | None = None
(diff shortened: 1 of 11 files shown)


jwortmann:
I was thinking about whether we could make a command wrapper that would allow to report back a result when (asynchronically) running a command. Then we could use that in a way like
```python
result = await run_command(view, 'lsp_some_command', {})
```
instead of the regular ST API call (which always returns `None`)
```python
view.run_command('lsp_some_command', {})
```

This could be useful for example when applying TextEdits or a WorkspaceEdit (e.g. LspApplyWorkspaceEditCommand), where it would be nice to know whether an operation failed, or to get back the number of applied edits and edited files.

PR #3012, opened 2026-08-15
Fix handling of unchanged diagnostics report

jwortmann:
I noticed that [UnchangedDocumentDiagnosticReport](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#unchangedDocumentDiagnosticReport) is not handled correctly. Unchanged diagnostics report means that the previous diagnostics (and as far as I understand, including the diagnostic's `range`) are still accurate. Currently we do "nothing" when we receive such a report:
https://github.com/sublimelsp/LSP/blob/bef834e8a09e3b083c62a3969721aec9c371cdff/plugin/session_buffer.py#L674-L676

But actually we need to redraw diagnostics regions in the view, because diagnostics are typically requested after a buffer change, and ST automatically adjusts/shifts existing region decorations if lines before the region we removed or added, or characters were inserted or removed on the same line before the region.

If lines or characters were inserted or removed *after* the diagnostics regions it didn't matter, because in that case the drawn regions are not affected anyway.

I added another `None` value as possible argument to `Session.handle_diagnostics_async`, which means that we received unchanged diagnostics, i.e. stored diagnostics don't need to be updated, but we still need to trigger a redraw.

### DIFF
--- a/plugin/core/sessions.py
+++ b/plugin/core/sessions.py
@@ -2223,10 +2223,11 @@ class Session(APIHandler, TransportCallbacks):
             debug("ignoring unsuitable diagnostics for", uri, "reason:", reason)
             return
         if diagnostics is not None:
-            # `None` means we received an UnchangedDocumentDiagnosticReport, in which case we still have to redraw
-            # diagnostic regions in the view to maintain the original positions.
             self.diagnostics.set_diagnostics(uri, identifier, diagnostics)
             mgr.on_diagnostics_updated()
+        # Even if we received an UnchangedDocumentDiagnosticReport (represented by the diagnostics argument being None)
+        # we still have to redraw the diagnostic regions in the view to ensure they keep their original positions after
+        # a buffer change.
         if session_buffer := self.get_session_buffer_for_uri_async(uri):
             self._publish_diagnostics_to_session_buffer_async(
                 session_buffer, self.diagnostics.get_diagnostics_for_uri(uri), version)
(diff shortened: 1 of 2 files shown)


plugin/core/sessions.py
```diff
@@ -2218,8 +2222,11 @@ def handle_diagnostics_async(
         if isinstance(reason, str):
             debug("ignoring unsuitable diagnostics for", uri, "reason:", reason)
             return
-        self.diagnostics.set_diagnostics(uri, identifier, diagnostics)
-        mgr.on_diagnostics_updated()
+        if diagnostics is not None:
+            # `None` means we received an UnchangedDocumentDiagnosticReport, in which case we still have to redraw
+            # diagnostic regions in the view to maintain the original positions.
+            self.diagnostics.set_diagnostics(uri, identifier, diagnostics)
+            mgr.on_diagnostics_updated()
```
rchl, line 2227: I'm coming here without refreshing my memory on how diagnostics rendering works but just reading that comment makes me a bit confused because it talks about still having to handle the `None` case but being inside the `!None` branch. Is it misplaced?

jwortmann: Yeah maybe it's a bit misplaced, I think I put it here because these two lines are the only part of this method which is affected by diagnostics being `None` or `not None`. And there is no `else` branch. I could put the comment after this if-block, if that makes it clearer?

rchl: I think it would.

jwortmann: Done in https://github.com/sublimelsp/LSP/pull/3012/changes/fed82d0481675c7967efdc9050450a140ef6cfa3 and tried to make the wording a bit more clear.

rchl (approved):

PR #3023, opened 2026-09-06
Docs: add instructions for Slang language server

jwortmann:
1. Add instructions for Slang language server. I noticed that there is also a helper package for this server at https://github.com/LDAP/LSP-slangd, but that package is not on Package Control, uses the old `AbstractPlugin`, installs an older version of the language server, and it looks like it doesn't provide any additional functionalities. So for now we can add the manual config.
2. Removed zero-width space in YamlPipelines name (I assume that was accidental).
3. Replaced old https://packagecontrol.io/ links with the new Package Control website https://packages.sublimetext.io/

### DIFF
--- a/docs/src/language_servers.md
+++ b/docs/src/language_servers.md
@@ -265,7 +265,7 @@ Follow installation instructions on [LSP-graphql](https://github.com/sublimelsp/
 ## Helm
 
 1. Install [helm-ls](https://github.com/mrjosh/helm-ls).
-2. Install the [Yaml​Pipelines](https://packagecontrol.io/packages/YamlPipelines) package from Package Control for syntax highlighting.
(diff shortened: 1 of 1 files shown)


rwols (approved):
LGTM. I would prefer these commits to be rebased instead of being squash-merged. I think changing the links deserves a separate atomic change, as well as the removal of the non-breaking space characters. But up to you.

rwols:
By the way… I once again encourage you to join the Discord server. It helps with alignment on where things go.

jwortmann:
> I would prefer these commits to be rebased instead of being squash-merged.

Sure, cleaned up the commits a bit for that now.

> By the way… I once again encourage you to join the Discord server. It helps with alignment on where things go.

Okay, I'll consider it... ;) but no guarantee that I will use it regularly.

PR #3028, opened 2026-09-14
Defined language identifiers (source.gjs and source.gts)

ijlee2:
## Background

Followed up on the changes made to [sublime-syntax-definition-template-tag](https://github.com/ijlee2/sublime-syntax-definition-template-tag) and [LSP-ember](https://github.com/ijlee2/LSP-ember). These repos provide syntax highlighting and LSP features, respectively, for `*.{gjs,gts}` files (JS/TS files with a `<template>` tag).

Currently, the syntax definitions use the identifiers `source.gjs` and `source.gts`, while LSP-ember needs to send the IDs `glimmer-js` and `glimmer-ts` (defined by the underlying dependency, [@glint/ember-tsc](https://github.com/typed-ember/glint/blob/v1.11.4-%40glint/ember-tsc/packages/ember-tsc/src/volar/ember-language-plugin.ts#L28-L35)).

Related: https://github.com/sublimehq/package_control_channel/pull/9559#issuecomment-5636800355

## References

- https://github.com/ijlee2/sublime-syntax-definition-template-tag/blob/4.0.1/src/Glimmer%20JS.sublime-syntax#L4
- https://github.com/ijlee2/sublime-syntax-definition-template-tag/blob/4.0.1/src/Glimmer%20TS.sublime-syntax#L4
- https://github.com/ijlee2/LSP-ember/blob/0.3.1/language-ids.sublime-settings (file will be removed)


protocol/__init__.py
```diff
@@ -570,6 +570,8 @@ class LanguageKind(StrEnum):
     FSharp = 'fsharp'
     GitCommit = 'git-commit'
     GitRebase = 'rebase'
+    GlimmerJs = 'glimmer-js'
+    GlimmerTs = 'glimmer-ts'
```
jwortmann: This is not part of the official LSP specs (this file is automatically generated).

Please delete this part and put the language ID strings directly as values into the dict in `constants.py`.

ijlee2: Ah, okay. Done.

(I didn't check if there are test files to be updated.)

jwortmann (approved):

